PR #827, opened 2022-12-01
feat(sdk): gcp bucket

### LINKED ISSUE #607: cloud.Bucket - GCP implementation

### LINKED ISSUE #675: Google Cloud Platform (GCP)
### Click 👍 to let us know you care about this
Tell us as much as possible about your use case in the comments below.

Chriscbr:
Closes #607
Closes #675

Introduces the Terraform GCP compilation target, as well as its first resource, the bucket.

Other changes:
- Refactors the internal `CdktfApp` class to resolve some issues where properties can't be added to subclasses
- Introduces some random name generation in the bucket implementation to support the requirement that resources have a globally unique ID.
- Ran `eslint` command in the SDK to clean up imports etc., and added the command to the SDK's build workflow (I think it was in the workflow in the past, but it accidentally was removed at some point)

TODO:
- [x] add tests
- [x] add "tf-gcp" target to CLI
- [ ] ~inflight client~ (skipped)
- [x] "public" prop
- [x] update CLI docs with `tf-gcp` target notes


staycoolcall911:
Keep

libs/wingsdk/src/target-tf-gcp/app.ts
```diff
@@ -0,0 +1,24 @@
+import { GoogleProvider } from "@cdktf/provider-google/lib/provider";
+import { Polycons } from "polycons";
+import { AppProps, CdktfApp, IApp } from "../core";
+import { PolyconFactory } from "./factory";
+
+export interface GcpAppProps extends AppProps {
+  readonly projectId?: string;
+}
```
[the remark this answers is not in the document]
Chriscbr: There doesn't seem to be a good way to infer the region automatically in terraform that I could find. So I've added `storageLocation` as a required field (which can be passed through an environment variable)

Chriscbr:
This change set is large as it is, so I'll create a separate issue for implementing the Bucket client in another PR.

libs/wingsdk/test/target-tf-gcp/bucket.test.ts
```diff
@@ -0,0 +1,40 @@
+import * as cloud from "../../src/cloud";
+import * as tfgcp from "../../src/target-tf-gcp";
+import { mkdtemp } from "../../src/util";
+import { tfResourcesOf, tfSanitize, treeJsonOf } from "../util";
+
+test("create a bucket", () => {
```
hasanaburayyan, line 6: let's add a multiple bucket test, since we are adding it with other providers (in PRs) I think it makes sense to add one here as well, just for sanity.

libs/wingsdk/src/target-tf-gcp/app.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * The Google Cloud project ID.
+   */
+  readonly projectId: string;
+
+  /**
+   * The Google Cloud storage location, used for all storage resources.
+   * @see https://cloud.google.com/storage/docs/locations
+   */
+  readonly storageLocation: string;
```
hasanaburayyan, line 20: Just curious, do you imagine we will have a separate location for other resources types? It looks like other resources also take a location, but from what I see initially the location values are not uniform.

It looks like the storage locations could potentially be used with queues https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_tasks_queue

But then there is something like kms key rings which use their own location list (shown by `gcloud kms locations list`)
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/kms_key_ring

Yea not really a comment more an observation, interesting to see different dynamics in the clouds. 😅

hasanaburayyan (commented):
🎉 Yay GCP, Im excited to get my hands on the changes with InnerApp

libs/wingsdk/src/target-tf-gcp/app.ts
Chriscbr: Yep, exactly. That's what I inferred from reading the GCP docs as well -  it seems like different resources have very different lists of regions. (I never thought I'd appreciate AWS regions before today 😅).

It's possible we can make an enum with a list of all of the available locations, and indicate which ones support which GCP resources etc., but I thought we can start with `storageLocation` and iterate from there.

libs/wingsdk/test/target-tf-gcp/bucket.test.ts
Chriscbr: Good idea, adding...

hasanaburayyan (approved):

PR #832, opened 2022-12-03
feat(sdk): tf-azure bucket implementation

hasanaburayyan:
WIP adding tfazure (bucket)

TODO:
- [x] add tests
- [x] add "tf-gcp" target to CLI
- [x] "location" prop from App level


libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+    // Consider cleaner ways to check if construct is instance of App.
+    const app = construct as App;
+    return app.appType === "TF_AZURE_APP" ? app : App.of(construct.node.scope);
+  }
+
+  /** The app type */
+  public readonly appType = "TF_AZURE_APP";
+  /** The apps props */
+  public readonly props: AzureAppProps;
```
Chriscbr, line 38: Let's expose the individual properties directly (like `location`). I think this reduces the amount of indirection if users want to access the information in their app code, since they can access e.g. `app.location` instead of `app.props.location`.

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -0,0 +1,122 @@
...
+import { BlobServiceClient } from "@azure/storage-blob";
+import { BucketDeleteOptions, IBucketClient } from "../cloud";
+
+export class BucketClient implements IBucketClient {
+  private readonly bucketName: string;
+  private readonly blobServiceClient: BlobServiceClient;
+  private readonly defaultAzureCredential: DefaultAzureCredential =
+    new DefaultAzureCredential();
+
+  constructor(bucketName: string, storageAccount: string) {
```
Chriscbr, line 11: If you run into issues trying to unit test this, one strategy you could try is adding an optional third parameter for passing a blobServiceClient instance. In the jest test, you would pass in a jest mock object that just implements or stubs out the methods you need. (Maybe there's a more elegant way to do this? 🤷‍♂️)

(only suggesting because I know jest can be a pain sometimes... ping me if you have any questions 😄)

Chriscbr (commented):
Looks fantastic! I added a couple ideas/suggestions.

Feel free to hit the "re-request review" button on GitHub (under the reviewers list) when it's ready for another pass 👍

Btw, we'll likely want to add azure as a compilation target in our integration tests suite (in `tools/hangar`) in a future issue/pr

libs/wingsdk/src/target-tf-azure/bucket.ts
```diff
@@ -0,0 +1,83 @@
...
+  constructor(scope: Construct, id: string, props: cloud.BucketProps) {
+    super(scope, id, props);
+
+    this.public = props.public ?? false;
+
+    const app = App.of(this);
+
+    if (app === undefined) {
+      throw new Error("Unable to find app in scope");
+    }
```
Chriscbr, line 29: `App.of` should throw if an app can't be found instead of returning undefined (fail fast by default). Users can always try/catch in the exceptional case that they want to do something if it fails

libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -0,0 +1,55 @@
...
+  /** The location context of the App */
+  public readonly location: string;
+
+  constructor(props: AzureAppProps) {
+    super({
+      ...props,
+      customFactory: props.customFactory ?? new PolyconFactory(),
+    });
+    this.location = props.location;
+    // Polycons.register(this, props.customFactory ?? new PolyconFactory());
```
Chriscbr, line 46: Remove

libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -0,0 +1,55 @@
...
+// import { Polycons } from "polycons";
+import { IApp, CdktfApp, AppProps } from "../core";
+import { PolyconFactory } from "./factory";
+import { APP_AZURE_TF_SYMBOL } from "./internal";
+
+/**
+ * Azure app props
+ */
+export interface AzureAppProps extends AppProps {
+  /** Location for resources to be deployed to */
```
Chriscbr, line 12: Are there any docs listing the possible locations that we could link to in the docstring?

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -0,0 +1,128 @@
...
+      this.bucketName
+    );
+    await containerClient.getBlockBlobClient(key).upload(body, body.length);
+  }
+
+  /**
+   * Get object content from bucket
+   *
+   * @param key Key of the object
+   * @returns string contnet of the object as string
```
Chriscbr, line 42:
```suggestion
   * @returns string content of the object as string
```

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -0,0 +1,128 @@
...
+    return downloaded;
+  }
+
+  /**
+   * List all keys in the bucket
+   *
+   * @param prefix Limits the response to keys that begin with the specified prefix
+   * TODO - add pagination support
+   */
+  public async list(prefix?: string | undefined): Promise<string[]> {
```
Chriscbr, line 68: `prefix: string | undefined` and `prefix?: string` have pretty much the same effect, so no need to go double-optional
```suggestion
  public async list(prefix?: string): Promise<string[]> {
```

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -0,0 +1,128 @@
...
+   * Delete an existing object using a key from the bucket
+   *
+   * @param key Key of the object
+   * @param opts Option object supporting additional strategies to delete item from a bucket
+   */
+  public async delete(
+    key: string,
+    opts?: BucketDeleteOptions | undefined
+  ): Promise<void> {
+    const mustExist = opts?.mustExist ?? false;
```
Chriscbr, line 92: nit
```suggestion
    opts: BucketDeleteOptions = {}
  ): Promise<void> {
    const mustExist = opts.mustExist ?? false;
```

libs/wingsdk/test/target-tf-azure/bucket.inflight.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    STORAGE_NAME,
+    mockBlobServiceClient
+  );
+  const response = await client.put(KEY, VALUE);
+
+  // THEN
+  expect(response).toEqual(undefined);
+});
+
+test("delete object from a bucket", async () => {
```
Chriscbr, line 63: Unit test for list?

Chriscbr (approved):
Awesome!! Nice error handling and everything. 👍

I added a couple small comments but feel free to merge after addressing them

staycoolcall911:
Congratulations, @hasanaburayyan, this is great!

.github/workflows/build.yml
```diff
@@ -250,7 +250,7 @@ jobs:
         runner:
           - ubuntu-latest
           # - macos-latest
-        node: [18, 19]
+        node: [18]
```
staycoolcall911, line 253: Let's reinstate our support of Node 19 - since sooner or later this is going to break if we stop testing it. Both @Chriscbr and @MarkMcCulloh offered to help you with that and I'm happy to assist too if this gets complicated

staycoolcall911 (commented):
First of all - congrats! What a great achievement - our first ever Azure impl!

Reviewed after it was already merged, so maybe just open a new PR with the fixes? Thank you!

.github/workflows/build.yml
staycoolcall911: Or actually if this gets complicated beyond Mark and Chris's skills I probably won't be able to help :)
But I'll find someone who can in that case...

hasanaburayyan: Going to put a PR up for it in just a moment. I think we can just update the cli.test.ts to ignore engines. testing now.

hasanaburayyan: @staycoolcall911 you will find the e2e re-enabled and above comments addressed in: https://github.com/winglang/wing/pull/1087

PR #899, opened 2022-12-14
fix(compiler): when array lit elements are an instance of something type infer the array type correctly

yoav-steinberg:
Fix compiler error on this
```wing
bring cloud;

let bucket1 = new cloud.Bucket() as "b1";
let bucket2 = new cloud.Bucket() as "b2";
let bucket3 = new cloud.Bucket() as "b3";
let a:Array<cloud.Bucket> = [bucket1, bucket2, bucket3]; // This was broken
let b = Array<cloud.Bucket>[bucket1, bucket2, bucket3]; // This too was broken
```
Missing:
- [ ]  apply same logic to all custom type and container literals (structs, set, map).
- [ ]  tests.

*By submitting this pull request, I confirm that my contribution is made under the terms of the 
[Monada Contribution License](https://docs.winglang.io/terms-and-policies/contribution-license.html)*.

PR #1002, opened 2022-12-29
fix(compiler): refactor type checker to distinguish between identifier kinds and actual types

### LINKED ISSUE #996: Unable to assign an enum value to a typed variable
### I tried this

```js
enum Foo {A,B,C}

let x: Foo = Foo.B;
```

### Expected

This should compile successfully

### Instead, this happened

```
Error at code.w:3:14 | Expected type "Foo code.w:1:6", but got "enum value Foo code.w:1:6.B code.w:1:13" instead
```

(notice also that the file type name seems to include the file location, so instead of saying `Foo.B`, it says `Foo code.w:1:6.B`).

### Component

Compiler

### Environment

```markdown
- Wing Version:
- OS:
- Node version:
```

### Anything else?

_No response_

yoav-steinberg:
Now the type environment (type_env.rs) contains a hierarchy of identifier kinds (custom types, variable or field names, namespaces), while the the custom types and variable kinds also reference a type which is the actual type information for that type/variable.
The actual type information is also associated with expressions.
This eliminates all the pseudo types for instances which were stored in the `type_check::Type` enum.
This also eliminates the possibility that something without a name (like `Type::String`) will be stored in the env.

I also took the opportunity to change some typing panics to correct error reporting and made real bugs a bit more verbose using `expect(...)`.

I also created a generic for our unsafe (cyclic) pointers used for referencing types and the environment hierarchy (see `UnsafeRef`). In this generic I utilized Rust's `Deref` trait to make it easier to work with and removed all the `.into()` calls and `From` trait implementations.

There's also an important bug fix here that's supposed to handle the ambiguity of a type name appearing inside a nested reference expression: `myvar.myfield.method_name().another_field` looks similar to `a.namespace.MyType.static_method().another_field` but in the latter `a.namespace.MyType` isn't really a hierarchy of expressions it's just a type. So we need to carefully handle this in `resolve_reference`. This basically fixes the simple case of `let x = MyEnum.A` without us thinking that `MyEnum` is a variable/instance.

This PR should also (hopefully) resolve #996 and #899.
Update: :point_up: it does solve them, added the tests :heavy_check_mark: 

Update2: following the CR I also added a check preventing instantiation of resources inflight. This isn't related to this PR but was mentioned in the CR and was easy to fix.


libs/wingc/src/jsify.rs
```diff
@@ -604,10 +606,10 @@ impl JSifier {
...
 		};
 
-		let async_prefix = match phase {
-			Phase::Inflight => "async ",
-			_ => "",
-		};
+		// let async_prefix = match phase {
+		// 	Phase::Inflight => "async ",
+		// 	_ => "",
+		// };
```
MarkMcCulloh, line 612: Not too complicated, can probably just be killed

```suggestion
```

libs/wingc/src/type_check.rs
```diff
@@ -104,76 +163,58 @@ impl PartialEq for Type {
...
+					let parent_type: &Type = &*parent;
 					return parent_type.eq(other);
 				}
 				false
 			}
 			(Self::Struct(l0), Self::Struct(_)) => {
 				// If we extend from `other` then treat both structs as equal (inheritance)
 				for parent in l0.extends.iter() {
-					let parent_type: &Type = (*parent).into();
+					let parent_type: &Type = &*parent;
```
MarkMcCulloh, line 183: It is strange to see `&*` everywhere knowing that it's not a no-op.

libs/wingc/src/type_check.rs
```diff
@@ -12,7 +13,70 @@ use type_env::TypeEnv;
...
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		let t: &T = self;
+		write!(f, "{}", t)
+	}
+}
+
+pub type TypeRef = UnsafeRef<Type>;
+
 #[derive(Debug)]
+pub enum IdentKind {
```
MarkMcCulloh, line 51: Wouldn't this technically be `ReferenceKind`?

libs/wingc/src/type_check.rs
```diff
@@ -628,11 +586,16 @@ impl<'a> TypeChecker<'a> {
...
 				// Lookup the type in the env
 				let type_ = self.resolve_type(class, env, statement_idx);
-				let (class_env, class_symbol) = match type_.into() {
-					&Type::Class(ref class) => (&class.env, &class.name),
-					&Type::Resource(ref class) => (&class.env, &class.name), // TODO: don't allow resource instantiation inflight
-					&Type::Anything => return Some(self.types.anything()),
-					_ => panic!("Expected {} to be a resource or class type but it's a {}", class, type_),
+				let (class_env, class_symbol) = match *type_ {
+					Type::Class(ref class) => (&class.env, &class.name),
+					Type::Resource(ref class) => (&class.env, &class.name), // TODO: don't allow resource instantiation inflight
```
MarkMcCulloh, line 591: This TODO is a oneliner right? Just need to check the current env and add an error if it's inflight

libs/wingc/src/type_check.rs
```diff
@@ -1522,55 +1480,94 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	fn expr_maybe_type(&mut self, expr: &Expr, env: &TypeEnv, statement_idx: usize) -> Option<TypeRef> {
+		// TODO: we currently don't handle parenthesized expressions correctly so something like `(MyEnum).A` or `std.(namespace.submodule).A` will return true, is this a problem?
```
MarkMcCulloh, line 1484: I would say this is a problem

libs/wingc/src/type_check.rs
```diff
@@ -12,7 +13,70 @@ use type_env::TypeEnv;
 
 use self::type_env::StatementIdx;
 
+pub struct UnsafeRef<T>(*const T);
```
MarkMcCulloh, line 16: If we keep this up I feel like we're going to reinvent RefCell 😁

MarkMcCulloh (commented):
Super important change, not fully reviewed yet but I like the changes so far

libs/wingc/src/type_check.rs
yoav-steinberg: The underlying issue here is that we discard of parentheses info during the parsing phase (because we can just express that in the tree structure). So the end result is that we have the same AST for `(MyEnum).A` and `MyEnum.A`. I don't think it's a big deal to fix. I think opening a separate issue can be enough for now.

libs/wingc/src/type_check.rs
yoav-steinberg: Of course this isn't a `RefCell`, which is safe. This is more of a curtain hiding the fact that I'm doing risky things under the hood. I wanted this because I thought the `IdentifierKind` might have the same issues as the Type (require cross reference) and I didn't want to duplicate the code. But eventually I didn't need it.
Then I just used the same logic to hide the unsafe `parent` pointer in the `TypeEnv`.

libs/wingc/src/type_check.rs
yoav-steinberg: I believe you're right.

libs/wingc/src/type_check.rs
MarkMcCulloh: Yeah, I totally understand the intent, was mostly a joke 👍

libs/wingc/src/type_check.rs
yoav-steinberg: It's a kind of [Identifier](https://en.wikipedia.org/wiki/Identifier_(computer_languages)) or symbol or (named) reference. They are all kind of interchangeable. We can change the name, I'm aiming for something that'll capture the idea of "something with a name". Currently, btw, we have 3 such things: namesapces, custom types, variables/fields.

libs/wingc/src/type_check.rs
yoav-steinberg: Open for suggestions... I like it being more concise than `into()` and also not giving the feeling that I'm doing some conversion here (because I'm just casting). Might be a way to make this implicit, but for now I think it's better than what we had.

libs/wingc/src/type_check.rs
```diff
@@ -216,177 +267,79 @@ impl Display for Type {
...
-		} else {
-			None
-		}
-	}
-
 	fn maybe_unwrap_option(&self) -> TypeRef {
-		if let &Type::Optional(ref t) = (*self).into() {
-			*t
+		if let Type::Optional(ref t) = **self {
+			t.clone()
```
MarkMcCulloh, line 314: Why clone here? Wouldn't this work still?
```suggestion
			*t
```

libs/wingc/src/type_check.rs
MarkMcCulloh: No I agree that this way is better, no suggestions from me until we revisit the safety of the whole thing

libs/wingc/src/type_check.rs
MarkMcCulloh: It seemed to mostly match up with the Reference concept we already have in the compiler which is why I had the suggestion. Another option would be `SymbolKind` which I think aligns with your framing as well.

MarkMcCulloh (approved):
Other than a possible enum rename and TODO this is looking good to me, nice one.

btw I changed the PR title chore->fix because while this it does change the compiler's behavior.

libs/wingc/src/type_check.rs
yoav-steinberg: I'm more for symbol or identifier because I feel they are more explicit about what this is. For example a namespace's name doesn't feel so much like a reference. 
[Looking at how golang calls this](https://github.com/golang/example/tree/master/gotypes#objects) they call it an "Object" and specifically note:
"_This is different from the everyday meaning of object in object-oriented programming._" I think they were aiming at the subject->object terms used in natural language grammar. The name is the subject and the **object** is the thing. But that felt way too confusing to me.

WDYT about also changing `TypeEnv` / _type_env.rs_ accordingly: `SymbolEnv` / _symbol_env.rs_?

libs/wingc/src/type_check.rs
yoav-steinberg: #1006

libs/wingc/src/type_check.rs
MarkMcCulloh: I am very much in favor of SymbolEnv

libs/wingc/src/jsify.rs
```diff
@@ -604,10 +606,6 @@ impl JSifier {
 			None => ("", "=> "),
 		};
 
-		let async_prefix = match phase {
```
eladb, line 607: Is this intentional

MarkMcCulloh: It's unused

yoav-steinberg: It's dead code. The compiler wars about it. Initially I commented it out but @MarkMcCulloh suggested just removing it since it's no biggy to rewrite when we'll need it.

libs/wingc/src/type_check.rs
yoav-steinberg: done.

Issue #864, opened 2022-12-08
Resource declaration: `this` in constructor and inflight fields


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#33-resources

This was split from #467 to represent what we need for the [TODO app epic](https://github.com/winglang/wing/pull/656).

- [x] support `this` in the constructor
- [ ] #1596

ekeren:
What would be a good use case for inflight fields? 

@eladb/ @Chriscbr ?

Chriscbr:
@ekeren

`queue.approx_size` is an inflight field

ekeren:
> queue.approx_size is an inflight field

@Chriscbr , I am assuming it is a getter that gets calculated on the actual resource.... not an actual QueueClient field 

I was thinking about a use case to save some actual field that store a value on the client

yoav-steinberg:
> I was thinking about a use case to save some actual field that store a value on the client

`queue._cached_approx_size` which isn't updated on each access if the getter is called too frequently.

ekeren:
created a ticket for inflight fields

Issue #542, opened 2022-11-11
Resource declaration: JSify inflight methods


yoav-steinberg:
wingc currently doesn't JSify inflight methods as inflight code (into some JS file to be used as an inflight asset). Instead it just treats them as preflight methods. This makes creating resources with inflight code useless.

Here: https://github.com/winglang/wing/blob/572499b6b20b14bb68daee56474461d10e09c23a/libs/wingc/src/jsify.rs#L359-L386 you can see inflight methods aren't handled specially.

So this:
```
resource R {
  init() {}
  ~inf() {}
}
```
turns into this (notice `inf` is treated as a preflight method):
```js
class R  {
  constructor() {}
  inf() {}
}
```
Which is a bug.
-------------------------------------------------------------
Sub tasks list by @yoav-steinberg:
- [ ] 1. Support resource initialization in wing (access `this` from `init`).
    - [x] 1. At first with no verification that all fields are initialized.  ** p1 **
    - [ ] 2. #1446 ** p2 **
- [ ] 2. JSIfy inflight and preflight for custom resources.
    - [x] 1. First assume all captures are resources and are passed via constructor.  ** p1 **
    - [x] 2. #1519 ** p1 **
    - [ ] 3. #1447 ** p2 **
- [ ] 3. Binding metadata phases (policy metadata):
    - [x] 1. All methods of resource are marked as used.
    - [ ] 2. #76 ** p1 **
    - [ ] 3. Basic CFA (control flow analysis) for generating the binding metadata. Compiler error (unless custom binding metadata is provided). At this phase we don’t allow stuff like variable assignments, passing a resource as an argument, returning a resource… 
    	#1449 ** p1 **
    - [ ] 4. More advanced CFA (this can be multiple stages). ** p2 **
- [ ] 4. Generate custom resources from inflight ~functions~ closures (until this is done we need to manually create resources that adhere to the required wingsdk interface). 
	#1448  ** p2 **
- [ ] 5. Support importing JSII interfaces (so we can understand stuff like IFunctionHandler).
	- [ ] #123  ** p1 **
	- [x] #1024  ** p1 **

staycoolcall911:
Closing since all sub tasks have an issue.

PR #1008, opened 2022-12-30
fix(compiler): support for `this` in resource/class init

yoav-steinberg:
This is the bare minimum just to make `init` useful. We still don't have verification that all fields are initialized etc.
See #542 
See #864

```wing
bring cloud;

resource R {
  name: str;

  init(name: str) {
    this.name = name; // This should now work, before `this` was an unknown symbol
  }
}

let r = new R("foo");
```

### DIFF
--- a/examples/tests/valid/resource.w
+++ b/examples/tests/valid/resource.w
@@ -5,3 +5,6 @@ resource R {
     this.name = name;
   }
 }
+
+let res = new R("Arr");
+print(resource.name);
--- a/examples/tests/valid/resource.w
+++ b/examples/tests/valid/resource.w
@@ -1,0 +1,7 @@
+resource R {
+  name: str;
+  
+  init(name: str) {
+    this.name = name;
+  }
+}
--- a/examples/tests/valid/resource.w
+++ b/examples/tests/valid/resource.w
@@ -7,4 +7,4 @@ resource R {
 }
 
 let res = new R("Arr");
-print(resource.name);
+print(res.name);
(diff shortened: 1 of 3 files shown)


MarkMcCulloh (changes requested):
Looks good 👍 just needs a test

yoav-steinberg:
> Looks good +1 just needs a test

Sure thing. Test added.

examples/tests/valid/resource.w
```diff
@@ -0,0 +1,7 @@
+resource R {
+  name: str;
+  
+  init(name: str) {
+    this.name = name;
+  }
+}
```
MarkMcCulloh, line 7: Maybe add something like this to ensure it's running the constructor as expected

```suggestion
resource R {
  name: str;
  
  init(name: str) {
    this.name = name;
  }
}

let res = new R("Arr");
print(resource.name);
```

libs/wingc/src/type_check.rs
```diff
@@ -1304,6 +1305,17 @@ impl<'a> TypeChecker<'a> {
 				// Create constructor environment and prime it with args
 				let mut constructor_env = SymbolEnv::new(Some(env), constructor_sig.return_type, false, env_flight, stmt.idx);
 				self.add_arguments_to_env(&constructor.parameters, constructor_sig, &mut constructor_env);
+				// Prime the constructor environment with `this`
```
eladb, line 1308: Is `this` also supported within methods?

yoav-steinberg: That's already working.

eladb (approved):

MarkMcCulloh (approved):

PR #1021, opened 2023-01-02
fix(compiler): constants for wingsdk resource names and other cleanups

yoav-steinberg:
All constants for wingsdk resource names are now define in a single place in _lib.rs_. Also use nested lookup function to get these types from the env when needed.

Also in this PR there's a non related fix for correctly handling unknown builtin types in the return value of a closure during parsing. Before this paniced:
```wing
let a = (): Set<str> => {
}
```
Now it report an error:
```
Error at ../../tmp/tmp.w:1:13 | builtin container type 'Set' is not supported yet see https://github.com/winglang/wing/issues/530
```

### DIFF
--- a/libs/wingc/src/type_check/jsii_importer.rs
+++ b/libs/wingc/src/type_check/jsii_importer.rs
@@ -7,13 +7,11 @@ use crate::{
 		symbol_env::StatementIdx, Class, FunctionSignature, Struct, SymbolKind, Type, TypeRef, Types, WING_CONSTRUCTOR_NAME,
 	},
 	utilities::camel_case_to_snake_case,
-	WINGSDK_DURATION, WINGSDK_INFLIGHT, WINGSDK_RESOURCE,
+	WINGSDK_ASSEMBLY_NAME, WINGSDK_DURATION, WINGSDK_INFLIGHT, WINGSDK_RESOURCE,
 };
 use colored::Colorize;
 use serde_json::Value;
 use wingii::jsii;
-
-const WINGSDK_ASSEMBLY: &'static str = "@winglang/wingsdk";
 
 trait JsiiInterface {
 	fn methods<'a>(&'a self) -> &'a Option<Vec<jsii::Method>>;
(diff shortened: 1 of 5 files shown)


staycoolcall911 (approved):
LGTM. Maybe add a test if it makes sense

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -145,6 +145,17 @@ impl SymbolEnv {
 		}
 	}
 
+	pub fn lookup_nested_str(&self, nested_str: &str, statement_idx: Option<usize>) -> Result<&SymbolKind, TypeError> {
```
MarkMcCulloh, line 148: Nice, I was writing something like this in another branch haha

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -60,13 +59,13 @@ impl<'a> JsiiImporter<'a> {
 					_ => panic!("TODO: handle primitive type {}", primitive_name),
 				}
 			} else if let Some(Value::String(type_fqn)) = obj.get("fqn") {
-				if type_fqn == WINGSDK_INFLIGHT_FQN {
+				if type_fqn == &format!("{}.{}", WINGSDK_ASSEMBLY, WINGSDK_INFLIGHT) {
```
MarkMcCulloh, line 62: Since the constants are static and `format!` is a macro, there's no runtime cost for this right?

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -7,14 +7,13 @@ use crate::{
...
+	WINGSDK_DURATION, WINGSDK_INFLIGHT, WINGSDK_RESOURCE,
 };
 use colored::Colorize;
 use serde_json::Value;
 use wingii::jsii;
 
-const WINGSDK_DURATION_FQN: &'static str = "@winglang/wingsdk.std.Duration";
-const WINGSDK_RESOURCE_FQN: &'static str = "@winglang/wingsdk.core.Resource";
-const WINGSDK_INFLIGHT_FQN: &'static str = "@winglang/wingsdk.core.Inflight";
+const WINGSDK_ASSEMBLY: &'static str = "@winglang/wingsdk";
```
MarkMcCulloh, line 16: Why not include this with the other constants?

libs/wingc/src/type_check/jsii_importer.rs
yoav-steinberg: I'm almost certain there actually is runtime cost here because the formatting macro does need to create a string and concat all this together. I was considering [const_format](https://github.com/rodrimati1992/const_format_crates/) to avoid this, but then decided not... WDYT should I pull that crate in?

MarkMcCulloh: probably an unneeded micro-optimization, I'm ok with it as-is

MarkMcCulloh (approved):

PR #1026, opened 2023-01-03
fix(compiler): diagnostics show the file and line as part of the type name

### LINKED ISSUE #997: Compiler diagnostics show the file:line as part of the type name
### I tried this

```js
enum Foo {A,B,C}

let x = Foo.H;
```

### Expected

```
Compilation failed with 1 errors
Error | Enum `Foo` (declared in code.w:1:6) does not contain value `H`
```

### Instead, this happened

```
Compilation failed with 1 errors
Error | Enum Foo code.w:1:6 does not contain value H
```

It seems like the type name includes the location (`Foo code.w:1:6`).

### Component

Compiler

### Environment

```markdown
- Wing Version:
- OS:
- Node version:
```

### Anything else?

_No response_

staycoolcall911:
This PR fixes 2 things:

1. We're changing the error message to be as appears in issue #997.
Note a couple of changes from #997: 
  a. We used "at" instead of "declared in" because it makes more sense in different contexts where this diagnostic message appears.
  b. We used quotes instead of ticks, since this is the convention on our compiler error messages.
A test for this already exists at: https://github.com/winglang/wing/blob/main/examples/tests/invalid/enums.w 

2. Also fixed a different error where the compiler would panic when we access a field as if it was a method. For example:
```ts
resource SomeResource {
    name: str;
    init() {
    }
}

let x = new SomeResource();
x.name();
```
Added this as a test.

Fixes #997 

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -724,7 +724,7 @@ impl<'a> TypeChecker<'a> {
 				} else {
 					self.expr_error(
 						exp,
-						format!("{} should be a function or method",function
+						format!("\"{}\" should be a function or method",function
 						)
 					);
 					return None;
@@ -1560,7 +1560,7 @@ impl<'a> TypeChecker<'a> {
 							if e.values.contains(property) {
 								return _type;
 							} else {
-								return self.general_type_error(format!("Enum {} does not contain value `{}`", _type, property.name));
+								return self.general_type_error(format!("Enum \"{}\" does not contain value \"{}\"", _type, property.name));
 							}
 						}
 						_ => {
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -724,7 +724,7 @@ impl<'a> TypeChecker<'a> {
 				} else {
 					self.expr_error(
 						exp,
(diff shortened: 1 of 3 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -719,9 +719,16 @@ impl<'a> TypeChecker<'a> {
...
 				// Make sure this is a function signature type
-				let func_sig = func_type
-					.as_function_sig()
-					.expect(&format!("{} should be a function or method", function));
+				let func_sig = if let Some(func_sig) = func_type.as_function_sig() {
+					func_sig
+				} else {
+					self.expr_error(
+						exp,
+						format!("\"{}\" should be a function or method",function
```
yoav-steinberg, line 727: I suspect this isn't formatted using our rustfmt settings. We should try to make sure that all commits use the same fmt settings to avoid extra "non functional" commits in the future.
```suggestion
						format!("\"{}\" should be a function or method", function
```

yoav-steinberg (approved):
see minor comment about code formatting

libs/wingc/src/type_check.rs
staycoolcall911: Thanks!

PR #1030, opened 2023-01-03
feat(compiler): basic global utility functions

### LINKED ISSUE #105: Utility Functions: panic, assert
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#12-utility-functions

3p3r:
resolves #105 


libs/wingc/src/jsify.rs
```diff
@@ -283,8 +283,25 @@ impl JSifier {
...
+					return format!(
+						"{{((msg) => {{console.error(msg, (new Error()).stack);process.exit(1)}})({})}}",
+						self.jsify_arg_list(args, None, None, false)
+					);
+				} else if matches!(&function, Reference::Identifier(Symbol { name, .. }) if name == "throw") {
+					return format!(
+						"{{((msg) => {{throw new Error(msg)}})({})}}",
+						self.jsify_arg_list(args, None, None, false)
+					);
 				} else if let Reference::NestedIdentifier { object, .. } = function {
```
Chriscbr, line 305: It looks like we've collected a lot of global functions. How about we extract this out into something like:

```rs
if let Some(output) = self.jsify_global_function(&function, ...) {
  return output;
} else if let Reference::NestedIdentifier { object, .. } = function {
  ...
}
```

libs/wingc/src/jsify.rs
```diff
@@ -283,8 +283,25 @@ impl JSifier {
 			ExprKind::Reference(_ref) => self.jsify_reference(&_ref, None, phase),
 			ExprKind::Call { function, args } => {
 				let mut needs_case_conversion = false;
+				// note: Globals are emitted here and wrapped in "{{ ... }}" blocks. Wrapping makes these emissions, actual
+				// statements and not expressions. this makes the runtime panic if these are used in place of expressions.
```
Chriscbr, line 287: Isn't this already prevented by the type checker? I'm not sure I see the need to add a second safety mechanism.

e.g. If I write this:
```
let x = print("Hello, world!");
```

and compile it in Wing, it should error when type checking is performed (before any JS is emitted)

libs/wingc/src/jsify.rs
```diff
@@ -283,8 +283,25 @@ impl JSifier {
...
 				if matches!(&function, Reference::Identifier(Symbol { name, .. }) if name == "print") {
-					return format!("console.log({})", self.jsify_arg_list(args, None, None, false));
+					return format!("{{console.log({})}}", self.jsify_arg_list(args, None, None, false));
+				} else if matches!(&function, Reference::Identifier(Symbol { name, .. }) if name == "assert") {
+					return format!(
+						"{{((cond) => {{if (!cond) throw new Error(`Runtime Assertion Failed For Condition: '{0}'`)}})({0})}}",
```
Chriscbr, line 292: Wondering if we should shorten this
```suggestion
						"{{((cond) => {{if (!cond) throw new Error(`assertion failed: '{0}'`)}})({0})}}",
```

Chriscbr (commented):
Woot - this looks mighty useful!

libs/wingc/src/jsify.rs
3p3r: You are correct. I just wanted to emit technically accurate code here, because spec does mention these are statements and not expressions. The block make these expression-statements at JS grammar level.

3p3r: From: https://github.com/tree-sitter/tree-sitter-javascript/blob/7a29d06274b7cf87d643212a433d970b73969016/grammar.js#L247

libs/wingc/src/jsify.rs
```diff
@@ -145,6 +145,33 @@ impl JSifier {
 			}
 		}
 	}
+	// note: Globals are emitted here and wrapped in "{{ ... }}" blocks. Wrapping makes these emissions, actual
```
eladb, line 148: Why double `{{`?

libs/wingc/src/jsify.rs
```diff
@@ -145,6 +145,33 @@ impl JSifier {
...
+					"{{((cond) => {{if (!cond) throw new Error(`preflight assertion failed: '{0}'`)}})({0})}}",
+					self.jsify_arg_list(args, None, None, false)
+				);
+			}
+			UtilityFunctionKind::Print => {
+				return format!("{{console.log({})}}", self.jsify_arg_list(args, None, None, false));
+			}
+			UtilityFunctionKind::Throw => {
+				return format!(
+					"{{((msg) => {{throw new Error(msg)}})({})}}",
```
eladb, line 163: Why are we wrapping in a function?

(Add comment explaining)

libs/wingc/src/jsify.rs
```diff
@@ -145,6 +145,33 @@ impl JSifier {
...
+					self.jsify_arg_list(args, None, None, false)
+				);
+			}
+			UtilityFunctionKind::Panic => {
+				return format!(
+					"{{((msg) => {{console.error(msg, (new Error()).stack);process.exit(1)}})({})}}",
```
eladb, line 169: Same q

libs/wingc/src/jsify.rs
```diff
@@ -145,6 +145,33 @@ impl JSifier {
 			}
 		}
 	}
+	// note: Globals are emitted here and wrapped in "{{ ... }}" blocks. Wrapping makes these emissions, actual
+	// statements and not expressions. this makes the runtime panic if these are used in place of expressions.
+	fn jsify_global_utility_function(&self, args: &ArgList, function_type: UtilityFunctionKind) -> String {
+		match function_type {
+			UtilityFunctionKind::Assert => {
+				return format!(
+					"{{((cond) => {{if (!cond) throw new Error(`preflight assertion failed: '{0}'`)}})({0})}}",
```
eladb: Why preflight? This should work for inflight as well, no?

eladb (approved):

libs/wingc/src/jsify.rs
3p3r: You are too quick! fb08498886bad95c2e00ca6ea8198d1c70cdade4

libs/wingc/src/jsify.rs
3p3r: [It's Rust syntax](https://doc.rust-lang.org/std/fmt/index.html#escaping). It's how a literal `{` is escaped in formatted strings.

libs/wingc/src/jsify.rs
3p3r: https://github.com/winglang/wing/pull/1030#discussion_r1061166157

libs/wingc/src/jsify.rs
3p3r: https://github.com/winglang/wing/pull/1030#discussion_r1061166157

libs/wingc/src/jsify.rs
```diff
@@ -284,7 +311,13 @@ impl JSifier {
 			ExprKind::Call { function, args } => {
 				let mut needs_case_conversion = false;
 				if matches!(&function, Reference::Identifier(Symbol { name, .. }) if name == "print") {
-					return format!("console.log({})", self.jsify_arg_list(args, None, None, false));
+					return self.jsify_global_utility_function(&args, UtilityFunctionKind::Print);
```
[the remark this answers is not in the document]
3p3r: I implemented the `Display` trait on that enum to move these string associations under the same enum definition: 190a4befd6dcab06b451727cf2c57cc9a057d6fe

examples/tests/error/utilities.w
```diff
@@ -0,0 +1,4 @@
+assert(false);
+print("W");
```
eladb, line 2: We need some tests that assert some expected results. For example, where can I see  the output of each one of these?

I understand from @MarkMcCulloh that there's a plan to snapshot errors so maybe that's going to be covered there. But this is something we should prioritize because those are not really sufficient to verify the behavior of these functions in my view.

libs/wingc/src/jsify.rs
eladb: Not exactly sure what's the answer here...

eladb (approved):

libs/wingc/src/jsify.rs
3p3r: Oh this one got confused with the comment below it. The reason for this one is just to stay consistent with the other ones.

libs/wingc/src/lib.rs
```diff
@@ -84,6 +84,36 @@ pub fn type_check(scope: &mut Scope, types: &mut Types) -> Diagnostics {
 		scope,
 		types,
 	);
+	add_builtin(
+		"assert",
```
yoav-steinberg, line 88: Lets avoid having these strings defined twice, something like:
```suggestion
		UtilityFunctionKind::Assert.to_string().as_str(),
```

libs/wingc/src/ast.rs
```diff
@@ -147,6 +147,25 @@ pub struct Stmt {
 	pub idx: usize,
 }
 
+#[derive(Debug)]
+pub enum UtilityFunctionKind {
```
yoav-steinberg, line 151: Maybe just
```suggestion
pub enum UtilityFunctions {
```

libs/wingc/src/lib.rs
3p3r: cac32eca312dcb0450370d25b930a10c6d859f9b

libs/wingc/src/ast.rs
3p3r: cac32eca312dcb0450370d25b930a10c6d859f9b

examples/tests/error/utilities.w
3p3r: I would love to have typed error checking in whatever solution we end up with. For example if code is expected to throw the following Error:
```ts
class WingCliException extends Error {}
```
It needs to have an assertion like this:
```ts
expect(call()).rejects.toThrowWith(WingCliException);
```

Issue #977, opened 2022-12-22
Algebraic data types / union-like classes


Chriscbr:
### Feature Spec

As a Wing user, I would like to be able to express enums where each choice may have one or more associated fields.

Examples:

```js
// Wing
enum Frequency {
  Cron(str),
  Rate(duration),
}

enum ArithmeticExpression {
    Number(num)
    Addition(ArithmeticExpression, ArithmeticExpression)
    Multiplication(ArithmeticExpression, ArithmeticExpression)
}

enum MatchAssertion {
  Absent,
  AnyValue,
  ContainsPattern(str),
}
```

Such an enum could be used like so:

```js
// Wing
let x = Frequency.Cron("0 0 * * *"); // typed as Frequency
let kind = nameof(x); // "Cron"

let message: str = switch x {
  Cron(_) -> "Running with a cron schedule.",
  Rate(d) -> "Running every ${d.minutes} minutes",
};
```

Since a switch statement is the only control flow that lets you safely unwrap an enum, it is the only way to extract the values of the associated fields.

> The name "switch" is used to emphasize that this is primarily a control structure for enums. I think if we chose the name "match" it might confused some users as they may expect the syntax to support matching against other types like tuples, primitives, etc. like Rust ([example](https://doc.rust-lang.org/rust-by-example/flow_control/match.html))

The compiler would translate such an enum into an enum-like class in JavaScript:

```js
// JavaScript
class Frequency {
  private constructor(
    public readonly discriminant: number,
    public readonly field1?: string,
    public readonly field2?: Duration
  ) {}
  public static Cron(field1: string) {
    return new Frequency(0, field1, null);
  }
  public static Rate(field2: Duration) {
    return new Frequency(1, null, field2);
  }
  public nameOf(): string {
    switch (this.discriminant) {
      case 0:
        return "Cron";
      case 1:
        return "Rate";
    }
  }
}
```

And the enum usage would be translated like so:

```js
// JavaScript
let x = Frequency.Cron("0 0 * * *");
let kind = x.nameOf();

let message: string = (function(x) {
  // this could also be implemented with if/else, doesn't really matter
  switch (x.discriminant) {
    case 0:
      return "Running with a cron schedule.";
    case 1:
      return `Running every ${x.field2.minutes} minutes`
  }
})(x);
```

### FAQ

**Q:** What about ordinary enums without any fields?
[...]

PR #1039, opened 2023-01-04
rfc: bring typescript

eladb:
The goal: to be able to `bring` any typescript library to be used in my Wing code.

This is an initial draft for spec of this feature.

[Rendered version](https://github.com/winglang/wing/blob/eladb/bring-typescript/rfcs/bring-ts.md)

### DIFF
--- a/rfcs/bring-ts.md
+++ b/rfcs/bring-ts.md
@@ -9,7 +9,21 @@
 ## Feature Spec
 
 Wing has great interoperability with the TypeScript ecosystem. You can bring any TypeScript library
-and use it Wing.
+and use it in Wing.
+
+### How does Wing bring typescript libraries?
+
+When the Wing compiler sees something like this:
+
+```js
+bring colors;
+```
+
+It will use node's standard module resolution algorithm to lookup for `colors`. When found, it will
+check if there's a `.jsii` manifest in the module. Since there isn't, it will check if the directory `vendor/colors.wlib`
+exists.
+
+This directory is expected to include a `.jsii` file and an `index.js` file.
 
 ### ts2wing
 
@@ -24,21 +38,26 @@ $ ts2wing --version
 0.1.2
 ```
 
-Let's say I want to use the [colors](https://www.npmjs.com/package/colors) library, first, let's
-import it using the `ts2wing` command:
+Let's say I want to use the [colors](https://www.npmjs.com/package/colors) library in my Wing code.
+
+First, I am going to install it via npm/yarm/pnpm or whatever:
+
+```sh
+npm i colors
+```
+
+Now, I am going to generate bindings for it using `ts2wing`:
 
 ```sh
 $ ts2wing colors
-Installing colors@1.4.0 from npm and all dependencies into vendor/colors.wlib.src...
-Processing type definitions (index.d.ts)...
-Generating Wing wrapper classes...
-Compiling with JSII...
-Generating API.md...
-Packaging Wing Library to vendor/colors.wlib
-Cleaning vendor/colors.wlib.src
+Found colors@1.4.0 under node_modules/colors...
+Reading definitions from index.d.ts...
+Generating Wing wrapper classes under vendor/colors/index.ts...
+Compiling JSII to vendor/colors/.jsii...
+Generating vendor/colors/API.md...
 ```
 
-### The .wlib File
+### The .wlib Directory
 
 This command created a Wing Library file under `vendor/colors.wlib`.
 
--- a/rfcs/bring-ts.md
+++ b/rfcs/bring-ts.md
@@ -11,31 +11,27 @@
 Wing has great interoperability with the TypeScript ecosystem. You can bring any TypeScript library
 and use it in Wing.
 
-### TypeScript adapters
+In order to be able to `bring` a TypeScript module into your Wing code, you will need
+a **Wing bindings library** for this module. Bindings can be hand-written or can be 
+automatically generated from `.d.ts` files using `ts2wing`.
 
-In order to be able to `bring` a TypeScript library into your Wing code, you will need
-a **TypeScript adapter** for this library. Adapters can be hand-written or can be automatically
-generated from `.d.ts` files using `ts2wing`.
+> There are certain type definitions that are not supported for automatic generation at the moment.
+> As this project evolves, we will add more patterns as needed.
 
-> At the moment, there will be type definitions that are not supported by `ts2wing`,
-> which means that you might need to manually add adapter code for those libraries. More on that later.
-
-### Manually creating an TypeScript adapter
+### Manual
 
 We will first start by demonstrating what it takes to manually create
-an adapter for a TypeScript library, and then we will show how to use `ts2wing` in order
-to generate the adapter automatically.
+bindings for a TypeScript library, and then we will show how to use generate
+one automatically.
 
 Let's say I want to use the [colors](https://www.npmjs.com/package/colors) library in my Wing code.
-
-Like this:
 
 ```js
 // hello-colors.w
 bring colors;
 ```
 
-First, simply install this library through npm (or any other node-compatible package manager):
+We install this library through npm (or any other node-compatible package manager):
 
 ```sh
 $ npm i colors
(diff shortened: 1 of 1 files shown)


rfcs/bring-ts.md
```diff
@@ -0,0 +1,248 @@
...
+$ ts2wing --version
+0.1.2
+```
+
+Let's say I want to use the [colors](https://www.npmjs.com/package/colors) library, first, let's
+import it using the `ts2wing` command:
+
+```sh
+$ ts2wing colors
+Installing colors@1.4.0 from npm and all dependencies into vendor/colors.wlib.src...
```
Joshswooft, line 32: option to install via yarn or pnpm in addition to npm would be nice

rfcs/bring-ts.md
```diff
@@ -0,0 +1,248 @@
...
+Generating Wing wrapper classes...
+Compiling with JSII...
+Generating API.md...
+Packaging Wing Library to vendor/colors.wlib
+Cleaning vendor/colors.wlib.src
+```
+
+### The .wlib File
+
+This command created a Wing Library file under `vendor/colors.wlib`.
```
Joshswooft, line 43: vendors is a nice folder name, very golang like :)

rfcs/bring-ts.md
```diff
@@ -0,0 +1,248 @@
...
+
+This will open your browser with the generated API documentation for the generated library. Cute,
+ha?
+
+### Consuming the library
+
+OK, let's use this library in our Wing code.
+
+```js
+bring colors;
```
Joshswooft, line 64: any chance of name collisions?

rfcs/bring-ts.md
```diff
@@ -0,0 +1,248 @@
...
+$ ts2wing edit colors
+OK, "colors" is ready to be edited under "vendor/colors.wlib.src".
+After you are done, make sure to run: ts2wing colors
+```
+
+Now, go ahead and make any modifications you like to `vendor/colors.wlib.src` and after
+you are finished, run:
+
+```sh
+$ ts2wing colors
```
Joshswooft, line 93: watch mode would be nice for this kind of stuff

rfcs/bring-ts.md
```diff
@@ -0,0 +1,248 @@
...
+
+This is all nice, but at the moment we believe that supporting type unions is not really required
+for building great cloud applications. In fact, we think that type unions are a potential source of
+bugs and complexity, and we want Wing to be easy to learn and use and scale across large teams in
+big companies. To that end, we don't support type unions (and we reserve the right to change our
+minds).
+
+To bridge this gap, the `ts2wing` tool will generate a "union-like class" in for each type union in
+the imported library. The class will have a set of static factory methods, one for each type in the
+union:
```
MarkMcCulloh, line 138: As a TS fan, this reads as "We think type unions are bad so instead of having them, we're doing something worse". 
It's fair to say we aren't convinced they're important to have in Wing and that it maybe is too difficult to implement right now so we're doing something easier.

Basically, the following specific part doesn't seem helpful, because it's strange to criticize the existence of something while attempting to mimic it:
> In fact, we think that type unions are a potential source of bugs and complexity, and we want Wing to be easy to learn and use and scale across large teams in big companies.

rfcs/bring-ts.md
eladb: Decoupled the npm installation and the wing bindings, so now users control exactly how to install.

rfcs/bring-ts.md
```diff
@@ -0,0 +1,248 @@
...
+This is it, now the updated `colors.wlib` will include your changes.
+
+### Type system adapters
+
+It is not possible to express every type in TypeScript within the Wing type system. This is because:
+
+1. Wing is young and it will take the type system time to evolve; 
+2. The design goals of Wing and TypeScript are different. TypeScript is designed to be able to
+   represent every possible type that can be expressed in JavaScript, while Wing is designed to be a
+   great language for cloud application programming.
```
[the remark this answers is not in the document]
MarkMcCulloh: > I don't agree - it seems to me like JS's dynamic typing is very much a motivating reason why unions exist in TypeScript.

Theoretically, TS did not have to implement ADTs like this right? I figure it could be expressed more [akin to the way you discuss it here](https://github.com/winglang/wing/issues/977) while still supporting legacy APIs. 
I see your point though and will agree that what I said originally isn't fair, the reason they had to tackle the problem **is** that javascript allowed it. However, I think the particular solution they chose is a really good one when it may not have been the only option. 
Does that mean we have to do it the same way they did? No, but saying it's bad is not a good reason.

> Other languages address unions in a way that results in less compiler complexity (=> faster compile times):

You say less compiler complexity, I say less expressiveness. If we want that (which I suppose we do, based on design tenants) then that's fine, but we shouldn't pretend like what TS did is worse.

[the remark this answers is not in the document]
eladb: I :heart: TypeScript's type system and I think it's beautiful and elegant and extremely powerful. And as we discussed, given the goals of the language, the type system is a work of art, it's inspiring and expressive.

But... as I mentioned, Wing doesn't have the same goals as TypeScript. And the decision whether to support type unions or not should be taken within the design goals and tradeoffs of Wing, not of TypeScript. And I don't think we need type unions in Wing. Like @Chriscbr says, most programming languages don't have support type unions. Yes, it might be less expressive for certain use cases, but there is also value in constrainting expressivity sometimes, especially in a language that aims to scale to large enterprises and huge multi-team projects.

I'll try to improve the reasoning, but this is not about aesthetics. The fact that we want to interop with TypeScript doesn't mean we need to carry over every design decision they made. Not because they are "wrong" for typescript, but because they might not be the right thing for Wing.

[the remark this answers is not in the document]
MarkMcCulloh: > The decision whether to support type unions or not should be taken within the design goals and tradeoffs of Wing, not of TypeScript

Agreed, I think union types are a nice idea within the goals and tradeoffs of Wing regardless of TS interop or why they exist in the context of TS. We disagree on this, which is totally okay, but I wanted to make it clear that I don't feel that way because of the reasons you mention.

> The fact that we want to interop with TypeScript doesn't mean we need to carry over every design decision they made.

Then this RFC should strictly be about "What is the experience of using TS libs in Wing" rather than having notes about why we chose not to have certain features of the TS type system in Wing itself. If someone asks why we don't support unions here as-is, then the discussion should be moved elsewhere and focused on Wing's type system design so that it's free of any influence of TS interop.

rfcs/bring-ts.md
```diff
@@ -0,0 +1,267 @@
...
+
+As you can see, since Wing doesn't support global functions, the `Global` wrapper class was
+generated and includes a static method for each public API exported from `colors`.
+
+Similarly to turning global functions to static methods, `ts2wing` applies other heuristics to
+create an [adaptation layer](#type-system-adapters) between the type systems.
+
+### Escape hatch
+
+As you might have expected, there could be cases where `ts2wing` won't be able to create an adapter
```
Joshswooft, line 96: Will a user be made aware of when the wing compiler can't create the necessary adapters? You could take this further and prompt them to use the edit command

rfcs/bring-ts.md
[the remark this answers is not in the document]
Joshswooft: "there is also value in constrainting expressivity sometimes, especially in a language that aims to scale to large enterprises and huge multi-team projects." - I agree with @eladb here. 

Gonna use Golang as the example here, which by design is very simple and not as expressive as other languages e.g. JavaScript. But if you look at any tutorial/repo about Go, all the code looks the same! Thats very powerful because it allows new users to learn concepts very quickly without worrying that they are learning outdated syntax. 

If you take JS as an example for this you have the callbacks vs async/await. Prototypes vs class vs functional approach. This can be quite overwhelming for new users and even become a barrier to some hesitant on learning a new language.

I guess looping back to the design tenants it also fits in: 

- Syntax design aims to be concise and minimal, while being "batteries included" 

So i guess the question is do we see unions as being necessary for the wing ecosystem and not just something we added as a way to support another language?

rfcs/bring-ts.md
```diff
@@ -0,0 +1,267 @@
...
+Generating Wing wrapper classes under vendor/colors/index.ts...
+Compiling JSII to vendor/colors/.jsii...
+Generating vendor/colors/API.md...
+```
+
+### The .wlib Directory
+
+This command created a Wing Library file under `vendor/colors.wlib`.
+
+> A `.wlib` file it a self-contained artifact that represents a reusable library in Wing. It
```
hasanaburayyan, line 64:
```suggestion
> A `.wlib` file is a self-contained artifact that represents a reusable library in Wing. It
```

rfcs/bring-ts.md
```diff
@@ -0,0 +1,267 @@
...
+Similarly to turning global functions to static methods, `ts2wing` applies other heuristics to
+create an [adaptation layer](#type-system-adapters) between the type systems.
+
+### Escape hatch
+
+As you might have expected, there could be cases where `ts2wing` won't be able to create an adapter
+for a certain TypeScript library. It will take us time to establish all the patterns and in parallel
+evolve the Wing type system.
+
+If you run `ts2wing` with the switch `--unpacked` like so:
```
hasanaburayyan, line 100: no example follows using `--unpacked`

rfcs/bring-ts.md
```diff
@@ -0,0 +1,267 @@
...
+
+#### Public dependencies are not supported
+
+Since we want to bundle all the dependencies of a library into a `.wlib`, it is not possible to
+export APIs from the library that uses types from other libraries. This is actually not very common
+in the JavaScript ecosystem (most library APIs just use standard types). To solve this, we need to
+the dependent library to also be "winfigied". This is possible, but for now, I think we can just
+fail if a library tries to export APIs that use types from a dependency. This is in a sense the
+difference between "public dependencies" and "private dependencies" and we need to model that well
+in order to avoid some horrible fitfalls that TypeScript fell into (e.g. it is impossible to use
```
hasanaburayyan, line 265:
```suggestion
in order to avoid some horrible pitfalls that TypeScript fell into (e.g. it is impossible to use
```

rfcs/bring-ts.md
```diff
@@ -0,0 +1,413 @@
...
+bugs and complexity, and we want Wing to be easy to learn and use and scale across large teams in
+big companies. To that end, we don't support type unions (and we reserve the right to change our
+minds).
+
+To bridge this gap, the `ts2wing` tool will generate a "union-like class" in for each type union in
+the imported library. The class will have a set of static factory methods, one for each type in the
+union:
+
+```ts
+declare class FooInitValue {
```
ekeren, line 250: I am trying to imagine the actual `bridge.ts` artifact in this example, hope I understand this correctly.
Also, this is the first time in my life that I am writing typescript code, and I have to say, It is Fucking amazing!!! 

So, bare with me cause I am a noob in typescript...

Imagine we have the following class Foo implementation:

```ts
// foo.ts
export class Foo {
  private value: string | number;
  constructor(value: string | number) {
    console.log(`The parameter is of type ${typeof value}`);
    this.value = value;
  }

  bar(): string[] | string | false {
    if (Math.random() > 0.5) return ["hello", "world"];
    if (Math.random() > 0.5) return "hello world";
    return false;
  }

  toString() {
    return `This is a Foo Class with value ${this.value}`;
  }

  getAnotherFoo(f: Foo) {
    console.log(`got another foo ${f}`);
  }
}
```

I assume we will need to result in a `foo_bridge.ts` file
```ts
// foo_bridge.ts
import { ExampleClass, globalExample, Foo as _inner_Foo } from "./original";

export class FooInitValue {
  static fromStr(value: string): FooInitValue {
    return new FooInitValueImp(value, undefined);
  }
  static fromNum(value: number): FooInitValue {
    return new FooInitValueImp(undefined, value);
  }
}
class FooInitValueImp {
  _str?: string;
  _number?: number;
  
  constructor(s?: string , n?: number){
    if (s) this._str = s; 
    else if (n) this._number = n;
    else throw new Error("Unreachable");
    
  }
  
}
export declare class FooBarResult{
  asBool(): boolean ;
  isBool(): boolean ;
  asMutArrayStr(): string[];
  isMutArrayStr(): boolean ;
  asString(): string ;
  isString(): boolean ;
}
class FooBarResultImpl implements FooBarResult {
  _boolean? : boolean;
  _mutArrayStr?: string[];
  _string?: string;
  asBool(): boolean {
    return this._boolean!;
  }
  isBool(): boolean {
    return typeof this._boolean !== "undefined";
  }
  isMutArrayStr(): boolean {
    return typeof this._mutArrayStr !== "undefined";
  }
  asMutArrayStr(): string[] {
    return this._mutArrayStr!;
  }
  isString(): boolean {
    return typeof this._string !== "undefined";
  }
  asString(): string {
    return this._string!;
  }
}

export class Foo {
  _boxed: _inner_Foo;
  constructor(v: FooInitValue){
    if (v instanceof FooInitValueImp){
      if (v._number) this._boxed = new _inner_Foo(v._number);
      else if (v._str)  this._boxed = new _inner_Foo(v._str);
      else throw new Error("Unreachable");
    }
    else {
      throw new Error("Unreachable");
    }
  }

  toString(){
    return this._boxed.toString();
  }

  bar():  FooBarResult {
    let innerRes = this._boxed.bar();
    let res = new FooBarResultImpl();
    if (typeof innerRes ===  "string" ) res._string = innerRes;
    else if (innerRes instanceof  Array) res._mutArrayStr = innerRes;
    else {
      res._boolean = innerRes;
    }
    return res;
  }
  getAnotherFoo(f: Foo ){
    return this._boxed.getAnotherFoo(f._boxed);
  }
}
```

And then the consumer of the bridge class will look like this:
```ts
import { Foo, FooBarResult, FooInitValue} from "./bridge";

let fooFromNumber = new Foo(FooInitValue.fromNum(10));
let fooFromString = new Foo(FooInitValue.fromStr("ten"));
fooFromNumber.getAnotherFoo(fooFromString);
let bar = new Foo(FooInitValue.fromNum(20)).bar();

if (bar.isBool()) {
  console.log(`bar ${bar} is bool = ${bar.asBool()}`);
}
if (bar.isMutArrayStr()) {
  console.log(`bar ${bar} is MutArrayStr = ${bar.asMutArrayStr()}`);
}
if (bar.isString()) {
  console.log(`bar ${bar} is String = ${bar.asString()}`);
}
```

First, is this correct? or close to being correct? 
Second, please notice that any exported entity that is linked to the exported original Foo also needs to be implemented in the bridge using the new `BoxedFoo`

PR #1043, opened 2023-01-04
fix(compiler): removing duplicate string when interpolate

### LINKED ISSUE #1041: String parsing is duplicating variables
### Community Note

When we create a string with interpolation, if we include more than 1 variable all variables except the last one are duplicated

### I tried this:

When compile this code

```w
bring cloud;

let q = new cloud.Queue();
let b = new cloud.Bucket();
let c = new cloud.Counter(initial: 1000);

let handler = inflight (message: str): str => {
  let value = c.inc();
  b.put("wing.txt", "${message}${value}");
};

q.on_message(handler);
```

### I expected this:

I expect this code on my inflight function (javascript)

```js
async handle(message) { const { b, c } = this; {
  const value = (await c.inc());
  (await b.put("wing.txt",`${message}${value}`));
} };
```

### Instead, this happened:

```js
async handle(message) { const { b, c } = this; {
  const value = (await c.inc());
  (await b.put("wing.txt",`${message}${message}${value}`));
} };
```

the `message` variable duplicate, if I add more variables all except the last will be duplicated

### Component:

Compiler

### Wing Version:

0.4.51

### Node.js Version:

v18.12.1

### Platform(s):

Linux

### Anything else?

I believe the problem is in the parser between lines 613 to 650 of the file `wing/libs/wingc/src/parser.rs`

When running this test
```bash
npx nx dev -- wing/examples/tests/valid/expressions_string_interpolation.w 
```

We can see duplicated data on this file `wing/examples/tests/valid/expressions_string_interpolation.w.out/preflight.js`

marciocadev:
When interpolate a string with more than one variable, all variables except the last one are duplicated

I create a simple validation on string parser file to remove the duplicated value

Please review this solution carefully as I do not program in Rust, but when I run the current test with this command
```bash
$ npx nx dev -- wing/examples/tests/valid/expressions_string_interpolation.w 
```
the resulting file was created correctly and can be found in the folder 
`wing/examples/tests/valid/expressions_string_interpolation.w.out/preflight.js`

Closes #1041 


libs/wingc/src/parser.rs
```diff
@@ -628,11 +628,10 @@ impl Parser<'_> {
 						let interpolation_end = interpolation_node.end_byte();
 
 						if interpolation_start != last_start {
```
MarkMcCulloh, line 630:
```suggestion
						if interpolation_start >= last_end
```

libs/wingc/src/parser.rs
```diff
@@ -628,11 +628,10 @@ impl Parser<'_> {
...
-							parts.push(InterpolatedStringPart::Static(
-								str::from_utf8(&self.source[last_start..interpolation_start])
-									.unwrap()
-									.into(),
-							));
+							let previous_string = str::from_utf8(&self.source[last_start..last_end]).unwrap();
+							let current_string = str::from_utf8(&self.source[last_start..interpolation_start]).unwrap();
+							let removing_duplication = current_string.replace(previous_string, "");
+							parts.push(InterpolatedStringPart::Static(removing_duplication,));
```
MarkMcCulloh, line 634:
```suggestion
							parts.push(InterpolatedStringPart::Static(
								str::from_utf8(&self.source[last_end..interpolation_start])
									.unwrap()
									.into(),
							));
```

MarkMcCulloh (changes requested):
Good catch! Now I have to wonder if this was ever somehow accidentally working 😅.
Your change does work, but I think with a slight tweak to the original logic it can work as expected. See comments.

Also, the snapshot tests need to be updated locally by running `npx nx test hangar`

marciocadev:
Whit this changes I remove the `last_start` variable too

marciocadev:
@MarkMcCulloh I made some changes but this break the cool_string example, I will explore a little more to resolve this

eladb (changes requested):
Can we add a test for this.

@MarkMcCulloh can you provide some guidance?

marciocadev:
I adding two test string on `expressions_string_interpolation.w` and run `npx nx test hangar`

Is this how we test it?

eladb:
> I adding two test string on `expressions_string_interpolation.w` and run `npx nx test hangar`
> 
> 
> 
> Is this how we test it?
> 
> 

LGTM! 🙏

MarkMcCulloh:
The changes look good! I just noticed that this PR is opened from a fork. Currently, our workflows require changes to come from branches on this repo so a new PR will need to be created. As a org member, you have the permissions to do this, but if you'd like I can take care of moving your commits to a new branch since you already made the lovely changes. 
Let me know if you could like some help @marciocadev

eladb:
Here's a [live example](https://youtu.be/kbl-SC3FaBU?t=4061) of this bug!

marciocadev:
> The changes look good! I just noticed that this PR is opened from a fork. Currently, our workflows require changes to come from branches on this repo so a new PR will need to be created. As an org member, you have the permissions to do this, but if you'd like I can take care of moving your commits to a new branch since you already made the lovely changes. Let me know if you could like some help @marciocadev

Hi @MarkMcCulloh I tried to send it directly but it's giving me a 403 error, even though I put the parsonal access token as password

PR #1045, opened 2023-01-05
feat(compiler): baby generics in std lib

MarkMcCulloh:


eladb:
This should be a `feat`, not a `fix`...

libs/wingc/src/jsify.rs
```diff
@@ -324,7 +324,12 @@ impl JSifier {
 					return self.jsify_global_utility_function(&args, UtilityFunctions::Throw);
 				} else if let Reference::NestedIdentifier { object, .. } = function {
 					let object_type = object.evaluated_type.borrow().unwrap();
-					needs_case_conversion = object_type.as_class_or_resource().unwrap().should_case_convert_jsii;
+					needs_case_conversion = if let Some(obj) = object_type.as_class_or_resource() {
+						obj.should_case_convert_jsii
+					} else {
+						// This should only happen in the case of `any`, which are almost certainly JSII imports.
```
yoav-steinberg, line 330: Just to be sure, even in case of JSII imports this shouldn't normally happen. It might be required because of unimplemented stuff. CMIIW.

Assuming I'm right, I had this thought that it might be good to print some notice level report in these cases, just so we're aware of what's happening. Makes sense?

libs/wingsdk/src/std/array.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Mutable Array
+ */
+export class MutArray {
+  /**
+   * The length of the array
+   */
+  get length(): number {
+    return 0 as any;
```
eladb, line 9:
```suggestion
    throw new Error("abstract");
```

libs/wingc/src/type_check.rs
```diff
@@ -1508,6 +1505,88 @@ impl<'a> TypeChecker<'a> {
...
 	}
 
+	fn pseudo_reify_std_class(&mut self, env: &SymbolEnv, original_type_name: &str, type_param: TypeRef) -> TypeRef {
+		let std_lookup = env.try_lookup("std", None).unwrap();
+		let std_namespace = &std_lookup.as_namespace().unwrap().env;
+		let original_type = std_namespace
+			.try_lookup(original_type_name, None)
+			.unwrap()
+			.as_type()
+			.unwrap();
```
yoav-steinberg, line 1515: Lets add appropriate class names as constants in _lib.rs_: `WINGSDK_ARRAY` or `STD_ARRAY` set to `"std.Array"` and then pass the dot delimited string to this function and use `env.netsted_lookup_str()`.

libs/wingsdk/src/std/array.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Get the value at the given index
+   * @param _index index of the value to get
+   * @returns the value at the given index
+   */
+  public at(_index: number): any {
```
eladb, line 55:
```suggestion
  public at(index: number): any {
```

libs/wingc/src/type_check.rs
```diff
@@ -1508,6 +1505,88 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	fn pseudo_reify_std_class(&mut self, env: &SymbolEnv, original_type_name: &str, type_param: TypeRef) -> TypeRef {
```
eladb, line 1508: Maybe add some docstrings here - the name is not intuitive to me.

libs/wingc/src/type_check.rs
```diff
@@ -1508,6 +1505,88 @@ impl<'a> TypeChecker<'a> {
...
+		});
+		let mut new_type = self.types.add_type(tt);
+		let new_type_class = new_type.as_mut_class_or_resource().unwrap();
+
+		// add stuff from env into new class
+		for (name, symbol) in original_type_class.env.iter() {
+			match symbol {
+				SymbolKind::Variable(v) => {
+					if let Some(sig) = v.as_function_sig() {
+						let return_type = if let Some(ret) = sig.return_type {
```
yoav-steinberg, line 1539: Won't this be cleaner with a `sig.return_type.map(...)` call?

ekeren:
I've tried to compile the build and run the example, and notice 2 problems

Here is the code used:
```ts (wing)
bring cloud;

let ar = ["one", "two", "three", "four"];
let s = ar.at(2);
let handler = inflight (body: str /* string arg */): str => {
  let ss = ar.at(2);
};
```

```sh
wing % ./apps/wing/bin/wing  compile --target sim   examples/tests/valid/file_counter.w
Compilation failed with 3 errors
Error at examples/tests/valid/file_counter.w:4:15 | Expected type to be one of "str", but got "num" instead
Error at examples/tests/valid/file_counter.w:6:12 | Cannot call preflight function "ar.at" while in inflight phase
Error at examples/tests/valid/file_counter.w:6:18 | Expected type to be one of "str", but got "num" instead
```

As you can see there is two issues:
- at is expecting a `str` type 
- at cannot be used in `inflight`

libs/wingc/src/type_check.rs
```diff
@@ -1508,6 +1505,88 @@ impl<'a> TypeChecker<'a> {
...
+			name: original_type_class.name.clone(),
+			env: new_env,
+			parent: original_type_class.parent,
+			should_case_convert_jsii: original_type_class.should_case_convert_jsii,
+		});
+		let mut new_type = self.types.add_type(tt);
+		let new_type_class = new_type.as_mut_class_or_resource().unwrap();
+
+		// add stuff from env into new class
+		for (name, symbol) in original_type_class.env.iter() {
```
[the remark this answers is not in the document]
MarkMcCulloh: An alternative is to either create a type alias in wingsdk like `type TypeParam = any;` and use it as a marker or add some JSII metadata. For now it's strictly targeted at the std namespace and the logic is pretty naive/shallow, so hopefully it's good enough for a while 🤞

libs/wingc/src/jsify.rs
MarkMcCulloh: After checking again I opted to keep the same logic but I made the comment more accurate (I hope).

MarkMcCulloh:
Good catch @ekeren, I added your code as a test and both of those issues are resolved now

libs/wingsdk/src/std/array.ts
```diff
@@ -0,0 +1,65 @@
+// These classes are used by Wing to provide JSII subsets of the JS Array class.
+// They should not be consumed directly by users.
+// TODO: These should be interfaces, currently Wing does not support interface JSII imports
+
+/**
+ * Mutable Array
+ */
+export class MutArray {
```
eladb, line 8: `MutArray` should extend `Array`...

(we can add this in a subsequent PR)

libs/wingsdk/src/std/array.ts
```diff
@@ -0,0 +1,65 @@
+// These classes are used by Wing to provide JSII subsets of the JS Array class.
+// They should not be consumed directly by users.
+// TODO: These should be interfaces, currently Wing does not support interface JSII imports
+
+/**
+ * Mutable Array
+ */
```
eladb, line 7: Thought...

Instead of using `any` as a placeholder by convention, can we use a placeholder type. Something like this:

```js
/**
 * This type is not going to be visible in Wing. It is just used
 * as a placeholder for type parameters. In TypeScript it is `== any`.
 * 
 * @placeholder
 */
export class T {}

/**
 * @typeparam T
 */
export class MutArray {
  push(item: T): void;
}
```

We export `T` as a "placeholder type" and then we add an annotation to `MutArray` that tells Wing that the class takes a type parameter. Then, when we import the type into Wing, we can replace `T` instead of `any`.

Does it make any sense?

libs/wingsdk/src/std/set.ts
```diff
@@ -0,0 +1,39 @@
+// These classes are used by Wing to provide JSII subsets of the JS Set class.
+// They should not be consumed directly by users.
+// TODO: These should be interfaces, currently Wing does not support interface JSII imports
+
+/**
+ * Mutable Set
+ */
+export class MutSet {
```
eladb, line 8:
```suggestion
export class MutSet extends Set {
```

libs/wingsdk/src/std/array.ts
MarkMcCulloh: Yeah that makes perfect sense, and I considered this (mentioned in another comment) but I decided on the simpler approach for now because this doesn't buy us anything yet. This would however be required if we want to support multiple type args. I'm on the fence about whether or not to include it in this PR, what do you think?

MarkMcCulloh: Discussed offline, decided to table this for now.

libs/wingsdk/src/std/array.ts
MarkMcCulloh: This is hard because JSII gets mad if you try to extend Array (it assumes the global array).

To support this, I decided to rename these to ImmutableArray and MutableArray, there's technically no reason the class name needed to match.

libs/wingsdk/src/std/array.ts
```diff
@@ -0,0 +1,48 @@
+// These classes are used by Wing to provide JSII subsets of the JS Array class.
+// They should not be consumed directly by users.
+// TODO: These should be interfaces, currently Wing does not support interface JSII imports
+
+/**
+ * Immutable Array
```
eladb, line 6: (or @wing-name)

```suggestion
 * Immutable Array
 * @name Array
```

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
MarkMcCulloh: I agree but will defer to a later PR, because the same optimization applies in several places

libs/wingsdk/src/std/array.ts
MarkMcCulloh: This is similar to the `@typeparam` thing where I'm concerned about preemptively adding too much use of JSII metadata when it's not technically necessary for this PR.

eladb: Nit picking not important.

eladb (approved):

libs/wingc/src/lib.rs
```diff
@@ -30,6 +30,10 @@ pub mod utilities;
 const WINGSDK_ASSEMBLY_NAME: &'static str = "@winglang/wingsdk";
 
 const WINGSDK_DURATION: &'static str = "std.Duration";
+const WINGSDK_ARRAY: &'static str = "std.ImmutableArray";
+// const WINGSDK_MUTARRAY: &'static str = "std.MutableArray";
```
eladb, line 34: ?

libs/wingsdk/src/std/array.ts
eladb: No problem

libs/wingc/src/lib.rs
MarkMcCulloh: Adding commented code is a sin indeed, will remove

eladb:
@MarkMcCulloh congrats! this is very cool

PR #1050, opened 2023-01-05
feat(compiler): set container type

staycoolcall911:
Adding `Set` to wing.

Examples:
```ts
let s2: Set<num> = {1, 2, 3};
let s3 = Set<num> {1, 2, 3};
let s4: Set<num> = Set<num> {1, 2, 3};

let bucket1 = new cloud.Bucket() as "bucket1";
let bucket2 = new cloud.Bucket() as "bucket2";
let bucket3 = new cloud.Bucket() as "bucket3";
let s6: Set<cloud.Bucket> = {bucket1, bucket2, bucket3};
```


libs/wingc/src/jsify.rs
```diff
@@ -394,6 +394,16 @@ impl JSifier {
...
+			ExprKind::SetLiteral { items, .. } => {
+				format!(
+					"Object.freeze({{{}}})",
+					items
+						.iter()
+						.map(|expr| self.jsify_expression(expr, phase))
+						.collect::<Vec<String>>()
+						.join(", ")
+				)
+			}
```
MarkMcCulloh, line 406: In the spec, Set us using the [global Set class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), which also does not a constructor to initialize it with values so you'd have to call add for each one with something like this (untested)

```suggestion
			ExprKind::SetLiteral { items, .. } => {
				format!(
					"Object.freeze(new Set(){})",
					items
						.iter()
						.map(|expr| format!(".add({})", self.jsify_expression(expr, phase)))
						.collect::<Vec<String>>()
						.join("")
				)
			}
```

MarkMcCulloh (changes requested):

examples/tests/valid/container_types.w
```diff
@@ -18,4 +18,10 @@ let m2: Map<num> = {"a":1, "b":2, "c":3};
...
 let m5 = {"a":bucket1, "b":bucket2, "c":bucket3};
-let m6: Map<cloud.Bucket> = {"a":bucket1, "b":bucket2, "c":bucket3};
\ No newline at end of file
+let m6: Map<cloud.Bucket> = {"a":bucket1, "b":bucket2, "c":bucket3};
+
+//Set tests
+let s2: Set<num> = {1, 2, 3};
+let s3 = Set<num> {1, 2, 3};
+let s4: Set<num> = Set<num> {1, 2, 3};
+let s6: Set<cloud.Bucket> = {bucket1, bucket2, bucket3};
```
eladb, line 27: Curious- i am assuming JS used referential equality to determine uniqueness. Correct? Does that mean that `add()` is o(n)? (We have to check if each element is already there)

staycoolcall911: I don't think it checks if an element is already there. It just overrides it

libs/wingc/src/jsify.rs
MarkMcCulloh: Wait, just kidding the constructor can take an array.

```suggestion
			ExprKind::SetLiteral { items, .. } => {
				format!(
					"Object.freeze(new Set([{}]))",
					items
						.iter()
						.map(|expr| self.jsify_expression(expr, phase))
						.collect::<Vec<String>>()
						.join(", ")
				)
			}
```

examples/tests/valid/container_types.w
eladb: But there isn't really a formal concept of a hashcode in JavaScript.

The [spec] is very illusive.

[spec]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#performance

MarkMcCulloh (approved):

Issue #256, opened 2022-10-12
Cannot submit PRs from forks


MarkMcCulloh:
PR builds should:
- Trigger based on "pr target" changes
- Build everything
- Test what is needed based on changes
- Support use in forks

"Releasable" builds should:
- Trigger based on "push to main"
- Build everything
- Test everything

Chriscbr:
Pinning this issue to improve visibility for anyone trying to contribute.

Right now we have a item in our short-term backlog to get our CI/CD workflows working on forks - until then, if you are someone looking to submit a pull request, please request access through https://www.winglang.io/ to get membership access to the repo, and then you can create a branch directly. Thanks!

monadabot:
Congrats! :rocket: This was released in [Wing 0.4.76](https://github.com/winglang/wing/releases/tag/v0.4.76).

PR #1052, opened 2023-01-05
fix(compiler): removing duplicate string when interpolating

### LINKED ISSUE #1041: String parsing is duplicating variables

marciocadev:
When interpolate a string with more than one variable, all variables except the last one are duplicated

I create a simple validation on string parser file to remove the duplicated value

Please review this solution carefully as I do not program in Rust, but when I run the current test with this command

```bash
$ npx nx dev -- wing/examples/tests/valid/expressions_string_interpolation.w
```
the resulting file was created correctly and can be found in the folder
`wing/examples/tests/valid/expressions_string_interpolation.w.out/preflight.js`

Closes #1041 

### DIFF
--- a/examples/tests/valid/expressions_string_interpolation.w
+++ b/examples/tests/valid/expressions_string_interpolation.w
@@ -4,3 +4,6 @@ let number = 1;
 
 let cool_string = "cool \"\${${regular_string}}\" test";
 let really_cool_string = "${number}${empty_string}\n${cool_string}\n\${empty_string}${"string-in-string"}!";
+
+let begining_with_cool_strings = "${regular_string} ${number} <- cool";
+let ending_with_cool_strings = "cool -> ${regular_string} ${number}";
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -633,18 +633,18 @@ impl Parser<'_> {
 							start_from = last_start;
 						}
 
-						if interpolation_start != last_start {							
+						if interpolation_start != last_start {
 							parts.push(InterpolatedStringPart::Static(
 								str::from_utf8(&self.source[start_from..interpolation_start])
-								.unwrap()
-								.into(),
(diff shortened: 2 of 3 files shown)


MarkMcCulloh (approved):
Excellent, thanks for both fixing and improving the original test. I pushed up a quick change where I ran rustfmt because I noticed some extra whitespace, but otherwise this was great 👍

eladb:
Thanks @marciocadev and apologies for the mess with the forks... We are on it (https://github.com/winglang/wing/issues/256)

PR #1055, opened 2023-01-05
feat(sdk): adding peek method to counter

### LINKED ISSUE #964: Counter `peek()`
### Summary

Add a method to retrieve the value of a Counter

### Feature Spec

```
/**
 * Returns the current value.
 * @returns the value of the counter.
 */
peek(): Promise<number>;
```

### Use Cases

Currently, there is no getter for the counter value. 

We use the `inc` method which returns the `previous` value:

`libs/wingsdk/src/cloud/counter.ts`
```
/**
 * Inflight interface for `Queue`.
 */
export interface ICounterClient {
  /**
   * Increments the counter atomically by a certain amount and returns the previous value.
   * @param amount amount to increment (default is 1).
   * @returns the previous value of the counter.
   */
  inc(amount?: number): Promise<number>;
}
```

When using it on the Console, we need to use it like this:
```
.query("counter.peek", {
    ...
    const response = await client.inc(0);
    return response;
  },
});
```

### Implementation Notes

_No response_

### Component

SDK

marciocadev:
The peek method return the current value of counter

Closes #964 

### DIFF
--- a/libs/wingsdk/src/cloud/counter.ts
+++ b/libs/wingsdk/src/cloud/counter.ts
@@ -70,7 +70,9 @@ export interface ICounterClient {
   inc(amount?: number): Promise<number>;
 
   /**
-   * Get the current value of the counter
+   * Get the current value of the counter,
+   * Using this API may introduce race conditions since the value can change between
+   * the time it is read and the time it is used in your code.
    * @returns current value
    * @inflight
    */
@@ -78,9 +80,7 @@ export interface ICounterClient {
 }
 
 /**
- * List of inflight operations available for `Counter`,
- * Using this API may introduce race conditions since the value can change between
- * the time it is read and the time it is used in your code.
+ * List of inflight operations available for `Counter`
  */
 export enum CounterInflightMethods {
   /** `Counter.inc` */
--- a/libs/wingsdk/src/cloud/counter.ts
+++ b/libs/wingsdk/src/cloud/counter.ts
@@ -68,6 +68,13 @@ export interface ICounterClient {
    * @inflight
    */
   inc(amount?: number): Promise<number>;
+
+  /**
+   * Get the current value of the counter
+   * @returns current value
+   * @inflight
+   */
+  peek(): Promise<number>;
 }
 
 /**
@@ -76,4 +83,6 @@ export interface ICounterClient {
 export enum CounterInflightMethods {
   /** `Counter.inc` */
   INC = "inc",
+  /** `Counter.peek` */
+  PEEK = "peek",
 }
--- a/libs/wingsdk/src/cloud/counter.ts
+++ b/libs/wingsdk/src/cloud/counter.ts
@@ -78,7 +78,9 @@ export interface ICounterClient {
 }
 
 /**
- * List of inflight operations available for `Counter`.
+ * List of inflight operations available for `Counter`,
+ * Using this API may introduce race conditions since the value can change between
+ * the time it is read and the time it is used in your code.
  */
 export enum CounterInflightMethods {
   /** `Counter.inc` */
--- a/libs/wingsdk/src/target-tf-aws/counter.inflight.ts
+++ b/libs/wingsdk/src/target-tf-aws/counter.inflight.ts
@@ -49,7 +49,7 @@ export class CounterClient implements ICounterClient {
     const result = await this.client.send(command);
     let currentValue = result.Item?.[VALUE_ATTRIBUTE].N;
     if (!currentValue) {
-      throw new Error(`${VALUE_ATTRIBUTE} attribute not found on table.`);
+      return this.initial;
     }
 
     return parseInt(currentValue);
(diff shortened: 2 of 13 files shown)


libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -35,7 +35,7 @@ export class Counter extends cloud.CounterBase {
     if (ops.includes(cloud.CounterInflightMethods.INC)) {
       host.addPolicyStatements({
         effect: "Allow",
-        action: ["dynamodb:UpdateItem"],
+        action: ["dynamodb:UpdateItem", "dynamodb:GetItem"],
         resource: this.table.arn,
       });
     }
```
Chriscbr, line 41: This condition just applies to when `inc` is called. Instead you'll want to add another if branch for the peek method:

```
    if (ops.includes(cloud.CounterInflightMethods.INC)) {
      // leave unchanged
    }
    if (ops.includes(cloud.CounterInflightMethods.PEEK)) {
      host.addPolicyStatements({
        effect: "Allow",
        action: ["dynamodb:GetItem"],
        resource: this.table.arn,
      });
    }
```

libs/wingsdk/src/cloud/counter.ts
```diff
@@ -68,6 +68,13 @@ export interface ICounterClient {
    * @inflight
    */
   inc(amount?: number): Promise<number>;
+
+  /**
+   * Get the current value of the counter
```
Chriscbr, line 73: We should add some kind of warning about the potential for race conditions:
```suggestion
   * Get the current value of the counter.
   * Using this API may introduce race conditions since the value can change between
   * the time it is read and the time it is used in your code.
```

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
```diff
@@ -35,4 +39,19 @@ export class CounterClient implements ICounterClient {
...
+      Key: { [HASH_KEY]: { S: COUNTER_ID } },
+    });
+
+    const result = await this.client.send(command);
+    let currentValue = result.Item?.[VALUE_ATTRIBUTE].N;
+    if (!currentValue) {
+      throw new Error(`${VALUE_ATTRIBUTE} attribute not found on table.`);
+    }
+
+    return parseInt(currentValue);
```
Chriscbr, line 55: Currently `Counter`'s implementation doesn't add an item to the DynamoDB table when the resource is deployed because doing so would require us to run code on deploy (something like [cdk triggers](https://docs.aws.amazon.com/cdk/api/v1/docs/triggers-readme.html)) - but such a resource/feature hasn't been added in the Wing SDK yet.

To work around this, we need to handle the situation where `inc` hasn't been called yet by returning the initial value if no item is found:
```suggestion
    const result = await this.client.send(command);
    let currentValue = result.Item?.[VALUE_ATTRIBUTE].N;
    if (!currentValue) {
      return this.initial;
    }

    return parseInt(currentValue);
```

libs/wingsdk/test/target-tf-aws/counter.inflight.test.ts
```diff
@@ -51,14 +54,39 @@ test("inc(5)", async () => {
...
+test("peek without initial value", async () => {
+  setupPeekMock({
+    expectedTableName: MOCK_TABLE_NAME,
+  });
+
+  // WHEN
+  const client = new CounterClient(MOCK_TABLE_NAME);
+  const response = await client.peek();
+
+  expect(response).toEqual(123);
```
Chriscbr, line 79: I think this should be 0

Chriscbr (changes requested):
Thanks for the pull request @marciocadev! Just a few comments and this should be ready to ship 👍

libs/wingsdk/src/target-tf-aws/counter.ts
marciocadev: I see, thank you @Chriscbr 

changind now

libs/wingsdk/src/cloud/counter.ts
```diff
@@ -68,12 +68,23 @@ export interface ICounterClient {
...
+   * @inflight
+   */
+  peek(): Promise<number>;
 }
 
 /**
- * List of inflight operations available for `Counter`.
+ * List of inflight operations available for `Counter`,
+ * Using this API may introduce race conditions since the value can change between
+ * the time it is read and the time it is used in your code.
```
Chriscbr, line 83: Oops - I think this is in the wrong place (it should be on the `peek` docstring)

marciocadev: Sorry, changing now

Chriscbr (approved):
Nice work!

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -40,6 +40,14 @@ export class Counter extends cloud.CounterBase {
...
 
+    if (ops.includes(cloud.CounterInflightMethods.INC)) {
```
eladb, line 43: Shouldn't that be `PEEK`?

eladb (commented):
Very cool! I think there was a small mistake

libs/wingsdk/src/target-tf-aws/counter.ts
marciocadev: Sorry, fixing now

PR #1065, opened 2023-01-08
fix(sdk): aws function names cannot be longer than 64 characters

### LINKED ISSUE #170: Enforce that generated AWS function names are not longer than 64 characters
Otherwise this results in a deployment error

marciocadev:
Enforce AWS Function Names limits and ensuring that WING_FUNCTION_NAME has the same value as FunctionName

Closes #170 

### DIFF
--- a/libs/wingsdk/src/target-tf-aws/function.ts
+++ b/libs/wingsdk/src/target-tf-aws/function.ts
@@ -114,9 +114,11 @@ export class Function extends cloud.FunctionBase {
       role: this.role.name,
     });
 
+    const functionName = this.sanitizeFunctionName(this.node.id);
+
     // Create Lambda function
     this.function = new LambdaFunction(this, "Default", {
-      functionName: this.sanitizeFunctionName(this.node.id),
+      functionName: functionName,
       s3Bucket: bucket.bucket,
       s3Key: lambdaArchive.key,
       handler: "index.handler",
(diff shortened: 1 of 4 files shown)


libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -114,9 +114,11 @@ export class Function extends cloud.FunctionBase {
       role: this.role.name,
     });
 
+    const functionName = this.sanitizeFunctionName(this.node.id);
+
     // Create Lambda function
     this.function = new LambdaFunction(this, "Default", {
-      functionName: this.sanitizeFunctionName(this.node.id),
```
eladb, line 119: `node.id` is only unique within the scope. We need something that is unique across the app.

The best practice from the CDK is to use a trimmed human-readable component (e.g. sanitized `node.id`) and suffix it using [`this.node.addr`] which is a app-unique 42 character hash (it's a hash of the construct path).

[`this.node.addr`]: https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_core.ConstructNode.html#addr

I would recommend creating a utility function in `util.ts` to help generate these names and use it everywhere in the AWS implementation.

See [CDK8s naming utilities](https://github.com/cdk8s-team/cdk8s-core/blob/2.x/src/names.ts) as inspiration (copying the maintainer @iliapolo)

eladb (changes requested):

libs/wingsdk/src/target-tf-aws/function.ts
marciocadev: This is cool, extending this approach to other resources like a Bucket, when we create a new bucket we can easily find it with a friendly name

```w
new Bucket() as "my-bucket";
```

Generate a Bucket name like `my-bucket-c8045fccc85a7`

eladb: Yes!

marciocadev: I will close this PR for now

PR #1076, opened 2023-01-08
feat: capture immutable values

### LINKED ISSUE #1061: Unable to capture immutable primitives
### Community Note

### I tried this:

```js
bring cloud;

let arr = ["hello", "world"];

new cloud.Function(inflight (e: str): str => {
  print("${arr.at(0)} ${arr.at(1)}");
});
```

### I expected this:

That when I invoke the cloud function, it will emit `hello world` to the log.

### Instead, this happened:

The resulting inflight code is invalid:

```js
async handle(e) { const {  } = this; {
  {console.log(`${(arr.at(0))} ${(arr.at(1))}`)};
} };
```

### Component:

Compiler, SDK

### Wing Version:

0.4.57

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

eladb:
It is now possible to capture immutable primitives (strings, booleans, numbers) and collections (array, set, map). This is done by serializing them to JSON and including in the inflight bundle.

We expanded the bindings API such that it now accepts `resources` and `data`. Resources are bound and `_toInflight`ed while data doesn't need any bindings and it is JSON-serialized into the bundle.

### SDK support

In order to support this in the SDK, two things needed to happen. First, the base `resource._bind()` method now only calls `_bind()` on the bound resource (and adds a connection) if it has such as method. Otherwise, it simply passes it through. This means that SDK users can technically pass any object and the SDK will not be defensive about it. *We currently assume "capture-safety" is the responsibility of the Wing compiler.* Similarly, we only call `toInflight` if the object has this method. Otherwise, we simply call `JSON.stringify`. 

There's a special case for `Set` because JSON.stringify doesn't work out of the box for sets.

The `Duration` type is serialized through its `toJSON` method into its respective fields.

The SDK now takes a snapshot of the inflight code. To avoid ugly paths in the code bundles, I am filtering out any lines that start with `//`. Not ideal, but harmless (famous last words).

### Compiler support

* Update `capture.rs` to identify primitives and collections and basically pass them as-is under `bindings`.
* Propagate a mutable `Diagnostics` vector into `capture.rs` so it is possible to report errors. Specifically, if users try capture an uncapturable type, an error is emitted.

### Testing

Added a couple of tests: `capture_collections.w` and `capture_primitives.w` to cover the various types of immutable objects.

Fixes #1061


libs/wingc/src/lib.rs
```diff
@@ -175,6 +175,11 @@ pub fn compile(source_file: &str, out_dir: Option<&str>) -> Result<CompilerOutpu
 	let mut diagnostics = parse_diagnostics;
 	diagnostics.extend(type_check_diagnostics);
 
+	// Analyze inflight captures
```
eladb, line 178: Moved this before error reporting so it is included

libs/wingsdk/src/core/internal.ts
```diff
@@ -62,3 +57,20 @@ ${Object.entries(clients)
...
+
+  // if this is a Set, we need to convert it to an array and reconstruct on the other side
+  if (typeof obj === "object" && (obj as any) instanceof Set) {
+    return NodeJsCode.fromInline(`new Set(${JSON.stringify(Array.from(obj))})`);
+  }
+
+  // otherwise, just JSON.stringify() it and reconstruct on the other side
+  //
+  // NOTE: this won't work if `obj` is not serializable, but this is not the ownership of the sdk
+  return NodeJsCode.fromInline(JSON.stringify(obj));
```
MarkMcCulloh, line 75: Maybe worth bringing in a 3rd party lib to simplify/harden the logic? https://github.com/terrymorse58/serialize-anything

I understand wanting to avoid external deps, but this would basically solve the issue for several data types (including Duration I think). Since we don't have that many right now, I wouldn't consider it a blocker.

libs/wingsdk/src/core/resource.ts
```diff
@@ -154,26 +154,30 @@ export abstract class Resource
...
 
-        // add connection metadata
-        for (const op of resources[field]) {
-          Resource.addConnection({
-            from: host,
-            to: obj,
-            relationship: op,
-          });
+        // if this is a bindable object, call `_bind()` and add a connection for each operation.
+        if (typeof obj === "object" && "_bind" in obj) {
```
MarkMcCulloh, line 164: Back in the days of the original polycons RFC, I remember thinking that capturing (or binding) should probably have an interface associated with it. IBindable or whatever.
 
Resources would obviously implement that interface, but then primitives would also have a similar Primitive class that implements it with plain old serialization. Then the compiler could emit calls to `Primitive.binding("my object")`). I figured this would be necessary if we wanted to be able to bind arbitrary classes.

Does that path make sense, and/or is it an unnecesarry complication?

libs/wingsdk/src/core/resource.ts
```diff
@@ -154,26 +154,30 @@ export abstract class Resource
...
         }
-        if (!("_bind" in obj)) {
-          throw new Error(
-            `Resource ${this.node.path} field ${key} does not have a bind method`
-          );
-        }
-        obj._bind(host, resources[field]);
 
-        // add connection metadata
-        for (const op of resources[field]) {
```
Chriscbr, line 170: This PR adds a lot of duck-typing into the SDK for handling whether we're calling `_bind` or `_toInflight` on resources or objects. We should avoid this if possible since it doubles the number of cases a contributor/maintainer needs to think in many parts of the SDK code.

We can simplify this by annotating whether the captured item is a resource or a piece of immutable data. The compiler has this information, so we can just model that in the compiler's `Capture` struct and in the SDK's `InflightBinding` interface:

```rs
// capture.rs
struct Capture {
	pub object: Symbol,
	pub kind: CaptureType,
}

enum CaptureKind {
	Resource(CaptureDef),
	ImmutableData,
}
```

```ts
/**
 * A resource binding.
 */
export interface InflightBinding {
  /**
   * The resource. 
   */
  readonly resource?: IResource;
  
  /**
   * A piece of immutable data.
   */
  readonly object?: any;

  /**
   * The list of operations used on the resource.
   */
  readonly ops: string[];
}
```

libs/wingsdk/src/util.ts
```diff
@@ -87,42 +81,38 @@ export function sanitizeValue(obj: any, options: SanitizeOptions = {}): any {
   return newObj;
 }
 
-export function directorySnapshot(root: string) {
+export function directorySnapshot(rooti: string) {
```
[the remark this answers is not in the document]
Chriscbr: Or `initialRoot` maybe

libs/wingsdk/src/cloud/function.ts
```diff
@@ -100,6 +101,13 @@ export abstract class FunctionBase extends Resource implements IInflightHost {
       external: ["aws-sdk"],
     });
 
+    // the bundled contains line comments with file paths, which are not useful for us, especially
+    // since they may contain system-specific paths. sadly, esbuild doesn't have a way to disable
+    // this, so we simply filter those out from the bundle.
+    const outlines = readFileSync(outfile, "utf-8").split("\n");
+    const isLineComment = (line: string) => !line.startsWith("//");
```
hasanaburayyan, line 108: Nit
This returns true when line is not a comment. Which can be confusing since since the filter here is meant to "filter for" and not "filter out". Might be more clear with
```suggestion
    const isNotLineComment = (line: string) => !line.startsWith("//");
    writeFileSync(outfile, outlines.filter(isNotLineComment).join("\n"));
```

libs/wingsdk/src/util.ts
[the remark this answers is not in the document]
eladb: There's no happiness here 😒

libs/wingsdk/src/core/internal.ts
eladb: I want us to be very explicit about how and what we are serializing.

We are not supposed to be serializing complex object graphs, just pure data (or resources thought their clients), so I don't see a real need for a heavy hammer.

libs/wingsdk/src/core/resource.ts
eladb: I'll look into it.

Alternatively I am tempted to explore the idea of basically implementing `toInflight` through `toJSON`. I have a gut feeling they are basically the same thing, but I need to think about it a little more.

libs/wingc/src/capture.rs
```diff
@@ -246,10 +248,21 @@ fn scan_captures_in_expression(exp: &Expr, env: &SymbolEnv, statement_idx: usize
 							})
 							.collect::<Vec<Capture>>(),
 					);
+				} else if t.is_immutable_collection() || t.is_primitive() {
```
yoav-steinberg, line 251: Shouldn't this only capture stuff defined in preflight?
```suggestion
				} else if (t.is_immutable_collection() || t.is_primitive()) && f == Phase::PreFlight {
```

libs/wingc/src/capture.rs
```diff
@@ -246,10 +248,21 @@ fn scan_captures_in_expression(exp: &Expr, env: &SymbolEnv, statement_idx: usize
...
+					res.push(Capture {
+						object: symbol.clone(),
+						def: CaptureDef { method: "*".to_owned() },
+					});
+				} else {
+					diagnostics.push(Diagnostic {
+						level: DiagnosticLevel::Error,
+						message: format!("Can't capture '{}' of type {} in preflight", symbol.name, t),
```
yoav-steinberg, line 259:
```suggestion
						message: format!("Can't capture '{}' of type {} in inflight", symbol.name, t),
```

yoav-steinberg (commented):
Very cool, only reviewed the compiler code (lazy me).

libs/wingsdk/src/core/resource.ts
eladb: Done (as you suggested).

I still feel there's a potential opportunity for a unified model (of the universe), but for now, we'll just do with the general theory of relativity.

libs/wingc/src/capture.rs
```diff
@@ -18,41 +17,47 @@ pub struct CaptureDef {
...
 		captures
 			.entry(capture.object.name)
 			.or_insert(BTreeSet::new())
-			.insert(capture.def);
+			.insert(capture.kind);
 	}
 	captures
 }
 
-pub fn scan_for_inflights_in_scope(scope: &Scope) {
```
Chriscbr, line 37: nit: We may be able to avoid passing around `diagnostics` everywhere by associating all of these scanning methods with a single struct like `CaptureScanner` that takes ownership of the diagnostics object, and making each method an implementation on the struct (so they each accept `&mut self` as an argument and can all `self.diagnostics.push(...)` inside).

Might not be a quick refactor, so feel free to skip

libs/wingsdk/src/core/resource.ts
```diff
@@ -154,26 +154,30 @@ export abstract class Resource
...
+        if (obj === undefined) {
           throw new Error(
             `Resource ${this.node.path} does not have field ${key}`
           );
         }
-        if (!("_bind" in obj)) {
-          throw new Error(
-            `Resource ${this.node.path} field ${key} does not have a bind method`
-          );
-        }
```
Chriscbr, line 166: Since you refactored so we should only have to call `_bind` on resources, can we add this safety check back?

Chriscbr (approved):
Nice work!

yoav-steinberg (approved):
compiler code looks good.

libs/wingsdk/src/core/resource.ts
eladb: Done

PR #1077, opened 2023-01-08
fix(compiler): implicitly import types outside of namespace when importing jsii module

### LINKED ISSUE #834: Implicitly resolve dependencies outside of namespace on JSII `bring`
`bring cloud` should work even if it contains a reference to something in wingsdk outside of `cloud` namespace.

```wing
bring cloud; // this should not fail even if something in `cloud` references `core.Something`
// But we have to to make sure this will fail compilation even if not explicit `bring core;` is called.
let x = new core.Something(); // I guess we need to somehow mark the `core` namespace as hidden unless explicitly brought in.
```

See discussion here: https://github.com/winglang/wing/pull/788#issuecomment-1336000236

yoav-steinberg:
Before this fix anything in the wingsdk needed to be contained within a single namespace. You couldn't, for example, have some class inherit from a class in another module. This is why we moved `FileBase` to be under `fs` instead of under `core`, because the JSII importer failed to import `fs.TextFile` because it included something from `core` in it.

This should resolve #834. 


libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -145,18 +188,32 @@ impl SymbolEnv {
...
+		)
 	}
 
-	pub fn lookup_nested(&self, nested_vec: &[&Symbol], statement_idx: Option<usize>) -> Result<&SymbolKind, TypeError> {
+	pub fn lookup_nested(
+		&self,
+		nested_vec: &[&Symbol],
+		ignore_hidden: bool,
+		statement_idx: Option<usize>,
+	) -> Result<&SymbolKind, TypeError> {
```
Chriscbr, line 216: Maybe some kind of comment like
```suggestion
    /// Pass `ignore_hidden: true` if it's OK to return types that have only been imported implicitly (such as through an inheritance chain), and false otherwise
	pub fn lookup_nested(
		&self,
		nested_vec: &[&Symbol],
		ignore_hidden: bool,
		statement_idx: Option<usize>,
	) -> Result<&SymbolKind, TypeError> {
```

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -37,15 +39,33 @@ impl JsiiInterface for jsii::InterfaceType {
...
-	pub namespace_name: String,
-	pub wing_types: &'a mut Types,
-	pub import_statement_idx: usize,
+	jsii_types: &'a wingii::type_system::TypeSystem,
+	assembly: &'a Assembly,
+	module_name: &'a str,
+	wing_types: &'a mut Types,
+	import_statement_idx: usize,
+	env: &'a mut SymbolEnv,
 }
```
Chriscbr, line 48: Add comments?

libs/wingsdk/src/core/files.ts
```diff
@@ -7,8 +7,8 @@ import {
 } from "fs";
 import { isAbsolute, join } from "path";
 import { IConstruct } from "constructs";
-import { FileBase } from "../fs";
```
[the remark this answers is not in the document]
Chriscbr: edit: I wrote this before our offline discussion - feel free to keep this as is

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -180,6 +237,12 @@ impl SymbolEnv {
 				}
 			}
 			let ns = if let Some(ns) = t.as_namespace() {
+				if ns.hidden && !ignore_hidden {
+					return Err(TypeError {
+						message: format!("Symbol \"{}\" is not a namespace", symb.name),
```
Chriscbr, line 242: Maybe we could add a unit test covering this edge case

yoav-steinberg: I invested a bit in making a test for this, solved some bugs and cleanups on the way. Thanks for this comment!

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -370,14 +387,23 @@ impl<'a> JsiiImporter<'a> {
 		}
 	}
 
+	fn is_fqn_construct_base(fqn: &str) -> bool {
+		// We treat both CONSTRUCT_BASE and WINGSDK_RESOURCE, as base constructs because in wingsdk we currently have stuff directly derived
+		// from `construct.Construct` and stuff derived `cloud.Resource` (which itself is derived from `constructs.Construct`).
+		// But since we don't support interfaces yet we can't import `cloud.Resource` so we just treat it as a base class.
+		// I'm also not sure we should ever import `cloud.Resource` because we might want to keep its internals hidden to the user:
```
Chriscbr, line 394:
```suggestion
		// But since we don't support interfaces yet we can't import `core.Resource` so we just treat it as a base class.
		// I'm also not sure we should ever import `core.Resource` because we might want to keep its internals hidden to the user:
```

Chriscbr (approved):
LGTM, just one nit.

PR #1078, opened 2023-01-09
feat(sdk): meaningful resource name

### LINKED ISSUE #170: Enforce that generated AWS function names are not longer than 64 characters

### LINKED ISSUE #1047: Physical resource names are meaningless
### Community Note

### I tried this:

I wrote this program:

```js
bring cloud;

new cloud.Bucket() as "bucket1";
new cloud.Bucket() as "bucket2";
```

And deployed it to AWS.

### I expected this:

(from https://github.com/winglang/wing/discussions/861) 

The names of the resources Wing creates in the cloud are auto-generated and not human friendly.  Furthermore, there doesn't seem to be a property that allows them to be set by the developer (I looked through the doc, but didn't see a name field).  I'd like at the minimum to be able to specify resource names so I'm not looking in buckets like: `terraform-20221125051333963700000005`.  Ideally there would be some automatic naming too that is more grok-able than a uuid.  I'm not sure whether you need a high level entity like a project to help with this, but maybe `account-region-project-stage-name`.

### Instead, this happened:

This was the name of my bucket `terraform-20221125051333963700000005`

### Component:

SDK

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

marciocadev:
Trying to give more human friendly names to the aws resources Wing create

Fixes #1047 
Closes #170 


libs/wingsdk/src/utils/aws/topic.name.ts
```diff
@@ -0,0 +1,29 @@
+import { Construct } from "constructs";
+import { NameOptions } from "../name-options";
+
+/**
+ * AWS rules for Topic Names
+ *
+ * Topic names are limited to 256 characters.
+ * You can use alphanumeric characters, hyphens (-) and underscores (_).
+ */
+export class TopicName {
```
eladb, line 10: Lots of duplicated code and boilerplate. Feels like all of these can basically be folded into one or two utility methods that are not resource specific and then we can call them from each resource with the specific constraints.

WDYT?

marciocadev: You're right, I will make some changes to avoid this

libs/wingsdk/src/utils/aws/aws-names.ts
```diff
@@ -0,0 +1,76 @@
+import { CaseConventions, NameOptions, ResourceType } from "../resource-names";
+
+export const awsResourceNameProps = (
```
eladb, line 3: Sorry for being a bit picky about this, but kind of feels like each one of these can go into a local `const` next to where it is used instead of in a centralized location.

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,72 @@
...
+  readonly regexMatch: RegExp;
+  readonly charReplacer: string;
+  readonly maxLen: number;
+  readonly minLen?: number;
+  readonly case?: CaseConventions;
+  readonly avoidStartWith?: string;
+  readonly prefix?: string;
+}
+
+export enum ResourceType {
```
eladb, line 21: This enum not needed I think

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,72 @@
...
+  AWS_COUNTER = "aws-counter",
+  AWS_FUNCTION = "aws-function",
+  AWS_QUEUE = "aws-queue",
+  AWS_TOPIC = "aws-topic",
+
+  AZURE_BUCKET = "azure-bucket",
+}
+
+export class ResourceNames {
+  public static of(resource: Construct, type: ResourceType): string {
```
eladb, line 32:
```suggestion
  public static of(resource: Construct, props: NameOptions): string {
```

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,72 @@
...
+      name = name.toLocaleUpperCase();
+    }
+
+    return name
+      .replace(opt!.regexMatch, opt!.charReplacer)
+      .substring(0, opt!.maxLen);
+  }
+}
+
+function resourceNameProps(type: ResourceType): NameOptions | undefined {
```
eladb, line 60: Not needed

libs/wingsdk/src/utils/aws/aws-names.ts
marciocadev: Ohhhh you mean which resource pass its options

This makes sense and all the logic regarding the resource will be in the resource itself

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -63,6 +64,21 @@ export class Counter extends cloud.CounterBase {
   private envName(): string {
     return `DYNAMODB_TABLE_NAME_${this.node.addr.slice(-8)}`;
   }
+
+  /**
+   * Couter (Table) names must be between 3 and 255 characters.
```
perpil, line 69: Counter

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -18,7 +19,7 @@ export class Counter extends cloud.CounterBase {
     super(scope, id, props);
 
     this.table = new DynamodbTable(this, "Default", {
-      name: `wingsdk-counter-${this.node.addr}`,
+      name: this.sanitizeName(`${this.node.id}-${this.node.addr}`),
```
perpil, line 22: This still appears to be a generated name of the format: `wingsdk-counter-nodeid-nodeaddr`
You can tell it is a counter now, but not which one.  This may be fine for the generated names, but I'd like a property that I can specify like `name` that will set it to what I've named it as is and maybe a `namePrefix` that I can set if I want a generated name that isn't the default: `namePrefixValue-nodeid-nodeaddr`

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -21,7 +26,9 @@ export class Bucket extends cloud.BucketBase {
 
     this.public = props.public ?? false;
 
-    this.bucket = new S3Bucket(this, "Default");
+    this.bucket = new S3Bucket(this, "Default", {
+      bucket: this.sanitizeName(`${this.node.id}-${this.node.addr}`),
```
perpil, line 30: Why is this called `bucket` and not something consistent across resources like `name`?

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -114,9 +115,11 @@ export class Function extends cloud.FunctionBase {
       role: this.role.name,
     });
 
+    const functionName = this.sanitizeName(`${this.node.id}-${this.node.addr}`);
```
perpil, line 118: I'd like to set the name via property.

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -114,9 +115,11 @@ export class Function extends cloud.FunctionBase {
...
+
     // Create Lambda function
     this.function = new LambdaFunction(this, "Default", {
-      functionName: this.sanitizeFunctionName(this.node.id),
+      functionName: functionName,
       s3Bucket: bucket.bucket,
```
perpil, line 123: Can you allow setting the bucket name via a property too `deploymentBucketName`?

libs/wingsdk/src/target-tf-aws/topic.ts
```diff
@@ -19,7 +20,9 @@ export class Topic extends cloud.TopicBase {
   constructor(scope: Construct, id: string, props: cloud.TopicProps = {}) {
     super(scope, id, props);
 
-    this.topic = new SnsTopic(this, "Default", {});
+    this.topic = new SnsTopic(this, "Default", {
+      name: this.sanitizeName(`${this.node.id}-${this.node.addr}`),
```
perpil, line 24: I'd like to be able to set name via `props`

libs/wingsdk/src/target-tf-aws/bucket.ts
marciocadev: The `bucket` property is from S3Bucket class

It's from CdkTF library

libs/wingsdk/src/target-tf-aws/counter.ts
marciocadev: `node.id` can be defined

if you create a `Counter` with this command `new cloudCounter() as "somethink";`

On AWS Console you will see a DynamoDB `wingsdk-counter-somethink-<addr>`

libs/wingsdk/src/target-tf-aws/function.ts
marciocadev: Same as `Counter`

`new cloud.Function() as "F1"`;

On AWS Console you will see `F1-<addr>`

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -114,9 +115,11 @@ export class Function extends cloud.FunctionBase {
       role: this.role.name,
     });
 
+    const functionName = this.sanitizeName(`${this.node.id}-${this.node.addr}`);
+
     // Create Lambda function
     this.function = new LambdaFunction(this, "Default", {
-      functionName: this.sanitizeFunctionName(this.node.id),
+      functionName: functionName,
```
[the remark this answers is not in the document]
marciocadev: No problem

perpil (commented):
There seem to be some inconsistencies across resources.  Some have a prefix like counter and some don't like function.  I would recommend deciding on whether to prefix or not and making it consistent.

There also isn't a way for me to specify the name verbatim.  I'd like to be able to pass in a `name` property.  I've left some details on one such way of implementing this on the counter.

libs/wingsdk/src/target-tf-aws/topic.ts
marciocadev: Same before

`new cloud.Topic() as "MyAmazingTopic";`

AWS Console -> `MyAmazingTopic-<addr>`

marciocadev:
@eladb I was talking with @perpil on Slack and we wondering if we can add a property name as optional on props interface or create a new ResourceProps interface with common params like

```ts
export interface ResourceProps {
  readonly name?: string;
}
```

so it can be used like

```ts
export interface FunctionProps extends ResourceProps { ... }
```

so if we create a `Function` with `name`, the function name would be what is defined
otherwise it would be the current name

eladb:
> @eladb
> 
> 
> 
> I was talking with @perpil on Slack and we wondering if we can create a new ResourceProps interface with common params like
> 
> 
> 
> ```ts
> 
> export interface ResourceProps {
> 
>   readonly name?: string;
> 
> }
> 
> ```
> 
> 
> 
> so it can be used like
> 
> 
> 
> ```ts
> 
> export interface FunctionProps extends ResourceProps { ... }
> 
> export interface CounterProps extends ResourceProps { ... }
> 
> ...
> 
> ```
> 
> 
> 
> so if we create a `Function` with `name`, the function name would be what is defined
> 
> otherwise it would be `<node.id>-<node.addr>`
> 
> 

Yes, this is something we can add, but let's discuss over a separate issue because there are some implications that we should think about:

1. Is it a resource specific thing or under ResourcePrpps
2. If the resource is part of a composite resource, it would be impossible to define two instances of the composite resource because the physical name will conflict, so we really don't want to encourage this. 

3. If this is needed for human readability, it might be better to assign resource tags based on the `display` api we are working on.
4. There might be very different behaviors across providers when it comes to physical names.

Can you create an issue with this context and we will discuss.

@Chriscbr this is technically something that should be included in the sdk rfc

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,38 @@
+export enum CaseConventions {
+  LOWERCASE = "lowercase",
+  UPPERCASE = "uppercase",
+}
+
+export interface NameOptions {
+  readonly resource?: string;
```
eladb, line 7: This prop seems to be ignored.

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,38 @@
...
+  readonly charReplacer: string;
+  readonly maxLen: number;
+  readonly minLen?: number;
+  readonly case?: CaseConventions;
+  readonly avoidStartWith?: string;
+  readonly prefix?: string;
+}
+
+export class ResourceNames {
+  public static of(name: string, props: NameOptions): string {
```
eladb, line 18: As mentioned in a previous review, the names should be constructed such that the `node.addr` is always included because it ensures that there will be no duplicates throughout the app (it is a hash of the path of the construct).

This implies that to `maxLen` should only effect the "human-readable" portion of the name, and not the address.

So a name should usually look like this:

```ts
const hash = node.addr;
const human = node.id.substring(0, props.maxLen - hash.length);
return ${human}-${hash};
```

(check my math!)

I would recommend passing in the construct node instead of `name: string`, so that the formatting logic can be centralized here.

If you go back to CDK8s implementation, you'll see this logic as well.

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,38 @@
...
+  LOWERCASE = "lowercase",
+  UPPERCASE = "uppercase",
+}
+
+export interface NameOptions {
+  readonly resource?: string;
+  readonly regexMatch: RegExp;
+  readonly charReplacer: string;
+  readonly maxLen: number;
+  readonly minLen?: number;
```
eladb, line 11: This seems to be ignored as well.

libs/wingsdk/src/target-tf-aws/queue.ts
```diff
@@ -118,6 +120,20 @@ export class Queue extends cloud.QueueBase {
   private envName(): string {
     return `QUEUE_URL_${this.node.addr.slice(-8)}`;
   }
+
+  /**
+   * Queue names are limited to 80 characters.
+   * You can use alphanumeric characters, hyphens (-), and underscores (_).
+   */
+  private sanitizeName(name: string): string {
+    const nameProps: NameOptions = {
```
eladb, line 129: nit: I would just put this as a const outside the class and use `ResourceNames.of()` directly in the ctor:

```ts
const NAME_OPTS = {
  maxLen: 80,
  regexMatch: /.../,
  charReplacer: "-"
};

export class Queue ... {
  constructor(...) {
    // ...
    this.queue = new SqsQueue(this, "Default", {
      name: ResourceNames.of(this, NAME_OPTS),
      // ----------------------^
      // see comment below about passing the construct so that the logic
      // that combines the id and addr is centralized under ResourceNames.
    });
  }
}
```

eladb (changes requested):
@marciocadev a few more tweaks are needed. 

(it's a very important aspect of the system, so bear with me until this is perfect).

libs/wingsdk/src/utils/resource-names.ts
marciocadev: @eladb 
I saw the limits of Bucket name size on Azure and max length was 24 characters, currently hash has 42 characters

If we apply the cut only on id I think it's not be userfull to Azure

Chriscbr: @marciocadev It should be okay to use a small suffix of the hash (like `node.addr.substring(-8)`) since that should still provide enough randomness to avoid conflicts

libs/wingsdk/test/target-sim/queue.test.ts
```diff
@@ -2,7 +2,7 @@ import * as cloud from "../../src/cloud";
 import { SimApp, Testing } from "../../src/testing";
 import { listMessages } from "./util";
 
-jest.setTimeout(10_000); // 5 seconds
+jest.setTimeout(100000); // 5 seconds
```
Chriscbr, line 5:
```suggestion
jest.setTimeout(100_000); // 100 seconds
```

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -117,6 +124,25 @@ export class Bucket extends cloud.BucketBase {
...
+   * with 'xn--' or ending with '-s3alias' and must not contain
+   * two adjacent dots.
+   */
+  private sanitizeName(name: string): string {
+    const nameProps: NameOptions = {
+      maxLen: 63,
+      case: CaseConventions.LOWERCASE,
+      regexMatch: /(([(.)\1+]+)[^a-z0-9\-]+)/g,
+      charReplacer: "-",
+      avoidStartWith: "xn--",
```
Chriscbr, line 141: I think we can make the API a bit more general by removing this field (`avoidStartsWith`). Instead of throwing an error when a generated name happens to start with "xn--", we can just add it to the `regexMatch` so it gets replaced with "-" anywhere in the string:

```suggestion
      regexMatcher: /(([(.)\1+]+)[^a-z0-9\-]+)|xn--/g,
      regexReplacer: "-"
```

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -117,6 +124,25 @@ export class Bucket extends cloud.BucketBase {
...
+   * You can use lowercase alphanumeric characters, dot(.) and dash (-),
+   * must begin and end with alphanumeric character, can not begin
+   * with 'xn--' or ending with '-s3alias' and must not contain
+   * two adjacent dots.
+   */
+  private sanitizeName(name: string): string {
+    const nameProps: NameOptions = {
+      maxLen: 63,
+      case: CaseConventions.LOWERCASE,
+      regexMatch: /(([(.)\1+]+)[^a-z0-9\-]+)/g,
```
Chriscbr, line 139: Could you add a comment explaining this regex? It will be helpful for maintaining this code :)

marciocadev: sure, I will put some comments in all

libs/wingsdk/src/utils/resource-names.ts
marciocadev: good, I think that leaves plenty of room for a descriptive name

libs/wingsdk/test/target-sim/queue.test.ts
marciocadev: thank you

libs/wingsdk/src/target-tf-aws/bucket.ts
marciocadev: the problem is if a node.id was `xn-- somethink` will be replaced to `-somenthink` and bucket names must begin and end with a letter or number

I was thinking in put `forbiddenStartWith` and `forbiddenEndWith`

Chriscbr: I see... another option might be to do the assertion/throw `bucket.ts` so `ResourceNames` only just responsible for name generation. But I'm ok leaving this in this field if you think there are multiple resources this could be useful for

libs/wingsdk/src/utils/resource-names.ts
Chriscbr, line 18: Let's give this method another name, like `generateName` or `makeName`. `of` is typically reserved for methods that find a resource in some `constructs` scope (like `Stack.of(scope)` or `App.of(scope)`).

libs/wingsdk/src/target-tf-aws/bucket.ts
marciocadev: I actually only see these limitations on AWS Buckets

I think validating on `bucket.ts` is a great option

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,39 @@
...
+  UPPERCASE = "uppercase",
+}
+
+export interface NameOptions {
+  readonly regexMatch: RegExp;
+  readonly charReplacer: string;
+  readonly maxLen: number;
+  readonly minLen?: number;
+  readonly case?: CaseConventions;
+  readonly forbidenStartwith?: string;
```
eladb, line 14: This doesn't seem to be used

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,39 @@
+import { Construct } from "constructs";
+
+export enum CaseConventions {
+  LOWERCASE = "lowercase",
+  UPPERCASE = "uppercase",
+}
+
+export interface NameOptions {
+  readonly regexMatch: RegExp;
```
eladb, line 9: This is currently "disallowedRegEx". I think we should flip the logic and make this "allowedRegEx". It's safer to say which characters are allowed then to try to cover all the disallowed cases (or use a negative regex).

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly charReplacer: string;
```
eladb, line 10: I'd call this `sep` and make it `-` by default.

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly minLen?: number;
+  readonly case?: CaseConventions;
+  readonly forbidenStartwith?: string;
+  readonly forbidenEndwith?: string;
+  readonly prefix?: string;
+}
+
+export class ResourceNames {
+  public static generateName(resource: Construct, props: NameOptions): string {
+    let hash = resource.node.addr.substring(0, 8);
```
eladb, line 21: Add some defensive check that maxLen is >8 because addr at least 8.

eladb (changes requested):

libs/wingsdk/src/utils/resource-names.ts
marciocadev: good catch

libs/wingsdk/src/utils/resource-names.ts
marciocadev: I agree, it makes more sense

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -2,10 +2,21 @@ import { DynamodbTable } from "@cdktf/provider-aws/lib/dynamodb-table";
...
 export const HASH_KEY = "id";
 
+/**
+ * Counter (Table) names must be between 3 and 255 characters.
+ * You can use alphanumeric characters, dot (.), dash (-), and underscores (_).
+ */
+const NAME_OPTS: NameOptions = {
+  maxLen: 255,
+  allowedRegEx: /[^a-zA-Z0-9\_\.\-]+/g,
+  prefix: "wingsdk-counter-",
```
eladb, line 17:
```suggestion
  prefix: "wing-counter-",
```

eladb (commented):
Great job!

Thanks for your patience and persistence 🙏

LGTM. @Chriscbr Any additional feedback?

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -2,10 +2,21 @@ import { DynamodbTable } from "@cdktf/provider-aws/lib/dynamodb-table";
...
 
 export const HASH_KEY = "id";
 
+/**
+ * Counter (Table) names must be between 3 and 255 characters.
+ * You can use alphanumeric characters, dot (.), dash (-), and underscores (_).
+ */
+const NAME_OPTS: NameOptions = {
+  maxLen: 255,
+  allowedRegEx: /[^a-zA-Z0-9\_\.\-]+/g,
```
Chriscbr, line 16: This API doesn't make sense to me, it's named "allowed regex" but it looks like we are describing what characters aren't allowed?

I think it needs to be changed to something like this:
```suggestion
  allowedRegex: "a-zA-Z0-9\_\.\-",
```

I don't think it matters if the API accepts a string or regex (whichever is easier to implement) as long as the name generator has the intended behavior of replacing all _disallowed_ characters.

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  readonly allowedRegEx: RegExp;
+  /**
+   * Word breaker
+   *
+   * @default "-"
+   */
+  readonly sep?: string;
+  readonly case?: CaseConventions;
+  readonly prefix?: string;
```
Chriscbr, line 24: Missing docstrings for these two fields

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,53 @@
+import { Construct } from "constructs";
+
+export enum CaseConventions {
+  LOWERCASE = "lowercase",
+  UPPERCASE = "uppercase",
+}
+
+export interface NameOptions {
```
Chriscbr, line 8:
```suggestion
/**
 * Options for `ResourceNames.generateName`
 */
export interface NameOptions {
```

libs/wingsdk/src/utils/resource-names.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Maximum name length must always be greater than the hash (>=8 characters)
+   */
+  readonly maxLen: number;
+  /**
+   * Regular expression that indicates which characters are valid
+   */
+  readonly allowedRegEx: RegExp;
```
Chriscbr, line 16: Can we make these fields optional? I think having no max length and allowing any characters makes sense as defaults. (be sure to mention defaults in the docstrings)

libs/wingsdk/src/target-tf-aws/counter.ts
marciocadev: I think you're right, I can change the variable name, but I think regex give us more ways to change the name

Like an aws bucket must not contain two adjacent dots so this regex can filter for us `/([^a-z0-9\.\-]+)|(\.{2,})/g`

But I will change the variable to `disallowedRegEx`

libs/wingsdk/src/utils/resource-names.ts
marciocadev: I can change to Optional but I don't know if some cloud provider don't have some type of limitations on name resources

I will make this Optional

Chriscbr: Alright, let's start with this / adjust if there's a use case 👍

Chriscbr (approved):

Chriscbr:
Looks great -- I made a few small tweaks to clean up the docstrings and change how a few resource names are generated (like using the `bucketPrefix` field provided by terraform). I'm excited to use this 😄

eladb (approved):

PR #1080, opened 2023-01-09
feat(sdk): visual annotations of resources

polamoros:


libs/wingsdk/src/core/tree.ts
```diff
@@ -134,6 +161,17 @@ function synthAttributes(
   return undefined;
 }
 
+function synthDisplay(construct: IConstruct): DisplayInfo | undefined {
+  if (construct instanceof Resource) {
```
eladb, line 165: Why is this needed?

libs/wingsdk/src/core/tree.ts
```diff
@@ -33,6 +34,31 @@ export interface ConstructTreeNode {
...
+
+  /**
+   * Description of the resource.
+   */
+  readonly description: string;
+
+  /**
+   * Whether the resource should be hidden from the UI.
+   */
+  readonly hidden: boolean;
```
eladb, line 61:
```suggestion
  /**
   * Title of the resource.
   * @default - The type and/or identifier of the resource
   */
  readonly title?: string;

  /**
   * Description of the resource.
   * @default - No description
   */
  readonly description?: string;

  /**
   * Whether the resource should be hidden from the UI.
   * @default false (visible)
   */
  readonly hidden?: boolean;
```

libs/wingsdk/src/core/resource.ts
```diff
@@ -229,3 +234,33 @@ export interface OperationAnnotation {
...
 }
+
+/**
+ * Information on how to display a resource in the UI.
+ */
+export class Display {
+  /**
+   * Title of the resource.
+   */
+  public title: string;
```
eladb, line 245: All of these should be optional (`?`).

libs/wingsdk/src/core/resource.ts
```diff
@@ -229,3 +234,33 @@ export interface OperationAnnotation {
...
+
+  /**
+   * Whether the resource should be hidden from the UI.
+   */
+  public hidden: boolean;
+
+  public constructor(
+    title: string = "",
+    description: string = "",
+    hidden: boolean = false
```
eladb, line 260:
```suggestion
    title?: string,
    description?: string,
    hidden?: boolean
```

libs/wingsdk/test/target-sim/__snapshots__/bucket.test.ts.snap
```diff
@@ -33,6 +33,11 @@ Object {
             "fqn": "constructs.Construct",
             "version": "10.0.25",
           },
+          "display": Object {
```
eladb, line 36: Omit `display` if if's all defaults (`undefined`).

eladb (changes requested):
@polamoros This looks awesome!

A few notes:
1. Add some unit tests for `display`.
2. Make the `Inflight` resource `hidden: true`.
3. Make the `Logger` resource `hidden: true`.

libs/wingsdk/src/core/tree.ts
```diff
@@ -33,6 +34,34 @@ export interface ConstructTreeNode {
    * Information on the construct class that led to this node, if available.
    */
   readonly constructInfo?: ConstructInfo;
+
+  /**
+   * Information on how to display this node in the UI.
+   */
+  readonly display?: DisplayInfo;
```
Chriscbr, line 41: Instead of adding a new field here, could we add this information as another attribute (like the existing connections and stateful attributes)? This can be done by updating the `_inspect` method:

https://github.com/winglang/wing/blob/aa8e4510cd508ed38213d0afda65d6f4cc860cd1/libs/wingsdk/src/core/resource.ts#L196-L208

(sorry in advance, this needs to be better documented :))

eladb: @Chriscbr I am curious, what's the value you see in using these attributes instead of simply extending the base schema?

Chriscbr: @eladb I guess there isn't much of a difference - my thought process back in the day was "it looks like tree.json in the CDK puts most optional metadata in this `attributes` field, so if we want to play well, we can put stuff there too." There's an also an argument that by putting it in this `attributes` field, it signals more clearly that the attribute isn't required / isn't guaranteed to exist. I'm ok with the current place if you feel it makes more sense though 👍

eladb: I feel there's value in lifting up everything into the schema as first class citizens and getting rid of "attributes" as a 2nd class protocol.

libs/wingsdk/src/cloud/logger.ts
```diff
@@ -12,6 +12,11 @@ export const LOGGER_SYMBOL = Symbol.for(LOGGER_TYPE);
 export abstract class LoggerBase extends Resource {
   public readonly stateful = true;
 
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+    this.display.hidden = true;
```
eladb, line 17: Maybe add a description as well?

libs/wingsdk/src/core/resource.ts
```diff
@@ -294,3 +304,52 @@ export interface OperationAnnotation {
...
+
+  /**
+   * Whether the resource should be hidden from the UI.
+   */
+  public hidden?: boolean;
+
+  public constructor(props?: DisplayProps) {
+    this.title = props?.title;
+    this.description = props?.description;
+    this.hidden = props?.hidden;
```
eladb, line 353:
```suggestion
  public constructor(props: DisplayProps = {}) {
    this.title = props.title;
    this.description = props.description;
    this.hidden = props.hidden;
```

libs/wingsdk/src/core/tree.ts
```diff
@@ -134,6 +164,17 @@ function synthAttributes(
   return undefined;
 }
 
+function synthDisplay(construct: IConstruct): DisplayInfo | undefined {
+  if (
+    construct instanceof Resource &&
```
eladb, line 169: This won't work for the Inflight class

eladb (approved):

libs/wingsdk/src/core/tree.ts
polamoros: fixed!

polamoros:
It seems `package-lock.json`s are not updated on main

libs/wingsdk/src/cloud/logger.ts
eladb: LOL that's not what I meant, I meant `this.display.description`

eladb (commented):
@Chriscbr can you take a look?

libs/wingsdk/src/core/resource.ts
```diff
@@ -294,3 +304,52 @@ export interface OperationAnnotation {
...
+
+  /**
+   * Description of the resource.
+   */
+  public description?: string;
+
+  /**
+   * Whether the resource should be hidden from the UI.
+   */
+  public hidden: boolean;
```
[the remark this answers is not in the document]
polamoros: done 🙏

libs/wingsdk/test/target-sim/logger.test.ts
```diff
@@ -33,3 +33,23 @@ test("inflight uses a logger", async () => {
   expect(listMessages(s)).toMatchSnapshot();
   expect(app.snapshot()).toMatchSnapshot();
 });
+
+test("Logger has display hidden property set to true", async () => {
```
Chriscbr, line 37: Could we also add a test somewhere that built-in resources (Bucket, Counter, Function) are not hidden by default?

libs/wingsdk/src/cloud/logger.ts
polamoros: Why a `display.description` of a Resource that will not be displayed on the Console?

eladb: Mmmm. Good question... The way I see it, we should include a description for all the resources in the SDK, whether they are displayed or not. The reason:

1. The description can be used in other places, not just the console.
2. We might add support in the console to show hidden nodes at some point, in which case it makes sense to have a description for those nodes.

But that's not really critical... We can also add that later (and perhaps better to add it to all resources at once).

polamoros: Agree on both points 👍 

Let me add a title and a description to all resources on this PR.

eladb: 👍

polamoros: I added titles and descriptions from the [SDK ref](https://docs.winglang.io/reference/sdk).. Not the best ones I think.. :)

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -25,6 +25,10 @@ export abstract class BucketBase extends Resource {
   public readonly stateful = true;
   constructor(scope: Construct, id: string, props: BucketProps) {
     super(scope, id);
+
+    this.display.title = "Bucket";
+    this.display.description = "Represents a cloud object store";
```
eladb, line 30:
```suggestion
    this.display.description = "A cloud object store";
```

libs/wingsdk/src/cloud/counter.ts
```diff
@@ -31,6 +31,10 @@ export abstract class CounterBase extends Resource {
 
   constructor(scope: Construct, id: string, props: CounterProps = {}) {
     super(scope, id);
+
+    this.display.title = "Counter";
+    this.display.description = "Represents a distributed atomic counter";
```
eladb, line 36:
```suggestion
    this.display.description = "A distributed atomic counter";
```

libs/wingsdk/src/cloud/function.ts
```diff
@@ -46,6 +46,9 @@ export abstract class FunctionBase extends Resource implements IInflightHost {
   ) {
     super(scope, id);
 
+    this.display.title = "Function";
+    this.display.description = "Represents a function";
```
eladb, line 50:
```suggestion
    this.display.description = "A cloud function (FaaS)";
```

libs/wingsdk/src/cloud/queue.ts
```diff
@@ -33,6 +33,10 @@ export abstract class QueueBase extends Resource {
   public readonly stateful = true;
   constructor(scope: Construct, id: string, props: QueueProps = {}) {
     super(scope, id);
+
+    this.display.title = "Queue";
+    this.display.description = "Represents a queue";
```
eladb, line 38:
```suggestion
    this.display.description = "A distributed message queue";
```

libs/wingsdk/src/cloud/topic.ts
```diff
@@ -17,6 +17,10 @@ export abstract class TopicBase extends Resource {
   public readonly stateful = true;
   constructor(scope: Construct, id: string, props: TopicProps = {}) {
     super(scope, id);
+
+    this.display.title = "Topic";
+    this.display.description = "Represents a topic";
```
eladb, line 22:
```suggestion
    this.display.description = "A pub/sub notification topic";
```

libs/wingsdk/src/core/tree.ts
```diff
@@ -134,6 +165,21 @@ function synthAttributes(
   return undefined;
 }
 
+function isIResource(construct: IConstruct): construct is IResource {
+  return construct instanceof Resource || construct instanceof Inflight;
```
eladb, line 169: Meh... @Chriscbr what do you think about this? Not ideal... but I guess `Inflight` is going away anyway, right?

eladb (approved):

libs/wingsdk/src/core/tree.ts
Chriscbr: Yeah, this looks okay to me. 🙂 BTW we could also opt for duck typing:

```ts
function isDisplayable(construct: IConstruct): boolean {
  if (construct.display) {
    return true;
  }
  return false;
}
```

skyrpex: ^ That would need either ts-ignore or a cast, right?

polamoros: I think we can let it as it is. What do you think?

eladb (approved):
🎉

libs/wingsdk/src/core/tree.ts
Chriscbr: Sounds good

PR #1082, opened 2023-01-09
fix: missing wingsdk in vscode extension (again)

MarkMcCulloh:
https://github.com/winglang/wing/pull/832 moved wingsdk to be a devDep, which no longer included any of its files in the final vscode vsix file. This moves the required `package.json` and `.jsii` to be treated more like the language server binaries instead.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -298,11 +298,6 @@ jobs:
       - name: Prepare language server binaries
         run: cp -r wing-language-server-* ./apps/vscode-wing/resources
 
-      - name: Prepare WingSDK .jsii
-        run: |
-          tar -xzf wingsdk/*.tgz package/.jsii
-          mv package/.jsii libs/wingsdk
-
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
@@ -314,6 +309,11 @@ jobs:
 
       - name: Install Dependencies
         uses: bahmutov/npm-install@v1
+
+      - name: Prepare WingSDK resource
+        run: |
+          tar -xzf wingsdk/*.tgz
+          mv package apps/vscode-wing/resources/wingsdk
 
       - name: Build
         uses: MansaGroup/nrwl-nx-action@v2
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -318,6 +318,7 @@ jobs:
       - name: Build
         uses: MansaGroup/nrwl-nx-action@v2
         with:
+          # build-multi assumes all required resources are present
           targets: "build-multi"
```
Chriscbr, line 322: Maybe we could specify dependent files with nx somehow?

apps/vscode-wing/src/extension.ts
```diff
@@ -67,9 +66,7 @@ async function startLanguageServer(context: ExtensionContext) {
     execSync(`chmod +x ${serverPath}`);
   }
 
-  const wingsdkManifestRoot = dirname(
-    require.resolve("@winglang/wingsdk/.jsii")
-  );
+  const wingsdkManifestRoot = context.asAbsolutePath("resources/wingsdk");
```
Chriscbr, line 69: Is this config/setup needed by all apps consuming the Wing LSP? I wonder if we could add this SDK-types-injection as a build step to the LSP instead so it's bundled with that code instead. (This would match up with how the Wing CLI bundles the compiler with the SDK today)

Chriscbr (approved):

apps/vscode-wing/src/extension.ts
MarkMcCulloh: Honestly, I've been thinking that we should embed the .jsii into wingc itself, it would greatly simplify the story for every dependency (vscode, lsp, playground) since it is always needed. 

I thought about this a long time ago and thought it was hacky, but as time passed it seems more and more reasonable.

.github/workflows/build.yml
MarkMcCulloh: For caching specifically, yes, but I'm not sure it would be super useful since we only use this target in the GH actions.

apps/vscode-wing/src/extension.ts
Chriscbr: Makes sense to me. 👍  Wing SDK isn't a "modular" piece of the compiler / they are both needed for the compiler to work, so coupling them seems natural

MarkMcCulloh: https://github.com/winglang/wing/issues/1083 👍

PR #1089, opened 2023-01-10
fix(docs): installing from github packages throws error on npm 9+

vanjaoljaca:
This arg is needed for newer versions of NPM


Chriscbr (approved):
Thanks for the fix @vanjaoljaca - it looks like people are encountering the same error on stack overflow [here](https://stackoverflow.com/questions/74482745/npm-login-not-allowing-login-to-github), so adding the error sounds like a good idea.

(note: we will be able to remove these instructions soon as we plan to publish wing directly on npm in the near future :-)

PR #1091, opened 2023-01-10
fix(compiler): implicit await missing in argument lists

### LINKED ISSUE #1073: await is not called inside functions
### Community Note

### I tried this:

I tried compiling the following wing code
```ts (wing)
bring cloud;

let q = new cloud.Queue();
let str_to_str = new cloud.Function(inflight (s: str): str => {

}) as "str_to_str";
let func = new cloud.Function(inflight (s: str): str => {
  str_to_str.invoke("one");
  print(str_to_str.invoke("two"));
}) as "func";
```

### I expected this:

await to be called on for both invocations of `str_to_str.invoke`

### Instead, this happened:

The output of the `func` is (in the playground) is:

```ts
async handle(s) { const { str_to_str } = this; {
  (await str_to_str.invoke("one"));
  {console.log((str_to_str.invoke("two")))};
} };
```

Notice that we should `await` the two calls for `str_to_str.invoke`

### Component:

Compiler

### Wing Version:

0.4.61

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

staycoolcall911:
See issue #1073 for the use case - in short, we didn't propagate the phase (e.g. `inflight`) when jsifying arg list, which led to skipping adding implicit await on inflight calls.

Credit for assistance: @yoav-steinberg 👍 

Fixes #1073 

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -149,27 +149,30 @@ impl JSifier {
 	}
 	// note: Globals are emitted here and wrapped in "{{ ... }}" blocks. Wrapping makes these emissions, actual
 	// statements and not expressions. this makes the runtime panic if these are used in place of expressions.
-	fn jsify_global_utility_function(&self, args: &ArgList, function_type: UtilityFunctions) -> String {
+	fn jsify_global_utility_function(&self, args: &ArgList, function_type: UtilityFunctions, phase: Phase) -> String {
 		match function_type {
 			UtilityFunctions::Assert => {
 				return format!(
 					"{{((cond) => {{if (!cond) throw new Error(`assertion failed: '{0}'`)}})({0})}}",
-					self.jsify_arg_list(args, None, None, false)
+					self.jsify_arg_list(args, None, None, false, phase)
 				);
 			}
 			UtilityFunctions::Print => {
-				return format!("{{console.log({})}}", self.jsify_arg_list(args, None, None, false));
+				return format!(
+					"{{console.log({})}}",
+					self.jsify_arg_list(args, None, None, false, phase)
+				);
 			}
 			UtilityFunctions::Throw => {
 				return format!(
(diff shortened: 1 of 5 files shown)


libs/wingc/src/jsify.rs
```diff
@@ -198,7 +208,7 @@ impl JSifier {
 		}
 
 		for arg in arg_list.pos_args.iter() {
-			args.push(self.jsify_expression(arg, Phase::Independent));
+			args.push(self.jsify_expression(arg, phase));
```
eladb, line 211: Well, well, well...

https://github.com/winglang/wing/pull/986/files#r1056930513

eladb (approved):
Nice.

eladb:
The title needs some work - for bugs, the title should reflect the problem, not the solution.
So maybe something like:

```
fix(compiler): implicit await missing in argument lists
```

libs/wingc/src/jsify.rs
staycoolcall911: 🤷‍♂️

yoav-steinberg (approved):

tools/hangar/src/cli.test.ts
```diff
@@ -138,7 +138,7 @@ test.each(validWingFiles)(
       expect(npx_tfJson).toMatchSnapshot("cdk.tf.json");
 
       // get all files in .wing dir
-      const dotWingFiles = await walk.async(path.join(targetDir, ".wing"), {
+      const dotWingFiles = await walk.sync(path.join(targetDir, ".wing"), {
```
Chriscbr, line 141: 🤯  What a clean fix @staycoolcall911

staycoolcall911: Yeah, but it didn't work :)
It's not hangar's fault that our snapshot tests are printing inflight files in a non-deterministic order. That actually happens before, either on our compiler or SDK. I'm trying to look into it, but would love pointers

PR #1100, opened 2023-01-10
chore: follow ups on immutable capture (#1076) and fix build

eladb:
This fixes the build break caused by #1076 as follows:

1. Missed some updates to generated API documentation.
2. Update package-lock.json
3. Emit an "unsupported capture" error only if the symbol is a preflight symbol (otherwise, it is not really a capture).
4. Synthesize `ops: ["a", "b", "c"]` instead of `ops: [a,b,c]` (regression).
5. Emit a "syntax error" when a nested identifier cannot be parsed (the test `missing_semicolon.w` failed with a panic because the type checker did not descend into the inflight closure through the error) <-- we need a more robust model for validating that there are no parse errors at all throughout the tree.

### DIFF
--- a/libs/wingc/src/capture.rs
+++ b/libs/wingc/src/capture.rs
@@ -268,7 +268,7 @@ fn scan_captures_in_expression(exp: &Expr, env: &SymbolEnv, statement_idx: usize
 						// unsupported capture
 						diagnostics.push(Diagnostic {
 							level: DiagnosticLevel::Error,
-							message: format!("Can't capture '{}' of type '{}' from preflight", symbol.name, t),
+							message: format!("Cannot reference '{}' of type '{}' from an inflight context", symbol.name, t),
 							span: Some(symbol.span.clone()),
 						});
 					}
--- a/libs/wingc/src/capture.rs
+++ b/libs/wingc/src/capture.rs
@@ -156,6 +156,7 @@ pub fn scan_for_inflights_in_expression(expr: &Expr, diagnostics: &mut Diagnosti
 			if let Phase::Inflight = func_def.signature.flight {
 				let mut func_captures = func_def.captures.borrow_mut();
 				assert!(func_captures.is_none());
+				assert!(func_def.statements.env.borrow().is_some()); // make sure env is defined
 				*func_captures = Some(collect_captures(scan_captures_in_inflight_scope(&func_def.statements, diagnostics)));
 			}
 		}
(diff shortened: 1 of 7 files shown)


libs/wingc/src/parser.rs
```diff
@@ -470,6 +470,9 @@ impl Parser<'_> {
 	}
 
 	fn build_nested_identifier(&self, nested_node: &Node) -> DiagnosticResult<Reference> {
+		if nested_node.has_error() {
```
eladb, line 473: For reference, the test `missing_semicolon.w` yields the following parse tree (see the `ERROR` node under `nested_identifier`):

```
source [0, 0] - [10, 0]
  short_import_statement [0, 0] - [0, 12]
    module_name: identifier [0, 6] - [0, 11]
  expression_statement [2, 0] - [8, 3]
    call [2, 0] - [8, 2]
      caller: reference [2, 0] - [6, 13]
        nested_identifier [2, 0] - [6, 13]
          object: call [2, 0] - [4, 2]
            caller: reference [2, 0] - [2, 13]
              nested_identifier [2, 0] - [2, 13]
                object: reference [2, 0] - [2, 2]
                  identifier [2, 0] - [2, 2]
                property: identifier [2, 3] - [2, 13]
            args: argument_list [2, 13] - [4, 2]
              positional_argument [2, 14] - [4, 1]
                inflight_closure [2, 14] - [4, 1]
                  parameter_list: parameter_list [2, 23] - [2, 30]
                    parameter_definition [2, 24] - [2, 29]
                      name: identifier [2, 24] - [2, 25]
                      type: builtin_type [2, 26] - [2, 29]
                  type: builtin_type [2, 32] - [2, 35]
                  block: block [2, 39] - [4, 1]
                    expression_statement [3, 2] - [3, 32]
                      call [3, 2] - [3, 31]
                        caller: reference [3, 2] - [3, 7]
                          nested_identifier [3, 2] - [3, 7]
                            object: reference [3, 2] - [3, 3]
                              identifier [3, 2] - [3, 3]
                            property: identifier [3, 4] - [3, 7]
                        args: argument_list [3, 7] - [3, 31]
                          positional_argument [3, 8] - [3, 15]
                            string [3, 8] - [3, 15]
                          positional_argument [3, 17] - [3, 30]
                            string [3, 17] - [3, 30]
                              template_substitution [3, 25] - [3, 29]
                                reference [3, 27] - [3, 28]
                                  identifier [3, 27] - [3, 28]
          ERROR [6, 0] - [6, 2]
            identifier [6, 0] - [6, 2]
          property: identifier [6, 3] - [6, 13]
      args: argument_list [6, 13] - [8, 2]
        positional_argument [6, 14] - [8, 1]
          inflight_closure [6, 14] - [8, 1]
            parameter_list: parameter_list [6, 23] - [6, 31]
              parameter_definition [6, 24] - [6, 30]
                name: identifier [6, 24] - [6, 25]
                type: builtin_type [6, 27] - [6, 30]
            type: builtin_type [6, 33] - [6, 36]
            block: block [6, 40] - [8, 1]
              expression_statement [7, 2] - [7, 32]
                call [7, 2] - [7, 31]
                  caller: reference [7, 2] - [7, 7]
                    nested_identifier [7, 2] - [7, 7]
                      object: reference [7, 2] - [7, 3]
                        identifier [7, 2] - [7, 3]
                      property: identifier [7, 4] - [7, 7]
                  args: argument_list [7, 7] - [7, 31]
                    positional_argument [7, 8] - [7, 15]
                      string [7, 8] - [7, 15]
                    positional_argument [7, 17] - [7, 30]
                      string [7, 17] - [7, 30]
                        template_substitution [7, 25] - [7, 29]
                          reference [7, 27] - [7, 28]
                            identifier [7, 27] - [7, 28]
```

Chriscbr: Huh, I didn't know about the `has_error` API - TIL.

I wonder if we could cover more parsing errors if we check entire subtrees for errors instead of individual nodes perhaps?

```diff
diff --git a/libs/wingc/src/parser.rs b/libs/wingc/src/parser.rs
index 561e6459..a120b18f 100644
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -97,7 +97,7 @@ impl Parser<'_> {
        }

        fn check_error<'a>(&'a self, node: Node<'a>, expected: &str) -> DiagnosticResult<Node> {
-               if node.is_error() {
+               if node.has_error() {
                        self.add_error(format!("Expected {}", expected), &node)
                } else {
                        Ok(node)
```

libs/wingc/src/capture.rs
```diff
@@ -239,32 +240,41 @@ fn scan_captures_in_expression(exp: &Expr, env: &SymbolEnv, statement_idx: usize
...
+						// capture as an immutable data type (primitive/collection)
+						res.push(Capture {
+							object: symbol.clone(),
+							kind: CaptureKind::ImmutableData,
+						});
+					} else {
+						// unsupported capture
+						diagnostics.push(Diagnostic {
+							level: DiagnosticLevel::Error,
+							message: format!("Can't capture '{}' of type '{}' from preflight", symbol.name, t),
```
Chriscbr, line 271: I might reword this to `Unsupported capture` to make it clear that this isn't necessarily user error

Chriscbr (approved):

libs/wingc/src/capture.rs
eladb: Changed to:

```
Cannot reference '{}' of type '{}' from an inflight context
```

libs/wingc/src/parser.rs
eladb: Trying this out

eladb: Well well it seems to have found a bug in one of our tests! Nice!

https://github.com/winglang/wing/pull/1100/commits/8d79023290f077225eee4951679e5ed52b95cade

PR #1108, opened 2023-01-10
chore: fork-based workflows

### LINKED ISSUE #256: Cannot submit PRs from forks

MarkMcCulloh:
## NOTE: This PR is not mergeable without admin override

Because the workflows won't run until merged to main. My fork at https://github.com/MarkMcCulloh/wing has example of the workflow running on a fork (AKA a place that doesn't have access to all the secrets). I propose an admin force merges this PR.

### What's going on here:
- build workflow now uses pull_request_target in addition to push (for `main` only)
- Ensured PROJEN_GITHUB_TOKEN is only used in places that would have access to it (pushes to main on our repo)
  - GITHUB_TOKEN should have all the permissions needed to do what we need in forks
- Simplified the self-mutation logic and verified it works in a fork. Eventually, the self mutator won't be specific to wingsdk :)

Fixes #256

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,13 @@
 name: Build
 
 on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
   push:
+    branches:
+      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


eladb (approved):

MarkMcCulloh:
@eladb Would you be able to force-merge this to bypass checks? I'm ready to revert if needed :)

Issue #1117, opened 2023-01-11
Mutable container type cannot be inferred


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

```js
let arr: MutArray<str> = [1,2,3];
```

(same for `MutMap` and `MutSet`)

### I expected this:

That `[1,2,3]` will be inferred as a mutable array

### Instead, this happened:

```
Expected type to be one of "MutArray<num>", but got "Array<num>" instead
```

### Component:

Compiler

### Anything else?

This works:

```js
let arr = MutArray<str>[1,2,3];
```

So this is a low-priority issue.

eladb:
See @yoav-steinberg's comment here: https://github.com/winglang/wing/pull/1118#discussion_r1067001218

In a sense, we want to allow mutable containers to be initialized from immutable (implicit casting).

MarkMcCulloh:
So if we have implicit casting, we can do this?

```
let a = [1, 2, 3];
let b: MutArray<num> = a;
b.push(4);
```

Because either:
1. `b` is a "mutable pointer" to `a`'s data, basically defeating the purpose of `a` being immutable
2. `b` is a copy of `a`

Chriscbr:
2 cents - I feel like the original compiler error is correct here.

Instead of:
```js
let arr: MutArray<num> = [1,2,3];
```

(this is surprising... [1, 2, 3] could be initializing an immutable or mutable array? 😱)

Why not instead write:

```js
let arr = [1,2,3].to_mut(); // inferred: MutArray<num>
```

eladb:
OK, I am happy to close this as "no fix". I like the explicit approach and `.to_mut()` makes things reasonably easy.

@yoav-steinberg @staycoolcall911 any objections to close this for now?

staycoolcall911:
No objections

yoav-steinberg:
Sure. Lets live a bit with the `.to_mut()` approach and see how it goes.

eladb:
Closing for now!

staycoolcall911:
I'm guessing we need to add `.to_mut()` to our `Array`, `Set`, `Map` std objects API - adding here #1116

PR #1118, opened 2023-01-11
feat: mutable arrays and sets

### LINKED ISSUE #663: Container Types: MutArray
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#112-container-types

### LINKED ISSUE #852: Container Types: MutSet
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#112-container-types

eladb:
Adds support for mutable arrays and sets.

Misc:
- Add `Set.has()`, `MutSet.clear()` and `MutSet.delete()`.
- Add tests for all std APIs.
- Add some invalid tests for partial equality.

Resolves #852 
Resolves #663 

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -390,14 +390,18 @@ impl JSifier {
 				)
 			}
 			ExprKind::ArrayLiteral { items, .. } => {
-				format!(
-					"Object.freeze([{}])",
-					items
-						.iter()
-						.map(|expr| self.jsify_expression(expr, phase))
-						.collect::<Vec<String>>()
-						.join(", ")
-				)
+
+				let item_list = items
+					.iter()
+					.map(|expr| self.jsify_expression(expr, phase))
+					.collect::<Vec<String>>()
+					.join(", ");
+
+				if is_mutable_collection(expression) {
+					format!("[{}]", item_list)
+				} else {
+					format!("Object.freeze([{}])", item_list)
+				}
 			}
 			ExprKind::StructLiteral { fields, .. } => {
 				format!(
@@ -420,14 +424,17 @@ impl JSifier {
 				)
 			}
 			ExprKind::SetLiteral { items, .. } => {
-				format!(
-					"Object.freeze(new Set([{}]))",
-					items
-						.iter()
-						.map(|expr| self.jsify_expression(expr, phase))
-						.collect::<Vec<String>>()
-						.join(", ")
-				)
+				let item_list = items
+					.iter()
+					.map(|expr| self.jsify_expression(expr, phase))
+					.collect::<Vec<String>>()
+					.join(", ");
+
+				if is_mutable_collection(expression) {
+					format!("new Set([{}])",	item_list)					
+				} else {
+					format!("Object.freeze(new Set([{}]))",	item_list)					
+				}
 			}
 			ExprKind::FunctionClosure(func_def) => match func_def.signature.flight {
 				Phase::Inflight => self.jsify_inflight_function(func_def),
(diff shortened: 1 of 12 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -811,6 +823,7 @@ impl<'a> TypeChecker<'a> {
 
 				let element_type = match *container_type {
 					Type::Array(t) => t,
+					Type::MutArray(t) => t,
```
yoav-steinberg, line 826: I wonder if we want to support literals defined at mutable. Maybe all container literals should be of the immutable variant, while allowing them to be assigned to a mutable variant:
```wing
let a: MutArray<num> =  // Here we're defining a mutable container
  [1,2,3];  // And assigning an immutable value to it during initialization.
```
This can also be left for the future and we can just "let" mutable literals exist for now...

libs/wingc/src/type_check.rs
```diff
@@ -86,8 +86,10 @@ pub enum Type {
 	Boolean,
 	Optional(TypeRef),
 	Array(TypeRef),
+	MutArray(TypeRef),
```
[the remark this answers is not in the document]
eladb: Cool!

> `let a1 = MutArray[1,2,3];`

I didn't even know this was something we support.

libs/wingc/src/type_check.rs
eladb: I think we would want both options.

This should work:

```js
let x = MutArray<num>[1,2,3];
```

And also this:

```js
let x: MutArray<num> = [1,2,3];
```

I've created #1117 to cover the second case.

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
eladb: OK, fixed.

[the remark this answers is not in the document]
staycoolcall911: It's `MutArray<num>[1,2,3]`

libs/wingsdk/test/target-sim/queue.test.ts
```diff
@@ -2,7 +2,7 @@ import * as cloud from "../../src/cloud";
 import { SimApp, Testing } from "../../src/testing";
 import { listMessages } from "./util";
 
-jest.setTimeout(10_000); // 5 seconds
```
Chriscbr, line 5: How about we change the default test timeout for all SDK tests? (requires changing projen file) We could probably just add `--testTimeout=300000` to the command

libs/wingc/src/jsify.rs
```diff
@@ -710,3 +717,11 @@ impl JSifier {
...
 	}
 }
+
+fn is_mutable_collection(expression: &Expr) -> bool {
+    if let Some(evaluated_type) = expression.evaluated_type.borrow().as_ref() {
+			evaluated_type.is_mutable_collection()
+		} else {
+			false
+		}
+}
```
Chriscbr, line 727: (style) I think it makes sense to associate these kinds of utility function with the target type, but I'll let the king of Rust style @yoav-steinberg weigh in:

```rs
impl Expr {
	fn type(&self) -> TypeRef {
		if let Some(evaluated_type) = expression.evaluated_type.borrow().as_ref() {
			evaluated_type
		} else {
			panic!("expr.type() was called before the expression's type was evaluated");
		}
	}
	fn is_mutable_collection(&self) -> bool {
		this.type().is_mutable_collection()
	}
}
```

Chriscbr (commented):
LGTM but I'll let Yoav / Mark give a final pass 🙂

MarkMcCulloh (approved):
+1 to Chris' feedback, otherwise lookin good

libs/wingsdk/test/target-sim/queue.test.ts
eladb: https://github.com/winglang/wing/issues/1120

libs/wingc/src/jsify.rs
eladb: I'll fast-follow with this after @yoav-steinberg gives his blessing. I rather get this PR merged first if that's okay.

PR #1122, opened 2023-01-11
fix(compiler): using `print()` in bad places causes panics

### LINKED ISSUE #1062: using `print()` in bad places causes compiler panics
### Community Note

### I tried this:

I tried writing some invalid code:

Example 1:
```
print("hey").get("x");
```

Example 2:
```
let x = "my name is ${print("chris")}";
```

Example 3:
```
let x = 5 + print("hello");
```

### I expected this:

I expected `wing compile` to show me relevant diagnostics, like `Expression must be a class or resource instance to access property "get", instead found type "void"`.

### Instead, this happened:

Instead, when I was trying this in VS Code my language server crashed due to compiler panics. All of the examples panic with an error like:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libs/wingc/src/type_check.rs:1691:37
```

### Component:

Compiler

### Wing Version:

0.4.59

### Node.js Version:

19.2.0

### Platform(s):

MacOS

### Anything else?

I think the issue is related to how the compiler models return types as `Option<Type>` / `Option<TypeRef>`. When `type_check_exp` is called on an expression like `print("hey")`, the void type is modeled as `Option::None`, so `None` is the value that is returned. When other parts of the type checker calls `unwrap()` on the result of `type_check_exp`, they all panic.

Perhaps we could model a `Void` option inside the `Type` enum?

Chriscbr:
Fixes #1062

The implementation adds a new type `void` to our type system, used to represent the return type of functions that do not return a value. This type is not user-constructible (you cannot create a `void` value) and the `void` type is only equal to itself during type checking. In doing so we're able to remove a _lot_ of `unwrap` calls that corresponded to compiler panics. I've included a few examples in the `print_invalid` unit test.


examples/tests/invalid/print_invalid.w
```diff
@@ -0,0 +1,8 @@
+print("hey").get("x");
```
eladb, line 1: wonderful!

eladb (commented):
Looks great to me

MarkMcCulloh (approved):

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -10,7 +10,7 @@ use super::UnsafeRef;
 pub struct SymbolEnv {
 	pub(crate) ident_map: HashMap<String, (StatementIdx, SymbolKind)>,
 	parent: Option<UnsafeRef<SymbolEnv>>,
-	pub return_type: Option<TypeRef>,
+	pub return_type: TypeRef,
```
Chriscbr, line 13: I feel like this might need some further modeling -- in some cases a SymbolEnv corresponds to an assignable block of code, like a function body (so we use this `return_type` field for type checking) but in other cases it doesn't correspond to a block of code (e.g. the environment inside a class definition). For now I inserted `void` where we previously used `None` in dummy `SymbolEnv` instances

MarkMcCulloh: I don't think your change made the ambiguity any worse, but I would definitely agree it needs fourth modeling

Chriscbr:
I'll wait for @yoav-steinberg's blessing - added a `pr/do-not-merge` label in the mean time.

examples/tests/invalid/print_invalid.w
```diff
@@ -0,0 +1,8 @@
...
+// ^^^^^^^^^ Expression must be a class or resource instance to access property "get", instead found type "void"
+
+let x = "my name is ${print("mister cloud")}";
+//      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected type to be one of "str,num", but got "void" instead
+
+let x = 5 + print("hello");
+//          ^^^^^^^^^^^^^^ Expected type to be "num", but got "void" instead
```
yoav-steinberg, line 8: Maybe change the test name to something like "can't use void in expressions" and add another test that verifies the same thing with a custom function (not `print`):
```wing
returns_nothing = () {};
let x = returns_nothing();
```

yoav-steinberg (approved):
I love this! I actually had a branch doing almost exactly the same change which I never used.

libs/wingc/src/type_check.rs
```diff
@@ -580,26 +594,26 @@ impl<'a> TypeChecker<'a> {
 	}
 
 	// Validates types in the expression make sense and returns the expression's inferred type
-	fn type_check_exp(&mut self, exp: &Expr, env: &SymbolEnv, statement_idx: usize) -> Option<TypeRef> {
-		let t = match &exp.kind {
+	fn type_check_exp(&mut self, exp: &Expr, env: &SymbolEnv, statement_idx: usize) -> TypeRef {
+		let t: TypeRef = match &exp.kind {
```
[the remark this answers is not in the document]
Chriscbr: Ah, I think I just temporarily added this to force a type error while refactoring this.. will remove (nice catch)

examples/tests/invalid/print_invalid.w
Chriscbr: Good idea

Chriscbr: BTW, found a bug, as it was possible to assign void to a variable - fixed and added it to the example.

PR #1133, opened 2023-01-11
chore: move rfcs to docs

MarkMcCulloh:
Random idea. I don't see why the RFCs should be separate from the rest of the documentation, it'd be nice to be able to reference it from the beautiful rendered site.

I renamed the RFCs based on date so that they're ordered. Seems like we got away from doing "RFC Numbers" but I still wanted a way to order them somehow. Also separated the "stories" because they are kinda different from the others.


eladb (approved):
Hell yeah!

Chriscbr:
Great idea 👍

PR #1144, opened 2023-01-12
feat: mutable and immutable maps

### LINKED ISSUE #98: Container Types: MutMap
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#112-container-types

### LINKED ISSUE #1129: Unable to capture both resources and data
### Community Note

### I tried this:

```js
bring cloud;

let a = ["hello", "world"];

new cloud.Function(inflight (event: str): str => {
  print(a.at(1));
  bucket.put("hello.txt", a.at(0));
});
```

### I expected this:

To work

### Instead, this happened:

Error:

```
preflight error: Unexpected identifier

  note: intermediate javascript code:
           },
         }
>>       data: {
           a: a,
         }
```

### Component:

Compiler

### Wing Version:

0.4.77

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

This is because we are capturing _both_ immutable data (`a`) and a resource (`bucket`) and the generated javascript doesn't add a `,` between the fields.

### LINKED ISSUE #1143: Captures are not recognized within a binary operator
### Community Note

### I tried this:

```js
let a = [1,2,3];
let h = inflight () => {
  assert(a.length == 3);
};
```

### I expected this:

The array `a` should be captured.

### Instead, this happened:

This is the inflight output (`a` is not bound):

```js
const h = new $stdlib.core.Inflight(this, "$Inflight1", {
  code: $stdlib.core.NodeJsCode.fromFile(require('path').resolve(__dirname, "proc.b13bc27d12233b0ccf054de44bb4a6c7a34e8272f2f4165e55b9efe0c87918b1/index.js")),
  bindings: {
    
    
  }
});
```

### Component:

Compiler

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

eladb:
Add a `MutMap<T>` type for mutable maps, and the APIs for `Map<T>`.

Changed how maps are jsified. Instead of JS objects, they are now JS `Map` objects. This required a different serialization logic.

Resolves #98

I ran into a few issues along the way:

- Fixed #1143 where captures were not propagating through binary operators
- Converted some of the tests to use `assert` instead of `print` inside inflight closures in preparation for running those tests during build.
- Add a negative test to verify that mutables cannot be captured
- In `capture.rs:250`, ignore the case where an identifier cannot be found (it used to panic). A diagnostics is already emitted by the type checker in this case.
- Discovered that serialization was not really working for `Set` because the VM that executes preflight is not the same one as the one where the SDK comes from. So propagated the types from the outer VM (in `compile.ts`). To make sure this is discovered, made serializer explicitly support the various types and throw an error in case it couldn't serialize.
- Do not propagate `console` into inflight VMs because they override `console.log` and that reflects on the outer scope.
- Fixed #1129 where resources and data cannot be captured together (and added a test).


examples/tests/valid/capture_collections.w
```diff
@@ -2,15 +2,18 @@ bring cloud;
 
 let arr = ["hello", "world"];
 let my_set = {"my", "my", "set"};
+let my_map = {"hello": 123, "world": 999};
 
 let handler = inflight (s: str): str => {
-  print(arr.at(0));
-  print(arr.at(1));
-  print("size=${arr.length}");
+  assert(arr.at(0) == "hello");
```
MarkMcCulloh, line 8: 👍 great idea

apps/wing/src/commands/compile.ts
```diff
@@ -107,6 +107,19 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
     },
     __dirname: workDir,
     __filename: artifactPath,
+
+    // since the SDK is loaded in the outer VM, we need these to be the same class instance,
+    // otherwise "instanceof" won't work between preflight code and the SDK. this is needed e.g. in
+    // `serializeImmutableData` which has special cases for serializing these types.
```
MarkMcCulloh, line 113: Would it be better to solve this by not using `require.resolve` above to preload wingsdk? I think we could instead override `require` in this VM to resolve from the current dir.

We may need to do something like that anyways once we have more complicated deps.

eladb: That's a good idea. Since this can be a bit messy I'll do it in a subsequent commit.

eladb: https://github.com/winglang/wing/issues/1148

libs/wingsdk/src/std/map.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Returns the number of elements in the map.
+   */
+  public get size(): number {
+    throw new Error("Abstract");
+  }
+
+  /**
+   * Returns a specified element from the map. If the value that is associated to the provided key
+   * is an object, then you will get a reference to that object and any change made to that object
+   * will effectively modify it inside the map.
```
Chriscbr, line 19: (thinking about docs) What counts as an "object" in Wing? (we want to make sure we are not leaking javascript impl details to the user)

libs/wingsdk/src/core/internal.ts
```diff
@@ -69,13 +69,38 @@ ${Object.entries(clients)
...
+function serializeImmutableData(obj: any): string {
+  switch (typeof obj) {
+    case "string":
+    case "boolean":
+    case "number":
+      return JSON.stringify(obj);
+
+    case "object":
+      if (Array.isArray(obj)) {
+        return JSON.stringify(obj);
```
Chriscbr, line 81: I think this should be a recursive call to `serializeImmutableData` (what if it's an Array of Maps etc.)

libs/wingsdk/src/core/internal.ts
```diff
@@ -69,13 +69,38 @@ ${Object.entries(clients)
...
+      }
+
+      if (obj instanceof Set) {
+        return `new Set(${JSON.stringify(Array.from(obj))})`;
+      }
+
+      if (obj instanceof Map) {
+        return `new Map(${JSON.stringify(Array.from(obj))})`;
+      }
```
Chriscbr, line 95: Recursion for these cases as well? (see other comment)

libs/wingsdk/src/core/internal.ts
eladb: 🤦 of course!

But this naturally brings up the fact that separating `data` and `resources` is artificial and won't hold water because it should be absolutely possible to capture
an array of resources:

```js
let arr = [new cloud.Bucket(), new cloud.Bucket()];
  for b in arr {
    b.put("file.txt", "hello");
  }
});
```

So we need a unified serialization model.

I'll open a bug and work on this in a subsequent commit.

eladb: https://github.com/winglang/wing/issues/1156

Chriscbr: I see, you're totally right. Sounds like good merit for joining them 👍

eladb: Updated the code to serialize data recursively

libs/wingsdk/src/core/internal.ts
eladb: Done

libs/wingsdk/src/std/map.ts
eladb: I think this is okay for now...

libs/wingsdk/test/target-sim/immutable-capture.test.ts
```diff
@@ -84,3 +84,15 @@ makeCaptureTest({
...
+  name: "map",
+  in: Object.freeze(
+    new Map([
+      ["foo", 123],
+      ["bar", 456],
+    ])
+  ),
+  out: "{ has_foo: this.my_capture.has('foo'), size: this.my_capture.size, get_bar: this.my_capture.get('bar') }",
+  expected: '{"has_foo":true,"size":2,"get_bar":456}',
+});
```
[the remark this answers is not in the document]
eladb: Renamed.

libs/wingc/src/capture.rs
```diff
@@ -246,46 +246,59 @@ fn scan_captures_in_expression(
...
+								res.push(Capture {
+									object: symbol.clone(),
+									kind: CaptureKind::ImmutableData,
+								});
+							} else {
+								// unsupported capture
+								diagnostics.push(Diagnostic {
+									level: DiagnosticLevel::Error,
+									message: format!(
+										"Cannot reference '{}' of type '{}' from an inflight context",
```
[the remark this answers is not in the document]
eladb: It seems like this is it the only diagnostics with quotes... So I guess we are using single quotes?

examples/tests/valid/capture_containers.w
```diff
@@ -0,0 +1,22 @@
...
+  assert(my_set.has("my"));
+  assert(my_set.size == 2);
+
+  assert(my_map.has("world"));
+  assert(my_map.size == 2);
+
+  assert(arr_of_map.at(0).has("bang"));
+};
+
+new cloud.Function(handler) as "test";
```
eladb, line 22: `as "test"` here is compliant with https://github.com/winglang/wing/issues/1145

libs/wingsdk/src/std/set.ts
```diff
@@ -18,6 +18,7 @@ export class ImmutableSet {
    * Returns a boolean indicating whether an element with the specified value exists in the set.
    * @param value The value to test for presence in the Set object.
    * @returns Returns `true` if an element with the specified value exists in the set; otherwise `false`.
+   * @macro "this.has(value)"
```
eladb, line 21:
```suggestion
```

Chriscbr (approved):

Issue #493, opened 2022-11-08
Story: Infrastructure Customizations


eladb:
### Summary

Apply non-functional infrastructure concerns on Wing apps using declarative, policy-driven syntax.

### Feature Spec

(hypothetical readme)

You can use **wing policy** files to apply infrastructure policies to your wing applications during compilation.

Use the `--policy` switch in the compiler to tell the compiler which policy to use. For example:

```sh
$ wing compile hello.w --target tf-aws --policy my-policy.wp
```

Where `my-policy.wp` is (this is just a mock up):

```css
resource aws_s3_bucket * {
  ...

  versioning {
    enabled = true
  }
}
```

The policy file is a bit like a CSS file. It *selects* which elements in the Terraform output to work on and then applies a bunch of attributes on it. This rule selects all `aws_s3_bucket` resources (hence the `*`) and enables versioning on the bucket.

### Use Cases

* Deploy this to AWS and make all the buckets encrypted
* Put all the resources in side the VPC
* I want this app to be HIPPA-compliant
* Set the property "foo_bar" of resource "s3_bucket.bufol.dkkdjf" to "false"
* Tag all the resources with this label
* Don't allow any IAM policy to perform `s3:putObject` operations.

hasanaburayyan:
Once this is implemented we will need to update the interim solution in https://github.com/winglang/wing/pull/1162 (azure location)

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.8](https://github.com/winglang/wing/releases/tag/v0.5.8).

PR #1162, opened 2023-01-16
feat(sdk): azure location

hasanaburayyan:
Workaround to read azure location from environment variable.


apps/wing/src/commands/compile.ts
```diff
@@ -92,6 +94,10 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
   const artifact = await readFile(artifactPath, "utf-8");
   log("artifact: %s", artifact);
 
+  let appProps = {
```
eladb, line 97:
```suggestion
  const appProps = {
```

apps/wing/src/index.ts
```diff
@@ -51,6 +51,10 @@ async function main() {
       "Target platform (options: 'tf-aws', 'tf-azure', 'sim')",
       "tf-aws"
     )
+    .option(
```
eladb, line 54: I am not entirely comfortable with exposing this as a first-class option because different targets (in the future) might have a very diverse (and distinct) set of settings, and I don't want this to become the union of all settings of all targets. I won't be a good DX.

Since this type of setting is target-specific, I feel it falls under the category of #493 (which we don't have at the moment).

Here's what I suggest as a temporary step:
1. Have the Azure target read the location from an environment variable `AZURE_LOCATION` and throw a descriptive error if it's not defined (also add some documentation in the --help of `wing compile`).
1. Add a comment to #493 describing this temporary workaround, so that we remember to address this when we introduce infra policies.

I think that eventually the DX is going to be something like a config file that you will pass to the compiler with target-specific settings (which will likely be passed into the app).

For now, an environment variable is likely a good enough solution.

eladb:
Also, title should reflect as much as possible the value to the user, not the code change.

So in this case, I would go with:

```
feat(sdk): allow controlling azure location
```

apps/wing/src/index.ts
hasanaburayyan: Yea reviewing back I think this makes total sense to just use an environment variable as a work around, since we just need someway to do this until we have Policy as Infrastructure.

libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -45,7 +45,15 @@ export class App extends CdktfApp implements IApp {
       ...props,
       customFactory: props.customFactory ?? new PolyconFactory(),
     });
-    this.location = props.location;
```
eladb, line 48: missing test

eladb (approved):

eladb:
@hasanaburayyan the PR description needs a refresh

PR #1163, opened 2023-01-16
feat: toddler generics

MarkMcCulloh:
Enhanced the "baby generics" feature:

- Use JSII metadata to denote type params in class
- Use empty class type instead of `any` for placeholder param

This makes the generics feature a bit more explicit and rigorous. It also should allow for multiple type arguments per class. There should not be a noticeable functional change introduced in this PR.


libs/wingc/src/type_check.rs
```diff
@@ -1631,82 +1636,111 @@ impl<'a> TypeChecker<'a> {
...
 		let original_type = env
 			.lookup_nested_str(original_fqn, true, None)
 			.unwrap()
 			.as_type()
 			.unwrap();
 		let original_type_class = original_type.as_class().unwrap();
+		let original_type_params = if let Some(tp) = original_type_class.type_parameters.as_ref() {
+			tp
+		} else {
+			return self.general_type_error(format!("\"{}\" does not have type parameters", original_fqn));
```
eladb, line 1654: Maybe this only needs to be an only error if `type_params.length > 0` and I would also make the error a bit more descriptive.

(also, kind of feels like this is some kind of a compiler internal error and not an error the developer can do something about)

libs/wingc/src/type_check.rs
```diff
@@ -1631,82 +1636,111 @@ impl<'a> TypeChecker<'a> {
 	/// # Returns
 	/// The hydrated type reference
 	///
-	fn hydrate_class_type_arguments(&mut self, env: &SymbolEnv, original_fqn: &str, type_param: TypeRef) -> TypeRef {
+	fn hydrate_class_type_arguments(
+		&mut self,
+		env: &SymbolEnv,
+		original_fqn: &str,
+		type_params: Vec<TypeRef>,
```
eladb, line 1643: Feels like this should be deduced from the @typeparam?

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -466,11 +466,26 @@ impl<'a> JsiiImporter<'a> {
 		// Create the new resource/class type and add it to the current environment.
 		// When adding the class methods below we'll be able to reference this type.
 		debug!("Adding type {} to namespace", type_name.green());
+		let type_params = jsii_class.docs.as_ref().and_then(|docs| {
+			// get "@typeparam" from doc custom map
```
eladb, line 470: document the syntax of `@typeparam` here

libs/wingsdk/src/std/util.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Generic type argument. This type is replaced at compile time.
```
eladb, line 2: Should this be `@hidden` somehow so it cannot be instantiated?

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -527,20 +542,24 @@ impl<'a> JsiiImporter<'a> {
...
-						// so they have to be escaped in the original docstring.
-						if fqn.starts_with("`") && fqn.ends_with("`") {
-							&fqn[1..fqn.len() - 1]
-						} else {
-							fqn
-						}
-					})
-					.and_then(|fqn| self.jsii_types.find_interface(fqn))
-			});
+			let client_interface = jsii_class
```
eladb, line 545: Curious - what was the motivation to add `.add_ref()`s here?

libs/wingc/src/type_check.rs
MarkMcCulloh: True, this should actually be a panic. Will update

libs/wingc/src/type_check.rs
MarkMcCulloh: The actual type of our collection types are the built-in `Types::Array` and such. This is where the type params are actually supplied.

```
let d: Array<num> = [1, 2, 3];
```

So when hydrating the JSII classes, we need to gather the args from that type info.

libs/wingsdk/src/std/util.ts
MarkMcCulloh: Yes is should, I didn't put it at first because for some reason I thought it would break compilation. Will add.

libs/wingc/src/type_check/jsii_importer.rs
MarkMcCulloh: The reasoning is because rust yelled at me :)

The real reason (I think) is because the usage of closures causes the `docs` to get borrowed. This is normally fine, but above I do added another borrow so I needed to make sure we're only dealing with references here.

eladb (approved):

Issue #118, opened 2022-09-20
For Statement


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#27-for

@yoav-steinberg:
Partial support exists. Still missing many elements:
- [x] Type check that expression is an iterable.
- [ ] Support the non `in` format.

marciocadev:
The `for` statement example is weird

the wing code looks like it will print from 0 to 100 (or 99)
```ts (wing)
for item in 0..100 {
  print(item);
}
```

but the Typescript equivalent print from 99 to 0
```ts
// calling 0..100 does not allocate, just returns an iterator
function* iterator(lim) { let i = lim; while (i--) yield i; }
const iter = iterator(100);
for (const val of iter) {
  console.log(val);
}
```

I think it would be better if 

`for item in 0..100` iterates from 0 to 99

```ts
function* iterator(start, end) {
  let i = start;
  while (i < end) yield i++;
  while (i > end) yield--;
}
const iter = iterator(100, 0);
for (const val of iter) {
  console.log(val);
}
```

 and `for item in 100..0` iterates from 100 to 1


```ts
function* iterator(start, end) {
  let i = start;
  while (i < end) yield i++;
  while (i > end) yield i--;
}
const iter = iterator(0, 100);
for (const val of iter) {
  console.log(val);
}
```

Chriscbr:
@marciocadev Good catch, this looks to me like a bug in the spec. 🙂 Your proposal makes sense. would you like to submit a fix?

staycoolcall911:
@MarkMcCulloh added type checking for iterable in #1412, still left to support the non `in` format of for loops.

marciocadev:
Wow, iterator works now? Amazing

monadabot:
Congrats! :rocket: This was released in [Wing 0.8.35](https://github.com/winglang/wing/releases/tag/v0.8.35).

PR #1166, opened 2023-01-16
fix(docs): for statement equivalent typescript code

marciocadev:
The `for` statement equivalent typescript code when iterate between 0 and 100 it is printing in order from 99 to 0

```ts (wing)
for item in 0..100 {
  print(item);
}
```
generates
```bash
99
98
...
1
0
```

This change fixes that and makes the statement more meaningful

That way when `for item in 0..100` we get a iterate `0, 1, ... , 98, 99`
and `for item in 100..0` we get a iterate `100, 99, ... , 2, 1`

Reference #118 


docs/04-reference/winglang-spec.md
```diff
@@ -862,8 +862,12 @@ The loop invariant in for loops is implicitly `readwrite` and re-assignable.
 >   console.log(item);
 > }
 > // calling 0..100 does not allocate, just returns an iterator
-> function* iterator(lim) { let i = lim; while (i--) yield i; }
-> const iter = iterator(100);
+> function* iterator(start, end) {
+>   let i = lim;
```
Chriscbr, line 866: What is lim?

marciocadev: an mistake 😖

Chriscbr (approved):
Perfect, thanks

Issue #109, opened 2022-09-20
Reassignability (`var`)


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#16-reassignability

staycoolcall911:
Reassignability for variables was implemented in #1180 as `let var x = 5;`.
We still need to implement reassignability for class/resource members:
```ts
resource Foo {
  let bucket: cloud.Bucket;
  var name: str?;
}
```

ekeren:
update syntax, the `var` is a modifier and the `let` is not requires for class/resource members

ekeren:
```[tasklist]
### Tasks
- [ ] https://github.com/octo-org/octo-repo/issues/45
- [ ] Draft issue title
```

staycoolcall911:
Closed via #1196

PR #1180, opened 2023-01-17
feat: `var` declarations

Chriscbr:
This change introduces the `var` modifier, which indicates that a variable can be reassigned to. For example:

```ts
let var sum = 0;
for item in [1, 2, 3] {
  sum = sum + item;
}
```

Partially addresses #109 (only implements the modifier for variables, not for class/resource fields)

Also fixes a bug where expressions used in an interpolated string are not recognized as captures.


examples/tests/valid/vars.w
```diff
@@ -0,0 +1,7 @@
+var j = 5;
+j = j + 1;
+
+let handler = inflight(x: str): str => {
```
eladb, line 4: Can we add a test?

```js
var x: str?;
assert(x == nil);

x = "hello";
assert(x != nil);
```

eladb (approved):

eladb:
> The compiler doesn't yet check that let variables cannot be reassigned to, but this has not been an issue to date in preflight since it is translated into const in JavaScript code.

Do we have a tracking issue for this?

Chriscbr:
Not yet. Will create

Also I need to add a test that "var"  symbols cannot be captured (right?)

examples/tests/valid/vars.w
Chriscbr: Added an assertion-like test

docs/04-reference/winglang-spec.md
```diff
@@ -402,6 +402,8 @@ Mixing `protected` and `internal` is not allowed.
 
 Re-assignment to variables that are defined with `let` is not allowed in Wing.
 
+Variables can be reassigned to by using the `var` keyword instead of `let`.
+
 Re-assignment to class fields is allowed if field is marked with `readwrite`.
```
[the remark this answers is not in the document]
Chriscbr: Updated the spec to use `var` instead of `readwrite` everywhere... so far it looks okay, we can adjust / switch back of course at any point 🙂

Chriscbr:
Awaiting feedback from @yoav-steinberg

docs/04-reference/winglang-spec.md
```diff
@@ -1147,8 +1149,8 @@ resource Foo {
   field6: bool;
 
   // re-assignable class fields, read about them in the mutability section
-  readwrite field4: num;
-  readwrite field5: str;
+  var field4: num;
```
yoav-steinberg, line 1152: The following suggestion came up, instead of `var` change this to `let var`. And in class definitions do:
```wing
class {
  x: num;
  var y: num; // I'm reassignable
}
```
This creates some consistency where `var` is a modifier on the symbol meaning it's reassignable. It also adds some "ugliness" or friction meant to discourage the use of `var`. 
We should also add a separate issue for adding a compiler warning if you create a `var` and never reassign to it.

libs/wingc/src/capture.rs
```diff
@@ -328,7 +338,27 @@ fn scan_captures_in_expression(
 			res.extend(scan_captures_in_expression(lexp, env, statement_idx, diagnostics));
 			res.extend(scan_captures_in_expression(rexp, env, statement_idx, diagnostics));
 		}
-		ExprKind::Literal(_) => {}
+		ExprKind::Literal(lit) => match lit {
+			Literal::String(_) => {}
+			Literal::InterpolatedString(interpolated_str) => {
```
yoav-steinberg, line 343: Did you add interpolated string support here which was missing? This doesn't seem relevant directly to `var`.

libs/wingc/src/type_check.rs
```diff
@@ -1279,6 +1291,7 @@ impl<'a> TypeChecker<'a> {
 				self.type_check_exp(e, env, stmt.idx);
 			}
 			StmtKind::Assignment { variable, value } => {
+				// TODO: check if variable can be assigned to
 				let exp_type = self.type_check_exp(value, env, stmt.idx);
 				let var_type = self.resolve_reference(variable, env, stmt.idx);
```
yoav-steinberg, line 1296: I think this should be easy to do. `resolve_reference` can return the `kind` and we'll have it. If we don't do that change we just make sure we have an issue for it.

libs/wingc/src/capture.rs
Chriscbr: Correct, it was a missing case that I discovered while testing (bug). I'll mention it in the description

docs/04-reference/winglang-spec.md
eladb: +1 on 

```js
let var x = 12;
```

docs/04-reference/winglang-spec.md
```diff
@@ -402,11 +402,13 @@ Mixing `protected` and `internal` is not allowed.
 
 Re-assignment to variables that are defined with `let` is not allowed in Wing.
 
-Re-assignment to class fields is allowed if field is marked with `readwrite`.
+Variables can be reassigned to by using the `var` keyword instead of `let`.
```
eladb, line 405:
```suggestion
Variables can be reassigned to by using `let var` instead of `let`.
```

docs/04-reference/winglang-spec.md
```diff
@@ -833,7 +835,7 @@ The `if` statement is optionally followed by `elif` and `else`.
 
 `for..in` statement is used to iterate over a array or set.  
 Type annotation after an iteratee (left hand side of `in`) is optional.  
-The loop invariant in for loops is implicitly `readwrite` and re-assignable.
+The loop invariant in for loops is implicitly re-assignable (like `var`).
```
eladb, line 838:
```suggestion
The loop invariant in for loops is implicitly re-assignable (`var`).
```

docs/04-reference/winglang-spec.md
eladb: By the way, tangentially related: when you declare a var in a class, you either must define its value in the ctor (static analysis) or it must be optional (and then we go into the `nil` world)

examples/tests/invalid/capture_reassignable.w
```diff
@@ -0,0 +1,8 @@
+bring cloud;
+
+var x = 5;
```
eladb, line 3:
```suggestion
let var x = 5;
```

docs/04-reference/winglang-spec.md
yoav-steinberg: > By the way, tangentially related: when you declare a var in a class, you either must define its value in the ctor (static analysis) or it must be optional (and then we go into the nil world)

In general I'm for either having default values for **everything** optional or not or forcing everything to be inited in the ctor. In the latter case, even an optional needs to be inited: `init() {self.my_optional_num = nil; // or undefined or nothing...}`
In the former case we can have strict rules saying `num`s default to 0, `str`s default to "", etc.

skinny85: What about `let mut` (and `mut` for fields), following Rust?

eladb: We are intentionally separating between what we call "mutable" and "reassignable", which I believe is a source of confusion in Rust.

In the following example, `x` *cannot be reassigned*, but it is *mutable* (there's a way to mutate it's value).

```js
let x = new MyObject();
x.foo = 12;
```

So we don't want to use the term `mut` to represent reassinability.

Chriscbr: Updated the syntax to `let var`

libs/wingc/src/type_check.rs
Chriscbr: #1186

eladb (approved):

libs/tree-sitter-wing/test/corpus/statements.txt
```diff
@@ -35,6 +36,11 @@ x = 2;
         name: (reference (identifier))
         value: (number)
     )
+    (variable_definition_statement
+        (reassignable)
```
yoav-steinberg, line 40: I'm for being verbose about the field names in the grammar tests:
```suggestion
        reassignable: (reassignable)
```

libs/tree-sitter-wing/test/corpus/statements.txt
```diff
@@ -152,6 +159,11 @@ class A {
                 name: (identifier)
                 type: (builtin_type)
             )
+            (class_member
+                (reassignable)
```
yoav-steinberg, line 163:
```suggestion
                reassignable: (reassignable)
```

yoav-steinberg (approved):

PR #1181, opened 2023-01-17
fix: add line numbering to playground - with suggested changes

ogre14t:


eladb (approved):
Looks like this PR merges from main and into your branch. I don't think a PR is required for this.

PR #1183, opened 2023-01-17
fix: cross-platform wing cli shebang

### LINKED ISSUE #1157: Shebang in Wing CLI is not portable
### Community Note

### I tried this:

See https://github.com/winglang/wing/issues/1000#issuecomment-1382618051

https://github.com/winglang/wing/blob/8e4f7b816b23d8a0313297b7829a655d3255827d/apps/wing/bin/wing#L1

`-S` is not portable on all distros. Generally speaking, multiple args in shebang is not portable.

### I expected this:

The shebanged file runs with the given node arguments

### Instead, this happened:

`wing: exec: line 11: illegal option -S`

### Component:

Other

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

Linux

### Anything else?

Alpine Linux under WSL

MarkMcCulloh:
This simplified shebang should be able to run anywhere. The downside is the now-required `spawn`, but I think that's worth revisiting later.

Note: this PR **does not** confirm windows support (although windows may work now 😁), but it does resolve #1157

### DIFF
--- a/apps/wing/bin/wing
+++ b/apps/wing/bin/wing
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S node --experimental-wasi-unstable-preview1 --no-warnings
+#!/usr/bin/env node
 
 // Only certain builds and versions of Node support WASI.
 // We first assume this script is loaded in the correct runtime environment, and


Chriscbr:
> The downside is the now-required spawn

Where is this? Maybe there is another file that was supposed to be committed?

Chriscbr (approved):

apps/wing/bin/wing
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S node --experimental-wasi-unstable-preview1 --no-warnings
```
eladb, line 1: Just for my understanding, it seems like we can simply split these into two scripts. An entrypoint that runs with regular node and the child that's executed with wasm support. Right?

Feels like there is no real additional value to making this a reentrant script.

MarkMcCulloh: It is technically possible that wing is being run with node that has the flag enabled already. In that case, the logic in this file does a have a use. I also don't think it's worse than separating, but we could do benchmarking to see if there's a perf difference.

eladb: 🤷🏻‍♀️

PR #1184, opened 2023-01-17
chore: ensure nx project names match package names

MarkMcCulloh:
There was some strange behavior causing nx failures in local builds.

Note: Will have to force-merge this due to the change to the workflow file.


apps/wing/project.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "wing",
```
Chriscbr, line 2: @MarkMcCulloh If we are keeping the name of this directory as "wing" then it might just be easier to keep it as wing in this file (not sure if this is related to the current build error) though I'm okay with either

MarkMcCulloh: There seems to be an issue with nx where it has trouble if the "project name" is different from the package.json name, so yeah unfortunately this will have to match the icky package name until we can get wing

Chriscbr (approved):

PR #1192, opened 2023-01-18
fix: model subtyping in compiler

Chriscbr:
This PR doesn't introduce any functional changes. Implements a small refactoring to our compiler to model the "subtype" relationship directly in the type checker instead of using the PartialEq trait.


yoav-steinberg:
Note that the main goal is to be more explicit about what we're doing when comparing types or checking if a type is a subtype.

libs/wingc/src/type_check.rs
```diff
@@ -178,8 +178,25 @@ pub struct EnumInstance {
...
 }
 
-impl PartialEq for Type {
-	fn eq(&self, other: &Self) -> bool {
+trait Subtype {
+	/// Returns true if `self` is a subtype of `other`.
+	///
+	/// For example, `str` is a subtype of `str`, `str` is a subtype of
+	/// `anything`, `str` is a subtype of `Json`, `str` is not a subtype of
+	/// `numb`, and `str` is not a subtype of `void`.
```
yoav-steinberg, line 186:
```suggestion
	/// `num`, and `str` is not a subtype of `void`.
```

libs/wingc/src/type_check.rs
```diff
@@ -194,23 +211,23 @@ impl PartialEq for Type {
...
 				// If our parent is equal to `other` then treat both resources as equal (inheritance)
 				if let Some(parent) = l0.parent.as_ref() {
 					let parent_type: &Type = &*parent;
-					return parent_type.eq(other);
+					return parent_type.is_subtype_of(other);
 				}
 				false
 			}
 			(Self::Struct(l0), Self::Struct(_)) => {
 				// If we extend from `other` then treat both structs as equal (inheritance)
```
yoav-steinberg, line 227: Need to remove the usage of "equal" from all these comments.
```suggestion
				// If we extend from `other` then I'm a subtype of it (inheritance)
```

libs/wingc/src/type_check.rs
```diff
@@ -1715,7 +1761,7 @@ impl<'a> TypeChecker<'a> {
 					SymbolKind::Variable(v, kind) => {
 						// Replace type params in function signatures
 						if let Some(sig) = v.as_function_sig() {
-							let new_return_type = if sig.return_type == *original_type_param {
+							let new_return_type = if sig.return_type.is_subtype_of(original_type_param) {
```
MarkMcCulloh, line 1764: Technically I think this should be `is_same_type_as` (as well as other uses of `is_subtype_of` in this method). However, in practice it basically won't matter and `is_subtype_of` would just be slightly faster. Was that your thinking as well?

MarkMcCulloh (commented):
Makes sense to me, nice, will leave to Yoav for approval

libs/wingc/src/type_check.rs
Chriscbr: I wasn't too familiar with how this hydrate method works so I think I just overlooked it. Fixed

yoav-steinberg (approved):

Issue #1186, opened 2023-01-18
Compiler should error if `let` variable is reassigned to


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

In Wing, all variables are non-reassignable by default. This code should not compile:

```
let var x = 5;
x = 7;
```

### I expected this:

I expected a compile-time error (within the first phase of the compiler that generates preflight code)

### Instead, this happened:

I received a preflight error:

```
preflight error: Assignment to constant variable.

  note: intermediate javascript code:

     const x = 5;
>>   x = 7;
   }
   }
  note: run with `NODE_STACKTRACE=1` environment variable to display a stack trace
```

### Component:

Compiler

### Wing Version:

0.4.102

### Node.js Version:

19.2.0

### Platform(s):

MacOS

monadabot:
Congrats! :rocket: This was released in [Wing 0.4.121](https://github.com/winglang/wing/releases/tag/v0.4.121).

PR #1196, opened 2023-01-19
fix(compiler): no error when trying to reassign to a `let` variable

yoav-steinberg:
now this:
```wing
let x = 9;
x = 10;
```
produces something like: `Error at ../../examples/tests/invalid/reassing_to_nonreassignable.w:2:1 | Variable x is not reassignable `

Also correctly import reassignability (let/var) from JSII class properties.

see #1186


yoav-steinberg:
Note that this us currently failing because one of our tests attempt to write to an argument passed to a function, but there's currently no way of specifying if an argument is reassignable:
```wing
let f = (x: num) => {
  x = x +1; // This write fails because `x` is non reassignable
};
```
Should we suppor the following syntax to fix this? @eladb @Chriscbr  WDYT?
```wing
let f = (var x: num) => {
  x = x +1;
};
```

eladb (approved):

Chriscbr:
@yoav-steinberg I'd be okay adding that syntax later if there is demand for it, though I feel like it's generally safer for parameters to not be reassignable by default.

libs/wingc/src/type_check.rs
```diff
@@ -1458,7 +1500,10 @@ impl<'a> TypeChecker<'a> {
 						name: WING_CONSTRUCTOR_NAME.into(),
 						span: name.span.clone(),
 					},
-					SymbolKind::Variable(constructor_type, VariableKind::Let),
+					SymbolKind::Variable(VariableInfo {
+						_type: constructor_type,
+						kind: VariableKind::Let,
```
Chriscbr, line 1505: Add a helper perhaps?

SymbolKind::make_variable(constructor_type, true)

Chriscbr (approved):
One suggestion to reduce some verbosity - otherwise LGTM. I like the var argument idea as well 👍

yoav-steinberg:
Update: fixed everything related to to function arguments not being `var`. 
Also added support for modifying non-var class fields in `init` so we can initialize classes (this also caused test failures).
Also added support (and tests) for checking reassignablity of class fields.
Additionally added support for mutable JSII class fields by marking them as reassignable on import.

examples/tests/valid/reassignment.w
```diff
@@ -3,3 +3,25 @@ assert(x == 5);
...
+let r = new R();
+r.inc();
+assert(r.f == 2);
+
+let f = (var arg: num):num => {
+  arg = 0;
+  return arg;
+}
+
+assert(f(1) == 0);
```
Chriscbr, line 27: nit: It could be worth testing this case:
```suggestion

let y = 1;
assert(f(y) == 0);
assert(y == 1); // y is not modified, since Wing functions are pass-by-value

```

Chriscbr (approved):

eladb:
@yoav-steinberg is this ready to be merged?

examples/tests/valid/reassignment.w
yoav-steinberg: Why not, i'll add it.

yoav-steinberg:
> @yoav-steinberg is this ready to be merged?

Not yet, I woke up this morning having a feeling this might break:
```wing
resource R {
  f: num;
  init() {
    if true {
      this.f = 1; // Access a non reassignable field from constructor but not in upper most scope. Need to make sure this works!
    }
  }
}

PR #1034, opened 2023-01-04
feat(cli): pretty preflight errors

Chriscbr:
Display pretty preflight errors, instead of the usual NodeJS errors.

Before:
Screenshot 2023-01-03 at 10 17 49 PM

After:
Screenshot 2023-01-03 at 10 18 50 PM

*By submitting this pull request, I confirm that my contribution is made under the terms of the 
[Monada Contribution License](https://docs.winglang.io/terms-and-policies/contribution-license.html)*.

PR #1199, opened 2023-01-20
fix: playground crashes due to memory leaks

MarkMcCulloh:
wingc is no longer a binary, now only a library. The cli and playground use it via WASM and pointer fun.
So, far this seems to increase the speed of `wing compile ...` by ~80%! It unfortunately slows down the playground a little bit, but now it no longer has a huge memory leak and won't just break after a while.

Right now the logic for `wingcInvoke` is duplicated for the cli and playground. `wingc` should eventually have a wrapper npm package that includes this logic. 

Added an "example" for wingc, which allows a similar workflow to the previous "cargo run". No doc updates needed, as it can be accessed via `nx dev` like before.


libs/wingc/src/lib.rs
```diff
@@ -48,6 +48,50 @@ pub struct CompilerOutput {
...
+	let results = compile(source_file, output_dir);
+	if let Err(mut err) = results {
+		// Sort error messages by line number (ascending)
+		err.sort_by(|a, b| a.cmp(&b));
+		eprintln!(
+			"Compilation failed with {} errors\n{}",
+			err.len(),
+			err.iter().map(|d| format!("{}", d)).collect::<Vec<_>>().join("\n")
+		);
+	}
```
Chriscbr, line 92: Can we output these errors in some kind of structured format (eg. JSON) so that the responsibility of error formatting etc. can be deferred to whoever is consuming wingc? Right now some error formatting is done in the wingc (like this snippet), and some in the CLI (eg. #1034), and I feel like if we're treating wingc like a library, it will be easier to maintain if we keep all of the formatting outside.

apps/wing/src/commands/compile.ts
```diff
@@ -87,6 +83,9 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
   log("invoking wingc with importObject: %o", importObject);
   wasi.start(instance);
 
+  log(`invoking wingc_compile with: "${wingFile};${workDir}"`);
+  await wingcInvoke(instance, "wingc_compile", `${wingFile};${workDir}`);
```
Chriscbr, line 87: Extract `wingc_compile` as a constant

libs/wingc/src/lib.rs
MarkMcCulloh: Totally agreed, but I think it'd be okay to leave that for a follow-up (I'll make an issue). In this PR I'm pretty much exclusively changing how wingc is invoked and otherwise changing as little else as possible.

libs/wingc/examples/compile.rs
```diff
@@ -1,17 +1,19 @@
+// Compiles a source file and prints any errors to stderr.
```
eladb, line 1: Do we really use this entrypoint? Maybe just get rid of it so we don't have to continue to maintain it...

libs/wingc/src/lib.rs
```diff
@@ -48,6 +48,50 @@ pub struct CompilerOutput {
 	pub diagnostics: Diagnostics,
 }
 
+// Hack needed for some WASM related reason
```
eladb, line 51: Any additional info on this "hack"?

apps/wing/src/commands/compile.ts
```diff
@@ -154,3 +155,32 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
...
+  const argMemoryBuffer = new Uint8Array(
+    exports.memory.buffer,
+    argPointer,
+    bytes.byteLength
+  );
+  argMemoryBuffer.set(bytes);
+
+  exports[func](argPointer, bytes.byteLength);
+
+  exports.wingc_free(argPointer, bytes.byteLength);
```
eladb, line 185: should this be under `finally` or something?

eladb (commented):
This is very exciting!

eladb:
@MarkMcCulloh please update the PR title to reflect the bug you are fixing (for `fix`) or the feature you are implementing for (`feat`).

libs/wingc/examples/compile.rs
MarkMcCulloh: To support fast iteration cycles on development of wingc-specific features, it's super useful to have an entrypoint of some kind. The surface area is very small so I think it's worthwhile.

MarkMcCulloh:
@eladb afaik the title is mostly accurate for my intent. There isn't really a bug or a feature, more of a gap of how this was originally implemented. I could say `fix: improve performance of cli wasm interop` but that wasn't technically my goal.

libs/wingc/src/lib.rs
MarkMcCulloh: I can try to put some more details in the comment, but some of it is a guess. I'll see what I can do/lookup.

eladb:
The mental model should be aligned to the reader of the changelog - our user, not the implementor.

The guidance is that "fix" should be used when a bug is fixed (and therefore the title should say what was fixed) and "feat" is when a new capability is introduced (and then the title should describe what can I do now that I couldn't do previously).

If at this point there isn't a clear user facing impact, you can simply use `chore`.

One option is perhaps to think it as playground fix for the crashes.

MarkMcCulloh:
This is an implementation detail with user-facing impact with performance impact in the cli and crash fixes in the playground. I can choose the playground fix, but it's arbitrary.

eladb:
(Just taking the opportunity to align)

A "fix" title should represent the problem, not the solution. You are "fixing this issue".

So, rewrite to:

fix(playground): crashes due to memory leaks

MarkMcCulloh:
There is no playground scope. I added it to the PR but due to using `pull_request_target` it has to be merged first.

.github/workflows/pull-request-lint.yml
```diff
@@ -27,6 +27,7 @@ jobs:
             examples
             build
             repo
+            playground
```
eladb, line 30: While you are at it:

```suggestion
            vscode
            playground
```

libs/wingc/examples/compile.rs
eladb: OK

eladb (approved):
LGTM

libs/wingc/src/lib.rs
MarkMcCulloh: Thanks for bringing this up, because I looked into it more and saw I was doing the wrong thing. For WASI libraries, it's actually looking for an `_initialize` function to be exposed rather than _start. 
When creating a `cdylib` like we are here, an _initialize function is automatically added so I just deleted this function.

PR #1212, opened 2023-01-22
fix: invalid grammar for `var` in classes

eladb:
Fix the tree sitter grammer to match the parsing tree of `var` in `class` to be the same as `let`.

This is currently broken in our build.

Also - update the workflow files to *exclude* projects from build and test.


Chriscbr (approved):

.github/workflows/build.yml
```diff
@@ -126,15 +126,16 @@ jobs:
         uses: MansaGroup/nrwl-nx-action@v2
         with:
           targets: "build"
-          projects: "winglang,sdk,wing-api-checker,jsii-docgen,wing-playground"
```
eladb, line 129: It seems like this list of projects effectively builds all the projects. no?

MarkMcCulloh: It's specifically leaving out the native-based projects (vscode and lsp)

MarkMcCulloh: So once this is merged, it'll take a lot longer

.github/workflows/build.yml
```diff
@@ -126,15 +126,16 @@ jobs:
...
+          all: true
           args: "--configuration=release --output-style=stream --verbose"
 
       # wingc require wingsdk to be built first in order to run
       - name: Test
         uses: MansaGroup/nrwl-nx-action@v2
         with:
           targets: "test"
-          projects: "wingc,sdk"
+          all: true
```
MarkMcCulloh, line 137: this will cause wingc to do a full native build. I would suggest adding the specific new projects you want here instead.

.github/workflows/build.yml
eladb: I'll exclude them.

MarkMcCulloh: For the purpose of this PR, I don't think this line needs to be changed

MarkMcCulloh (changes requested):

.github/workflows/build.yml
eladb: added "exclude"

MarkMcCulloh:
side note: It would probably be useful to do workflow changes in a fork and preview it there first, otherwise you’ll have to wait for it to merge.

.github/workflows/build.yml
MarkMcCulloh: the test target for wingc runs cargo test, which uses the native platform target by default

eladb: As discussed, we are already running `wingc` tests.

MarkMcCulloh (approved):

Issue #1182, opened 2023-01-17
Cloud cost tracking


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Wing SDK now has CostTracker, a resource you can use to help you measure the costs of your cloud application.

Costs can be registered to methods like so:

CostTracker.of(this).assignCost(Bucket.put, 0.01);

Then, in unit tests you can measure the total costs through the Wing simulator:

assert(CostTracker.of(app).total() == 0.06);

When deployed to a cloud, the cost tracker will automatically aggregate costs across your running inflight code (in FaaS, containers, etc) and report it as a metric for your own use:

CostTracker.of(this).metric_total().on_threshold(5.00, inflight (value) => { /* stuff */ });

### Use Cases

Unit test how expensive different operations are in your cloud application, or how expensive your application is so far.

### Component

SDK

PatMyron:
Would build on https://github.com/infracost/infracost/ so costs that have already been mapped by them don't need to be re-mapped in every application

staycoolcall911:
Another take on cost tracking is to have the CLI print estimated monthly cost for the provisioned infrastructure upon compile:
```
> wing compile -t tf-aws app.w
Total estimated monthly cost for provisioned resources: $238
```

eladb:
https://arxiv.org/pdf/2309.04954.pdf

PR #1218, opened 2023-01-22
rfc: infrastructure customization through plugins

### LINKED ISSUE #493: Story: Infrastructure Customizations

hasanaburayyan:
Just initial thoughts so far.

The goal: provide a plugin system that allows users wing to provide non-functional requirements to their wing applications.

[Rendered Version](https://github.com/winglang/wing/blob/hasanaburayyan/wing-plugin-system/docs/05-rfcs/2023-01-21-wing-plugins.md)

Resolves #493


eladb:
I think the title of this RFC should be more user-focused. Focus on the reason why someone would use this feature (the problem), not the way it's implemented.

Plugins are the solution :-)

hasanaburayyan:
> I think the title of this RFC should be more user-focused. Focus on the reason why someone would use this feature (the problem), not the way it's implemented.
> 
> Plugins are the solution :-)

@eladb Okay this is a mental model I need to practice a bit 😄 so something like "rfc: apply non-functional requirements"?

eladb:
> rfc: apply non-functional requirements

Yes, exactly. The narrative should also be around this theme.

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+
+Implementing a simple wing plugin system
+
+## Background
+
+Developers using wing should whenever possible only care about functional aspects of an application's infrastructure. However, there must be a way to define non-functional requirements as that will be crucial for wide adoption. For instance, an organization may have specific security requirements to adhere to for compliance and regulation sake, such as encryption at rest, versioning, replication etc... 
+
+Applying these non-functional requirements will mean that the wing compiler should provide the ability for users to write plugins that can hook into the compilation process and apply business logic to resources.
+
+## Proposal
```
eladb, line 20: Try to rephrase the first section as a topic in the Wing documentation as if this capability was already available ("work backwards" from the user). 

For example, see how [this issue] is phased.

[this issue]: https://github.com/winglang/wing/issues/1182#issue-1536336793

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+- Security policies in an organization require the use of customer managed keys for all resources that support encryption at rest. 
+
+- To better manage and understand costs, an organization requires that all resources are tagged with a certain set of tags.
+
+- An organization's security team has decided that all IAM policies should not contain Allow "*" actions. A plugin can be written to scan the synthesized config and throw an error if a policy statement is found, and where it is found.
+
+- A development team is importing an existing stateful resource into their wing code, however the access to the resource should only ever be readonly. The devops team decides to write a plugin to ensure that no resource in the wing code is ever granted a policy that would allow write access to the imported stateful resource.
+
+## Examples
+These examples are meant to be illustrative of the plugin system, and not necessarily the best way to implement the use cases.
```
eladb, line 64: I think the next step of this RFC is to actually try to implement some of the use cases (or all of them). This is how we will validate and demonstrate that the solution can solve the problem.

(if you feel it will be useful, don't hesitate to start implementing or prototype this in the background).

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Wing Plugins
+description: How to implement wing plugins
+---
+
+# Wing Plugin System
```
eladb, line 6: I am not super comfortable with the name "plugin", but I am okay with it for now. Just wanted to share that :-)

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Background
+
+Developers using wing should whenever possible only care about functional aspects of an application's infrastructure. However, there must be a way to define non-functional requirements as that will be crucial for wide adoption. For instance, an organization may have specific security requirements to adhere to for compliance and regulation sake, such as encryption at rest, versioning, replication etc... 
+
+Applying these non-functional requirements will mean that the wing compiler should provide the ability for users to write plugins that can hook into the compilation process and apply business logic to resources.
+
+## Proposal
+
+Initially the plugin system can be crafted in simple manner, allowing the ability for users to pass in additional `--plugins` option to the wing cli. These plugin files will be Javascript that contains steps for applying non-functional infrastructure requirements.
```
eladb, line 22: I would tell this story through a concrete example (e.g. one of the use cases) and describe exactly how to apply it to a Wing application.

eladb (changes requested):

docs/05-rfcs/2023-01-21-wing-plugins.md
hasanaburayyan: I think I started less comfortable with it, but the more research I have done the more comfortable I have become with it. But yea I totally get it, If we want to get to a point where we have a really nice declarative syntax like you described here: https://github.com/winglang/wing/issues/493 then the name plugin does not work in my opinion. However, this writing raw JS files fits the plugin theme IMO but It may be opening too big of a can of worms (security wise).

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+          const actions = statement.Action as string[];
+          
+          if (actions.includes("*") && statement.Effect === "Allow") {
+            throw new Error(`Found a wildcard allow action in policy: ${key}`);
+          }
+        }
+      }
+    }
+  }
+}
```
ekeren, line 114: Can you also describe how you are going to "plug" the plugin?  Is this a compiler flag?

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+    * Allows the plugin access to the resulting terraform config. 
+    */
+    postSynthesis(config) {
+      // do something after synthesis
+    }
+  }
+```
+Additional iterations on the implementation can be made to allow for more complex plugin systems, such as allowing users to define their plugins in a special query language, or implement their plugins using higher level apis.
+
+## Use Cases
```
ekeren, line 45: I would also add the following use cases:

> For every bucket in app, turn on versioning for it and  create a duplicate bucket in a different region that replicate it

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+Developers using wing should whenever possible only care about functional aspects of an application's infrastructure. However, there must be a way to define non-functional requirements as that will be crucial for wide adoption. For instance, an organization may have specific security requirements to adhere to for compliance and regulation sake, such as encryption at rest, versioning, replication etc... 
+
+Applying these non-functional requirements will mean that the wing compiler should provide the ability for users to write plugins that can hook into the compilation process and apply business logic to resources.
+
+## Proposal
+
+Initially the plugin system can be crafted in simple manner, allowing the ability for users to pass in additional `--plugins` option to the wing cli. These plugin files will be Javascript that contains steps for applying non-functional infrastructure requirements.
+
+A wing plugin class consists of the following methods:
+```ts
```
eladb, line 25: I'd just go with a simple set of exported JavaScript functions instead of requiring class extension:

```js
// runs after the app is initialized, just before "synth" is called.
exports.preSynthesis = function(app) {
  // "app" is the construct tree
};

// runs after "synth" is called with the resulting terraform config
// returns an updated terraform config.
exports.postSynthesis = function(config) {
  return config; // <-- config can be mutated (optionally)
};
```

We also talked today about adding:

```js
// these hooks run after postSynthesis and 
exports.validate = function(config) {
  // does not return a value, config is immutable at this stage
}
```

We can consider (P2) first-class support for visitor pattern:

```js
exports.forEach = function(construct) {
  // here
};
```

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+- An application team is provided a VPC in their account that was provisioned by the platform engineering team. It is required that all the resources in the application are deployed into this VPC.
+
+- Security policies in an organization require the use of customer managed keys for all resources that support encryption at rest. 
+
+- To better manage and understand costs, an organization requires that all resources are tagged with a certain set of tags.
+
+- An organization's security team has decided that all IAM policies should not contain Allow "*" actions. A plugin can be written to scan the synthesized config and throw an error if a policy statement is found, and where it is found.
+
+- A development team is importing an existing stateful resource into their wing code, however the access to the resource should only ever be readonly. The devops team decides to write a plugin to ensure that no resource in the wing code is ever granted a policy that would allow write access to the imported stateful resource.
+
```
eladb, line 62: Add an example of how someone can publish a *reusable plugin* with some degrees of freedom. For example, an `AwsVpc` plugin which can be used by simply importing it from
npm and instantiating:

```js
const { AwsTerraformVpcPlugin } from "wing-vpc";

module.exports = new AwsTerraformVpc({ internetAccess: false });
```

docs/05-rfcs/2023-01-21-wing-plugins.md
eladb, line 22: Mention that people can pass multiple plugins and that they will be executed in order (for each hook).

docs/05-rfcs/2023-01-21-wing-plugins.md
hasanaburayyan: I need to look into this more, I have questions about how this would work using the node vm. For now I left the examples using classes but Ill get them converted once I figure this out.

eladb: OK, I'll stay tuned :-) I don't think there should be a problem...

eladb:
@hasanaburayyan I took the liberty and re-titled this "infrastructure customization through plugins". WDYT?

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,258 @@
...
+      const resources = config.resource;
+      if (resources.aws_iam_role_policy) {
+        const policyKeys = Object.keys(resources.aws_iam_role_policy);
+        for (const key of policyKeys) {
+          const policy = resources.aws_iam_role_policy[key].policy;
+          const statements = JSON.parse(policy).Statement;
+          for (const statement of statements) {
+            const actions = statement.Action;
+            if (actions.includes("*") && statement.Effect === "Allow") {
+              throw new Error(`Found a wildcard allow action in policy: ${key}`);
```
schosterbarak, line 127: Opinion: Policy as code and object mutation (what I thought the plugin was) are two different use cases.

hasanaburayyan:
> @hasanaburayyan I took the liberty and re-titled this "infrastructure customization through plugins". WDYT?

That you are a word smith 😃 sounds great to me

docs/05-rfcs/2023-01-21-wing-plugins.md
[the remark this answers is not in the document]
hasanaburayyan: That would be really cool, in the original issue there was discussion around using like a wing policy file with its own declarative syntax. Either way we go In my opinion it has to evolve from pure js, in my mind since these plugins will be probably mostly be written by platform teams I think a declarative dsl might be the most welcomed approach... but 🤷 just thinking outloud.

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,115 @@
...
+    * Allows the plugin access the root app node.
+    */
+    preSynthesis(app) {
+      // do something before synthesis
+    }
+    /**
+    * This method hooks in after the wing sdk code is synthesized.
+    * Allows the plugin access to the resulting terraform config. 
+    */
+    postSynthesis(config) {
```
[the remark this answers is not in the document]
hasanaburayyan: Yea I think these plugins (atleast in the first iteration) have to be pretty target specific and provisioning engine specific. I wonder what the best way to handle to illustrate that is.. maybe the plugin can contain some metadata on its compatible targets.

docs/05-rfcs/2023-01-21-wing-plugins.md
hasanaburayyan: hmm, I had not thought of them as being very different. I think that the former implies the latter. Where policies are implemented via mutations (in most cases)

Maybe room for discussions and clarifications, would be happy to hear more 😃

Jazzepi:
One thing I've run into with IAC is that you tend to have multiple people touching it with a varying collection of roles and responsibilities.

Some things I think should be addressed, though it really depends on what "stage" you are in for your plugin system.
1. Distribution - How are you going to package your plugins? A zip file? Any imposed folder structure? A single naked file?
2. End user configuration - I'm a consumer using a plugin, how can I pass it configuration to the plugin?
3. Environment information for the plugin - Is there anything you'd want to pass to the plugin so the author can change behavior? Environment variables? Host information like CPU/memory/Operating System? Version of Winglang you're running in?
4. Permission controls - If I'm working with infrastructure I'm giving some random person's code control over it's provisioning if there's no controls it's fairly trivial for them to do all sorts of malicious things like deploying crypto coin miners, putting in network level access backdoors, etc. etc. I think you need a way to scope the permissions that the plugin is granted so Winglang can bail out and not run if you try to run with a plugin but don't give it the permissions it's requesting.

hasanaburayyan:
> Some things I think should be addressed, though it really depends on what "stage" you are in for your plugin system.
> 
> 1. Distribution - How are you going to package your plugins? A zip file? Any imposed folder structure? A single naked file?
> 2. End user configuration - I'm a consumer using a plugin, how can I pass it configuration to the plugin?
> 3. Environment information for the plugin - Is there anything you'd want to pass to the plugin so the author can change behavior? Environment variables? Host information like CPU/memory/Operating System? Version of Winglang you're running in?
> 4. Permission controls - If I'm working with infrastructure I'm giving some random person's code control over it's provisioning if there's no controls it's fairly trivial for them to do all sorts of malicious things like deploying crypto coin miners, putting in network level access backdoors, etc. etc. I think you need a way to scope the permissions that the plugin is granted so Winglang can bail out and not run if you try to run with a plugin but don't give it the permissions it's requesting.

Yea I think these are totally valid concerns. Ive got some local changes to this RFC that Im planning to finish cleaning up and push but I think they will address the first 3 points so stay tuned! 

As for the permissions control this is a very interesting and challenging point, to an extent easy enough to limit the context in which the plugin runs though that does not solve the issue of plugins injecting backdoors, or deploying malicious resources (as is the issue with 3rd party tooling).

❤️ Great feedback, some things to ponder for sure!

docs/05-rfcs/2023-01-21-wing-plugins.md
```diff
@@ -0,0 +1,258 @@
...
+For instance if you have a plugin that applies a tag to all resources, and another plugin that creates additional resources, you need to ensure that the plugin that applies the tags is executed after the plugin that creates the resources (or else the tags will not be applied to the newly created resources).
+
+Another important thing to note is the phases of the compilation process that the plugin system hooks into. The plugin system hooks into the `preSynthesis`, `postSynthesis`, and `validation` phases of the compilation process. 
+
+- The `preSynthesis` phase is executed before the wing sdk code is synthesized which means your plugin will have access to the cdktf root app
+- The `postSynthesis` phase is executed after the wing sdk code is synthesized which means your plugin will have access to the resulting terraform config.
+- The `validation` phase is executed after the `postSynthesis` phase and is meant to be used to validate the resulting config, thus it is considered an immutable phase.
+
+```js
+  class MyWingPlugin {
```
eladb, line 33: Perhaps it makes most sense to describe this as a TypeScript interface:

```ts
interface IWingPlugin {
  preSynthesis(app: Construct): void;
  postSynthesis(config: any): any;
  validate(config: any): void;
}
```

eladb:
Some inspiration for what plugins can become: https://github.com/winglang/wing/issues/1365

docs/05-rfcs/2023-01-21-wing-plugins.md
hasanaburayyan: So I do have a lack of experience with the Node VM library but essentially when loading in the plugin
```typescript
const script = new vm.Script(some_path);
const result = script.runInContext(some_context);
```
the result from running the script is the last line evaluated in the script. 

Thus if I had a plugin file like
```js
exports.preSynthesis = function(app) {
	// do something
};

exports.postSynthesis = function(config) {
  return config;
};
``` 
Then I would later on be able call the postSynthesis because its export was read as the last line in the script
```typescript
vm.runInNewContext('postSynthesis(config);', {
	config,
	postSynthesis: result,
});
```

However trying the same for preSynthesis would fail since the export for that was not stored in the result
```typescript
// Wont work
vm.runInNewContext('preSynthesis(config);', {
	config,
	preSynthesis: result,
});
```
(this is all just a thought dump sorry)
For me this is why I originally went with wrapping these methods in a class and exporting the class.

As well for reference I have taken much inspiration from Rocket.chat
https://www.youtube.com/watch?v=yuEKbs934e0

hasanaburayyan: hmm, yea I think thats a fair point. This being a compile time system we dont have the same security concerns for running everything in VM with limited context. 

This is maybe as you said something we can punt, thanks for the input Ill give it a whirl.

eladb: The value I see in running in a VM is that it will force us to be more explicit about the surface area we expose to plugins and that's generally a Good Thing.

Looking at your `vm` code above, did you include an `exports` object in the VM context?

Something like:

```js
const hooks = {};
vm.runInNewContext(readFileSync(resolvedPluginModule), { exports: hooks });
hooks.preSynthesis()
```

hasanaburayyan: I had not, that does solve things 😅.

docs/05-rfcs/2023-01-21-wing-plugins.md
hasanaburayyan: So I took this and ran a bit maybe 😁 I went with a walkthrough approach. Not quite done but I think it fits more with the narrative.

eladb (approved):
Ship it

ekeren:
@ShaiBer 

Is it for the same use cases we have today just implemented across different providers?

ekeren:
Or is there some different use case for deployment agnostic plugins

ShaiBer:
> @ShaiBer
> 
> Is it for the same use cases we have today just implemented across different providers?

Yes.
I want to be able to move between clouds and retain (as much as possible) my non functional concerns.
With provier specific plugins you need to rewrite the plugins for each provider.
I feel like there should be a way to abstract this (or at least most of it, and then we'll have provider specific plugins for the provider agnostic ones :) )

hasanaburayyan:
@ShaiBer I think the way we get to that point is implementing some sort of DSL kinda like described in the original issue https://github.com/winglang/wing/issues/493. 

In the model for the beta the plugin is responsible to be agnostic (if desired). I think there is a case where the plugin may have a series of compatible targets which is what this card is for https://github.com/winglang/wing/issues/1474
in which I might write a plugin that is either compatible for (tf-aws | cf-aws | tf-azure)

Just some thoughts.

ShaiBer:
Cool, makes sense

Issue #1204, opened 2023-01-20
cloud.Function timeout property


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

See the `timeout` prop from the current draft SDK specification: https://docs.winglang.io/rfcs/2023-01-11-sdk-spec#function

### Use Cases

Allow cloud functions to run for different amounts of time

### Component

SDK

Chriscbr:
Note: Implementing timeout in the simulator could be non-trivial since inflights are not simulated in different threads, so I think it's OK if we don't support this initially. (For now we can say simulating `timeout` is P2 🙂).

staycoolcall911:
@RaphaelManke I saw you started working on this in #1220 , so I'm assigning this issue to you to prevent others from working on this simultaneously (true story 😄). Thanks for working on it!

RaphaelManke:
Thank you. I am not 100% sure to proceed in near future. So if someone is willing to continue feel free to work further. 

Initially I started to work on this issue to check out how the framework and sdk works. 

If I have some time, I will continue on that.

monadabot:
Congrats! :rocket: This was released in [Wing 0.4.141](https://github.com/winglang/wing/releases/tag/v0.4.141).

PR #1220, opened 2023-01-22
feat(sdk): extend function with timeout property 

RaphaelManke:
Closes: https://github.com/winglang/wing/issues/1204

This PR adds a timeout property to the `cloud.Function` class so that the execution time of a function can be limited. 
This property is defined in the rfc for the Function class. 
The RFC specifies a default value for timeout of 1 minute.

This PR adds a implementation for the timeout property for the simulator and aws function.

In the AWS function implementation the value of the timeout property is directly mapped to the cdktf property of the aws lambda function construct property.

The simulator implementation uses the node [vm options](https://nodejs.org/docs/latest-v16.x/api/vm.html#vmrunincontextcode-contextifiedobject-options) which offer a timeout property to cancel executions after a defined timeout. 

An alternative implementation of the timeout for the simulator has been not been evalutated because the implemented version resulted in the expected behaviour with a minimal effort. 
Side efects are not known yet. 
In case of a timeout in the simulator the User gets informed that a script execution has timed out.
This message could be improved to give the user feedback which inflight function has timed out.


Chriscbr (commented):
I think this looks good! Maybe for the simulator target we can throw an error if someone specifies a `timeout` value?

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -75,7 +77,7 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
     return this.context.withTrace({
       message: `Invoke (payload="${JSON.stringify(payload)}").`,
       activity: async () => {
-        return vm.runInContext(wrapper, context);
+        return vm.runInContext(wrapper, context, { timeout: this.timeout });
```
Chriscbr, line 80: Aha, so there's a timeout option! Brilliant :)

RaphaelManke: yes, I thought that would be clean solution, to make sure the script execution is really stopped.

RaphaelManke:
> Looks like there's a build error related to [this Wing example](https://github.com/winglang/wing/blob/main/examples/tests/invalid/struct_expansion.w) with this line in the logs:
> 
> ```
> wingc: thread 'sanity::cannot_compile_invalid_files' panicked at 'Not all fields of FunctionProps (at src/cloud/function.ts:21:1) are initialized.', libs/wingc/src/type_check.rs:902:21
> ```
> 
> What's weird is that `timeout` is an optional property, so I think the correct behavior is for the code to compile successfully. This is likely a bug in the compiler. You're welcome to dig into the compiler code if you'd like, but I think it's also totally okay to just update the example to specify the `timeout` for now, and someone can fix the error in a separate PR. 🙂

I am not sure how to fix the example. 
How do I test this build locally?
How do I use the Duration class from the std lib to set a timeout? 
what I'd like to do is 
```rust
new cloud.Function(
  handler, 
  cloud.FunctionProps {
    env: Map<str> {}
    timeout: std.Duration.fromMinutes(1)
  }
);

```

Chriscbr:
@RaphaelManke To represent a duration in Wing, you can write `1s`, `1m`, or `1h` (with any numbers)

To run all of the tests you can follow the instructions here: https://docs.winglang.io/contributors/handbook#-how-do-i-run-e2e-tests

Chriscbr (approved):

Chriscbr:
Thanks @RaphaelManke for contributing this! 🙏

staycoolcall911:
Congrats on your first PR @RaphaelManke 👏

PR #1226, opened 2023-01-23
fix(sdk): azure share resources at app level

hasanaburayyan:
This PR moves storage account and resource group resources to the app level to be shared across all resources defined within the app.

Also minor refactor on inflight client to only create 1 instance of containerClient.

### DIFF
--- a/libs/wingsdk/src/target-tf-azure/app.ts
+++ b/libs/wingsdk/src/target-tf-azure/app.ts
@@ -13,6 +13,27 @@ export interface AzureAppProps extends AppProps {
 }
 
 /**
+ * ResourceGroup names are limited to 90 characters.
+ * You can use alphanumeric characters, hyphens, and underscores,
+ * parentheses and periods.
+ */
+const RESOURCEGROUP_NAME_OPTS: NameOptions = {
+  maxLen: 90,
+  disallowedRegex: /([^a-zA-Z0-9\-\_\(\)\.]+)/g,
+};
+
+/**
+ * StorageAccount names are limited to 24 characters.
+ * You can only use alphanumeric characters.
+ */
+const STORAGEACCOUNT_NAME_OPTS: NameOptions = {
+  maxLen: 24,
+  case: CaseConventions.LOWERCASE,
+  disallowedRegex: /([^a-z0-9]+)/g,
+  sep: "",
+};
+
+/**
  * An app that knows how to synthesize constructs into a Terraform configuration
  * for Azure resources.
  */
@@ -38,6 +59,8 @@ export class App extends CdktfApp implements IApp {
    * @link https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/resource_group#location
    * */
   public readonly location: string;
+  public readonly resourceGroup: ResourceGroup;
+  public readonly storageAccount: StorageAccount;
 
   constructor(props: AzureAppProps) {
     super({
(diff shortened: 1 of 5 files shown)


libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -61,5 +86,19 @@ export class App extends CdktfApp implements IApp {
...
       writable: false,
     });
+
+    // Create a resource group and storage account for all resources in this app
+    this.resourceGroup = new ResourceGroup(this, "ResourceGroup", {
+      location: this.location,
+      name: ResourceNames.generateName(this, RESOURCEGROUP_NAME_OPTS),
+    });
+
+    this.storageAccount = new StorageAccount(this, "StorageAccount", {
```
hasanaburayyan, line 96: I think it's possible that storage account can be implemented as a singleton, but looking into the azure docs on storage account I think it's nearly always going to be the case we want to create 1 per app. Still pondering this...

eladb: We could implement this as a lazy property.

hasanaburayyan: Yup, agreed was just chatting with @staycoolcall911 about using lazy loading here.

libs/wingsdk/test/target-tf-azure/bucket.test.ts
```diff
@@ -48,13 +48,15 @@ test("bucket name valid", () => {
   const app = new tfazure.App({ outdir: mkdtemp(), location: "East US" });
   const bucket = new cloud.Bucket(app, "The-Uncanny-Bucket");
   const output = app.synth();
+  console.log(output);
+  console.log(bucket.node.addr.substring(0, 8));
```
Chriscbr, line 52: Clean up

Chriscbr (commented):
Let's unit test this with an app that has more than one bucket

libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -61,5 +86,49 @@ export class App extends CdktfApp implements IApp {
       enumerable: false,
       writable: false,
     });
+
+    // At runtime the resolved app class is an instance of InnerApp which means that all the methods
+    // declared in this class are tossed out. This is a workaround to attach the getters to InnerApp class
+    // at runtime.
+    this._resourceGroup = undefined as any;
+    Object.defineProperty(this, "resourceGroup", {
```
hasanaburayyan, line 94: @eladb @staycoolcall911 @Chriscbr Wanted to draw your attention to this. Im still trying to understand why, but essentially because the instance of app is that of InnerApp from the CdktfApp during runtime creating a getter (or any other public method) gets tossed out. @MarkMcCulloh might have better explanation, but he rescued me on this one 😅. Though during our debugging we thought it might merit some more team discussion around this.

libs/wingsdk/test/target-tf-azure/bucket.test.ts
hasanaburayyan: 👍 cleaned up

libs/wingsdk/src/target-tf-azure/app.ts
eladb: Not sure I understand...

Is this what you tried to do and didn't work?

```ts
class Foo {
  private _resourceGroup?: ResourceGroup;

  public get resourceGroup() {
    if (!this._resourceGroup) {
      this._resourceGroup = new ResourceGroup(this, "ResourceGroup", ...);
    }

    return this._resourceGroup;
  }
}
```

hasanaburayyan: Thats right that getter method would never be called, because the only methods/fields available are those of the InnerApp class. Though what I dont understand is why when we create an instance of the app directly 
```ts
const app = new tfazure.App({ outdir: mkdtemp(), location: "East US" });
console.log(app);
```
this logged app is an InnerApp

Chriscbr: @hasanaburayyan @eladb Odd... actually I think I might have also encountered some issues like this while trying to set up stuff on the GCP bucket resource. I think it's possible this might be fixable by refactoring away the `CdktfApp` - take a look at what I did in my PR here (in core/app.ts):

https://github.com/winglang/wing/pull/827/files#diff-380d8cb14bea95361d57e8a63cc67a761a28a2e36214a025446ffef5f4c5adfe

I haven't cleaned up everything yet but I think with some tweaking it might be able to fix the issue.

hasanaburayyan: Yea I think refactoring out the InnerApp class is going to be the fix, once that PR is merged Ill get this rebased and fixed to be cleaner!

hasanaburayyan: Yupp, rebased the new changes to Cdktf class and things work the way they should 😁

eladb (approved):
Nice and elegant.

Curious what was the issue with the properties...

PR #1231, opened 2023-01-23
feat(sdk): bucket.add_object

Chriscbr:
Code written from the third Wingly Update twitch show. Still need to write unit tests and probably at least one integration test before shipping it.

I promise to get back to this PR just like GCP bucket (#827) 😆 


libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -103,6 +104,14 @@ export class Bucket extends cloud.BucketBase {
     }
   }
 
+  public addObject(key: string, body: string): void {
+    new S3Object(this, `S3Object-${key}`, {
+      bucket: this.bucket.bucket,
+      key,
```
marciocadev, line 110: @Chriscbr Restrictions on object names in S3 are quite broad, but there are characters that need special treatment and others to avoid as per this link

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

Perhaps we can ignore these limitations for the moment (I imagine very few people want to create a file with @ or $) or we can change the resource-names so that it is not limited to resources but to everything related to names

And I imagine that this can also affect the logical id of the object

libs/wingsdk/src/target-tf-azure/bucket.ts
```diff
@@ -92,8 +93,23 @@ export class Bucket extends cloud.BucketBase {
...
+  public addObject(key: string, body: string): void {
+    // Blob naming conventions:
+    // https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#blob-names
+    const blobName = key
+      .split("/")
+      .map((s) => encodeURIComponent(s))
+      .join("/");
 
-    this.storageContainer;
+    new StorageBlob(this, `Blob-${key}`, {
```
marciocadev, line 106: I don't know much about CdkTF, but I imagine there are limits on the length of the logical id string

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -35,6 +35,11 @@ export abstract class BucketBase extends Resource {
 
     props;
   }
+
+  /**
+   * Add a file to the bucket that is uploaded when the app is deployed.
+   */
+  public abstract addObject(key: string, body: string): void;
```
hasanaburayyan, line 42: I did not see one but I wonder if we have any issue that references future support of blobs. Maybe a TODO: with the issue link might mitigate questions from those who wonder why we are using just string for body. 🤷

hasanaburayyan (approved):
❤️ the azure support as well. Small question on whether to add a TODO comment but other than that LGTM.

libs/wingsdk/src/cloud/bucket.ts
marciocadev: I had the same question, but I thought that since the language didn't have blob support yet that's why str was used

libs/wingsdk/src/target-tf-aws/bucket.ts
Chriscbr: Logical IDs don't have any hard restrictions on characters or length AFAIK.

Based on the docs you linked, maybe we could offer some warnings if users choose file names with special characters, but I think it's OK to keep things simple for now. For the size of the string, I think every cloud provider has some restrictions on user input sizes, but since the key name is entirely user-provided, I think it's OK to let the cloud provider throw an error instead of handling it in the Wing SDK.

libs/wingsdk/src/target-tf-azure/bucket.ts
Chriscbr: No restrictions that I'm aware of

libs/wingsdk/src/cloud/bucket.ts
Chriscbr: I've added a comment to the description 👍

marciocadev (approved):

Chriscbr:
GitHub says I still need an approval from @winglang/maintainers

staycoolcall911 (approved):

PR #1247, opened 2023-01-24
fix(sdk): unable to capture resources inside immutable data containers

### LINKED ISSUE #1156: Unable to capture an array/map/set of resources
### Community Note

### I tried this:

I tried to capture an array with two buckets:

```js
bring cloud;

let arr = [new cloud.Bucket(), new cloud.Bucket()];

  let b1 = arr.at(0);
  let b2 = arr.at(1);
  b1.put("hello", "world");
});
```

### I expected this:

To work

### Instead, this happened:

The capturing mechanism currently treats data and resources differently, which means that in this case the buckets are not bound as resources.

### Component:

Compiler, SDK

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

See https://github.com/winglang/wing/pull/1144#discussion_r1069527720

eladb:
This is an initial step towards merging `data` and `resources` in the binding system. We now allow any immutable data object to also include resources (e.g. an array of buckets, map of queues, etc).

To support this, we use duck-typing to determine if an object is a resource (e.g. it implements `_toInflight` and `_bind`), and call these methods in their relevant place respectively.

Fixes #1156


libs/wingsdk/src/core/internal.ts
```diff
@@ -103,8 +105,14 @@ function serializeImmutableData(obj: any): string {
         return `new Map(${serializeImmutableData(Array.from(obj))})`;
       }
 
-      // structs are just objects
-      if (obj instanceof Object) {
+      // if the object is a resource (i.e. has a "_toInflight" method"), then we lift and add to the
+      // list of bindings.
+      if (typeof (obj as IResource)._toInflight === "function") {
+        return (obj as IResource)._toInflight().text;
+      }
```
Chriscbr, line 112: This comment does not match this code.

BTW - I feel like it's okay to treat constructs without `_toInflight` or `_bind` as still resources - they're just resources that don't have any inflight functionality. For example we could provide a default `_toInflight()` implementation that returns an empty inflight client class, and a bind method that returns nothing.

libs/wingsdk/src/core/resource.ts
```diff
@@ -353,3 +341,69 @@ export class Display {
...
+        (obj as IResource)._bind(host, ops);
+
+        // add connection metadata (use "?" to indicate that we don't know the operation)
+        for (const op of ops) {
+          Resource.addConnection({
+            from: host,
+            to: obj,
+            relationship: op,
+          });
+        }
```
Chriscbr, line 394: Update comment

libs/wingsdk/src/core/resource.ts
```diff
@@ -353,3 +341,69 @@ export class Display {
...
+
+        return;
+      }
+
+      // structs are just plain objects
+      if (obj.constructor.name === "Object") {
+        Object.values(obj).forEach((item) => bindObject(item, host, ops));
```
Chriscbr, line 401: I feel like there's something we're not quite modeling here -- the list of operations passed to this function are specifically operations about `obj`. (Or at least, that was the original intent). If a struct contains a bucket and a key value store, we shouldn't mix up performing `put` on a `cloud.Bucket` with performing `put` on a `cloud.KeyValueStore`. Do you see what I mean?

libs/wingsdk/src/core/internal.ts
eladb: > BTW - I feel like it's okay to treat constructs without _toInflight or _bind as still resources - they're just resources that don't have any inflight functionality. For example we could provide a default _toInflight() implementation that returns an empty inflight client class, and a bind method that does nothing

That's an interesting idea. There is even a use case for referencing things like resource attributes from inflight.

```js
bring "aws-cdk" as aws;

let bucket = new aws.s3.Bucket();

  print(bucket.bucket_name);
};
```

I'll open an issue and follow up.

eladb: https://github.com/winglang/wing/issues/1251

libs/wingsdk/src/core/resource.ts
eladb: DONE

libs/wingsdk/src/core/resource.ts
eladb: Yes, good point. In the `struct` case, `ops` is just going to be `["?"]` anyway, so I think we are "fine".

As discussed yesterday, we need to revisit how "ops" are determined/propagated anyway.

Chriscbr (approved):

Chriscbr:
@eladb @MarkMcCulloh I'm not sure why this got automatically merged, it looks like on the last commit before mergify commented, the E2E tests were failing. Is this an issue with our configuration somewhere?

MarkMcCulloh:
@Chriscbr I think I see the issue. The self mutation workflow also has a job named "build" so when that passes it considers it good to merge. In this PR I changed the name https://github.com/winglang/wing/pull/1250

Chriscbr:
Ahh, nice catch 👍

PR #1248, opened 2023-01-24
feat(examples): naive distributed lock

Chriscbr:
Another Wing example for inspiration!

See the source code for more explanation on limitations of this lock implementation. It's possible kind of lock might be practical if it was augmented with a timeout mechanism -- i.e. it internally tracked _who_ requested the lock and what time it was requested, and if the lock was not released after that amount of time, whoever requested the lock would be evicted/stopped and the lock would be released.

Another idea is to tweak this resource into a [semaphore](https://en.wikipedia.org/wiki/Semaphore_(programming)). A semaphore is basically a lock that can be acquired by up to N workers.


examples/proposed/lock.w
```diff
@@ -0,0 +1,83 @@
+bring cloud;
+
+/**
+ * A lock that can only be acquired by one inflight at a time.
+ *
+ * Internally, this uses a cloud.Counter to keep track of a number.
+ * If the number is 0, the lock is free. If the number is 1, the lock
+ * is acquired.
+ *
+ * This is a very simple lock, and it has a number of limitations:
```
eladb, line 10: Worth talking about how can we solve these problems with wing.

eladb: Maybe a team time brainstorm

eladb (approved):

flyingImer:
you beat me on this example, but let me share mine as well once it is ready

Chriscbr:
> you beat me on this example, but let me share mine as well once it is ready

yes! I'd love to see it (we can definitely add multiple examples) :)

examples/proposed/lock.w
```diff
@@ -0,0 +1,83 @@
...
+  * - It is not reentrant. If a single inflight tries to acquire the lock
+  *   twice, it may block forever.
+  * - It is not fair. If multiple inflights are waiting to acquire the lock,
+  *   there is no guarantee that they will acquire the lock in the order
+  *   that they requested it.
```
MarkMcCulloh, line 15: tbh I would not expect a lock to be fair. 
If I needed a fair lock I would be okay with requiring a `FairLock` resource or something since there's gonna have to be overhead.

examples/proposed/lock.w
```diff
@@ -0,0 +1,83 @@
...
+  * - It is not stable. If an inflight crashes while holding the lock, the
+  *   lock will be permanently stuck in the acquired state.
+ */
+resource Lock {
+  counter: cloud.Counter;
```
MarkMcCulloh, line 20: I feel like Counter is not a sufficient structure here.

Using something that supports:
- key/value storage
- TTLs

Would allow the lock to be both reentrant and "stable" in crash conditions.

Obviously we don't have that now, but the closest thing could be some kind of `Cache` I guess

eladb: DynamoDB has support for TTLs so that might be something to consider.

examples/proposed/lock.w
Chriscbr: Fair enough (ha). I've seen in Java and some other languages, locks sometimes have a way to configure fairness so e.g. under contention, the lock is always given to the longest-awaited thread. Maybe it's not something you always expect as a developer, but this kind of feature can help avoid some forms of deadlock I think.

MarkMcCulloh: To be clear, I think being able to configure fairness is great and I agree with what you're saying, I just meant as a default for something simply called `Lock` I think it's okay if it's unfair.

examples/proposed/lock.w
MarkMcCulloh: Yeah I think under-the-hood dynamo would be okay for this, although a little slow compared to Elasticache if you're okay with a volatile lock.

flyingImer:
> 

> > you beat me on this example, but let me share mine as well once it is ready
> 
> yes! I'd love to see it (we can definitely add multiple examples) :)

it's here: https://github.com/winglang/wing/pull/1260 😃

PR #1256, opened 2023-01-25
chore: missing snapshot and fix premature merging

MarkMcCulloh:
#1247 was merged prematurely, missing this snapshot update.

Mergify sees the "build" success from the SDK mutation job and attempts to merge, but that was not intentional. So we rename it here.

### DIFF
--- a/.github/workflows/wingsdk-mutation.yml
+++ b/.github/workflows/wingsdk-mutation.yml
@@ -20,7 +20,7 @@ env:
   CI: "true"
 
 jobs:
-  build:
+  mutate:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


Chriscbr (approved):

.github/workflows/wingsdk-mutation.yml
```diff
@@ -20,7 +20,7 @@ env:
   CI: "true"
 
 jobs:
-  build:
+  mutate:
```
staycoolcall911, line 23: @MarkMcCulloh - maybe add a comment here in a subsequent PR, just to explain why we called it `mutate` instead of `build` and to prevent someone from refactoring it to `build` again to be consistent with names across our ymls...

PR #1258, opened 2023-01-25
fix(compiler): std objects can't be used inflight

### LINKED ISSUE #1242: `str` should support .includes() method on inflight
### Community Note

### I tried this:

```ts
bring cloud;

let a = "some string";

 assert(a.includes("some"));
})
```

### I expected this:

To work.

### Instead, this happened:

Compilation failed with 1 errors
Error at code.w:6:9 | Cannot call preflight function "a.includes" while in inflight phase

### Component:

SDK

### Wing Version:

0.4.131

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

MarkMcCulloh:
By default, class imported through jsii have been preflight only. This PR changes any class that is not a resource (construct) to be phase independent. This Fixes #1242 by allowing the use of the primitive APIs in any phase as-is.


libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -413,6 +413,11 @@ impl<'a> JsiiImporter<'a> {
 	fn import_class(&mut self, jsii_class: wingii::jsii::ClassType, namespace_name: &str) {
 		let mut is_resource = false;
 		let type_name = &self.fqn_to_type_name(&jsii_class.fqn);
+		let phase = if jsii_class.assembly == WINGSDK_ASSEMBLY_NAME && namespace_name == WINGSDK_STD_MODULE {
```
eladb, line 416: This feels pretty special-casy... Is it possible to add some JSII annotation to these APIs that will tell Wing that they are phase independent?

I also believe we decided that classes are phase independent, no? (i.e. they can be instantiated and used from both phases). WDYT?

eladb (changes requested):

libs/wingc/src/type_check/jsii_importer.rs
MarkMcCulloh: I mention the jsii annotation in the description, and I suppose I'm fine with it but it doesn't feel particularly useful right now. The only things that stay the same between preflight/inflight are serializable primitives, which are all in `std` right now and we have no means of having anywhere else.

Not sure what it means for a class to be phase independent, we'd need to have a way of cloning the class definition across boundaries and accounting for mutability somehow. Not really something this PR is addressing.

eladb: Being phase independent is not the same as allowing capture. It just means that it can be instantiated either in preflight or in inflight.

Specifically, primitives and immutable collections can also be captured across inflight boundaries. But that's covered already.

Does that make sense?

Bottom line -- I think we should just consider classes to phase independent, and therefore all std type APIs are also phase independent.

yoav-steinberg: Another option is to consider anything that's a construct/resource to be preflight while everything else is phase independent. I think that's a better approach for now and it should resolve the issue you're having with using std stuff inflight.

yoav-steinberg (approved):

eladb:
Description probably needs an update

PR #1259, opened 2023-01-25
fix(docs): dangled `let` modifier for a `var` variable

flyingImer:

### DIFF
--- a/docs/04-reference/winglang-spec.md
+++ b/docs/04-reference/winglang-spec.md
@@ -406,7 +406,7 @@ Variables can be reassigned to by adding the `var` modifier:
 
 ```ts
 // wing
-let var sum = 0;
+var sum = 0;
 for item in [1,2,3] {
   sum = sum + item;
 }


docs/04-reference/winglang-spec.md
```diff
@@ -406,7 +406,7 @@ Variables can be reassigned to by adding the `var` modifier:
 
 ```ts
 // wing
-let var sum = 0;
+var sum = 0;
```
eladb, line 409: nope, our syntax is `let var` :-)

eladb:
@flyingImer our syntax is `let var` and not just `var`. I am actually just writing a post explaining the reasoning behind it. Stay tuned.

I think this PR can be closed.

docs/04-reference/winglang-spec.md
flyingImer: oh was not aware of it, tried to look for reference in the doc, but no luck. probably I missed it somewhere, any pointer?

flyingImer: `Let let be let. (Elad B. 2022)` I figured

PR #1260, opened 2023-01-25
feat(examples): naive counting semaphore

flyingImer:
A naive implementation, not quite the right encapsulation I wanted.

For example, I feel as a distributed semaphore in a cloud environment should not be trapped in a while loop if fails to acquire a share (intentionally leave out the `acquire` implementation for now). Instead, it should be dispatched to something similar in a cloud resource for later pickup, like a queue, keep the async nature of being on cloud. However, I would like to get some early feedback to see if I'm getting the right vibe in my first winglang example.


examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+
+  // probably key-value store is wanted,
+  // so that a specific resource instance can be released
+  // rather than naively releasing a count
+  public inflight release() {
+    if this._counter.peek() <= 0 {
+      return;
+    }
+
+    this._counter.dec();
```
flyingImer, line 43: seems this is missing from the counter interface

examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+  init(props: CountingSemaphoreProps) {
+    // pseudocode: input validation
+    this.limit = props.availableResources;
+    this._counter = new cloud.Counter();
+  }
+
+  // some stable unique instance id is wanted in the inflight context
+  // so that a reousce instance can be properly claimed and later released
+  // when used in conjunction with a key-value store
+  public inflight try_acquire(): bool {
```
flyingImer, line 22: compiler seems to be not happy with `public inflight`

wondering if `public` modifier is necessary in this case? or recommended in general? as I saw some inconsistency across examples in the repo, please advice

examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+  _counter: cloud.Counter;
+
+  // need some ttl solution here,
+  // probably in-house once replaced with a key-value store
+  init(props: CountingSemaphoreProps) {
+    // pseudocode: input validation
+    this.limit = props.availableResources;
+    this._counter = new cloud.Counter();
+  }
+
```
flyingImer, line 18: for some reason I wanted to do something below from the spec, but none of them works with the current complier.

let me know if they are expected or not - unlike the `struct` one, it has a nice message saying WIP.

```
public inflight var my_variable: bool; // <- `inflight` and `var` do not seem to play together as of now
inflight init() {} // <- got complaints about have `init()` defined already, which, in the context, is a preflight - like the kitchen sink example
```

examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+let resource_1 = new CountingSemaphore({ availableResources: 2 });
+let resource_2 = new CountingSemaphore({ availableResources: 3 });
+
+let queue = new cloud.Queue();
+/**
+ * An example handler to work with two shared resources.
+ */
+queue.on_message(inflight (message: str) => {
+  let is_resource_1_acquired = resource_1.try_acquire();
+  if !is_resource_1_acquired {
```
flyingImer, line 60: posted a question over slack on this NOT operator on bool type: https://winglang.slack.com/archives/C048QCN2XLJ/p1674689428632649

examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+  // need some ttl solution here,
+  // probably in-house once replaced with a key-value store
+  init(props: CountingSemaphoreProps) {
+    // pseudocode: input validation
+    this.limit = props.availableResources;
+    this._counter = new cloud.Counter();
+  }
+
+  // some stable unique instance id is wanted in the inflight context
+  // so that a reousce instance can be properly claimed and later released
```
Chriscbr, line 20:
```suggestion
  // so that a resource instance can be properly claimed and later released
```

examples/proposed/counting-semaphore.w
Chriscbr: I think these are both compiler bugs

examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+    this._counter.dec();
+  }
+
+  public inflight is_at_capacity(): bool {
+    return this._counter.peek() >= this.limit;
+  }
+}
+
+let resource_1 = new CountingSemaphore({ availableResources: 2 });
+let resource_2 = new CountingSemaphore({ availableResources: 3 });
```
Chriscbr, line 52: Nice, thanks for showing the usage of the resource

Chriscbr (approved):
great example!

examples/proposed/counting-semaphore.w
```diff
@@ -0,0 +1,75 @@
...
+ */
+queue.on_message(inflight (message: str) => {
+  let is_resource_1_acquired = resource_1.try_acquire();
+  if !is_resource_1_acquired {
+    // brutally error out to re-enqueue
+    throw("Failed to acquire resource 1");
+  }
+  let is_resource_2_acquired = resource_2.try_acquire();
+  if !is_resource_2_acquired {
+    // brutally error out to re-enqueue
```
flyingImer, line 66: should release resource_1 here, will update it

examples/proposed/counting-semaphore.w
flyingImer: bug reported: https://github.com/winglang/wing/issues/1271

examples/proposed/counting-semaphore.w
flyingImer: bug reported: https://github.com/winglang/wing/issues/1271

PR #1264, opened 2023-01-26
fix(compiler): unable to use unary `!` with boolean

MarkMcCulloh:


ekeren:
@MarkMcCulloh , maybe it is worth adding an `invalid` test that make sure we don't use binaryUnaryOps on Numbers

```ts (wing)
let n = 5;
if !n {
}
```

examples/tests/valid/statements_if.w
```diff
@@ -2,13 +2,16 @@ bring cloud;
...
+  let f = false;
   if true && x + 2 == 4 {
     if true && x + 3 == 4 {
       assert(false);
     } elif true && x + 3 == 6 {
       assert(false);
     } elif false || x + 3 == 5 {
       assert(true);
+    } elif !f {
+      assert(!!!f);
```
Chriscbr, line 14: Can we add a test that `+-+f` is invalid? And a positive test for numeric unary expressions (like `+x`, `-x`) if we don't have those already?

MarkMcCulloh: Actually, why is `+` even a unary op? I can add additional cases, but we might as well remove unary plus.

Chriscbr: I suppose you could use it to emphasize that a value is positive, like
```
arr.push(+1);
arr.push(-1);
```

MarkMcCulloh: I mean I know js has it, but it doesn't seem useful for us to inherit it. If we do, then `+-+f` is actually valid

yoav-steinberg: I have nothing against removing unary plus, not strongly opinionated on this.

yoav-steinberg (approved):

examples/tests/valid/statements_if.w
Chriscbr: I'm also ok with removing. I think `--f` should also be an error (since we don't a dedicated `++` or `--` operator in the language)

MarkMcCulloh: Alright, made changes:
- tree-sitter now parses unary `--` to detect it, compilation intentionally fails
  - this is needed, because otherwise --x is basically the same as -(-x) which is valid but probably is not what the user was expecting to happen
- remove unary plus (in code and in spec)

examples/tests/invalid/statements_if.w
```diff
@@ -0,0 +1,5 @@
+let n = 0;
+--n;
```
eladb, line 2: Maybe add the compiler diagnostics as comments like in other tests (eventually we will check them)

eladb (approved):
👍

PR #1267, opened 2023-01-26
chore: pin rust version and ensure `node_options` propagates

MarkMcCulloh:
rust 1.67.0 broke our wasm builds. https://github.com/rust-lang/rust/releases/tag/1.67.0

Still not entirely sure how, but this is currently breaking all builds. Will look into fixing the root cause after this is merged to unblock others.

While looking into issues, I realized that our entrypoint for wing did not preserve any existing NODE_OPTIONS, which seemed like a bad idea, so I kept the change in here.


MarkMcCulloh:
The build here will not succeed because it's not using the new version, you can see success in my fork here https://github.com/MarkMcCulloh/wing/actions/runs/4019332667

staycoolcall911 (approved):

PR #1269, opened 2023-01-26
fix(compiler): should show errors for missing semicolons or braces

### LINKED ISSUE #1266: Compiler should show an error for missing semicolon or parenthesis
### Community Note

### I tried this:

Tried compiling these:
```ts
let x = 1
```
Note the missing semicolon.

```ts
if (true) {
```
Note the missing closing parenthesis.

### I expected this:

To see an error message - e.g. missing ;, missing } 

### Instead, this happened:

These 2 code samples compiled successfully.

### Component:

Compiler

### Wing Version:

0.4.143

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

Implementation note:
This should be very easy to fix.

It turns out tree-sitter actually has that information when something is missing. 
The code in [here](https://github.com/winglang/wing/blob/b50e45058ab015dce41fc304e823f2e80ac3b5b3/libs/wingc/src/parser.rs#L848) traverses the CST and  throws a compiler error for any tree-sitter parsing error. Seems like we should simply tweak this to also handle node.is_missing().

Chriscbr:
Fixes #1266

After making the fix work, I discovered there were many syntax errors in our valid examples, so I fixed all of them as well.

I also noticed this sequence of commands were failing on a fresh repo:

```
npm i
cd libs/wingc
npx nx run build-native
npx nx run test
...
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "Expected to find .jsii file in ../wingsdk, but no such file found"', libs/wingc/src/type_check.rs:1729:18
```

The same issue applies with `npx nx run dev`.

To fix this I've added the sdk as a dependency of `wingc:dev` and `wingc:test`.

### DIFF
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -154,13 +154,6 @@ impl Parser<'_> {
 	fn build_scope<'a>(&self, scope_node: &Node) -> Scope {
 		let mut cursor = scope_node.walk();
 
-		// check if the last child is a missing closing brace
-		if let Some(last_child) = scope_node.children(&mut cursor).last() {
-			if last_child.kind() == "}" && last_child.is_missing() {
-				_ = self.add_error::<Node>(format!("'}}' expected."), &last_child);
-			}
-		}
-
 		Scope {
 			statements: scope_node
 				.named_children(&mut cursor)
@@ -254,14 +247,6 @@ impl Parser<'_> {
 			"ERROR" => return self.add_error(format!("Expected statement"), statement_node),
 			other => return self.report_unimplemented_grammar(other, "statement", statement_node),
 		};
-
-		// check if the last child is a missing semicolon
-		let mut cursor = statement_node.walk();
-		if let Some(last_child) = statement_node.children(&mut cursor).last() {
-			if last_child.kind() == ";" && last_child.is_missing() {
-				self.add_error::<Node>(format!("';' expected."), &last_child)?;
-			}
-		}
 
 		Ok(Stmt {
 			kind: stmt_kind,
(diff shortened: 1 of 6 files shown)


libs/wingc/src/parser.rs
```diff
@@ -154,6 +154,13 @@ impl Parser<'_> {
 	fn build_scope<'a>(&self, scope_node: &Node) -> Scope {
 		let mut cursor = scope_node.walk();
 
+		// check if the last child is a missing closing brace
+		if let Some(last_child) = scope_node.children(&mut cursor).last() {
+			if last_child.kind() == "}" && last_child.is_missing() {
+				_ = self.add_error::<Node>(format!("'}}' expected."), &last_child);
```
Chriscbr, line 160: Aside - does anyone know why some methods in the parser module return an `AST type` (like this method) while other methods return a `DiagnosticResult<AST type>`? is one of them more correct?

MarkMcCulloh: Generally `DiagnosticResult` is more correct. If using the unwrapped one, it's because either it can't fail or we don't handle the failure. The latter is more likely 🥲 

In the case of this method, building a scope can't really _fail_. It can be empty, but if you already invoked this method you're going to get some kind of scope back.

MarkMcCulloh (approved):

staycoolcall911 (approved):

libs/wingc/src/parser.rs
```diff
@@ -857,7 +872,7 @@ impl Parser<'_> {
 	fn report_unhandled_errors(&self, root: &Node) {
 		let iter = traverse(root.walk(), Order::Pre);
 		for node in iter {
-			if node.is_error() && !self.error_nodes.borrow().contains(&node.id()) {
+			if (node.is_error() || node.is_missing()) && !self.error_nodes.borrow().contains(&node.id()) {
```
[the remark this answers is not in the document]
Chriscbr: Oh, duh - I didn't think of that. 🤦‍♂️  Yeah, the `missing` nodes actually include information about what kind of thing is missing, so no need for all the special case handling. Fixed

eladb (approved):
Nice

PR #1274, opened 2023-01-27
feat(sdk): adding dec method to counter

marciocadev:
I created the `dec` method but I think it looks strange because without limits the counter can have negative values

Maybe it's worth giving the option of creating lower and higher limits for `Counter`, generating an exception if they are exceeded

If this makes sense let me know and I'll implement it


ShaiBer (approved):
Very cool!

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
```diff
@@ -40,6 +40,28 @@ export class CounterClient implements ICounterClient {
     return parseInt(newValue) - amount;
   }
 
+  public async dec(amount = 1): Promise<number> {
+    const command = new UpdateItemCommand({
```
eladb, line 44: Is it possible to simply call `inc()` here with a negative value?

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
```diff
@@ -40,6 +40,28 @@ export class CounterClient implements ICounterClient {
...
+      TableName: this.tableName,
+      Key: { [HASH_KEY]: { S: COUNTER_ID } },
+      UpdateExpression: `SET ${VALUE_ATTRIBUTE} = if_not_exists(${VALUE_ATTRIBUTE}, :${INITIAL_VALUE_TOKEN}) - :${AMOUNT_TOKEN}`,
+      ExpressionAttributeValues: {
+        [`:${AMOUNT_TOKEN}`]: { N: `${amount}` },
+        [`:${INITIAL_VALUE_TOKEN}`]: { N: `${this.initial}` },
+      },
+      ReturnValues: "UPDATED_NEW",
```
eladb, line 52: Why not return the old value here?

eladb (changes requested):

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
marciocadev: Yes, I tested it and it works

I was following the [Counter Spec](https://docs.winglang.io/reference/wingsdk-spec#counter) but `inc` can be used to decrement the value

Perhaps stating that `inc` accepts both positive and negative values ​​is a better alternative

eladb: What I meant is that we can simply implement `dec()` using `inc()`. I still think we need both methods.

```ts
async dec(amount = 1) {
  return this.inc(-1 * amount);
}
```

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
marciocadev: If it's the first time returning old values ​​it doesn't work because there's nothing in the table

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
marciocadev: I will try this but have to make some test

please do not merge

libs/wingsdk/src/cloud/counter.ts
```diff
@@ -73,6 +73,14 @@ export interface ICounterClient {
    */
   inc(amount?: number): Promise<number>;
 
+  /**
+   * Decrement the counter, returning the previous value.
+   * @param amount amount to decrement (default is 1).
+   * @returns the previous value of the counter.
```
eladb, line 79: Missing `@default`

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
```diff
@@ -40,6 +40,10 @@ export class CounterClient implements ICounterClient {
     return parseInt(newValue) - amount;
   }
 
+  public async dec(amount = 1): Promise<number> {
```
eladb, line 43: Perhaps we can put this in `CounterBase`? Seems like there's no need to implement this for every target?

eladb (approved):
Looks great. See minor comment about providing a default implementation in the base class

libs/wingsdk/src/cloud/counter.ts
marciocadev: Good catch!

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
marciocadev: Great idea, I'll change and do some tests

Issue #1132, opened 2023-01-11
Fragile publish workflow


MarkMcCulloh:
See error here https://github.com/winglang/wing/actions/runs/3896316568/jobs/6652941088#step:6:28

Occasionally, we fail to publish the vscode extension with this intermittent. At this point in the workflow, we already published NPM packages but have not yet created a release/tag. This unfortunately means that we cannot rerun the build with the same commit because it's won't increment the version (versions are based on last tag) and thus try (and fail) to publish packages.

The workflow should more resilient and/or auto-remediate the issue.

Chriscbr:
Can we make the publishing step idempotent (check if the library is published to github packages / npm / etc. and skip if that version is already published)?

MarkMcCulloh:
We can, although we should also ensure that idempotency isn't hiding another underlying issue because under normal circumstances we would never even attempt to publish the same version more than once.

staycoolcall911:
@MarkMcCulloh - I have assigned the issue to you since I noticed you're working on it

staycoolcall911:
Microsoft vaguely told Mark it was fixed... If this doesn't hit us again by the end of sprint 12, we can close this issue

Chriscbr:
Looks like it happened again - in this case, publishing seemed to succeed in this run (https://github.com/winglang/wing/actions/runs/4024885362/jobs/6917693013) but logging into NPM was cancelled (unknown root cause).

MarkMcCulloh:
Looks like it failed because a newer push to main cancelled it, so that's another fragility to account for.

btw, currently, the quickest workaround to being stuck in this is to manually push a new tag to the previous commit to force the next build to increment.

Chriscbr:
@MarkMcCulloh do you think it makes sense to disable the workflow concurrency limit on the main branch? I see how limiting concurrency saves some compute on pull requests, but it seems like it causes more pain/surprise than its worth on the main branch.

MarkMcCulloh:
I'm not sure what pain you're talking about other than the slight chance of canceling during a publish workflow, which is a solvable issue by adding the idempotency you mention. 

If two things get pushed at the same time it's pretty painful to have to wait for a commit to get built that's going to be obsolete right after.

Chriscbr:
[...]

PR #1276, opened 2023-01-27
fix(build): builds on main branch are cancelled by concurrency limits

Chriscbr:
Related to #1132

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,8 @@ on:
   workflow_dispatch: {}
 
 concurrency:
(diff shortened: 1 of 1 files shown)


MarkMcCulloh (changes requested):
With our current workflow as-is, having builds run at the same time would cause issues.

See some discussion here https://github.com/winglang/wing/issues/1132#issuecomment-1406652491

To summarize: We would have to start tagging in the beginning of the workflow to avoid versions running into each other

Chriscbr:
Closing for now

PR #1277, opened 2023-01-27
chore: idempotent publish workflow

### LINKED ISSUE #1132: Fragile publish workflow

MarkMcCulloh:
- Create tag before actually publishing anything
  - This is needed to avoid a mix of things for a given version with the idempotent publishing

- vscode and npm will not attempt to publish if the version already exists

This is difficult to test, may have to iterate after this is merged.

Fixes #1132


.github/workflows/build.yml
```diff
@@ -365,7 +365,23 @@ jobs:
         with:
           node-version: ${{ env.NODEJS_VERSION }}
 
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODEJS_VERSION }}
```
Chriscbr, line 371: Duplicate

.github/workflows/build.yml
```diff
@@ -365,7 +365,23 @@ jobs:
...
+
+      - name: Tag commit
+        uses: tvdias/github-tagger@v0.0.1
+        with:
+          repo-token: "${{ secrets.PROJEN_GITHUB_TOKEN }}"
+          tag: "v${{ needs.prepare.outputs.version }}"
```
Chriscbr, line 377: Is this step idempotent? If not can we make it idempotent?

MarkMcCulloh: It's not, I can see if I can get it to be though

MarkMcCulloh: done (hopefully)

Chriscbr (approved):

PR #1285, opened 2023-01-28
feat(sdk): add api resource tf-aws

RaphaelManke:
# Cloud.Api
## SDK Cloud Api
The SDK Cloud Api is a Resource that contains a OpenApi specification. The Specification is generated by adding route to a empty specification. The specification is then used while synth to Terraform to describe the AWS Api Gateway. Therefore each endpoint or route is represented in the OpenApi specification and has a x-amazon-apigateway-integration property that describes the integration with the Lambda function.

## AWS Api implementation
The AWS implementation of the Cloud.Apiresource is split into two parts. The exported class Api implements the SDK specification. It orchestrates the interaction with the "Parent class" ApiBase for example calling the OpenApi manipulation methods to add new routes. The second part of the AWS implementation is a Abstraction class called WingRestApi. This class should encapsulate all the Terraform specific things. This includes creating a OpenApi extension that glues together the route and the handling lambda function as well as setup the required IAM policies.
All route handler use the same inflight wrapper from api.onrequuest.inflight.ts to handle a inflight request.
This inflight handler maps the AWS ApiGateway event payload to the ApiREquest model from the SDK specification. The ApiRequest model is then passed to the handler function. The handler function returns a ApiResponse model which is then mapped to the AWS ApiGateway response payload.


libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * AWS Implementation of `cloud.Api`.
+ */
+export class Api extends cloud.ApiBase {
+  private readonly api: ApiGatewayRestApi;
+  constructor(scope: Construct, id: string, props: cloud.ApiProps = {}) {
+    super(scope, id, props);
+    this.api = new ApiGatewayRestApi(this, "api", {
+      name: id,
```
Chriscbr, line 30: This name should be generated using the ResourceNames utility

examples/tests/valid/api.w
```diff
@@ -0,0 +1,9 @@
+bring cloud;
+
+let api = new cloud.Api();
+
+let handler = inflight (message: str): str => {
+  return "Hello World";
+};
+// failed when using '/' at the beginning of the path
+api.get("hello", handler);
```
Chriscbr, line 9: I think the public API should expect a slash at the beginning. Even though it doesn't add any extra information per se, it improve readability and helps with maintaining code (imagine someone grepping for "/hello" in their codebase). It's also pretty common in other REST API libraries I've tried (expressjs, flask).

If needed, the individual backends (sim.Api, tfaws.Api) can strip it out so the correct terraform is generated.

```suggestion
api.get("/hello", handler);
```

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+    new ApiGatewayIntegration(this, `integration-${hash}`, {
+      restApiId: this.api.id,
+      resourceId: apiResource.id,
+      httpMethod: apiMethod.httpMethod,
+      integrationHttpMethod: "POST",
+      type: "AWS_PROXY",
+      uri: _tf_function.invokeArn,
+    });
+
+    const region = new DataAwsRegion(this, "region");
```
Chriscbr, line 82: Let's refactor this a little bit so that the region information is provided by the `App` class. 

Here we want to have:
```suggestion
    const region = App.of(this).region;
```

To make that work, in `target-tf-aws/app.ts` we can add:

```ts
private awsRegionProvider: DataAwsRegion | undefined;
public get region(): string {
  if (!this.awsRegionProvider) {
    this.awsRegionProvider = new DataAwsRegion(this, "Region");
  }
  return this.awsRegionProvider.name;
}
```

You can also copy the implementation of `App.of` from `target-tf-azure/app.ts`

libs/wingsdk/test/target-tf-aws/api.test.ts
```diff
@@ -0,0 +1,16 @@
+import { Api } from "../../src/cloud";
+import * as tfaws from "../../src/target-tf-aws";
+import { mkdtemp } from "../../src/util";
+import { tfResourcesOf, tfSanitize, treeJsonOf } from "../util";
+
+test("api with GET route", () => {
```
Chriscbr, line 6: We should probably include these as test cases:
- api with no routes
- api with one GET route
- api with two different GET routes
- api with two different GET routes, where one is a prefix of the other (e.g. "/hello" and "/hello/world")
- throws when get() is called with the same route twice

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+/**
+ * Represents a resource with an inflight "handle" method that can be passed to
+ * `Api.get`.
+ *
+ * @inflight `wingsdk.cloud.IApiOnGetRequestHandlerClient`
+ */
+export interface IApiOnGetRequestHandler extends IResource {}
```
Chriscbr, line 146:
```suggestion
 * @inflight `wingsdk.cloud.IApiGetRequestHandlerClient`
 */
export interface IApiGetRequestHandler extends IResource {}
```

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * Inflight client for `IGetRequestHandler`.
+ */
+export interface IApiOnGetHandlerHandlerClient {
+  /**
+   * Function that will be called when a get request is made to the url.
+   * @inflight
+   */
+  handle(message: string): Promise<void>;
+}
```
Chriscbr, line 157: This should be in `cloud/api.ts` (the inflight interface should be the same among all clouds)

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+    const apiMethod = new ApiGatewayMethod(this, `method-${hash}`, {
+      restApiId: this.api.id,
+      resourceId: apiResource.id,
+      httpMethod: "GET",
+      // FIXME: this cannot be hardcoded
+      authorization: "NONE",
+    });
+
+    const functionHandler = convertBetweenHandlers(
+      this.node.scope!, // ok since we're not a tree root
```
Chriscbr, line 56: For Api it's okay if we make all each of the route resources be children of `Api` (instead of siblings)
```suggestion
      this,
```

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+      `${this.node.id}-OnGetRequestHandler-${hash}`,
+      inflight,
+      join(__dirname, "api.ongetrequest.inflight.js"),
+      "ApiOnGetRequestHandlerClient"
+    );
+    const fn = new cloud.Function(
+      this.node.scope!, // ok since we're not a tree root
```
Chriscbr, line 63:
```suggestion
      this,
```

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+      `${this.node.id}-OnGetRequest-${hash}`,
+      functionHandler,
+      props
+    );
+    if (!(fn instanceof Function)) {
+      throw new Error("Queue only supports creating tfaws.Function right now");
+    }
+    const _tf_function: Function = fn;
```
Chriscbr, line 71:
```suggestion
    const awsFunction: Function = fn;
```

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+  }
+
+  /** @internal */
+  public _bind(host: core.IInflightHost, ops: string[]): void {
+    if (!(host instanceof Function)) {
+      throw new Error("queues can only be bound by tfaws.Function for now");
+    }
+
+    super._bind(host, ops);
```
Chriscbr, line 131: This bind method is only needed when people try to use a `Api` resource inside a Wing [inflight](https://docs.winglang.io/concepts/inflights). For this first PR we don't need to support that (it's a secondary use case) so I think we can just throw an error here:
```suggestion
    throw new Error("Method not implemented.");
```

libs/wingsdk/src/target-tf-aws/api.ongetrequest.inflight.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { IFunctionHandlerClient } from "../cloud";
+
+export class ApiOnGetRequestHandlerClient
+  implements IApiOnGetHandlerHandlerClient
+{
+  private readonly handler: IFunctionHandlerClient;
+  constructor({ handler }: { handler: IFunctionHandlerClient }) {
+    this.handler = handler;
+  }
+  public async handle(event: any) {
```
Chriscbr, line 11: Can we add type information for the `any` here? I think the types are available in `@types/aws-lambda`, check out the code here: https://docs.aws.amazon.com/lambda/latest/dg/typescript-handler.html#async-typescript

This class will need to transform the AWS request input into an cloud-agnostic `cloud.ApiRequest` struct, and then transform the `cloud.ApiResponse` returned by `handler.handle(event)` back into AWS response event object.

Chriscbr (changes requested):
This looks like a great start @RaphaelManke! I gave the code a first pass - let me know if you have any questions or hit any blockers.

libs/wingsdk/src/target-tf-aws/api.ts
RaphaelManke: I didn't know that there is helper for that. 
I'll start with a simple pattern, because I could't find any docs about RestApi name limitations in AWS.

libs/wingsdk/src/target-tf-aws/api.ts
RaphaelManke: ok, I'll change that. Can you explain me what that function does? Is this convert actually needed in this case?

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,136 @@
...
+import { convertBetweenHandlers } from "../convert";
+import { Code } from "../core/inflight";
+import { Function } from "./function";
+import { ApiGatewayIntegration } from "@cdktf/provider-aws/lib/api-gateway-integration";
+import { IResource } from "../core";
+
+/**
+ * AWS Implementation of `cloud.Api`.
+ */
+export class Api extends cloud.ApiBase {
```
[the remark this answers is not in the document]
RaphaelManke: I get your point, but do you have an idea how the OpenAPI spec can be generated? The only advantage I can see right now, is that I don't have to build the ApiRessources from the api path. 

If I take a look at the aws docs https://docs.aws.amazon.com/de_de/apigateway/latest/developerguide/api-gateway-request-validation-sample-api-swagger.html
the openapi spec needs to be extended with amazon specific extensions. Even if the general spec with its routes and methods would be independent, the openapi extensions would be vendor specific.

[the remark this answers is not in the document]
RaphaelManke: maybe a library like this one can help generating the openapi spec
https://github.com/metadevpro/openapi3-ts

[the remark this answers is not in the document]
Chriscbr: I like the idea...

Maybe in terms of implementation we could take a look at how CDK generates APIs from specs, I think they support a feature like this here: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.SpecRestApi.html

BTW, there might be a chance the `spec` will need to be a lazy property since `post()` / `get()` / `add_route()` would modify the spec. We don't have something like this in the SDK yet because we haven't had a use case yet, but here is a test showing how they work: https://github.com/cdk8s-team/cdk8s-core/blob/2.x/test/tokens.test.ts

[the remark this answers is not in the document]
eladb: Yes good point about the lazy property. Alternatively I was thinking we can offer "preSynth" hooks on any construct in the Wing SDK. 

We had this at some point in the CDK and removed it because it's  quite abusable, but I think there's a use case for it.

It's also aligned with @hasanaburayyan's work on plugins.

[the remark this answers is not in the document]
Chriscbr: Got it. Maybe we can discuss this in a separate issue -- my only concern with respect to synth hooks is that they don't always play well with escape hatches.

(Concretely, I could imagine someone creating a wing library where an AwsIamRole is created during presynth (using terraform-cdk). Then, when I want to make a library that composes over their library, if I want to do something like create a resource per Iam statement or if I want to apply CDKTF escape hatches, all my library logic will have to be in presynth, and it seems like we're back at square one where someone wants a postPreSynth stage etc.)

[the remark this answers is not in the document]
Chriscbr: In any case I'll leave it up to @RaphaelManke which direction you'd like to choose, the OpenAPI direction sounds promising but we can also continue with the direction that does not use OpenAPI to start and maybe refactor later. 🙂

libs/wingsdk/src/target-tf-aws/api.ts
Chriscbr: Sure, I'd be happy to. The purpose of `convertBetweenHandlers` is to perform a conversion to and from the APIs that the Wing user implements against in their Wing code, and the interface of the input/output expected by the cloud provider.

For example, in Wing, the end goal is to let a user write something like:

```js
bring cloud;

let api = new cloud.Api();
api.get("/ducks", inflight (req: cloud.ApiRequest): ApiResponse => {
  // do stuff with req
  return ApiResponse {
    status: 200,
    body: "Ducks created"
  };
});
```

When this is implemented on other clouds like Azure and GCP, it will still be the same code! However, an AWS Lambda that is hooked up to an AWS API gateway receives a different shape of input, and expect a different shape of output. Perhaps it expects `statusCode` instead of `status`, or includes an `httpMethod` instead of `method`. That's fine, as long as the SDK wraps the user's inflight code in another function that transforms the fields for it.

So, conceptually that's the purpose. In terms of implementation, the function accepts an `Inflight` (class from the SDK) which is treated as an immutable data structure. Since it's immutable, we cannot mutate it with the wrapped code, and instead it creates a new `Inflight` with the new code. (At least, that's how it works at the moment). Let me know if that explanation helps at all 😅

libs/wingsdk/src/target-tf-aws/api.ongetrequest.inflight.ts
```diff
@@ -1,14 +1,27 @@
-import { IApiOnGetHandlerHandlerClient } from "./api";
+// TODO: remove this line and discuss with @Chriscbr about how to handle this
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
```
Chriscbr, line 3: I'm OK leaving the eslint warning if it compiles :) We could also just hand-copy a simple version of the API / the fields we care about and duplicate it into our source code (I feel like it's unlikely they introduce breaking changes since it's a live AWS service)

RaphaelManke: I see two option how to resolve this error. Either move the dependecy from devDependency to dependency or add a exclusion for @types packages in the eslint rule

libs/wingsdk/src/target-tf-aws/api.ts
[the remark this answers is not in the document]
RaphaelManke: for now I added a `addResource` helper function that adds a route to the api. I still like the openapi spec idea because it would also enable the wing console to render the api spec and let the user invoke the api there. But I am not very familiar with the Token resolution and the lazy method.

libs/wingsdk/test/target-tf-aws/api.test.ts
RaphaelManke: WIP I added some test cases :smile:

libs/wingsdk/src/target-tf-aws/api.ongetrequest.inflight.ts
Chriscbr: I like the second option (first is also ok).

Another thing you could try is:
```ts
import type { ... } from "aws-lambda";
```

RaphaelManke: the import type version works

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,274 @@
...
+    //     const deps: string[] = [];
+    //     console.log(this.node.children);
+    //     console.log("produce dependson");
+    //     for (const dep of this.redeployTrigger) {
+    //       deps.push(this.node.findChild(dep).node.id);
+    //     }
+    //     console.log("dependsOn", deps);
+    //     return deps;
+    //   },
+    // });
```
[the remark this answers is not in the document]
Chriscbr: Hey @RaphaelManke - yeah this seems like a tricky issue. From what I can see, it seems to me like CDKTF doesn't really like when you provide lazy list values (I tried a couple variations of your code and hit similar issues). I've pushed a commit that works around this by using an escape hatch - I think this is OK as long we can specify all dependencies ourselves.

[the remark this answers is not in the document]
RaphaelManke: Ok your suggestion works (I had to add the resource type prior to the `friendlyUniqueId`). 
Now there is one small timing issue left, not sure if that can be solved. 

When a route is removed, terraform detects the change and deletes the integration, lambda etc. But this also needs a deployment. Since the dependent resources are deleted they are not part of the dependsOn. This results in the following behaviour. 
The deployment is created prior to deleting the resources from the apigateway. So the old resources are still part of the apigateway at the time the deployment is done. Since a deployment is a snapshot of the resources configured at this time the old resource is still part of the snapshot. 

Now one idea I have is to use the [sleep](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep) to wait a few seconds before creating a deployment. But this doesn't feel to be a good solution because it slows down the deployment by a fixed time value and does not guarantee the deployment always uses the new configuration because if the deletion takes longer than the configured sleep time it would still create a wrong api state afterwards.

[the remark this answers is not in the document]
Chriscbr: > The deployment is created prior to deleting the resources from the apigateway

Hmm -- I'm wondering if this is caused by the `create_before_delete` lifecycle option that is enabled on the deployment? I see that the Terraform docs recommends adding that field, but I wonder if it could be causing problems in the route-removal scenario...

I'm not sure of the best solution is at this moment. One idea I had from the terraform angle is that there is an option to set the "count" of a resource to 0, so that it's still in your template and can be depended on. But I don't know how we would generate this in Wing, since it would require remembering what resources you previously had in your app...

I'm okay with trying the sleep approach for now. I've posted a question on StackOverflow related to your issue so if I get an answer with a better solution I'll update you 🙏

libs/wingsdk/src/target-tf-aws/api.ts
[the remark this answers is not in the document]
RaphaelManke: Sorry for the back and forth, I now refactored to use the openApi approach and generate a spec file and the generate the restApi from that. This makes the generation actually very simple because one does not have to create resource, method and integration as terraform resources. On top of that it fixes the deployment issue. Not a deleted resource is not deployed anymore. I will continue to follow the the openapi spec approach.

libs/wingsdk/src/target-tf-aws/api.ts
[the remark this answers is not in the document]
RaphaelManke: switching to the apispec approach fixes the problem 
https://github.com/winglang/wing/pull/1285#discussion_r1106407345

examples/tests/valid/api.w
```diff
@@ -0,0 +1,12 @@
+bring cloud;
+
+let api = new cloud.Api();
+
+let handler = inflight (message: str): str => {
+  return "Hello World";
+};
+// failed when using '/' at the beginning of the path
```
eladb, line 8: > `// failed when using '/' at the beginning of the path`

Not sure I understand this comment

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    props;
+
+    this.display.title = "Api";
+
+    this.display.description = "A cloud API";
+  }
+  /**
+   * Add a function to handle GET requests to a route.
+   */
```
eladb, line 40: Add `@param`s

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,91 @@
...
+    super(scope, id);
+
+    if (!scope) {
+      return;
+    }
+
+    props;
+
+    this.display.title = "Api";
+
```
eladb, line 35: Redundant newline:

```suggestion
```

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,91 @@
...
+    this.display.description = "A cloud API";
```
eladb, line 36:
```suggestion
    this.display.description = "A REST API endpoint";
```

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  /**
+   * Add a function to handle GET requests to a route.
+   */
+  public abstract get(
+    route: string,
+    inflight: Inflight,
+    props?: ApiOnGetProps
```
eladb, line 44:
```suggestion
    props?: ApiGetProps
```

libs/wingsdk/src/cloud/function.ts
```diff
@@ -193,12 +193,15 @@ export interface IFunctionHandler extends IResource {}
 /**
  * Inflight client for `IFunctionHandler`.
  */
-export interface IFunctionHandlerClient {
+export interface IFunctionHandlerClient<
```
eladb, line 196: We can't use generics in public API

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,91 @@
...
+    inflight;
+    props;
+    throw new Error("Method not implemented.");
+  }
+}
+
+/**
+ * Inflight client for `IGetRequestHandler`.
+ */
+export interface IApiGetHandlerHandlerClient<ApiRequest, ApiResponse> {
```
eladb, line 85: Why is this specific to "get"?

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   *  Create a function to handle GET requests to a route.
+   *
+   * @param route
+   * @param inflight
+   * @param props
+   */
+  public get(route: string, inflight: Inflight, props?: ApiOnGetProps): void {
```
eladb, line 74: I would add at least another method initially (e.g. `post`) to make sure you are not designing only for one use case

libs/wingsdk/src/target-tf-aws/api.ongetrequest.inflight.ts
```diff
@@ -0,0 +1,25 @@
...
+{
+  private readonly handler: IHandler;
+  constructor({ handler }: { handler: IHandler }) {
+    this.handler = handler;
+  }
+  public async handle(
+    request: APIGatewayProxyEvent
+  ): Promise<APIGatewayProxyResult> {
+    const resp = await this.handler.handle(request);
+    // TODO: this is just a WIP to get the API working, we need to figure out what needs to be returned by the inflight function
```
eladb, line 19: I believe the [spec] stipulates this should be an `ApiResponse` struct.

[spec]: https://docs.winglang.io/reference/wingsdk-spec#api

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,217 @@
...
+  POST = "post",
+  PUT = "put",
+  DELETE = "delete",
+  PATCH = "patch",
+  HEAD = "head",
+  OPTIONS = "options",
+  ANY = "any",
+}
+
+export class WingRestApi extends Construct {
```
eladb, line 114:
```suggestion
class WingRestApi extends Construct {
```

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,217 @@
...
+  public readonly api: ApiGatewayRestApi;
+  private readonly deployment: ApiGatewayDeployment;
+  private readonly region: string;
+  private apiSpec: any = {
```
eladb, line 118: Feels like all the OpenAPI generation stuff can go into `cloud.ApiBase` and reused across implementations.

eladb (changes requested):
Initial comments.

examples/tests/valid/api.w
RaphaelManke: this is leftover from the old implementation where I had to create multiple `apigateway_resources` which mustn't contain the slash. Now with the openapi approach this is explicitly needed (which makes the implementation easier)

libs/wingsdk/src/target-tf-aws/api.ts
RaphaelManke: I will try that

examples/tests/valid/api.w
RaphaelManke: This is now solved.

libs/wingsdk/src/cloud/api.ts
RaphaelManke: It was the starting point with a lot of copy pasting from the bucket and queue implementation. I think you are right that this can be used for all http methods.

libs/wingsdk/src/target-tf-aws/api.ongetrequest.inflight.ts
RaphaelManke: yes you are right.
But I have problem here as well. 
We try to describe the api request on a cloud independent "language". That means we say for example that the response has a key `status` whereas when implemented the aws apigateway expect a key `statusCode`. 
I am not sure how to model that with types? 
If you look at this handle method. It should implement the specification from the cloud spec. But in the real world the request does look different. On which level do I map the `APIGatewayProxyEvent` to the `ApiRequest` from the cloud spec?

If I change the handle method to align with the `ApiRequest` interface this will be not work in the real world because the apigateway in aws creates a  `APIGatewayProxyEvent`.

libs/wingsdk/test/target-tf-aws/api.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  const output = app.synth();
+
+  // THEN
+  // expect(tfResourcesOf(output)).toEqual(["aws_api_gateway_rest_api"]);
+  expect(tfResourcesOfCount(output, "aws_api_gateway_rest_api")).toEqual(1);
+  expect(extractApiSpec(output)).toMatchSnapshot();
+  expect(tfResourcesOfCount(output, "aws_lambda_function")).toEqual(1);
+});
+
+test("api with multiple GET route and one lambda", () => {
```
Chriscbr, line 89: I think this is the same as the test above

Chriscbr (commented):
Great progress, the OpenAPI spec direction looks promising 😊

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,144 @@
...
+  PUT = "PUT",
+  /** Delete */
+  DELETE = "DELETE",
+  /** Connect */
+  CONNECT = "Connect",
+  /** Options */
+  OPTIONS = "OPTIONS",
+  /** Patch */
+  PATCH = "PATCH",
+}
```
RaphaelManke, line 104: running `npx nx wing -- compile --target=tf-aws ../../examples/tests/valid/api.w` ind the `wing/apps` folder produces a error 
```
thread '<unnamed>' panicked at 'Expected cloud.HttpMethod to be defined: TypeError { message: "Unknown symbol \"HttpMethod\" in namespace \"cloud\"", span: WingSpan { start: Point { row: 0, column: 0 }, end: Point { row: 0, column: 0 }, start_byte: 0, end_byte: 0, file_id: "" } }', libs/wingc/src/type_check/jsii_importer.rs:132:14
stack backtrace:
wasm://wasm/09bd1d76:1
RuntimeError: unreachable
    at __rust_start_panic (wasm://wasm/09bd1d76:wasm-function[14371]:0x53cd5d)
    at rust_panic (wasm://wasm/09bd1d76:wasm-function[6112]:0x487c9b)
    at std::panicking::rust_panic_with_hook::hc5c73bd02fe928d4 (wasm://wasm/09bd1d76:wasm-function[996]:0x2a9a6d)
    at std::panicking::begin_panic_handler::{{closure}}::hcd8b0196239bb506 (wasm://wasm/09bd1d76:wasm-function[6875]:0x4a780e)
    at std::sys_common::backtrace::__rust_end_short_backtrace::h634ecf1fa70ae144 (wasm://wasm/09bd1d76:wasm-function[13423]:0x535a2f)
    at rust_begin_unwind (wasm://wasm/09bd1d76:wasm-function[10014]:0x4ffca2)
    at core::panicking::panic_fmt::h1024bc10a158814e (wasm://wasm/09bd1d76:wasm-function[12630]:0x52b9c5)
    at core::result::unwrap_failed::h4f10a8b6d25baa65 (wasm://wasm/09bd1d76:wasm-function[7985]:0x4cd4f7)
    at core::result::Result<T,E>::expect::hb29bc82c2258788a (wasm://wasm/09bd1d76:wasm-function[3682]:0x3fcc6d)
    at wingc::type_check::jsii_importer::JsiiImporter::lookup_or_create_type::h4e62f125c2534baf (wasm://wasm/09bd1d76:wasm-function[635]:0x23e9d2)
```

and I have no idea how to fix that. Maybe some missing export? Do I have to register this enum somewhere?
@Chriscbr can you have a look at this?

libs/wingsdk/test/target-tf-aws/api.test.ts
RaphaelManke: yes

libs/wingsdk/src/cloud/api.ts
Chriscbr: @RaphaelManke ah, I think this is a limitation as the compiler don't support importing enums quite yet. https://github.com/winglang/wing/issues/1484 For now we might have to use strings in the API a placeholder 😅

RaphaelManke: ok i commented it out.

libs/wingsdk/src/target-tf-aws/api.ts
RaphaelManke: started to move it up see https://github.com/winglang/wing/pull/1285/commits/6b5141121e1cd57ce4f8145121a763bb045e76fe

The idea would be that the apiBase has a function to add a route to the apiSpec and gets some extension values from the cloud specific implementation that is then added to the route. At the end the ApiBase gives back the finalized apiSpec.

libs/wingsdk/src/cloud/function.ts
RaphaelManke: ok i removed it

libs/wingsdk/src/cloud/api.ts
RaphaelManke: added a post method as a second one. Once those two work its easy to add the missing ones.

libs/wingsdk/src/target-tf-aws/api.ongetrequest.inflight.ts
RaphaelManke: Have a look at this commit https://github.com/winglang/wing/pull/1285/commits/8a16798f032af234ee5f769c78a7f34dc8161de1 in file 
libs/wingsdk/src/target-tf-aws/api.onrequest.inflight.ts

What do you think?

Chriscbr: Looks great 👍

RaphaelManke:
# Api - status quo

## SDK Cloud Api

The SDK Cloud Api is a Resource that contains a OpenApi specification. The Specification is generated by adding route to a empty specification. The specification is then used while synth to Terraform to describe the AWS Api Gateway. Therefore each endpoint or route is represented in the OpenApi specification and has a `x-amazon-apigateway-integration` property that describes the integration with the Lambda function.

### What works
- The Api Resource has a OpenApi specification as well as getter and setter methods to extend the specification or retrieve it while synth to Terraform.
- All possible HTTP methods are supported (but not all are implemented yet).

### Open Questions
- OpenApi specification typing - should we create our own typing or is there one we can use
- Typing for a JSON response - I didn't find an existing typing for a JSON response. I created a own one for now.
- Cors as part of this PR
- The Enum for HTTP methods isn't working because of a bug see [#1484](https://github.com/winglang/wing/issues/1484).
- There is a `IApiEndpointHandlerClient` interface which isn't used (yet). I am not sure about the this client concept here. Do I have to implement this interface anywhere? 

## AWS Api implementation

The AWS implementation of the `Cloud.Api`resource is split into two parts. The exported class `Api` implements the SDK specification. It orchestrates the interaction with the "Parent class" ApiBase for example calling the OpenApi manipulation methods to add new routes. The second part of the AWS implementation is a Abstraction class called `WingRestApi`. This class should encapsulate all the Terraform specific things. This includes creating a OpenApi extension that glues together the route and the handling lambda function as well as setup the required IAM policies.
All route handler use the same inflight wrapper from `api.onrequuest.inflight.ts` to handle a inflight request. 
This inflight handler maps the AWS ApiGateway event payload to the `ApiREquest` model from the SDK specification. The `ApiRequest` model is then passed to the handler function. The handler function returns a `ApiResponse` model which is then mapped to the AWS ApiGateway response payload.

### What works
- The OpenApi extension is created and added to the OpenApi specification.
- A handler function only added once to the stack.
### Open Questions
- Is this inflight handler the correct way to handle inflight requests? I am unsure because there is interface that can be used to define how the handler should look like in general. 
- Is is correct that the Inflight should return a `ApiResponse` object? Or should it just return a JSON object that is then transformed to the `ApiResponse` object?


## Examples
I added a example File to play around with the Api resource and to be able to deploy it to a real AWS account. 
Now I am looking for more examples to test the Api resource. I am not sure if I am using the Api resource in the correct way.

These examples are in my mind: 

**Return simple string body**
```js
bring cloud;

let api = new cloud.Api();

let handler = inflight (message: str): cloud.ApiResponse => {
  return {
    "body": "Hello",
    "status": 200,
  };
};
api.get("/abc", handler);
```

**Return JSON body**
```js
bring cloud;

let api = new cloud.Api();

let handler = inflight (message: str): cloud.ApiResponse => {
  return {
    "body": {
      "foo": "bar",
    },
    "status": 200,
  };
};
api.get("/abc", handler);
```

**Call another resource within inflight**
```js
bring cloud;

let api = new cloud.Api();
let counter = new cloud.Counter();

let handler = inflight (message: str): cloud.ApiResponse => {
  let count = counter.inc();

  return {
    "body": {
      "counter": count,
    },
    "status": 200,
  };
};
api.get("/abc", handler);
```

Right now all of the examples throw an error during synth. I am not sure if I am using the Api resource in the correct way. I would be happy to get some feedback on this.


## Next Steps
I think the next steps would be:
- discuss the current implementation and if it is the correct way to implement the Api resource
    - clarify the open questions
- define some examples that should work
- rebase the branch to the latest main branch (I didn't do this yet because it often messes up the comments to specific lines of code)
- fix the code to make the examples work
- validate if the code documentation is correct and self explaining enough 

I think we should not extend the feature set in this PR. I would like to keep it small and discuss the implementation first. After this PR is merged we can add more features like CORS, Authentication, Content-type header, etc. What do you think?

Chriscbr:
Hey @RaphaelManke - this looks great. Just letting you know I'm working on reviewing your PR, thanks for your patience 🙏

Chriscbr (approved):
I've left a few small comments but they're all very tiny so I've approved the PR. The code was very easy to read 😄 

I tried out compiling the example you provided and I was able to deploy and run it on AWS! Here's a screenshot to show:
Screenshot 2023-02-21 at 5 49 13 PM

I did notice an issue where the headers were not being returned properly, but I did some investigation and I think it could be an issue with the compiler (currently `Map<str>` values are translated into JavaScript `Map` objects, and I'm not sure if this is correct - I'll open a separate issue). In the meantime, I think we can use some code like this so either `Map` or plain JSON will work:

```ts
function mapCloudApiResponseToApigatewayResponse(resp: ApiResponse) {
  return {
    statusCode: resp.status,
    body: JSON.stringify(convertToJson(resp.body)),
    headers: convertToJson(resp.headers),
  };
}

// TODO - remove this once the compiler stops emitting Map objects instead of records
function convertToJson(value: Map<string, any> | Record<string, any>) {
  if (value instanceof Map) {
    return Object.fromEntries(value);
  }
  return value;
}
```

> * OpenApi specification typing - should we create our own typing or is there one we can use

Types are definitely an improvement wherever possible, but I'm fine with the current approach since they're internal. 🙂  The limitation you might find is that we cannot use types from external libraries in our public APIs (public fields, return types, etc.) will cause a build error due to restrictions from JSII, so these types would only be useful as an internal implementation detail. (In order to provide public types, we would have to define our own interfaces into the SDK code.)

> * Typing for a JSON response - I didn't find an existing typing for a JSON response. I created a own one for now.

👍  This is something we'll iterate on more as we add Json support to the language, stay tuned...

> * Cors as part of this PR

Let's leave it out for now.

> * The Enum for HTTP methods isn't working because of a bug see [#1484](https://github.com/winglang/wing/issues/1484).

Thanks for mentioning this. We have enums support planned for the compiler in the near future.

> * There is a `IApiEndpointHandlerClient` interface which isn't used (yet). I am not sure about the this client concept here. Do I have to implement this interface anywhere?

This is due to some technical debt we have 😅.  In the future we would like many functions in the SDK that currently expect `core.Inflight` will instead expect `IApiEndpointHandlerClient` or other interfaces with more precise type information. This way, the compiler will not allow any inflight functions to be passed into `api.get` -- only those inflight functions that are typed as `inflight (input: cloud.ApiRequest): cloud.ApiResponse`. We can't make the change yet because the compiler doesn't fully support interfaces, so for now they are just placeholders. So the code in your PR is all good!

> * Is this inflight handler the correct way to handle inflight requests? I am unsure because there is interface that can be used to define how the handler should look like in general.

See my answer above.

> * Is is correct that the Inflight should return a `ApiResponse` object? Or should it just return a JSON object that is then transformed to the `ApiResponse` object?  

I think the code as it is now is good -- Inflight does not specify a specific return type for functions but ApiOnRequestHandlerClient is responsible for returning a cloud.ApiResponse so that will be used.   The main purpose of `ApiResponse` is to provide a standard API shape so the same response a user returns in their code can be converted to the response requested by a cloud provider (so e.g. we can implement `cloud.Api` on Azure in the future).

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,229 @@
...
+export interface ApiProps {}
+/**
+ * The OpenAPI spec.
+ */
+// TODO: This should be a type from the OpenAPI spec. We can either use a external package or define it ourselves.
+export type OpenApiSpec = any;
+
+/**
+ * The OpenAPI spec extension for a route.
+ * TODO: Unsure how to type this.
```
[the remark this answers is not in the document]
RaphaelManke: Which documentation do you refere to? OpenApi documenatation like this one? https://spec.openapis.org/oas/v3.0.3

libs/wingsdk/src/cloud/api.ts
```diff
@@ -0,0 +1,229 @@
...
+   * @param method The method to add.
+   * @param apiSpecExtension The extension to add to the api spec for this route and method.
+   *
+   * @internal
+   * */
+  public _addToSpec(
+    route: string,
+    method: string,
+    apiSpecExtension: OpenApiSpecExtension
+  ) {
```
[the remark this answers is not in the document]
RaphaelManke: Thas a good catch :) The AWS implementation does not need any thypes for path parameters to work. But I'll check how that needs be presented in the api spec.

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -0,0 +1,272 @@
...
+import { ApiGatewayRestApi } from "@cdktf/provider-aws/lib/api-gateway-rest-api";
+import { ApiGatewayStage } from "@cdktf/provider-aws/lib/api-gateway-stage";
+import { LambdaPermission } from "@cdktf/provider-aws/lib/lambda-permission";
+
+import { Lazy } from "cdktf";
+import { Construct } from "constructs";
+import { App } from "./app";
+import { Function } from "./function";
+import { core } from "..";
+import * as cloud from "../cloud";
```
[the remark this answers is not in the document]
RaphaelManke: Why do you prefere a wildcard import? Named imports are better for tree-shaking, aren't they?

RaphaelManke:
I am now confident that the Api Resource is working 😃 
Now I was thinking if the Api Spec is incomplete? Don't we need some inflight method to actually invoke the Api? This would enable us to invoke the api from e.g. a lambda. But maybe this is a new feature...

libs/wingsdk/src/cloud/api.ts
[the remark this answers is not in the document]
Chriscbr: Yep 👍

examples/tests/valid/api.w
```diff
@@ -0,0 +1,23 @@
...
+//     },
+//     status: 200,
+//     headers: {
+//       "Content-Type": "text/plain",
+//     },
+//   };
+//   return resp;
+// };
+
+// api.get("/hello/world", handler);
```
Chriscbr, line 23: Can we uncomment this?

libs/wingsdk/src/target-tf-aws/api.onrequest.inflight.ts
```diff
@@ -0,0 +1,76 @@
...
+function convertMapToObject(map: Json | undefined) {
+  if (!map) {
+    return undefined;
+  }
+  if (!(map instanceof Map)) {
+    // the inflight returns a Map instead of a plain object
+    throw new Error("Expected a Map");
+  }
+  return Object.fromEntries(map);
+}
```
Chriscbr, line 52: Sorry for the bother 😅 - I think we can remove this conversion since we have merged in a fix here: https://github.com/winglang/wing/pull/1653

libs/wingsdk/test/target-tf-aws/api.onrequest.inflight.test.ts
```diff
@@ -0,0 +1,288 @@
...
+      body: JSON.stringify({}),
+      headers: {},
+      path: "/",
+      httpMethod: "GET",
+    };
+
+    const handlerResponse = {
+      status: 200,
+      body: new Map().set("key", "value"),
+      headers: new Map().set("header-1", "value-1"),
```
Chriscbr, line 21:
```suggestion
      body: { "key": "value" },
      headers: { "header-1": "value-1" }
```

libs/wingsdk/test/target-tf-aws/api.test.ts
```diff
@@ -0,0 +1,297 @@
...
+const extractApiSpec = (output: any) => {
+  const jsonOutput = JSON.parse(output);
+  const api = jsonOutput.resource.aws_api_gateway_rest_api as {
+    body: string;
+  }[];
+  const body: string = Object.values(api)[0].body;
+  return JSON.parse(body) as { paths: any };
+};
+
+test("api with GET route at root", () => {
```
Chriscbr, line 16: Add a unit test where a path parameter is used? (like `/hello/{world}`)

Chriscbr (commented):
A few small fixes but otherwise looks ready to ship 😄

Chriscbr:
> Don't we need some inflight method to actually invoke the Api? This would enable us to invoke the api from e.g. a lambda. But maybe this is a new feature...

Yes I think for now that will be a separate feature. This would be `Api`'s inflight methods

examples/tests/valid/api.w
RaphaelManke: Unless you have way to skip a hangar test for this file not until we have a simulator implementation.

Chriscbr: Ah I see... maybe just add a comment at the top of the file, or we can remove this test for now

RaphaelManke:
I added a way to skip a example file for hangar tests. 
On top of that I think we are Ready for a final review or merge 😃

Chriscbr (approved):
🚀 🚀

staycoolcall911:
@RaphaelManke - this is huge! Nicely done!

Issue #1253, opened 2023-01-25
Use versions and aliases to prevent downtime upon deployment


schosterbarak:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

When deploying a lambda there is some down time.
There are various approaches to solving it: 
https://www.aws-community-day.de/slides/Continuous%20deployment%20of%20Serverless%20resources.pdf |https://docs.aws.amazon.com/cdk/api/v1/docs/aws-lambda-readme.html#versions

### Use Cases

minimize deployment downtime

### Component

Compiler, SDK

staycoolcall911:
Thanks @schosterbarak, great use case. 

@eladb, @hasanaburayyan - is that something we may be able to solve with a [wing plugin](https://github.com/winglang/wing/pull/1218)? If so, then maybe we can just add this use case to the RFC and close the issue.

eladb:
I think we can bake this behavior by default to the `Function` implementation for `tf-aws`.

eladb:
Yes, definitely not P2.

(also, you can make as good-first-issue)

ShaiBer:
> Can I try solving this? 🙌🏻

Sure @raywonkari , I assigned you to it now, so no one will try to work on it in parallel.
Please let us know if you need any help.

staycoolcall911:
Already supported - closing

PR #1373, opened 2023-01-31
feat(sdk): enable publishing lambda function version

raywonkari:
- This PR enables publishing function versions, which could help avoid downtime while deploying new functions.
- return both ARN as well as Qualified ARN, so we could use whichever value we want

https://github.com/winglang/wing/issues/1253


eladb:
Thanks!

Can you please add some details about this in the description?

Is there any implication on users? 

The PR title should reflect what problem does this solve or what new capability is now supported.

revitalbarletz:
This issue is correlated to the issue @schosterbarak has opened as a suggestion to prevent downtime time upon deployment - https://github.com/winglang/wing/issues/1253

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -35,6 +35,8 @@ export class Function extends cloud.FunctionBase {
   private policyStatements?: any[];
   /** Function ARN */
   public readonly arn: string;
+  /** Function Qualified ARN */
```
eladb, line 38: Why not always return this instead of ARN?

If the "safe" version is the qualified arn, then we should only use it

eladb (changes requested):

eladb (commented):
LGTM. @Chriscbr ?

hasanaburayyan:
> Did you have a chance to test this on AWS @raywonkari ? We don't have rigorous integration tests yet but we just want to make sure simple apps like app from the [Wing getting started guide](https://docs.winglang.io/getting-started) still work.
> 
> Also I was wondering if the `invoke` method on `cloud.Function` is used, will it still always go to the latest published version?

@Chriscbr To your point, there are definitely implications with the versioning change for invoking the function. When running the invoke command in the client you need to append the alias or version to (name | arn). In this case @raywonkari already updated the function arn to have the version using the `qualified_arn` attribute of the function. 

```typescript
this.arn = this.function.qualifiedArn;
```

For sure we should test this with the GSG app, but I think the invoke target flow should be fine. Its the permissions Im curious about does changing instance variable arn to be the qualified arn cause us binding issues elsewhere. So maybe it would be worth testing this on an application that uses the Function bind method maybe like:

```typescript
bring cloud;

let f1 = new cloud.Function(inflight (message: str): str => {
  bucket.put("hello.txt", "Hello");
  return "Hello";
});

queue.on_message(inflight (message: str): str => {
  let res = f1.invoke(message);
  return res;
});
```

My hypothesis is that IAM policies are going to have to be updated each time a new function version is deployed which I dont think we want. We could make use of aliases if we want to only allow invocations on the latest version but yea.. just some thoughts.

libs/wingsdk/src/target-tf-aws/function.ts
hasanaburayyan: @eladb @raywonkari  I just noticed this was a discussion 😅 I made a comment about this below. I think what this will end up doing is binding the IAM permissions to a specific version which means that every function change will also deploy a new IAM policy update. Which if we want to limit the policy to only allow the function to invoke the latest version we could do that with an alias that would not require a new policy change per deploy.

It may be worth thinking about exposing both arn maybe?

eladb: So maybe we should add a `unqualifiedArn` or something like that, so `arn` will point to the qualified one which is the current version (which I am assuming is what users usually want), and if you want to point to any version, you can use the `unqualifiedArn`

(design tenet: best practice should be the path of least resistance).

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -35,6 +35,8 @@ export class Function extends cloud.FunctionBase {
   private policyStatements?: any[];
   /** Function ARN */
```
eladb, line 36: Please add information in the docstring explaining exactly what's going to be returned from each type of ARN property.

tools/hangar/README.md
```diff
@@ -1,6 +1,6 @@
 # Hangar - Wing Toolchain Test Suite
 
-See [contributing.md](../../CONTRIBUTING.md#🧪-how-do-i-run-e2e-tests) to learn how to run the e2e tests.
+[How to run E2E tests?](https://docs.winglang.io/contributors/development#-how-do-i-run-tests)
```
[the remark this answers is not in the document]
staycoolcall911: Got it yeah, you are correct - good change 😄

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -146,7 +149,8 @@ export class Function extends cloud.FunctionBase {
         : Duration.fromMinutes(1).seconds,
     });
 
-    this.arn = this.function.arn;
+    this.arn = this.function.qualifiedArn;
```
hasanaburayyan, line 152: Let’s change the arm that gets used in _bind to use the unqualified arn to avoid undesired recreations for each version

hasanaburayyan (changes requested):

libs/wingsdk/src/target-tf-aws/function.ts
hasanaburayyan: Yea you are right, I thought we used the instance variable arn in the _bind method. I think this is actually really confusing since we are basically inverting the exposed arns between terraform resource Function and our AWS function.

Maybe just my mental model but I think I would want to deal with the unqualified arn more often/by-default over the qualified arn. Perhaps instead do

```typescript
this.arn = this.function.arn
this.qualifiedArn = this.function.arn
```

@eladb @Chriscbr any strong feelings/thoughts here?

eladb: It is confusing to users or or maintainers? externally we only expose "arn" and "unqualifiedArn"..

See the discussion above, we wanted to help users to the right thing most times, which is to use the qualified ARN. If we have `arn` they will just use it.

Here's an alternative suggestion: we can expose `qualifiedArn` and `unqualifiedArn` and then it will require the user to understand these concepts.

Not sure that's a better solution because it adds cognitive friction where it's not really needed

hasanaburayyan: I think it’s confusing to both users and maintainers. By definition the qualified arn is the arn + suffix, and unqualified is just arn with no suffix. So I think to expose the qualified arn as just arn could lead to some misunderstandings. 

I’m also unsure (thinking out loud) that using the qualified arn is desired in most instances, I imagine most users will only care about invoking latest version of lambda code which the unqualified arn will always point to latest.

eladb: Okay makes sense. Let's just leave it with ARN and qualifiedArn

eladb (approved):

libs/wingsdk/src/target-tf-aws/function.ts
staycoolcall911: @raywonkari - I believe what they mean is if you could add a @returns docstring (annotation in the comments) over arn and qualifiedArn to explain what they are.

hasanaburayyan (approved):

Issue #1053, opened 2023-01-05
`Json` and `MutJson`


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Note: this should replace [this section of the spec](https://docs.winglang.io/reference/spec#114-struct-type), so please update this section as part of this task.

We introduce two new built-in types `Json` and `MutJson`:

```ts (wing)
let json = Json {
  a: "some string",
  b: 15, 
  c: [ "a", "mb"], 
  j: { 
    a: "sss",
  }
};

let mut_json: MutJson = new MutJson(json);
mut_json.d = "this is allowed"; 
let j : Json = mut_json.to_json();
```

### Use Cases

A couple of question to consider: 
- How do we convert Wing's `struct` from and to `Json`?
- What is the API to parse string to Json?
- Do we support both `json.a` and `json['a']` and `json.get("a")`? 
- How does the code looks like to get the string value from a path, e.g.:
```ts (wing)
let j = { 
  a: "something"
};
let some_value: Json = j.get("a"); // get method returns a Json object
let str_value = some_value.str_value(); // ?? what to do if a is not a string
```
- Another question I might add is how the compiler should choose to infer whether `{ "a": 1, "b": 2 }` is a Json or Map<num> (or if we should avoid guessing/inference altogether -- so the user is required to specify Json { ... } or Mut<num> { ... }).

### Implementation Notes

Originally this type was called [Struct](https://docs.winglang.io/reference/spec#114-struct-type)

### Component

Compiler, SDK

Sub tasks list by @hasanaburayyan 

- [ ] Assignment
	- [x] Literals **(P1)**
	- [x] From native types **(P1)**
	- [x] To native types `from_json()` **(P1)**
		- [ ] unsafe mode **(P2)**
		- [ ] `try_from_json()` **(P2)**
- [x] Json nested properties **(P1)**
	- [x] get and set
	- [ ] #1680
	- [ ] `[]` index notation
- [x] Print Json (directly + in string interpolation) **(P1)**
- [x] #1679
	- [x] Json.clone() && Json.clone_mut() **(P1)**
	- [x] Json.keys()  **(P1)**
	- [x] Json.values()  **(P1)**
	- [x] Json.delete() **(P1)**
	- [x] Json.to_str() **(P1)**
	- [x] Json.parse() **(P1)**
	- [ ] Json.equals() **(P2)**
[...]

PR #1394, opened 2023-01-31
rfc: task list with json, uuid and duration

ekeren:
This is a proposal for story 13, that includes: 
- Json and MutJson
- Using `bring jsii` in `inflight` (replace counter with uuid.v4())
- Using duration
- Bucket cross region replication 
- VScode autocomplete capabilities 
- Least privilege permissions 

@yoav-steinberg, please notice that we are using a bunch of static functions in this code, mainly for 
str<->Json parsing and writing. Can you suggest an alternative api for static functions in this case? 

### DIFF
--- a/docs/06-contributors/999-rfcs/000-stories/story-13.md
+++ b/docs/06-contributors/999-rfcs/000-stories/story-13.md
@@ -1,0 +1,169 @@
+# User Story 13 - Task List with uuid, Json and duration
+
+> **Status**: Expected released on 2023/02/16
+
+The following code is an iteration of Wing's task list.
+This version of it includes the following functional changes:
+- Using npm's uuid package inflight instead of the counter resource 
+- Storing Json instead of a string
+- Using duration for estimated effort
+
+Also, it includes some non-functional requirements: 
+- The Bucket should be replicated across different regions 
+  - +100 if we are able to do the same for azure as well
+- Least privileged permissions are granted, to resources 
+- The code should work on sim, aws, azure
+- VScode should be able to autocomplete 
+- Console should be able to show all resources, and interact with them
+
+
+
+```ts (wing)
+bring cloud;
+// need to run: npm install uuid --save
+bring uuid;
+
+/**
+ * Represents a cloud task list.
+ */
+resource TaskList {
+  /** stores the tasks */
+  _bucket: cloud.Bucket;
+  
+  init() {
+    this._bucket = new cloud.Bucket();
+  }
+
+  /** 
+   * Adds a task to the task list.
+   * @returns The ID of the new task.
+   */
+  inflight add_task(title: str): str {
+    let id = uuid.v4();
+    let j = Json { 
+      title: title, 
+    };
+    print("adding task ${id} with data: ${j}"); //j should be printed out nicely 
+    this._bucket.put(id, j.to_str());
+    return id;
+  }
+
+  /** 
+   * Gets a task from the task list.
+   * @param id - the id of the task to return
+   * @returns the title of the task (optimistic)
+   */
+  inflight get_task(id: str): Json {
+    let blob = this._bucket.get(id);
+    return Json.parse(blob)
+  }
+
+  /** 
+   * Sets effort estimation on a test
+   * @param id - the id of the task to return
+   * @param effort_estimation - the time (duration) estimated for this task
+   * @returns The ID of the existing task.
+   */
+  inflight add_estimation(id: str, effort_estimation: duration) {
+    let j = this.get_task(id).to_mut();
+    j.set("effort_estimation", effort_estimation);
+    this._bucket.put(id, j.to_str());
+    return id;
+  }
+
+  /** 
+   * Removes a task from the list
+   * @param id - the id of the task to be removed
+   */
+  inflight remove_tasks(id: str): str {
+    print("removing task ${id}");
+    this._bucket.delete(id);
+    return id;
+  }
+
+   /** 
+    * Gets the tasks ids 
+    * @returns set of task id
+    */
+  inflight list_task_ids(): Set<str> {
+    let result = MutSet<str> {};
+    for id in this._bucket.list() {
+      result.add(id);
+    }
+    return result.to_immut();
+  }
+
+   /** 
+    * Find tasks with title that contains a term
+    * @param term - the term to search
+    * @returns set of task id that matches the term
+    */
+  inflight find_tasks_with(term: str): Array<str> {
+    print("find_tasks_with: ${term}");
+    let task_ids = this.list_task_ids();
+    print("found ${task_ids.size} tasks");
+    let output = MutArray<str>[];// #1172
+    for id in task_ids {
+      let j = this.get_task(id); 
+      let title = Str.from_json(j.get("title"));
+      if title.contains(term) { 
+        print("found task ${id} with title \"${title}\" with term \"${term}\"");
+        output.push(id);
+      }
+    }
+    
+    print("found ${output.len} tasks which match term '${term}'");
+    return output.to_immut();
+  }
+}
+
+// --------------------------------------------
+// testing
+// --------------------------------------------
+
+resource Test {
+  tasks: TaskList; 
+  _test: inflight (Test): void;
+
+  inflight before() {
+    for id in this.tasks.list_task_ids() {
+      this.tasks.remove_tasks(id);
+    }
+  }
+
+  init(name: str, test: inflight (Test): void) {
+    this.tasks = new TaskList();
+    this._test = test;
+    new cloud.Function({
+      this.before();
+      this._test(this);
+    }) as "test:${name}";
+  }
+}
+
+new Test("get and find task", inflight (t: Test) => {
+  t.tasks.add_task("clean the dishes");
+  let result = t.tasks.find_tasks_with("clean the dishes");
+  assert(result.len == 1);
+  let t = t.tasks.get_task(result.at(0));
+  assert("clean the dishes" == Str.from_json(t.get("title")));
+}) as "a1";
+
+new Test("get, remove and find task", inflight (t: Test) => {
+  t.tasks.add_task("clean the dishes");
+  t.tasks.add_task("buy dishwasher soap");
+  t.tasks.remove_tasks(tasks.find_tasks_with("clean the").at(0));
+  let result = t.tasks.find_tasks_with("clean the dish");
+  assert(result.len == 0);
+}) as "a2";
+
+new Test("effort_estimation", inflight (t: Test) => {
+  let id = t.tasks.add_task("clean the dishes");
+  let var j = t.tasks.get_task(id);
+  assert(!j.get("effort_estimation")); //  make sure effort estimation default nil
+  task.add_estimation(id, 2h);
+  j = t.tasks.get_task(id);
+  assert(2h == j.get("effort_estimation"));
+}) as "a3";
+
+```
--- a/docs/06-contributors/999-rfcs/000-stories/story-13.md
+++ b/docs/06-contributors/999-rfcs/000-stories/story-13.md
@@ -11,7 +11,7 @@ This version of it includes the following functional changes:
 Also, it includes some non-functional requirements: 
 - The Bucket should be replicated across different regions 
   - +100 if we are able to do the same for azure as well
-- Least privileged permissions are granted, to resources 
+- Least privileged permissions are granted
 - The code should work on sim, aws, azure
 - VScode should be able to autocomplete 
 - Console should be able to show all resources, and interact with them
@@ -161,9 +161,9 @@ new Test("effort_estimation", inflight (t: Test) => {
   let id = t.tasks.add_task("clean the dishes");
   let var j = t.tasks.get_task(id);
   assert(!j.get("effort_estimation")); //  make sure effort estimation default nil
-  task.add_estimation(id, 2h);
+  task.add_estimation(id, 4h);
   j = t.tasks.get_task(id);
-  assert(2h == j.get("effort_estimation"));
+  assert(4h == j.get("effort_estimation"));
 }) as "a3";
 
 ```


docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+- Least privileged permissions are granted
+- The code should work on sim, aws, azure
+- VScode should be able to autocomplete 
+- Console should be able to show all resources, and interact with them
+
+
+
+```ts (wing)
+bring cloud;
+// need to run: npm install uuid --save
```
eladb, line 23: I think we will need a jsii wrapper for uuid in order for this to work.

```suggestion
// need to run: npm install @winglibs/uuid --save
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+  init() {
+    this._bucket = new cloud.Bucket();
+  }
+
+  /** 
+   * Adds a task to the task list.
+   * @returns The ID of the new task.
+   */
+  inflight add_task(title: str): str {
+    let id = uuid.v4();
```
eladb, line 42: This would be something like a static method because we don't have free floating functions.

```suggestion
    let id = uuid.Uuid.v4();
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+    let j = Json { 
+      title: title, 
+    };
+    print("adding task ${id} with data: ${j}"); //j should be printed out nicely 
+    this._bucket.put(id, j.to_str());
```
eladb, line 47:
```suggestion
    this._bucket.put_json(id, j);
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+    return id;
+  }
+
+  /** 
+   * Gets a task from the task list.
+   * @param id - the id of the task to return
+   * @returns the title of the task (optimistic)
+   */
+  inflight get_task(id: str): Json {
+    let blob = this._bucket.get(id);
```
eladb, line 57:
```suggestion
    let blob = this._bucket.get_json(id);
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /** 
+   * Sets effort estimation on a test
+   * @param id - the id of the task to return
+   * @param effort_estimation - the time (duration) estimated for this task
+   * @returns The ID of the existing task.
+   */
+  inflight add_estimation(id: str, effort_estimation: duration) {
+    let j = this.get_task(id).to_mut();
```
eladb, line 68:
```suggestion
    let j = this.get_task(id).clone_mut();
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+    j.set("effort_estimation", effort_estimation);
+    this._bucket.put(id, j.to_str());
```
eladb, line 70:
```suggestion
    this._bucket.put_json(id, j);
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+   /** 
+    * Gets the tasks ids 
+    * @returns set of task id
+    */
+  inflight list_task_ids(): Set<str> {
+    let result = MutSet<str> {};
+    for id in this._bucket.list() {
+      result.add(id);
+    }
+    return result.to_immut();
```
eladb, line 93:
```suggestion
    return result.clone();
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+      let j = this.get_task(id); 
+      let title = Str.from_json(j.get("title"));
+      if title.contains(term) { 
+        print("found task ${id} with title \"${title}\" with term \"${term}\"");
+        output.push(id);
+      }
+    }
+    
+    print("found ${output.len} tasks which match term '${term}'");
+    return output.to_immut();
```
eladb, line 116:
```suggestion
    return output.clone();
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+  inflight before() {
+    for id in this.tasks.list_task_ids() {
+      this.tasks.remove_tasks(id);
+    }
+  }
+
+  init(name: str, test: inflight (Test): void) {
+    this.tasks = new TaskList();
+    this._test = test;
+    new cloud.Function({
```
eladb, line 137: This really feels like a (very nice) hack... let's think how we can make this a little nicer...

ekeren: We are still hacking this for now. It turns out that this hack will not work for this sprint... 

I'll have to change this for now (probably tomorrow when we know what is actually in)

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren: I find it counter intuitive that clone returns a different object type. 
I would expect reserving the `clone` for 🤷  cloning

ekeren: BTW, I also don't like the `to_immut()` API 😊

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /** 
+   * Gets a task from the task list.
+   * @param id - the id of the task to return
+   * @returns the title of the task (optimistic)
+   */
+  inflight get_task(id: str): Json {
+    let blob = this._bucket.get(id);
+    return Json.parse(blob)
```
staycoolcall911, line 58:
```suggestion
    return Json.parse(blob);
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /** 
+   * Sets effort estimation on a test
+   * @param id - the id of the task to return
+   * @param effort_estimation - the time (duration) estimated for this task
+   * @returns The ID of the existing task.
+   */
+  inflight add_estimation(id: str, effort_estimation: duration) {
```
staycoolcall911, line 67:
```suggestion
  inflight add_estimation(id: str, effort_estimation: duration): str {
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren: I don't expect `str` to be a subtype of `Json` 

```ts (wing)
print_json(j: Json) {
   if (j.some_bool_value) {
      print(j.some_str_value)
   }
}

print_json("I am not a json");
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: Life is tuff

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: What's the problem?

ekeren: I believe that there is a contractual difference between the two API:
1. `this._bucket.put(id, j.to_str());` - We are saving the string representation of a json
1. `this._bucket.put_json(id, j);` - We are saving a json (it doesn't say how it is saved)

Personally, I prefer the  number 1, if someone wants to use a Bucket for storing json's, he can create his own JsonBucketStore. If someone wants to create his own Bucket for storing parquet files or csv files he can creates his own abstractions on top of the regular bucket

ekeren: But think goes back to me believing that we should have a thin API for our SDK, which may contradicts with "always optimize for the common case" in some cases

docs/06-contributors/999-rfcs/000-stories/story-13.md
MarkMcCulloh: I'm a fan of clone or copy. Maybe copy is better because there's less implication of "deep" copy (which is not what we're doing). It's also 1 fewer letter :)

`result.copy()` and `result.copy_mut()` are nice because the effect is clear, the fact that the type changes doesn't feel as important to put in the name because it's already in the signature.

docs/06-contributors/999-rfcs/000-stories/story-13.md
MarkMcCulloh: is uuid stuff important enough to include in the stdlib?

ekeren:
for myself, need to add a definition and adjust the code for permissions

docs/06-contributors/999-rfcs/000-stories/story-13.md
Chriscbr: Ok, maybe I jumped the gun here -- so even though `x` in my example above could be a string-ish value, maybe it's not correct to say `Json "hi"` is the same as `"hi"`.

Eyal raised this question ("if string is a subtype of json, then is this allowed?") that stumped me...

```
"hello".some_bool_value
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+> **Status**: Expected released on 2023/02/16
+
+The following code is an iteration of Wing's task list.
+This version of it includes the following functional changes:
+- Using npm's uuid package inflight instead of the counter resource 
+- Storing Json instead of a string
+- Using duration for estimated effort
+
+Also, it includes some non-functional requirements: 
+- The Bucket should be replicated across different regions
```
hasanaburayyan, line 12: Should the story indicate that this is toggled featured rather than change in implementation of underlying resources?

eladb: I think this should be a plugin and the story should be explicit about how this plugin is applied.

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+- Storing Json instead of a string
+- Using duration for estimated effort
+
+Also, it includes some non-functional requirements: 
+- The Bucket should be replicated across different regions 
+  - +100 if we are able to do the same for azure as well
+- Least privileged permissions are granted
+- The code should work on sim, aws, azure
+- VScode should be able to autocomplete 
+- Console should be able to show all resources, and interact with them
```
eladb, line 17: Let's add "Windows support" (we want to mitigate the risk)

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+The following code is an iteration of Wing's task list.
+This version of it includes the following functional changes:
+- Using npm's uuid package inflight instead of the counter resource 
+- Storing Json instead of a string
+- Using duration for estimated effort
+
+Also, it includes some non-functional requirements: 
+- The Bucket should be replicated across different regions 
+  - +100 if we are able to do the same for azure as well
+- Least privileged permissions are granted
```
eladb, line 14: I think we should punt this to the next story

docs/06-contributors/999-rfcs/000-stories/story-13.md
```diff
@@ -0,0 +1,169 @@
...
+- The code should work on sim, aws, azure
+- VScode should be able to autocomplete
```
eladb, line 16: 👍 👍

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: I actually think it is important enough to include in the stdlib at some point, but for the sake of this story, @ekeren wanted some simple external library, and this is good enough I believe.

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: We don't have a design tenet to create a "thin API". I believe in a rich, "batteries included", SDK that will make people's lives as easy as possible. If we know there's a common case to put JSON objects into buckets, we should support `put_json()` and not make every person in the world implement it themselves.

> if someone wants to use a Bucket for storing json's, he can create his own JsonBucketStore

This is exactly what I want to avoid :-)

Also, think of `put_json()` simply as an overload (we don't have method overloads at the moment, so this is the best we can do).

We should definitely talk about this more generally from a design tenets perspective, and not in this sprint.

I think a good place to have the conversation about JSON is over https://github.com/winglang/wing/issues/1053 and maybe we should expand our SDK design tenets to cover this question about "minimal" versus "batteries included" API and how we offer "overloads" in our APIs (<-- @Chriscbr).

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: I don't think `bucket.put()` should always take a JSON because there are definitely use cases to store objects that are not JSON (e.g. binary blobs).

Maybe in the future we should define a "Serializable" interface ("trait") and have both `str` and `Json` (and `Blob`) implement it, and then `put()` just takes a `Serializable`.

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: We are having two separate discussions here:

1. Should we call this method `copy` or `clone`?
2. What should be the name of the method that returns an immutable copy and a mutable one?

The rationale of having `copy`/`clone` return an immutable version is because we want Wing to be ["immutable by default"](https://docs.winglang.io/blog/2023/02/02/good-cognitive-friction). So I am actually comfortable with the fact that this is counter intuitive (this is what I was calling "good cognitive friction").

I am less comfortable with `clone()` returning a *mutable* copy because this means that users can mistakingly use a mutable object and we want that to be a very explicit thing.

So in my view, the alternative to `copy()`+`copy_mut()` are `copy_immut()`+`copy_mut()` (that is, no `clone()`), which TBH, is quite ugly.

Bottom line, I still think `clone()` (or `copy()`) should return the immutable version, and if the user is "surprised" and has to explicitly request the mutable copy, I am happy about it.

(I am less opinionated about `clone` versus `copy` :-))

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: Why do we need this actually? Why not just a bunch of tests?

eladb (dismissed):

docs/06-contributors/999-rfcs/000-stories/story-13.md
Chriscbr: @eladb I totally agree - I don't think only accepting `Json` should be our endgame. My original comment was just trying to point out that if our choices for story 13 are between the following:

```
// API A
resource Bucket {
  ...
  inflight put(key: str, value: str) { ... }
  inflight put_json(key: str, value: Json) { ... }
}
```

```
// API B
resource Bucket {
  ...
  inflight put(key: str, value: Json) { ... }
}
```

Then API B might be cleaner and less confusing for the beta / pre-beta.

eladb: But then how would you store a string in the bucket?

`"hello".to_json()` returns `"hello"` (with quotes).

How about we do API A and rename `put()` to `put_str()`.

(basically `put_json()` and `put_str()` and `get_json()` and `get_str()`).

Chriscbr: Ahh, gotcha - `"hello"` can be parsed back into JSON while `hello` cannot. Not sure why I got so tripped up on this... yeah separate methods for these make sense to me now.

docs/06-contributors/999-rfcs/000-stories/story-13.md
staycoolcall911: This is the next thing @yoav-steignberg will be working on. In sprint 13 we're not expecting a full solution for "permissions", just some progress.

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren: Windows support is only for the wing cli. 
Once the cli works on windows,  @ainvoner will do a spike to see what is missing in the console

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren: Do we want to use the`@winglibs` namespace  for this?

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren:
```suggestion
    return result.copy_mut();
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren: yes, this will be discussed in further details on the actual ticket

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren:
```suggestion
    return output.copy();
```

docs/06-contributors/999-rfcs/000-stories/story-13.md
ekeren: moving them to simple `cloud.Function` tests

ekeren: done

docs/06-contributors/999-rfcs/000-stories/story-13.md
eladb: We don't have to. We can just publish something like wing-uuid for now

staycoolcall911 (approved):
@ekeren - I think we can move forward with this, so I'm approving. Let's see if we need to limit the scope for the next sprint tick and or create an additional stories for bring jsii + permisssions

PR #1395, opened 2023-01-31
feat(compiler): `string.contains()` and allowing small js macros in jsii metadata

### LINKED ISSUE #1243: `str` API should have the `contains()` method
Currently we support `.includes()` (direct delegation to javascript). We would like to "rename" this method to `.contains()` in wing.

MarkMcCulloh:
Notes:
- Changed Calls to have expressions instead of references
  - Ended up adding support for IIFEs, so I added a test for that as well
- Added a way to denote the JS that gets emitted when invoking a function definition
  - The builtin globals now uses this for their custom justification
  - This string has two builtin placeholders to use when emitting, `$self$` and `$args$`
    - To handle the string replacements properly (avoid over-eager replacement), added a nice crate https://crates.io/crates/aho-corasick
- Added `@macro` for docstring jsii metadata, which can be set for methods of a class to set the JS emit customization above (I'm open for name suggestions)

Fixes #1243


libs/wingc/src/ast.rs
```diff
@@ -409,16 +409,23 @@ impl BinaryOperator {
...
-	Identifier(Symbol),
+pub struct Reference {
+	pub kind: ReferenceKind,
+	pub evaluated_type: RefCell<Option<TypeRef>>,
+	pub span: WingSpan,
+}
+
+#[derive(Debug)]
+pub enum ReferenceKind {
+	Identifier { symbol: Symbol },
```
Chriscbr, line 420: This decoupling looks good, though I'm wondering doesn't this mean we can end up with structures like this?

```rust
Expr {
  kind: ExprKind::Reference(Reference {
    kind: ReferenceKind::Identifier { .. }
    evaluated_type: // A
  }),
  evaluated_type: // B
}
```

Is there is a reason we need the `evaluated_type` data in both places given that they will always have the same value, in this case anyways? (disclaimer, I don't fully grasp all of the functions `Reference` serves in the compiler)

(random thought - something about this reminds me a little bit of lvalues vs rvalues https://eli.thegreenplace.net/2011/12/15/understanding-lvalues-and-rvalues-in-c-and-c)

MarkMcCulloh: Yeah that situation is possible and in fact would probably happen often. It occurs any time an expression is just a reference, e.g. `x + a.b` where both `x` and `a.b` are such expressions.

The reason I added the additional evaluated_type is primarily for the case of call expressions like `a.b()`. In this case, the evaluated type of the whole expression is not the same as the Reference `a.b` and there is no way to find the type of the reference unless you're inside the TypeChecker.

There is definitely a smell of lvalues and rvalues here. References are our only lvalues currently so they are unique. `Expr` are our rvalues but they are allowed to wrap References. Not sure if it affects any changes here other than some reasoning why Reference can get some special treatment.

MarkMcCulloh:
@Chriscbr @yoav-steinberg Per offline discussion, updated this PR so that `Calls` use `Expressions` instead of `References`.

libs/wingc/src/capture.rs
```diff
@@ -178,40 +176,15 @@ fn scan_for_inflights_in_arglist(args: &ArgList, diagnostics: &mut Diagnostics)
...
 	args: &ArgList,
 	env: &SymbolEnv,
 	statement_idx: usize,
 	diagnostics: &mut Diagnostics,
 ) -> Vec<Capture> {
 	let mut res = vec![];
-	if let Reference::NestedIdentifier { object, property } = reference {
-		res.extend(scan_captures_in_expression(&object, env, statement_idx, diagnostics));
-
-		// If the expression evaluates to a resource we should check what method of the resource we're accessing
```
yoav-steinberg, line 191: I see you got rid of this. I guess it makes sense because it's not relevant to capturing, it's more of a type-checking thing and doesn't affect the output.

libs/wingc/src/capture.rs
```diff
@@ -312,22 +285,33 @@ fn scan_captures_in_expression(
...
+								.expect("Expected resource property to be a variable")
+								._type,
+							phase,
+						),
+						Err(type_error) => {
+							panic!("{}", type_error);
+						}
+					};
+
+					let func = prop_type.as_function_sig().unwrap();
```
yoav-steinberg, line 306: We're not in _call_ anymore, this is the generic scan captures function for _expressions_ so I don't think we can simply `unwrap` here and assume it'll be a function sig.
On the other hand I'm not sure why we even have all this "if the expression evaluates to a resource we should check what method of the resource we're accessing". This doesn't affect the otuput all it does is adds a debug print. Maybe worth keeping for the future, but you'll need to fix this to verify `prop_type` is a function sig.

libs/wingc/src/jsify.rs
```diff
@@ -352,13 +309,38 @@ impl JSifier {
...
+					let arg_string = self.jsify_arg_list(&args, None, None, needs_case_conversion, phase);
+					let expr_string = match &function.kind {
+						ExprKind::Reference(reference) => self.jsify_reference(reference, Some(needs_case_conversion), phase),
+						_ => format!("({})", self.jsify_expression(function, phase)),
+					};
+
+					if let Some(function_sig) = function_type.as_function_sig() {
+						if let Some(js_override) = &function_sig.js_override {
+							let self_string = &match &function.kind {
+								//
```
yoav-steinberg, line 322: missing comment?

libs/wingc/src/jsify.rs
```diff
@@ -352,13 +309,38 @@ impl JSifier {
...
+							let ac = AhoCorasick::new(patterns);
+							return ac.replace_all(js_override, replace_with);
+						}
+					}
+
+					format!("({}{}({}))", auto_await, expr_string, arg_string)
+				} else {
+					panic!(
+						"Unable to call expressions, missing type information: {}",
+						function.span
```
yoav-steinberg, line 341:
```suggestion
					panic!(
						"Unable to call expressions at {}, it's not a closure",
						function.span
```

libs/wingc/src/lib.rs
```diff
@@ -133,12 +136,15 @@ pub fn type_check(scope: &mut Scope, types: &mut Types) -> Diagnostics {
 	let env = SymbolEnv::new(None, types.void(), false, false, Phase::Preflight, 0);
 	scope.set_env(env);
 
+	// note: Globals are emitted here and wrapped in "{{ ... }}" blocks. Wrapping makes these emissions, actual
```
yoav-steinberg, line 139:
```suggestion
	// note: Globals are emitted here and wrapped in "{ ... }" blocks. Wrapping makes these emissions, actual
```

libs/wingc/src/jsify.rs
MarkMcCulloh: Technically this block would only happen if the `evaluated_type` for the expressions is empty. That being said, this panic does belong a couple of lines above where we do as_function_sig so I added it there. I also slightly improved this panic to look a bit more like your suggestion.

libs/wingc/src/capture.rs
MarkMcCulloh: Good catch, added logic to handle that. This was indeed failing in the case of something like `this._counter.add()`. Also made a comment noting that if we want to add some sort of capturing logic inside a reference while recursing it, this is probably where it'd go.

yoav-steinberg (approved):

Issue #1024, opened 2023-01-02
jsii importer fails when trying to import an interface


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

Change the signature of `cloud.Function` to take an `IFunctionHandler` instead of `Inflight`.

### I expected this:

To work.

### Instead, this happened:

I added some traces to the importer which show the JSII FQN:

```
Importing JSII type @winglang/wingsdk.cloud.BucketProps
Importing JSII type @winglang/wingsdk.cloud.BucketDeleteOptions
Importing JSII type @winglang/wingsdk.cloud.CounterProps
Importing JSII type @winglang/wingsdk.cloud.IFunctionHandler
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libs/wingc/src/type_check/jsii_importer.rs:100:36
stack backtrace:
wasm://wasm/0958d686:1
RuntimeError: unreachable
    at __rust_start_panic (wasm://wasm/0958d686:wasm-function[19777]:0x51111e)
    at rust_panic (wasm://wasm/0958d686:wasm-function[5571]:0x3eb2c7)
    at std::panicking::rust_panic_with_hook::hd67762ecc8bae640 (wasm://wasm/0958d686:wasm-function[828]:0x238631)
    at std::panicking::begin_panic_handler::{{closure}}::h4aed573334d346d1 (wasm://wasm/0958d686:wasm-function[6849]:0x41db21)
    at std::sys_common::backtrace::__rust_end_short_backtrace::h33c000da160739e7 (wasm://wasm/0958d686:wasm-function[19510]:0x5102bf)
    at rust_begin_unwind (wasm://wasm/0958d686:wasm-function[16491]:0x4f0c68)
    at core::panicking::panic_fmt::hddd782b3234e7a63 (wasm://wasm/0958d686:wasm-function[16802]:0x4f4f6a)
    at core::panicking::panic::h99048373a2fbc45b (wasm://wasm/0958d686:wasm-function[14508]:0x4d2c2a)
    at core::option::Option<T>::unwrap::hf23f2d4e877cd587 (wasm://wasm/0958d686:wasm-function[8732]:0x4578d6)
    at wingc::type_check::jsii_importer::JsiiImporter::lookup_or_create_type::h15d4ece3799f019b (wasm://wasm/0958d686:wasm-function[806]:0x232f3f)
[...]

Issue #397, opened 2022-10-31
Jsii variadic parameter support


Chriscbr:
Tracking issue to support variadic parameters from JSII APIs.

yoav-steinberg:
Probably depends on https://github.com/winglang/wing/issues/125

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.66](https://github.com/winglang/wing/releases/tag/v0.24.66).

yoav-steinberg:
Re opening since we still don't support import variadics from JSII: https://github.com/winglang/wing/blob/6ae93e3950b12b6c1b3881e4d347263574222c4e/libs/wingc/src/type_check/jsii_importer.rs#L455-L462

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.91](https://github.com/winglang/wing/releases/tag/v0.24.91).

Issue #107, opened 2022-09-20
Storage Modifiers (static)


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#14-storage-modifiers

staycoolcall911:
We have completed this in #1662 with some leftovers #1668 and #1669

Issue #125, opened 2022-09-20
Function Declaration: Variadic Arguments


staycoolcall911:
#### 3.6.3 Variadic Arguments
If the last argument of a function type is the `...args` keyword followed by an
`Array` type, then the function accepts typed variadic arguments. Expansion of
variadic arguments is not supported currently and the container of variadic
arguments is accessible with the `args` key like a normal array instance.
```TS
let f = (x: num, ...args: Array<num>) => {
  log(x + y + args.len);
}
// last arguments are expanded into their array
f(1, 2, 3, 4, 5, 6, 34..100);
```

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.66](https://github.com/winglang/wing/releases/tag/v0.24.66).

PR #1397, opened 2023-02-01
feat(compiler): bring jsii

### LINKED ISSUE #477: Bring JSII modules
Add the ability to use external jsii libraries in wing

## Description

As a wing programmer I would like to be able to do:
```sh
npm install --save  cdk-constants
```
and then use this jsii library, in the following manner:
```ts (wing)
bring "cdk-constants" as cdk_constants;

print(cdk_constants.ManagedPolicies.AWS_LAMBDA_BASIC_EXECUTION_ROLE);

let something = inflight (s: str): str => {
  print(cdk_constants.ManagedPolicies.AWS_LAMBDA_BASIC_EXECUTION_ROLE);
};
```

### Spec reference

Chriscbr:
Closes #477 

This PR adds the underpinnings of support for `bring`-ing JSII libraries. The following list of limitations is not comprehensive:

1. Our compiler doesn't support static class/resource members yet, so types with static members may not be imported properly. (#107)
2. Our compiler doesn't support interfaces, so many Terraform libraries currently fail to import. (#1024)
3. Our compiler doesn't support variadic parameters, so methods with such parameters are not available. (#397)
4. Our compiler doesn't support importing enums yet. (#1484)
5. Constructs named `App` cannot be imported (these constructs typically do not have a `scope` and `id` so we would need to model them differently in the compiler in order to make them usable) (#1485)

 The core of the changes in this PR are in the `jsii_importer`. The JSII importer now imports all types into nested namespaces inside a new symbol environment named `libraries`, based on the fully-qualified names (FQN) of types. For example, when the SDK is imported through `bring cloud;`, the jsii importer creates a symbol named `@winglang/sdk` in the global namespace that points to a `Namespace` struct. This namespace has its own symbol environment (a lookup table), and a symbol named `cloud` is added to it pointing to another `Namespace`. That namespace's symbol environment is then given symbols pointing to each type in the `cloud` namespace.

The motivation for this model is that no type will ever get imported multiple times (it's not possible for `cloud.Bucket` to get imported twice, even if you were theoretically able to do `bring cloud; bring sdk;`). I'm not sure if this will hold water forever (will Wing apps allow multiple copies of a library in its dependency closure...?) but for now it seemed like a helpful assumption.

Types are automatically inserted into the correct namespace -- so whenever types are needed from other namespaces (for example, `cloud.BucketBase` needs to reference it's base type, `core.Resource`), a new namespace is created ad-hoc in the correct place (in this example, as a sub-namespace of `@winglang/sdk`).

Finally, when all types are imported into appropriate namespaces, the JSII importer will create a symbol in the global symbol  environment named `cloud` which points to the `cloud` namespace within `@winglang/sdk` inside `libraries` . This is the basis of the `bring "my-lib" as my_whatever` import name assignment.

Other changes:
- Namespaces are now allocated in a vector so that we can construct multiple pointers to a single namespace.
- Clean up comments (the convention in Rust is that docstrings are commented with three slashes -- once we do this, the comments appears on mouse hover in IDEs like VS Code)
- StmtKind::Use is renamed StmtKind::Bring
- Several parts of the compiler have been updated to use &Path instead of &str wherever a path is being referenced
- Create a "FQN" type for organizing FQN-related operations.

 Future work:
 - https://github.com/winglang/wing/issues/1414
- https://github.com/winglang/wing/issues/1487
- https://github.com/winglang/wing/issues/1488
- https://github.com/winglang/wing/issues/1515
- https://github.com/winglang/wing/issues/1516
- https://github.com/winglang/wing/issues/1517


ekeren:
@Chriscbr , 

Currently variadic parameters are not inside the beta, do you think this is an issue?

examples/tests/valid/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
+        "jsii-code-samples": "1.7.0"
```
MarkMcCulloh, line 3: Should we create an issue to manage a package like this in the monorepo? It should be small so the benefits of avoiding external deps and network lookups for tests would outweigh the cost of maintaining it.

libs/wingc/src/lib.rs
```diff
@@ -201,15 +210,26 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
...
 }
 
-pub fn compile(source_file: &str, out_dir: Option<&str>) -> Result<CompilerOutput, Diagnostics> {
+pub fn compile(source_path: &Path, out_dir: Option<&Path>) -> Result<CompilerOutput, Diagnostics> {
+	assert!(source_path.is_file(), "Source path must be a file");
+	assert!(
+		source_path.extension().unwrap() == "w",
+		"Source file must have .w extension",
+	);
+	assert!(source_path.is_absolute(), "Source path must be absolute");
```
MarkMcCulloh, line 219: This is exposing a very easily fixed bug in the playground :)

https://github.com/winglang/wing/blob/main/apps/wing-playground/worker.js#L48

In this PR could you update this to be "/code.w" instead of "code.w"?
Otherwise the assertion here fails.

apps/wing-language-server/src/prep.rs
```diff
@@ -31,7 +31,7 @@ pub fn parse_text(source_file: &str, text: &[u8]) -> ParseResult {
 	let mut scope = wing_parser.wingit(&tree.root_node());
 
 	let mut types = type_check::Types::new();
-	let type_diag = type_check(&mut scope, &mut types);
+	let type_diag = type_check(&mut scope, &mut types, &Path::new(source_file));
```
MarkMcCulloh, line 34: This path will not be absolute, so this will fail with the newly added is_absolute checks. 
Technically this doesn't even have to be a file path, it can be an arbitrary uri.

libs/wingc/src/type_check.rs
```diff
@@ -540,32 +560,37 @@ impl Types {
 	}
 
 	pub fn number(&self) -> TypeRef {
-		self.get_typeref(self.numeric_idx)
+		self.get_type_ref(self.numeric_idx)
```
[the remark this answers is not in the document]
Chriscbr: My mental model was that "ref to a Type" is what a "TypeRef" is, but maybe I'm confused

examples/tests/valid/package.json
Chriscbr: Sure, will do

apps/wing-language-server/src/prep.rs
Chriscbr: Got it. So, we need an absolute source file path is so that the compiler can locate JSII libraries in the file system and load their .jsii files with wingii etc. I'm sort of bundling two use cases here (one is to use the file name for error messages and diagnostics, the other is to look at the parent directories), but in either case I think it might simplify our design if we say that in our wingc-as-a-library, `compile` must be invoked with an absolute file path.

@MarkMcCulloh Is there a way we can convert the URI to a file path or just decide to throw if it can't be converted into one? (after all, I don't think we should accept arbitrary URIs like `tel:+1-555-555-5555` as input to our compiler)

MarkMcCulloh: https://github.com/winglang/wing/blob/main/apps/wing-language-server/src/main.rs#L202

params.uri has a `to_file_path()` that should be good, so you could change that line to something like:

```rs
let parse_result = parse_text(
	params
		.uri
		.to_file_path()
		.expect("Only 'file' scheme supported")
		.to_str()
		.unwrap(),
	&params.text.as_bytes(),
);
```

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
MarkMcCulloh: Our TypeRef is a raw pointer, so it is a kind of ref but I guess it's more particular than that.
Whether or not that should dictate the name of this method, idk I could go either way.

libs/wingc/src/type_check/fqn.rs
```diff
@@ -0,0 +1,161 @@
...
+	/// - `["ns1"]`
+	/// - `["ns1", "ns2"]`
+	///
+	/// But it is not part of these filters:
+	/// - `["ns1", "ns2", "ns3"]`
+	/// - `["ns2"]`
+	pub fn is_in_namespace<T: AsRef<str>>(&self, namespace_filter: &[T]) -> bool {
+		self.namespaces().count() >= namespace_filter.len()
+			&& self
+				.namespaces()
```
yoav-steinberg, line 92: Note that calling `namespaces()` here multiple times is inefficient. It might hint that we should store the FQN as a vector and not call `spilt` each time we want some info out of it. Not critical at this point, but just saying...

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -290,6 +312,67 @@ impl SymbolEnv {
 		Ok(t)
 	}
 
+	/// Pass `include_hidden: true` if it's OK to return types that have only been imported implicitly (such as through an inheritance chain), and false otherwise
```
yoav-steinberg, line 315: Maybe a TODO about the code duplication here with `lookup_nested` or perhaps just find a way to remove the code duplication.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -137,6 +144,10 @@ impl<'a> JsiiImporter<'a> {
 					"map" => self.wing_types.add_type(Type::Map(wing_type)),
 					_ => panic!("Unsupported collection kind '{}'", collection_kind),
 				}
+			} else if let Some(Value::Object(_)) = obj.get("union") {
+				// Wing does not support union types, so we'll model it the same way as if
+				// we saw an "any" in a JSII library
```
yoav-steinberg, line 149: Do we have a mechanism to emit compiler warnings/info?
```suggestion
				// Currently Wing does not support union types (see https://github.com/winglang/wing/issues/977), 
				// so we'll model it the same way as if we saw an "any" in a JSII library.
```

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -354,6 +411,11 @@ impl<'a> JsiiImporter<'a> {
 				arg_types.push(wing_type);
 				// Define the rest of the arguments and create the method signature
 				if let Some(params) = &m.parameters {
+					if self.has_variadic_parameters(params) {
+						debug!("Skipping method {} with variadic parameters", m.name);
```
yoav-steinberg, line 415: Tracking issue https://github.com/winglang/wing/issues/397 or https://github.com/winglang/wing/issues/125.
Emit compiler wanrinig?

libs/wingc/src/lib.rs
```diff
@@ -267,36 +287,35 @@ pub fn compile(source_file: &str, out_dir: Option<&str>) -> Result<CompilerOutpu
...
+	};
+
+	fn get_wing_files<P>(dir: P) -> impl Iterator<Item = PathBuf>
+	where
+		P: AsRef<Path>,
+	{
+		fs::read_dir(dir)
+			.unwrap()
+			.map(|entry| entry.unwrap().path())
+			.filter(|path| path.extension().map(|ext| ext == "w").unwrap_or(false))
```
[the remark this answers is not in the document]
staycoolcall911: Can we add a test?

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
Chriscbr: Reverted

libs/wingc/src/type_check/symbol_env.rs
Chriscbr: My rust-fu isn't good enough... a TODO comment will have to do

libs/wingc/src/type_check/jsii_importer.rs
Chriscbr: https://github.com/winglang/wing/issues/1475

libs/wingc/src/type_check/jsii_importer.rs
Chriscbr: Added links to issues

libs/wingc/src/type_check/fqn.rs
Chriscbr: It's a fair point. I think for now this is cheaper in the respect that most places in the jsii importer would need to do a clone if FQN took ownership of the string, but I can see the argument for both sides... seems like a classic memory/speed tradeoff. :-)

libs/wingc/src/type_check/fqn.rs
```diff
@@ -0,0 +1,161 @@
...
+impl<'a> From<&'a str> for FQN<'a> {
+	fn from(fqn: &'a str) -> Self {
+		if fqn.split('.').count() < 2 {
+			panic!("Invalid FQN: {}", fqn);
+		}
+		FQN(fqn)
+	}
+}
+
+impl<'a> From<&'a String> for FQN<'a> {
```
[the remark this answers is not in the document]
Chriscbr: Not sure, I got hit several obstacles trying to de-duplicate these. This feels like it could be a Rust learning opportunity, so I've opened a StackOverflow question: https://stackoverflow.com/questions/75378250/how-to-reuse-implementation-of-fromstr-and-fromstring-in-rust

libs/wingc/src/lib.rs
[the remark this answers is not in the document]
Chriscbr: @yoav-steinberg No need to wonder, I can just add a check for `path.is_file()` here 😄 

@staycoolcall911 This function is just used as a helper for finding test files

libs/wingc/src/type_check/fqn.rs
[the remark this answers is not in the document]
Chriscbr: Ok, yeah you were right - I've just removed it and called `.as_str()` to convert where needed

libs/wingc/src/lib.rs
```diff
@@ -210,15 +219,26 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
...
 }
 
-pub fn compile(source_file: &str, out_dir: Option<&str>) -> Result<CompilerOutput, Diagnostics> {
+pub fn compile(source_path: &Path, out_dir: Option<&Path>) -> Result<CompilerOutput, Diagnostics> {
+	assert!(source_path.is_file(), "Source path must be a file");
+	assert!(
+		source_path.extension().unwrap() == "w",
+		"Source file must have .w extension",
+	);
+	assert!(source_path.is_absolute(), "Source path must be absolute");
```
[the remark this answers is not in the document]
Chriscbr: I tried removing this requirement, but using relative paths seem to cause bugs with the crate we are using for node module resolution ([link](https://github.com/goto-bus-stop/node-resolve)). This is the algorithm we depend on for locating jsii and npm libraries. The main issue is that the implementation happens to use the `Path::canonicalize` function to normalize paths, which does not work when compiled to wasm. I found a workaround, but said workaround only works with absolute paths (i.e. paths like "../../examples/tests/valid/bring_jsii.w" won't work). So, this seemed like a better tradeoff than re-implementing the node resolution algorithm ourselves.

(another third option could be to suggest a fix in the upstream library `canonicalize`)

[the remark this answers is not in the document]
MarkMcCulloh: A very fancy option is that we can actually "import" a module resolution function via extern ([like here in the lsp](https://github.com/winglang/wing/blob/f9d5efe48af59b60f0f2e597dfaddb0b24d4565f/libs/wingc/src/lsp/mod.rs#L46)). That way module resolution can be offloaded to the runtime consuming wingc.

libs/wingc/src/lib.rs
```diff
@@ -116,13 +120,18 @@ pub fn parse(source_file: &str) -> (Scope, Diagnostics) {
...
 		None => {
-			panic!("Failed parsing source file: {}", source_file);
+			panic!("Failed parsing source file: {}", source_path.display());
 		}
 	};
 
 	let wing_parser = Parser {
 		source: &source[..],
-		source_name: source_file.to_string(),
+		source_name: source_path
```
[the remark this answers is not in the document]
Chriscbr: Updated

libs/wingc/src/lib.rs
```diff
@@ -132,7 +141,7 @@ pub fn parse(source_file: &str) -> (Scope, Diagnostics) {
 	(scope, wing_parser.diagnostics.into_inner())
 }
 
-pub fn type_check(scope: &mut Scope, types: &mut Types) -> Diagnostics {
+pub fn type_check(scope: &mut Scope, types: &mut Types, source_path: &Path) -> Diagnostics {
```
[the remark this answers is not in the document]
Chriscbr: Added a comment

libs/wingc/src/lib.rs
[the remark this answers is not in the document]
Chriscbr: Actually I've decided to leave for the future (the parser is only using the immediate file name, not the full path, so it would just add a lot of unnecessary clones at the moment). I think this should be done in a separate refactoring which adds a `Parser::new` method etc.

libs/wingc/src/type_check.rs
```diff
@@ -1740,25 +1823,48 @@ impl<'a> TypeChecker<'a> {
...
+			let assembly_name = wingii_types
+				.load(manifest_root.as_str(), Some(wingii_loader_options))
+				.unwrap();
+			(wingii_types, assembly_name)
+		} else {
+			let mut wingii_types = wingii::type_system::TypeSystem::new();
+			let wingii_loader_options = wingii::type_system::AssemblyLoadOptions { root: true, deps: true };
+			let source_dir = self.source_path.parent().unwrap().to_str().unwrap();
+			let assembly_name = wingii_types
+				.load_dep(library_name.as_str(), source_dir, &wingii_loader_options)
```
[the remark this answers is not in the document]
Chriscbr: Added a comment. TL;DR it's because the SDK we are loading from a specific location (provided by the WINGSDK_MANIFEST_ROOT environment variable) whereas other modules we need to search for from the user's project root directory. One is doing a lookup and one is not, hence the different arguments.

There might be better ways to clean this up but I can't quite envision what the better API would be -- definitely something we can iterate on

libs/wingc/src/type_check.rs
```diff
@@ -554,9 +555,13 @@ impl Types {
 		types.push(Box::new(Type::Void));
 		let void_idx = types.len() - 1;
 
+		let void_ref = UnsafeRef::<Type>(&*types[void_idx] as *const Type);
+		let libraries = SymbolEnv::new(None, void_ref, false, false, Phase::Preflight, 0);
```
yoav-steinberg, line 559: This "hack" for setting the void return value of the env kind of shows why this doesn't fit perfectly into a `SymbolEnv`.

It might be wroth attempting your suggestion of turning our `SymbolEnv` into a trait. And then adding the extra info (`return_type`, `is_init`, `is_class`, 'statement_idx`) to concrete implementations. Another challenge here is to see if we can model different enums of `SymbolKind` for each concrete implementation. 

I can currently think of the following concrete types/implementations of a `SymbolEnv` in our code:
* `CodeEnv`: A `SymbolEnv` associated with a `Scope` - has `return_type`, `is_init`, `statement_idx` fields. Contains Types, Vars, Namespaces.
* `ClassEnv`: A `SymbolEnv` associated with a class/resource/struct type (maybe enums too). Contains Vars only.
* `LibraryEnv`: A `SymbolEnv` associated with imported libraries. Contains Types and Namespaces only.

Because this seems complex I'm fine with what you did here, just add a comment that you can't call `self.void()` from withing the static `new`, and this is just a hack to create the top level env for mapping lib names to Symbols.

yoav-steinberg (approved):

libs/wingc/src/type_check.rs
```diff
@@ -1412,21 +1459,59 @@ impl<'a> TypeChecker<'a> {
...
+				let namespace_filter: Vec<String>;
+				// alias is the symbol we are giving to the imported library or namespace
+				let alias: &Symbol;
+
+				if module_name.name.starts_with('"') && module_name.name.ends_with('"') {
+					// case 1: bring "library_name" as identifier;
+					if identifier.is_none() {
+						self.stmt_error(
+							stmt,
+							format!("bring \"{}\" must be assigned to an identifier", module_name.name),
```
eladb, line 1480:
```suggestion
							format!("bring \"{}\" must be assigned to an identifier (e.g. bring \"foo\" as foo)", module_name.name),
```

libs/wingc/src/type_check.rs
```diff
@@ -1412,21 +1459,59 @@ impl<'a> TypeChecker<'a> {
...
+						// we use the module name as the identifier.
+						// For example, `bring fs` will import the `fs` namespace from @winglang/sdk and assign it
+						// to an identifier named `fs`.
+						WINGSDK_CLOUD_MODULE | WINGSDK_FS_MODULE => {
+							library_name = WINGSDK_ASSEMBLY_NAME.to_string();
+							namespace_filter = vec![module_name.name.clone()];
+							alias = identifier.as_ref().unwrap_or(&module_name);
+						}
+						WINGSDK_STD_MODULE => {
+							self.stmt_error(stmt, format!("Redundant import of \"{}\"", WINGSDK_STD_MODULE));
```
eladb, line 1504:
```suggestion
							self.stmt_error(stmt, format!("Redundant bring of \"{}\"", WINGSDK_STD_MODULE));
```

libs/wingc/src/type_check.rs
```diff
@@ -1412,21 +1459,59 @@ impl<'a> TypeChecker<'a> {
...
+							return;
+						}
+						_ => {
+							self.stmt_error(stmt, format!("\"{}\" is not a built-in module", module_name.name));
```
eladb, line 1508: Eventually this syntax is also going to be used to bring 3rd party `@winglibs` but that's for later...

libs/wingc/src/type_check.rs
```diff
@@ -1740,25 +1825,50 @@ impl<'a> TypeChecker<'a> {
...
-	fn add_module_to_env(&mut self, env: &mut SymbolEnv, module_name: String, statement_idx: usize) {
-		// TODO Hack: treat "cloud" or "std" as "_ in wingsdk" until I figure out the path issue
-		if module_name == "cloud" || module_name == "fs" || module_name == WINGSDK_STD_MODULE {
-			let mut wingii_types = wingii::type_system::TypeSystem::new();
+	fn add_module_to_env(
+		&mut self,
+		env: &mut SymbolEnv,
+		library_name: String,
+		namespace_filter: Vec<String>,
+		identifier: &Symbol,
```
eladb, line 1833: Semantic inconsistency -- `identifier` was called `alias` before (I think):



```suggestion
		alias: &Symbol,
```

libs/wingc/src/type_check.rs
```diff
@@ -1740,25 +1825,50 @@ impl<'a> TypeChecker<'a> {
...
+		statement_idx: usize,
+	) {
+		let mut wingii_types = wingii::type_system::TypeSystem::new();
```
eladb, line 1836: Why is this `mut`? Are we reassigning `wingii_types` somewhere?

libs/wingc/src/type_check.rs
```diff
@@ -1740,25 +1825,50 @@ impl<'a> TypeChecker<'a> {
...
+
+		// Loading the SDK is handled different from loading any other jsii modules because with the SDK we provide an exact
+		// location to locate the SDK, whereas for the other modules we need to search for them from the source directory.
+		let assembly_name = if library_name == WINGSDK_ASSEMBLY_NAME {
+			// in runtime, if "WINGSDK_MANIFEST_ROOT" env var is set, read it. otherwise set to "../wingsdk" for dev
```
eladb, line 1841: > otherwise set to "../wingsdk" for dev

This smells like a potential security issue. Usually dev behavior is something you want to be explicit, not implicit.

(but that's for a future fix, maybe open an issue).

libs/wingc/src/type_check/fqn.rs
```diff
@@ -0,0 +1,152 @@
+use std::fmt::{Display, Formatter};
+
+use crate::{CONSTRUCT_BASE, WINGSDK_ASSEMBLY_NAME, WINGSDK_RESOURCE};
+
+/// Represents a fully-qualified name (FQN) of a type in a JSII library.
+/// For example, `@aws-cdk/aws-ec2.Vpc` is a FQN.
+/// The FQN uniquely identifies a type.
```
eladb, line 7:
```suggestion
/// The FQN uniquely identifies a type within the JSII ecosystem.
```

libs/wingc/src/type_check/fqn.rs
```diff
@@ -0,0 +1,152 @@
...
+		let mut parts = self.0.split('.');
+		parts.next();
+		parts.next_back().unwrap();
+		parts
+	}
+
+	/// Returns true if the FQN represents a "construct base class".
+	///
+	/// TODO: this is a temporary hack until we support interfaces.
+	pub fn is_construct_base(&self) -> bool {
```
eladb, line 62: Feels weird that this is under `FQN`... Makes more sense as a free-floating function to me.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -148,107 +159,160 @@ impl<'a> JsiiImporter<'a> {
...
-			.unwrap()
-			.to_string()
-	}
+		self.setup_namespaces_for(&type_fqn);
 
-	fn import_type(&mut self, type_fqn: &str) {
-		// Hack: if the class name is a construct base then we treat this class as a resource and don't need to define it
-		if Self::is_fqn_construct_base(type_fqn) {
+		// Check if this is a JSII interface and import it if it is
+		let jsii_interface = self.jsii_types.find_interface(type_fqn.as_str());
```
eladb, line 191: Shouldn't `find_interface()` and `find_class()` take an `FQN`?

```suggestion
		let jsii_interface = self.jsii_types.find_interface(type_fqn);
```

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -558,7 +638,8 @@ impl<'a> JsiiImporter<'a> {
 			if let Some(params) = &initializer.parameters {
 				for (i, param) in params.iter().enumerate() {
 					// If this is a resource then skip scope and id arguments
-					if is_resource {
+					// TODO hack - skip this check if the resource's name is "App"
```
eladb, line 641: Why BTW? What's wrong with a resource named `App`?

If this is for the SDK `App` resource, then we could just add `@hidden` attribute to it or something, no?

eladb (commented):
Amazing work!

libs/wingc/src/lib.rs
```diff
@@ -258,36 +278,35 @@ pub fn compile(source_file: &str, out_dir: Option<&str>) -> Result<CompilerOutpu
...
-		}
-		files
+	use std::{
+		fs,
+		path::{Path, PathBuf},
+	};
+
+	fn get_wing_files<P>(dir: P) -> impl Iterator<Item = PathBuf>
+	where
+		P: AsRef<Path>,
```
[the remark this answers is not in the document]
Chriscbr: Here I was just out some Rust patterns -- in this case using the type parameter makes the method more flexible (accepts more kinds of input) by accepting any type that can be converted into a &Path. Based on the Rust docs this includes a lot of types like String, Path, PathBuf, and so on: 
Screenshot 2023-02-13 at 11 52 44 AM

.github/workflows/build.yml
```diff
@@ -206,6 +206,18 @@ jobs:
           working-directory: tools/hangar
           install-command: npm ci --ignore-scripts
 
+      - name: Install Example Dependencies (1/2)
```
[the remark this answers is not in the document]
Chriscbr: The reason I added it here is that the E2E test job is not installing all of the dependencies in the repo, only the ones for the hangar. We could install all dependencies, but my reasoning at the time was that it would extend the build times by up to a minute so I'd keep it scoped for now.

libs/wingc/src/type_check/jsii_importer.rs
Chriscbr: These are methods from `wingii` -- at some point it probably makes sense to use this `FQN` struct there and just import / use that type in `wingc`, but I decided against doing that work it in this PR because that library leverages code generation so it might be nontrivial.

libs/wingc/src/type_check/jsii_importer.rs
Chriscbr: https://github.com/winglang/wing/issues/1485

(I'll add a link to the issue in the comment)

Chriscbr: This is about importing other `App` constructs, like `cdktf.App` or `cdk8s.App`

libs/wingc/src/type_check/fqn.rs
Chriscbr: 👍

libs/wingc/src/type_check.rs
Chriscbr: Added

libs/wingc/src/type_check.rs
Chriscbr: Updated

libs/wingc/src/type_check.rs
Chriscbr: This needs to be `mut` because we are calling a mutable method on the struct (wingii_types.load()`) - in Rust, `mut` doesn't just refer to reassignment, but any kind of mutation to the variable or its contents.

eladb: Cool, thanks!

eladb (approved):

libs/wingc/src/type_check.rs
Chriscbr: https://github.com/winglang/wing/issues/1518

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -2,126 +2,131 @@
...
 `;
 
 exports[`capture_mutables.w > wing test (--target sim) - invalid > stderr 1`] = `
 "Compilation failed with 5 error(s)
-Error at ../../../../examples/tests/invalid/capture_mutables.w:8:10 | Cannot reference 'a' of type 'MutArray<str>' from an inflight context
-Error at ../../../../examples/tests/invalid/capture_mutables.w:10:10 | Cannot reference 's' of type 'MutSet<num>' from an inflight context
-Error at ../../../../examples/tests/invalid/capture_mutables.w:12:10 | Cannot reference 'm' of type 'MutMap<bool>' from an inflight context
-Error at ../../../../examples/tests/invalid/capture_mutables.w:14:10 | Cannot reference 'a_cloned' of type 'MutArray<str>' from an inflight context
-Error at ../../../../examples/tests/invalid/capture_mutables.w:14:19 | Unknown symbol \\"size\\""
+Error at capture_mutables.w:8:10 | Cannot reference 'a' of type 'MutArray<str>' from an inflight context
```
Chriscbr, line 20: The changes to the path names in this diff is because I changed the Parser so it's initialized in `lib.rs` with the source file's `file_name` and not the entire (relative or absolute) path. Definitely something we can change back if it becomes confusing 🙂

PR #1415, opened 2023-02-01
fix(compiler): jsii fails to import primitive collections

MarkMcCulloh:

### DIFF
--- a/libs/wingc/src/type_check/jsii_importer.rs
+++ b/libs/wingc/src/type_check/jsii_importer.rs
@@ -120,7 +120,7 @@ impl<'a> JsiiImporter<'a> {
 				// TODO: Handle non-primitive collections
 				let primitive_type = element_type
 					.get("primitive")
-					.expect("'primitive' is required for collection element types")
+					.expect("non-primitive collection types are not yet supported")
 					.as_str()
 					.expect("'primitive' must be a string");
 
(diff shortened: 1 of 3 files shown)


libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -104,9 +104,38 @@ impl<'a> JsiiImporter<'a> {
...
+				let element_type = d
+					.get("elementtype")
+					.expect("'elementtype' is required for collection types")
+					.as_object()
+					.expect("'elementtype' must be an object");
+
+				// TODO: Handle non-primitive collections
+				let primitive_type = element_type
+					.get("primitive")
+					.expect("'primitive' is required for collection element types")
```
Chriscbr, line 123: Should the error message be something like "non-primitive collection types not yet supported"?

MarkMcCulloh: done, good callout

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -104,9 +104,38 @@ impl<'a> JsiiImporter<'a> {
...
+					.expect("'elementtype' must be an object");
+
+				// TODO: Handle non-primitive collections
+				let primitive_type = element_type
+					.get("primitive")
+					.expect("non-primitive collection types are not yet supported")
+					.as_str()
+					.expect("'primitive' must be a string");
+
+				let wing_type = match primitive_type {
```
eladb, line 127: Can we add some tests to cover all types of collections?

Issue #287, opened 2022-10-18
Proposal: implicit scope for any function


eladb:
I am wondering if the notion of implicit scopes, which we already have for resource initializers, can be extended to methods/functions as well. 

So basically any method/functions/constructor that has a first argument called `scope` of type `Construct` is called without explicitly specifying the scope, and can be customised with "in SCOPE" syntax of course.

This comes up quite a lot, especially in static factory methods. It means that any code that runs in wing always have access to the parent scope that called it. Something that we struggled with in the CDK quite a lot.

The follow up question is what happens in Wing code. If I want to access `scope` from any method/function, does it mean the compiler will implicitly include an additional first parameter to the function signature. That's a breaking change... 

Perhaps all wing functions should always have an implicit scope as the first argument? Don't kill me!

Chriscbr:
A possible use case for this would be for the SDK's `Schedule`/`Timer`/`Cron` resource:

```ts
let timer = cloud.Timer.fromCron("* * * * 6"); // implicit scope and id
timer.on_tick(inflight () => {
  // ...
});
```

eladb:
yes, I think the "factory" use case is strong. I was thinking that we need to give factories a more first-class status in the language. In a sense an initializer and a static factory method are sort of the same thing.

Chriscbr:
Here's a random idea. What if whenever a function in a JSII/Wing library (instance method, static method, constructor, etc.) has its first two arguments named "scope" and "id", then the Wing compiler will automatically remove them from the signature and automatically fill in the scope and id during jsification?

(I think this would need to be fleshed out because of edge cases, e.g. does this apply to both classes and resources? Are there any places in Wing where there isn't a valid implicit scope?)

eladb:
@Chriscbr yes I like this a lot.

marciocadev:
Keep please

PR #1416, opened 2023-02-01
feat(sdk): add schedule resource

### LINKED ISSUE #1286: Tracking issue: work left todo to implement the SDK Spec
Since we finalized the [SDK Spec](https://docs.winglang.io/reference/wingsdk-spec) we want to update the [SDK Roadmap](https://github.com/orgs/winglang/projects/3/views/1).

### P1
- [ ] #1289
- [ ] #1290
- [ ] #1309
- [ ] #1310

### P2
- [ ] #1293
- [ ] #1294
- [ ] #1297
- [ ] #1298
- [ ] #1301
- [ ] #1302
- [ ] #1305
- [ ] #1306
- [ ] #1313
- [ ] #1314
- [ ] #1317
- [ ] #1321
- [ ] #1325
- [ ] #1329
- [ ] #1333
- [ ] #1337
- [ ] #1341
- [ ] #1291
- [ ] #1311

### P3
- [ ] #1345
- [ ] #1349
- [ ] #1353
- [ ] #1216
- [ ] #1292
- [ ] #1312
- [ ] #1295
- [ ] #1296
- [ ] #1299
- [ ] #1300
- [ ] #1303
- [ ] #1304
- [ ] #1307
- [ ] #1308
- [ ] #1315
- [ ] #1316

### Good first issues
- [ ] #1378
- [ ] #1379
- [ ] #1380
- [ ] #1381
- [ ] #1382
- [ ] #1383
- [ ] #1384
- [ ] #1385
- [ ] #1386
- [ ] #1387
- [ ] #1388
- [ ] #1389
- [ ] #1390
- [ ] #1375

### LINKED ISSUE #1290: cloud.Schedule - AWS implementation
https://docs.winglang.io/reference/wingsdk-spec#schedule

marciocadev:
~~WORK IN PROGRESS~~

I'm having some problems implementing Schedule, in the documentation it should be created from a static method but I don't know how to capture the `scope` this way, it may be necessary to change the compiler but I don't think it's recommended

It's a work in progress, but I can already create a schedule in AWS, but the creation of the schedule happens in the `on_tick` event because the EventBridge Scheduler can only be created with a defined target (function)

Closes #1290


revitalbarletz:
@marciocadev this is great, I see that you have some questions/problem with how to capture the scope. I will reach out through slack.

hasanaburayyan:
@marciocadev thanks so much for the patience 🙏 I went ahead and created a new issue to support static methods in the compiler. https://github.com/winglang/wing/issues/1456 

For this issue lets move forward with completing schedule resource using the same api experience as other resources. Making cron and rate both optional, but 1 required. (Which is what you have been working on 😃)
```typescript
let schedule = new cloud.Schedule({cron: "* * * * *"});
let otherSchedule = new cloud.Schedule({rate: 1h});

schedule.on_tick(inflight () => {
  // stuff
});
```

libs/wingsdk/test/target-tf-aws/schedule.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    "aws_iam_role", // role for function
+    "aws_iam_role_policy", // policy for role
+    "aws_iam_role_policy_attachment", // execution policy for role
+    "aws_lambda_function", // processor function
+    "aws_s3_bucket", // S3 bucket for code
+    "aws_s3_object", // S3 object for code
+    "aws_scheduler_schedule", // main schedule
+  ]);
+  expect(tfSanitize(output)).toMatchSnapshot();
+  expect(treeJsonOf(app.outdir)).toMatchSnapshot();
```
hasanaburayyan, line 33: Let's add a check in these tests to ensure the scheduler was created with a rate or cron. 

```typescript
expect(
    cdktf.Testing.toHaveResourceWithProperties(output, "aws_scheduler_schedule", {
      schedule_expression: "rate(2 minutes)",
    })
  ).toEqual(true);
```

libs/wingsdk/test/target-tf-aws/schedule.test.ts
```diff
@@ -0,0 +1,85 @@
...
```
hasanaburayyan, line 61: Same as above, except for cron

libs/wingsdk/test/target-tf-aws/schedule.test.ts
```diff
@@ -0,0 +1,85 @@
+import * as cloud from "../../src/cloud";
+import * as std from "../../src/std";
+import * as tfaws from "../../src/target-tf-aws";
+import { Testing } from "../../src/testing";
+import { mkdtemp } from "../../src/util";
+import { tfResourcesOf, tfSanitize, treeJsonOf } from "../util";
+
```
hasanaburayyan, line 7: Another case worth testing is the expected behavior with multiple schedules, or 1 schedule to multiple functions? Add some assertions about expectations in regards to iam resources maybe. 

We have been trying to get better about corner case coverage in our tests 😅 slowly but surely

libs/wingsdk/src/cloud/schedule.ts
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Trigger events at a periodic rate.
+   * @example 1m
+   * @default undefined
+   */
+  readonly rate?: Duration;
+
+  /**
+   * Trigger events according to a cron schedule.
+   * @example "0/1 * ? * * *"
```
hasanaburayyan, line 25: I think we should add a link to the AWS schedule cron syntax since its not standard. Thoughts?
https://docs.aws.amazon.com/scheduler/latest/UserGuide/schedule-types.html#cron-based

hasanaburayyan (changes requested):

hasanaburayyan:
@marciocadev As usual awesome work 🚀 ! I gave the code a once over and added a few comments nothing major. Let me know what you think.

libs/wingsdk/src/cloud/schedule.ts
marciocadev: Yes, I agree, it is worth putting this information

libs/wingsdk/test/target-tf-aws/schedule.test.ts
marciocadev: Nice catch, 1 schedule with multiple roles gave an error due to role id and policy id but I added node.id and hash and now it works

libs/wingsdk/src/cloud/schedule.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Global identifier for `Schedule`.
+ */
+export const SCHEDULE_TYPE = "wingsdk.cloud.Schedule";
+
+/**
+ * Properties for `Schedule`.
+ */
+export interface ScheduleProps {
```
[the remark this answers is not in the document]
marciocadev: Nice, with #287 I can easily change this to work with static "from" methods

libs/wingsdk/src/cloud/schedule.ts
hasanaburayyan: hmm yea I was not even thinking about this, it looks like even the major cloud providers dont agree on a standard cron format.
AWS:   { <minute> <hour> <day-of-month> <month> <day-of-week> <year> }
Azure: { <second> <minute> <hour> <day-of-month> <month> <day-of-week> }
GCP:   { <minute> <hour> <day-of-month> <month> <day-of-week> }

So basically AWS gives <year> and Azure gives <second>. @Chriscbr correct me if Im wrong but it looks like our spec uses the standard unix format (just like GCP). 

Since we dont want the cloud detail leaking to the user I think we should just offer the 5 fields (as it shows in the spec) and maybe update the SDK spec to clarify valid cron syntax. 

Then bear with me as the coffee needs to still kick in 😄 I also think maybe we might need to update the string depending on the target. 

@Chriscbr what about offering experience similar to CDK { minute: '30', hour: '2' }
https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-events.CronOptions.html

hasanaburayyan: Cool, so @marciocadev lets change the references to use the 5 field format and for the AWS implementation we may just need to append a wildcard for year field. I imagine the API fails if you pass in only 5 fields versus the expected 6 but fact check me there 😁

libs/wingsdk/src/cloud/schedule.ts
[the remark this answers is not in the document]
marciocadev: I think we can 'validate' the five value string by counting the spaces between them (4 blanks) at the moment and adding * at the end when deploying to AWS or * at the beginning when deploying to Azure.

libs/wingsdk/src/cloud/schedule.ts
marciocadev: I think we can 'validate' the five value string by counting the spaces between them (4 blanks) at the moment and adding * at the end when deploying to AWS or * at the beginning when deploying to Azure.

libs/wingsdk/src/target-tf-aws/schedule.ts
```diff
@@ -0,0 +1,134 @@
...
+        "cron only accepts five values (minute, hour, day of month, month, day of week)."
+      );
+    }
+
+    /*
+     * The schedule cron format consists of 5 values (minutes, hour, day of month,
+     * month, and day of week), but the AWS EventBridge Schedule uses a 6-string
+     * format adding the year at the bottom of the string.
+     *
+     * That way, we add * at the bottom.
```
hasanaburayyan, line 44:
```suggestion
     * We append * to the cron string for year field.
```

hasanaburayyan (commented):
🔥 this is gonna be awesome! I add a few nits around wording in some comments let me know what you think.

I love the test coverage by the way!

marciocadev:
> 🔥 this is gonna be awesome! I add a few nits around wording in some comments let me know what you think.
> 
> I love the test coverage by the way!

Thanks @hasanaburayyan, and thank you about the comments, it's better now

libs/wingsdk/src/cloud/schedule.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Functionality shared between all `Schedule` implementations.
+ */
+export abstract class ScheduleBase extends Resource {
+  public readonly stateful = true;
+  constructor(scope: Construct, id: string, props: ScheduleProps = {}) {
+    super(scope, id);
+
+    this.display.title = "Schedule";
+    this.display.description = "A cloud schedule";
```
Chriscbr, line 41: Can we be more descriptive - what does a cloud schedule do?

libs/wingsdk/src/target-tf-aws/schedule.ontick.inflight.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export class ScheduleOnTickHandlerClient
+  implements IScheduleOnTickHandlerClient
+{
+  private readonly handler: IFunctionHandlerClient;
+  constructor({ handler }: { handler: IFunctionHandlerClient }) {
+    this.handler = handler;
+  }
+  public async handle(event: any) {
+    await this.handler.handle(event.detail);
```
Chriscbr, line 14: How do we know event has a `detail` field? We should annotate `event` with an interface like `event: AwsScheduleEvent` for documentation

libs/wingsdk/src/target-tf-aws/schedule.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    if (rate && cron) {
+      throw new Error("rate and cron cannot be configured simultaneously.");
+    }
+    if (!rate && !cron) {
+      throw new Error("rate or cron need to be filled.");
+    }
+    if (rate && rate.seconds < 60) {
+      throw new Error("rate can not be set to less than 1 minute.");
+    }
```
Chriscbr, line 32: Can we create an issue about supporting more frequent rates? I think this should be possible with some extra machinery -- see for example https://constructs.dev/packages/cdk-lambda-subminute/v/2.0.109?lang=typescript or https://constructs.dev/packages/paws-lambda-time-trigger/v/1.0.2?lang=typescript

Chriscbr (approved):
Looks great @marciocadev!

staycoolcall911:
Wow @marciocadev , this is incredible!

hasanaburayyan (approved):

libs/wingsdk/src/target-tf-aws/schedule.ontick.inflight.ts
marciocadev: I followed the way the topic was implemented

How about importing @types/aws-lambda into the project and using `ScheduleEvent`?

This library already has all the interfaces of the AWS resources and could be used to replace `any` in other implementations as well.

libs/wingsdk/API.md
```diff
@@ -890,6 +890,181 @@ with a fresh copy without any consequences.
 ---
 
 
+### Schedule <a name="Schedule" id="@winglang/sdk.cloud.Schedule"></a>
```
staycoolcall911, line 893: Need to also update /docs/04-reference/wingsdk-api.md.
You don't need to do these changes manually, of course. Just run `npm run build` from the libs/wingsdk folder and all your code changes will be reflected in the generated docs API.md and copied to wingsdk-api.md

libs/wingsdk/src/target-tf-aws/schedule.ts
marciocadev: Done issue #1551

libs/wingsdk/src/target-tf-aws/schedule.ontick.inflight.ts
Chriscbr: Yes that's a good idea 👍 

I think something like `import type { ... } from 'aws-lambda'` works

marciocadev: @Chriscbr `ScheduleOnTickHandlerClient` implements `IScheduleOnTickHandlerClient`, and the interface has `handle(message: string): Promise<void>;`

I can't change the event type without changing the interface that will be use for all future Schedule implementations

libs/wingsdk/API.md
marciocadev: done

libs/wingsdk/src/target-tf-aws/schedule.ontick.inflight.ts
Chriscbr: I just took another look at the spec, the current API we're looking for is:
```js
on_tick(fn: inflight () => void, opts: ScheduleOnTickProps?): void;
```

So actually we don't need to pass anything to the handler - I think this line should just be
```ts
await this.handler.handle();
```
and IScheduleOnTickHandlerClient should be updated accordingly. Let me know if that works

marciocadev: The `handler` object came from `IFunctionHandlerClient`
It needs an `event: string` as input, actually I believe it should receive an `event: any` or an `event: { [key:string]: any } ` because each resource sends an event with different parameters

```ts
export interface IFunctionHandlerClient {
  /**
   * Entrypoint function that will be called when the cloud function is invoked.
   * @inflight
   */
  handle(event: string): Promise<void>;
}
```

Chriscbr: @marciocadev Thanks for explaining. For now let's update IScheduleOnTickHandlerClient so `handle` accepts zero arguments, and in the implementation (ScheduleOnTickHandlerClient) you can simply invoke `this.handler.handle` with an empty string. You're right, we should allow `IFunctionHandlerClient` to support more types like JSON etc. but I think we should leave that change out of this PR since affects a larger surface area of the SDK

Issue #1365, opened 2023-01-30
Custom Wing targets and resource injection using plugins


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Remember all these Wing targets you used to love? The ones behind the `--target` switch of `wing`?

Don't worry, they are not going anyway!

Not only they are not going away, now, anyone can create and share targets through [Wing Plugins](#1218). In fact, you can completely customize how resources are created in Wing and inject an alternative instance.

So how does this work?

Plugins can now implement a `newResource` hook:

```ts
newResource(type: string, scope: Construct, id: string, ...args: any[]): Construct | undefined;
```

This hook gets called whenever a new resources is defined (`new Foo()`) in preflight code. The hook can use the resource fully qualified wing `type`, construct `scope`, user-defined `id` and the list of arguments passed to the initializer (constructor) to produce a new instance of this resource. 

(if `undefined` is returned, the next plugin can give this an attempt, or fall through to an error).

So, for example, here is how the `tf-aws` target is implemented:

```js
const wing = require('@winglang/sdk`);
const tfaws = wing.tfaws;
const cloud = wing.cloud;

exports.newResource = function(type, scope, id, ...args) {
  switch (type) {
    case cloud.BUCKET_TYPE:   return new tfaws.Bucket(scope, id, args[0]);
    case cloud.FUNCTION_TYPE: return new tfaws.Function(scope, id, args[0], args[1]);
    case cloud.QUEUE_TYPE:    return new tfaws.Queue(scope, id, args[0]);
    case cloud.LOGGER_TYPE:   return new tfaws.Logger(scope, id);
    case cloud.COUNTER_TYPE:  return new tfaws.Counter(scope, id, args[0]);
    case cloud.TOPIC_TYPE:    return new tfaws.Topic(scope, id, args[0]);
  }

  return undefined;
};
```

This means that, for example, specifying a `--target tf-gcp` is the same as adding `--plugin "@winglang/sdk.targets.tfgcp"` as the last switch.

```sh
wing compile --target tf-gcp
# same as:
wing compile --plugin "@winglang/sdk.targets.tfgcp"
```

Since returning `undefined` indicates "pass", you can use this to implement cool things like `dropbox-bucket.js`:

```js
const { DropBoxBucket } = require('@winglang/dropbox');

exports.newResource = (type, scope, id, ...args) {
[...]

PR #1418, opened 2023-02-02
feat: customize infrastructure with plugins

hasanaburayyan:
Looking for some feedback on the POC code I threw together for wing plugins. Allows passing in a list of plugins with preSynth, postSynth, and validate hooks.


apps/wing/src/index.ts
```diff
@@ -49,6 +49,10 @@ async function main() {
       .choices(["tf-aws", "tf-azure", "tf-gcp", "sim"])
       .makeOptionMandatory()
     )
+    .option(
+      "-p, --plugins [plugin...]",
+      "JS plugin to use for compilation"
```
eladb, line 54:
```suggestion
      "Compiler plugins"
```

apps/wing/src/index.ts
```diff
@@ -49,6 +49,10 @@ async function main() {
...
+    )
     .action(compile);
```
eladb, line 56: Please add a doc topic about plugins under docs/

libs/wingc/src/jsify.rs
```diff
@@ -34,8 +34,10 @@ function __app(target) {
 	}
 }
 const $App = __app(process.env.WING_TARGET);
+const $Plugins = plugins;
```
eladb, line 37: What's the use of this global? Can we avoid it?

libs/wingsdk/src/core/app.ts
```diff
@@ -49,6 +50,12 @@ export interface AppProps {
    * @default - use the default polycon factory included in the Wing SDK
    */
   readonly customFactory?: IPolyconFactory;
+
+  /**
+   * Plugins to apply on app.
```
eladb, line 55: Describe what are the strings here? (They are paths to JavaScript files). Are they relative/absolute. If relative, relative to what?

(The safest option is to require that paths are always absolute at this layer and have the cli resolve any relative paths against cwd)

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
+import fs from "fs";
+import { resolve } from "path";
+import vm from "vm";
+import * as cdktf from "cdktf";
+import { IApp } from "./app";
+
+/**
+ * Behavioral interface of a plugin hook.
+ */
+export interface IHook {
```
eladb, line 10:
```suggestion
export interface IHooks {
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+  /** preSynth phase */
+  preSynth(app: IApp): void;
```
eladb, line 12:
```suggestion
  preSynth(app: IConstruct): void;
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Plugin manager is responsible for loading hooks from plugins and
+ * managing their executions during the various phases of the plugin lifecycle.
+ */
+export class PluginManager {
+  private hooks: IHook[];
+  private context?: vm.Context;
+
+  constructor(plugins: string[], context?: vm.Context) {
```
eladb, line 27: Add doc strings for all parameters.

You might want to consider putting the `context` inside a props.

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  /**
+   * Add plugins to the plugin manager based on the plugin type.
+   *
+   * @param plugin the plugin to add
+   */
+  public add(plugin: string) {
+    // maybe we support other plugin types in the future (e.g. npm modules)
```
eladb, line 42: Prefer "early bailout":

```js
if (!ends with js) {
  throw error;
}

// happy path continues here
// ...
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  /**
+   * Loads hooks from a plugin file and adds them to the list of hooks
+   * that will later be run in the various plugin phases.
+   *
+   * @param filePath path to the plugin file
+   */
+  public addPluginFromFilePath(filePath: string) {
```
eladb, line 58: Why public?

```suggestion
  private addPluginFromFilePath(filePath: string) {
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+    const hooks: IHook = {} as any;
```
eladb, line 59: Avoid `as any`. It hides bugs.

In this case, if you didn't use it it would have forced you to validate the `exports` from the plug-in before you assign it into an `IHook` object (Good Thing).

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+    const context =
+      this.context ??
+      vm.createContext({ require, console, exports: hooks, cdktf, process });
```
eladb, line 62: cdktf?

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+   * after synthesis has occurred then overwrite the config file on disk
+   * with the modified one.
+   *
+   * @param config Terraform config to be passed to the postSyth hooks
+   * @param synthesizedStackPath path to the synthesized stack json file
+   */
+  public postSynth(config: any, synthesizedStackPath: string): any {
+    for (const hook of this.hooks) {
+      if (hook.postSynth) {
+        config = hook.postSynth(config);
```
eladb, line 98: I think if postsynth returns undefined, let's skip it

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * This phase is immutable, allowing plugins to perform validations
+   * on the Terraform config after all modifications have been made.
+   *
+   * @param config Terraform config to be passed to the validate hooks
+   */
+  public validate(config: any) {
+    for (const hook of this.hooks) {
+      if (hook.validate) {
+        hook.validate(config);
```
eladb, line 120: From an error reporting perspective, I think we need to track the plug-in name for each hook and if a hook throws an exception, we should wrap it with additional diagnostics such as the name of the plugin.

libs/wingsdk/test/core/plugin-manager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+const PLUGIN_CODE = `
+var s3_bucket = require("@cdktf/provider-aws/lib/s3-bucket");
+
+exports.preSynth = function(app) {
+  new s3_bucket.S3Bucket(app, "PluginBucket", {bucket: "plugin-bucket"})  
+};
+
+exports.postSynth = function(config) {
+  //TODO: Not sure why I needed to do a JSON.parse here
```
eladb, line 17: This feels like something we need to figure out because that's not the desired behavior

libs/wingsdk/test/core/plugin-manager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  fs.writeFileSync(pluginFile, PLUGIN_CODE);
+
+  // WHEN
+  const app = new tfaws.App({ outdir: tmpDir }); // exclude plugins from constructor for control
+  const pm = new PluginManager([pluginFile]);
+  new tfaws.Bucket(app, "Bucket", {});
+
+  const synthOutput = app.synth();
+  // TODO: Maybe find a better way to get the path to the synthesized stack
+  const synthesizedStackPath = `${app.outdir}/cdktf.out/stacks/root/cdk.tf.json`;
```
eladb, line 86: @Chriscbr can help

eladb (changes requested):
Great stuff here! Can't wait to play with this. 

Let's add some initial documentation in the docsite as part of this PR (use content and examples from your RFC)

Maybe the best place for it is the CLI reference docs https://docs.winglang.io/reference/cli

libs/wingsdk/test/core/plugin-manager.test.ts
Chriscbr: 👍  After [this PR](https://github.com/winglang/wing/pull/1435) the terraform file should just be at the root of `app.outdir`

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,124 @@
+import fs from "fs";
```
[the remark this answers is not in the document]
hasanaburayyan: Yea I agree, I was actually pondering this over dinner tonight. Some way for a plugin to advertise its compatibility, then the plugin manager to check maybe. Ill give some ideas a whirl this weekend.

hasanaburayyan:
Adding `do-not-merge` label this I have a some enhancements and refactors I want to apply but thought it be worth getting early thoughts before going down any routes further

libs/wingc/src/jsify.rs
hasanaburayyan: I could pass the plugins as an env var string (then split it in preflight.js). Curious, because I may be missing it, what the downside of leaving as is would be. Just name collisions?

libs/wingsdk/src/core/plugin-manager.ts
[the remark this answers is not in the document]
eladb: P2

libs/wingsdk/src/core/plugin-manager.ts
eladb: ICompilationHook?

Chriscbr: 👍

libs/wingsdk/src/core/plugin-manager.ts
hasanaburayyan: Yea good idea, I added a test case for this as well

docs/04-reference/wingcli.md
```diff
@@ -133,6 +133,16 @@ JavaScript bundles that include inflight code that executes on compute platform
...
 
+### Plugins
+
+Additionally the wing `compile` command can be provided an optional list of plugins to use during the compilation process.
+
+```sh
+$ wing compile --target tf-aws ENTRYPOINT.w --plugins PLUGIN1 PLUGIN2
+```
+The plugins can be absolute paths or relative paths (to the current working directory) to javascript files. For more the structure 
+or how to create a plugin see [Compiler Plugins](/reference/compiler-plugins).
```
Chriscbr, line 144:
```suggestion
The plugins can be absolute paths or relative paths (to the current working directory) to JavaScript files. For more on the structure 
on how to create a plugin, see [Compiler Plugins](/reference/compiler-plugins).
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+A single plugin can implement any number of stages. However, it is recommended to only implement the stages that are needed.
+
+## Compilation Stages
+
+Each compilation stage is meant to serve a specific purpose. The following is a list of the stages and their purpose:
+
+### `preSynth` stage
+
+This stage is called before the Wing begins to synthesize. In the context of a cdktf target, this stage will have access to the root node of the construct tree. Allowing for the plugin to add or manipulate constructs before the synthesis process begins.
```
Chriscbr, line 24:
```suggestion
This stage is called before the Wing begins to synthesize. In the context of a terraform-based target like `tf-aws`, this stage will have access to the root node of the construct tree. This allows the plugin to add or change [CDK for Terraform](https://github.com/hashicorp/terraform-cdk) constructs before the tree is synthesized into Terraform configuration.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+The following example adds bucket to root node, note this plugin is intended to be used with the `tf-aws` target.
```
Chriscbr, line 26:
```suggestion
The following example adds a bucket to the root node (note that this plugin is intended to be used with the `tf-aws` target).
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+  // app is the root node of the construct tree
+  new s3_bucket.S3Bucket(app, "MyPluginBucket", {
+    bucket: "my-plugin-bucket",
+  });
+}
+```
+
+### `postSynth` stage
+
+This stage is called after the Wing has finished synthesizing. In the context of a cdktf target, this stage will have access to the raw terraform JSON configuration. Allowing for manipulation of the JSON that is written to the compiled output directory.
```
Chriscbr, line 41: On naming, I wonder if it makes sense to say that the `postSynth` **stage** is when all `postSynth` **hooks** (or **compiler hooks**) are called by the compiler. IMO it's not the stage that gets called, but the hook. Might be super pedantic though 😅

```suggestion
### `postSynth` stage

This stage runs after Wing has finished synthesizing artifacts. When compiling to a terraform-based target like `tf-aws`, the hook will have access to the raw terraform JSON configuration, allowing for manipulation of the JSON that is written to the compiled output directory.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+This stage is useful for adding customizations that can not be applied in the context of the preSynth phase. Its worth noting that this is not meant as a validation phase since this stage is mutable by subsequent plugins.
+
+The following example manipulates the terraform configuration to use a s3 backend. For brevity, the example uses hard coded values. In practice these would be configurable.
```
Chriscbr, line 45:
```suggestion
The following example manipulates the terraform configuration to use a S3 backend. For brevity, the example uses hard coded values. In practice these would be configurable.
```

Chriscbr: "In practice these would be configurable" like through an environment variable, or something else? (just checking if I understand)

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+      region: "us-east-1",
+    }  
+  }
+  return config;
+}
+```
+
+### `validate` stage
+
+This stage is called right after the `postSynth` stage and provided the same context object. In the context of a cdktf target, this is the same terraform JSON configuration. However, this stage is immutable which allows for validation of the configuration without worrying about another plugin mutating after the fact.
```
Chriscbr, line 62:
```suggestion
This stage is called right after the `postSynth` stage and provided the same context object. In the context of a terraform-based target like `tf-aws`, this is the same Terraform JSON configuration. However, this stage does not allow configuration to be mutated, which allows plugins to validate the configuration without worrying about another plugin mutating after the fact.
```

libs/wingsdk/src/core/app.ts
```diff
@@ -58,6 +59,12 @@ export interface AppProps {
    * @default - use the default polycon factory included in the Wing SDK
    */
   readonly customFactory?: IPolyconFactory;
+
+  /**
+   * Absolute paths to plugin javascript files.
```
Chriscbr, line 64:
```suggestion
   * Absolute paths to plugin JavaScript files.
```

libs/wingsdk/test/core/plugin-manager.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  // THEN
+  expect(tfResourcesOfCount(output, "aws_s3_bucket")).toEqual(1);
+  expect(
+    cdktf.Testing.toHaveResourceWithProperties(output, "aws_s3_bucket", {
+      bucket: "plugin-bucket",
+    })
+  ).toEqual(true);
+});
+
+test("postSynth can modify terraform config", () => {
```
Chriscbr, line 76: Maybe add a test that `validate` isn't allowed to modify terraform config?

libs/wingsdk/test/core/plugin-manager.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  const pluginFile = join(tmpDir, "plugin.js");
+  fs.writeFileSync(pluginFile, PLUGIN_CODE);
+
+  // WHEN
+  const app = new tfaws.App({ outdir: tmpDir }); // exclude plugins from constructor for control
+  const pm = new PluginManager([pluginFile]);
+  new tfaws.Bucket(app, "Bucket", {});
+
+  const synthOutput = app.synth();
+  const synthesizedStackPath = `${app.outdir}/main.tf.json`;
```
Chriscbr, line 88: How about we make this path a constant on `CdktfApp`?
```suggestion
  const synthesizedStackPath = app.terraformManifestPath;
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,190 @@
...
+        require,
+        console,
+        exports: pluginHooks,
+        process,
+        __dirname: pluginDir,
+      });
+
+    const pluginCode = readFileSync(resolve(filePath), "utf8");
+    const script = new vm.Script(pluginCode);
+    script.runInNewContext(context);
```
Chriscbr, line 96: I'm not sure but I think `runInContext(context)` might work here since `context` has already been contextified.

docs/04-reference/winglang-plugins.md
hasanaburayyan: me word no good 😄 I agree it makes sense it has been updated

docs/04-reference/winglang-plugins.md
hasanaburayyan: Yea I think that configuration in this first iteration of plugins is up to the plugin's creator. So in most cases I would imagine an environment variable, but in the RFC one of the potentials is like a class that can be instantiated from your own plugin with props passed in. Maybe even a config file exists. I added some of those as possibilities for clarity maybe.

libs/wingsdk/src/core/plugin-manager.ts
hasanaburayyan: Yea I kept it for a while incase a custom context was passed then the context does not get contextified. I think that 
 allowing a custom context to be used adds complexity where I dont have a use case for it now. So for now I removed the ability to provide a context and changed this back to just `runInContext`

docs/04-reference/wingcli.md
```diff
@@ -133,6 +133,16 @@ JavaScript bundles that include inflight code that executes on compute platform
 
 You can deploy your stack to GCP using Terraform ([example](/getting-started/aws)).
 
+### Plugins
+
+Additionally the wing `compile` command can be provided an optional list of plugins to use during the compilation process.
```
eladb, line 138:
```suggestion
Additionally the `compile` command can be provided an optional list of plugins to use during the compilation process.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Compiler Plugins
+id: compiler-plugins
+description: Wing Compiler Plugins Reference
+keywords: [Wing reference, Wing language, language, Wing language spec, Wing programming language, compiler plugins]
+---
+
+Wing compiler plugins are used to extend the Wing compiler. They can be used to add non-functional customizations to the compilation's resulting infrastructure. For example, you can use a compiler plugin to enforce a tagging convention on all resources in the resulting infrastructure.
```
eladb, line 8: wrap to 100 width please ([rewrap] is amazing). Long lines make it harder to provide PR comments.

[rewrap]: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

eladb, line 8:
```suggestion
Wing compiler plugins are used to extend the Wing compiler. They can be used to customize the compilation output such as infrastructure definitions. For example, you can use a compiler plugin to enforce a tagging convention on all resources in the resulting infrastructure.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Plugin Design
+
+A compiler plugin is a JavaScript file that exports multiple functions. These functions are called by the Wing compiler at various points during compilation also referred to as stages.
```
eladb, line 12:
```suggestion
A compiler plugin is a JavaScript file that exports multiple functions (called "hooks"). These hooks are called by the compiler at various points during compilation.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+At each stage the compiler plugin is provided with a context object that allows the plugin to hook into the compilation process and apply customizations.
+
+A single plugin can implement any number of stages. However, it is recommended to only implement the stages that are needed.
+
+## Compilation Stages
```
eladb, line 18: I would use the term "hooks" instead of "stages". Hooks is more general purpose and doesn't necessarily imply sequential ordering. For example, we want to add a resource factory hook (#1365), and those are not "stages".

"The compiler offers hooks that can be used to customize its behavior at various points during the compilation."

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+Each compilation stage is meant to serve a specific purpose. The following is a list of the stages and their purpose:
+
+### `preSynth` stage
```
eladb, line 22:
```suggestion
### `preSynth` hook
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+This stage is called before the Wing begins to synthesize. In the context of a terraform-based target like `tf-aws`, this stage will have access to the root node of the construct tree. This allows the plugin to add or change [CDK for Terraform](https://github.com/hashicorp/terraform-cdk) constructs before the tree is synthesized.
```
eladb, line 24:
```suggestion
This stage is called before the compiler begins to synthesize. In the context of a Terraform-based target like `tf-aws`, this stage will have access to the root node of the construct tree. This allows the plugin to add or change [CDK for Terraform](https://github.com/hashicorp/terraform-cdk) constructs before the tree is synthesized.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+
+The following example adds a bucket to the root node (note this plugin is intended to be used with the `tf-aws` target).
```
eladb, line 26: > note this plugin is intended to be used with the `tf-aws` target.

Feels like we should offer some way for a plugin to validate that it is used within the right context. Let's open an issue to track this need?

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+At each stage the compiler plugin is provided with a context object that allows the plugin to hook into the compilation process and apply customizations.
+
+A single plugin can implement any number of stages. However, it is recommended to only implement the stages that are needed.
+
+## Compilation Stages
+
+Each compilation stage is meant to serve a specific purpose. The following is a list of the stages and their purpose:
+
+### `preSynth` stage
+
```
eladb, line 23: For each hook, please add API reference (you can use TypeScript notation)

```js
preSynth(app: Construct): void;
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+  // app is the root node of the construct tree
+  new s3_bucket.S3Bucket(app, "MyPluginBucket", {
+    bucket: "my-plugin-bucket",
+  });
+}
+```
+
+### `postSynth` stage
+
+This stage runs after the Wing has finished synthesizing artifacts. When compiling to a terraform-based target like `tf-aws`, the hook will have access to the raw terraform JSON configuration, allowing for manipulation of the JSON that is written to the compiled output directory.
```
eladb, line 41:
```suggestion
This stage runs after artifacts where synthesized. When compiling to a Terraform-based target like `tf-aws`, the hook will have access to the raw Terraform JSON configuration, allowing for manipulation of the JSON that is written to the compiled output directory.
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,76 @@
...
+      region: "us-east-1",
+    }  
+  }
+  return config;
+}
+```
+
+### `validate` stage
+
+This stage is called right after the `postSynth` stage and provided the same context object. In the context of a terraform-based target like `tf-aws`, this is the same Terraform JSON configuration. However, does not allow configuration to be mutated, which allows plugins to validate the configuration without worrying about another plugin mutating after the fact.
```
eladb, line 62: Please fix "terraform" to "Terraform" across the board.

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+  pluginName: string;
+  /** preSynth phase */
+  preSynth?(app: IConstruct): void;
+  /** postSynth phase */
+  postSynth?(config: any): any;
+  /** validate phase */
+  validate?(config: any): void;
+}
+
+enum PluginPhases {
```
eladb, line 24:
```suggestion
enum CompilationPhases {
```

libs/wingsdk/src/core/app.ts
```diff
@@ -91,9 +103,12 @@ export class CdktfApp extends Construct implements IApp {
 
     super(cdktfStack, "Default");
 
+    this.pluginManager = new PluginManager(props.plugins ?? []);
```
eladb, line 106: I am not super happy with the fact that this is only for `CdktfApp`... We need to do some architectural refactoring at the app level at some point to generalize this a bit.

(not for now).

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+      this.add(plugin);
+    }
+  }
+
+  /**
+   * Add plugins to the plugin manager based on the plugin type.
+   *
+   * @param plugin the plugin to add
+   */
+  public add(plugin: string) {
```
eladb, line 50:
```suggestion
  public add(pluginAbsolutePath: string) {
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+    // maybe we support other plugin types in the future (e.g. npm modules)
+    if (!plugin.endsWith(".js")) {
+      throw new Error(
+        `Currently only javascript files are supported as plugins. Got: ${plugin}`
+      );
+    }
+    this.addPluginFromFilePath(plugin);
```
eladb, line 57: Why do we need both `add()` and `addPluginFromFilePath()`?

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Loads hooks from a plugin file and adds them to the list of hooks
+   * that will later be run in the various plugin phases.
+   *
+   * @param filePath absolute file path to the plugin
+   */
+  private addPluginFromFilePath(filePath: string) {
+    const pluginHooks: ICompilationHook = {
+      pluginName: filePath, // can be overwritten by plugin (ex: exports.pluginName = "my-plugin")
```
eladb, line 68: This is not documented

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Behavioral interface of a plugin hook.
+ */
+export interface ICompilationHook {
+  /**
+   *  name of plugin can be set by plugin as export
+   *
+   * @default - absolute path to plugin file
+   */
+  pluginName: string;
```
eladb, line 15:
```suggestion
  name: string;
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  /**
+   * Loads hooks from a plugin file and adds them to the list of hooks
+   * that will later be run in the various plugin phases.
+   *
+   * @param filePath absolute file path to the plugin
+   */
+  private addPluginFromFilePath(filePath: string) {
+    const pluginHooks: ICompilationHook = {
```
eladb, line 67:
```suggestion
    const hooks: ICompilationHook = {
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+   * Call all preSynth hooks
+   *
+   * This phase is mutable, allowing plugins to modify the app construct
+   * before any synthesis takes place.
+   *
+   * @param app app construct to be passed to the preSynth hooks
+   */
+  public preSynth(app: IConstruct) {
+    for (const hook of this.hooks) {
+      if (hook.preSynth) {
```
eladb, line 108: Early bailout:

```js
if (!hook.preSynth) { continue; }
// happy flow...
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+    });
+
+    const pluginCode = readFileSync(resolve(filePath), "utf8");
+    const script = new vm.Script(pluginCode);
+    script.runInContext(context);
+
+    this.hooks.push(pluginHooks);
+  }
+
+  private throwHookInvocationError(
```
eladb, line 88: Move privates to the end of the class

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+          config = hook.postSynth(config) ?? config; // ignores undefined return values
+        } catch (err) {
+          this.throwHookInvocationError(
+            PluginPhases.POST_SYNTH,
+            hook.pluginName,
+            err
+          );
+        }
+      }
+    }
```
eladb, line 145: Can we reuse this pattern instead of copy&pasting for every hook?

Something like:

```js
return this.callAllHooks(
  PluginPhases.POST_SYNTH, 
  hook => hook.postSynth(config) ?? config
);
```

eladb (changes requested):
Nice!

docs/04-reference/winglang-plugins.md
hasanaburayyan: Oh my! Rewrap is amazing, my markdown life is forever changed 😁

docs/04-reference/winglang-plugins.md
hasanaburayyan: Yea I have an idea on how we can do this, Ill add it as a suggestion to the issue.

libs/wingsdk/src/core/plugin-manager.ts
hasanaburayyan: The thought here was that maybe plugins are not always absolute paths to some JS file. Maybe we have some wing standard plugins that users can choose to toggle like "tag-resources" or "s3-backend" etc.. so in this case the add() method would control how we add plugins, (as absolute paths or something else). 

Though right now its not useful.

docs/04-reference/winglang-plugins.md
hasanaburayyan: https://github.com/winglang/wing/issues/1474

docs/04-reference/wingcli.md
```diff
@@ -133,6 +133,16 @@ JavaScript bundles that include inflight code that executes on compute platform
...
 You can deploy your stack to GCP using Terraform ([example](/getting-started/aws)).
 
+### Plugins
+
+Additionally the `compile` command can be provided an optional list of plugins to use during the compilation process.
+
+```sh
+$ wing compile --target tf-aws ENTRYPOINT.w --plugins PLUGIN1 PLUGIN2
+```
+The plugins can be absolute paths or relative paths (to the current working directory) to JavaScript files. For more on the structure
```
Chriscbr, line 143:
```suggestion
Each plugin can be an absolute path or relative path to a JavaScript file. For more 
```

docs/04-reference/winglang-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+This hook is called right after the `postSynth` hook and provided the same
+context object. In the context of a  Terraform-based target like `tf-aws`, this
+is the same Terraform JSON configuration. However, does not allow configuration
+to be mutated, which allows plugins to validate the configuration without
+worrying about another plugin mutating after the fact.
+
+The following example validates that buckets all have versioning enabled if they
+don't it will throw an  error during compilation.
```
Chriscbr, line 104:
```suggestion
The following example validates that buckets all have versioning enabled
and throw an error during compilation if they don't.
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+   * This hook is immutable, allowing plugins to perform validations
+   * on the Terraform config after all modifications have been made.
+   *
+   * @param config Terraform config to be passed to the validate hooks
+   */
+  public validate(config: any) {
+    this.callAllHooks(CompilationPhase.VALIDATE, (hook) =>
+      hook.validate(config)
+    );
+  }
```
Chriscbr, line 129: So clean, I love it 😄

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    const hooks: ICompilationHook = {
+      name: pluginAbsolutePath,
+    };
+
+    const pluginDir = resolve(pluginAbsolutePath)
+      .split("/")
+      .slice(0, -1)
+      .join("/");
```
Chriscbr, line 65: If pluginAbsolutePath is absolute do we really need to call `resolve` on it?

If the function expects an absolute path and gets a non-absolute path then it would probably be better to throw / fail fast

Also we can use a helper from the `path ` module:
```suggestion
    const pluginDir = dirname(pluginAbsolutePath)
```

libs/wingsdk/src/core/plugin-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const context = vm.createContext({
+      require,
+      console,
+      exports: hooks,
+      process,
+      __dirname: pluginDir,
+    });
+
+    const pluginCode = readFileSync(resolve(pluginAbsolutePath), "utf8");
```
Chriscbr, line 75: See previous comment
```suggestion
    const pluginCode = readFileSync(pluginAbsolutePath, "utf8");
```

Chriscbr (commented):
Sweet!!!  🚀  Great work Hasan

Looks good to me, though I'll let Elad give another pass if he wants :^)

apps/wing/src/commands/compile.ts
```diff
@@ -200,3 +201,18 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
...
+ * Resolves a list of plugin paths as absolute paths, using the current working directory
+ * if absolute path is not provided.
+ * 
+ * @param plugins list of plugin paths (absolute or relative)
+ * @returns list of absolute plugin paths or relative to cwd
+ */
+function resolvePluginPaths(plugins: string[]): string[] {
+  const resolvedPluginPaths: string[] = [];
+  for (const plugin of plugins) {
+    resolvedPluginPaths.push(isAbsolute(plugin) ? plugin : resolve(process.cwd(), plugin));
```
eladb, line 215: I suspect `resolve(process.cwd(), absPath)` will just return the absolute path:

```suggestion
    resolvedPluginPaths.push(resolve(process.cwd(), plugin));
```

eladb (approved):
🎉

apps/wing/src/commands/compile.ts
hasanaburayyan: Ah reading the docs for that now I think you are right, thats really slick!

PR #1427, opened 2023-02-02
feat(compiler): copy() and copy_mut() for builtin collections

MarkMcCulloh:
The main challenge here is having methods of a generic class that return another generic. I initially wanted to do a recursive style to have this work for any class but I didn't quite figure out a good way to handle it. I think when we add some sort of caching for this hydration process then we could add the recursion back in. 

For now I think it's okay to stick with our builtins, I think this is the most common situation.


examples/tests/invalid/capture_mutables.w
```diff
@@ -2,11 +2,15 @@ let a = MutArray<str>["hello"];
 let s = MutSet<num>{12};
 let m = MutMap<bool>{"hello": true};
 
+let a_cloned = (Array<str>["hello"]).mut_clone();
```
eladb, line 5:
```suggestion
let a_cloned = (Array<str>["hello"]).copy_mut();
```

libs/wingsdk/src/std/array.ts
```diff
@@ -27,6 +27,28 @@ export class ImmutableArray {
...
+
+  /**
+   * Create a mutable shallow copy of this array
+   *
+   * @macro [...($self$)]
+   *
+   * @returns a MutableArray with the same values as this array
+   */
+  public copyMut(): MutableArray {
+    throw new Error("Abstract");
```
eladb, line 39:
```suggestion
    throw new Error("Macro");
```

libs/wingsdk/src/std/array.ts
```diff
@@ -27,6 +27,28 @@ export class ImmutableArray {
...
+  }
+
+  /**
+   * Create an immutable shallow copy of this array
+   *
+   * @macro Object.freeze([...($self$)])
```
eladb, line 45: This is too cool!

libs/wingsdk/src/std/array.ts
```diff
@@ -27,6 +27,28 @@ export class ImmutableArray {
     index;
     throw new Error("Abstract");
   }
+
+  /**
+   * Create a mutable shallow copy of this array
+   *
+   * @macro [...($self$)]
```
eladb, line 34: Damn so cool

libs/wingsdk/src/std/array.ts
```diff
@@ -27,6 +27,28 @@ export class ImmutableArray {
     index;
     throw new Error("Abstract");
   }
+
+  /**
+   * Create a mutable shallow copy of this array
```
eladb, line 32: TBH, why are we doing a shallow copy instead of a deep clone? In most cases you'd want a deep clone...

Maybe we need both?

examples/tests/valid/std_containers.w
```diff
@@ -1,7 +1,12 @@
 let s_array = ["one", "two", "three", "four"];
+let mut_array = s_array.copy_mut();
```
eladb, line 2: Let's make sure tests cover all of these APIs.

eladb (changes requested):

eladb:
+1 on `copy()` and `copy_mut()`. I saw your discussion on Team Time and loved the result ❤️

libs/wingsdk/src/std/array.ts
MarkMcCulloh: I think we should have both eventually, but the purpose of this PR I think shallow is not only good enough but probably expected. `deep` things are not usually the default due to the cost, so this probably shouldn't be any different.

libs/wingsdk/src/std/array.ts
MarkMcCulloh: Added to the other macros as well 👍

libs/wingsdk/src/std/array.ts
eladb: Sounds good to me. Maybe raise an issue and we can gauge need

eladb (approved):

libs/wingsdk/src/std/array.ts
MarkMcCulloh: https://github.com/winglang/wing/issues/1450

libs/wingc/src/type_check.rs
```diff
@@ -1854,7 +1854,26 @@ impl<'a> TypeChecker<'a> {
 							let new_return_type = if sig.return_type.is_same_type_as(original_type_param) {
 								new_type_arg
 							} else {
-								sig.return_type
+								// Handle generic return types
+								// TODO: If a generic class has a method that returns another generic, it must be a builtin
```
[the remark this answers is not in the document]
MarkMcCulloh: https://github.com/winglang/wing/issues/1451

examples/tests/invalid/capture_mutables.w
```diff
@@ -2,11 +2,15 @@ let a = MutArray<str>["hello"];
...
+
 let i = inflight () => {
   assert(a.length == 1);
   //     ^ Cannot reference 'a' of type 'MutArray<str>' from an inflight context
   assert(s.size == 1);
   //     ^ Cannot reference 's' of type 'MutSet<num>' from an inflight context
   assert(m.size == 1);
   //     ^ Cannot reference 'm' of type 'MutMap<bool>' from an inflight context
+  assert(a_cloned.size == 1);
+  //     ^ Cannot reference 'a_cloned' of type 'MutArray<str>' from an inflight context
```
Chriscbr, line 15: Maybe we can tweak the messaging to make it clearer why it can't be referenced (we mention the type but it won't be obvious to all users that the type being mutable is the issue)

"Cannot reference mutable value 'a_cloned' of type 'MutArray<str>' from an inflight context"

PR #1432, opened 2023-02-03
feat(compiler): user-defined resources

yoav-steinberg:
See #542 

This adds initial support for custom resources including generating their inflight clients and handling preflight resource field capturing to be used inflight.

Huge thanks (:clap::clap::clap:) to @Chriscbr and @eladb for lots of help on the design side of this and getting it to tie into the SDK.

```wing
bring cloud;
resource MyResouce {
  inner_bucket: cloud.Bucket;

  init(b: cloud.Bucket) {
    this.inner_bucket = b;
  }

  inflight my_put() {
    this.inner_bucket.put("hello", "cloud");
  }
}

r = new MyResource(new cloud.Bucket());
  r.my_put();
});
```

Not show in the example there's also support for composition (a custom resource creating/using another custom resource). And initial (and partially broken) support for inheritance (a custom resource inheriting another custom resource).

Missing:
- [ ] tests.

There are also some cleanup and refactors here:
* Added `global` static method to `Symbol` for generating arbitrary `Symbol`s. We should probably have some symbol struct without the span field to eliminate this, but for now it makes the ambiguity of using a `String` vs `Symbol` easier to handle.
* In ast made `CustomType` and `Class` structs to be used by the `Type` and `StmtKind` enums respectively. Instead of just inlining the data in the enum variants. This might create some naming confusion but I find these complex variants need some container type to easily use them throughout the code.
* Made the `SymbolEnv` iterator have an optional `with_ancestors` flag. So we can iterate a single level or run up the entire env during iteration. This is very useful when we want to get all fields of a class vs only the fields defined in a subtype without the parent's fields.
* Now for each variable name stored in the `SymbolEnv` we also store its phase. This makes it easier to know the phase of variables without access to the AST. This introduces some conceptual complexity because a `SymbolEnv` also has a phase. So this is how it works:
  * the phase (flight field) in a `VariableInfo` stored in the env means the phase this variable exists in.
  * the phase (flight field) in a `SymbolEnv` means what phase this **environment** was created in.
  In most cases the env phase and the var phase are the same. But for example a resource's env is always preflight while some of it's methods are inflight. Those methods will have a phase different from the env they're defined in.

Finally I want to note (perhaps only to myself) that this PR doesn't take advantage of the "capture" phase of our compiler. It turned out too complex for me to do it that way and I abandoned that path. So that phase currently only handles capturing of inflight closures which should also be removed by turning them into resource sugar later on. I/we need to rethink the "capture" phase - maybe it's just code analysis for binding annotations (permissions)? Ideally I'd like the "jsify" phase to be dumb and just deal with, well, jsifying and move anything smart to the earlier capture phase.


Chriscbr:
> We should probably have some symbol struct without the span field to eliminate this

That or maybe we need some way to model that a symbol is generated or external to the user's source code. Either way the global() method sounds good to me

eladb:
Wooohooo! 🎉🎉

Nit about the title: I rather call these "user-defined resources" (as opposed to resources which are part of the standard library).

libs/wingc/src/capture.rs
```diff
@@ -160,6 +165,7 @@ pub fn scan_for_inflights_in_expression(expr: &Expr, diagnostics: &mut Diagnosti
 					&func_def.statements,
 					diagnostics,
 				)));
+				println!("Collected captures: {:?}", func_captures);
```
eladb, line 168: Debug print?

libs/wingc/src/jsify.rs
```diff
@@ -56,6 +57,10 @@ impl JSifier {
 		}
 	}
 
+	fn js_resolve_file(file_name: &str) -> String {
```
eladb, line 60:
```suggestion
	fn js_resolve_path(file_path: &str) -> String {
```

libs/wingc/src/jsify.rs
```diff
@@ -102,6 +107,7 @@ impl JSifier {
 
 		if self.shim {
 			output.push(format!("const {} = require('{}');", STDLIB, STDLIB_MODULE));
+			output.push(format!("const core = $stdlib.core;"));
```
eladb, line 110:
```suggestion
			output.push(format!("const $core = $stdlib.core;"));
```

As a convention, maybe prefix with $ symbols that are owned by the compiler

libs/wingc/src/jsify.rs
```diff
@@ -440,7 +450,7 @@ impl JSifier {
 						module_name
 					}),
 					if module_name.name.starts_with("\"./") {
-						// TODO so many assumptions here, would only wort with a JS file
+						// TODO so many assumptions here, would only work with a JS file
```
eladb, line 453: Tracking issue for this TODO?

(Let's try to make this a habit so we have better visibility of our tech debt and we can prioritize for beta)

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
 	fn jsify_class_member(&self, member: &ClassMember) -> String {
 		format!("{};", self.jsify_symbol(&member.name))
 	}
+
+	fn jsify_resource(
```
eladb, line 699: Perhaps some commentary on how this function works. 

Maybe even an example of how the output looks like

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+
+		// Get fields to be captured by resource's client
+		let captured_fields = self.get_captures(resource_type);
+
+		// Jsify inflight client
+		let inflight_methods = methods
+			.iter()
+			.filter(|(_, m)| m.signature.flight == Phase::Inflight)
+			.map(|t| *t)
+			.collect::<Vec<_>>();
```
eladb, line 721: Curious: do we really need this "collect"? Why?

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+		// Jsify class
+		let resource_class = format!(
+			"class {}{} {{\n{}\n{}\n{}\n}}",
+			self.jsify_symbol(name),
+			if let Some(parent) = parent {
+				format!(" extends {}", self.jsify_type(parent))
+			} else {
+				format!(" extends {}", WINGSDK_RESOURCE)
+			},
+			self.jsify_resource_constructor(constructor, parent.is_none()),
```
eladb, line 742: Will this also work if there is no Ctor?

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+			preflight_methods
+				.iter()
+				.map(|(n, m)| format!(
+					"{} = {}",
```
eladb, line 746: How does the output look like? Why is there a `=` here?

libs/wingc/src/type_check.rs
```diff
@@ -57,15 +61,24 @@ pub enum SymbolKind {
...
+// Information about a variable in the environment
 #[derive(Debug, Clone)]
 pub struct VariableInfo {
+	// Type of the variable
 	pub _type: TypeRef,
+	// Can the variable be reassigned?
 	pub reassignable: bool,
+	// The phase in which this variable exists
+	pub flight: Phase,
 }
```
Chriscbr, line 73: nit: style (triple-slash is the convention for documentation comments in Rust)
```suggestion
/// Information about a variable in the environment
#[derive(Debug, Clone)]
pub struct VariableInfo {
	/// Type of the variable
	pub _type: TypeRef,
	/// Can the variable be reassigned?
	pub reassignable: bool,
	/// The phase in which this variable exists
	pub flight: Phase,
}
```

libs/wingc/src/type_check.rs
```diff
@@ -57,15 +61,24 @@ pub enum SymbolKind {
...
 
+// Information about a variable in the environment
 #[derive(Debug, Clone)]
 pub struct VariableInfo {
+	// Type of the variable
 	pub _type: TypeRef,
+	// Can the variable be reassigned?
 	pub reassignable: bool,
+	// The phase in which this variable exists
+	pub flight: Phase,
```
Chriscbr, line 72: What would it mean for a variable to exist in an "independent" phase / how do I create such a variable? Maybe add some explanation

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+			let statement_str = self.jsify_statement(
+				constructor.statements.env.borrow().as_ref().unwrap(),
+				statement,
+				Phase::Preflight,
+			);
+			let result = statement_str.split("\n");
+			for l in result {
+				lines.push(format!("  {}", l));
+			}
+		}
```
eladb, line 803: Feels like this bit should be reused with normal methods

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+				format!(
+					"const {}_client = this.{}._toInflight();",
+					inner_member_name, inner_member_name,
+				)
+			})
+			.collect::<Vec<_>>();
+
+		let client_relative_path = format!("clients/{}.inflight.js", resource_name.name);
+		format!(
+			"_toInflight() {{\n{}\nconst self_client_path = {};\nreturn core.NodeJsCode.fromInline(`(new (require(\"${{self_client_path}}\")).{}_inflight({{{}}}))`);\n}}",
```
eladb, line 836: Can we perhaps make this more readable by using an array of strings and joining them using `\n`?

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+			.iter()
+			.map(|(inner_member_name, _, _)| {
+				format!(
+					"const {}_client = this.{}._toInflight();",
+					inner_member_name, inner_member_name,
+				)
+			})
+			.collect::<Vec<_>>();
+
+		let client_relative_path = format!("clients/{}.inflight.js", resource_name.name);
```
eladb, line 834: Put "clients" in a constant

libs/wingc/src/type_check.rs
```diff
@@ -1110,7 +1124,7 @@ impl<'a> TypeChecker<'a> {
 		}
 
 		// Verify that all non-optional fields are present and are of the right type
-		for (k, v) in expected_struct.env.iter().map(|(k, v)| {
+		for (k, v) in expected_struct.env.iter(true).map(|(k, v, _)| {
```
Chriscbr, line 1127: WDYT of `iter_with_ancestry()` instead of `iter(true)`? Feels slightly more readable to me

libs/wingc/src/jsify.rs
```diff
@@ -710,6 +695,315 @@ impl JSifier {
...
+					.iter()
+					.map(|(name, _, _)| name.clone())
+					.collect_vec()
+					.join(", ")
+			)
+		} else {
+			"".to_string()
+		};
+
+		// TDOD jsify inflight fields
```
eladb, line 882: Tracking issue 🙏🫶

libs/wingc/src/type_check.rs
```diff
@@ -1459,19 +1456,16 @@ impl<'a> TypeChecker<'a> {
...
 				constructor,
 				is_resource,
-			} => {
-				let env_flight = if *is_resource {
-					Phase::Preflight
-				} else {
-					Phase::Inflight
-				};
+			}) => {
+				// Resources canno't be defined inflight
```
Chriscbr, line 1467:
```suggestion
				// Resources cannot be defined inflight
```

libs/wingc/src/jsify.rs
```diff
@@ -521,49 +531,23 @@ impl JSifier {
...
+			}) => self.jsify_class(
+				env,
+				name,
+				*is_resource,
+				phase,
+				parent,
+				constructor,
+				&members.iter().collect::<Vec<_>>(),
+				&methods.iter().collect::<Vec<_>>(),
+			),
```
Chriscbr, line 550: Pass the AstClass directly perhaps?

```
StmtKind::Class(class) => self.jsify_class(env, phase, class),
```

Chriscbr (commented):
Incredible! 😄

libs/wingc/src/ast.rs
```diff
@@ -83,7 +92,23 @@ pub enum Type {
...
 	MutSet(Box<Type>),
 	FunctionSignature(FunctionSignature),
-	CustomType { root: Symbol, fields: Vec<Symbol> },
+	CustomType(CustomType),
+}
+
+// In the future this may be an enum for type-alias, class, etc. For now its just a nested name.
+// Also this root,fields thing isn't really useful, should just turn in to a Vec<Symbol>.
+#[derive(Debug, Clone)]
+pub struct CustomType {
```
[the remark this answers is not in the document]
yoav-steinberg: I'm done with this, just note that this is for **any** user defined type (any type with a non-builtin name), so if we have type aliases in the future they will also go here.

libs/wingc/src/ast.rs
```diff
@@ -184,6 +209,16 @@ pub struct ElifBlock {
 	pub statements: Scope,
 }
 
+#[derive(Debug)]
+pub struct Class {
+	pub name: Symbol,
+	pub members: Vec<ClassMember>,
+	pub methods: Vec<(Symbol, FunctionDefinition)>,
+	pub constructor: Constructor,
```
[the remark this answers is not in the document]
yoav-steinberg: Here's the naming issue as I see it:
* **constructors** create something from scratch.
* **initializers** initialize something but don't (necessarily?) create it.
* **factories** create one or more things from scratch.
I'm leaning towards an approach where we **only** have factories, we may want to call them something else maybe `init`. But that might not be the correct word.
To make things more complicated there might a distinction between what resources do (they initialize something that _may_ already exist - hence init) and what classes/structs/jsons... do (they construct something new).
I'm waiting a bit with the suggested name change here to get feedback. Also it's not strictly related to this PR.

libs/wingc/src/ast.rs
```diff
@@ -184,6 +209,16 @@ pub struct ElifBlock {
...
+	pub parent: Option<Type>,
```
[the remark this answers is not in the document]
yoav-steinberg: This won't work, since a `Type` is a name (in this case a `UserDefinedType`) and a `Class` is a definition of a class (init,fields,methods...). The `parent` field should be the **name** of the parent type not the parent type definition itself.
It **might** be possible to change this to a  `UserDefinedType` I'll give it a try.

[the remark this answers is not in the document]
yoav-steinberg: >  It might be possible to change this to a UserDefinedType I'll give it a try.

done.

libs/wingc/src/capture.rs
yoav-steinberg: right. removed.

libs/wingc/src/jsify.rs
staycoolcall911: Let's add a test for that

libs/wingc/src/lib.rs
```diff
@@ -193,18 +193,15 @@ pub fn type_check(scope: &mut Scope, types: &mut Types) -> Diagnostics {
...
+	let sym = Symbol::global(name);
 	scope
 		.env
 		.borrow_mut()
 		.as_mut()
 		.unwrap()
 		.define(
 			&sym,
-			SymbolKind::make_variable(types.add_type(typ), false),
+			SymbolKind::make_variable(types.add_type(typ), false, Phase::Independent),
```
staycoolcall911, line 204: Do all our builtin functions have to be Phase::Independent?
I know the current 4 ones are, but maybe it makes sense to pass the phase as a parameter (as an anecdote @MarkMcCulloh provided a suggestion for an inflight builtin [here](https://winglang.slack.com/archives/C048QCN2XLJ/p1675353044361399)).

libs/wingc/src/type_check.rs
```diff
@@ -1522,14 +1516,14 @@ impl<'a> TypeChecker<'a> {
 				};
 
 				// Create a the real class environment to be filled with the class AST types
-				let mut class_env = SymbolEnv::new(parent_class_env, self.types.void(), true, false, env_flight, stmt.idx);
+				let mut class_env = SymbolEnv::new(parent_class_env, self.types.void(), true, false, env.flight, stmt.idx);
 
 				// Add members to the class env
```
staycoolcall911, line 1521:
```suggestion
				// Add fields to the class env
```

libs/wingc/src/type_check.rs
```diff
@@ -1671,11 +1665,11 @@ impl<'a> TypeChecker<'a> {
 				let mut struct_env = SymbolEnv::new(None, self.types.void(), true, false, env.flight, stmt.idx);
 
 				// Add members to the struct env
```
staycoolcall911, line 1667:
```suggestion
				// Add fields to the struct env
```

libs/wingc/src/jsify.rs
yoav-steinberg: done

libs/wingc/src/jsify.rs
```diff
@@ -56,6 +58,10 @@ impl JSifier {
 		}
 	}
 
+	fn js_resolve_path(file_name: &str) -> String {
```
staycoolcall911, line 61:
```suggestion
	fn js_resolve_path(file_path: &str) -> String {
```

libs/wingc/src/jsify.rs
```diff
@@ -56,6 +58,10 @@ impl JSifier {
...
+		format!("require('path').resolve(__dirname, \"{}\")", file_name)
```
staycoolcall911, line 62:
```suggestion
		format!("require('path').resolve(__dirname, \"{}\")", file_path)
```

libs/wingc/src/jsify.rs
staycoolcall911: @yoav-steinberg - perhaps it's enough to add a comment in #864

libs/wingc/src/jsify.rs
```diff
@@ -707,9 +694,319 @@ impl JSifier {
...
+					"{} = {}",
+					n.name,
+					self.jsify_function(None, &m.parameters, &m.statements, phase)
+				))
+				.collect::<Vec<String>>()
+				.join("\n")
+		)
+	}
+
+	// Get the type and capture info for fields that are captured in the client of a the given resource
```
staycoolcall911, line 977:
```suggestion
	// Get the type and capture info for fields that are captured in the client of the given resource
```

libs/wingc/src/jsify.rs
```diff
@@ -707,9 +694,319 @@ impl JSifier {
...
+	fn get_captures(&self, resource_type: TypeRef) -> Vec<(String, TypeRef, Vec<String>)> {
+		resource_type
+			.as_resource()
+			.unwrap()
+			.env
+			.iter(true)
+			.filter(|(_, kind, _)| {
+				let var = kind.as_variable().unwrap();
+				// We capture preflight non-reassignable fields, we currently only support capturing resources
+				// need to add immutable primitives in the future
```
staycoolcall911, line 987:
```suggestion
				// TODO: need to add immutable primitives in the future
```

staycoolcall911 (commented):
Great job @yoav-steinberg - this is a major milestone

libs/wingc/src/lib.rs
MarkMcCulloh: I think this should be fine as-is, the symbol itself can be phase independent while it represents an inflight function (my proposal).

libs/wingc/src/ast.rs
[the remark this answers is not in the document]
eladb: OK, let's keep the names as is for now. We can revisit later.

libs/wingc/src/capture.rs
```diff
@@ -63,14 +63,19 @@ pub fn scan_for_inflights_in_scope(scope: &Scope, diagnostics: &mut Diagnostics)
...
 				methods,
 				constructor,
 				parent: _,
-				is_resource: _,
-			} => {
+				is_resource,
+			}) => {
+				// If this is a resource then we need to capture all its members
+				if *is_resource {
+					// TODO: what do I do with these?
```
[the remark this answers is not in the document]
yoav-steinberg: Added a top comment in the file with a general description of where I stand regarding the compiler's capture phase (more or less copied from my comment in the top of this PR). It also references some open issues. Hope that's good enough for now.

libs/wingc/src/jsify.rs
yoav-steinberg: We use `inflight_methods` in two places so I figured we can collect this and use it twice. Not perfect but was quicker than passing an iterator to `jsify_resource_client` and then dealing with resetting / cloning the iter.

libs/wingc/src/jsify.rs
yoav-steinberg: We currently force a ctor even if it's empty.
When we change that then `constructor` will be an `Option` and I'll fix it accordingly (we need to call `super(scope,id)` in any case).

libs/wingc/src/jsify.rs
yoav-steinberg: looks like this:
```wing
// wing
resource Foo {
  init() {}
  a_method() {}
}
```
turns into preflight js:
```js
class Foo extends $stdlib.core.Resource {
  constructor(scope, id) {
    super(scope, id);
  }
  a_method = () => {}
```

libs/wingc/src/jsify.rs
yoav-steinberg: ended up using [indoc](https://github.com/dtolnay/indoc)
should eventually change a lot of the jsification to use it, pretty cool.

libs/wingc/src/jsify.rs
yoav-steinberg: done

libs/wingc/src/type_check.rs
yoav-steinberg: It's just that the default hsould be to iter with the ancestry. I could create a `iter_without_ancestry` but that feels clunky.

libs/wingc/src/type_check.rs
yoav-steinberg: We currently don't support independent phase so I'm not sure. The general idea is that if you have a:
```wing
struct Foo {
  a: num
}
```
Then there's nothing preventing us from using the type `Foo` in either phase (and therefore also exchanging it between them. In that case `a` can be defined phase independent.
I don't feel like adding any documentation about this yet since there are too many unclear aspects and we don't support it yet.

libs/wingc/src/jsify.rs
yoav-steinberg: much better :sparkles: 
thanks.

libs/wingc/src/lib.rs
yoav-steinberg: Actually, @MarkMcCulloh, the phase of the symbol is deduced from the phase of the `SymbolEnv` it's defined in (in this case the global scope is preflight). The argument passed to `make_variable` is the phase of the function itself (in this case hard-coded phase independent).
@staycoolcall911 I think that when we even have an inflight builtin we can easily change this. I'm not sure Mark's `includejs` is necessarily inflight only...

libs/wingc/src/type_check.rs
Chriscbr: Where are we iterating without ancestry?

yoav-steinberg: > Where are we iterating without ancestry?

I guess that nowhere... I did this at some point to handle different inheritance related issues. For example you only want to initialize the current resource client's class fields in `init` and you need to pass all other `init` arguments to the client classes `super`. Another case is just to make sure you're jsifying the current resources methods and assuming parent class methods were already jsifyied.
In some places where I just iter over the AST and not the env this isn't an issue. In other places I eventually iterred with ancestry and filtered out the current classes stuff from the ancestors stuff because I acutally neeeded both.
I still think we'll eventually need this so I'm hesitant whether to remove it.

staycoolcall911 (approved):

eladb:
a huge milestone for Wing! - @yoav-steinberg @Chriscbr @MarkMcCulloh @staycoolcall911 - great job!

ShaiBer:
💪💪💪💪

Chriscbr:
Awesome work!

Should we create an issue to add tests for this? @yoav-steinberg

yoav-steinberg:
> 
> Should we create an issue to add tests for this? @yoav-steinberg 

We should

PR #1435, opened 2023-02-03
feat!: compile artifacts into target-specific directories

### LINKED ISSUE #815: Compiler output structure
If I run:

```sh
wing compile -t sim file1.w
```

but also:

```sh
wing compile -t sim file2.w
```

I'll end up with the following files:

```
cwd/
  target/
    .wing/
    file1.w
    file2.w
    tree.json # this file has been overwritten by the second compilation
  file1.w
  file2.w
```

Is it a know issue? Do we want to handle this?

Related to https://github.com/winglang/wing/pull/770.

Chriscbr:
This PR updates the CLI so that all compiler artifacts (like terraform files, simulator files, etc.) are now synthesized into target-specific directories (closes #815). For example, `wing compile -t tf-aws hello.w` will generate terraform artifacts inside `target/hello.tfaws/` while `wing compile -t tf-azure` will generate terraform artifacts inside `target/hello.tfazure/`. This avoids some problems users had where compiling to different targets would override previously compiled artifacts.

The simulator target now generates a single `.wsim` which is generated directly inside `target/`. (Under the hood, the `.wsim` file now contains the `tree.json` file that was previously generated outside).

This is a breaking change as it may require using a new version of the Wing console that understands how to read the updated .wsim file.

### DIFF
--- a/libs/wingsdk/src/core/app.ts
+++ b/libs/wingsdk/src/core/app.ts
@@ -115,7 +115,7 @@ export class CdktfApp extends Construct implements IApp {
     rmSync(this.cdktfApp.outdir, { recursive: true });
 
     // write outdir/tree.json
-    synthesizeTree(this);
+    synthesizeTree(this, this.outdir);
 
     // return a cleaned snapshot of the resulting Terraform manifest for unit testing
     const tfConfig = this.cdktfStack.toTerraform();
--- a/libs/wingsdk/src/core/app.ts
+++ b/libs/wingsdk/src/core/app.ts
@@ -1,11 +1,14 @@
+import { mkdirSync, readdirSync, renameSync, rmSync } from "fs";
 import { join } from "path";
 import * as cdktf from "cdktf";
 import { Construct, IConstruct } from "constructs";
 import { IPolyconFactory, Polycons } from "polycons";
 import stringify from "safe-stable-stringify";
-import { Files } from "./files";
+// import { Files } from "./files";
 import { synthesizeTree } from "./tree";
 import { Logger } from "../cloud/logger";
+
+const TERRAFORM_STACK_NAME = "root";
 
 /**
  * A Wing application.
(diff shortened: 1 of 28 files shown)


Chriscbr:
FYI @ainvoner @skyrpex @polamoros  I'm labeling this with `do-not-merge` until the console is ready to support these new changes. The only change that affects the console AFAIK is that the `.wsim` file now contains `tree.json` inside it directly (it is no longer generated outside).

libs/wingsdk/src/core/app.ts
```diff
@@ -99,17 +99,46 @@ export class CdktfApp extends Construct implements IApp {
...
+  /**
+   * Move files from `outdir/cdktf.out/stacks/root` to `outdir`.
+   */
+  private moveCdktfArtifactsToOutdir(): void {
+    const cdktfOutdir = this.cdktfApp.outdir;
+    const cdktfStackDir = join(cdktfOutdir, "stacks", TERRAFORM_STACK_NAME);
+
+    const files = readdirSync(cdktfStackDir, { withFileTypes: true });
+    for (const file of files) {
+      if (file.isFile()) {
```
eladb, line 135: TIL

libs/wingsdk/src/target-sim/app.ts
```diff
@@ -45,6 +45,9 @@ export class App extends Construct implements core.IApp {
     // write simulator.json file into workdir
     this.synthSimulatorFile(workdir);
 
+    // write tree.json file into workdir
+    core.synthesizeTree(this, workdir);
```
eladb, line 49: Is this going to break the Wing Console?

Chriscbr: Probably. But I think it should be possible to update the console to work in a backwards-compatible way

eladb: Backward compatibility is the responsibility of the "server". In our case it's the wing toolchain, not the Console.

The "right thing" to do here is to emit both the old format and the new format until the console is updated to support the new format.

Chriscbr: Ok, good point. 🙂 Updated to be backwards compatible... I tested out the changes with the latest version of the Wing console and it works as expected.

Chriscbr:
@eladb Even if we're not breaking the wing console I'm guessing we should still make this a minor version bump, yeah? We're still changing where users are going to find their terraform files etc.

eladb (approved):

Issue #1386, opened 2023-01-31
cloud.Function.memory()


staycoolcall911:
https://docs.winglang.io/reference/wingsdk-spec#function

staycoolcall911:
@raywonkari - I saw you started working on this in #1437, so I assigned it to you. Please let me know if you need assistance.

Chriscbr:
Added by #1437

PR #1437, opened 2023-02-04
feat(sdk): ability to define lambda memory

raywonkari:
https://github.com/winglang/wing/issues/1386

introduces a new parameter to `FunctionProps` i.e., `memory`. This enables us to specify memory of the lambda function.


libs/wingsdk/src/cloud/function.ts
```diff
@@ -30,6 +30,12 @@ export interface FunctionProps {
    * @default 1m
    */
   readonly timeout?: Duration;
+
+  /**
+   * The amount of memory to allocate to the function, in MB.
+   * @default 128
```
revitalbarletz, line 36: Is there a max value as well? Do we want to force that max?

libs/wingsdk/src/cloud/function.ts
```diff
@@ -30,6 +30,12 @@ export interface FunctionProps {
...
+   */
+  readonly memory?: number;
```
Chriscbr, line 38: @raywonkari We should also implement or handle this field for other cloud targets. (What happens if `memory` is configured while you run your app on the Wing simulator or on Azure?)

- For the simulator, I think it is OK if we do not simulate memory limits since it would require significant rearchitecture, so we should just add a comment in the constructor [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-sim/function.ts) saying that "props.memory is unused since we are not simulating it"
- For Azure, I don' t think the memory is directly configurable so I would throw an error in the constructor [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-tf-azure/function.ts) if `props.memory` is configured (if the value is not undefined). cc @hasanaburayyan Let me know if this sounds ok

libs/wingsdk/src/cloud/function.ts
Chriscbr: Great idea -- for any cloud targets that support his field, we should try and check that the user passes a valid input. We should start by checking the AWS docs and see what are the constraints are (like minimum and maximum) and see if they can be implemented as code

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -144,6 +144,7 @@ export class Function extends cloud.FunctionBase {
       timeout: props.timeout
         ? props.timeout.seconds
         : Duration.fromMinutes(1).seconds,
+      memorySize: props.memory ? props.memory : undefined,
```
Chriscbr, line 147: Can we add tests showing what the expected behavior is when memory is configured on a cloud.Function for each cloud?

Chriscbr (changes requested):

Chriscbr:
> could be a bug or I might be messing things up, but in the above example, the new env vars were not appended to the overall env vars in the cdk tfoutput. I was only seeing the below vars:
> 
> ```
>         "environment": {
>           "variables": {
>             "BUCKET_NAME_d755b447": "${aws_s3_bucket.root_cloudBucket_4F3C4F53.bucket}",
>             "WING_FUNCTION_NAME": "cloud-Function-c8d2eca1"
>           }
>         },
> ```

Is this an issue also present in the latest version of wing? Could you create a separate issue for it perhaps?

yoav-steinberg:
Hi,
My 2 cents here is that this really feels like a non-functional aspect of `cloud.Function`. I think our current solution for this is using the [plugin system](https://docs.winglang.io/contributors/rfcs/2023-01-21-wing-plugins).
Since this seems like something actually useful, maybe we need to provide a collection of non-functional plugins for specific cloud providers and add a pluging for tf-aws lambda memory instead of this approach?

@hasanaburayyan Does this make sense?

eladb:
> Hi, My 2 cents here is that this really feels like a non-functional aspect of `cloud.Function`. I think our current solution for this is using the [plugin system](https://docs.winglang.io/contributors/rfcs/2023-01-21-wing-plugins). Since this seems like something actually useful, maybe we need to provide a collection of non-functional plugins for specific cloud providers and add a pluging for tf-aws lambda memory instead of this approach?
> 
> @hasanaburayyan Does this make sense?

I think memory and timeout are not purely "non-functional" and they do have a place in our SDK since they are aspects of the resource definition that developers should be aware of.

I suspect the term "non-functional" might not be the right distinction...

Chriscbr:
@yoav-steinberg I'm also on the fence about this once but I think it's OK to try it out and see how well the water holds for this API (I'm not concerned about a slippery slope where we start including every operational aspect in our APIs). Here's how I described it in the SDK spec:

> Operational aspects of resources should not leak into the core API surface area, except when they are essential to the functional behavior of the resource and the user's mental model.

One question one might ask is whether this is something that is AWS specific -- for example I didn't see a way to configure memory for Azure functions (@hasanaburayyan does this sound accurate?) -- but it looks like there is an option to configure memory for Google Cloud functions.

Chriscbr:
@raywonkari We are still interested in this PR! 😄 I see there's still a few outstanding comments from my review, but let me know if there's anything we can do to help bring this over the finish line.

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -138,6 +138,11 @@ export class Function extends cloud.FunctionBase {
 
     const name = ResourceNames.generateName(this, FUNCTION_NAME_OPTS);
 
+    // validate memory size
+    if (props.memory && (props.memory < 128 || props.memory > 10240)) {
+      throw new Error("Memory should be in between 128 & 10240");
```
Chriscbr, line 143:
```suggestion
      throw new Error("Memory for AWS Lambda function should be in between 128 and 10240");
```

Chriscbr (approved):
Great work @raywonkari! I left one small suggestion for the error message but this is good to merge as is 👍

staycoolcall911:
Nice! Congrats @raywonkari for completing this! 
(and @Chriscbr for assisting)

PR #1441, opened 2023-02-06
chore: use pull_request instead of pull_request_target for builds

MarkMcCulloh:
We switched to pull_request_target a while back while making sure fork workflows would work. This made sense for the self-mutation workflow, but this isn't necessary for the build workflow. 
If triggered from a PR, the workflow does not use any secrets or attempt to write anything so it doesn't need those permissions. The only time secrets are read is on `push` to main.

The primary benefit of this change is that now we can easily test workflow changes in PRs. It's also just the generally recommended way for doing these workflows.

The explicit `ref` and `repository` settings were removed because pull_request_target didn't set it in the expected way. No longer needed, PRs will checkout the merge commit now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,9 +32,6 @@ jobs:
       - name: Checkout
         if: github.event_name == 'push' && github.repository == 'winglang/wing'
         uses: actions/checkout@v3
-        with:
-          ref: ${{ github.event.pull_request.head.ref || github.ref }}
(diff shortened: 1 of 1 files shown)


Chriscbr:
> The only time secrets are read is on push to main.

I see there are some secrets used elsewhere in build.yml (in the E2E test job), does that matter at all / can that be removed?

Chriscbr (approved):
Conditionally approved, just check the comment I left

eladb (approved):

PR #1453, opened 2023-02-06
feat: wasm language server and simplified build system

### LINKED ISSUE #821: Ensure tree-sitter queries stay valid
### Summary

For those not using vscode, the tree-sitter query we have provides syntax highlighting. Currently our grammar changes and can easily break the `highlights.scm` query.

### Feature Spec

tree-sitter cli should run highlight tests in our GH workflows to ensure they don't break as we change the grammar.

### Use Cases

_No response_

### Implementation Notes

_No response_

### Component

_No response_

### LINKED ISSUE #1396: Remove npm registry setup from hangar
### Community Note

### Feature Spec

When hangar was initially created, we were making use of private GitHub packages. There was a desire to make sure this functionality worked as intended, so hangar did a very complete E2E flow that involved mocking and downloading from a local NPM registry.

This is fun and all, but it is very slow and adds a lot of time to the tests. At this point we use public npm for distribution and there is not much benefit for hangar to still do this.,

### Use Cases

Should hopefully x2 the speed of our tests

### Implementation Notes

_No response_

### Component

Development Environment, Contributor Experience

MarkMcCulloh:
The Wing cli now has a `lsp` command which starts the language server. With this change, there are no more native parts of our toolchain. There are several ramifications for this:
- Nx no longer has to worry about `native` vs `wasm`, everything is just `build`
- The github workflow similarly only has one build job now where it builds everything
- The existing wing-language-server project has been removed, most of the code migrated to wingc under the wls module
- vscode now just uses npx to resolve the wing cli for the language server, alternatively an env var.

In the process of the simplification, Fixed #1396 by removing the npm registry stuff from hangar.
Also Fixes #821 because I removed the rust tests from tree-sitter-wing so we actually run all the tests.

**With all these changes, the build time is cut in half!**

### Future Improvement
Testing the rust crates still requires native builds. To make it faster, it only does a debug build. I attempted to use a wasmtime to instead run the tests through that, but ran into too many hurdles so decided to leave it.

### Intentional regression
Removed yarn from hangar. Had issues forcing it to resolve the SDK tarball. This worked before because mocking the registry is pretty brute force. This is specifically a problem with hangar, not the usage of wing with yarn. In order for it to work, we have to hack around yarn to force certain resolutions. At that point, what's the value in testing with it?


apps/wing/src/wingc.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  return instance;
+}
+
+/**
+ * Assumptions:
+ * 1. The called WASM function is expecting a pointer and a length representing a string
+ * 2. The string will be UTF-8 encoded
+ * 3. The string will be less than 2^32 bytes long  (4GB)
+ * 4. the WASI instance has already been started
```
eladb, line 71:
```suggestion
 * 4. The WASI instance has already been started
```

apps/wing/src/wingc.ts
```diff
@@ -0,0 +1,117 @@
...
+const LOW_MASK = 2n ** 32n - 1n;
+const HIGH_MASK = BigInt(32);
+
+export interface WingCLoadOptions {
+  imports?: Record<string, any>;
+  preopens?: Record<string, string>;
+  env?: Record<string, string>;
+}
+
+export async function loadWingc(options: WingCLoadOptions) {
```
eladb, line 32: We are already in the context of "wingc.ts":

```suggestion
export async function load(options: WingCompilerLoadOptions) {
```

apps/wing/src/wingc.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * Assumptions:
+ * 1. The called WASM function is expecting a pointer and a length representing a string
+ * 2. The string will be UTF-8 encoded
+ * 3. The string will be less than 2^32 bytes long  (4GB)
+ * 4. the WASI instance has already been started
+ */
+export function wingcInvoke(
```
eladb, line 73:
```suggestion
export function invoke(
```

apps/wing/src/wingc.ts
```diff
@@ -0,0 +1,117 @@
...
+  instance: WebAssembly.Instance,
+  func: WingCFunction,
```
eladb, line 75:
```suggestion
  func: WingCompilerFunction,
```

libs/tree-sitter-wing/package.json
```diff
@@ -18,10 +18,7 @@
   "dependencies": {
     "nan": "^2.12.1"
   },
-  "devDependencies": {
-    "tree-sitter-cli": "^0.20.6"
-  },
   "scripts": {
-    "test": "tree-sitter test"
+    "test": "npx -y tree-sitter-cli@0.20.7 test"
   }
```
eladb, line 23: why not as a normal devDep?

package.json
```diff
@@ -19,7 +19,7 @@
     "contributors:check": "all-contributors check",
     "postinstall": "run-p install:*",
     "docs": "./scripts/docsite.sh",
-    "install:wasi": "if [ \"$CI\" != \"true\" ]; then bash scripts/setup_wasi.sh; fi",
+    "install:wasi": "bash scripts/setup_wasi.sh",
```
eladb, line 22: Is this still needed?

libs/tree-sitter-wing/package.json
MarkMcCulloh: running npm install for this package runs node-gyp, which is both slow and caused issues in the build because it's expected that the c output had already been generated

I just realized I can add `--ignore-scripts` to fix that though so I'll add it back.

apps/wing/src/wingc.ts
```diff
@@ -0,0 +1,120 @@
...
+      );
+
+      toFree.push([returnPtr, returnLen]);
+
+      return new TextDecoder().decode(extractedBuffer) + "";
+    }
+  } finally {
+    toFree.forEach(([pointer, length]) => {
+      exports.wingc_free(pointer, length);
+    });
```
Chriscbr, line 118: I'm curious what happens if we don't free everything

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+mod semantic_token;
+
+pub struct FileData {
+	pub contents: String,
+	pub tree: Tree,
+	pub diagnostics: Diagnostics,
+}
+
+thread_local! {
+	/// Under normal cases, wingc is not in full control of the process in which it is running.
```
Chriscbr, line 30: What does this mean ("under normal cases")?

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+	pub fn send_notification(
+		notification_type: *const u8,
+		notification_type_length: u32,
+		data: *const u8,
+		data_length: u32,
+	);
+}
+
+/// Sends a notification to the client, not part of the typical request/response cycle.
+/// On wasm32, this is expected to be implemented whatever is consuming wingc.
```
Chriscbr, line 47:
```suggestion
/// On wasm32, this is expected to be implemented by whatever is consuming wingc.
```

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+#[allow(dead_code)]
+fn send_log(message: &str) {
+	let notification = serde_json::json!({
+		"type": lsp_types::MessageType::INFO,
+		"message": message.to_string(),
+	});
+
+	unsafe {
+		let json = serde_json::to_string(&notification).unwrap();
+		let notif_type = "window/logMessage".to_string().into_bytes().leak();
```
Chriscbr, line 212: Should this be a constant somewhere?

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+		} else {
+			Some(SemanticTokensResult::Tokens(SemanticTokens {
+				result_id: None,
+				data: relative_tokens,
+			}))
+		}
+	})
+}
+
+#[allow(dead_code)]
```
Chriscbr, line 203: Why? (add a comment)

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+		.collect_vec();
+
+	let notification = serde_json::json!({
+		"uri": uri,
+		"diagnostics": final_diags,
+	});
+
+	unsafe {
+		let json = serde_json::to_string(&notification).unwrap();
+		let notif_type = "textDocument/publishDiagnostics".to_string().into_bytes().leak();
```
Chriscbr, line 247: Constant

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+use self::{completions::completions_from_ast, semantic_token::semantic_token_from_ast};
+
+mod completions;
+mod semantic_token;
+
+pub struct FileData {
+	pub contents: String,
+	pub tree: Tree,
+	pub diagnostics: Diagnostics,
+}
```
Chriscbr, line 27: Can we add comments to this?

tools/hangar/src/cli.test.ts
```diff
@@ -99,36 +91,21 @@ async function enterTestDir(testDir: string) {
...
-enum InvocationType {
-	Direct,
-	NPX,
-}
+async function runWingCommand(command: string[], wingFile: string) {
+  const isError = path.dirname(wingFile).endsWith("error");
 
-async function runWingCommand(type: InvocationType, command: string[], wingFile: string) {
-	const isError = path.dirname(wingFile).endsWith("error");
+  const cmd: string[] = [npxBin, "winglang", ...command, wingFile];
```
Chriscbr, line 97: Why not invoke the winglang bin directly? Are we trying to test `npx`-related behavior, or are the tests faster this way, or is there another reason?

Chriscbr (commented):
Looks awesome 🔥 🔥 🚒

libs/wingc/src/lsp/mod.rs
MarkMcCulloh: My intent was "When wingc is consumed as a WASM library", so I suppose I could write that instead 😁

apps/wing/src/wingc.ts
MarkMcCulloh: We leak the memory. How much that matters depends on the use case. For compilation it's not really a big deal but for the lsp it means memory usage for the process would grow over time.
I figured I would be a good memory custodian and always free.

tools/hangar/src/cli.test.ts
MarkMcCulloh: Yeah good point. I think initially there was an intent to test workflows, and npx is probably a common one. but since hangar installs the cli, npx is really not doing much. npx is just slower.

Changed it to direct invocation.

libs/wingc/src/lsp/mod.rs
MarkMcCulloh: I just deleted it instead

libs/wingc/src/lsp/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+use self::{completions::completions_from_ast, semantic_token::semantic_token_from_ast};
+
+mod completions;
+mod semantic_token;
+
+/// The result of running wingc on a file
+pub struct FileData {
+	/// Text data contained in the file (utf8)
+	pub contents: String,
```
Chriscbr, line 26:
```suggestion
	/// Text data contained in the source file (utf8) for matching up with diagnostics
	/// TODO: how to support multiple source files
	pub contents: String,
```

MarkMcCulloh: This struct is stored in a hashmap, mapping `file -> FileData`. So it currently supports "multiple files". Although what that may technically mean in the future could change. Either way not sure the TODO is needed.

Chriscbr: Oh ok, got it. I was just a little confused at first when I saw the struct was about the output of the compiler and it contained the input/source code of compilation - but in the context of the LSP I figured it made sense

Chriscbr (approved):
🚢

PR #1454, opened 2023-02-06
fix: windows path delimiters break cli

MarkMcCulloh:
The first of many additions of `replace(/\\/g, "/")` :)

While the CLI does work on windows, if the user provides any `\`s in the path it breaks the wasm preopens.

### DIFF
--- a/apps/wing/src/commands/compile.ts
+++ b/apps/wing/src/commands/compile.ts
@@ -9,6 +9,7 @@ import { mkdir, readFile } from "fs/promises";
 import { WASI } from "wasi";
 import debug from "debug";
 import * as chalk from "chalk";
+import { normalPath } from "src/util";
 
 const log = debug("wing:compile");
 const WINGC_COMPILE = "wingc_compile";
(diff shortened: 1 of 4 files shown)


apps/wing/src/commands/compile.ts
```diff
@@ -48,6 +48,13 @@ export interface ICompileOptions {
   readonly target: Target;
 }
 
+/**
+ * Normalizes paths from windows to posix.
+ */
+function normalPath(path: string) {
```
MarkMcCulloh, line 54: ℹ️  I hesitated on adding a utils.ts or something, but I can do that if this is ugly

staycoolcall911: Since you're calling that `.replace(...)` from multiple places and since almost every app or lib on our monorepo has some sort of a utils module, I would add utils.ts, but I'm leaving it to you to decide and approving your PR anyway.

staycoolcall911 (approved):

apps/wing/src/commands/compile.ts
MarkMcCulloh: Alright you convinced me

apps/wing/src/commands/compile.ts
```diff
@@ -3,19 +3,20 @@
...
 import { WASI } from "wasi";
 import debug from "debug";
 import * as chalk from "chalk";
+import { normalPath } from "../util";
 
 const log = debug("wing:compile");
 const WINGC_COMPILE = "wingc_compile";
 
-const WINGC_WASM_PATH = resolve(__dirname, "../../wingc.wasm");
+const WINGC_WASM_PATH = normalPath(resolve(__dirname, "../../wingc.wasm"));
```
eladb, line 17: Why not `resolve(__dirname, "..", "..", "wingc.wasm")`?

PR #1457, opened 2023-02-06
fix: windows absolute paths break cli

MarkMcCulloh:
Followup to https://github.com/winglang/wing/pull/1454

My iteration cycle on fixes for Windows is a bit slow. Also added an if check to do the normalization cause it would be wasted cycles outside of windows.

### DIFF
--- a/apps/wing/src/util.ts
+++ b/apps/wing/src/util.ts
@@ -2,6 +2,10 @@
  * Normalizes paths from windows to posix.
  */
 export function normalPath(path: string) {
-  return path.replace(/\\/g, "/");
+  if (process.platform === "win32") {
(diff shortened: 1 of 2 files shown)


apps/wing/src/util.ts
```diff
@@ -2,6 +2,10 @@
  * Normalizes paths from windows to posix.
  */
 export function normalPath(path: string) {
-  return path.replace(/\\/g, "/");
+  if (process.platform === "win32") {
+    return path.replace(/\\/g, "/").replace(/^[a-zA-Z]:/, "");
```
Chriscbr, line 6: Can you add a comment explaining these regex's?

Chriscbr (approved):

PR #1458, opened 2023-02-06
feat(docs): spec for `json`

eladb:
Add a specification for how Wing works with JSON by addressing `Json` and `MutJson` types and a various of patterns across the surface.

Part of #1053

Feedback is welcome!


docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
-> console.log(data.name);
-> ```
+The type of elements within objects and arrays is still `Json`.
 
-</details>
+To access a field within an object, use the `.` notation:
+
+```js
+let boom: Json = json_obj.boom;
+```
```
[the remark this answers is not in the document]
Chriscbr: (idea: We could say that `json_obj.boom` is just sugar for `Json.get(json_obj, "boom")`)

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+let schema = Contact.schema();
+schema.validate(j);
+```
+
+##### 1.1.4.3. MutJson
+
+To define a mutable JSON container, use the `MutJson` type:
+
+```js
+let my_obj = MutJson { hello: "dear" };
```
Chriscbr, line 350: Are there any mutable operations you can do on a `MutJson "hello"` or `MutJson true`?

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+my object is: {
+  boom: 123
+}
+```
+
+It is also legal to just print a json object:
+
+```js
+print(json_mut_obj);
+```
```
Chriscbr, line 456: Another use case idea:
- Can I enumerate all keys, values, or key-value pairs from a Json?

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+
+```js
+assert(Json.equals(json_string, Json "hello"));
+assert(Json.equals(json_obj, { boom: [ 1, 2, 3 ] }));
+assert(!Json.equals(Json { hello: [ 1, 2, 3 ] }, Json { hello: [ 1, 2 ] }));
+```
+
+The `Json.diff(lhs: Json, rhs: Json): JsonPatch` static method can be used to calculate the deep
+difference between two JSON values. It returns a list of differences in
+[json-patch](https://jsonpatch.com/) format.
```
Chriscbr, line 411:
```suggestion
The `Json.diff(lhs: Json, rhs: Json): JsonPatch` static method can be used to calculate the deep
difference between two JSON values. It returns a list of differences in
[json-patch](https://jsonpatch.com/) format. (P2)
```

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+let j2 = Json {
+  baz: "boo",
+  hello: ["world"]
+};
+
+assert(Json.diff(j1, j2) = [
+  { op: JsonPatch.REPLACE, path: "/baz", value: "boo" },
+  { op: JsonPatch.ADD, path: "/hello", value: ["world"] },
+  { op: JsonPatch.REMOVE, path: "/foo" }
+]);
```
Chriscbr, line 428: How is the list of returned operations ordered? Or should `Json.diff` return a `Set` of operations?

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+  b: my_num,
+  c: my_bool,
+  d: my_arr
+};
+```
+
+##### 1.1.4.2. Parsing
+
+To convert a `Json` to a strong-type, use the `parse_json()` static method on the relevant target
+type:
```
Chriscbr, line 238: I think there are two kinds of "parsing" that we are interested in:
1. casting Json to structured types
2. parsing raw strings into Json

I wonder if calling (1) "parsing" could be confusing? Also I think (2) might be missing from the spec

docs/04-reference/winglang-spec.md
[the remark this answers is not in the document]
hasanaburayyan: What about multiple nested fields with spaces or special punctuation? 
So like `"//"."$metadata"` is the path then you would do `Json.get(Json.get(json_obj, "//"), "$metadata")`.

perhaps providing index like sugar `json_obj["//"]["$metadata"]` would be a nicer experience where you cant use `.` notation

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+```
+
+To access an array element, use the `[]` notation:
+
+```js
+let item2: Json = json_array[2];
+```
+
+It is also possible to assign the native `str`, `num`, `bool` and `Array<T>` values and they will
+implicitly be casted to `Json`:
```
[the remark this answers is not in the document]
eladb: No, it is not a valid statement (at least for now). It makes sense to allow implicit assignment from *safe* to *unsafe* but not the other way around.

In this specific example, the compiler can technically tell that the `Json` is just a string and since it is immutable, it can technically allow implicit assignment. But it's not a very interesting case because why would anyone even use this idiom? (they don't need the `Json`).

I'll add this as a clarification to the spec.

docs/04-reference/winglang-spec.md
eladb: Yes, I am really not in love with this name.

How about `from_json`?

```js
let s: str = str.from_json(my_json.foo.bar.baz);
```

This is also aligned with the factory method pattern (CDK uses `fromXxx` for all factory methods). I know @yoav-steinberg has some cool ideas on formalizing factory methods, which means that if we have 1st class support for them then those `from_json` methods will be included.

docs/04-reference/winglang-spec.md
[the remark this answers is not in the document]
eladb: I added syntax to support accessing non-symbolic keys through `[]` (same as JavaScript).

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,287 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+let my_arr = Array<num>.parse_json(json_array);
+```
+
+>
+> **A NOTE ABOUT NAMING**
+>
+> `parse_json` is very long and pretty ugly. We should think of alternatives.
+> One option is to put all these methods under `Json.as_xxx()`,
+> so `Json.as_num(Json 1234) == 1234`. But this won't work very well for user-defined
+> structs...
```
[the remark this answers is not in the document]
eladb: This requires using types as values which is not really something we have in the language today. Feels like something we can avoid for now.

docs/04-reference/winglang-spec.md
eladb: Not really, but for the sake of completeness, you can still define it like that.

eladb: added a clarification

docs/04-reference/winglang-spec.md
eladb: added

docs/04-reference/winglang-spec.md
eladb: yap

docs/04-reference/winglang-spec.md
eladb: I think this can't really be a set. The order can affect the result. I'll leave it for json-diff to take care of .

docs/04-reference/winglang-spec.md
eladb: OK, changed to `from_json` and added `Json.parse(s): Json` for the 2nd use case. Dah!

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,288 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+
+let my_array = Json [1,2,3,"hello"];
+Array<num>.parse_json(my_array); // RUNTIME ERROR: unable to parse `[1,2,3,"hello"]` as an array of `num`.
+```
+
+Use `unsafe: true` to disable this check at your own risk:
+
+```js
+let x = num.parse_json(Json [1,2,3], unsafe: true);
+print(x + 5); // OUTPUT: 1,2,35
```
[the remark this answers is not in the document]
eladb: Changed to a "positive example" and also indicated this is P2.

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,389 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+It is also possible to assign the native `str`, `num`, `bool` and `Array<T>` values and they will
+implicitly be casted to `Json`:
+
+```js
+let my_str: str = "hello";
+let my_num: num = 183;
+let my_bool: bool = true;
+let my_arr: Array<num> = [1,2,3];
+
+let json_string = Json {
```
flyingImer, line 294:
```suggestion
let json_obj = Json { 
```

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,389 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+
+
+Use `unsafe: true` to disable this check at your own risk:
+
+```js
+let x = num.from_json(Json [1,2,3], unsafe: true);
+print(x + 5); // OUTPUT: 1,2,35
+```
+
+For each `from_json()`, there is a `try_from_json()` method which returns an optional `T?` which
```
flyingImer, line 344: does `try_from_json()` have `unsafe: true` knob as well?

out of nowhere, was thinking `from_json(my_json, try: true)`, not sure if it fits winglang semantics overall

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,389 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+
+> NOTE: `values()` and `entries()` return an array inside a `Json` object because at the moment we
+> cannot represent heterogenous arrays in Wing.
+
+##### 1.1.4.3 JSON arrays
+
+To access an array element, use the `[]` notation:
+
+```js
+let item2: Json = json_array[2];
```
Chriscbr, line 278: I would emphasize that this should be the inferred type (if/when we have type inference)
```suggestion
let item2 = json_array[2]; // type: Json
```

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,389 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+
+To obtain an array of all the values, use `Json.values(o)`. To obtain an array of all key/value
+pairs use `Json.entries(o)` (P2):
+
+```js
+assert(Json.values(j).equals(Json [ 123, [ 1, 2, 3 ] ]));
+assert(Json.entries(j).equals(Json [
+  [ "hello", 123 ],
+  [ "world", [ 1, 2, 3 ] ]
+]));
```
Chriscbr, line 267: I would actually expect:
```suggestion
assert(Json.values(j).equals([ Json 123, Json [ 1, 2, 3 ] ]));
assert(Json.entries(j).equals([
  [ Json "hello", Json 123 ],
  [ Json "world", Json [ 1, 2, 3 ] ]
]));
```
TL;DR we can provide a more specific type to users -- `Array<Json>` instead of `Json` in the first example, and `Array<Array<Json>>` instead of `Json` in the second example.

docs/04-reference/winglang-spec.md
```diff
@@ -172,45 +172,389 @@ inflight (arg1: <type1>, arg2: <type2>, ...) => <type>
...
+```js
+let item2: Json = json_array[2];
+```
+
+Trying to index a value that is not an array will return JavaScript `undefined`.
+
+##### 1.1.4.4 Assignment from native types
+
+It is also possible to assign the native `str`, `num`, `bool` and `Array<T>` values and they will
+implicitly be casted to `Json`:
```
Chriscbr, line 286: Does this mean if a function has a `Json` parameter I can pass it a `str` and it will be upcasted (downcasted? I'm bad at directions so I have no idea which is correct 😅 ) for free?

eladb: Yes, this should work I believe.

docs/04-reference/winglang-spec.md
eladb: > `from_json(my_json, try: true)`

That won't work because it implies that the return type of `from_json` must always be and optional `T?`.

docs/04-reference/winglang-spec.md
eladb: cool

staycoolcall911 (approved):
Great - I think we can start implementing it and see if anything is still unclear or should be altered

Issue #630, opened 2022-11-15
Wing Console Windows support


ekeren:
### Summary

Ability to run Wing console on windows

### Component

Wing Console

Note: when this is done we should consider implementing #875

ainvoner:
- [ ] CI multi-platform
- [x] tests
- [ ] artifacts
- [ ] S3 upload
- [x] icon
- [x] installer
- [x] quit app - zombies
- [x] open app using protocol
- [x] open app with file input
- [x] cancel file input - process still exists in task manager
- [x] application manu
- [ ] auto updater
- [x] multiple windows (simulators)
- [ ] update docs

ekeren:
Nice work @ainvoner

staycoolcall911:
Wow nice @ainvoner!!!

PR #1462, opened 2023-02-07
feat: windows support for toolchain

### LINKED ISSUE #1000: Windows support for Wing CLI
As a user with a Windows machine, I would like to be able to use Wing.

Our CLI used to work with Windows before, so we need to:
- [x] Fix the CLI, so it works on Windows machines
- [x] Add a test to our CI to continuously build on a Windows machine (and break when it doesn't)

Related:
#644 
#630

MarkMcCulloh:
- Undoes some changes from previous PRs regarding windows, ideally tries to do the _path normalization_ as infrequently and late as necessary
- Rewrite of hangar to no longer use https://github.com/google/zx. It had issues with Windows and/or Vitest. It was useful before when hangar was more complicated, but now just using https://github.com/sindresorhus/execa instead
  - hangar is functionally the same

Fixes #1000


eladb (approved):
Ship it yeah!

apps/wing/src/wingc.ts
```diff
@@ -24,19 +24,27 @@ export interface WingCompilerLoadOptions {
...
+    // .jsii access
+    [WINGSDK_MANIFEST_ROOT]: WINGSDK_MANIFEST_ROOT,
+    ...(options.preopens ?? {}),
+  } as Record<string, string>;
+
+  if (process.platform === "win32") {
+    for (const [key, value] of Object.entries(preopens)) {
+      delete preopens[key];
+      preopens[normalPath(value)] = value;
+    }
```
Chriscbr, line 37: Mutating the same collection you're iterating over.... living life dangerously I see? 😄

MarkMcCulloh: 😎 
Technically iterating over `Object.entries(preopens)`, so should be completely safe to mutate the original afaik

revitalbarletz:
Does it mean that win developers can now contribute as well?

staycoolcall911:
Generally yes, but wing console is not supported yet on windows, see [Wing Console Windows support](https://github.com/winglang/wing/issues/630)

MarkMcCulloh:
@revitalbarletz If you mean can windows developers build the wing repo? That is untested and I don't think we've talked about ever supporting that

revitalbarletz:
@MarkMcCulloh can you please elaborate what is included in this change?

MarkMcCulloh:
@revitalbarletz This PR adds a couple of fixes for Windows and includes it in our E2E testing. This is specifically for running Wing as a user, e.g. running `wing compile -t sim cool.w` on Windows in a windows-native shell like PowerShell.

However:
- Someone using Windows is not well supported to develop/contribute to this project. They may be able to get everything working, but it's not an officially considered development environment
- The Wing Console is not directly affected by the changes here and requires its own work to support Windows in any way

Issue #926, opened 2022-12-19
Snapshot tests: invalid examples should also have snapshots of the error messages


staycoolcall911:
### Summary

Snapshot tests for invalid examples

### Feature Spec

We currently have 2 types of examples https://github.com/winglang/wing/tree/main/examples/tests: valid and invalid.

For valid examples - we generate snapshots of the compiled js.
For invalid examples - we just expect them to fail, whether it's panic or a specific error message. Of course this type of behavior is flawed because we want to snapshot the reason for failure and notice when/if it changes. We also want to be able to put multiple invalid statements in each file and get error message for each.

### Use Cases

See for example:
```ts
wing compile ../../examples/tests/invalid/container_types.w
```
Which prints out multiple error messages and we want all of them to appear in the snapshot.

### Component

Compiler

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.29](https://github.com/winglang/wing/releases/tag/v0.5.29).

PR #1463, opened 2023-02-07
fix: compiler errors cause distraction exception

MarkMcCulloh:
Ever since wingc stopped being a binary, invoking `compile` doesn't cause an error unless there's a panic. This means that execution isn't stopped in the CLI for regular errors but the rest of it expects the preflight js to exist, causing it to throw trying to read the file.


eladb (changes requested):
Test?

libs/wingc/src/lib.rs
```diff
@@ -78,11 +78,16 @@ pub unsafe extern "C" fn wingc_compile(ptr: u32, len: u32) {
...
-		eprintln!(
-			"Compilation failed with {} errors\n{}",
+		let result = format!(
+			"Compilation failed with {} error(s)\n{}",
 			err.len(),
 			err.iter().map(|d| format!("{}", d)).collect::<Vec<_>>().join("\n")
 		);
+
+		let leaked = result.into_bytes().leak();
+		combine_ptr_and_length(leaked.as_ptr() as u32, leaked.len() as u32)
```
yoav-steinberg, line 88: This feels WASM only? Does this break native builds?

MarkMcCulloh: This function is intentionally meant to only be used with WASM. There is an additional compile function that can be used by the native side with proper types.

Also with regards to breaking native builds, you may want to see [this PR](https://github.com/winglang/wing/pull/1453) if you haven't yet.

MarkMcCulloh:
@eladb tests will be coming in https://github.com/winglang/wing/issues/926

eladb:
> @eladb tests will be coming in #926

OK

libs/wingc/src/lib.rs
Chriscbr, line 88: If we have a `ptr_to_string` would it make sense to abstract this into a `string_to_ptr` function? I think there might be other places we do this as well

MarkMcCulloh: Makes sense, it's used in two places and will be used in more over time

MarkMcCulloh:
Recontextualized in https://github.com/winglang/wing/pull/1499

Issue #1392, opened 2023-01-31
time api language spec


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

The https://docs.winglang.io/reference/spec#65-roadmap has left the Date and Time api on the TODO list. 

While I was writing user story, I noticed that Javascript's `Date.now()` api feels like a good api for vm based programming language, but it felt awkward use it on wing where something like `cloud.Time.now()` might be more suitable.

### Use Cases

Data.now()

### Component

Language Design

eladb:
I'd like us to have a standard type called `datetime` (like `duration`) and then:

```js
datetime.now()
```

Feels right.

I'll add to the spec

staycoolcall911:
Sounds good @eladb - I guess we simply want to add as a wrapper over the javascript object, like we did with `str`. Need to decide what's our api and add a reference doc.

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.36](https://github.com/winglang/wing/releases/tag/v0.5.36).

PR #1470, opened 2023-02-07
feat(docs): spec for `duration` and `datetime`

eladb:
Spec for `duration` and `datetime` standard types.

Resolves https://github.com/winglang/wing/issues/1392


docs/04-reference/winglang-spec.md
```diff
@@ -216,6 +216,80 @@ it cannot be modified anymore.
...
+```
+
+Duration objects are immutable and can be referenced across inflight context.
+
+#### 1.1.6 `Datetime`
+
+The `Datetime` (alias `datetime`) type represents a single moment in time in a platform-independent
+format. It uses the
+[`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+JavaScript object under the hood and encapsulate an integral number that represents milliseconds
```
staycoolcall911, line 247:
```suggestion
JavaScript object under the hood and encapsulates an integral number that represents milliseconds
```

staycoolcall911 (commented):
Looks good. I'll let another team member review. 
The next person to LGTM - please approve.

docs/04-reference/winglang-spec.md
```diff
@@ -216,6 +216,80 @@ it cannot be modified anymore.
 
 ---
 
+#### 1.1.5 `Duration`
+
+The `Duration` (alias `duration`) type represents a time duration.
+
+Duration literals are numbers with `m`, `s`, `h` suffixes:
```
MarkMcCulloh, line 223: What about having a duration literal for every component of a datetime, for consistency:

```
year: Y;
month: M;
day: D;
hour: h;
min: m;
sec: s;
ms: ms;
```

Will probably also be useful for cron-like stuff later on.

MarkMcCulloh (commented):
I like this, the choice to stick to UTC is good although we should get also start thinking about APIs to do TZ conversion as well. Maybe something like ZonedDataTime ala [Temporal](https://tc39.es/proposal-temporal/docs/index.html)

docs/04-reference/winglang-spec.md
```diff
@@ -216,6 +216,80 @@ it cannot be modified anymore.
...
+assert(three_hours.minutes == 180);
+```
+
+Duration objects are immutable and can be referenced across inflight context.
+
+#### 1.1.6 `Datetime`
+
+The `Datetime` (alias `datetime`) type represents a single moment in time in a platform-independent
+format. It uses the
+[`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
```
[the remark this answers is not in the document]
eladb: Yes, fair enough

docs/04-reference/winglang-spec.md
eladb: Wild! But maybe later?

MarkMcCulloh: Very wild, although we should at least also have `ms` initially since that's pretty common/uncontroversial.

docs/04-reference/winglang-spec.md
```diff
@@ -216,6 +216,80 @@ it cannot be modified anymore.
...
+  hour: num;
+  min: num;
+  sec: num;
+  ms: num;
+}
+
+class Datetime {
+  static now(): Datetime;                 // returns the current DateTime
+  static from_iso(iso: str): Datetime;    // creates a instance from an ISO-8601 string
+  static from_components(c: DatetimeComponents): Datetime;
```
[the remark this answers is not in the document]
eladb: great point about UTC :thumbsup: - I see what you mean about `now()` having no other reasonable meaning. I still think `Datetime` should take timezones into account and has a `timezone` component, but I am going to remove `now()` and only keep `utc_now()`.

docs/04-reference/winglang-spec.md
```diff
@@ -216,6 +216,80 @@ it cannot be modified anymore.
...
+  year: num;
+  month: num;
+  day: num;
+  hour: num;
+  min: num;
+  sec: num;
+  ms: num;
+}
+
+class Datetime {
```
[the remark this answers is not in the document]
eladb: Done

docs/04-reference/winglang-spec.md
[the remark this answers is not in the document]
eladb: I've added `system_now()` to allow non-cloud use cases (because we are a general purpose language after all), but it's putting off :-)

docs/04-reference/winglang-spec.md
```diff
@@ -560,6 +560,91 @@ print(json_mut_obj);
...
+  min: num;
+  sec: num;
+  ms: num;
+  tz: num; // timezone offset in minutes from UTC
+}
+
+class Datetime {
+  static utc_now(): Datetime;             // returns the current time in UTC timezone
+  static system_now(): Datetime;          // returns the current time in system timezone
+  static from_iso(iso: str): Datetime;    // creates a instance from an ISO-8601 string
```
eladb, line 610:
```suggestion
  static from_iso(iso: str): Datetime;    // creates an instance from an ISO-8601 string
```

Chriscbr (approved):
LGTM

PR #1472, opened 2023-02-07
fix: unable to run tree-sitter tests from clean repo

MarkMcCulloh:
Previously avoided running post-install scripts before because the tree-sitter output isn't committed (nodegyp fails). This causes `tree-sitter test` to fail because the tree-sitter-cli doesn't work. There's no reason to avoid committing the tree-sitter output, and conventionally that is what people do.


Chriscbr:
@MarkMcCulloh this is fixing a bug (a bug in the contributor experience is still a bug 🙂) so I've changed it to `fix`, lmk if that's okay

Chriscbr (approved):
🙏

Issue #1479, opened 2023-02-08
Intermittent build issue in wing playground and wing cli


MarkMcCulloh:
https://github.com/winglang/wing/actions/runs/4124106745/jobs/7122963127#step:8:385

I've seen this more than once, root cause is unclear. Not able to replicate locally and it happens in Actions rarely.

staycoolcall911:
@MarkMcCulloh - is this a regression after you introduced the [WASM build PR](https://github.com/winglang/wing/pull/1453)? I was going to ask you how you changed the bundling of wingsdk - maybe it's related

MarkMcCulloh:
It would make sense that we'd have a regression like this after a big change like that, but I can't pinpoint anything in particular other than aggravating an existing race condition during the build process

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.14](https://github.com/winglang/wing/releases/tag/v0.5.14).

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.22](https://github.com/winglang/wing/releases/tag/v0.5.22).

PR #1483, opened 2023-02-08
chore: limit parallelism for root `npm install`

MarkMcCulloh:
Trying to figure out ways to address #1479

Still not clear on root cause, but I have a hunch it's related to the npm install process. So with that, I made it so that the root "install everything" script doesn't run all installs at once to hopefully avoid whatever corruption is happening. Frankly this is probably a decent idea anyways, with no throttling we were surely hitting some io (net or fs) wall anyways.

Also, I saw that the lockfile for wing-playground was missing a `name` for wingsdk. Not sure if it matters, since I don't think module resolution cares about lockfiles, but hey why not make sure that's in here.


package.json
```diff
@@ -17,17 +17,17 @@
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate",
     "contributors:check": "all-contributors check",
-    "postinstall": "run-p install:*",
     "docs": "./scripts/docsite.sh",
+    "postinstall": "run-p -l --max-parallel 3 'install:*'",
```
MarkMcCulloh, line 21: ℹ️  `-l` adds labels to the command, which I think helps clear up what the command does. Not related to the fix, just figured it looks nice.

package.json
```diff
@@ -17,17 +17,17 @@
...
     "install:jsii-docgen": "npm --prefix apps/jsii-docgen install",
-    "install:wing-playground": "npm --prefix apps/wing-playground install",
-    "install:winglang": "npm --prefix apps/wing install",
     "install:sdk": "npm --prefix libs/wingsdk install",
-    "install:hangar": "npm --prefix tools/hangar install"
+    "install:tree-sitter-wing": "npm --prefix libs/tree-sitter-wing install",
+    "install:winglang": "npm --prefix apps/wing install",
+    "install:hangar": "npm --prefix tools/hangar install",
+    "install:wing-playground": "npm --prefix apps/wing-playground install",
+    "install:vscode-wing": "npm --prefix apps/vscode-wing install"
```
Chriscbr, line 30: Alphabetical order perhaps?

Chriscbr (approved):

package.json
MarkMcCulloh: I attempted to put it in dependency order in case that was part of the thrashing and `run-p` would vaguely respect the order, but after saying that out loud it's silly so yeah I'll alphabetize.

Issue #436, opened 2022-11-03
Optionality


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#17-optionality

Spec revised in #1495. This lists the basic Optionality features belonging to this issue, while more advanced features are listed in #2103

- [x] Support `?` 
- [x] Support `x: T?` 
- [x] Support `??` 
- [x] Make sure optional primitives are captureable  - done in #2258
- [ ] Support `?.`
- [x] Support the `nil` keyword
- [x] Support `if let`

## More info 
Add support for the following `?` and `??` operations
```ts (wing)
print(k ?? "k is null");

if !k? {
  print("k is null");
} else { 
 print(k);
}
```

Also notice that currently `str?` is not capturable (and it should be)

```ts (wing)
bring cloud;

let f = (): str? =>{
  print("running f");
};
let k = f();

new cloud.Function(inflight () => {
 let c = k;
}) as "test";
```
  
produces the following compilation error:
```sh
Compilation failed with 1 error(s)
Error at aaa.w:9:10 | Cannot reference 'k' of type 'str?' from an inflight context
```

eladb:
@staycoolcall911 i think we will need something like this for beta. Otherwise there’s no way to unwrap optionals in a safe way.

I think ?? is going to be the only way to unwrap optionals in v1 (since we don’t have static analysis), so this is going to be pretty important.

Assign this to me and I’ll do a small spec revision to make sure we cover all the bases

ekeren:
@eladb  we are using this issue to implement ? and ?? 

you should work on completing the RFC for optionality

monadabot:
Congrats! :rocket: This was released in [Wing 0.17.48](https://github.com/winglang/wing/releases/tag/v0.17.48).

PR #1495, opened 2023-02-09
feat(docs): optionality spec

### LINKED ISSUE #224: Should the language support optional chaining?
```ts
class Dog {
	name: str;
	... //init with name field
}

class MyDogs {
	d1: Dog;
	d2: Dog?;
	... //init with d1 field
};

let dog1 = new Dog("Tarzan");
let myDogs = new MyDogs(d1);
let dog2Name = myDogs.d2?.name; //is this supported or does it throw a compilation error?
```

eladb:
Finalize the initial spec for optional types in Wing.

We are ambitiously trying to to avoid explicit `nil`s in the language to promote safety. Please read this with a critical eye.

Part of #436
Fixes #224 


docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+`throw()` can be used as an r-value. This promotes defensive and safe programming.
+
+```js
+let address = my_person.address ?? throw("address is required");
+```
+
+#### 1.7.6 Optional return types
+
+If a function returns an optional type, use the `return ?` statement to indicate that the value is
+not defined.
```
eladb, line 898: yes, I know this is effectively null, which might mean we can't avoid it. Would love thoughts here.

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+    return ?;
+  }
+
+  return parts.at(1);
+}
+```
+
+#### 1.7.7 Lazy evaluation (P2)
+
+The `?=` expression can be used to implement lazy evaluation:
```
eladb, line 913:
```suggestion
The `=?` expression can be used to implement lazy evaluation:
```

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+  var full_name: str?;
+  
+  init(first: str, last: str) {
+    this.first = first;
+    this.last = last;
+  }
+
+  // lazy evaluation
+  full_name(): str {
+    return this.full_name ?= "${this._first} ${this._last}";
```
eladb, line 928:
```suggestion
    return this.full_name =? "${this._first} ${this._last}";
```

docs/04-reference/winglang-spec.md
MarkMcCulloh: +1 to anything other than `return ?;`

`return ?;` Doesn't semantically match the uses above.

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+    return ?;
+  }
+
+  return parts.at(1);
+}
+```
+
+#### 1.7.7 Lazy evaluation (P2)
+
+The `=?` expression can be used to implement lazy evaluation:
```
MarkMcCulloh, line 913: This implies a sort of equivalence between optional and lazy values, which I think would be confusing. We don't want to overload the meaning of `?` more than we have to.

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+
+class Foo {
+  my_opt: num?;
+}
+```
+
+In the above struct, the `address` field, the `foo` argument and the `my_opt` field are all marked
+as optionals using `?`.
+
+This means, for example, that when a `Person` can be initialized without defining the `address`
```
yoav-steinberg, line 817:
```suggestion
This means, for example, that a `Person` can be initialized without defining the `address`
```

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+x = "hello"; // OK
+
+x = nil;
+//  ^^^ ERROR: no explicit `nil` in this language.
+```
+
+> You can be a smart-ass and do this:
+> ```js
+> let make_nil = (): str? => { return ?; }
+> x = make_nil();
```
yoav-steinberg, line 842: Why not support assigning a "not assigned" value:
```
let var x: num? = 7;
x = ?;
```
I guess we can play around with not supporting this and only allowing the smart-ass solution above, for a start. WDYT?

docs/04-reference/winglang-spec.md
MarkMcCulloh: Also `lazy` probably deserves its own separate spec update not tied to optionals.

docs/04-reference/winglang-spec.md
eladb: Yes, I considered it, but I felt we want felt like this should be more explicit. But I am open to trying this out because `return ?;` reads more like "return some value but I don't know what it is", which is not what we want to convey.

(eventually I suspect we will end up with `return nil;` being the only place where you can actually specify `nil`).

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
 > ```
+> Good luck with that.
 
-</details>
+#### 1.7.2 Testing if an optional has a value
+
+Object of type `str?` cannot be used within `==` or `!=` expressions because we don't have a null
+value. Instead, use the the `?` operator to get a `bool` that indicates if the optional is defined
+or not:
```
MarkMcCulloh, line 850: We also intend to have `?.` right?
```
let test = maybe?.stuff;
```

Having an `?` operator makes this ambiguous.

OlafConijn:
what about optional chaining?

given the following structure
```
interface A {
  b?: B
}

interface B {
  c: str;
}
```

and the following statements
```
const myA: A = {};
const myC_1 = myA.b.c; //would this be valid? I would expect myC to be str?
const myC_2 = myA.b?.c; //would this be valid? I would expect myC to be str?
const myC_3 = myA.b!.c;
```

I haven't looked at the Json type yet, but specifically, when parsing json I could see this type of thing to be useful.
If I think of the 3 statements above then I would prefer `myC_1` over `myC_2` as the optionally would natively flow through the object graph. largely unsure about `myC_3`

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+// can be negated
+if !my_person.address? {
+  print("address is not defined");
+}
+```
+
+#### 1.7.3 Unwrapping within conditions
+
+The `if let` statement can be used to test if an optional is defined an *unwrap* it into a
+non-optional variable defined inside the block:
```
yoav-steinberg, line 869: I was thinking about the supper common use case of conditionally unwrapping an optional and then using the inner value. Here are some things I things I'd like which I think rust/swift don't do nicely:
1) I want the unwrapped symbol to have the same name as the optional's symbol in my inner scope:
```
let x: num? = 7;
if x? {
  print(x); // The symbol x was implicitly defined to the unwrapped `x` and is shadowing the optional `x` by the `x?` above.
}
```
This of course can't work in all cases, but adding a `let` keyword here still feels too much so:
```
if my_person.address?:address {
  print(address); // address symbol is the unwrapped my_person.address
}
```

2) I want the conditional unwrapping expression to be used in more complex expressions. This can work nicely by combining the `?` operator that returns a boolean with the `if let` functionality:
```
if x? && print_enabled() && my_presoon.address?:address {
  print("${address} ${x}");
}

if x? || print_enabled { // compilation error because there's a code path where we'll enter the `if` without unwrapping
}
```

docs/04-reference/winglang-spec.md
MarkMcCulloh: Also if we ever have that classic ternary using `?`

yoav-steinberg (approved):
like it

docs/04-reference/winglang-spec.md
OlafConijn: i like `return nil` more than `return ?`, having the compiler check you can only "return nil" seems a fair solution.

i might like `return nil` more than `return;`. Mainly because the compiler (and error messages) will need to give nil a name (it is a concept that exists, also if you want to prevent it) and you might as well unambiguously use the name you give it: `return nil`

eladb (commented):
Change `return ?` to `return nil`

docs/04-reference/winglang-spec.md
eladb: Done

docs/04-reference/winglang-spec.md
eladb: yes, that was my thinking. let's try without this and see what blows up.

docs/04-reference/winglang-spec.md
eladb: I think ternaries are important and I want to support them.
Let me think about this.

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
 
 ### 1.7 Optionality
 
-Symbol `?` can mark a type as optional.  
-Optionality means the value behind type can be either present or nil.
+Tri-state "nullity" values (`null`, `undefined` in JavaScript) are a primary source of bugs in
```
Chriscbr, line 785: Even dual-state nullity values are a source of bugs, let alone JavaScript's version 😅

docs/04-reference/winglang-spec.md
eladb: I was trying to avoid this static analysis scenarios as P1

docs/04-reference/winglang-spec.md
Chriscbr, line 842: What does this "?" represent here? a none-option value?

docs/04-reference/winglang-spec.md
yoav-steinberg: > We also intend to have `?.` right?
> 
> ```
> let test = maybe?.stuff;
> ```
> 

What will go into `test` if `maybe` is nil? I think the only way to do this is:
```
let test = (maybe ?? default).stuff;
```

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+
+// or within a condition
+if my_person.address? {
+  print("address is defined but i do not care what it is");
+}
+
+// can be negated
+if !my_person.address? {
+  print("address is not defined");
+}
```
Chriscbr, line 863: (meta) It's funny, on one hand I feel like we are rationally making things more safe by removing "null" from the language, but also I sort of miss the human-readable code like Python has:

```python
if address is not None:
  do_stuff()
```

I don't have a concrete suggestion here, just sharing as food for thought

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+```
+
+#### 1.7.3 Unwrapping within conditions
+
+The `if let` statement can be used to test if an optional is defined an *unwrap* it into a
```
ShaiBer, line 868:
```suggestion
The `if let` statement can be used to test if an optional is defined and *unwrap* it into a
```

docs/04-reference/winglang-spec.md
Chriscbr: Maybe the word "lazy" is confusing here - I think what's happening here is just short circuit evaluation (sort of how like && doesn't evaluate the second argument if the first is false)

MarkMcCulloh: Oh is `=?` supposed to be `only assign if the lvalue is nil`?

skyrpex:
My boring (and biased) take is that given that JS is one of the most used languages in the world (which also feels easy to use), it would be beneficial to inherit some things from it, like:
- [Optional chaining (?.)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
- [Nullish coalescing assignment (??=)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)
- [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

Regarding the divergence of `null` and `undefined`, I'm not sure if there's a way to avoid making the differentiation in our language. In JSON, undefined doesn't exist but null does...

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
 
 ### 1.7 Optionality
 
-Symbol `?` can mark a type as optional.  
-Optionality means the value behind type can be either present or nil.
+Tri-state "nullity" values (`null`, `undefined` in JavaScript) are a primary source of bugs in
+software. Being able to guarantee that a value will never be null makes it possible to write safe
+code without thinking about nullity.
```
flyingImer, line 787: ❤️ this idea

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+}
+```
+
+#### 1.7.3 Unwrapping within conditions
+
+The `if let` statement can be used to test if an optional is defined an *unwrap* it into a
+non-optional variable defined inside the block:
+
+```js
+if let address = my_person.address {
```
hasanaburayyan, line 872: Originally I was not a big fan of this during our `let x = x` debate but mulling it over some I think this is pretty intuitive. (Swift docs sold me ) 

What about `while let`

docs/04-reference/winglang-spec.md
hasanaburayyan: It really is just a nice syntax, <3 the zen of python

hasanaburayyan: It really is just a nice syntax, <3 the zen of python

docs/04-reference/winglang-spec.md
skinny85: I think `test` will have the type `T?`, where `T` is the type of `maybe.stuff`.

docs/04-reference/winglang-spec.md
skinny85, line 872: Isn't this weird that it's:

```js
if my_person.address? {
  // stuff
```

, but it's 

```js
if let address = my_person.address {
```

? Is there a way to make these consistent?

?

docs/04-reference/winglang-spec.md
skinny85: I echo Mark's confusion here - I think `=?` needs more explanation, I also don't understand how it's different than just `=`, which already returns its value 😛.

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+not defined.
+
+```js
+let parse_last_name = (full_name: str): str? => {
+  let parts = full_name.split(" ");
+  if parts.len < 2 {
+    return ?;
+  }
+
+  return parts.at(1);
```
skinny85, line 907: There might be a reason to have a version of `at()` that returns `T?` for `Array<T>`, but I understand what's the motivation behind the example 🙂.

skinny85 (commented):
Just my two random cents, feel free to ignore 🙂.

docs/04-reference/winglang-spec.md
Chriscbr: Yeah on second reading I'm not sure what this syntax does, though it might just be an issue with the example

docs/04-reference/winglang-spec.md
```diff
@@ -782,30 +782,153 @@ let f = (arg1: num, var arg2: num) => {
...
+Let's look at various use cases for how to interact with optionals.
+
+#### 1.7.2 Assignment
+
+A reassignable variable or field can only be assigned to a concrete value. There is no way to
+explicitly assign a "null" value:
+
+```js
+let var x: str?;
+x = "hello"; // OK
```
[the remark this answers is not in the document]
ekeren: @yoav-steinberg , I am not sure 
```ts
let x = 123?;
```

makes sense to me. Do you mean?
```ts
let var x = 123?;
```

docs/04-reference/winglang-spec.md
eladb: Wen't with `return nil`

docs/04-reference/winglang-spec.md
eladb: Going with `if let` for now to support these use cases without support for complex expressions and without requiring static analysis.

docs/04-reference/winglang-spec.md
eladb: If think `if !address? { .. }` reads quite well, no?

docs/04-reference/winglang-spec.md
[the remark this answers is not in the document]
eladb: We can consider something like this in the future.

docs/04-reference/winglang-spec.md
eladb: We can consider `while let` in the future as needed.

eladb: Not sure I understand why you feel this is inconsistent?

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+  print("the ip address is defined and it is: ${ip_address}");
+}
+```
+
+#### 1.7.6 Roadmap
+
+In the future we will consider the following additional sugar syntax:
+
+* `x ?? throw("message")` to unwrap `x` or throw if `x` is not defined.
+* `x =?? value` returns `x` or assigns a value to it and returns it to support lazy evaluation/memoization.
```
eladb, line 1088:
```suggestion
* `x ??= value` returns `x` or assigns a value to it and returns it to support lazy evaluation/memoization.
```

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+Here's a quick summary of how optionality works in Wing:
 
-> ```TS
-> let x? = 44;
-> let y = x ?? 55;
-> ```
+* `x: T?` marks `x` as "optional of T". This means that `x` can store a value of type `T` but it can also be in a state where there is no value stored there.
+* To test if `x` has a value or not, use `x?`, which returns a `bool`. If x has a value, `x?` will be `true`, otherwise it will be `false`.
+* The `if let y = x { } else { }` is a special if statement that binds `y` inside the block only if `x` includes a value. Otherwise, the `else` block will be executed.
+* The `x?.y` notation can be used to access object fields only if there is a value. It returns an undefined optional if the value is not defined.
```
Chriscbr, line 885:
```suggestion
* The `x?.y` notation can be used to access object fields only if there is a value. It returns an empty optional if the value is not defined.
```
(simpler language)

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+It is also possible to use `=` in the field declaration to indicate the default value (similar to function arguments):
+
+```js
+struct FormatOpts {
+  radix: num = 10;
+}
+
+let opts = FormatOpts {};
+assert(opts.radix == 10);
+```
```
Chriscbr, line 920: We should clarify that a value can be omitted from a struct literal if the field is optional type OR if the field has a default value associated with it.

It would help if we could make it clearer on whether `radix` is of type `num` or `num?` in the `FormatOpts` example.

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
-> ```TS
-> let x? = 44;
-> let y = x ?? 55;
-> ```
+* `x: T?` marks `x` as "optional of T". This means that `x` can store a value of type `T` but it can also be in a state where there is no value stored there.
+* To test if `x` has a value or not, use `x?`, which returns a `bool`. If x has a value, `x?` will be `true`, otherwise it will be `false`.
+* The `if let y = x { } else { }` is a special if statement that binds `y` inside the block only if `x` includes a value. Otherwise, the `else` block will be executed.
+* The `x?.y` notation can be used to access object fields only if there is a value. It returns an undefined optional if the value is not defined.
+* The `x ?? y` notation will return the value in `x` (if there is one) or `y` otherwise (`y` must be of type `T`, not `T?`).
+* A default value can also be supplied using `= y` in the declaration of the field or argument, which implies that the type will always be defined.
```
Chriscbr, line 887: This phrasing is confusing to me -- every field in a struct always has a type defined. That seems like a vacuous statement.

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+let opts = FormatOpts {};
+assert(opts.radix == 10);
+```
+
+##### 1.7.1.2 Declaration in variables
+
+Use `T?` to indicate that a variable is optional:
+
+```js
+let var y: num?;
```
Chriscbr, line 927: This is a violation of another section of the spec -- see the section on declarations. Every `let` statement must have `= <value>` clause.

(I'm totally happy with changing this or saying optionals are an exception, just trying to keep us honest 🙂)

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+```
+
+Non-optional arguments can only be used before all optional arguments:
+
+```js
+let my_fun = (a: str, x?: num, y: str): void = { /* ... */ };
+//-----------------------------^^^^^^ ERROR: cannot declare a non-optional argument after an optional
+```
+
+If a function uses a keyword argument struct as the last argument, and there are other optional arguments before,
```
Chriscbr, line 991: (nit) curious what "keyword argument" means here

docs/04-reference/winglang-spec.md
Chriscbr, line 887: This introduction is awesome (I appreciate the bulleted summary) - though I think it's missing one critical piece, which is an example that demonstrates why nullity is modeled in the language in the first place. Namely, by shifting null-pointer-exceptions from being a runtime error to a compile time error instead:

```
let increment = (x: num, by: num?): num => {
  return x + by; // Error: "by" is `num?` (optional type), it must be unwrapped first
};
```

Chriscbr (approved):
Cool to see how this spec has evolved! 👍

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+
+If a function returns an optional type, use the `return nil;` statement to indicate that the value is
+not defined.
+
+> This is currently the only scenario in the language where `nil` is allowed.
+
+```js
+struct Name { first: str, last: str };
+
+let try_parse_name = (full_name: str): Name? => {
```
hasanaburayyan, line 1010: Is it valid to do something like?
```js
if (try_parse_name(some_str)) {
	// do stuff
}
```
or would it need to be
```js
if (try_parse_name(some_str)?) {
	// do stuff
}
```

hasanaburayyan (approved):
Seeya latter null!  😁

docs/04-reference/winglang-spec.md
skinny85: The proposal has two very similar syntaxes for using a value in a boolean context that are slightly different.

docs/04-reference/winglang-spec.md
skinny85: Yeah, `try_at()` is a great name for that.

docs/04-reference/winglang-spec.md
eladb: It makes sense to me that `let var` of an optional s won't require an initial value. There are many use cases for this. Let me check.

docs/04-reference/winglang-spec.md
skinny85: How about:

```js
if let address = my_person.address? {
  // stuff...
```

?

skinny85: Would differentiate it from just a "regular" `let address = my_person.address`.

skinny85: I think there's already some confusion about the two syntaxes: https://github.com/winglang/wing/pull/1495/files#r1134886913.

skinny85 (approved):
Looks great to me, with the only comment that I would love the `let` syntax to be consistent with the `?` syntax (https://github.com/winglang/wing/pull/1495#discussion_r1102328509).

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
-> ```TS
-> let x? = 44;
-> let y = x ?? 55;
-> ```
+* `x: T?` marks `x` as "optional of T". This means that `x` can store a value of type `T` but it can also be in a state where there is no value stored there.
+* To test if `x` has a value or not, use `x?`, which returns a `bool`. If x has a value, `x?` will be `true`, otherwise it will be `false`.
+* The `if let y = x { } else { }` is a special if statement that binds `y` inside the block only if `x` includes a value. Otherwise, the `else` block will be executed.
+* The `x?.y` notation can be used to access object fields only if there is a value. It returns an empty optional if the value is not defined.
+* The `x ?? y` notation will return the value in `x` (if there is one) or `y` otherwise (`y` must be of type `T`, not `T?`).
+* A default value can also be supplied using `= y` in the declaration of the field or argument, which implies that the type will always be defined.
```
yoav-steinberg, line 887: by
> using `= y`

Do you mean that these are valid:
```wing
let x: num? = 7;
let var y: num? = 7; // `y` is `var` so it might not always be defined.
```
Also if the optional variable is marked `var` then this might not be true:
> which implies that the type will always be defined

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
 
-</details>
+```js
+struct Person {
+  name: str;
+  address: str?;
+}
+```
+
+In the above struct, the `address` field, the `foo` argument and the `my_opt` field are all marked
```
yoav-steinberg, line 900: where are `foo` and `my_opt`? I can't see them anywhere "above".

docs/04-reference/winglang-spec.md
yoav-steinberg: I don't like this implicit nullness. I believe that very early on we'll have a case where I just want to say that there's nothing inside an optional and we'll need some syntax for that. I'm for not using the words null/nil/undefined just so it's clear. Good candidates are none/nothing/empty.

There's also this issue:
`let func = (x: num?) => { do something };`
does the argument `x` have a default "empty" value or is it a regular positional argument?

Chriscbr: > There's also this issue: `let func = (x: num?) => { do something };` does the argument `x` have a default "empty" value or is it a regular positional argument?

My current mental model of how this will work in the language/compiler is `x` will be type `num?` (that is, when you try using it inside the function, you need unwrap the value in some way) -- but a `num?` with a default value of nil/empty. Thus it's possible to call func() and it will still work.

yoav-steinberg: > Thus it's possible to call func() and it will still work.

But what about a `let func = (x: num?, y:num?) => { do something };`?
When I call `func(7)` which of the two should I assign the `7` to?
I think default values only make sense for named (non-positional) argument.

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+```js
+let address: str = my_person.address ?? "Planet Earth";
+```
+
+#### 1.7.5 Optional chaining using `?.`
+
+The `?.` syntax can be used for optional chaining. Optional chaining returns a value of type `T?` which must be unwrapped in order to be used.
+
+```js
+if let ip_address = options.networking?.ip_address {
```
yoav-steinberg, line 1078: To be clear about what's happening here (correct me where I'm wrong):
`networking` is of type `T1?`
`ip_address` is of type `T2`
The expression `options.networking` is a `T1?`
The expression `options.networking?` is a `T1`
The expression `options.networking?.ip_address` is a `T2?`
so `if let ip_address = xxx` is unwrapping a `T2?` into `ip_address`.
So if I get this correctly given a `T1?.T2?.T3.T4` we get a `T4?`.

Also it's a bit confusing that `?` is used both for optional chaining and for evaluating to `bool`:
<code>T1?.T2; // This is a `T2?`</code>
<code>T1?; // This is a `bool`</code>
<code>T1?.T2?; // This is a `bool`</code>
I guess we can try this a bit and see how it feels...

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+  print("the ip address is defined and it is: ${ip_address}");
+}
+```
+
+#### 1.7.6 Roadmap
+
+In the future we will consider the following additional sugar syntax:
+
+* `x ?? throw("message")` to unwrap `x` or throw if `x` is not defined.
+* `x ??= value` returns `x` or assigns a value to it and returns it to support lazy evaluation/memoization (inspired by [Nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)).
```
yoav-steinberg, line 1088: Other things I'd like:
* multiple `if lets` in a single expression. Can be initially supported only with `&&` operator:
```wing
if let x = x && let y = y { .. }
```
* Support `&&` operator in `if let`s:
```wing
if its_raining && let x = x { .. }
```
* Support implicit naming of unwrapped variable for `if let`s:
```wing
if let x { print("unwrapped x: ${x}"); }
```

Chriscbr: > * Support implicit naming of unwrapped variable for `if let`s:
> 
> ```
> if let x { print("unwrapped x: ${x}"); }
> ```

I think this is handled by the other syntax? I could be mistaken

```js
if let address = maybe_address {
  print("unwrapped address: ${address}");
}
```

yoav-steinberg: Imaging this case:
```wing
if let address = address { print("${address}"}; }
// We can have sugar here so this will be the same as:
if let address { print("${address}"}; }
// Or we can make the syntax a bit different to make this case clearer
if unwrap address { print("${address}"}; }
```

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+let address = my_person.address else {
+  throw("address is required");
+};
+
+print("address is ${address}");
+```
+
+#### 1.7.4 Unwrapping or default value using `??`
+
+The `??` operator can be used to unwrap or provide a default value. This returns a value of `T` that can safely be used.
```
yoav-steinberg, line 1067: Note that in TypeScript the type inference for `a ?? b` tries to find the lowest common ancestor between `a` and `b`'s types. So if `a` is `TSuper` and `b` is `TSub` (and `TSub` extends `TSuper`) then the type of the expression will be `TSuper`.
If there's no common ancestor it'll evaluate to a union type of `a` and `b`'s types.
I think in wing we can try to evaluate to the common ancestor, but fail in other cases.

docs/04-reference/winglang-spec.md
yoav-steinberg, line 1088: Another future thing to support is `??` for different types if they have a common ancestor (and also thing of interfaces).

docs/04-reference/winglang-spec.md
Chriscbr: I think this is similar to how default arguments work in TypeScript and Python (correct me if I'm wrong)

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
 
-> ```TS
-> let x? = 44;
-> let y = x ?? 55;
-> ```
+* `x: T?` marks `x` as "optional of T". This means that `x` can store a value of type `T` but it can also be in a state where there is no value stored there.
+* To test if `x` has a value or not, use `x?`, which returns a `bool`. If x has a value, `x?` will be `true`, otherwise it will be `false`.
+* The `if let y = x { } else { }` is a special if statement that binds `y` inside the block only if `x` includes a value. Otherwise, the `else` block will be executed.
+* The `x?.y` notation can be used to access object fields only if there is a value. It returns an empty optional if the value is not defined.
+* The `x ?? y` notation will return the value in `x` (if there is one) or `y` otherwise (`y` must be of type `T`, not `T?`).
```
Chriscbr, line 886: I don't think we need to limit `y` to be of type `T`. Chaining `??` together is a pretty normal/valid use case. Take this example:

```
struct Config {
  path: str?
}

fn path_from_env(): str? {
  return std.env("MY_APP_PATH");
}

let c = Config { ... };

let path = c.path ?? path_from_env() ?? "/hardcoded_path";
```

First, it tries to get the path from the struct. If that option is empty, it falls back to a path from an environment variable. If that environment variable isn't set, then it falls back to a hardcoded path.

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
-variable by always providing a default, in case the variable is nil. This forces
-a value to be present for the l-value (left hand side of the assignment operator
-and guarantees what's returned is type-stripped from its `?` keyword).
+Here's a quick summary of how optionality works in Wing:
 
-> ```TS
-> let x? = 44;
-> let y = x ?? 55;
-> ```
+* `x: T?` marks `x` as "optional of T". This means that `x` can store a value of type `T` but it can also be in a state where there is no value stored there.
```
ekeren, line 882: I am trying to understand what is the idiom for lazy instantiation in the current version
```ts (wing) 
resource {
  _hashcode: str? 
  _compute_hashcode() : str {
  } 

  hashcode(): str { 
    if let var h = this._hashcode {
      return h;
    } else {
      h = this.calculate_hashcode();
      this._hashcode = h;
      return h;
    } 
  }
} 
```
The above should work, but it is very verbose

staycoolcall911 (commented):
I added to the PR description: "Fixes #224".
Just a few more comments/suggestions:

docs/04-reference/winglang-spec.md
eladb: Alternatively with `let/else` (note that the field needs to be marked `var`):

```js
resource Foo {
  var _hashcode: str?;

  hashcode() {
    let hc = this._hashcode else {
      this._hashcode = this._compute_hashcode();
      return this.hashcode();
    };

    return hc;
  }
}
```

Eventually, this is why we need `??=`:

```js
resource Foo {
  var _hashcode: str?;
  
  hashcode() {
    return this._hashcode ??= this._compute();
  }
}
```

docs/04-reference/winglang-spec.md
eladb: https://github.com/winglang/wing/issues/1868

docs/04-reference/winglang-spec.md
eladb: Fixed. Sorry.

docs/04-reference/winglang-spec.md
eladb: Let's try this out and see how it feels in real life. Resolving for now.

docs/04-reference/winglang-spec.md
eladb: Done

docs/04-reference/winglang-spec.md
eladb: OK, the last expression in the chain must be of type `T`.

docs/04-reference/winglang-spec.md
eladb: Rephrased.

TIL "vacuous"

docs/04-reference/winglang-spec.md
eladb: No, in struct fields:

```js
struct Foo {
  age: num = 20;
}
```

And in function arguments:

```js
let print_age = (age: num = 88) => {
  print("my age: ${age}");
};
```

Added clarification.

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,234 @@ let f = (arg1: num, var arg2: num) => {
...
+
+  init(opt: num?) {
+    this.my_opt = opt;
+  }
+
+  set_my_var(x: str) {
+    this.my_var = x;
+  }
+}
+```
```
flyingImer, line 969: I was thinking about resetting:
```js
class Foo {
  var my_var1: num?;
  var my_var2: str?;

  set_my_var1(x: num?) {
    this.my_var1 = x;
  }

  set_my_var2(x: str) {
    this.my_var2 = x;
  }
  
  reset() {
    // my_var1 and my_var2 are supposed to be null again, but how?
  }
}
```
I guess I can do some dumb thing:
```js
class Foo {
  var my_var1: num?;
  var my_var2: str?;

  set_my_var1(x: num?) {
    this.my_var1 = x;
  }

  set_my_var2(x: str?) {
    this.my_var2 = x;
  }
  
  reset() {
    let new_foo = new Foo();
    this.my_var1 = new_foo.my_var1;
    this.my_var2 = new_foo.my_var2;
  }
}
```
But this definitely makes me feel I'm walking around some language limitations

flyingImer: Reading through the paragraphs below, I found I can do something a bit more nature:
```js
class Foo {
  var my_var1: num?;
  var my_var2: str?;

  set_my_var1(x: num?) {
    this.my_var1 = x;
  }

  set_my_var2(x: str?) {
    this.my_var2 = x;
  }

  reset_my_var1(): num? { // still weird to have return type, then I will need more code to do the right encapsulation
    return nil;
  }

  reset_my_var2(): str? {
    return nil;
  }
  
  reset() {
    this.my_var1 = this.reset_my_var1();
    this.my_var2 = this.reset_my_var2();
  }
}
```

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,234 @@ let f = (arg1: num, var arg2: num) => {
...
+// or
+let parse_int = (x: str, radix: num = 10, opts: ParseOpts = ParseOpts {}): num { /* ... */ };
+```
+
+##### 1.7.1.5 Declaration in function return types
+
+If a function returns an optional type, use the `return nil;` statement to indicate that the value is
+not defined.
+
+> This is currently the only scenario in the language where `nil` is allowed.
```
flyingImer, line 1016: I believe this is conflicted with an example in section 1.8
```
let i1? = nil;
```

eladb: Fixed

docs/04-reference/winglang-spec.md
eladb: It's when you can assign the fields of a struct through keyword arguments. Added another example.

docs/04-reference/winglang-spec.md
```diff
@@ -869,30 +869,223 @@ let f = (arg1: num, var arg2: num) => {
...
+//-----------------------------^^^^^^ ERROR: cannot declare a non-optional argument after an optional
+```
+
+If a function uses a keyword argument struct as the last argument, and there are other optional arguments before,
+it also has to be declared as optional:
+
+```js
+let parse_int = (x: str, radix: num?, opts?: ParseOpts): num { /* ... */ };
+// or
+let parse_int = (x: str, radix: num = 10, opts: ParseOpts = ParseOpts {}): num { /* ... */ };
```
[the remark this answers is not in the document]
eladb: Let's see how it feels in the real world.

docs/04-reference/winglang-spec.md
eladb: added

eladb: Leaving as is for now

docs/04-reference/winglang-spec.md
eladb: Added to roadmap

docs/04-reference/winglang-spec.md
eladb: Added `= nil` so now every let must have a value, even if it's optional.

eladb:
OK, finalized this.

PR #1496, opened 2023-02-09
chore: pin node/npm version for build workflow

### LINKED ISSUE #1479: Intermittent build issue in wing playground and wing cli

MarkMcCulloh:
Ok so during npm install symlinks were not being created where required. I believe this is some sort of regression with a new minor/patch version of node or npm. Using volta we already pin the node version locally, so it makes sense to do it in the workflow as well.

Removed any explicit node versioning where it is unnecessary.

Also combined all of the build into 1 command, because it shows the power of nx.

Fixes #1479 🤞 


.github/workflows/build.yml
```diff
@@ -100,10 +99,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Setup Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: ${{ env.NODEJS_VERSION }}
+      - name: Setup Node/NPM
```
Chriscbr, line 102:
```suggestion
      - name: Setup Node/NPM with fixed version
```

Chriscbr, line 102: Should we add this to the other jobs? Feels like we should probably use the same version of node/npm for everything in a given build if possible

Chriscbr (approved):
> during npm install symlinks were not being created where required  

Would enabling this flag make any difference maybe? https://docs.npmjs.com/cli/v8/commands/npm-install#install-links

.github/workflows/build.yml
MarkMcCulloh: By default the runner images are stable and come with stable node versions, for the purpose of the "Prepare" and "Publish" jobs I think that's pretty good. 

For E2E we intentionally pin it to different versions, no need to use volta for that since we aren't using the workspace version anyways.

MarkMcCulloh:
> Would enabling this flag make any difference maybe

[Funnily enough, I think that's part of the regression.](https://github.com/npm/cli/pull/6142) 

I had already added that option to the npmrc, I mistakenly thought that fixed the issue (npmrc doesn't apply to subdirs, whoops). 
Even though npm reverted the change, I will go ahead and set the flag as well here.

.github/workflows/build.yml
Chriscbr: Got it. I was just trying to think of ways to reduce environmental differences between different CI jobs, just like we try to reduce the number of differences between CI environments and dev environments. Maybe it would make more sense if at some point we combined all of the work into a single job or something, anyway future optimizations I guess 😅

PR #1499, opened 2023-02-10
fix: unclear compiler errors and lack of error snapshot testing

### LINKED ISSUE #926: Snapshot tests: invalid examples should also have snapshots of the error messages

MarkMcCulloh:
Ever since wingc stopped being a binary, calling `compile` doesn't cause an error unless there's a panic. This means that execution isn't stopped in the CLI for regular errors but the rest of the cli expects the preflight js to exist, causing it to throw trying to read the file.

To help verify this, added snapshot testing for the `invalid` tests to hangar.

Fixes #926


tools/hangar/src/cli.test.ts
```diff
@@ -144,15 +151,53 @@ test.each(validWingFiles)(
...
+
+    const stderrSanitized = stderr
+      // Remove absolute paths
+      .replaceAll(invalidTestDir.replaceAll("\\", "/"), "<TEST_DIR>")
+      // Normalize line endings
+      .replaceAll("\r\n", "\n")
+      // For the error about a missing semicolon, remove the char number
+      // because it's usually the last char of the line which is 
+      // different due to windows line endings (`\r\n`)
+      .replace(/:\d+ \| ';' expected./g, ":0 | ';' expected.");
```
MarkMcCulloh, line 196: As far as I could tell, this was the best of the worst ways to solve this.

Chriscbr: 🙈 

Maybe we could normalize the line endings in the compiler (before processing the source file(s)) instead?

tools/hangar/src/cli.test.ts
```diff
@@ -144,15 +151,53 @@ test.each(validWingFiles)(
...
+      path.join(invalidTestDir, wingFile),
+      args,
+      false
+    );
+
+    const stderr = out.stderr;
+
+    const stderrSanitized = stderr
+      // Remove absolute paths
+      .replaceAll(invalidTestDir.replaceAll("\\", "/"), "<TEST_DIR>")
```
Chriscbr, line 190: Could we just use relative paths instead? I feel like showing relative paths to your source files in compiler error messages by default is a better user experience since it reduces the amount of unimportant information on your screen

tools/hangar/src/__snapshots__/cli.test.ts.snap
```diff
@@ -4129,3 +4129,206 @@ constructor() {
...
+
+exports[`wing test sorted_errors_no_span.w (--target sim) - invalid > stderr 1`] = `
+"Compilation failed with 6 error(s)
+Error | Class c (at <TEST_DIR>/sorted_errors_no_span.w:3:7)'s parent \\"any\\" is not a class
+Error | Class e (at <TEST_DIR>/sorted_errors_no_span.w:7:7)'s parent \\"any\\" is not a class
+Error at <TEST_DIR>/sorted_errors_no_span.w:1:14 | Expected type to be \\"num\\", but got \\"str\\" instead
+Error at <TEST_DIR>/sorted_errors_no_span.w:3:17 | Expected b (at <TEST_DIR>/sorted_errors_no_span.w:3:17) to be a type but it's a variable
+Error at <TEST_DIR>/sorted_errors_no_span.w:6:14 | Expected type to be \\"num\\", but got \\"str\\" instead
+Error at <TEST_DIR>/sorted_errors_no_span.w:7:17 | Expected b (at <TEST_DIR>/sorted_errors_no_span.w:7:17) to be a type but it's a variable"
+`;
```
Chriscbr, line 4265: These snapshots look great! 🎉  Huge improvement for reviewing compiler changes

tools/hangar/src/cli.test.ts
MarkMcCulloh: I considered that but changing code in the compiler to account for this adds possible platform-specific behavior that doesn't actually help users and in fact may harm them (e.g. character position would actually be incorrect and not reflect reality)

Chriscbr (approved):

eladb (approved):

PR #1502, opened 2023-02-10
fix(compiler): errors do not show inflight in type name

### LINKED ISSUE #1501: Confusing compiler error when inflight keyword is missing
### Community Note

### I tried this:

```
bring cloud;

new cloud.Function((name: str): str => {
    return "Hello ${name}";
});
```

### I expected this:

An error message telling me that the function I passed is not inflight

### Instead, this happened:

Expected type to be "fn(any): any", but got "fn(str): str" instead

### Component:

Compiler

### Wing Version:

0.5.23

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

We should show the word `inflight` as part of the type (i.e. `inflight fn(any): any`)

Chriscbr:
Fixes #1501

### DIFF
--- a/examples/tests/invalid/cloud_function_expects_inflight.w
+++ b/examples/tests/invalid/cloud_function_expects_inflight.w
@@ -1,0 +1,6 @@
+bring cloud;
+
+new cloud.Function((name: str): str => {
+    return "Hello ${name}";
+});
+// ^ Expected type to be "inflight (any): any", but got "preflight (str): str" instead
--- a/libs/wingc/src/ast.rs
+++ b/libs/wingc/src/ast.rs
@@ -101,21 +101,23 @@ impl Display for Type {
 			Type::Set(t) => write!(f, "Set<{}>", t),
 			Type::MutSet(t) => write!(f, "MutSet<{}>", t),
 			Type::FunctionSignature(sig) => {
-				write!(
-					f,
-					"fn({}): {}",
-					sig
-						.parameters
-						.iter()
-						.map(|a| format!("{}", a))
-						.collect::<Vec<String>>()
-						.join(", "),
-					if let Some(ret_val) = &sig.return_type {
-						format!("{}", ret_val)
-					} else {
-						"void".to_string()
-					}
-				)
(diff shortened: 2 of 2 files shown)


examples/tests/invalid/cloud_function_expects_inflight.w
```diff
@@ -0,0 +1,6 @@
+bring cloud;
+
+new cloud.Function((name: str): str => {
+    return "Hello ${name}";
+});
+// ^ Expected type to be "inflight (any): any", but got "preflight (str): str" instead
```
Chriscbr, line 6: I feel like `preflight (str): str` could be confusing (since isn't preflight the defaul" phase in a Wing file?)? So maybe the error should be:

```
Expected type to be "inflight (any): any", but got "(str): str" instead
```

But this opens the question of how do we indicate a phase-independent function? `anyflight (str): str` maybe?

Curious what other people think

Chriscbr: I'm tempted to say the overly-verbose `preflight (str): str` in diagnostics is better as a starting point since it reinforces the idea of preflight vs inflight

MarkMcCulloh: I think keeping `preflight` in the type is good, especially in the context of a diagnostic. It helps to make the difference stick out more. Plus, general consistency is nice.


For phase independent, I think excluding any keyword is ok but I think having one would be great too.

```
Expected type to be "independent (any): any", but got "preflight (str): str" instead
```

I mean it looks pretty verbose but I wouldn't call it confusing. Not sure about `independent` of course, we should probably throw around more options for that.

MarkMcCulloh (approved):

examples/tests/invalid/cloud_function_expects_inflight.w
Chriscbr: Let's try without the extra keyword and see how it goes

eladb (commented):
Nice

PR #1508, opened 2023-02-12
fix(sdk): reduce default lambda timeout to 30 seconds

### LINKED ISSUE #1507: InvalidParameterValueException: Queue visibility timeout: 30 seconds is less than Function timeout: 60 seconds
### Community Note

### I tried this:

I tried the hello-world example. I attempted to deploy using Terraform (`terraform apply`), and I got this error:

```
[...]
│ Error: creating Lambda Event Source Mapping (arn:aws:sqs:us-east-1:9386XXXXXXX:cloud-Queue-c86e03d8): InvalidParameterValueException: Queue visibility timeout: 30 seconds is less than Function timeout: 60 seconds
│ {
│   RespMetadata: {
│     StatusCode: 400,
│     RequestID: "fa7dcc28-e2a6-4ac2-8606-e9fb37730353"
│   },
│   Message_: "Queue visibility timeout: 30 seconds is less than Function timeout: 60 seconds",
│   Type: "User"
│ }
│ 
│   with aws_lambda_event_source_mapping.root_cloudQueue_EventSourceMapping_A2041279,
│   on main.tf.json line 64, in resource.aws_lambda_event_source_mapping.root_cloudQueue_EventSourceMapping_A2041279:
│   64:       }
```

### I expected this:

The SQS queue to have a correct value for [`visibility_timeout_seconds`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue#visibility_timeout_seconds)

### Instead, this happened:

See error above.

### Component:

Other

### Wing Version:

0.5.26

### Node.js Version:

v18.14.0

### Platform(s):

MacOS

### Anything else?

_No response_

staycoolcall911:
p0 - hello.w terraform plan doesn't work.

The queue visibility timeout is somehow set by default to 30 seconds, while the lambda timeout is configured to 60 seconds by us.
To fix this as quickly as possible we simply reduce the lambda timeout to 30 seconds as well. In a later PR we will increase both to 60 seconds by default.

Fixes #1507  


ShaiBer (approved):

eladb:
Curious- why is this breaking the example? I'd expect the lambda to handle the message much faster than 30 seconds....

staycoolcall911:
> Curious- why is this breaking the example? I'd expect the lambda to handle the message much faster than 30 seconds....
https://stackoverflow.com/questions/28713366/what-does-visibility-timeout-mean-for-aws-sqs

eladb:
> > Curious- why is this breaking the example? I'd expect the lambda to handle the message much faster than 30 seconds....
> 
> https://stackoverflow.com/questions/28713366/what-does-visibility-timeout-mean-for-aws-sqs

The timeout only kicks in if the message wasn't handled after the allotted time. The fact that the queue's visibility timeout is 30s and lambda is 60s doesn't mean that it actually takes the function over 30 seconds to handle the batch.

I can't imagine that `hello.w` takes over 30 seconds to handle the message.

I am curious if there's something we don't fully understand or perhaps we have a bug in our SQS=>Lambda connection.

staycoolcall911:
Updated the PR description to reflect the fact that this was an error received on `terraform apply`.

eladb:
We should create an issue to add this check in the cloud.Function layer during compilation. 

This is not aws-specific

staycoolcall911:
> We should create an issue to add this check in the cloud.Function layer during compilation.
> 
> This is not aws-specific

#1526

PR #1520, opened 2023-02-13
fix(compiler): errors do not show inflight in type name (attempt 2)

### LINKED ISSUE #1501: Confusing compiler error when inflight keyword is missing

Chriscbr:
Fixes #1501

### DIFF
--- a/libs/wingc/src/ast.rs
+++ b/libs/wingc/src/ast.rs
@@ -79,18 +79,18 @@ impl Display for Phase {
 }
 
 #[derive(Debug, Clone)]
-pub enum Type {
+pub enum TypeAnnotation {
 	Number,
 	String,
 	Bool,
 	Duration,
-	Optional(Box<Type>),
-	Array(Box<Type>),
-	MutArray(Box<Type>),
-	Map(Box<Type>),
-	MutMap(Box<Type>),
-	Set(Box<Type>),
-	MutSet(Box<Type>),
+	Optional(Box<TypeAnnotation>),
+	Array(Box<TypeAnnotation>),
+	MutArray(Box<TypeAnnotation>),
+	Map(Box<TypeAnnotation>),
+	MutMap(Box<TypeAnnotation>),
+	Set(Box<TypeAnnotation>),
+	MutSet(Box<TypeAnnotation>),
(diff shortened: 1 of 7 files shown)


Chriscbr:
There's a bit of duplication because we have a `Type` enum in the `ast` module and a `Type` enum in the `type_checker` module. (I changed the `Display` trait on only one of them, hence why the bug wasn't fixed). This doesn't seem ideal, but I couldn't think of a clean way to avoid the duplication. do you have thoughts on this @yoav-steinberg?

libs/wingc/src/ast.rs
```diff
@@ -117,32 +117,36 @@ impl Display for Type {
...
-					.collect::<Vec<String>>()
-					.join(", ");
-				let ret_type_str = if let Some(ret_val) = &sig.return_type {
-					format!("{}", ret_val)
-				} else {
-					"void".to_string()
-				};
-				write!(f, "{phase_str}({params_str}): {ret_type_str}",)
-			}
+			Type::FunctionSignature(sig) => write!(f, "{}", sig),
```
MarkMcCulloh, line 120: I don't feel like this is duplication, our naming just makes it seem like it.
One of these corresponds to the AST node for a function type annotation, the other is our representation of a function type in our type system. 

The `Type` used in this file should probably be renamed to `TypeAnnotation` or something to help with that.

MarkMcCulloh (approved):
Nice fix, just noticed the phase was missing in the type text a minute ago too 🎉

libs/wingc/src/ast.rs
Chriscbr: Ah great idea, I like `TypeAnnotation` 👍

PR #1521, opened 2023-02-13
fix: unable to reference immutable data in inflight methods

### LINKED ISSUE #1519: Unable to reference data fields in inflight methods of user-defined resources
### Community Note

### I tried this:

```js
bring cloud;

resource Foo {
  s: str;
  init() { 
    this.s = "hello";
  }
  inflight test() {
    assert(this.s == "hello");
  }
}

let f = new Foo();
new cloud.Function(inflight () => { f.test(); }) as "test";
```

### I expected this:

Test pass

### Instead, this happened:

```
Error: assertion failed: '(this.s === "hello")'
```

### Component:

Compiler

### Wing Version:

0.5.28

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

eladb:
Include primitives and collections when identifying preflight fields referenced by inflight methods. Then, for each field, instead of calling `._toInflight()` directly, call a new protected internal method on `Resource` called `._lift(x)` which reuses `serializeImmutableData` to lift the object into an inflight context. Yes, lift...

The test `user-defined-resources-captures` verifies that all types of capturable types can be referenced from inflight methods.

Fixes #1519


libs/wingc/src/type_check.rs
```diff
@@ -515,6 +515,21 @@ impl TypeRef {
...
+			Type::Number => true,
+			Type::String => true,
+			Type::Duration => true,
+			Type::Boolean => true,
+			Type::Array(v) => v.is_capturable(),
+			Type::Map(v) => v.is_capturable(),
+			Type::Set(v) => v.is_capturable(),
+			_ => false,
+		}
+	}
```
Chriscbr, line 532: 👨‍🍳 👌

libs/wingc/src/jsify.rs
```diff
@@ -718,9 +718,9 @@ impl<'a> JSifier<'a> {
 	/// 	}
 	/// }
 	/// _toInflight() {
-	/// 	const b_client = this.b._toInflight();
+	/// 	const b_client = this._lift(b);
```
Chriscbr, line 721: I wonder if we'll at some point want `lift` to be some kind of method on the inflight host (not the resource). For example, a `cloud.Function` would be responsible for "lifting" a value into a JS string, and a `cloud.Workflow` could "lift" a value into some state machine language value or AST representation, etc.

(for future work)

flyingImer:
this is 🚀, just hit this rough edge and saw this pr!

Chriscbr (approved):
Pretty elegant approach 👍

examples/tests/valid/user-defined-resources-captures.w
```diff
@@ -0,0 +1,73 @@
...
+resource MyResource {
+  my_resource: cloud.Bucket;
+  my_str: str;
+  my_num: num;
+  my_bool: bool;
+  array_of_str: Array<str>;
+  map_of_num: Map<num>;
+  set_of_str: Set<str>;
+
+  array_of_queues: Array<cloud.Queue>;
```
yoav-steinberg, line 12: Do we also want to add a test for `Array<UserDefinedResource>`?
Also what about:
```wing
struct MyStruct {
  b: cloud.Bucket;
  n: num;
}
```
And testing `Array<MyStruct>`?

libs/wingc/src/jsify.rs
yoav-steinberg: Interesting concept, but I think as long as out inflights are in a functions/imperitive language (and not a state machine) it makes sense the way it is now.

libs/wingc/src/type_check.rs
```diff
@@ -515,6 +515,21 @@ impl TypeRef {
...
+		match **self {
+			Type::Resource(_) => true,
+			Type::Enum(_) => true,
+			Type::Number => true,
+			Type::String => true,
+			Type::Duration => true,
+			Type::Boolean => true,
+			Type::Array(v) => v.is_capturable(),
+			Type::Map(v) => v.is_capturable(),
+			Type::Set(v) => v.is_capturable(),
```
yoav-steinberg, line 529: `Type::Struct`?

yoav-steinberg (approved):
Amazing!

libs/wingsdk/src/core/resource.ts
```diff
@@ -203,6 +204,19 @@ export abstract class Resource
...
+  /**
+   * "Lifts" a value into an inflight context. If the value is a resource (i.e. has a `_toInflight`
+   * method), this method will be called and the result will be returned. Otherwise, the value is
+   * returned as-is.
+   *
+   * @param value The value to lift.
+   * @returns a string representation of the value in an inflight context.
+   * @internal
+   */
+  protected _lift(value: any): string {
```
yoav-steinberg, line 217: 🛫🛫🛫

staycoolcall911 (commented):
Looks like some of the hangar snapshot tests are failing.
You should probably `npm i` (from root folder) before running hangar to update the snapshots (`npx nx hangar test`).

libs/wingc/src/jsify.rs
eladb: yes, we'll get there! For now, let's say that the resource "lifts" the value into an inflight context.

libs/wingc/src/type_check.rs
eladb: I can add that but can't really test it yet I think

eladb: added

examples/tests/valid/user-defined-resources-captures.w
yoav-steinberg: We do however have struct literals, I think this should work:
```
array_of_structs: Array<cloud.QueueProps> [
  cloud.QueueProps { timeout: 60s, initialMessages: ["Chicks", "with", "bricks", "come"] },
  cloud.QueueProps { timeout: 120s, initialMessages: ["Chicks", "with", "blocks", "come"] }
];

Issue #1468, opened 2023-02-07
Use pascal case for all primitive types


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

This is a proposal to make all types in Wing use pascal-case, so primitive types have a similar look & feel like container and user-defined types:

* `str` => `Str`
* `num` => `Num`
* `bool` => `Bool`
* `duration` => `Duration`

### Use Cases

To reduce cognitive load, all types should have a similar look & feel.

As an example, see the [spec for `Json`](#1458), which proposes the all built-in types will have a `from_json()` method.

Consider this:

```js
Str.from_json(j);
Array<num>.from_json(j);
MyStruct.from_json(j);
```

Versus this:

```js
str.from_json(j);
Array<num>.from_json(j);
MyStruct.from_json(j);
```

The `str.from_json` version feels a bit off.

### Implementation Notes

I think we should just try to rename to pascal case. If we see a lot of a resistance from users,  we can consider simply adding aliases `str` => `Str` (but then we still have `Str.from_json()`).

### Component

Language Design

MarkMcCulloh:
I disagree with the cognitive load argument for the case change, I don't think that's giving enough credit to developers who see that maybe these types are different (cause they are).

If it this ends up happening, then I would say the real cognitive load is that they're abbreviated. I would be more in favor of this (with aliasing):

```
str <=> String
num <=> Number
bool <=> Bool
duration <=> Duration
```

eladb:
@MarkMcCulloh I like your suggestion. Yes, after submitting this issue I had a weird gut feeling that it's the wrong direction...

eladb:
@MarkMcCulloh lmk if you this makes you happy.

Chriscbr:
By aliasing do we mean these are both valid?

```
let fn = (message: str) => { `Hello, ${message}` };
let fn = (message: String) => { `Hello, ${message}` };
```

(Maybe we can also offer `~` as an alias for `inflight`? 😝)

I'm not sure I am convinced of the strong need/benefit for aliases here

MarkMcCulloh:
I actually don't want aliases either, but I also dislike the pascal casing for primitives. So if we decide to do Pascal casing anyways, I'd love to be able to avoid it with aliases.

eladb:
[...]

PR #1524, opened 2023-02-14
feat(compiler): `json` support

hasanaburayyan:
Current changes support:
- Json literals
- prints
- `get` and `set`
- from native types support. 

Not supported:
- [Static methods](https://github.com/winglang/wing/issues/1679)
- [dot and [] access notation](https://github.com/winglang/wing/issues/1680)


libs/tree-sitter-wing/grammar.js
```diff
@@ -502,6 +503,23 @@ module.exports = grammar({
...
+      $.json_array_literal,
+      $.string,
+      $.number,
+      $.bool
+    ),
+    json_array_literal: ($) => seq(
+     "[", commaSep(field("element", $.expression)), "]"
+    ),
+    json_map_literal: ($) => seq(
+     "{", commaSep(field("member", $.map_literal_member)), "}"
```
hasanaburayyan, line 521: The problem with this is that nested maps within Json will not be Json map literals and not persist the heterogenous behavior. 

I could keep going and make a json_map_literal_member but then I feel like Im basically copying the whole grammar into json versions

hasanaburayyan:
@yoav-steinberg @Chriscbr @MarkMcCulloh wanted to see about getting some discussion going here on Json and JsonMut literals. I find myself wanting to add a lot more to the grammar but maybe there is a way to go about this differently, so this PR has my initial crazy ideas.

libs/tree-sitter-wing/grammar.js
staycoolcall911: I hope I understand you correctly, but if you need to reuse logic from map or any other element of the grammar, you can easily do so, since this is javascript. See for example the `commaSep()` function at the bottom of the file.

To get help, you can schedule some time with @MarkMcCulloh, our master of grammar 😄

staycoolcall911 (commented):
Please rename JsonMut to MutJson across this PR.

examples/tests/valid/json.w
```diff
@@ -0,0 +1,22 @@
+let json_string  = Json "hello";
+let json_number  = Json 123;
+let json_bool    = Json true;
+let json_array   = Json [ 1, 2, 3 ];
+let json_obj     = Json { boom: 123 };
+
+let x = {
+  y: "Hello"
+};
+
```
eladb, line 10: missing some `assert`s for converting values from `Json` to strong types.

examples/tests/valid/json.w
```diff
@@ -0,0 +1,22 @@
...
+  hello: x.y, 
+  world: [ 1, "cat", 3 ],
+  "boom boom": { 
+    hello: {
+      world: "cool"
+    },
+    cool: [1, 2, "c"]
+  }
+};
+
```
eladb, line 21: `assert`s for mutation?

libs/tree-sitter-wing/grammar.js
MarkMcCulloh: As unfortunate as it is, I think we may have to deal with creating json-specific grammar nodes. Even though most of the json stuff is super similar to existing rules it's not exactly the same. 
I don't think our grammar will makes it feasible to allow the rest of our `expressions` inside json literals (due to the inconsistency) so it'll have to be limited.

I think it would work to have:
- json_element (any of the following)
  - json_primitive (`string|number|boolean|identifier`)
  - json_map (`{identifier|string: json_element}`)
  - json_array (`[json_element]`)

examples/tests/valid/json.w
```diff
@@ -0,0 +1,24 @@
+let json_string  = Json "hello";
```
eladb, line 1: As part of #1468 we should create an alias for this `json`, which will be beautiful to signal the fact that we think about it as a primitive type in our little cloud language.

libs/tree-sitter-wing/grammar.js
```diff
@@ -515,7 +516,44 @@ module.exports = grammar({
...
+    ),
+    json_map_literal: ($) => seq(
+      optional(field("type", $.json_container_type)),
+     "{", commaSep(field("member", $.json_map_literal_member)), "}"
+    ),
+    json_map_literal_member: ($) => seq(
+      optional(field("type", $.json_container_type)),
+      seq(choice($.identifier, $.string), ":", $._json_value)
+    ),
+    json_reference: ($) => choice($.identifier, $.json_nested_identifier),
```
eladb, line 548: please add some comments to explain the new grammar rules.

libs/wingc/src/ast.rs
```diff
@@ -84,6 +84,8 @@ pub enum TypeAnnotation {
 	String,
 	Bool,
 	Duration,
+  MutJson,
```
eladb, line 87: Maybe you need to use tabs instead of spaces? For some reason our rust code uses tabs... (@MarkMcCulloh, should we normalize this across our repo?

libs/wingc/src/capture.rs
```diff
@@ -362,6 +362,19 @@ fn scan_captures_in_expression(
 				res.extend(scan_captures_in_expression(&v, env, statement_idx, diagnostics));
 			}
 		}
+    ExprKind::JsonLiteral {expression, type_: _ } => {
```
eladb, line 365: do we have tests for capture references inside json values?

libs/wingc/src/ast.rs
hasanaburayyan: 😅 yea I noticed this when I pushed, I thought my VS code was smart enough to detect what was going on in the file.

libs/wingc/src/jsify.rs
```diff
@@ -388,7 +390,7 @@ impl<'a> JSifier<'a> {
 					self.jsify_expression(rexp, phase)
 				)
 			}
-			ExprKind::ArrayLiteral { items, .. } => {
+			ExprKind::ArrayLiteral { items, .. } | ExprKind::JsonArrayLiteral { items } => {
```
eladb, line 393: add a comment explaining this case.

libs/wingc/src/jsify.rs
```diff
@@ -411,15 +416,31 @@ impl<'a> JSifier<'a> {
...
 			}
+      ExprKind::JsonMapLiteral { fields, type_ } => {
+        let f = fields
+          .iter()
+          .map(|(key, expr)| format!("[ \"{}\", {} ]", key, self.jsify_expression(expr, phase)))
+          .collect::<Vec<String>>()
+          .join(", ");
+        
+        match type_ {
+          TypeAnnotation::MutJson => format!("new Map([{}])", f),
```
eladb, line 427: Why are we using a `Map` here as the underlying struct instead of a plain-old javascript object?

libs/wingc/src/ast.rs
MarkMcCulloh: We are already forcing tabs if you run rustfmt (https://github.com/winglang/wing/blob/main/.rustfmt.toml)

eladb: > Maybe you need to use tabs instead of spaces? For some reason our rust code uses tabs... (@MarkMcCulloh, should we normalize this across our repo?

@MarkMcCulloh ☝️

libs/tree-sitter-wing/grammar.js
```diff
@@ -414,7 +415,7 @@ module.exports = grammar({
             "Array",
             "Set",
             "Map",
-            "Promise",
+            "Promise"
```
yoav-steinberg, line 418: Why do this change?
```suggestion
            "Promise",
```

libs/tree-sitter-wing/grammar.js
```diff
@@ -428,7 +429,7 @@ module.exports = grammar({
             choice(
               "MutSet",
               "MutMap",
-              "MutArray",
+              "MutArray"
```
yoav-steinberg, line 432:
```suggestion
              "MutArray",
```

libs/tree-sitter-wing/grammar.js
```diff
@@ -515,7 +516,44 @@ module.exports = grammar({
     await_expression: ($) => prec.right(seq("await", $.expression)),
     defer_expression: ($) => prec.right(seq("defer", $.expression)),
     parenthesized_expression: ($) => seq("(", $.expression, ")"),
+    
+    json_literal: ($) => choice(
+      seq(field("type", choice($.json_container_type)), field("element", $._json_value)),
```
yoav-steinberg, line 521: `$.json_container_type` is already wrapped in a `choice`, this is redundant.
```suggestion
      seq(field("type", $.json_container_type), field("element", $._json_value)),
```

libs/tree-sitter-wing/grammar.js
yoav-steinberg, line 548: I don't see why this is needed. Doesn't a json reference look the same as any reference. Is this `a.b.c` a json reference or some other reference? I would expect tree-sitter to fail generating the parser on ambiguous grammar.

libs/tree-sitter-wing/grammar.js
```diff
@@ -515,7 +516,44 @@ module.exports = grammar({
...
+    ),
+
+    json_container_type: ($) => choice(
+      "Json",
+      "MutJson"
+    ),
+
+    _json_value: ($) => choice(
```
yoav-steinberg, line 529: I don't think we need a `_json_value`. It seems like any place it's used is basically a place where we can have some expression that evaluates to a json type. So is json value just an `$.expression`.

You might be worried that if you define some `json { a: function_call(1) }` it'll be fine as long as `function_call(1)` returns a json type but it'll fail if it evaluates to a non-json type. In that case it's the **type-checker**'s job to give an error not the grammar/parser. Perhaps this is a source of confusion, or perhaps I don't understand something here.

libs/tree-sitter-wing/grammar.js
```diff
@@ -538,6 +539,20 @@ module.exports = grammar({
       seq($.identifier, ":", $.expression),
     structured_access_expression: ($) =>
       prec.right(seq($.expression, "[", $.expression, "]")),
+
+    json_literal: ($) => seq(field("type", $.json_container_type), field("element", $.json_element)),
+    
+    json_element: ($) => choice(
+      $._literal,
+      $.map_literal,
+      $.array_literal
```
yoav-steinberg, line 548: There are two things that `Json` literals have that "normal" literals don't which we should make sure we take care of:
1) I think json maps accept both `key: val` and `"key": val` format. This means that we need to change the grammar for `map_literal` to support the string variant (or is it even an `$.expression` variant)? Later on during the parsing phase we can report and error if someone tries to use it out of place.
2) `array_literal`s need to be uniformly typed when not a `Json` array. This needs to be verified in the type checker.

**1** above might introduce a world of complexity if we support map keys as expressions, so it might be worth starting without this, just string literals or simple identifiers.

libs/wingc/src/ast.rs
```diff
@@ -316,6 +320,10 @@ pub enum ExprKind {
 		// We're using an ordered map implementation to guarantee deterministic compiler output. See discussion: https://github.com/winglang/wing/discussions/887.
 		fields: BTreeMap<Symbol, Expr>,
 	},
+	JsonLiteral {
+		type_: TypeAnnotation,
```
yoav-steinberg, line 324: Do we need a `TypeAnnotation` here? I mean a JsonLiteral is always either a `Json` or a `MutJson`, maybe just a `is_mut: bool` is enough here. Will reduce more complex error checks down the line.

libs/wingc/src/jsify.rs
```diff
@@ -395,7 +404,7 @@ impl<'a> JSifier<'a> {
 					.collect::<Vec<String>>()
 					.join(", ");
 
-				if is_mutable_collection(expression) {
+				if is_mutable_collection(expression) || *self.is_json.borrow() {
```
yoav-steinberg, line 407: Why special treatment for json here? Shouldn't this be related to whether this is a `MutJson` or not and not generally a json?

libs/wingc/src/jsify.rs
```diff
@@ -411,15 +420,50 @@ impl<'a> JSifier<'a> {
...
-				format!(
-					"Object.freeze(new Map([{}]))",
-					fields
-						.iter()
-						.map(|(key, expr)| format!("[ \"{}\", {} ]", key, self.jsify_expression(expr, phase)))
-						.collect::<Vec<String>>()
-						.join(", ")
-				)
+			ExprKind::JsonLiteral { type_, element } => {
+				*self.is_json.borrow_mut() = true;
```
yoav-steinberg, line 424: maybe call this `in_json` instead?

libs/wingc/src/lib.rs
```diff
@@ -131,6 +131,7 @@ pub fn parse(source_path: &Path) -> (Scope, Diagnostics) {
 		source_name: source_path.to_str().unwrap().to_string(),
 		error_nodes: RefCell::new(HashSet::new()),
 		diagnostics: RefCell::new(Diagnostics::new()),
+		is_json: RefCell::new(false),
```
yoav-steinberg, line 134:
```suggestion
		in_json: RefCell::new(false),
```
maybe clearer (if you agree then probably relevant to all `is_json` fields as well)

libs/wingc/src/type_check.rs
```diff
@@ -2241,6 +2273,16 @@ impl<'a> TypeChecker<'a> {
...
+					Type::MutJson => {
+						todo!(); // TODO: Need to add SDK support for MutJson
+						 // let new_class = self.hydrate_class_type_arguments(env, WINGSDK_MUT_JSON, vec![self.types.json()]);
+						 // self.get_property_from_class(new_class.as_class().unwrap(),	property)
+					}
+					Type::Json => {
+						todo!(); // TODO: Need to add SDK support for Json
+						 // let new_class = self.hydrate_class_type_arguments(env, WINGSDK_JSON, vec![self.types.json()]);
+						 // self.get_property_from_class(new_class.as_class().unwrap(),	property)
+					}
```
yoav-steinberg, line 2285: I don't think our Json spec is related to generics in any way. So I don't understand why we need to "hydrate" here.

libs/wingc/src/type_check.rs
```diff
@@ -134,6 +134,8 @@ pub enum Type {
 	Duration,
 	Boolean,
 	Void,
+	Json,
+	MutJson,
```
yoav-steinberg, line 138: Now that we have these types we need to type check them when they appear in literals. Basically we need to verify that all fields within a `Json` literal are valid json types. So something like this won't work:
```
let b = new cloud.Bucket();
let arr = Json [b,b,b];
```
You'll probably need to add code for this in `type_check_exp`

libs/wingc/src/type_check.rs
```diff
@@ -1106,6 +1130,12 @@ impl<'a> TypeChecker<'a> {
 
 				container_type
 			}
+			ExprKind::JsonLiteral { type_, element } => {
+				*self.is_json.borrow_mut() = true;
+				let json_type = self.resolve_type_annotation(type_, env, statement_idx);
```
yoav-steinberg, line 1135: I think the `type_` here is always either `Json` or `MutJson` so there's no need to `resolve_type..` here.
this should ways return `self.types.[mut_]json()`.
What is missing here is some recursion into the literals elements and verifying they are of a valid json type.

yoav-steinberg (commented):
nice!
I think there's some work needed in the type_checking phase.

apps/wing/src/commands/compile.ts
```diff
@@ -131,6 +131,7 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
     __dirname: workDir,
     __filename: artifactPath,
     $plugins: resolvePluginPaths(options.plugins ?? []),
+    console,
```
eladb, line 134: add a comment explaining why this is needed

examples/tests/valid/json.w
```diff
@@ -0,0 +1,16 @@
...
+let json_number  = Json 123;
+let json_bool    = Json true;
+let json_array   = Json [ 1, 2, 3 ];
+let json_obj     = Json { boom: 123 };
+
+let json_mut_obj = Json {
```
staycoolcall911, line 7: Should it be `MutJson`?
```suggestion
let json_mut_obj = MutJson {
```

libs/wingc/src/jsify.rs
```diff
@@ -411,15 +420,50 @@ impl<'a> JSifier<'a> {
...
+							format!("Object.freeze({})", self.jsify_expression(element, phase))
+						}
+					}
+					_ => format!("{}", self.jsify_expression(element, phase)),
+				};
+				*self.is_json.borrow_mut() = false;
+				json_string
+			}
+			ExprKind::MapLiteral { fields, type_ } => {
+				if *self.is_json.borrow() {
```
staycoolcall911, line 439: Did I understand correctly that we are handling the jsification of json maps in this clause?
I'm a little confused if we should consider `Json` and `MutJson` in here too (then we'll need to check that and apply `Object.freeze()`).

libs/wingc/src/type_check.rs
staycoolcall911: @yoav-steinberg - it's for accessing properties.
We want to support `.get(propertyName)` (and `.set` too) - to do that we will use the "baby macros" mechanism.
Supporting dot notation will only come later.

staycoolcall911 (commented):
Looks great @hasanaburayyan - you're getting there...

libs/wingc/src/jsify.rs
hasanaburayyan: I believe since we only need to freeze the whole map or none of it the JsonLiteral ExprKind handles adding `Object.freeze()` then the sub maps dont need to be individually frozen. 

const x = Object.freeze({some: { deep: { object: {}}}});
or
const x = {some: { deep: { object: {}}}};

examples/tests/valid/json.w
hasanaburayyan: Yupp it should, I committed some testing junk 😅

libs/wingc/src/type_check.rs
hasanaburayyan: So actually if we just return something like
```rust
Type::MutJson => {
		VariableInfo {
              _type: self.types.mut_json(),
              flight: env.flight,
              reassignable: true 
           }
}
Type::Json => {
     VariableInfo { 
           _type: self.types.json(),
           flight: env.flight,
           reassignable: false 
	}
}
```

Then we have dot notation. Since the jsified Json uses JS objects rather than maps. But Im not sure about how to determine the proper `flight`

libs/wingc/src/jsify.rs
hasanaburayyan: So here was the thinking..  the JsonLiteral will take care of adding the `Object.freeze()` or not at the root level. This way json elements will inherit their mutability from the JsonLiteral.

so that
```js
let x = MutJson {
 	y: {
		z: [1, 2, 3]
	}
}

let a = MutJson {
 	b: {
		c: [1, 2, 3]
	}
}

// Should jsify to
const x = {"y":{"z":[1, 2, 3]}};
const a = Object.freeze({"b":{"c":[1, 2, 3]}});
```

libs/wingc/src/ast.rs
hasanaburayyan: You are right it was much simpler 😅

libs/wingc/src/type_check.rs
yoav-steinberg: Actually I think @staycoolcall911 is right here. `hydrate_class_type_arguments` isn't only for "baby generics" it's also for returning some std class from the sdk for a given "native" wing type. So your original commented code seems more-or-less correct. We just need to pass an empty `type_params` vector because there are not type params here (`Json` isn't generic).
If this is still unclear lets try to have a session about this..

libs/tree-sitter-wing/grammar.js
hasanaburayyan: For 1 it looks like we already do support both formats with normal map_literals

2. Yes agree, this has been using RefCell booleans

libs/wingc/src/type_check.rs
hasanaburayyan: hmm, yea right now the acceptable json types are pretty much everything, I want to trim this down but curious how that would look since we would have to allow for sub arrays within an array and without maybe like a serializable interface not sure how we could limit the allowed types of arrays. IE. `[ "a", 2, [b, b, b]]`

libs/wingsdk/src/std/json.ts
```diff
@@ -0,0 +1,49 @@
...
+  public get(key: string): Json {
+    key;
+    throw new Error("Macro");
+  }
+}
+
+/**
+ * Mutable Json
+ */
+export class MutJson {
```
eladb, line 22: What about accessing array elements? Maybe `getAt()` and `setAt()`?

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -174,12 +185,6 @@ exports[`try_no_catch_or_finally.w > wing test (--target sim) - invalid > stderr
 Error at ../../../../examples/tests/invalid/try_no_catch_or_finally.w:1:1 | Missing \`catch\` or \`finally\` blocks for this try statement"
 `;
 
-exports[`types_string_interpolation.w > wing test (--target sim) - invalid > stderr 1`] = `
```
eladb, line 177: Is it intentional that we lost these tests?

eladb (changes requested):
Amazing work! Nothing major to call out.

libs/wingc/src/jsify.rs
```diff
@@ -304,7 +310,14 @@ impl<'a> JSifier<'a> {
 						.iter()
 						.map(|p| match p {
 							InterpolatedStringPart::Static(l) => format!("{}", l),
-							InterpolatedStringPart::Expr(e) => format!("${{{}}}", self.jsify_expression(e, phase)),
+							InterpolatedStringPart::Expr(e) => {
+								match *e.evaluated_type.borrow().expect("Should have type") {
+									Type::Json | Type::MutJson => {
+										format!("${{JSON.stringify({}, null, 2)}}", self.jsify_expression(e, phase))
+									}
+									_ => format!("${{{}}}", self.jsify_expression(e, phase)),
```
staycoolcall911, line 318: @eladb - this is the line of code that added support for `bool` in string interpolation (and actually any expression), which is why @hasanaburayyan deleted `examples/tests/invalid/types_string_interpolation.w`

libs/wingc/src/ast.rs
```diff
@@ -84,6 +84,8 @@ pub enum TypeAnnotation {
 	String,
 	Bool,
```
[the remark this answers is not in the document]
MarkMcCulloh: Yes we are supposed to commit the generated parser src, this a convention for tree-sitter parsers

tools/hangar/src/__snapshots__/invalid.test.ts.snap
staycoolcall911: Yes - see [my comment](https://github.com/winglang/wing/pull/1524#discussion_r1119953551).

examples/tests/invalid/types_string_interpolation.w
```diff
@@ -1,7 +0,0 @@
-let bool_example = true;
-
-"invalid ${bool_example}";
-// ERR     ^^^^^^^^^^^^ expression must be a string or number
```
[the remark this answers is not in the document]
staycoolcall911: See [my comment](https://github.com/winglang/wing/pull/1524#discussion_r1119953551).

libs/wingc/src/jsify.rs
hasanaburayyan: Ah yea I meant to start a conversation about why we were not allowing prior. 😅

libs/wingsdk/src/std/json.ts
hasanaburayyan: 😅 index an element in an array! Who would need to do that? This is totally a use case I missed here, tunnel visioned on Maps. Ill add `getAt()` and `setAt()` apis to support this. Im sure this will evolve as `dot` and `[]` notations are added.

libs/wingc/src/jsify.rs
```diff
@@ -421,14 +435,30 @@ impl<'a> JSifier<'a> {
 						.join("\n")
 				)
 			}
+			ExprKind::JsonLiteral { is_mut, element } => {
+				self.set_json(true);
```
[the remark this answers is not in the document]
hasanaburayyan: I went ahead and refactored to use context objects

libs/wingsdk/src/std/json.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Returns a specified element at a given index from Json Array
+   *
+   * @macro ($self$)[$args$]
+   *
+   * @param index The index of the element in the Json Array to return
+   * @returns The element at given index in Json Array, or undefined if index is not valid
+   */
+  public getAt(index: number): Json {
```
staycoolcall911, line 26: So currently every Json object has the `get_at()` method even if it's not a json array?  Let's see if we can come up with a more typesafe API (no suggestion at the moment - only complains :) ).

libs/wingsdk/src/std/json.ts
```diff
@@ -0,0 +1,88 @@
...
+    index;
+    throw new Error("Macro");
+  }
+}
+
+/**
+ * Mutable Json
+ */
+export class MutJson {
```
staycoolcall911, line 35: Can we have `MutJson` extend `Json` to avoid duplicated methods? We did the same at array.ts.
```suggestion
export class MutJson extends Json {
```

libs/wingc/src/jsify.rs
```diff
@@ -41,6 +41,10 @@ const $App = __app(process.env.WING_TARGET);
 const TARGET_APP: &str = "$App";
 
 const INFLIGHT_OBJ_PREFIX: &str = "$Inflight";
+pub struct JSifyContext {
+	pub in_json: bool,
```
staycoolcall911, line 45: Great change - I like it much better than `self.in_json`!

staycoolcall911 (commented):
Looks great - are we ready to ship it?

libs/wingsdk/src/std/json.ts
hasanaburayyan: The issue is that Json get methods return Json/MutJson whereas the arrays return a generic type. So for instance if MutJson class extended Json and did not override the get method it would return a `Json` when using .get() which means you couldn't do something like 

```ts
let x = MutJson {
	y: {
		z: "cool"
	}
}

let a = x.get("y");

a.set("z", "beans");
   ^^^ No set method on Json
```

This could be solved potentially with a behavioral interface, but I think we might rethink this entirely with `dot` and `[]` notation.

libs/wingsdk/src/std/json.ts
hasanaburayyan: Yea I feel you here, actually if it wasnt for the fact that `get()` only accepts a string that method could have handled arrays as well 😅 I dont like this DX but I think we will solve this with `dot` and `[]` notation

libs/wingsdk/src/std/json.ts
staycoolcall911: Ok, thanks for clarifying. I guess we could have `get` return the generic type T1, no?
In any case, this might suffice for now as we'll give it another pass when we implement the `dot` notation.

staycoolcall911 (approved):

eladb (approved):
Amazing!

staycoolcall911:
This is a major milestone @hasanaburayyan! Well done!

libs/wingsdk/src/std/json.ts
eladb: This is fine. `Json` is by definition not type-safe. If you know the structure, you can use `from_json` to turn it into a type-safe object.

libs/wingc/src/type_check.rs
```diff
@@ -339,6 +342,9 @@ impl Subtype for Type {
 				let r: &Type = &*r0;
 				self.is_subtype_of(r)
 			}
+			// This allows us for assignment from native types without allowing assignment to native types
+			// e.g. assert("hello" == x.world) but NOT assert(x.world == "hello")
+			(_, Self::Json) | (_, Self::MutJson) => true,
```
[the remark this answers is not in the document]
hasanaburayyan: Yea you are right, this was the "make it work" for assignment from native types to Json. I have an idea for tightening this up. We have an open bug for clearing this up https://github.com/winglang/wing/issues/1717

PR #1528, opened 2023-02-14
chore: revert "chore: test user defined resources"

MarkMcCulloh:
Reverts winglang/wing#1525

The test changed here breaks on windows and the PR was merged without passing tests

PR #1529, opened 2023-02-14
fix: invalid inflight client for custom resources on windows

MarkMcCulloh:
fixing forward instead of reverting https://github.com/winglang/wing/pull/1528

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -63,8 +63,7 @@ impl<'a> JSifier<'a> {
 	fn js_resolve_path(path_name: &str) -> String {
 		format!(
 			"require('path').resolve(__dirname, \"{}\").replace(/\\\\/g, \"/\")",
-			// normalize path separators to posix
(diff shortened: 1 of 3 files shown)


libs/wingc/src/jsify.rs
```diff
@@ -62,6 +62,8 @@ impl<'a> JSifier<'a> {
 
 	fn js_resolve_path(path_name: &str) -> String {
 		format!("require('path').resolve(__dirname, \"{}\")", path_name)
+			// escape backslashes for windows paths
```
staycoolcall911, line 65: Just verifying: do we want to limit this to windows only? I'm assuming this is ok on linux/macOS, right?

staycoolcall911 (approved):
Approving - pending a small question I had

libs/wingc/src/jsify.rs
MarkMcCulloh: Technically posix does allow `\` in paths, so let me see if I can do this a little differently

PR #1537, opened 2023-02-15
fix(sdk): environment variable already set error

### LINKED ISSUE #1531: Environment variable already set error
### Community Note

### I tried this:

I compiled this file to `sim` or `tf-aws`

```
bring cloud;

resource Foo {
  b: cloud.Bucket;
  init(b: cloud.Bucket) {
    this.b = b;
  }
  inflight do_stuff() {
    this.b.put("foo", "bar");
  }
}

let foo = new Foo(bucket);
  foo.do_stuff();
  bucket.list();
}) as "test";
```

### I expected this:

It works

### Instead, this happened:

```
preflight error: Environment variable "BUCKET_HANDLE_2cd0933d" already set.
  note: run with `NODE_STACKTRACE=1` environment variable to display a stack trace
```

### Component:

Compiler, SDK

### Wing Version:

0.5.36

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

I think the SDK should just no-op when `addEnvironment` is called on a function with the same key-value pair multiple times (this feels reasonable)

Chriscbr:
Fixes #1531

Refactors how the SDK performs binding so that when the tree of resources is being constructed, bindings only get "registered" to a resource, and during a `preSynthesize` hook the actual bindings are performed (by calling `_bind` methods). The purpose of this is so that `_bind` is not called on the same resource-host pair multiple times.

For example, if `resourceA` uses `bucketA`'s `put` operation, and `resourceB` uses `bucketA`'s `list` operation (and both `resourceA` and `resourceB` were used in the same cloud function), the SDK would previously call `bucketA._bind(host, ["put"])` followed by `bucketA._bind(host, ["list"])`. Now, the SDK will only call `bucketA._bind(host, ["put", "list"])`.


libs/wingsdk/src/core/resource.ts
```diff
@@ -39,6 +39,12 @@ export interface IResource extends IInspectable, IConstruct {
    */
   _bind(host: IInflightHost, ops: string[]): void;
 
+  /**
+   * TODO
```
Chriscbr, line 43: Add docstrings

libs/wingsdk/test/target-tf-azure/__snapshots__/captures.test.ts.snap
```diff
@@ -21,9 +21,7 @@ exports[`function with a bucket binding requiring only read 2`] = `
     \\"azurerm_linux_function_app\\": {
       \\"root_Function_9E43762E\\": {
         \\"app_settings\\": {
-          \\"BUCKET_NAME_e51fbb13\\": \\"\${azurerm_storage_container.root_Bucket_D7A55221.name}\\",
           \\"FUNCTIONS_WORKER_RUNTIME\\": \\"node\\",
-          \\"STORAGE_ACCOUNT_e51fbb13\\": \\"\${azurerm_storage_account.root_StorageAccount_3B92EC74.name}\\",
```
Chriscbr, line 26: These shouldn't be removed, need to investigate

libs/wingsdk/src/cloud/function.ts
```diff
@@ -69,8 +69,8 @@ export abstract class FunctionBase extends Resource implements IInflightHost {
```
Chriscbr, line 69: TODO: add test for the original issue in `examples/tests`

Chriscbr:
TODO: also deduplicate operations by using a `Set`

libs/wingsdk/src/core/resource.ts
```diff
@@ -106,6 +112,8 @@ export abstract class Resource
     });
   }
 
+  private readonly bindMap: Map<IResource, Set<string>> = new Map();
```
MarkMcCulloh, line 115: Not sure if this will be used for any different binding in the future, but for now we should limit the typing?

```suggestion
  private readonly bindMap: Map<IInflightHost, Set<string>> = new Map();
```

libs/wingsdk/src/target-tf-azure/function.ts
```diff
@@ -152,11 +152,13 @@ export class Function extends cloud.FunctionBase {
         },
       },
       httpsOnly: true,
-      appSettings: {
-        ...this.env,
-        WEBSITE_RUN_FROM_PACKAGE: `https://${this.storageAccount.name}.blob.core.windows.net/${functionCodeBucket.storageContainer.name}/${functionCodeBlob.name}`,
-        FUNCTIONS_WORKER_RUNTIME: functionRuntime,
-      },
+      appSettings: Lazy.anyValue({
```
MarkMcCulloh, line 155: Hopefully this isn't too much of a "gotcha" over time, having to know what values may be mutated by bindings in presynth.

MarkMcCulloh (approved):
I think this change makes a lot of sense and LGTM, will put a label to give Elad a chance to peep it maybe.

libs/wingsdk/src/core/resource.ts
Chriscbr: Good idea

eladb (approved):
LGTM.

Not in love with the naming but since they are all internal for now, I am okay :-)

libs/wingsdk/src/core/app.ts
```diff
@@ -121,6 +123,12 @@ export class CdktfApp extends Construct implements IApp {
    * for unit testing.
    */
   public synth(): string {
+    // call preSynthesize() on every construct in the tree
+    if (!this.presynthed) {
+      preSynthesizeAllConstructs(this);
+      this.presynthed = true;
```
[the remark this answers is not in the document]
Chriscbr: Sure, I'll do that. (I added this because some tests were calling synth() twice, once in order to start the simulator, and a second time to obtain a snapshot, so alas...)

libs/wingsdk/src/target-tf-azure/function.ts
Chriscbr: Yeah these issues are hard to deal with. The original Wing language proposal mentions an idea of how the language might be able to handle tokens and other lazy values more natively, which could avoid the need for code like this in userland perhaps. https://docs.winglang.io/contributors/rfcs/2022-05-28-winglang-reqs#opaque-primitives-tokens

Issue #111, opened 2022-09-20
Error Handling


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#19-error-handling

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.39](https://github.com/winglang/wing/releases/tag/v0.5.39).

PR #1539, opened 2023-02-15
feat(compiler): add try-catch support

yoav-steinberg:
See #111 


examples/tests/valid/try_catch.w
```diff
@@ -0,0 +1,22 @@
...
+}
+assert(x == "finally");
+
+try {
+  x = "I got here";
+} catch e {
+  x = "caught";
+} finally {
+  assert(x == "I got here");
+  x = "finally";
```
eladb, line 20: can you also add `try-finally`?

libs/tree-sitter-wing/grammar.js
```diff
@@ -223,6 +224,16 @@ module.exports = grammar({
         field("block", $.block),
       ),
 
+    try_catch_statement: ($) =>
+      seq(
+        "try",
+        field("block", $.block),
+        "catch",
+        optional(field("exception_identifier", $.identifier)),
```
eladb, line 232: technically the `catch` block is optional (you should be able to do `try/finally`)

eladb (approved):

staycoolcall911 (approved):
Nice! Necessity is the mother of invention :)
LGTM. One question: what are all the changes you made to tree-sitter-wing/src/parser.c?

libs/tree-sitter-wing/grammar.js
staycoolcall911: If we add that, we need to also update the spec: https://docs.winglang.io/reference/spec#19-error-handling

MarkMcCulloh:
Including the changes to the generated parser is a convention for tree-sitter grammars ([example](https://github.com/tree-sitter/tree-sitter-rust))

It's beneficial during a fresh clone (for new contribs and CI) because `npm install` in that project expects the generated parser to exist. 

It's included as "generated" so the github UI doesn't actually show you the changes, so it shouldn't be too big of a distraction.

yoav-steinberg:
> It's included as "generated" so the github UI doesn't actually show you the changes, so it shouldn't be too big of a distraction.

image

:point_up: here it seems _parser.c_ isn't marked as generated (or is is just a GH bug that they don't display the correct message if the diff it big?)

MarkMcCulloh:
I'm going to assume it's a bug/feature, we explicitly mark everything in `src` as generated here

https://github.com/winglang/wing/blob/06dbe4e9c98c7ea78efd8525d9c04a4e0fcfcf43/libs/tree-sitter-wing/.gitattributes#LL1C6-L1C6

libs/tree-sitter-wing/grammar.js
yoav-steinberg: updated code and spec.

examples/tests/valid/try_catch.w
yoav-steinberg: done.

libs/wingc/src/type_check.rs
```diff
@@ -1846,6 +1846,65 @@ impl<'a> TypeChecker<'a> {
...
+						false,
+						env.flight,
+						stmt.idx,
+					);
+
+					// Add the exception variable to the catch block
+					if let Some(exception_var) = &catch_block.exception_var {
+						match catch_env.define(
+							exception_var,
+							SymbolKind::make_variable(self.types.string(), false, env.flight),
```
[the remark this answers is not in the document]
yoav-steinberg: Yes because of this: 
https://github.com/winglang/wing/blob/276611e0868b805e8bc771243b97ed4d61bd2cc1/libs/wingc/src/lib.rs#L180
and this:
https://github.com/winglang/wing/blob/276611e0868b805e8bc771243b97ed4d61bd2cc1/libs/wingc/src/jsify.rs#L604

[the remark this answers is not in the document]
Chriscbr: 👍  Clean

PR #1541, opened 2023-02-15
fix(compiler): can't call `return` in `try`/`catch`/`finally`, no error when missing `catch`/`finally`

yoav-steinberg:
* Support `return` in `try`/`catch`.
 * Error if both `catch` and `finally` are missing.


staycoolcall911 (approved):

Chriscbr:
Nice!

PR #1533, opened 2023-02-14
fix: compiling with invalid file causes panic

Chriscbr:
Fixes #1532

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Monada Contribution License](https://docs.winglang.io/terms-and-policies/contribution-license.html)*.

PR #1552, opened 2023-02-15
chore: add test for compiling with missing source file

Chriscbr:
https://github.com/winglang/wing/pull/1533#discussion_r1106810918

I did some minor refactoring so that the `compile` function can be unit tested more easily (it now throws an error instead of printing and exiting immediately).


staycoolcall911:
Thanks for adding that!

PR #1555, opened 2023-02-15
chore: remove unnecessary clones

Chriscbr:
Cloning usually introduces additional memory allocations (i.e. performance costs), and it doesn't seem to be necessary here.


eladb:
Wow. Is there a systematic way to avoid this? Linter?

eladb (approved):

Chriscbr:
I don't think there are for this one, but there is a popular Rust linter named clippy that can catch a decent number of things: https://github.com/rust-lang/rust-clippy

PR #1556, opened 2023-02-16
fix: unable to load jsii modules in vscode, missing highlights, and verbose diagnostics for certain types

### LINKED ISSUE #1488: Cannot import jsii modules from outside of the Wing project directory
#1397 implements basic support for importing jsii libraries in Wing. jsii libraries are found using the node module resolution algorithm (so all you need to do is `npm install <library>` to use the library in your Wing project).

However, since we are invoking the compiler with WebAssembly, we must explicitly declare what paths in the file system to give it access to - today, we only give it access to the parent directory of your Wing source file. This means npm modules in parent `node_modules` directories will not be found, even importing them in the JavaScript would work.

To implement this, we would need to either grant access to the user's entire file system, or add some kind of dynamic checking of  for `brought` file system paths, or handle some of the importing logic outside of the WebAssembly runtime.

Examples:

```
bring "my-lib" as my_lib1; // where "my-lib" is in a parent or grandparent node_modules folder
bring "/path/outside/of/wing/project" as my_lib2;
```

Do you have any use cases where you would like to import a library from outside of your Wing project directory? If so, please tell us and give this issue a +1 (👍) to help us prioritize this feature.

### LINKED ISSUE #1530: `bring` jsii library causes VS Code extension to crash
### Community Note

### I tried this:

```
bring "jsii-code-samples" as jsii;
```

### I expected this:

It works

### Instead, this happened:

Wing Language Server crashed

### Component:

IDE Extension

### Wing Version:

0.5.34

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

_No response_

### LINKED ISSUE #1549: fn is highlighted like builtin keyword
### Community Note

### I tried this:

I wrote this in a Wing file in VS Code
```
bring cloud;

  print("hi");
});
let fn2 = new cloud.Function(inflight () => {
  print("hi");
});
```

### I expected this:

VS Code extension highlights `fn` and `fn2` with the same color

### Instead, this happened:

`fn` is highlighted in the same color as `inflight` and `let` and `bring`

### Component:

IDE Extension

### Wing Version:

0.5.36

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

_No response_

MarkMcCulloh:
Kinda a mishmash, was working on the language server this came out of it

Fixes #1549
Fixes #1530
Fixes #1488

### DIFF
--- a/apps/wing/src/wingc.ts
+++ b/apps/wing/src/wingc.ts
@@ -25,12 +25,14 @@ export interface WingCompilerLoadOptions {
 
 export async function load(options: WingCompilerLoadOptions) {
   const preopens = {
+    "/": "/",
     // .jsii access
     [WINGSDK_MANIFEST_ROOT]: WINGSDK_MANIFEST_ROOT,
     ...(options.preopens ?? {}),
   } as Record<string, string>;
 
   if (process.platform === "win32") {
+    preopens["C:\\"] = "C:\\";
     for (const [key, value] of Object.entries(preopens)) {
       delete preopens[key];
       preopens[normalPath(value)] = value;
--- a/examples/tests/valid/bring_jsii_path.w
+++ b/examples/tests/valid/bring_jsii_path.w
@@ -1,0 +1,9 @@
+bring cloud;
+bring "./node_modules/jsii-code-samples" as jsii_code_samples;
+
+let hello = new jsii_code_samples.HelloWorld();
+let greeting = hello.say_hello("wingnuts");
+
+new cloud.Function(inflight (m: str): str => {
+  assert(greeting == "Hello, wingnuts");
+}) as "test:say_hello";
--- a/libs/wingii/src/node_resolve.rs
+++ b/libs/wingii/src/node_resolve.rs
@@ -11,6 +11,8 @@ use std::{
 
 use serde_json::Value;
 
+use crate::util::package_json::is_path_dependency;
+
 static ROOT: &str = "/";
 static NODE_EXTENSIONS: [&str; 3] = [".js", "json", ".node"];
 static NODE_MAIN_FIELDS: [&str; 1] = ["main"];
@@ -68,7 +70,7 @@ pub fn resolve_from(target: &str, basedir: &Path) -> Result<PathBuf, Box<dyn Err
 	};
 
 	// 3. If X begins with './' or '../' or '/'
-	if target.starts_with("./") || target.starts_with("../") || target.starts_with("/") {
+	if is_path_dependency(target) {
 		let path = basedir.join(target);
 		return resolve_as_file(&path)
 			.or_else(|_| resolve_as_directory(&path))
--- a/libs/wingii/src/node_resolve.rs
+++ b/libs/wingii/src/node_resolve.rs
@@ -67,8 +67,8 @@ pub fn resolve_from(target: &str, basedir: &Path) -> Result<PathBuf, Box<dyn Err
 		basedir
 	};
 
-	// 3. If X begins with './' or '/' or '../'
-	if target.starts_with("./") || target.starts_with('/') || target.starts_with("../") {
+	// 3. If X begins with '.' or '/'
+	if target.starts_with(".") || target.starts_with('/') {
 		let path = basedir.join(target);
 		return resolve_as_file(&path)
 			.or_else(|_| resolve_as_directory(&path))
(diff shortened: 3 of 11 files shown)


apps/wing/src/wingc.ts
```diff
@@ -25,12 +25,14 @@ export interface WingCompilerLoadOptions {
 
 export async function load(options: WingCompilerLoadOptions) {
   const preopens = {
+    "/": "/",
```
MarkMcCulloh, line 28: idk if this is controversial, but we are not using WASM for it's security sandboxing right now. Anything limiting of filesystem access has to be handled in the runtime invoking wasm.

apps/wing/src/wingc.ts
```diff
@@ -25,12 +25,14 @@ export interface WingCompilerLoadOptions {
...
     // .jsii access
     [WINGSDK_MANIFEST_ROOT]: WINGSDK_MANIFEST_ROOT,
     ...(options.preopens ?? {}),
   } as Record<string, string>;
 
   if (process.platform === "win32") {
+    preopens["C:\\"] = "C:\\";
```
MarkMcCulloh, line 35: tbh it may be impossible to grant "full access" to windows, until we delegate fs access to node we will have issues here

apps/wing/src/wingc.ts
```diff
@@ -25,12 +25,14 @@ export interface WingCompilerLoadOptions {
 
 export async function load(options: WingCompilerLoadOptions) {
   const preopens = {
+    "/": "/",
     // .jsii access
     [WINGSDK_MANIFEST_ROOT]: WINGSDK_MANIFEST_ROOT,
```
MarkMcCulloh, line 30: Even though we add "/", adding other explicit preopens is useful in the case of Windows and if any files we need are in drives other than C.

apps/wing/src/wingc.ts
eladb: Why is this needed?

apps/wing/src/wingc.ts
eladb: Do we have an issue about this?

examples/tests/valid/bring_jsii_path.w
```diff
@@ -0,0 +1,9 @@
+bring cloud;
+bring "./node_modules/jsii-code-samples" as jsii_code_samples;
```
eladb, line 2: Why do we need `./node_modules` here? Shouldn't this behave like `require`?

libs/wingii/src/node_resolve.rs
```diff
@@ -67,8 +67,8 @@ pub fn resolve_from(target: &str, basedir: &Path) -> Result<PathBuf, Box<dyn Err
 		basedir
 	};
 
-	// 3. If X begins with './' or '/' or '../'
-	if target.starts_with("./") || target.starts_with('/') || target.starts_with("../") {
+	// 3. If X begins with './' or '../' or '/'
+	if target.starts_with("./") || target.starts_with("../") || target.starts_with("/") {
```
eladb, line 71: Extract to a utility function with some docs an reuse

examples/tests/valid/bring_jsii_path.w
MarkMcCulloh: We don't need it, but this file is testing the behavior of importing from a path rather than a node module (https://github.com/winglang/wing/issues/1488)

apps/wing/src/wingc.ts
MarkMcCulloh: preopens grant the WASI runtime access to the host filesystem, and without some fancy analysis we don't know if the user will attempt to import libraries from anywhere on the system. 
This is also very useful for the lsp, which can open a file anywhere on the user's system at a moment's notice.

apps/wing/src/wingc.ts
MarkMcCulloh: https://github.com/winglang/wing/issues/1560

examples/tests/valid/bring_jsii_path.w
MarkMcCulloh: there is already another test that does `bring "jsii-code-samples"`

eladb (approved):

apps/wing/src/wingc.ts
Chriscbr: Yeah I wasn't sure whether to add this to my bring jsii PR or not (security concerns and all) but I think it makes sense in the end. We could always add some kind of "secure" compile API/option later if needed

Issue #1538, opened 2023-02-15
`wing` command line `print` support


yoav-steinberg:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

The wing `print` command is a useful diagnostics and debugging tool. You may want to use it when compiling or simulating your application from the command line. In this case `print` will output to stdout.

Any preflight call to `print` will output to stdout during compilation `wing compile`. Any inflight call to `print` will output to stdout when simulating the app in `wing test`.

```wing
// sample.w
bring cloud;
print("Hi from preflight!");
new cloud.Function((s:str):str => {
  print("Hi from inflight!");
}) as "test";
```
Now a `wing compile -t sim sample.w` should print:
```
Hi from preflight!
```
While a `wing test sample.w` should output:
```
Hi from preflight!
Hi from inflight!
```

### Use Cases

Debugging/Diagnostics (see above)

### Implementation Notes
**For Preflight prints:**
The context that preflight runs in will need to be given access to the console in the compile command.
```ts
const context = vm.createContext({
    require: preflightRequire,
	console, // <==== like so
    process: {
      env: {
        WINGSDK_SYNTH_DIR: synthDir,
        WING_TARGET: options.target,
      },
    },
    __dirname: workDir,
    __filename: artifactPath,
	...
```
**For Inflight prints:**
We explicitly do not propagate the console to the vm context that the simulator functions run in, this is because inflight
 function bind console.log to the global $logger object. So some extra thought into these prints is needed.

### Component

SDK, CLI

Chriscbr:
> While a wing test sample.w should output:
> ```
> Hi from preflight!
> Hi from inflight!
> ```

This could be a confusing in the respect that "Hi from inflight!" is associated with the one of the tests, while "Hi from preflight" is associated with the entire app / program. Maybe we could highlight when logs are associated with specific tests, e.g.

```
$ wing test sample.w
Compiling...
Hi from preflight!
(root/test) Hi from inflight!
pass | hello.w | root/test
```

We can look to other testing frameworks for inspiration

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.82](https://github.com/winglang/wing/releases/tag/v0.5.82).

PR #1564, opened 2023-02-16
feat: emit "print" logs in tests

eladb:
Emits logs produced by calls to `print` (only in inflight for now) when running `wing test`.


eladb:
Screenshot: 

Screenshot 2023-02-16 at 20 06 18

apps/wing/src/commands/test.ts
```diff
@@ -40,7 +40,7 @@ async function testOne(entrypoint: string) {
   if (results.length === 0) {
     results.push({
       pass: true,
-      path: "(no inflight tests)",
+      path: '',
```
flyingImer, line 43: was it previously misplaced? or the message "no inflight tests" will be gone if no inflight test is found?

eladb:
> If we add `console` to the compile command context I think this PR could close out: https://github.com/winglang/wing/issues/1538
> 
> 

I'll pick it up in a subsequent pull request.

apps/wing/src/commands/test.ts
eladb: It's omitted now. Do you think it's important?

MarkMcCulloh: I would personally expect a message telling me I have no tests. If there was no output I would assume something is broken.

apps/wing/src/commands/test.ts
```diff
@@ -49,14 +49,48 @@ async function testOne(entrypoint: string) {
...
-    const error = result.error ? `\n${chalk.red(result.error)}` : "";
+    const details = new Array<string>();
 
-    const sep = chalk.gray(" | ");
-    const row = [status, basename(entrypoint), pathWithPadding + error].join(
-      sep
-    );
+    // add any log messages that were emitted during the test
+    for (const log of result.traces.filter(t => t.type == "log")) {
+      details.push(chalk.gray(log.data.message));
```
MarkMcCulloh, line 56: Would it be too verbose to include more from the trace? Maybe atleast `log.sourcePath` could be added here.

```suggestion
      details.push(chalk.gray(`[${log.sourcePath}] ${log.data.message}`));
```

MarkMcCulloh (approved):
Very snazzy. Other than a minor comment and snapshot updates LGTM

apps/wing/src/commands/test.ts
flyingImer: ditto, that message was helpful when I started with `wing test`

eladb: ok!

apps/wing/src/commands/test.ts
eladb: Since at the moment, `sourcePath` is going to be the test inflight context, it will result in a weird duplication: 
Screenshot 2023-02-19 at 18 12 54


Let's start without it and we can add later ....

eladb:
Updated:

Screenshot 2023-02-19 at 18 25 05

Issue #1559, opened 2023-02-16
can't capture resource with no inflight fields 


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

The following code:
```ts (wing) 
bring cloud;

resource A {
  field: str;
  init() { this.field = "hey"; }
  // inflight f(): str{
  //   return "hello";
  // }
}

let a = new A();
new cloud.Function(inflight () => {
 assert("hey" == a.field);
}) as "test";
```

### I expected this:

I expected `wing test` to work

### Instead, this happened:

```sh
wingly % wing test a.w
fail | a.w | root/test
ReferenceError: a is not defined
    at Handler.handle (evalmachine.<anonymous>:26:26)
    at exports.handler (evalmachine.<anonymous>:32:12)
    at evalmachine.<anonymous>:35:9
    at Script.runInContext (node:vm:140:12)
    at Object.runInContext (node:vm:290:6)
    at Object.activity (/Users/eyalkeren/.nvm/versions/node/v18.4.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:86:27)
    at Object.withTrace (/Users/eyalkeren/.nvm/versions/node/v18.4.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:107:50)
    at Function.invoke (/Users/eyalkeren/.nvm/versions/node/v18.4.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:83:29)
    at Simulator.runTest (/Users/eyalkeren/.nvm/versions/node/v18.4.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:295:22)
    at async Simulator.runAllTests (/Users/eyalkeren/.nvm/versions/node/v18.4.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:270:26)
```

### Component:

Compiler

### Wing Version:

0.5.44

### Platform(s):

MacOS

### Anything else?

When adding an inflight method to resource A, the example works

```ts (wing)
bring cloud;

resource A {
  field: str;
  init() { this.field = "hey"; }
  inflight f(): str{
    return "hello";
  }
}

let a = new A();
new cloud.Function(inflight () => {
 assert("hey" == a.field);
}) as "test";
[...]

PR #1565, opened 2023-02-16
fix(compiler): capture fields of resource with no inflight methods

yoav-steinberg:
This was a rabbit hole :rabbit2: :hole: thanks @Chriscbr 

The problem is that when _capture.rs_ wants to capture a resource is iterates over all its inflight methods and creates a list of the methods that are (potentially) used by the inflight closure. This is required to pass to the sdk later on. If there are no inflight methods we just "skip" capturing which is a bug.

The quick fix is that if there are any capturable fields in the resource we just add the resource as an immutable data capture (in addition to a resource capture, see `CaptureKind`). But this caused the captured resource to be handled twice in the sdk's `makeHandler`. And then we ran into ordering issues. So the quick fix here was to change the order of some code in the sdk so the double capture won't cause issues.

### note about a possibly better implementation

As @Chriscbr mentioned in #1559, another approach can be not to capture the resource when accessing its immutable fields but instead capture the field itself. But this approach leads us to needing to capture "expressions" because in
```wing
new cloud.Function(inflight ()=> {
  print(some_resource.inner_resource.field);
});
```
capturing the nested identifier `some_resource.inner_resource` requires capturing a preflight expression and turning it into an inflight symbol. This introduces the ability to capture any preflight expression (as long as it evaluates to immutable data). Something like:
```wing
new cloud.Function(inflight ()=> {
  print(some_resource.preflight_method_that_returns_a_resource().field);
});
```
This mechanism can be very cool but requires the capture mechanism and jsifier to do some trickery when jsifying the inflight closure:
1) replacing preflight expressions with a generated symbol.
2) adding the generated symbol to the bindings.
3) initializing the binding with the JS code that'll produce the result of the expression.

All this seems less critical for now, so we'll just cpature the entire resource and rely on the fact that its inner fields are "lifted".


yoav-steinberg:
@eladb would be interesting to hear your thoughts about capturing preflight expressions (see top comment).

eladb:
> @eladb would be interesting to hear your thoughts about capturing preflight expressions (see top comment).

Absolutely! I believe that we have already discussed this in the context of capturing attributes of arbitrary terraform resources into inflight:

```js
let s = new kineses.Stream(..);

let i = inflight () => {
  print(s.arn);
};
```

In the above example Stream is a simple CDK construct and not a Wing resource, so it doesn't have an inflight client, but it should be possible to access immutable preflight fields so it will be possible to interact with it in inflight code.

examples/tests/valid/capture_resource_with_no_inflight.w
```diff
@@ -0,0 +1,13 @@
+bring cloud;
+
+// This resource has no inflight methods (no client) but it still needs to be captured
+// so we can access its fields.
```
eladb, line 4:
```suggestion
// but we can still access its fields.
```

examples/tests/valid/capture_resource_with_no_inflight.w
```diff
@@ -0,0 +1,13 @@
...
+resource A {
+  field: str;
+  init() { this.field = "hey"; }
+}
+
+let a = new A();
+new cloud.Function(inflight () => {
+ assert("hey" == a.field);
+}) as "test";
```
eladb, line 13: Do we have a test where we access a preflight field of a resource that does have inflight methods? (Just to cover that case as well)

eladb (approved):

examples/tests/valid/capture_resource_with_no_inflight.w
yoav-steinberg: That's **is** test. This is what this PR fixes!
There are two different way to get this done: what I did in this PR and a more complex way but possibly better that's mentioned in the top comment.
In any case this test is a test of accessing a preflight field of a resource without inflight methods.

examples/tests/valid/capture_resource_with_no_inflight.w
yoav-steinberg: rephrased

yoav-steinberg:
@Chriscbr I left this as a draft PR just in case you're up to going over that SDK code and removing the duplicate handing of resource vs data binding, since they all call `serializeImmutableData`. If you're not going to handle that I'll remove the draft here and we'll get this merged.

Chriscbr:
@yoav-steinberg I probably won't have time to get to this soon so I'd just merge this. I've created a separate tracking issue for the leftover work: https://github.com/winglang/wing/issues/1605

yoav-steinberg:
> @yoav-steinberg I probably won't have time to get to this soon so I'd just merge this. I've created a separate tracking issue for the leftover work: https://github.com/winglang/wing/issues/1605

Excellent, I'll try to get this merged.

PR #1568, opened 2023-02-17
feat(vscode): add hover information

MarkMcCulloh:
https://user-images.githubusercontent.com/1237390/219514995-4b9d43c8-1b2e-4bda-8ea1-74508949b307.mp4

### What

Added hover for symbols info in vscode. 

This should include all "symbols", but some things don't work correctly yet:
- nested custom types (e.g. `Bucket` from `cloud.Bucket`) don't have info
- named arguments don't have info
- nested identifiers are missing partial info (e.g. reassignable, phase)
- nested functions look ugly

### How
- refactored the lsp module into several files
- Added a file for AST traversal. I tried to make an actual iterator implementation but I gave up and did something _good enough_. 

### Misc
- Added the lsp endpoint for "documentSymbols", since I was already getting symbol info
  - This allows the symbols to show up in the outline, and using `@` in the command palette to go to symbols. It's pretty incomplete though (mostly lacking hierarchy)
- Made sure vscode highlights 0, negative numbers, and decimals
- Removed semantic highlighting for now. Wasn't really working, will revisit later. It will be needed to differentiate highlight for things like class construction and enum usage.

Followups:
- https://github.com/winglang/wing/issues/1569
- https://github.com/winglang/wing/issues/1570


libs/wingc/src/lsp/ast_traversal.rs
```diff
@@ -0,0 +1,887 @@
+use lsp_types::Position;
```
MarkMcCulloh, line 1: If you're looking at this file, you may be thinking: "Hey shouldn't there be some `impl Iterator`s in here?"

You are correct, but iterating over our tree is pretty complicated due to how non-uniform and recursive it is.

I tried it and couldn't finish the task, so I did this instead. I'm open to help on it if I'm just being dumb and it's not that crazy.

libs/wingc/src/lsp/ast_traversal.rs
```diff
@@ -0,0 +1,887 @@
...
+		}
+	}
+
+	None
+}
+
+/// Traverse the given statement context and return all expressions inside it (recursive).
+pub fn get_expressions_from_statement<'a>(context: TreeLocationContext) -> Vec<TreeLocationContext> {
+	let mut return_expressions: Vec<TreeLocationContext> = vec![];
+	match &context.statement.kind {
```
eladb, line 49: Intuitively it feels like we should already have this type of traversal in the compiler code. Is there a way to reuse?

eladb (commented):
Hover looks awesome!


I saw (in the screenshot) that when hovering method it shows only the type of arguments and not the names. Is it possible to also include the names?

libs/wingc/src/lsp/ast_traversal.rs
```diff
@@ -0,0 +1,887 @@
...
+	pub expression: Option<&'a Expr>,
+}
+
+/// Traverse the AST and find the symbol at the given position.
+pub fn find_symbol<'a>(scope: &'a Scope, position: &'a Position) -> Option<(TreeLocationContext<'a>, &'a Symbol)> {
+	let all_statements = get_statements_from_scope_no_context(scope);
+	for context in all_statements {
+		let statement = context.statement;
+		if !statement.span.contains(position) {
+			continue;
```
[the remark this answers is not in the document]
MarkMcCulloh: In this loop I'm trying to find the statement that contains the given position, so this `continue` is to ignore this statement while iterating.

MarkMcCulloh:
> Is it possible to also include the names?

@eladb Unfortunately it would be pretty tough since we don't store any of that info away from the declaration itself. It's definitely an improvement I'd love to see though, perhaps for a followup since this PR is already pretty big.

eladb:
> > I saw (in the screenshot) that when hovering method it shows only the type of arguments and not the names. Is it possible to also include the names?
> 
> 
> 
> @eladb Unfortunately it would be pretty tough since we don't store any of that info away from the declaration itself. It's definitely an improvement I'd love to see though, perhaps for a followup since this PR is already pretty big.

No problem. Can you open a tracking issue?

libs/wingc/src/lsp/ast_traversal.rs
MarkMcCulloh: Yeah this traversal code is similar to something we've written a few times. Unfortunately, I'm not sure the code I've written here is good enough for the rest of the compiler. Mostly because the compiler phases may need to have the iteration ordered in a certain way that I'm not sure these satisfy.

It's something I'd love to figure out soon though because I'm tired of writing this code.

https://github.com/winglang/wing/issues/1570

libs/wingc/src/lsp/hover.rs
```diff
@@ -0,0 +1,122 @@
...
+			};
+			let reference = if let Some(expr) = expr {
+				if let ExprKind::Reference(reference) = &expr.kind {
+					Some(reference)
+				} else {
+					None
+				}
+			} else {
+				None
+			};
```
Chriscbr, line 58: I think this is slightly more idiomatic rust
```suggestion
			let expression_type = expr.as_ref().and_then(|expr| {
				let t = expr.evaluated_type.borrow();
				t.clone()
			});
			let reference = expr.as_ref().and_then(|expr| {
				if let ExprKind::Reference(reference) = &expr.kind {
					Some(reference)
				} else {
					None
				}
			});
```

libs/wingc/src/lsp/hover.rs
```diff
@@ -0,0 +1,122 @@
...
+			} else {
+				&symbol.span
+			};
+
+			let env_ref = scope.env.borrow();
+			let env = env_ref.as_ref().unwrap();
+
+			let mut symbol_lookup = env.lookup_ext(symbol, None);
+			if symbol_lookup.is_err() {
+				// There is a bug with the .parent of immediate children of the root scope, so attempt to look up the symbol in the root scope
```
Chriscbr, line 70: Create an issue?

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -25,9 +25,6 @@ pub struct SymbolEnv {
 	statement_idx: usize,
 }
 
-// TODO See TypeRef for why this is necessary
-unsafe impl Send for SymbolEnv {}
```
Chriscbr, line 29: 🎉

libs/wingc/src/lsp/document_symbols.rs
```diff
@@ -0,0 +1,126 @@
...
+					SymbolKind::VARIABLE
+				} else {
+					continue;
+				}
+			}
+			StmtKind::Class(c) => {
+				if c.name.name == symbol.name {
+					SymbolKind::CLASS
+				} else {
+					// eventually we should handle the entire class symbol hierarchy
```
Chriscbr, line 84: What's an example of a case the code won't handle? (just curious)

libs/wingc/src/lsp/ast_traversal.rs
Chriscbr: Definitely something I'd be down to try

libs/wingc/src/lsp/ast_traversal.rs
```diff
@@ -0,0 +1,887 @@
...
+				if let Some(exception_var) = &catch_block.exception_var {
+					return_symbols.push((context, exception_var));
+				}
+			}
+		}
+		While { .. } => {}
+		If { .. } => {}
+		Return { .. } => {}
+		Expression { .. } => {}
+		Scope(_) => {}
```
Chriscbr, line 771:
```suggestion
		// These nodes do not define any new symbols
		While { .. } => {}
		If { .. } => {}
		Return { .. } => {}
		Expression { .. } => {}
		Scope(_) => {}
```

libs/wingc/src/lsp/ast_traversal.rs
MarkMcCulloh: @Chriscbr Sweet, as a followup to this PR or as part of this one?

Chriscbr: Either is cool with me 😄

libs/wingc/src/lsp/document_symbols.rs
```diff
@@ -0,0 +1,126 @@
...
+		} else {
+			0
+		}
+	} else {
+		eprintln!("Failed to parse 'onHover' text document: {}", parse_string);
+		0
+	}
+}
+
+pub fn on_document_symbols<'a>(params: lsp_types::DocumentSymbolParams) -> Option<Vec<DocumentSymbol>> {
```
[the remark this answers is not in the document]
MarkMcCulloh: I'm actually not 100% sure if there is a semantic difference or not (from the perspective of the lsp), I agree with the sentiment though so I'll make that change.

libs/wingc/src/lsp/document_symbols.rs
MarkMcCulloh: A big part of the DocumentSymbol thing is that it can represent hierarchy. e.g.

```
variable
closure
	class
		method
		variable
```

This whole file is just making all symbols flat, as if they were all defined at the root. It's not useless right now, but it is inaccurate.

Chriscbr (approved):
🚢 it!

PR #1571, opened 2023-02-17
feat: isomorphic wasi

MarkMcCulloh:
### Current State

wingc is currently consumed in two ways:
- in the cli, via node's `wasi` modules
- in the playground, via https://github.com/wasmerio/wasmer-js

This is ok, but there are several downsides:
- Some code duplication with initialization and invocation of wingc
- wasmer-js is frankly not well maintained
- node's `wasi` only works in node
  - not a big deal, but prevented usage in other node-compatible runtimes or https://webcontainers.io/
- node `wasi` required node18 or newer and more importantly requires the use of an experimental flag

### New Change

I noticed that [bun](https://bun.sh/) recently started using [wasi-js](https://www.npmjs.com/package/wasi-js) for wasi. This library is a better-maintained fork of wasmer-js. It's part of https://cowasm.org/ which is a super cool project.

So I switched the CLI to use this library to load wingc, and made the playground consume the cli as a library.

## TODO

- [x] Windows?
- [x] Node16?
- [x] Docs


package.json
```diff
@@ -26,7 +26,7 @@
     "install:wasi": "bash scripts/setup_wasi.sh",
     "install:wing-api-checker": "npm --prefix apps/wing-api-checker --install-links=false install",
     "install:wing-playground": "npm --prefix apps/wing-playground --install-links=false install",
-    "install:winglang": "npm --prefix apps/wing --install-links=false install",
+    "install:winglang": "npm --prefix apps/wing --install-links=false install --ignore-scripts",
```
Chriscbr, line 29: What does this change do? (just curious for learning)

apps/wing/src/wingc.ts
```diff
@@ -21,9 +18,16 @@ export interface WingCompilerLoadOptions {
   imports?: Record<string, any>;
   preopens?: Record<string, string>;
   env?: Record<string, string>;
+  fs?: any;
+  wingcWASMData?: Uint8Array;
+  wingsdkManifestRoot?: string;
```
Chriscbr, line 23: Please add docstrings

Can `fs` be a more precise type here? What is `wingcWASMData`?

apps/wing/src/wingc.ts
```diff
@@ -1,15 +1,12 @@
 import debug from "debug";
-import { readFile } from "fs/promises";
-import { resolve } from "path";
-import { WASI } from "wasi";
+import { readFileSync } from "fs";
 import { normalPath } from "./util";
+import WASI from "wasi-js";
```
Chriscbr, line 4: (checking my understanding) So unlike the `wasi` module which is a node builtin, `wasi-js` works both browser-side and server-side? Big if true

Chriscbr (approved):

apps/wing/src/wingc.ts
MarkMcCulloh: Yessir

package.json
MarkMcCulloh: I'm trying to find a way around this, but a preinstall script was causing an error: https://github.com/winglang/wing/actions/runs/4206774554/jobs/7300707121#step:6:213

winglang currently doesn't have any other install hooks, so this isn't a complete deal breaker but I'll see if there's some other way to fix it.

MarkMcCulloh: Caused by https://github.com/pnpm/only-allow/issues/13

Keeping this changed in though, because ignoring scripts doesn't hurt us atm

apps/wing/src/wingc.ts
```diff
@@ -15,11 +15,59 @@ export type WingCompilerFunction =
...
+
+  /**
+   * A filesystem for the WASI instance to use.
+   *
+   * @default - The `fs` module from Node.js
+   */
   fs?: any;
+
+  /**
+   * The bytes of of the `wingc.wasm` data loaded into memory.
```
Chriscbr, line 60:
```suggestion
   * The bytes of the `wingc.wasm` data loaded into memory.
```

Chriscbr (approved):

eladb (commented):
Goodness!

PR #1574, opened 2023-02-18
chore(compiler): visitor trait

### LINKED ISSUE #1570: Consistent AST traversal methods
### Community Note

### Feature Spec

We need a way to traverse the wing AST without rewriting so many `match`es

### Use Cases

Code duplication and maintenance headaches from ensuring traversal is correct in several places.

### Implementation Notes

_No response_

### Component

_No response_

Chriscbr:
Introduces a new Visit trait in the compiler that allows us to use the visitor pattern. The visitor pattern makes it easy to add new compiler passes since it handles the logic of traversing the AST tree for you, only requiring you to specify implementations of methods for the kinds of AST nodes you're interested in.

I used the trait to clean up `capture.rs` and remove some boilerplate.

No user-impacting changes.

Closes #1570 

Misc changes:
- renamed `args` to `arg_list` to be consistent with the struct name, and the identically-named field in `ExprKind::New`
- renamed `Expr::Binary.lexp` and `rexp` to `left` and `right` (copilot suggested it)


libs/wingc/src/visit.rs
```diff
@@ -0,0 +1,270 @@
...
+	}
+	fn visit_reference(&mut self, e: &'ast Reference) {
+		visit_reference(self, e);
+	}
+	fn visit_function_definition(&mut self, e: &'ast FunctionDefinition) {
+		visit_function_definition(self, e);
+	}
+	fn visit_args(&mut self, e: &'ast ArgList) {
+		visit_args(self, e);
+	}
```
Chriscbr, line 49: We can add a lot more cases for visiting more specific types of nodes (e.g. `visit_expr_binary_op`), the only limitation is we don't have dedicated structs with this information because they're currently inlined into the `Expr` enum variants (like [here](https://github.com/winglang/wing/blob/b0e799f14fb4a864eb26007ae45cb418389c5649/libs/wingc/src/ast.rs#L304-L309)). I can create those structs but I left it out of this PR to avoid changing too much at once / making the code harder to review.

libs/wingc/src/visit.rs
```diff
@@ -0,0 +1,272 @@
...
+///
+/// For example:
+///
+/// ```ignore
+/// impl<'ast> Visit<'ast> for ExprVisitor {
+///   fn visit_item_fn(&mut self, exp: &'ast Expr) {
+///     println!("Expr with span={}", exp.span);
+///
+///     // Delegate to the default impl to visit any nested expressions.
+///     visit::visit_expr(self, exp);
```
yoav-steinberg, line 20: IIUC this pattern uses these `visit::visit_X(self, X_struct)` "static" methods as the "default" implementation to achieve some sort of "base class" implementation effect (without inheritance). Would you say that's correct, just so I get what we're doing here?

libs/wingc/src/visit.rs
```diff
@@ -0,0 +1,272 @@
...
+///   }
+/// }
+/// ```
+pub trait Visit<'ast> {
```
yoav-steinberg, line 24: I think this begs to be a rust macro that receives a list of AST node types and generates the entire trait. Not critical but if I'm write here lets at least add a TODO.

libs/wingc/src/visit.rs
```diff
@@ -0,0 +1,272 @@
...
+	fn visit_scope(&mut self, e: &'ast Scope) {
```
yoav-steinberg, line 25: Would be cool if we could find some way to pass some [`FnOnce`](https://doc.rust-lang.org/std/ops/trait.FnOnce.html) here so we can just provide a closure of "what to do" in this node that returns a `bool` telling the default visitor implementation if we want to continue to traverse deeper into this node or return. This way we don't need to copy the default implementation for each visit function we create.

yoav-steinberg (approved):
I like this a lot :rocket:
Any work of refactoring our original quick and dirty code is highly appreciated!
Added some thoughts...

libs/wingc/src/visit.rs
MarkMcCulloh: I think this is a good option to have, but sometimes closures are more annoying to deal with (due to borrowing rules) so it shouldn't be the only way. I originally implemented something like this (visitor with FnMut) in my hover PR and ended up getting rid of it.

MarkMcCulloh:
While I think this is a good implementation that we should have, one downside is the each visitor function doesn't know anything about where it is in the tree. Maybe this is more a problem with our AST nodes idk (should they have a .parent?). This is fine for traversal but the actual visitor may need that info.

In my PR for the vscode hover, I specifically need to know what scope things are in so I can do a lookup.
Later on, when working on vscode completion, if you're inside an expression we'll also need to know what kind of statement/scope you're in (to provide a contextually useful completion).

yoav-steinberg:
> While I think this is a good implementation that we should have, one downside is the each visitor function doesn't know anything about where it is in the tree...

Maybe we can easily pass some `&mut T` to all functions where `T` is a some traverse state information that gets updated/read inside the visit functions if needed. @hasanaburayyan just needed something like that to know if we're inside a json literal in _parser.rs_.

hasanaburayyan:
> > While I think this is a good implementation that we should have, one downside is the each visitor function doesn't know anything about where it is in the tree...
> 
> Maybe we can easily pass some `&mut T` to all functions where `T` is a some traverse state information that gets updated/read inside the visit functions if needed. @hasanaburayyan just needed something like that to know if we're inside a json literal in _parser.rs_.

Yea I really like this, I think potentially maybe another extension of this is could be to implement `preVisit` and `postVisit` methods. This combined with `&mut T` references would make it easy to keep track things like nested depth during traversal.

eladb (commented):
Beautiful

staycoolcall911 (approved):
Awesome, I love it

libs/wingc/src/capture.rs
```diff
@@ -48,141 +52,46 @@ fn collect_captures(capture_list: Vec<Capture>) -> Captures {
 	captures
 }
 
-pub fn scan_for_inflights_in_scope(scope: &Scope, diagnostics: &mut Diagnostics) {
```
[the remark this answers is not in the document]
Chriscbr: Ah, I see. I think it's doable, the second traversal is just a little bit more tricky because as implemented because of the environment and statement index passing. Not sure if I can get to it in this PR but it's a good callout.

libs/wingc/src/visit.rs
Chriscbr: Exactly. To be totally honest before this PR I wasn't even aware this was available, but turns out this is a well supported feature of traits in Rust (even mentioned in the Rust book here: https://doc.rust-lang.org/book/ch10-02-traits.html). 😄

libs/wingc/src/visit.rs
Chriscbr: Totally. It's definitely possible to code-generate this kind of code. This would also make it easier to maintain multiple traits, like separate a `VisitMut` trait variation that gives mutable access to each tree node instead of read-only access.

Chriscbr:
> While I think this is a good implementation that we should have, one downside is the each visitor function doesn't know anything about where it is in the tree. Maybe this is more a problem with our AST nodes idk (should they have a .parent?). This allows for traversal but the actual visitor may need that info for its work.
> 
> In my PR for the vscode hover, I specifically need to know what scope things are in so I can do a lookup. Later on, when working on vscode completion, if you're inside an expression we'll also need to know what kind of statement/scope you're in (to provide a contextually useful completion).

I think this is something that can be handled by storing mutable state whatever struct implements the `Visit` trait. (Note that each method accepts a `&mut self`). The visitor traverses the tree in depth-first order so that is a property that implementors can rely on.

PR #1575, opened 2023-02-18
feat(examples): plugin examples

hasanaburayyan:
Made a place to start gathering some working plugin examples

### DIFF
--- a/examples/plugins/permission-boundary.js
+++ b/examples/plugins/permission-boundary.js
@@ -1,0 +1,20 @@
+var iam_role = require("@cdktf/provider-aws/lib/iam-role");
+var cdktf = require("cdktf");
+
+class PermissionBoundaryAspect {
+  constructor(permissionBoundaryArn) {
+    this.permissionBoundaryArn = permissionBoundaryArn;
+  }
+
+  visit(node) {
+    if (node instanceof iam_role.IamRole) {
+      node.permissionsBoundary = this.permissionBoundaryArn;
+    }
+  }
+}
+
+exports.preSynth = function(app) {
+  if (!process.env.PERMISSION_BOUNDARY_ARN) {throw new Error("env var PERMISSION_BOUNDARY_ARN not set")}
+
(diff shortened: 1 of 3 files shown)


eladb (changes requested):
Love these!

Can we add a test for each example that will verify that it's working during build?

staycoolcall911 (approved):
So great to have the start of our plugins library!

hasanaburayyan:
> Love these!
> 
> Can we add a test for each example that will verify that it's working during build?

I added the test cases as part of the plugin-manager test suite. I thought about using snapshot but I think its more valuable to test hard assertions against the outputs of the plugins.

eladb (approved):
Very nice!

Should we add links to these examples in our docs?

libs/wingsdk/test/core/plugin-manager.test.ts
```diff
@@ -196,3 +198,94 @@ test("plugins are run in order they are passed in", () => {
   // THEN
   expect(tfResourcesOfCount(postSynthOutput, "aws_s3_bucket")).toEqual(0);
 });
+
+describe("test plugin examples", () => {
+  const EXAMPLES_DIR = "../../examples/plugins";
```
Chriscbr, line 203: FYI @hasanaburayyan for the future, I think we should host these examples into the SDK directory, or move the tests to elsewhere (like the hangar). 

One reason is puristic  -- I think it will be easier to maintain the SDK long term when it's self contained, so that none of source code or test code depends on other libraries in the monorepo.

But the practical reason is that by referencing the examples directory, we are introducing a dependency that NX (our task runner) isn't aware of. So if we want to keep the tests here / reference examples from outside, we should update our NX configuration files -- otherwise it's possible for someone to ask NX to run the SDK tests, update an example, and then when they ask NX to run the SDK tests again, it will think no files have changed (because the examples directory wasn't mentioned as a dependency) and it will just return the cached result.

hasanaburayyan: @Chriscbr thanks for the catch and great points. I’m going to get these moved (just been procrastinating)

PR #1576, opened 2023-02-18
feat(sdk): adding awscdk provider

marciocadev:
# Adding aws-cdk provider to Wing

## Code sample
```ts (wing)
bring cloud;
let b = new cloud.Bucket() as "wing-cdk-bucket";
b.add_object("wing-file.txt", "Hello World!");
```

When we compile the above code with the command
```bash
$ wing compile -t awscdk app.w
```
the `target/app.awscdk` folder will be created

deploying/destroying the stack
```bash
$ cd target/app.awscdk
$ npm install
$ npm run deploy
$ npm run destroy
```


docs/04-reference/wingsdk-api.md
```diff
@@ -3027,7 +3027,7 @@ Represents the bag of attributes as key-value pairs.
 
 - *Extends:* constructs.IConstruct
 
-- *Implemented By:* core.CdktfApp, sim.App, testing.SimApp, tfaws.App, tfazure.App, tfgcp.App, core.IApp
+- *Implemented By:* awscdk.App, core.AwsCdkApp, core.CdktfApp, sim.App, testing.SimApp, tfaws.App, tfazure.App, tfgcp.App, core.IApp
```
eladb, line 3030: Update docs/cli

libs/wingsdk/src/core/awscdk-app.ts
```diff
@@ -0,0 +1,178 @@
...
+      "    return;",
+      "  }",
+      "",
+      "  const s = spinner();",
+      "  s.start(\"Removing stack\");",
+      "  execSync(`aws cloudformation delete-stack --stack-name ${appName}`);",
+      "  s.stop(\"Stack removed\");",
+      "",
+      "  s.start(\"Deleting stack-name file\");",
+      "  unlinkSync(\"./stack-name\");",
```
eladb, line 95: What's the rational of not simply instructing users to use the CDK cli here?

libs/wingsdk/src/core/awscdk-app.ts
```diff
@@ -0,0 +1,178 @@
...
+      "const { intro, outro, spinner, text, note } = require(\"@clack/prompts\");",
+      "const { writeFileSync, readFileSync, existsSync } = require(\"fs\");",
+      "const { execSync } = require(\"child_process\");",
+      "const color = require(\"picocolors\");",
+      "",
+      "async function main() {",
+      "  intro(`${color.bgCyan(color.black(\" CDK App \"))}`);",
+      "",
+      "  var appName;",
+      "  if (existsSync(\"./stack-name\")) {",
```
eladb, line 122: Similarly to terraform, let's just tell users to install the CDK cli and do:

```sh
$ wing compile -t awscdk hello.w
$ cd target/hello.awscdk
$ cdk deploy
```

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -0,0 +1,87 @@
...
+export class Bucket extends cloud.BucketBase {
+  private readonly bucket: S3Bucket;
+  private readonly public: boolean;
+
+  constructor(scope: Construct, id: string, props: cloud.BucketProps) {
+    super(scope, id, props);
+
+    this.public = props.public ?? false;
+
+    const bucketName = ResourceNames.generateName(this, BUCKET_PREFIX_OPTS);
```
eladb, line 32: Contrary to terraform, the recommended practice for cloudformation (cdk) is not to supply a name at all and let the engine allocate one.

So this is not needed

libs/wingsdk/src/target-awscdk/factory.ts
```diff
@@ -0,0 +1,38 @@
...
+  public resolve(
+    type: string,
+    scope: IConstruct,
+    id: string,
+    ...args: any[]
+  ): IConstruct {
+    switch (type) {
+      case BUCKET_TYPE:
+        return new Bucket(scope, id, args[0]);
+      // case FUNCTION_TYPE:
```
eladb, line 24: Remove the commented code

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -0,0 +1,87 @@
...
+  //   }
+
+  //   if (ops.includes(cloud.BucketInflightMethods.PUT)) {
+
+  //   }
+  // }
+
+  /** @internal */
+  public _toInflight(): core.Code {
+    return core.InflightClient.for(__filename, "BucketClient", [
```
eladb, line 77: Does this use the same bucket client as the terraform provider? 

(It technically can)

eladb (changes requested):
This is great.
I am a bit concerned by the fact that this will increase our package size dramatically but it's a premature optimization

marciocadev:
> This is great. I am a bit concerned by the fact that this will increase our package size dramatically but it's a premature optimization

I think you're right. Every new provider increases the size of `Wing`.

Perhaps it would be better to wait for a package manager of some kind so that we can download what we need.

I will close this PR for now

eladb:
No! I would love for this to be supported, even at the cost of package size.

We will figure out how to optimize later!

If you are inclined, happy to work with you to get this over the finish line

marciocadev:
Back to work 😄

libs/wingsdk/src/target-awscdk/bucket.ts
marciocadev: I can remove it, but I'm a little worried about this change. Personally, I never let the CDK create my resource names because they are difficult to find in the console.

libs/wingsdk/src/core/awscdk-app.ts
marciocadev: I'm not sure how to use cdk deploy in this case. As far as I know, cdk deploy uses cdk.json to synthesize and deploy a stack. However, when we run wing compile -t awscdk app.w, the cdk.out directory is created with the necessary files to deploy, so we need to find a way to deploy only those files. I'm not sure how to configure cdk.json for that purpose.

The alternative I found was to use cdk-assets to upload the required files to the CDK's S3 folder by using the root.assets.json file, and then execute aws cloudformation deploy with the root.template.json file.

eladb: `cdk deploy` accepts a `cdk.out` directory in the `--app` option:

```sh
cdk deploy --app path/to/cdk.out
```

marciocadev: It works, cool, I will remove the scripts

marciocadev: I think I've hit a roadblock.
Using the `cdk deploy` command works, but in the console, all the created stacks have the same name. I tried using `node.addr`, but it doesn't change the stack even if I use a different path.

eladb: I see... I think you'll need to load the stack name from an environment variable like `CDK_STACK_NAME`

See an example of AZURE_LOCATION.

docs/04-reference/wingsdk-api.md
marciocadev: Done

docs/04-reference/wingcli.md
```diff
@@ -133,6 +133,25 @@ JavaScript bundles that include inflight code that executes on compute platform
...
+$ wing compile --target awscdk cdk-app.w
+```
+
+The output includes both a AWS-CDK configuration file (under `target/<file name>.awscdk`) and
+JavaScript bundles that include inflight code that executes on compute platform such as AWS Lambda.
+
+You can deploy your stack to AWS using `cdk deploy` command
+```sh
+$ cdk deploy --app target/cdk-app.awscdk
+```
```
Chriscbr, line 153: Calling it `cdk-app.w` feels a bit redundant here 😛 

````suggestion
```sh
$ export CDK_STACK_NAME="my-project"
$ wing compile --target awscdk app.w
```

The output includes both a AWS-CDK configuration file (under `target/<file name>.awscdk`) and
JavaScript bundles that include inflight code that executes on compute platform such as AWS Lambda.

You can deploy your stack to AWS using `cdk deploy` command
```sh
$ cdk deploy --app target/app.awscdk
```
````

File: libs/wingsdk/src/core/awscdk-app.ts [hunk 8]
```diff
@@ -0,0 +1,81 @@
...
+
+  private readonly cdkApp: cdk.App;
+  private readonly cdkStack: cdk.Stack;
+
+  constructor(props: CdkAppProps) {
+
+    const stackName = props.stackName ?? process.env.CDK_STACK_NAME;
+    if (stackName === undefined) {
+      throw new Error(
+        "A CK stack name must be specified through the CDK_STACK_NAME environment variable."
```
Chriscbr, line 39:
```suggestion
        "A CDK stack name must be specified through the CDK_STACK_NAME environment variable."
```

libs/wingsdk/src/core/awscdk-app.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    this.outdir = outdir;
+    this.cdkApp = cdkApp;
+    this.cdkStack = cdkStack;
+
+    // register a logger for this app.
+    Logger.register(this);
+  }
+
+  /**
```
Chriscbr, line 68:
```suggestion
/**
```

libs/wingsdk/src/core/awscdk-app.ts
```diff
@@ -0,0 +1,81 @@
+import { Construct } from "constructs";
```
Chriscbr, line 1: I would define this all in `target-awscdk/app.ts` since the logic does not need to be shared with any other targets

Chriscbr (commented):
Very cool work :)

libs/wingsdk/src/core/awscdk-app.ts
marciocadev: Great 😃

It's really make more sense

libs/wingsdk/src/target-awscdk/factory.ts
```diff
@@ -0,0 +1,27 @@
...
+    id: string,
+    ...args: any[]
+  ): IConstruct {
+    switch (type) {
+      case BUCKET_TYPE:
+        return new Bucket(scope, id, args[0]);
+      case LOGGER_TYPE:
+        return new Logger(scope, id);
+      default:
+        throw new Error(`Type ${type} not implemented for tf-aws target.`);
```
staycoolcall911, line 24:
```suggestion
        throw new Error(`Type ${type} not implemented for awscdk target.`);
```

libs/wingsdk/src/target-awscdk/bucket.ts
marciocadev: Good catch

I will change `__filename` to `__filename.replace("awscdk", "tf-aws")`

marciocadev: Good catch

marciocadev:
Now Function works

```ts (wing)
bring cloud;
let b = new cloud.Bucket() as "wing-cdk-bucket";
b.add_object("wing-file.txt", "Hello World!");

new cloud.Function(inflight(s:str):str => {
  b.put("inflight-file.txt", "Inside now.");
});
```

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -0,0 +1,108 @@
...
+    return `BUCKET_NAME_${this.node.addr.slice(-8)}`;
+  }
+}
+
+Bucket._annotateInflight("put", {});
+Bucket._annotateInflight("get", {});
+Bucket._annotateInflight("delete", {});
+Bucket._annotateInflight("list", {});
+Bucket._annotateInflight("put_json", {});
+Bucket._annotateInflight("get_json", {});
```
Chriscbr, line 108: Let's add support for public_url
```suggestion
Bucket._annotateInflight("get_json", {});
Bucket._annotateInflight("public_url", {});
```

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -0,0 +1,108 @@
...
+      });
+    }
+    if (ops.includes(cloud.BucketInflightMethods.GET)) {
+      host.addPolicyStatements({
+        effect: Effect.ALLOW,
+        actions: ["s3:GetObject*", "s3:GetBucket*", "s3:List*"],
+        resources: [`${this.bucket.bucketArn}`, `${this.bucket.bucketArn}/*`],
+      });
+    }
+    if (ops.includes(cloud.BucketInflightMethods.LIST)) {
```
Chriscbr, line 63:
```suggestion
    if (ops.includes(cloud.BucketInflightMethods.LIST) || ops.includes(cloud.BucketInflightMethods.PUBLIC_URL)) {
```

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    if (ops.includes(cloud.BucketInflightMethods.PUT)) {
+      host.addPolicyStatements({
+        effect: Effect.ALLOW,
+        actions: ["s3:PutObject*", "s3:Abort*"],
+        resources: [`${this.bucket.bucketArn}`, `${this.bucket.bucketArn}/*`],
+      });
+    }
+    if (ops.includes(cloud.BucketInflightMethods.GET)) {
```
Chriscbr, line 56: or PUT_JSON or GET_JSON

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  public addObject(key: string, body: string): void {
+    new BucketDeployment(this, `S3Object-${key}`, {
+      destinationBucket: this.bucket,
+      sources: [Source.data(key, body)],
+    });
+  }
+
+  /** @internal */
+  public _bind(host: core.IInflightHost, ops: string[]): void {
```
Chriscbr, line 44: I'm thinking that our code for inflights, and anything else that can be reused between AWS targets can go into `libs/wingsdk/src/shared-aws`.

For example, we could create a function `calculateBucketPermissions(arn: string, ops: string[]): PolicyStatements[]` that is shared between both `tf-aws` and `awscdk` to reduce the amount of repetition here.

Chriscbr (approved):
This is amazing work @marciocadev! Looks pretty much ready to ship, I just left a few small comments.

libs/wingsdk/src/target-awscdk/bucket.ts
marciocadev: Nice

marciocadev: Hey @Chriscbr 

I did what you suggested and I think it was really cool. I just had to be careful with the policies because CDK and TF use slightly different objects (CDK policies has `actions` and `resources` keys and Terraform has `action` and `resource`)

docs/04-reference/wingcli.md
```diff
@@ -133,6 +133,24 @@ JavaScript bundles that include inflight code that executes on compute platform
...
+$ export CDK_STACK_NAME="my-project"
+$ wing compile --target awscdk app.w
+```
+
+The output includes both a AWS-CDK configuration file (under `target/<file name>.awscdk`) and
+JavaScript bundles that include inflight code that executes on compute platforms such as AWS Lambda.
+
+You can deploy your stack to AWS by installing the [AWS CDK Toolkit](https://docs.aws.amazon.com/cdk/v2/guide/cli.html) and running:
+```sh
+$ cdk deploy --app target/app.awscdk
```
Chriscbr, line 152:
````suggestion
```sh
$ cdk deploy --app target/app.awscdk
```
````

Chriscbr (approved):

Issue #1578, opened 2023-02-19
failed to install the newest version of winglang


ainvoner:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

`npm i -g winglang`

### I expected this:

winglang newest version to be installed globally while using npm version 8 or 9 major versions

### Instead, this happened:

**npm version 8.6.0**

```
npm ERR! code 1
npm ERR! path C:\Users\ainvoner\AppData\Roaming\npm\node_modules\winglang\node_modules\wasi-js
npm ERR! command failed
npm ERR! command C:\Windows\system32\cmd.exe /d /s /c npx only-allow pnpm
npm ERR! 'only-allow' is not recognized as an internal or external command,
npm ERR! operable program or batch file.
```

**npm version 9.2.0**

```
npm notice 
npm notice New minor version of npm available! 9.2.0 -> 9.5.0
npm notice Changelog: https://github.com/npm/cli/releases/tag/v9.5.0
npm notice Run npm install -g npm@9.5.0 to update!
npm notice 
npm ERR! code 127
npm ERR! path /Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/wasi-js
npm ERR! command failed
npm ERR! command sh -c npx only-allow pnpm
npm ERR! npm WARN exec The following package was not found and will be installed: only-allow@1.1.1
npm ERR! sh: only-allow: command not found
npm ERR! npm notice 
npm ERR! npm notice New minor version of npm available! 9.2.0 -> 9.5.0
npm ERR! npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.5.0>
npm ERR! npm notice Run `npm install -g npm@9.5.0` to update!
npm ERR! npm notice
```

**npm version 9.5.0**
```
npm ERR! code 127
npm ERR! path /Users/ainvoner/.nvm/versions/node/v18.0.0/lib/node_modules/winglang/node_modules/wasi-js
npm ERR! command failed
npm ERR! command sh -c npx only-allow pnpm
npm ERR! sh: only-allow: command not found

```

### Component:

CLI

### Wing Version:

0.5.53

### Node.js Version:

v18.0.0

### Platform(s):

MacOS, Windows

trondhindenes:
I'm getting the same

staycoolcall911:
[...]

PR #1584, opened 2023-02-19
feat: isomorphic wasi (again) and patched wasi-js

MarkMcCulloh:
This PR reintroduces https://github.com/winglang/wing/pull/1571 with 3 additional changes to avoid [the last regression](https://github.com/winglang/wing/issues/1578):
- `patch-package` is used to mess with wasi-js, to remove the preinstall script from their package.json. The patching is performed during packaging, where it's actually effective (for downstream consumers).
- wasi-js is included as a bundled dep for winglang to ensure that the patched `package.json` is appropriately patched in time to avoid any hooks. Normally this isn't needed with patch package, but due to the particular mutation I'm not sure this is avoidable
- Added checks to hangar to ensure we avoid introducing script hooks. I verified that these tests pass with the latest version of the repo but fail the version noted in the regression.

The changes from the original PR are untouched.


apps/wing/patches/wasi-js+1.7.3.patch
```diff
@@ -0,0 +1,12 @@
...
+--- a/node_modules/wasi-js/package.json
++++ b/node_modules/wasi-js/package.json
+@@ -54,7 +54,6 @@
+     "typescript": "^4.8.4"
+   },
+   "scripts": {
+-    "preinstall": "npx only-allow pnpm",
+     "clean": "rm -rf dist",
+     "build": "tsc -b",
+     "tsc": "tsc -w",
```
MarkMcCulloh, line 12: ℹ️ Notable change `#1`

apps/wing/package.json
```diff
@@ -35,8 +36,12 @@
     "is-installed-globally": "^0.4.0",
     "open": "^8.4.0",
     "update-notifier": "^6.0.2",
-    "vscode-languageserver": "^8.0.2"
+    "vscode-languageserver": "^8.0.2",
+    "wasi-js": "^1.7.3"
   },
+  "bundledDependencies": [
+    "wasi-js"
+  ],
```
MarkMcCulloh, line 44: ℹ️ Notable change `#2`

tools/hangar/src/package.setup.ts
```diff
@@ -52,18 +55,33 @@ export default async function () {
...
+  nodeAssert.equal(
+    installResult.exitCode,
+    0,
+    `Failed to install npm deps: \n${installResult.stderr}`
+  );
+  nodeAssert.doesNotMatch(
+    installResult.stdout,
+    />/,
+    `Install contains unexpected script hook: \n${installResult.stdout}`
+  );
```
MarkMcCulloh, line 68: ℹ️ Notable change `#3`

tools/hangar/src/package.setup.ts
```diff
@@ -1,6 +1,7 @@
 import { execa } from "execa";
 import * as fs from "fs-extra";
 import * as path from "path";
+import * as nodeAssert from "node:assert";
```
MarkMcCulloh, line 4: ℹ️  Also added proper assertions here for prettier errors. Note that we can't use vitest assertions here because this is not an actual test.

eladb (approved):

Issue #1265, opened 2023-01-26
Hangar tests for Windows and Mac


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Validate that the hangar tests we run that perform commands on the Wing CLI (like compiling, testing, etc.) also work on Windows and Linux in our CI/CD.

### Use Cases

Make sure Wing works on all platforms in an automation fashion

### Implementation Notes

If the tests take substantially longer to run on these platforms than our current Linux tests, then instead of running them on every pull request commit, we could run them on every commit to the main branch, or daily, etc.

### Component

Compiler

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.63](https://github.com/winglang/wing/releases/tag/v0.5.63).

PR #1601, opened 2023-02-20
chore: add canary test and only run e2e matrix for release builds

MarkMcCulloh:
With this change:
1. In PRs, we only run E2E tests in the same context that we build everything (linux + node18 currently)
2. In release (push to main), we run all E2E tests in the super-matrix (linux/windows/mac + node18/node19)
3. Every hour, we run a small canary test to validate that we can install and run a basic wing file (using the flow advertised on https://www.winglang.io)
  - After this PR merges and it's verified to be working, will add a Better Uptime monitor 

Fixes https://github.com/winglang/wing/issues/1265


.github/workflows/canary.yml
```diff
@@ -0,0 +1,41 @@
...
+      matrix:
+        runner:
+          - ubuntu-latest
+          - windows-latest
+          - macos-latest
+        node:
+          - 18
+          - 19
+        install_command:
+          - npm install -g winglang
```
MarkMcCulloh, line 21: Was initially going to add yarn here as well, but found out that yarn got rid of global installs at some point.
I think maybe this is initially good enough, since `npm install -g winglang` is what we actually tell people to do.

.github/workflows/build.yml
```diff
@@ -124,8 +124,8 @@ jobs:
         with:
           targets: "build,test,package"
           all: true
-          # Exclude E2E testing
-          args: "--exclude=hangar --configuration=release --output-style=stream --verbose"
+          # If pushing, exclude hangar because we're going to run it in a separate job
+          args: "${{ github.event_name = 'push' ?? '--exclude=hangar ' || '' }}--configuration=release --output-style=stream --verbose"
```
Chriscbr, line 128: Maybe we can tweak this slightly to model this in the NX config directly, i.e. `--configuration=prbuild` and `--configuration=release`.

Also can we make `github.event_name == 'push'` into a workflow scoped variable (`RUN_FULL_E2E_TESTS=...`)? I think it could be handy if we could tweak the conditions so we can e.g. opt into the full multi-platform test suite if we want by adding a GitHub label (doesn't have to be in this PR)

.github/workflows/canary.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+      - name: Get Latest Wing Version
+        id: get-version
+        run: |
+          echo version=$(curl -s https://registry.npmjs.org/winglang | jq -r '.["dist-tags"].latest') >> $GITHUB_OUTPUT
```
Chriscbr, line 31:
```suggestion
        run: npm view winglang version
```

MarkMcCulloh: Thanks! Forgot about that one

.github/workflows/build.yml
MarkMcCulloh: > Also can we make github.event_name == 'push' into a workflow scoped variable

Should be able to do that, it's better semantically anyways 👍 

> Maybe we can tweak this slightly to model this in the NX config directly

Not sure if nx is meant to work that way (e.g. using configurations to conditionally run certain targets).

> opt into the full multi-platform test suite if we want by adding a GitHub label

For this and the previous point, I think this should be solved by actually having separate workflows.

Chriscbr (approved):

Issue #1553, opened 2023-02-15
Use npm 9 in our CI/CD


Chriscbr:
This change should not affect how consumers use Wing

MarkMcCulloh:
Explored this issue in https://github.com/winglang/wing/pull/1603 and found that `npm pack` did not include bundled dependencies in the resulting tarball with npm 9, causing missing `@aws-sdk` packages failing during preflight.

MarkMcCulloh:
Closing, moved to pnpm

PR #1603, opened 2023-02-20
chore: node19 and npm9 for development

MarkMcCulloh:
- Updated docs
- Update volta pin, and pinned all `package.json`s in the repo to make sure they're consistent (volta looks for the nearest `package.json`)
- Added a .npmrc file to all projects that required `--install-links=false` to be set and removed that flag from the root. This change was made because it helps prevent lockfile thrashing if you `npm install` outside of the root installer.
- Added warning checks to hangar, just to make sure we don't introduce ugly errors (such as node engine warnings)

Note: One of the transitive azure dependencies of wingsdk has an `engine` entry that doesn't include node19. You will see the warning in local dev, but users shouldn't see it during installation. Here's an issue to track if needed https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/5700. Doesn't seem like a big deal.


tools/hangar/src/package.setup.ts
```diff
@@ -66,6 +66,11 @@ export default async function () {
     />/,
     `Install contains unexpected script hook: \n${installResult.stdout}`
   );
+  nodeAssert.doesNotMatch(
```
Chriscbr, line 69: (nit) Any reason this can't just be `assert`?

Chriscbr (approved):

tools/hangar/src/package.setup.ts
MarkMcCulloh: So originally when I tried `assert` I remember there being some issue where it was using the `vitest` `assert` global instead. I figured vitest was hijacking something.

I must have just made a mistake before because it works fine now, thanks for bringing it up.

MarkMcCulloh:
https://github.com/winglang/wing/issues/1553#issuecomment-1437813784

PR #1611, opened 2023-02-21
fix: corrupted wing terminal, inconsistant dev node version, other small changes

MarkMcCulloh:
During the process of trying to finish https://github.com/winglang/wing/pull/1603, I found some nice improvements to make
- Fixed the issue where running `wing` would corrupt the current terminal (i.e. intercepting control characters)
- If esbuild fails during inflight bundling, the error is thrown is a way that makes it more clear it's an issue with function bundling
- `wing` process now exits with `1` when it fails in preflight (previously 0)
  - This also improves error reporting in hangar, because previously hangar would fail due to synthesized files not being found. Now it properly detects the failed (`1`) process and can show you the output
- Ensure there are no warnings in hangar install
- Use `files` in wingsdk `package.json` instead of `.npmignore`
  - This convention makes so much more sense than `.npmignore`, it makes sure we don't include extraneous stuff in the package
- Made sure `volta` config is present in all package.json files, otherwise sometimes your node version can instead revert to your default
- Added .npmrc to projects that rely on symlinked installs and removed the corresponding flag from the room package installer thing. 
  - This should help avoid lockfile thrashing due to `npm install` being run outside of the root.
- Added warning checks to hangar, just to make sure we don't introduce ugly errors (such as node engine warnings)
- wingsdk mutation workflow now uses volta for its node version (to be consistent with the main workflow)


libs/wingsdk/src/cloud/function.ts
```diff
@@ -96,20 +96,26 @@ export abstract class FunctionBase extends Resource implements IInflightHost {
...
+        nodePaths: [join(__dirname, "..", "..", "node_modules")],
+        target: "node16",
+        platform: "node",
+        absWorkingDir: tempdir,
+        outfile,
+        minify: false,
+        logLevel: "silent",
+        external: ["aws-sdk"],
+      });
+    } catch (e) {
```
yoav-steinberg, line 116: I this `catch` all we really needed to avoid the terminal garbling?

yoav-steinberg (approved):

libs/wingsdk/src/cloud/function.ts
MarkMcCulloh: It's the loglevel: "silent" above

Chriscbr: Looks like the issue is still present for me @MarkMcCulloh

PR #1616, opened 2023-02-21
fix: cache npm dependencies to speed gh actions

raywonkari:
Future work: add sub paths as this is a monorepo by utilising the `cache-dependency-path` parameter.
More on that [here](https://github.com/actions/setup-node#caching-global-packages-data)


MarkMcCulloh:
We use https://github.com/bahmutov/npm-install, so npm packages should already be cached. 
Also we don't use `actions/setup-node` everywhere (volta is used during actual builds).

The caching could be useful in the canary workflow, but it's also possible that defeats the purpose of it (to warn us about issues install wing). 
What do you think?

PR #1622, opened 2023-02-22
feat(sdk): bucket.try_get

Chriscbr:
Adds a `try_get` method to `cloud.Bucket` which can be used instead of `get` when an object may be expected to not exist in the bucket. It's important to note that calling `try_get` is not simply the same as wrapping `get` with a try/catch block and handling any error, because `try_get` can still throw if an unrelated error occurred (such as a network error, permissions error, etc.).

In order to implement this, I had to overcome a technical hurdle with JSII -- JSII does not allow asynchronous methods that return an "optional" (JSII defines this as a type that is union-ed with `undefined` or `null`). In other words, returning `Promise<string | undefined>` caused a compiler error. I can't really fathom why given that `Promise<string | number>` *is* allowed. To work around this I patched over some code in `jsii` to allow this API. Oh well. 

TODO: test on Azure


Chriscbr:
I've been having some issues trying to test on Azure (I don't know how the Azure SDK throws errors at the moment) so for now it's not supported there yet, but it works on AWS and the simulator.

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -56,9 +57,16 @@ export class BucketClient implements IBucketClient {
...
     ).toString();
 
+    // TODO: throw InflightError and enable support for tryGet
+
     return downloaded;
   }
 
+  public async tryGet(key: string): Promise<string | undefined> {
+    key;
+    throw new Error("Method not implemented.");
```
hasanaburayyan, line 67: open an issue for later?

hasanaburayyan (commented):
I didn't get to finish reviewing, but so far this looks dope. Ive got questions about the patch. Maybe we can do a quick huddle tomorrow if you dont mind answering some dumb questions 😅

ekeren:
@Chriscbr, why did you decide to go and implement the `try_get`?

Chriscbr:
Keep

Chriscbr:
The code here is pretty stale, closing for now

Issue #1487, opened 2023-02-08
Remove code for supporting hidden namespaces


Chriscbr:
In order to support importing different namespaces from the same library (such as the `cloud` and `fs` namespaces of the SDK), a mechanism was added to the compiler where symbol environments can hide namespaces, and `lookup` methods can choose to include or skip results that are within hidden namespaces.

Once the changes in https://github.com/winglang/wing/pull/1397 are merged, all imported types will be created in a hierarchy reflecting how the types are organized within the source libraries, and module declarations will just be pointers into this data structure. For example, `bring foo as baz` adds a symbol to the global namespace named `baz` that points to  `Types.libraries.get("foo")` (pseudocode), and in the future if a syntax like `bring foo.bar as baz` is added, then this would add a symbol pointing to `Types.libraries.get("foo").get("bar")`.

Since no unused symbols need to be added to the global environment anymore (like a `core` symbol pointing to a hidden namespace), I think these parts of the compiler code can be removed.

staycoolcall911:
@Chriscbr - just making sure - this hidden namespaces mechanism was added to also bring nested imports in packages (e.g. `bring cloud` should also bring `core.Resource` IIRC).
I'm assuming you checked with @yoav-steinberg that this could be removed, right?

Chriscbr:
Yep. We have some unit tests that check the different cases for importing libraries, so I think as long as they continue to pass, it should be ok

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.68](https://github.com/winglang/wing/releases/tag/v0.5.68).

PR #1631, opened 2023-02-22
feat: bring cdktf

Chriscbr:
Fixes some limitations with the compiler's jsii importer so that it's now possible to import types from most CDKTF libraries (tested with `@cdktf/provider-aws`).

Notes:
- There were some issues with the jsii importer not unhiding namespaces correctly -- the plan is still to just remove this namespace-hiding logic (see #1487), but to avoid mixing concerns I just fixed the issues I found in this PR.
- Since the compiler doesn't support interfaces yet, I imported these types as `anything`. It's not possible skip importing them because many other resource methods reference interfaces in their parameters (e.g. "cdktf.IResolvable") and our jsii importer doesn't support gracefully failing in these kinds of situations.
- Fixed an issue with symbol shadowing (copied from #1610)

### DIFF
--- a/examples/tests/valid/bring_cdktf.w
+++ b/examples/tests/valid/bring_cdktf.w
@@ -1,0 +1,9 @@
+bring "@cdktf/provider-aws" as aws;
+
+new aws.s3Bucket.S3Bucket(
+  bucket_prefix: "hello",
+  versioning: aws.s3Bucket.S3BucketVersioning {
+    enabled: true,
+    mfa_delete: true,
+  },
+) as "Bucket";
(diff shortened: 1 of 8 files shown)


examples/tests/valid/bring_cdktf.w
```diff
@@ -0,0 +1,9 @@
+bring "@cdktf/provider-aws" as aws;
+
+new aws.s3Bucket.S3Bucket(
```
eladb, line 3: tears of joy! 😂

eladb (approved):
👏

Issue #470, opened 2022-11-06
Operators: advanced behavior


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#63-operators

Missing the following functionality:
- [x] Floor Division `\`
- [x] Power `^`
- [x] Equality https://github.com/monadahq/winglang-spec/blob/rev1.3/README.md#636-equality

mortax:
ah ha. so this actually explains why using floor division (or exponents) don't compile:

```
$ cat wing.w         
let floor = 10 \ 5;

print("${floor}");
$ wing compile wing.w
Compilation failed with 2 errors
Error at wing.w:1:16 | Unknown parser error.
Error at wing.w:1:16 | Unknown parser error.
{1} $ 
```

is there an ETA when floor division and exponent support will exist? it was confusing since they are [documented](https://docs.winglang.io/reference/spec#632-logical-operators).

staycoolcall911:
Hi @mortax, thank you for your question.

Wing is still in alpha. The [language spec](https://docs.winglang.io/reference/spec) defines our beta, where Wing could be used to build some real world apps.
You can see the progress on our spec implementation and the roadmap [here](https://github.com/orgs/winglang/projects/1/views/1).
I'll also frankly add that as an early alpha project, we still don't have an ETA for our beta.

I'm actually going to speak about that and other aspects of how-we-work on Wing today on our Twitch show, "The Wingly Update Name TBD".
Click [here](https://t.winglang.io/wingly02) on Sunday Jan 8th, 1:30pm ET to view the show and ask us questions live!

mortax:
perfect. that answers all of my current questions. i'll watch the roadmap and have another look once the aspects i need are implemented. cheers!

staycoolcall911:
Both operators were implemented (thanks @flyingImer) and equality is detailed in a different issue. Closing this issue.

PR #1640, opened 2023-02-23
feat(compiler): add floor division support

flyingImer:
Get `floor division` implemented: https://github.com/winglang/wing/issues/470


flyingImer:
still need to clean up the jsify part. Ideas welcome!

libs/wingc/src/jsify.rs
```diff
@@ -371,6 +371,13 @@ impl<'a> JSifier<'a> {
 					BinaryOperator::Sub => "-",
 					BinaryOperator::Mul => "*",
 					BinaryOperator::Div => "/",
+					BinaryOperator::FloorDiv => {
+						return format!(
+							"Math.floor({} / {})",
```
staycoolcall911, line 376: hmmm... no special ideas how to simplify that @flyingImer. @yoav-steinberg - do you have any idea?

flyingImer: LMK

Will keep it in back of my mind as digging more into the art part of coding rust

examples/tests/valid/expressions_binary_operators.w
```diff
@@ -6,3 +6,5 @@ let xyz = ((y * y) / (x * x)) * z;
 let xf = 1.0;
 let yf = (-20.22) * xf;
 let zf = (xf + (yf) - -0.01);
+let xyzf = 501 \ (99 + 1);
+assert(xyzf == 5);
```
yoav-steinberg, line 10: I'm assuming we're aiming for integer division here. So lets also add a test for negative numbers which will show that using `Math.floor(..)` is a bug here.
```suggestion
let xyzf = 501 \ (99 + 1);
assert(xyzf == 5);
let xyznf = -501 \ (99 + 1);
assert(xyznf == -5);
```

libs/wingc/src/jsify.rs
yoav-steinberg, line 376: to deal with negatives
```suggestion
							"Math.trunc({} / {})",
```

yoav-steinberg: Maybe do a `let js_left = self.jsify_expression(left, phase);` and `let js_right = self.jsify_expression(right, phase);` before entering this `match`, to de-dup the call to `jsify_expression`. But other than that this seems fine.

yoav-steinberg (changes requested):
nice :rocket:

libs/wingc/src/jsify.rs
flyingImer: TIL

examples/tests/valid/expressions_binary_operators.w
```diff
@@ -6,5 +6,13 @@ let xyz = ((y * y) / (x * x)) * z;
 let xf = 1.0;
 let yf = (-20.22) * xf;
 let zf = (xf + (yf) - -0.01);
+let fxzy = 5 ^ (1^5);
+assert(fxzy == 5);
```
yoav-steinberg, line 10: According to the spec this is the "exponent" operator. I'm not sure what this means. It's probably supposed to be a **power** operator. But in either case what you added here acts like the javascript `^` operator which is a "bitwise xor" operator. So to clarify:
```
// JavaScript
7^3 = 4 // Bitwise XOR: binary 111 xor binary 011 = binary 100 = 4
// Power  - this is what I think we want with the ^ operator
7^3 = Math.pow(7,3) = 343
// Exponent - x^y - this is unclear. It may mean the exponent of x but then this is a unary operator and there's no need for y
Math.exp(7)
```
@staycoolcall911 I suggest we either delete `^` from the spec for now, or fix the spec to say it's a power operator (and not exponent).

yoav-steinberg (changes requested):
This seems fine, see my comment re "exponent". I suggest we either clarify this with @staycoolcall911 or just remove it from the PR so we can get it merged quickly.

examples/tests/valid/expressions_binary_operators.w
flyingImer: @yoav-steinberg to be clear, the "exponent" operator (https://github.com/winglang/wing/pull/1654) was jsified as `**`, which is function-wise equivalent to `Math.pow`. Tho the meaning of syntax `^` in the spec is a bit different from the javascript. 

I think it can be a separate PR if @staycoolcall911 thinks "exponent" the spec and the implementation should be removed. WDYT

staycoolcall911: Good catch @yoav-steinberg.

I suggest:
1. Since this PR is not related to "exponent", I think @yoav-steinberg can approve it as is ("exponent" was actually added in a previous PR #1654 which is already merged in).
2. @flyingImer - could you please add another PR to fix the name of `exponent` to `power` in the spec? 
I'd say please also change the operator from `^` to `**`, because `^` is confusing. Thanks!

Happy to hear if there are other opinions/suggestions.

yoav-steinberg: And if this PR adds a test for power (exponen) lets make it a test that really shows this is power:
```suggestion
let fxzy = 5 ^ (2^3);
assert(fxzy == 390625);
```

flyingImer: @yoav-steinberg will make that happen in the following PR

yoav-steinberg (approved):
:+1:

examples/tests/valid/expressions_binary_operators.w
staycoolcall911: Resolved the conversation and @yoav-steinberg approved your PR. Thank you for your patience and for following up @flyingImer 🙏

Issue #1644, opened 2023-02-23
Memory corruption in SymbolEnv during post-compilation traversal


MarkMcCulloh:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

In the language server, we introspect Scopes and occasionally must do lookups in Symbol environment. As-is, this usually works fine.

This is a scenario that does not work:

```
if true {
  print("duh");
}
```

### I expected this:

`print` is defined at the root scope and is available in the inner scope. We should be able to lookup `print` in the inner env because the symbol is available there.

### Instead, this happened:

The data pointed to by env.parent had been lost, causing it to either be corrupted or gone. This causes lookups in the inner scope to be take a long time or occasionally panic.

### Component:

Compiler, IDE Extension

monadabot:
Congrats! :rocket: This was released in [Wing 0.6.6](https://github.com/winglang/wing/releases/tag/v0.6.6).

PR #1646, opened 2023-02-23
fix: `visitor` for lsp and use user's `wing` binary in vscode

### LINKED ISSUE #1634: `while` is not highlighted as a keyword
### Community Note

### I tried this:

```js
bring cloud;

resource A {
  cache_hits: cloud.Counter;
  init() {
    this.cache_hits = new cloud.Counter();
  }
  inflight hits(): num {
    return this.cache_hits.peek();
  }
  inflight reset_cache() {
    while this.cache_hits.peek() > 0 {
      this.cache_hits.dec();
    }
  }
}
```

### I expected this:

`while` to be highlighted in the same color in my IDE as `bring` or `return`

### Instead, this happened:

Screenshot 2023-02-22 at 3 26 53 PM

`while` is a different color (using Wing VS Code extension)

### Component:

Compiler

### Wing Version:

0.5.65

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

_No response_

MarkMcCulloh:
Main Changes:
- Added symbol visiting in base Visitor
  - **(?) Any perf concerns?** Most visitors in the compiler probably don't need this, but it's very useful for the lsp
- Deleted the other ad-hoc traversal for lsp, use Visitor instead
- Stop using npx for wing cli in vscode. It's slow/confusing and causes a disconnect between the local compiler and vscode
  - prompts added to aid the user in installing wing if not available

Misc:
- Removed RwLock from lsp. We are running the lsp single-threaded so there's no point.
- Cleaned up the tm grammar for vscode a little more
- Use .replace for refcell instead of pointer assignment, this seems to be the technically correct way to do what's happening there

Fixes #1634

Followups Created:
- https://github.com/winglang/wing/issues/1644


apps/vscode-wing/src/extension.ts
```diff
@@ -21,11 +26,62 @@ export async function activate(context: ExtensionContext) {
...
+                );
+              })
+              .catch((e) => {
+                void window.showErrorMessage(
+                  `Failed to install Wing v${extVersion}: ${e}`
+                );
+                wingBin = "npx";
+              });
+          }
+        );
```
Chriscbr, line 75: Extract this into a separate function like `interactiveWingInstall()`

apps/vscode-wing/src/extension.ts
```diff
@@ -21,11 +26,62 @@ export async function activate(context: ExtensionContext) {
...
-    : ["-y", "-q", `winglang@${context.extension.packageJSON.version}`, "lsp"];
+  const extVersion = context.extension.packageJSON.version;
+  const configuredWingBin = workspace
+    .getConfiguration(CFG_WING)
+    .get<string>(CFG_WING_BIN, "wing");
+  let wingBin = env.WING_BIN ?? configuredWingBin;
+
+  if (wingBin !== "npx") {
+    // check if wing is installed
+    const result = which.sync(wingBin, { nothrow: true });
```
Chriscbr, line 37: I'd consider extracting this one line into a function like `assertWingIsInstalled()` to avoid the need for the comment.

libs/wingc/src/lsp/document_symbols.rs
```diff
@@ -1,13 +1,154 @@
...
+impl DocumentSymbolVisitor {
+	pub fn new() -> Self {
+		Self {
+			document_symbols: vec![],
+		}
+	}
+}
+
+impl Visit<'_> for DocumentSymbolVisitor {
+	fn visit_stmt(&mut self, statement: &Stmt) {
```
Chriscbr, line 21: Cool use of the visitor 👍 Curious if based on the name it's possible to use the visit_symbol method you added, or if there's a reason it wasn't used here?

libs/wingc/src/lsp/hover.rs
```diff
@@ -27,45 +203,52 @@ pub unsafe extern "C" fn wingc_on_hover(ptr: u32, len: u32) -> u64 {
...
 		let parse_result = files.get(&params.text_document_position_params.text_document.uri.clone());
 		let parse_result = parse_result.unwrap();
 
 		let position = params.text_document_position_params.position;
 
 		let root_scope = &parse_result.scope;
-		let root_env_ref = root_scope.env.borrow();
-		let root_env = root_env_ref.as_ref().unwrap();
+		let root_env = root_scope.env.borrow();
+		let root_env = root_env.as_ref().unwrap();
```
Chriscbr, line 213:
```suggestion
		let root_env = root_env.as_ref().expect("All scopes should have a symbol environment");
```

libs/wingc/src/lsp/hover.rs
```diff
@@ -27,45 +203,52 @@ pub unsafe extern "C" fn wingc_on_hover(ptr: u32, len: u32) -> u64 {
...
 				} else {
 					None
 				}
 			});
 			let span = if let Some(_) = reference {
-				&expr.unwrap().span
+				&expr.expect("Missing type for expression").span
 			} else {
 				&symbol.span
 			};
```
Chriscbr, line 237: What is the reason for using the parent Expr's span instead of the symbol's span? (maybe consider adding a comment)

libs/wingc/src/lsp/hover.rs
```diff
@@ -76,7 +259,7 @@ pub fn on_hover<'a>(params: lsp_types::HoverParams) -> Option<Hover> {
```
Chriscbr, line 259: Maybe we can extract some of this into separate functions just to make the code a bit more self-documenting / reduce the amount of nesting. Here's a sketch of one way it could look:
```rust
if let Ok(symbol_lookup) = symbol_lookup {
  format_symbol_with_type_info(symbol_name, symbol_kind)
} else if let Some(reference) = reference {
  format_reference_symbol(symbol_name, reference)
} else {
  format_unknown_symbol(symbol_name)
}
```

libs/wingc/src/lsp/hover.rs
```diff
@@ -1,13 +1,189 @@
...
+			return;
+		}
+
+		// Handle situations where symbols are actually defined in inner scopes
+		match &node.kind {
+			StmtKind::ForLoop {
+				iterator,
+				iterable,
+				statements,
+			} => {
```
Chriscbr, line 72: Ah, I see why this is needed... I guess it has to do with the fact that this statement is introducing a new symbol, but that symbol isn't part of the parent scope like the visitor expects/wants in the general case. Feels like a modeling issue but I'm not quite sure how to fix it

libs/wingc/src/lsp/hover.rs
```diff
@@ -27,45 +203,52 @@ pub unsafe extern "C" fn wingc_on_hover(ptr: u32, len: u32) -> u64 {
 pub fn on_hover<'a>(params: lsp_types::HoverParams) -> Option<Hover> {
 	FILES.with(|files| {
 		let files = files.borrow();
-		let files = files.read().unwrap();
 		let parse_result = files.get(&params.text_document_position_params.text_document.uri.clone());
 		let parse_result = parse_result.unwrap();
```
Chriscbr, line 207: expect()

libs/wingc/src/lsp/document_symbols.rs
MarkMcCulloh: DocumentSymbol is more concerned about how symbols are introduced rather than every usage. Currently, I decided to only handle cases with statements, which are probably the most common way this happens right now.

libs/wingc/src/lsp/hover.rs
MarkMcCulloh: image
This is maybe controversial since TS and rust don't do this, but I kinda like that this highlights the entire relevant part of a reference rather than just the field.

I'll add a comment

libs/wingc/src/lsp/hover.rs
MarkMcCulloh: Agreed, I went over this whole part of the code again and I think it's a lot better now.

Chriscbr (approved):

PR #1652, opened 2023-02-23
chore: simplify binding model in wingc/sdk

### LINKED ISSUE #1605: Refactor SDK inflight bindings API to remove resources/data distinction
Today when the compiler generates preflight code in JS, it separates resources in data into different categories. This is no longer needed since we treat them the same in the SDK (they are all handled by [serializeImmutableData](https://github.com/winglang/wing/blob/e7551dcda97eab78aa17a9bc08385ff78d8503fe/libs/wingsdk/src/core/internal.ts#L86), which should probably be renamed).

So, instead of generating:

```js
const fn = new cloud.Function(this,"test",new $stdlib.core.Inflight(this, "$Inflight1", {
  code: $stdlib.core.NodeJsCode.fromFile(require('path').resolve(__dirname, "proc.f461cbb2996b2e6c0a4f548706c4e2ed3844d744058fa69f536f5929c7fe2371/index.js").replace(/\\/g, "/")),
  bindings: {
    resources: {
      bucket: {
        resource: bucket,
        ops: ["delete","get","list","put"]
      },
    },
    data: {
      name: name,
    },
  }
}));
```

It should generate:

```js
const fn = new cloud.Function(this,"test",new $stdlib.core.Inflight(this, "$Inflight1", {
  code: $stdlib.core.NodeJsCode.fromFile(require('path').resolve(__dirname, "proc.f461cbb2996b2e6c0a4f548706c4e2ed3844d744058fa69f536f5929c7fe2371/index.js").replace(/\\/g, "/")),
  bindings: {
    bucket: {
      obj: bucket,
      ops: ["delete","get","list","put"]
    },
    name: {
      obj: bucket,
    },
  }
}));
```

(aside: Once we implement a compiler pass that transforms anonymous closures into ad-hoc resource declarations, we should not need to generate `core.Inflight` anymore, so that would also close out this issue)

Chriscbr:
Closes #1605

### DIFF
--- a/docs/04-reference/wingsdk-api.md
+++ b/docs/04-reference/wingsdk-api.md
@@ -2006,48 +2006,48 @@ The maximum amount of time the function can run.
 
 ---
 
-### InflightBindings <a name="InflightBindings" id="@winglang/sdk.core.InflightBindings"></a>
+### InflightBinding <a name="InflightBinding" id="@winglang/sdk.core.InflightBinding"></a>
 
-Inflight bindings.
+An inflight binding.
 
-#### Initializer <a name="Initializer" id="@winglang/sdk.core.InflightBindings.Initializer"></a>
+#### Initializer <a name="Initializer" id="@winglang/sdk.core.InflightBinding.Initializer"></a>
 
 ```wing
 bring core;
 
-let inflight_bindings = core.InflightBindings{ ... }
+let inflight_binding = core.InflightBinding{ ... }
 ```
 
 #### Properties <a name="Properties" id="Properties"></a>
 
 | **Name** | **Type** | **Description** |
 | --- | --- | --- |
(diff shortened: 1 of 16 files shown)


MarkMcCulloh (approved):
LGTM

This is a super satisfying change, the uniformity makes much more sense.

(noticed you marked Yoav as a reviewer, so I'll throw on the label to give time for him to review if you'd like)

Chriscbr:
Let's just merge it in while the code is fresh - @yoav-steinberg if you have comments still don't hesitate to leave them here and I can address them in a follow up PR if needed 🙂

eladb (commented):
NICE! Thanks for cleaning up after me!

yoav-steinberg:
Noticed this a bit late, but this is very good. Thanks.

Issue #1613, opened 2023-02-21
Env variables are not appended to the final CDKTF output


raywonkari:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

- Create test.w file with content pasted below
- Generate CDK output by running `wing compile --target tf-aws test.w`
- Run `cat target/test.tfaws/main.tf.json` to see the JSON

```
bring cloud;

let my_bucket = new cloud.Bucket();

let function_props = cloud.FunctionProps{
  env: { "test": "value" },
  timeout: 2m,
  memory: 512
};

new cloud.Function(inflight() => {
  my_bucket.put("hello", "world");
}, function_props);
```

### I expected this:

The lambda function definition should contain 3 env vars

```
"environment": {
  "variables": {
    "BUCKET_NAME_d755b447": "${aws_s3_bucket.root_cloudBucket_4F3C4F53.bucket}",
    "WING_FUNCTION_NAME": "cloud-Function-c8d2eca1",
    "test": "value"
  }
},
```

### Instead, this happened:

The lambda function definition has only 2 env vars

```
"environment": {
  "variables": {
    "BUCKET_NAME_d755b447": "${aws_s3_bucket.root_cloudBucket_4F3C4F53.bucket}",
    "WING_FUNCTION_NAME": "cloud-Function-c8d2eca1"
  }
},
```

### Component:

SDK

### Platform(s):

MacOS

Chriscbr:
I believe the reason for this is that the compiler is currently translating `Map` objects into actual [JavaScript `Map` objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) - which is not what the SDK expects.

In theory this makes complete sense (a map is a map, right?). In practice, JavaScript is weird -- and plain objects are what most JS libraries use for representing open-ended bags of properties. In particular, JSII requires that the public APIs in our SDK use `Record<string, T>` TypeScript types in public APIs and not the TypeScript `Map<string, T>`. So code-generating `Map` objects bites us because the SDK's JavaScript/TypeScript code is written to expect plain JS objects, but the Wing compiler is producing JS `Map` objects instead (which are not interoperable).

This problem also won't just affect SDK, it will affect every JSII/CDKTF library our users try to import. So I think plain old JS objects are the only way to go.

Chriscbr:
Code that generates `Map` statements:
[...]

PR #1653, opened 2023-02-24
feat(compiler): change maps to js objects

hasanaburayyan:
These changes update the generated preflight of wing maps to use JavaScript objects instead of Maps. Which addresses the bug seen in: https://github.com/winglang/wing/issues/1613

**Note**: This changes the wing Map API a bit to have to use `size()` method rather than accessor `.size` this can be reverted back once the macro system can be used on getter/setter methods. Maybe when we introduce midlife crisis macros 😁 

These changes were brought to you by the absolutely freaking fantastic Macro system.

**Misc**
1. Also fixed bug where compiler would only create immutable maps even if using MutMap<>


examples/tests/valid/capture_containers.w
```diff
@@ -14,7 +14,7 @@ let handler = inflight (s: str): str => {
   assert(my_set.size == 2);
 
   assert(my_map.has("world"));
-  assert(my_map.size == 2);
+  assert(my_map.size() == 2);
```
eladb, line 17: Do we have a tracking issue to fix this? This is not the API we want.

libs/wingc/src/jsify.rs
```diff
@@ -411,14 +411,17 @@ impl<'a> JSifier<'a> {
...
 			ExprKind::MapLiteral { fields, .. } => {
-				format!(
-					"Object.freeze(new Map([{}]))",
-					fields
-						.iter()
-						.map(|(key, expr)| format!("[ \"{}\", {} ]", key, self.jsify_expression(expr, phase)))
-						.collect::<Vec<String>>()
-						.join(", ")
-				)
+        let f = fields
```
eladb, line 414: indentation issue...

libs/wingc/src/jsify.rs
```diff
@@ -411,14 +411,17 @@ impl<'a> JSifier<'a> {
...
+          .iter()
+          .map(|(key, expr)| format!("\"{}\":{}", key, self.jsify_expression(expr, phase)))
+          .collect::<Vec<String>>()
+          .join(",");
+
+				if is_mutable_collection(expression) {
```
eladb, line 420: Is it possible that MutMap did not work as a mutable collection before?

examples/tests/valid/mut_container_types.w
```diff
@@ -33,6 +33,6 @@ let m2: MutMap<num> = MutMap<num>{"hello": 123};
 let m3 = MutMap<cloud.Bucket>{"b1": bucket1, "b2": bucket2};
 let m4: MutMap<str> = m1;
 assert(m1.has("hello"));
```
eladb, line 35: can you add some asserts that mutate the maps?

libs/wingsdk/src/std/map.ts
```diff
@@ -12,9 +12,14 @@ import { T1 } from "./util";
 export class ImmutableMap {
   /**
    * Returns the number of elements in the map.
+   *
+   * TODO: For now this has to be a method rather than a getter as macros only work on methods (for now)
```
eladb, line 16: add reference to tracking issue please

libs/wingsdk/src/std/map.ts
```diff
@@ -23,28 +28,33 @@ export class ImmutableMap {
    * If the value that is associated to the provided key is an object, then you will get a reference
    * to that object and any change made to that object will effectively modify it inside the map.
    *
+   * @macro $self$[$args$]
```
eladb, line 31: Add a test that verifies this is working (map of maps):

```js
map.get("foo").get("bar")
```

libs/wingsdk/src/std/map.ts
```diff
@@ -72,29 +82,37 @@ export class ImmutableMap {
...
    */
   public clear(): void {
-    throw new Error("Abstract");
+    throw new Error("Macro");
   }
 
   /**
    * Removes the specified element from a map.
+   *
+   * @macro delete $self$[$args$]
```
eladb, line 94:
```suggestion
   * @macro (delete $self$[$args$])
```

libs/wingsdk/src/std/map.ts
```diff
@@ -72,29 +82,37 @@ export class ImmutableMap {
 export class MutableMap extends ImmutableMap {
   /**
    * Removes all elements
+   * @macro $self$={}
```
eladb, line 85: does this work? Do we have a test for this?

libs/wingsdk/src/std/map.ts
```diff
@@ -72,29 +82,37 @@ export class ImmutableMap {
...
   public delete(key: string): boolean {
     key;
-    throw new Error("Abstract");
+    throw new Error("Macro");
   }
 
   /**
    * Adds or updates an entry in a Map object with a specified key and a value.
+   *
+   * TODO: revisit this macro after we support indexed params
```
eladb, line 107: What do you mean by indexed params?
+ every TODO must have a tracking issue ref

eladb (changes requested):

libs/wingsdk/src/std/map.ts
hasanaburayyan: Great catch, I thought I saw a test for this but my eyes must have mixed up `copy` and `clear` This in-fact did NOT work because you cant reassign const 😅 I updated it to iterate the keys and call delete. Also added test to confirm.

libs/wingsdk/src/std/map.ts
hasanaburayyan: Tracking issue added. I meant to say indexed args. `$args[0]$` type of thing

libs/wingsdk/src/std/map.ts
```diff
@@ -72,29 +82,38 @@ export class ImmutableMap {
...
+    throw new Error("Macro");
   }
 
   /**
    * Removes the specified element from a map.
+   *
+   * @macro (delete $self$[$args$])
+   *
    * @param key The key
    * @returns true if the element was in the map
```
MarkMcCulloh, line 98: the return value of the `delete` operator is different, it'll return true even if the property doesn't exist. In fact, it may even throw in some cases if we start using `strict` mode.

Maybe this to make it more wishy-washy?

```suggestion
   * @returns true if the given key is no longer present
```

Not ideal, but idk how much wrapper code we want to make for this usecase.

libs/wingsdk/src/std/map.ts
```diff
@@ -55,7 +65,7 @@ export class ImmutableMap {
   /**
    * Create an immutable shallow copy of this map
    *
-   * @macro Object.freeze(new Map($self$))
+   * @macro Object.freeze({...$self$})
```
MarkMcCulloh, line 68: `$self$` could be a multi-factor expression and it's not emitted with parens. This is probably good just-in-case.

```suggestion
   * @macro Object.freeze({...($self$)})
```

libs/wingsdk/src/std/map.ts
```diff
@@ -23,28 +28,33 @@ export class ImmutableMap {
...
     key;
-    throw new Error("Abstract");
+    throw new Error("Macro");
   }
 
   /**
    * Create a mutable shallow copy of this map
    *
-   * @macro new Map($self$)
+   * @macro {...$self$}
```
MarkMcCulloh, line 57:
```suggestion
   * @macro {...($self$)}
```

libs/wingsdk/src/std/map.ts
```diff
@@ -72,29 +82,38 @@ export class ImmutableMap {
...
    */
   public clear(): void {
-    throw new Error("Abstract");
+    throw new Error("Macro");
   }
 
   /**
    * Removes the specified element from a map.
+   *
+   * @macro (delete $self$[$args$])
```
MarkMcCulloh, line 95:
```suggestion
   * @macro (delete ($self$)[$args$])
```

libs/wingsdk/src/std/map.ts
```diff
@@ -23,28 +28,33 @@ export class ImmutableMap {
...
   public get(key: string): T1 {
     key;
-    throw new Error("Abstract");
+    throw new Error("Macro");
   }
 
   /**
    * Returns a boolean indicating whether an element with the specified key exists or not.
+   *
+   * @macro ($args$ in $self$)
```
MarkMcCulloh, line 44:
```suggestion
   * @macro ($args$ in ($self$))
```

libs/wingsdk/src/std/map.ts
MarkMcCulloh, line 31:
```suggestion
   * @macro ($self$)[$args$]
```

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -17,10 +17,11 @@ Error at ../../../../examples/tests/invalid/bring_jsii.w:4:1 | Cannot find modul
 `;
 
 exports[`capture_mutables.w > wing test (--target sim) - invalid > stderr 1`] = `
-"Compilation failed with 5 error(s)
+"Compilation failed with 6 error(s)
 Error at ../../../../examples/tests/invalid/capture_mutables.w:8:10 | Cannot reference 'a' of type 'MutArray<str>' from an inflight context
 Error at ../../../../examples/tests/invalid/capture_mutables.w:10:10 | Cannot reference 's' of type 'MutSet<num>' from an inflight context
 Error at ../../../../examples/tests/invalid/capture_mutables.w:12:10 | Cannot reference 'm' of type 'MutMap<bool>' from an inflight context
+Error at ../../../../examples/tests/invalid/capture_mutables.w:12:20 | Expected type to be \\"num\\", but got \\"(ImmutableMap): num\\" instead
```
MarkMcCulloh, line 24: This doesn't look right, that test is still using `.size`. It should be updated so it's not invalid for the wrong reason 😄

MarkMcCulloh: In general I would go through all the examples and make sure they're updated, unless it's the intent of the test.
Docs too, in case they're used there. Even though this is temporary we don't want to convey the wrong info.

MarkMcCulloh (approved):
LGTM, minor comments (mostly about parentheses)

examples/tests/invalid/immutable_container_types.w
```diff
@@ -0,0 +1,4 @@
+let m1 = Map<str>{"a": "hi"};
+
+m1.set("a", "bye");
+// ^ Unknown symbol "set" (TODO: better error message https://github.com/winglang/wing/issues/1660)
```
eladb, line 4: 😙

eladb, line 4:
```suggestion
// ^^^ Unknown symbol "set" (TODO: better error message https://github.com/winglang/wing/issues/1660) 
```

libs/wingc/src/jsify.rs
eladb: So it basically worked by accident

eladb (approved):
Very nice

PR #1654, opened 2023-02-24
feat(compiler): add exponent support

flyingImer:
Implement exponent mentioned: https://github.com/winglang/wing/issues/470.

NOTE: exponentiation operator precedence was not defined in the language spec, so I followed the JavaScript spec (considering we are jsifying them), putting it in between Parentheses and Unary minus. I inserted it before `NIL_COALESCING` in PREC, not sure if it would be an issue. LMK


eladb (approved):
Nice!

Issue #1169, opened 2023-01-16
Cannot capture enums


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:
`wing test`
```js
bring cloud;

enum SomeEnum {
    ONE, TWO, THREE
}

let two: SomeEnum = SomeEnum.TWO;

new cloud.Function(inflight (e: str) => {
  assert(two == SomeEnum.TWO);
}) as "test";
```

### I expected this:

To work

### Instead, this happened:

```
ReferenceError: SomeEnum is not defined
     │     at Handler.handle (evalmachine.<anonymous>:20:100)
     │     at exports.handler (evalmachine.<anonymous>:25:5)
     │     at evalmachine.<anonymous>:27:9
     │     at Script.runInContext (node:vm:141:12)
     │     at Object.runInContext (node:vm:291:6)
     │     at /Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:133:12
     │     at new Promise (<anonymous>)
     │     at runInSandbox (/Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:114:12)
     │     at Object.activity (/Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:55:24)
     └     at Object.withTrace (/Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:82:50)
```

### Component:

Compiler

monadabot:
Congrats! :rocket: This was released in [Wing 0.16.1](https://github.com/winglang/wing/releases/tag/v0.16.1).

monadabot:
Congrats! :rocket: This was released in [Wing 0.16.1](https://github.com/winglang/wing/releases/tag/v0.16.1).

PR #1662, opened 2023-02-24
feat(compiler): static support

yoav-steinberg:
See #107

also in this PR:
  * de-dup diagnostics by storing them in an `IndexSet` instead of a `Vec`.
  * Added a new AST reference type `TypeMember` for referencing enum fields and class statics. Currently the grammar has no way of distinguishing between a nested reference and a type property reference. So during type-checking I do a (unsafe) convert in the AST tree from one type to the other. If we change the grammar to distinguish between the two we can avoid the unsafe here.
  * Changed `Reference::NestedReferece` to `Reference::InstanceMember` to match nicely with the new `Reference::TypeMember`.

**Not in this PR**
  - Support static field initialization (#1668)
  - Support accessing preflight static stuff inflight (#1669)


examples/tests/valid/static_members.w
```diff
@@ -0,0 +1,15 @@
+resource Foo {
+  instance_field: num;
+
+  static f: num;
```
eladb, line 4: I _think_ it might be okay for now not to support static fields...

libs/tree-sitter-wing/grammar.js
```diff
@@ -148,29 +150,22 @@ module.exports = grammar({
...
-        $._type_annotation,
-        ";"
-      ),
-    inflight_class_member: ($) =>
+    class_field: ($) =>
       seq(
         optional(field("access_modifier", $.access_modifier)),
-        field("phase_modifier", $._inflight_specifier),
+        optional(field("static", $.static)),
+        optional(field("phase_modifier", $._inflight_specifier)),
```
eladb, line 164: Does this mean we can have `static inflight` fields and methods?

(Kind of feels like we need)

libs/tree-sitter-wing/test/corpus/statements/class_and_resource.txt
```diff
@@ -0,0 +1,173 @@
...
+    preflight_func() {}
+    async preflight_func() {}
+    public preflight_func2() {}
+    inflight inflight_func() {}
+    public inflight inflight_func2() {}
+    pf_member: str;
+    inflight if_member: str;
+    public inflight var if_member2: str;
+    static inflight inflight_field: num;
+    static inflight inflight_method() {}
```
eladb, line 90: 👍

libs/wingc/src/ast.rs
```diff
@@ -164,6 +164,7 @@ pub struct FunctionDefinition {
 	pub signature: FunctionSignature,
 	#[derivative(Debug = "ignore")]
 	pub captures: RefCell<Option<Captures>>,
+	pub _static: bool,
```
eladb, line 167: Why does this have an underscore prefix?

If it is because `static` is a reserved word, then maybe call this `is_static`?

libs/wingc/src/capture.rs
```diff
@@ -240,6 +239,9 @@ fn scan_captures_in_expression(
 					}
 				}
 			}
+			Reference::TypeProperty { .. } => {
+				// TODO: handle access to static preflight memebers from inflight
```
eladb, line 243: Issue reference please 

Also- should we emit a diagnostic here?

libs/wingc/src/ast.rs
yoav-steinberg: yes, i'll fix

examples/tests/valid/static_members.w
yoav-steinberg: I might take you up on this since it'll remove the need to handle static field initialization which is a bit complex.

libs/tree-sitter-wing/grammar.js
yoav-steinberg: I think it should work, need to add a test.

libs/wingc/src/lsp/sync.rs
```diff
@@ -106,7 +106,12 @@ fn partial_compile(source_file: &str, text: &[u8]) -> FileData {
 	let mut capture_visitor = CaptureVisitor::new();
 	capture_visitor.visit_scope(&scope);
 
-	let diagnostics = vec![parse_diag, type_diag, capture_visitor.diagnostics].concat();
+	let mut diagnostics = Diagnostics::new();
+	for diags in [parse_diag, type_diag, capture_visitor.diagnostics].iter() {
+		for diag in diags.iter() {
+			diagnostics.insert(diag.clone());
+		}
+	}
```
Chriscbr, line 114: Feels like there's probably a way to do this without the extra clones...

libs/wingc/src/type_check.rs
```diff
@@ -2129,38 +2185,50 @@ impl<'a> TypeChecker<'a> {
...
+					}
 				}
 			}
 		}
 
 		return new_type;
 	}
 
-	fn expr_maybe_type(&mut self, expr: &Expr, env: &SymbolEnv, statement_idx: usize) -> Option<TypeRef> {
+	fn expr_maybe_type(
```
Chriscbr, line 2198: Docstring (purpose of the method isn't clear from its name IMO)

libs/wingc/src/type_check.rs
```diff
@@ -2129,38 +2185,50 @@ impl<'a> TypeChecker<'a> {
...
+						path.push(symbol.clone());
 						break;
 					}
 					Reference::NestedIdentifier { object, property } => {
-						path.push(property);
+						path.push(property.clone());
 						curr_expr = &object;
 					}
+					Reference::TypeProperty { .. } => {
+						panic!("Type property references cannot be a type name because they have a property");
```
Chriscbr, line 2219: Can we make this into a diagnostic?

libs/wingc/src/type_check.rs
```diff
@@ -2179,30 +2247,27 @@ impl<'a> TypeChecker<'a> {
...
+						_type: user_type_annotation,
+						property: property.clone(),
 					};
+					// Replace the reference with the new one, this is unsafe because `reference` isn't mutable and theoretically someone may
+					// hold anoter reference to it. But our AST doesn't hold up/cross references so this is safe as long as we return right.
+					let const_ptr = reference as *const Reference;
+					let mut_ptr = const_ptr as *mut Reference;
+					unsafe {
+						// We dont' use the return value but need to call replace so it'll drop the old value
+						_ = std::mem::replace(&mut *mut_ptr, new_ref);
```
Chriscbr, line 2268:
```suggestion
						let old_value = std::mem::replace(&mut *mut_ptr, new_ref);
						std::mem::drop(old_value);
```

libs/wingc/src/type_check.rs
```diff
@@ -2283,19 +2349,62 @@ impl<'a> TypeChecker<'a> {
...
 
 				if force_reassignable {
-					VariableInfo {
-						_type: res._type,
-						reassignable: true,
-						flight: res.flight,
-					}
+					let mut res = res.clone();
+					res.reassignable = true;
+					res
```
Chriscbr, line 2359: Just for fun, here's a third way to do this (random rust trivia)
```suggestion
					VariableInfo {
					  reassignable: true
					  ..res.clone(),
					}
```

Chriscbr (changes requested):
This is looking promising...

libs/tree-sitter-wing/test/corpus/statements/class_and_resource.txt
```diff
@@ -0,0 +1,173 @@
...
+    init() {}
+    preflight_func() {}
+    async preflight_func() {}
+    public preflight_func2() {}
+    inflight inflight_func() {}
+    public inflight inflight_func2() {}
+    pf_member: str;
+    inflight if_member: str;
+    public inflight var if_member2: str;
+    static inflight inflight_field: num;
```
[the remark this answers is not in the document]
yoav-steinberg: they are

libs/wingc/src/diagnostic.rs
```diff
@@ -6,7 +7,7 @@ use crate::debug;
 pub type FileId = String;
 pub type CharacterLocation = Point;
 pub type ByteIndex = usize;
-pub type Diagnostics = Vec<Diagnostic>;
+pub type Diagnostics = IndexSet<Diagnostic>;
```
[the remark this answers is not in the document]
yoav-steinberg: What lead to me adding this is that I now call `resolve_reference` in some cases more than once on the same reference in the type-checker. Specifically when type checking a call expression I first type check the expression that evaluates to the function type (usually a reference to a function/method name) and then later I call `resolve_reference` again just to check if that method/function is static or not so I know if I should expect a `this` first arg to it or not.
The double call to `resolve_reference` can lead to duplicate "unknown symbol" errors.

But then I noticed that some other duplicates where removed from our invalid snapshot tests, so I guess there were other scenarios that caused duplicates. It makes sense because there's no way for me to know if some function in our compiler produces an error and I don't want to limited to not calling functions more than once on the same inputs.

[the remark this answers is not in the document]
yoav-steinberg: > Why IndexSet vs HashSet or BTreeSet? (add comment maybe)

The answer is that we want to preserve the order in which the errors are collected. I'll add a comment.

libs/wingc/src/lsp/sync.rs
yoav-steinberg: Played around with this a bit. I had a version that uses `IndexSet.union` but that was clunky and broke when we added the capture visitor (after a merge from main). Maybe worth andding a `concat` like thing `IndexSet` and PRing it to them..
I'm open to any improvements, but eventually just wanted to get it done...

libs/wingc/src/type_check.rs
yoav-steinberg: In the current implementation this is clearly a bug. This method is used to check if we need to turn a `Reference::NestedIdentifier` into a `Reference::TypeProperty`. So we should never get here if the input expression is a `Reference::TypeProperty`. I might refactor this to make it clearer and get rid of this `panic`.
In general I'm really against adding diagnostics when we know it's a bug, because then we might never notice we have a bug in our code and treat it the same as bad input.

libs/wingc/src/type_check.rs
yoav-steinberg: I think it'll drop the old value just because it goes out of scope. the `_ =` here is just to get rid of the compiler warning. So I don't think we need the explicit `drop`. I added the comment just to be explicit about what's going on here.

libs/wingc/src/type_check.rs
yoav-steinberg: 😲!

yoav-steinberg: Stockholm syndrome or not, that's simply cool.

libs/wingc/src/type_check.rs
Chriscbr: Ah ok, I wasn't sure - sounds good 👍

yoav-steinberg: Added some comments and also renamed the variants in `Reference` to be clearer (I hope).

libs/wingc/src/type_check.rs
yoav-steinberg: done.

examples/tests/valid/static_members.w
eladb: Fuck

yoav-steinberg: > I guess we can start with just supporting this in jsii import without supporting the syntax to declare and define the initial value in Wing. No?

Exactly my thoughts. Added a tracking issue and resolving this.

libs/tree-sitter-wing/grammar.js
yoav-steinberg: This works as long as the type is define inflight. See example in _static_members.w_ test. I've opened a tracking issue (#1669) for the more complex scenario, hopefully you're ok merging as is.

staycoolcall911: I'm ok with that. Let's merge that PR in. #1669 is p1 IMO, it is similar to #1169. 
Resolving - @eladb, please unresolve or let's speak if you disagree.

libs/wingc/src/capture.rs
staycoolcall911: @yoav-steinberg opened issue #1669.
After an offline chat with @yoav-steinberg I understand emitting the right diagnostic here calls for some refactoring. I suggest we merge the PR as is and since #1669 is p1 we should address it before beta.

Chriscbr (approved):

PR #1665, opened 2023-02-26
rfc(docs): tasklist with api and db

ekeren:
This is just an initial RFC for TaskList with API - see [rendered](https://github.com/winglang/wing/blob/ekeren/task-list-with-api-and-db/docs/06-contributors/999-rfcs/000-stories/story-15.md)


docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+  api: cloud.Api;
+  model: TaskListModel;
+  init(model: ITaskListModel) {
+    this.model = model;
+    this.api = new cloud.Api();
+    
+    this.api.post("/tasks", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let title = Str.from_json(req.body.title);
+      let id = this.model.add(title);
+      return new cloud.Api.ApiResponse(status:202, body: Json.format(id));
```
staycoolcall911, line 46:
```suggestion
      return new cloud.Api.ApiResponse(status:201, body: Json.format(id));
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    this.api.get("/tasks/:id", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let id = Str.from_json(req.params.id);
+      try {
+        let title = this.model.get(id);
+        return new cloud.Api.ApiResponse(status:202, body: Json.format(title));
```
staycoolcall911, line 53:
```suggestion
        return new cloud.Api.ApiResponse(status:200, body: Json.format(title));
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+      } catch {
+        return new cloud.Api.ApiResponse(status:404);
```
staycoolcall911, line 55: Why 404? Are we treating every exception as "not found"?

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+      }
+    });
+    
+    this.api.delete("/tasks/:id", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let id = Str.from_json(req.params.id);
```
staycoolcall911, line 60:
```suggestion
      let id = str.from_json(req.params.id);
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+      try {
+        this.model.delete(id);
+        return new cloud.Api.ApiResponse(status:204);
+      } catch {
+        return new cloud.Api.ApiResponse(status:404);
+      }
+    });
+
+    this.api.get("/tasks", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let search = Str.try_from_json(req.query.search) ?? ".*"; // Do we want a regex primitive? Regex.from_str(".*")
```
staycoolcall911, line 70:
```suggestion
      let search = str.try_from_json(req.query.search) ?? ".*"; // Do we want a regex primitive? Regex.from_str(".*")
```

staycoolcall911, line 70: `try_from_json()` is p2 - #1053 
Let's either remove it from here or bump it to p1.

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+      let results = this.model.find(search);
+      return new cloud.Api.ApiResponse(status:202, body: Json.format(results));
```
staycoolcall911, line 72:
```suggestion
      return new cloud.Api.ApiResponse(status:200, body: Json.format(results));
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+    this.api = new cloud.Api();
+    
+    this.api.post("/tasks", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let title = Str.from_json(req.body.title);
+      let id = this.model.add(title);
+      return new cloud.Api.ApiResponse(status:202, body: Json.format(id));
+    });
+
+    this.api.get("/tasks/:id", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let id = Str.from_json(req.params.id);
```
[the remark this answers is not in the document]
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+
+resource TaskListApi {
+  api: cloud.Api;
+  model: TaskListModel;
+  init(model: ITaskListModel) {
+    this.model = model;
+    this.api = new cloud.Api();
+    
+    this.api.post("/tasks", inflight (req: cloud.Api.ApiRequest): cloud.Api.ApiResponse => {
+      let title = Str.from_json(req.body.title);
```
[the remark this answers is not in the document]
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+
+  inflight add(title: str): str {
+
+  }
+
+  inflight remove(id: str) {
+
+  }
+
+  inflight find(r: str): Set<str> {
```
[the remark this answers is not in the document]
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,79 @@
...
+  inflight remove(id: str);
+  inflight find(r: str): Set<str>;
+}
+
+resource TaskListModel implementes ITaskListModel {
+
+  init() {
+
+  }
+  inflight get(id: str): str {
```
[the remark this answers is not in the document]
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: fixed, will be seen in coming commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: I think this is good enough for now

ekeren: Do you disagree?

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,169 @@
...
+
+## New landuage and SDK features it introduces
+
+- [x] bring untyped
+  - [x] bring external npm package (axios)
+  - [x] bring an internal nodejs stdlib (RegEx)
+- [x] enum & duration that can be included inside json
+- [x] It leverages setting explicit permissions (using the `this.inflight` API, described [here](https://github.com/winglang/wing/pull/1610))
+- [x] Using interface 
+- [x] uuid as standard type (missing spec)
```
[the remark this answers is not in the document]
ekeren: I am more worried about the P1. 

I agree that having a builtin type for network has higher value then uuid, but I want to make sure we complete all P1 (make sure the system is functionally complete) vs optimizing for the DX, which is clearly high on the list. 

Your comment does makes think that bringing uuid as a standard type should be removed from this list, it is also not a p1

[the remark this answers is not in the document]
ekeren: removed the uuid and added another untyped interface question

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Discussion topics
+- `get` vs `on_get` - Should the `cloud.api` API have the `on_` prefix to match `cloud.bucket` API and also to allow calling
+the api get/post/delete/put commands (`api.get(url)` vs `api.on_get(path, inflight ())`
+- Routing params - I have used express's synatx for dynamic parts of the path for the api gateway `get("/task/:id"`, 
+but then I noticed that [aws](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-step-by-step.html) 
+uses the `{id}` syntax. What should be the the right syntax?
+- Bring internal nodejs types - how do we bring somethink that we can't require, like `RegEx`
+- From anything to Json - can I cast something from bring untyped to Json? 
+- Calling await
```
eladb, line 31: ??

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+- [ ] bring cdktf
+- [ ] use redis instead of bucket
+- [ ] code that updates estimation and duration from REST post command
+
+## Discussion topics
+- `get` vs `on_get` - Should the `cloud.api` API have the `on_` prefix to match `cloud.bucket` API and also to allow calling
+the api get/post/delete/put commands (`api.get(url)` vs `api.on_get(path, inflight ())`
+- Routing params - I have used express's synatx for dynamic parts of the path for the api gateway `get("/task/:id"`, 
+but then I noticed that [aws](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-step-by-step.html) 
+uses the `{id}` syntax. What should be the the right syntax?
```
eladb, line 28: For preflight - I think I prefer `.get()` and `.post()` because these are not really "events". There is a single callback for each route, while usually it is possible to subscribe multiple handlers for events. Having said that I am not that opinionated... Let's test the water with the current spec and see how it feels.

Interesting idea about inflight `get()`, `post()` as a way to issue REST HTTP requests...

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+- Bring internal nodejs types - how do we bring somethink that we can't require, like `RegEx`
+- From anything to Json - can I cast something from bring untyped to Json? 
+- Calling await 
+
+## Code 
+```ts (wing)
+
+bring cloud;
+
+// TODO discuss how we bring untyped something like RegEx from JavaScript
```
eladb, line 38: How about this:

```js
bring untyped js;

// and then
new js.RegExp(...)
js.Math.max()
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+
+// TODO discuss how we bring untyped something like RegEx from JavaScript 
+// PLACEHOLER for bringing something from Javascript stdlib
+bring untyped RegExp from_js("RegExp"); 
+bring untyped Math from_js("Math"); 
+
+
+// prerequisite: npm install axios
+// PALCEHOLDER for bringing some external module
+bring untyped axios require_from_js("axios");
```
eladb, line 46: I believe this is the syntax we had for bring untyped:

```suggestion
bring untyped "axios" as axios; 
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+// prerequisite: npm install axios
+// PALCEHOLDER for bringing some external module
+bring untyped axios require_from_js("axios"); 
+
+enum Status {
+  Uncompleted,
+  Completed
+}
+
+interface ITaskListModel {
```
eladb, line 53: Nit: Personally I've never liked to use the word "model" for this. Everything is a model...

```suggestion
interface ITaskList {
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+  inflight find(r: RegExp): Array<str>;
+  inflight set_status(id: str, status: Status): str;
+  inflight set_estimation(id: str, estimation: duration): str;
+}
+
+resource TaskListModel implementes ITaskListModel {
+  _bucket: cloud.Bucket;
+  init() {
+    this._bucket = new cloud.Bucket();
+    this.inflights.add("get", ref: "this._bucket", op: "get");
```
eladb, line 66: All of these should be implicit by the end of the sprint I believe. @yoav-steinberg ?

docs/06-contributors/999-rfcs/000-stories/story-15.md
eladb, line 53: Do we really need an interface here or is this just because we want to get interfaces implemented?

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+      status: Status.Uncompleted
+    };
+    print("adding task ${id} with data: ${j}"); 
+    return this_add(id, js);
+  }
+
+  inflight remove(id: str) {
+    print("removing task ${id}");
+    this._bucket.delete(id);
+    return id;
```
eladb, line 100: Mismatch between interface and method

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+  model: TaskListModel;
+  init(model: ITaskListModel) {
+    this.model = model;
+    this.api = new cloud.Api();
+    
+    // TODO add this.put
+    
+    this.api.post("/tasks", inflight (req: cloud. Api.ApiRequest): cloud.ApiResponse => {
+      let var title = str.from_json(req.body.title);
+      if title == random {
```
eladb, line 133: what does that mean? how is it related to this story?

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+        // PLACEHOLDER - can I cast an untyped ?
+        let random_task: Json = await axios.get('https://www.boredapi.com/api/activity');
+        title = str.from_json(random_task.data.activity); 
+      } 
+      let id = this.model.add(title);
+      return new cloud.ApiResponse(status:201, body: Json.format(id));
```
eladb, line 139: I think `ApiResource` is a struct (@Chriscbr?):

```suggestion
      return cloud.ApiResponse { status:201, body: Json.format(id) };
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+    });
+
+    this.api.get("/tasks/:id", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
+      let id = str.from_json(req.params.id);
+      try {
+        let title = this.model.get(id);
+        return new cloud.ApiResponse(status:200, body: Json.format(title));
```
eladb, line 146:
```suggestion
        return cloud.ApiResponse {status:200, body: Json.format(title)};
```

eladb (commented):
Some initial comments

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: love it

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: Because I want interface implemented 🤷 

I want people to help user realize "Oh, so a resource is a class"
Another thing that will help is moving forward with your turn-resource-into-a-class-modifier idea.

BTW, I prefer 
```ts 
scoped class TaskList {}
```
vs 
```ts 
resource class TaskList {}
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: I've added an easter egg  to our TODO list, if you add a title random, the system will fetch a random task from the Internet. 

We wanted to use network for something, and thought this is a nice idea

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
+# User Story 15 - Task List with Api Gateway and Redis
+
+> **Status**: Draft, Expected released on 2023/03/16
+
+
+The following code is an inital implementation of TaskList with api gateway and a redis db
```
staycoolcall911, line 6:
```suggestion
The following code is an initial implementation of TaskList with api gateway and a redis db 
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: > here is a single callback for each route

Nice point

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,173 @@
...
+
+- [x] bring untyped
+  - [x] bring external npm package (axios)
+  - [x] bring an internal nodejs stdlib (RegEx)
+  - [x] How does untyped works with numeric operations?
+  - [x] Do we cast untyped? 
+- [x] enum & duration that can be included inside json
+- [x] It leverages setting explicit permissions (using the `this.inflight` API, described [here](https://github.com/winglang/wing/pull/1610))
+- [x] Using interface 
+- [x] Using await when calling a untyped function that returns a promise
```
staycoolcall911, line 18:
```suggestion
- [x] Using await when calling an untyped function that returns a promise
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: If that is the case, I would work to add a non trivial example that can't be calculated by the compiler and describe how we want the compiler to behave. (error/warning/ default behavior)

Waiting for @yoav-steinberg

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: We call await on a network call (using axios)

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: Done

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,171 @@
...
+
+bring cloud;
+
+// TODO discuss how we bring untyped something like RegEx from JavaScript 
+// PLACEHOLER for bringing something from Javascript stdlib
+bring untyped js;
+
+// prerequisite: npm install axios
+// PALCEHOLDER for bringing some external module
+bring untyped "axios" as axios;
```
ekeren, line 44: @eladb 

Once we've added `bring untyped js;`
It makes me wonder why aren't we allowing:

```ts
let axios = js.require("axios");
```
instead of 
```ts
bring untyped "axios" as axios; 
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,171 @@
...
+- [ ] use redis instead of bucket
+- [ ] code that updates estimation and duration from REST post command
+
+## Discussion topics
+- `get` vs `on_get` - Should the `cloud.api` API have the `on_` prefix to match `cloud.bucket` API and also to allow calling
+the api get/post/delete/put commands (`api.get(url)` vs `api.on_get(path, inflight ())`
+- Routing params - I have used express's synatx for dynamic parts of the path for the api gateway `get("/task/:id"`, 
+but then I noticed that [aws](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-step-by-step.html) 
+uses the `{id}` syntax. What should be the the right syntax?
+- Bring internal nodejs types - how do we bring somethink that we can't require, like `RegEx`
```
staycoolcall911, line 29:
```suggestion
- Bring internal nodejs types - how do we bring something that we can't require, like `RegEx`
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: And then we should wonder if we even need `bring untyped` as a mechanism and the `untyped` keyword vs having some dedicated way to being `js`

docs/06-contributors/999-rfcs/000-stories/story-15.md
staycoolcall911: `bring untyped js` seems too wide. Can we limit the scope somehow? `bring untyped js.RegExp as RegExp`?

docs/06-contributors/999-rfcs/000-stories/story-15.md
eladb: Yes, I am leaning towards a modifier as well. I would start with `resource class` only since its a smaller change and then we can decide if we want to revisit the term.

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,186 @@
...
+        result.push(id);
+      }
+    }
+    return result.copy();
+  }
+
+  inflight set_status(id: str, status: Status): str {
+    let j = Json.clone_mut(this.get(id));
+    j.status = status;
+    this._add(id, j);
```
MarkMcCulloh, line 121: `_add` takes a Json, not MutJson.

```suggestion
    this._add(id, Json.clone(j));
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,186 @@
...
+    
+    // TODO add this.put
+    
+    this.api.post("/tasks", inflight (req: cloud. Api.ApiRequest): cloud.ApiResponse => {
+      let var title = str.from_json(req.body.title);
+      // Easter Egg - if you add a todo with the single word "random" as the title, 
+      //              the system will fetch a random task from the internet
+      if title == "random" {
+        // PLACEHOLDER - can I cast an untyped ?
+        let random_task = axios.get('https://www.boredapi.com/api/activity');
```
ekeren, line 148: @eladb , following the comments from the team, I understand that all function calls (in inflight context), and not just inflight method on resources are implicitly using await

Is that the intent?

Chriscbr (commented):
I really like the separation between `TaskList` and `TaskListApi` in this example!

One variation I thought of (for a future user story, or just as a thought experiment) was what if `TaskListApi` accepted any number of `ITaskList`'s and made `TaskListApi` responsible for organizing the APIs like so:

```
/task_lists/{task_list_id}/tasks/{task_id}
/task_lists/1/tasks/1
/task_lists/1/tasks/2
/task_lists/2/tasks/1
...
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,186 @@
...
+
+    // TODO add more permissions for remove, find
+  }
+
+  inflight get(id: str): Json {
+     return Json.parse(this._redis.GET(id));
+  }
+  
+  inflight _add(id: str, j: Json): str {
+    this._redis.SET(id , Json.format(j));
```
Chriscbr, line 82: In some other languages "format" is associated with format strings like `printf` in C, `println!` in Rust, or f-strings in Python. We could consider renaming to
```suggestion
    this._redis.SET(id , Json.to_str(j));
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,186 @@
...
+      return cloud.ApiResponse { status:201, body: Json.format(id) };
+    });
+
+    this.api.get("/tasks/{id}", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
+      let id = str.from_json(req.params.id);
+      try {
+        let title = this.task_list.get(id);
+        return cloud.ApiResponse {status:200, body: Json.format(title)};
+      } catch {
+        return cloud.ApiResponse(status:400);
```
Chriscbr, line 161: Is the second syntax `cloud.ApiResponse` supported? I think only the first way of creating structs has been written right now - the potential issue I see with the second is that it makes `cloud.ApiResponse` look like a method call or a constructor (missing the `new` keyword).

docs/06-contributors/999-rfcs/000-stories/story-15.md
```diff
@@ -0,0 +1,186 @@
...
+    let result = MutArray<str>[]; 
+    let ids = this._redis.SMEMBERS("todo");
+    for id in ids {
+      let j = Json.parse(this._redis.GET(id));
+      // Notice that there is autocasting from untyped to bool here 
+      if r.test(j.title) {
+        result.push(id);
+      }
+    }
+    return result.copy();
```
Chriscbr, line 115: I'm still unsure about this API - to me, it feels like converting `MutArray<T>` into `Array<T>` should not require copying data or memory allocation. (It's totally possible I'm not thinking about this right...)

I wonder if it makes sense to say `MutArray<T>` is a subtype of `Array<T>`? From a behavioral perspective, `MutArray<T>` has all of the same behaviors as `Array<T>` (like `get`, `len`, etc.), and includes _extra behaviors_ like `append`, `delete`, etc. If this is the case, then it should be safe to return `result` directly since `MutArray<T>` is a subtype of the annotated return type `Array<str>`.

Thoughts? cc @eladb @yoav-steinberg

MarkMcCulloh:
```
let mutable_array = MutArray<num> [1, 2, 3];

// If MutArray is allowed to be assigned like this without cloning
let immutable_array: Array<num> = mutable_array;

// immutable_array has changed, thus defeating the point of it being immutable
mutable_array.push(4)
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
MarkMcCulloh: Alternatively, to maybe make it more clear that this is effectively a cast:

```suggestion
    this._redis.SET(id , Json.as_str(j));
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
Chriscbr: 🤦‍♂️  right, I see... (great example btw)

It's sort of like a 3 line math proof, and the question whether the step from line 1 to 2 is wrong, or the step from line 2 to 3 is wrong.

In Rust, step 2 to 3 is invalid because `mutable_array` gets "moved" to immutable_array, so it can't be used by the old variable anymore. But since we don't do that kind of static analysis in our language (doesn't really make sense I think?) then step 2 to 3 must be OK. Thus the subtyping between steps 1 and 2 is wrong for our language.

So I guess as a rule of thumb, to the extent that our language has subtyping, in order to figure out whether Y is a subtype of X, we have to consider whether the relationship breaks any "mutation" invariants, and not just whether Y inherits the same set of behaviors. (at least that's my initial takeaway)

Chriscbr: In that case `result.copy()` makes sense, so I'm resolving this thread 👍

docs/06-contributors/999-rfcs/000-stories/story-15.md
eladb: `require` is a node.js thing, not a JS thing, and it's also not a "real function". I wouldn't conflate module imports and function calls (that's a big mistake node didthe JS community is still trying to recover from using `import`).

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: I also found format confusing.

@eladb  FTI

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren, line 161:
```suggestion
        return cloud.ApiResponse { status: 400 };
```

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: > `bring untyped js` seems too wide. Can we limit the scope somehow? `bring untyped js.RegExp as RegExp`?

Here are my thoughts about this: 

I think that the difference here is about the maturity of wing.

If you use wing today, `bring untyped js;` will be more empathic to your needs, because, at beta, we are going to see the bring untyped quite often in wing code. Once wing is closer to GA, we expect the developer to use `bring untyped` less common, in that case the `bring untyped js.RegExp as RegExp;` is more empathic because it shows in a glance all the untyped imports this code is bringing.

@eladb , this is a question of language UX

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren:
```suggestion
```

Following https://github.com/winglang/wing/pull/1665/files#r1120267023

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: fixed, thanks

staycoolcall911 (approved):
I think we can merge that in and solve the DX around `bring untyped js` during the sprint.

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: This is resolved as part of the last commit

docs/06-contributors/999-rfcs/000-stories/story-15.md
ekeren: changed to to_str

Issue #1641, opened 2023-02-23
cloud.Counter.reset()


staycoolcall911:
https://docs.winglang.io/reference/wingsdk-spec#counter

YarivLevy81:
Hi there, first attempt to contribute to wing. 
https://github.com/winglang/wing/pull/1667

staycoolcall911:
@YarivLevy81 thanks so much for picking up this issue. I've assigned it to you. Let me know if you need any help or pointers

Chriscbr:
Implemented in #1667

PR #1667, opened 2023-02-26
feat(sdk): counter.reset()

YarivLevy81:
Implementation of Counter.reset() as requested, accepts an optional value and defaults to 0.

#1641


staycoolcall911:
This looks nice and straight forward @YarivLevy81 - great job!
I'll leave it for @Chriscbr, @hasanaburayyan or @eladb to give it another review and approve/request changes.

libs/wingsdk/src/cloud/counter.ts
```diff
@@ -89,6 +89,13 @@ export interface ICounterClient {
    * @inflight
    */
   peek(): Promise<number>;
+
+  /**
+   * Reset a counter to a given value.
+   * @param value value to reset (default is 0)
+   * @inflight
+   */
+  reset(value?: number): Promise<void>;
```
hasanaburayyan, line 98: This value is not optional in the sdk spec https://github.com/winglang/wing/pull/1633/files, but given this is a reset method and not set I think it makes for a nicer DX. I think we should update the spec to reflect this default behavior @Chriscbr wdyt?

libs/wingsdk/src/target-tf-aws/counter.inflight.ts
```diff
@@ -56,4 +57,24 @@ export class CounterClient extends CounterClientBase {
...
+      ReturnValues: "UPDATED_NEW",
+    });
+
+    const result = await this.client.send(command);
+    let newValue = result.Attributes?.[VALUE_ATTRIBUTE].N;
+    if (!newValue) {
+      throw new Error(`${VALUE_ATTRIBUTE} attribute not found on table.`);
+    }
+
+    return;
```
hasanaburayyan, line 78: Since we are not trying to do any early bailout with this return I think we can just remove it.

libs/wingsdk/test/target-tf-aws/counter.inflight.test.ts
```diff
@@ -54,6 +54,26 @@ test("inc(5)", async () => {
   expect(response).toEqual(prevValue); // returns previous value
 });
 
+test("reset", async () => {
```
hasanaburayyan, line 57:
```suggestion
test("reset(0)", async () => {
```

hasanaburayyan (approved):
This is great 🚀 few minor nits.

libs/wingsdk/src/cloud/counter.ts
YarivLevy81: Yea makes sense. I think that as a developer I would expect `counter.reset()` to just set to counter to 0.
I'll change the spec to with your decision as I fix the nits.

eladb: Of course 0 should be the default. 

Make sure to document with a @default doc tag

YarivLevy81:
@hasanaburayyan 
I resolved all conversations

libs/wingsdk/src/cloud/counter.ts
flyingImer: curious - why 0 over the initial value as default?

eladb: You are right. That's a good point.

But... this requires "capturing" the value from preflight to the inflight client. In wing this is trivial but in typescript it require a few hoola hoops.

Maybe open an issue to follow up.

staycoolcall911: Resolving for now

staycoolcall911:
@skyrpex, @polamoros, @ainvoner - I've opened an issue to add support for this to the console  #1672

PR #1681, opened 2023-02-27
chore: replace polycons with a `new()` factory method

eladb:
Since we are optimizing the SDK to be used through Wing, we can remove the dependency on `polycons` and clean up much of the boilerplate needed in the base classes. This is the first step towards #1365.

To do this, we've changed the compiler to emit a call to:

```js
this.node.root.new(my_fqn, MyResource, scope, id, ...args)
```

instead of:

```js
new MyResource(scope, id, ...args)
```

We do this only for resources for which we actually have an FQN, currently only those imported from JSII. For resources without an FQN (such as user-defined types), we continue to emit normal `new`.

The node returned by `this.node.root` is our `App`, which is target-specific. Therefore, it can determine the concrete behavior based on the target. In the future we will be able to allow plugins to customize further. For now, I removed the `customFactory` option from app as it is not being used and will be replaced with a hook.

The `new()` factory has the following signature:

```js
new(fqn, ctor, scope, id, ...args)
```

Where `fqn` is the JSII FQN of the type we wish to create (e.g. `@winglang/sdk.cloud.Bucket`). `ctor` is the constructor object (e.g. `Bucket`) and the rest are the standard construct arguments.

The `ctor` is passed in order to allow the `core.App` base implementation to fall back to  simply instantiate the type requested (e.g. when defining raw CDKTF resources). If `ctor` is `undefined`, it implies that the resource is an *abstract type* and therefore must be resolved by the app. If it is not resolved, this is the error:

```
Unable to create an instance of abstract type "@winglang\/sdk.cloud.Schedule"
```

Types are now identified by their JSII FQNs, but in order to maintain backwards compatibility with the Wing Console, the legacy type naming scheme (e.g. `wingsdk.cloud.Bucket`) is still used for the simulator manifest. We will change this in a subsequent commit in coordination with the Wing Console.

This change allowed removing most of the boilerplate hackery required to implement polycons: most `XxxBase` classes were deleted as well as the phantom polycons classes. Yey!

To make it easier to create `cloud.Xxx` resources in the SDK code (either when resources use each other or in tests), each resource class has a static method called `cloud.Xxx.newXxx()` which basically calls `this.root.new()` with the resource's FQN. This is how you would define a bucket:

```js
cloud.Bucket.newBucket(scope, "MyBucket");
```


libs/wingsdk/src/core/app.ts
```diff
@@ -55,24 +40,75 @@ export interface AppProps {
...
+   * @param ctor the constructor of the class to instantiate (undefined for abstract classes)
+   * @param scope the scope of the resource
+   * @param id the id of the resource
+   * @param args the arguments to pass to the resource
+   * @returns the new instance
+   * @throws if the FQN is not supported
+   */
+  public new(
+    fqn: string,
+    ctor: any | undefined,
```
MarkMcCulloh, line 91: JSII might let you get away with using an actual type here, something like:

```suggestion
    ctor: new (scope: Construct, id: string, ...args: any[]) => Construct | undefined,
```

It only really matters for TS, but hey why not. I did not test this and jsii might yell.

MarkMcCulloh, line 91: Would it make sense to have 2 different versions of `new`, one that takes a ctor and one that doesn't? `new` and `newAbstract` or something.
Clears up the intent of the caller, and lets the type system do the work of null checking for us.

libs/wingsdk/test/core/app.test.ts
```diff
@@ -0,0 +1,51 @@
+import { Construct } from "constructs";
+import { App } from "../../src/core/app";
+
+describe("new()", () => {
+  test("allows derived classes to override", () => {
```
Chriscbr, line 5: I spent a few minutes trying to think of edge cases but I couldn't come up with too much...

I suppose e.g. if I created an abstract resource that extends another abstract resource, it will also need a Wing plugin in order to reify it into a real class instance, but that probably makes sense.

Chriscbr (commented):
So much boilerplate removed! 🛫 

If you have extra time to update this section of the docs, it would be great - otherwise, I can take it as an issue at some point: https://docs.winglang.io/contributors/wingsdk#polycons

docs/04-reference/wingsdk-api.md
```diff
@@ -7,1245 +7,6 @@ keywords: [Wing sdk, sdk, Wing API Reference]
 
 # API Reference <a name="API Reference" id="api-reference"></a>
 
-## Constructs <a name="Constructs" id="Constructs"></a>
-
-### Bucket <a name="Bucket" id="@winglang/sdk.cloud.Bucket"></a>
```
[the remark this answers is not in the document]
eladb: Will check

libs/wingsdk/src/target-tf-gcp/bucket.ts
```diff
@@ -57,7 +57,7 @@ export class Bucket extends cloud.BucketBase {
 
     this.bucket = new StorageBucket(this, "Default", {
       name: bucketName + "-" + randomId.hex,
-      location: App.of(this).storageLocation,
+      location: (App.of(this) as App).storageLocation,
```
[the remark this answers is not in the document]
eladb: It returns the abstract `core.App`. We need the concrete one.

libs/wingsdk/src/core/app.ts
eladb: I'll try it out. Much better of course

libs/wingsdk/package.json
```diff
@@ -80,8 +80,7 @@
     "@cdktf/provider-google": "^5.0.2",
     "@cdktf/provider-random": "^5.0.0",
     "cdktf": "0.15.2",
-    "constructs": "~10.1.228",
-    "polycons": "^0.1.1"
```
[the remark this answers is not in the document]
eladb: Indeed. Served us well!

libs/wingsdk/src/cloud/queue.ts
```diff
@@ -27,20 +27,29 @@ export interface QueueProps {
...
 /**
- * Functionality shared between all `Queue` implementations.
+ * Represents a queue.
+ *
+ * @inflight `@winglang/sdk.cloud.IQueueClient`
  */
-export abstract class QueueBase extends Resource {
+export abstract class Queue extends Resource {
+  /**
+   * Create a new `Queue` instance.
```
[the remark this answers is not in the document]
eladb: Yes good point.

Let me see how hard it is to add something like `@undocumented` to avoid the underscore shit

libs/wingsdk/test/core/app.test.ts
eladb: I am sure new edge cases will soon emerge

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -19,20 +19,30 @@ export interface BucketProps {
...
+ * Represents a cloud object store.
+ *
+ * @inflight `@winglang/sdk.cloud.IBucketClient`
  */
-export abstract class BucketBase extends Resource {
+export abstract class Bucket extends Resource {
+  /**
+   * Create a new bucket.
+   */
+  public static newBucket(
```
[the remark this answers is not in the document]
eladb: It's a good practice to give static factory methods explicit names because if you extend the class, the name becomes ambigious:

```js
class MyBucket extends Bucket { }
```

And now:

```js
const b = MyBucket.new(...); // returns `Bucket`
```

Per @Chriscbr's suggestion I am going to make these internal, so we can tweak this later.

libs/wingsdk/src/core/app.ts
eladb: Yes, that's a good call out. Let me have a look

libs/wingc/src/jsify.rs
```diff
@@ -274,26 +274,59 @@ impl<'a> JSifier<'a> {
...
+						.expect("expected expression")
+						.as_resource()
+						.expect("expected resource")
+						.fqn
+						.clone()
+				} else {
+					None
+				};
+
+				if ! is_resource || fqn.is_none() {
```
eladb, line 321: Any rustaceans in the room to help me write this more idiomatically?

libs/wingsdk/src/core/app.ts
eladb: Done

eladb: Sadly JSII didn't like this.

docs/04-reference/wingsdk-api.md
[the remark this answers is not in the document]
eladb: Oh... they are all abstract classes now... Investigating...

libs/wingsdk/src/cloud/queue.ts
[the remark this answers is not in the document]
eladb: Went with @internal

docs/04-reference/wingsdk-api.md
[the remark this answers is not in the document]
eladb: Fixed

libs/wingc/src/jsify.rs
Chriscbr: @eladb (you nerd sniped me)

One idea I had was to try flipping the conditional, and using `if let` with an "&&" condition -- but it turns out this syntax isn't stable yet so it's possible so you will get a compiler warning 😅:

```rust
if let Some(fqn) = fqn && is_resource {
  // this.node.root ...
} else {
  // new ...
}
```

So the other way I think you can do this to avoid the extra `expect` is to combine the values into a tuple and then match:

```rust
if let (Some(fqn), true) = (fqn, is_resource) {
	if is_abstract {
		format!("this.node.root.newAbstract(\"{}\",{})", fqn, args)
	} else {
		format!("this.node.root.new(\"{}\",{},{})", fqn, ctor, args)
	}
} else {
	format!("new {}({})", ctor, args)
}
```

Chriscbr (approved):

libs/wingc/src/jsify.rs
MarkMcCulloh: Is there currently a reason to check `is_resource`? I understand that in the future that check would be needed, but based on the code above if is_resource is true then fqn is always `Some` at this point.

so just:

```rust
if let Some(fqn) = fqn {
	if is_abstract {
		format!("this.node.root.newAbstract(\"{}\",{})", fqn, args)
	} else {
		format!("this.node.root.new(\"{}\",{},{})", fqn, ctor, args)
	}
} else {
	format!("new {}({})", ctor, args)
}
```

If you want to keep the check in for the future, then tbh I think what you put initially is perfectly legible. You could even instead chuck an assert for `is_resource == true` inside the `if let` if you wanted to make sure this gets updated later.

MarkMcCulloh (approved):

libs/wingc/src/jsify.rs
eladb: I'll merge it for now and will revisit later. (just don't have time to deal with it right now and it's a big enough change that's worth not lingering on...)

PR #1682, opened 2023-02-28
feat(compiler): minimal capturing from inflight methods

eladb:
Analyze inflight methods to identify exactly which fields are being referenced and which operations are performed on them. Previously, we assumed all inflight methods access all fields across all their inflight operations.

## Implementation notes

This is currently implemented during jsification. The function `find_inflight_references()` iterates over all methods in a resource. For each method it uses a `FieldReferenceVisitor` to analyze the method and identify all references to `this.xxx`.

In some cases we will only fail at runtime. For example see [`inflight_ref_resource_collection.w`](https://github.com/winglang/wing/blob/eladb/infer-refs/examples/tests/valid/inflight_ref_resource_collection.w) and [`inflight_ref_unknown_op.w`](https://github.com/winglang/wing/blob/eladb/infer-refs/examples/tests/valid/inflight_ref_unknown_op.w) are both supposed to fail and require that explicit annotations are added. This will be handled as part of #1610.

The `JSifier` now maintains a `diagnostics` list so it will be possible to report errors (this turned it into a mutable creature).

We still want to bail out *before* jsification if there are any errors in previous phases because otherwise, the jsifier cannot infer certain types, etc.

Tests are prefixed with `inflight_ref_*` (both valid and invalid).

Co-author: @yoav-steinberg 


examples/tests/valid/inflight_ref_resource_collection.w
```diff
@@ -0,0 +1,16 @@
...
+
+resource Test {
+  array: Array<cloud.Bucket>;
+
+  init() {
+    this.array = [new cloud.Bucket()];
+  }
+
+  inflight test() {
+    assert(this.array.at(0).list().length == 0);
```
eladb, line 11: We don't know which operation is performed on the contents of the array, and I am planning to make this fail as part of https://github.com/winglang/wing/pull/1610

So for now, this is will only fail at runtime (not ideal, but I am on it).

examples/tests/invalid/inflight_ref_resource_sub_field.w
```diff
@@ -0,0 +1,32 @@
...
+resource Test {
+  another: Another;
+
+  init() {
+    this.another = new Another();
+  }
+
+  inflight test() {
+    let res = this.another.func();
+    this.another.my_queue.push("message");
```
eladb, line 26: This fails in *preflight*, not during compilation.

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
 
 pub struct JSifier<'a> {
+	pub diagnostics: Diagnostics,
 	pub out_dir: &'a Path,
 	shim: bool,
 	app_name: String,
 	inflight_counter: RefCell<usize>,
 }
 
+/// Analysizes a resource inflight method and returns a list of fields that are referenced from the
```
Chriscbr, line 59:
```suggestion
/// Analyzes a resource inflight method and returns a list of fields that are referenced from the
```

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
+		(self.references, self.diagnostics)
+	}
+}
+
+impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
+	fn visit_reference(&mut self, node: &'ast Reference) {
+		match node {
+    	Reference::InstanceMember { object, property } => {
+				match &object.kind {
+        	ExprKind::Reference(r) => {
```
Chriscbr, line 99: We shouldn't mix spaces and tabs. Maybe we can add linting as a build step

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
+							}
+							_ => {}
+						}
+					}
+					_ => {}
+    		}
+
+				self.path.insert(0, property.name.clone());
+			},
+			_ => {}
```
Chriscbr, line 217: Why are we not handling these cases? (_ => {}) (Maybe we can add comments saying which cases don't need to be handled by design and which are future TODOs, if possible)

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
+																level: DiagnosticLevel::Error,
+																message: format!(
+																	"Unable to reference \"{}\" from inflight method \"{}\" because it is not an inflight method",
+																	self.path.join("."), 
+																	self.method_name.name,
+																),
+																span: Some(property.span.clone()),
+															});
+														}
+													}
```
Chriscbr, line 190: Refactor into helper function? like `verify_operation_is_inflight`

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
+
+									// lookup our field in the class environment
+									match &*resource_type {
+            				Type::Resource(r) => {
+											let field_kind = r.env
+												.lookup(property, None).expect("unable to find field in resource env")
+												.as_variable().expect("field is not a variable");
+
+											// we only care about preflight fields (inflight fields are normal references)
+											if field_kind.flight == Phase::Preflight {
```
Chriscbr, line 117: Reorder conditions to bail out early -- `if field_kind.flight != Phase::Preflight` then `break` or `continue` etc. (reduce nesting)

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
+														for (name, typeref) in r.methods(true) {
+															if name == op {
+																match &*typeref {
+																	Type::Function(f) => {
+																		if f.flight == Phase::Inflight {
+																			found_method = true;
+																		}
+																	}
+																	_ => {}
+																}
```
Chriscbr, line 174: I think we can use `typeref.as_function_sig()` here

libs/wingc/src/jsify.rs
```diff
@@ -41,21 +42,190 @@ const $App = __app(process.env.WING_TARGET);
...
+												}
+
+												debug!("field \"this.{}\" is referenced from inflight method {}", property.name, self.method_name.name);
+
+												// get/create an "ops" set for this field
+												let mut ops = if let Some(ops) = self.references.get(&property.name) {
+													ops.clone()
+												} else {
+													IndexSet::new()
+												};
```
Chriscbr, line 156: The idiomatic way to get-and-mutate a value in a map in Rust without having to call `get` and `insert` two separate times is to use the [Entry](https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html) API -- I think it can probably be used here? 
```suggestion
												let mut ops = self.references.entry(property.name.clone()).or_default();
```

examples/tests/valid/inflight_ref_resource.w
```diff
@@ -0,0 +1,16 @@
+bring cloud;
+
+resource Test {
+  b: cloud.Bucket;
+
+  init() {
+    this.b = new cloud.Bucket();
+  }
```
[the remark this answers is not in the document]
eladb: added another test

libs/wingc/src/jsify.rs
eladb: Fixed

libs/wingc/src/jsify.rs
eladb: 👍

libs/wingc/src/jsify.rs
eladb: All of them don't need to be handled by design. Rewrote as simple `if let` statements to signal that this is all by design.

libs/wingc/src/jsify.rs
eladb: Done

libs/wingc/src/jsify.rs
eladb: Done!

libs/wingc/src/jsify.rs
eladb: done

libs/wingc/src/jsify.rs
```diff
@@ -1149,3 +1169,181 @@ fn is_mutable_collection(expression: &Expr) -> bool {
...
+		if name != *method_name { continue }
+
+		if let Some(sig) = typeref.as_function_sig() {
+			if sig.flight == Phase::Inflight {
+				return true
+			}
+		}
+	}
+
+	false
```
yoav-steinberg, line 1348: the iteration here is bad, try something like this, 
```suggestion
   let method_flight = resource_class.env.try_lookup(method_name, None).map(|s| s.as_variable().unwrap().flight)
   if !matches(method_flight, Some(Phase::Inflight)) {
     true
   } else {
     false
    }     
```

examples/tests/invalid/inflight_ref_invalid.w
```diff
@@ -0,0 +1,17 @@
+resource Foo {
+  mut_array: MutArray<str>;
+  var reassignable: num;
```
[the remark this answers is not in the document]
eladb: DONE!

libs/wingc/src/jsify.rs
```diff
@@ -1103,6 +1100,29 @@ impl<'a> JSifier<'a> {
 		)
 	}
 
+	/// Get the type and capture info for fields that are captured in the client of the given resource
+	/// Returns a map from method name to a map from field name to a set of operations
+	fn find_inflight_references(&mut self, resource_class: &AstClass) -> BTreeMap<String, BTreeMap<String, IndexSet<String>>> {
```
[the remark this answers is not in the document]
eladb: Done

libs/wingc/src/jsify.rs
eladb: OK, changed.

Curious - why is it bad?

(performance wise it has the same performance, right?)

yoav-steinberg: performance. `try_lookup` does a lookup in the env (O(1)), while iterating over the methods and comparing them is O(number_of_methods).

libs/wingc/src/jsify.rs
```diff
@@ -1149,3 +1168,178 @@ fn is_mutable_collection(expression: &Expr) -> bool {
...
+/// Checks if a resource class has an inflight method with the given name
+fn has_inflight_method(resource_class: &Class, method_name: &String) -> bool {
+	let method_flight = resource_class.env
+		.try_lookup(method_name, None)
+		.map(|s| s.as_variable().unwrap().flight);
+	
+	return match method_flight {
+		Some(Phase::Inflight) => true,
+		_ => false
+	}
```
yoav-steinberg, line 1344: Actually even this isn't ideal because I don't verify it's actually a method (and also we might want to verify it's not static). Anyway I added a helper method to do all this in `main`. So something like this:
```suggestion
	if let Some(method_var) = resource_class.get_method(method_name) {
	  method_var.flight == Phase::Inflight && !method_var.is_static
	} else {
  	  false
  	}
}
```

libs/wingc/src/jsify.rs
```diff
@@ -1149,3 +1168,178 @@ fn is_mutable_collection(expression: &Expr) -> bool {
...
+				return true
+			}
+
+			debug!("field \"this.{}\" is referenced from inflight method {}", property.name, self.method_name.name);
+
+			// get/create an "ops" set for this field
+			let ops = self.references.entry(property.name.clone()).or_default();
+
+			if ! self.path.is_empty() {
+				let op = self.path[0].clone();
```
yoav-steinberg, line 1303: I got a little mixed up with this, why do we need only the top of the `path`?
And What happens when we access an inner field: `this.some_field.inner_field.do()`?

libs/wingc/src/jsify.rs
eladb: Updated.

libs/wingc/src/jsify.rs
eladb: See [this test](https://github.com/winglang/wing/blob/eladb/infer-refs/examples/tests/invalid/inflight_ref_resource_sub_field.w).

Not ideal, but it's a start.

eladb: OK, just saw this. Next time hit "unresolve conversation" if you think there's still something to cover...

eladb: See https://github.com/winglang/wing/pull/1682/commits/b4a30ffad00e0dad97d19ac14a496463408ed036 - basically failing this should be covered in the next PR.

eladb:
@yoav-steinberg this is ready for another round

libs/wingc/src/jsify.rs
```diff
@@ -1148,3 +1167,175 @@ fn is_mutable_collection(expression: &Expr) -> bool {
...
+			_ => panic!("'this' is not a resource"),
+		};
+
+		// lookup our field in the class environment
+		let field_kind = r.env
+			.lookup(property, None).expect("unable to find field in resource env")
+			.as_variable().expect("field is not a variable");
+
+		// we only care about preflight fields (inflight fields are normal references)
+		if field_kind.flight == Phase::Preflight {
```
Chriscbr, line 1263: Perhaps
```suggestion
		// we only care about preflight fields (inflight fields are normal references)
		if field_kind.flight != Phase::Preflight {
			return false
		}
```

libs/wingc/src/jsify.rs
```diff
@@ -1148,3 +1167,175 @@ fn is_mutable_collection(expression: &Expr) -> bool {
...
+}
+
+/// Checks if a resource class has an inflight method with the given name
+fn has_inflight_method(resource_class: &Class, method_name: &String) -> bool {
+	if let Some(method_var) = resource_class.get_method(method_name) {
+		method_var.flight == Phase::Inflight && !method_var.is_static
+	} else {
+		false
+	}
+}
```
Chriscbr, line 1341: nit: I don't think this needs to be a free floating function. Also it's more idiomatic to use `&str` than `&String`, since the latter requires the string to be heap-allocated:
```suggestion
impl Class {
	/// Checks if a resource class has an inflight method with the given name
	fn has_inflight_method(method_name: &str) -> bool {
		// ...
	}
}
```

yoav-steinberg (approved):

libs/wingc/src/jsify.rs
eladb: Thank you never nester!

libs/wingc/src/jsify.rs
yoav-steinberg: I feel adding `has_inflight_method` to `Class` is too specific. I'll refactor and push a cleanup.

eladb: I inlined this condition

Issue #1673, opened 2023-02-27
Differentiate files and directories when listing bucket contents


polamoros:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

`bucket.list` returns information about each file/dir inside a bucket.

```
const response = await bucket.list();

[
  {
    name: "a",
    type: "directory",
    size: "196", // bytes
    last_modified: "1057984701"
  },
  {
    name: "hello.txt",
    type: "file",
    size: "12", // bytes
    last_modified: "1057984701"
  }
]
```

`bucket.list` has an option to retrieve files in a specific path.

```
const response = await bucket.list(undefined, 'a');

[
  {
    name: "hello_inside_a.txt",
    type: "file",
    size: "12", // bytes
    last_modified: "1057984701"
  }
]
```

### Use Cases

As a Console user, I would like to be able to see files and directories inside buckets.

<img width="184" alt="image" src="https://user-images.githubusercontent.com/5547636/221523671-7d3a76eb-8dc0-4b9a-94a3-d13a66ffb56b.png">

<img width="676" alt="image" src="https://user-images.githubusercontent.com/5547636/221523741-d03de3c4-99f8-40d7-a98d-01cd62374558.png">

### Component

SDK

skyrpex:
I'd propose using an options object for the list method:

```ts
const response = await bucket.list({ prefix: "my_prefix", path: "my_dir/" });
```

Chriscbr:
So, the interesting thing here is that technically `cloud.Bucket` only provides an abstraction for files. That is, you cannot create or delete directories on AWS S3 or Azure Blob Storage etc. This is because at the underlying hardware level, they are not storing your objects using a file system, but simply as a distributed key-value store  -- the fact that most keys resemble file paths is just convention. 😅 

*However*, as you've noted, we do want to provide this illusion of being able browse objects as if it were a file system. Since this use case is so common, the way AWS and Azure APIs support this (and how I understand their web consoles work under the hood) is they offer a "delimiter" option in their API. The way it works is like this:

Given these files:
```
pictures/January/photo1.png
pictures/January/photo2.png
pictures/February/photo1.png
pictures/February/photo2.png
```

Then you can call the List API on AWS with `prefix: "pictures/", delimiter: "/"` to obtain a list of pseudo-folders inside `pictures`:

```
pictures/January/
pictures/February/
```

My proposal would be:
[...]

PR #1692, opened 2023-03-01
feat(sdk): change bucket simulator put() to make directories recursively to avoid 'no such file or directory' on fs.writeFile

### LINKED ISSUE #1674: Create directories inside buckets
### Community Note

### Feature Spec

When uploading a file to a bucket, I can create it inside a directory of my choice.

### Use Cases

Currently, if I try to upload a file in a path that doesn't exist (`new-directory`), I get the following error:

image

The directory should be created if it doesn't exist.

### Implementation Notes

_No response_

### Component

SDK

YarivLevy81:
Should resolve #1674.
I used `mkdir` with `{recursive:true` to make sure the path exists. 
Additionally, changed the `list()` implementation to list by directories, instead of reading for fileDir and filtering by prefix - hope it makes sense. When trying to mimic s3, it did made sense to me.


libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -1,11 +1,11 @@
 import * as fs from "fs";
 import * as os from "os";
-import { join } from "path";
-import { ISimulatorResourceInstance } from "./resource";
-import { BucketSchema } from "./schema-resources";
-import { exists } from "./util";
-import { BucketDeleteOptions, IBucketClient } from "../cloud";
-import { ISimulatorContext } from "../testing/simulator";
+import {dirname, join} from "path";
+import {ISimulatorResourceInstance} from "./resource";
```
YarivLevy81, line 4: This was automatically added by Intellij, I can revert this.

eladb (approved):
Good stuff

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -57,10 +59,7 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     return this.context.withTrace({
       message: `List (prefix=${prefix ?? "null"}).`,
       activity: async () => {
-        const fileNames = await fs.promises.readdir(this.fileDir);
-        return prefix
-          ? fileNames.filter((fileName) => fileName.startsWith(prefix))
-          : fileNames;
+        return fs.promises.readdir(`${this.fileDir}/${prefix ?? ""}`);
```
Chriscbr, line 62: Does `list` read all files out recursively (even in sub-directories)? I think we need to add a unit test case like this to match up with how AWS's ListObjects API works (and other clouds too):

```ts
await bucket.put("foo");
await bucket.put("foo/bar");
await bucket.put("foo/bar/baz");
const resp = await bucket.list(); // no prefix
expect(resp).toEqual([
  "foo",
  "foo/bar",
  "foo/bar/baz",
]);
```

Chriscbr (changes requested):

libs/wingsdk/src/target-sim/bucket.inflight.ts
YarivLevy81: It appears that the behavior of the list may be influenced by whether it represents `s3api list-objects` or `s3 list`. 
In my opinion, considering that the AWS s3 console likely utilizes `s3 list`, I would anticipate a similar behavior in the wing console. Nonetheless, I am a newcomer to this platform, and you possess a better understanding of the system, so please feel free to correct me if I am wrong.

Chriscbr: For now I think we will start with modeling after `s3api list-objects` as it's the lower-level of the two APIs. (`s3 list` can be implemented by calling `list-objects` and using the `delimiter` option, and then formatting the output a bit). There's some more discussion about the API here: https://github.com/winglang/wing/issues/1673

In any case it's probably fine to handle this issue separately / it doesn't need to be part of this PR, so I'll approve 😄

Chriscbr (approved):

YarivLevy81:
Looks like some unrelated failures happened on SDK mutation?

PR #1707, opened 2023-03-02
fix(compiler): diagnostics issues and cleanups

yoav-steinberg:
* Correct diagnostics for binary expr checking
* test equal operator on different types
* Revert diagnostics set back to vector and make sure there are no duplicate errors.
* Colors in error reporting
* Don't validate non boolean result binary operators evaluate to a number (this isn't necessarily the case and produces a dup error in some cases).
* Unary operators evaluate to the expected type if they receive a wrong argument - reduce redundant errors.
* Avoid calling resolve_reference twice to check if a call is static - reduce redundant errors.
* Fix bug where hydration didn't mark class fields as non-static.
* Generate diagnostic if `get_property_from_class` tries to access a static field.
* Updated failure snapshots accordingly.

color diags:
image


eladb (approved):

libs/wingc/src/diagnostic.rs
```diff
@@ -108,20 +107,30 @@ pub enum DiagnosticLevel {
...
+			DiagnosticLevel::Error => write!(f, "{}", "Error".bold().red()),
+			DiagnosticLevel::Warning => write!(f, "{}", "Warning".bold().yellow()),
+			DiagnosticLevel::Note => write!(f, "{}", "Note".bold().blue()),
+		}
+	}
+}
+
 impl std::fmt::Display for Diagnostic {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
 		// TODO: implement a Display for DiagnosticLevel (instead of Debug formatting)
```
Chriscbr, line 129: We can remove this comment

Chriscbr (approved):

examples/tests/valid/capture_containers.w
```diff
@@ -18,7 +18,7 @@ let handler = inflight (s: str): str => {
   assert(my_map.size() == 2);
 
   assert(arr_of_map.at(0).has("bang"));
-  assert("world" == j.get("b"));
+  assert(j.get("b") == "world");
```
Chriscbr, line 21: Is this just a stylistic change?

Let's add a check that both should work:
```suggestion
  assert(j.get("b") == "world");
  assert("world" == j.get("b"));
```

yoav-steinberg: It's not a stylistic change see #1717. I discussed with @hasanaburayyan and for now this switch is a workaround to get the tests to pass until we handle the issue.

PR #1709, opened 2023-03-02
chore(compiler): improve jsii import by removing unnecessary clones

MarkMcCulloh:
This change required manual changes to the generated jsii manifest schema, so I went ahead and removed the auto-updating logic from wingii for now. The jsii spec is pretty stable so I think this okay.

My local compile with bring_cdktf went from 36 seconds to 20 seconds.


libs/wingii/src/lib.rs
```diff
@@ -322,7 +332,7 @@ pub mod type_system {
 							interfaces.insert(iface);
 						}
 					}
-					interfaces.insert(iface_type);
+					interfaces.insert(iface_type.clone());
```
MarkMcCulloh, line 335: Had to add a new clone here, but this is much less expensive than the ones we avoided

Chriscbr (approved):
🙏  awesome performance improvement!

This is what we need!

eladb (commented):
Awesome!

Issue #435, opened 2022-11-03
Phase independent (unphased) functions, methods, fields


staycoolcall911:

Chriscbr:
related - https://github.com/winglang/wing/issues/7174

PR #1711, opened 2023-03-03
rfc: `unphased` functions

Chriscbr:
This doc proposes how phase-independent functions should look and behave in Winglang. Related to #435

From a compiler perspective, these functions would essentially be code generated in both the preflight JS class and the inflight JS class.


docs/04-reference/winglang-spec.md
```diff
@@ -765,15 +767,95 @@ let handler2 = inflight() => {
...
 
 Preflight resource methods and initializers can receive an inflight function as
 an argument. This enables resources to define code that will be executed on the
 deployed resource (lambda functions, docker, virtual machines etc).
 
 [`▲ top`][top]
 
+#### 1.3.1 Phase-independent code
+
+Code that is not dependent on the phase of execution can be designated as phase-independent using the `?inflight` modifier.
```
skyrpex, line 781: What if all functions that don't have an explicit phase modifier are maybe-inflights by default?

Chriscbr: @skyrpex If I'm understanding correctly, you're proposing one of two alternatives. Either (1) to have the compiler infer that a function is phase-independent where possible, OR (2) to make phase-independent the default for functions, and make preflight-only functions require a keyword (like `preflight () => {}`).

I'll try to explain why I chose my current proposal over those options (will also add to the PR description)

----

For the first alternative, I feel like the phase of a function isn't something we should allow the compiler to infer -- not because it's not technically feasible, but of user experience problems it could introduce.

In the future, devs will be able to create a Wing libraries that export resources with public preflight methods and public inflight methods. When using a Wing library, consumers should be able to rely on the fact that the method always works in that phase in the future, even if it the implementation changes -- just like today you can rely on the fact that a if a imported TypeScript function returns a boolean, then it will always return a boolean in the future (unless there is a breaking change).

If the compiler inferred whether a function should be phase-independent or preflight, then changing the implementation of the function so that it captures some mutable data could change the function's type to preflight, thus causing a potentially breaking change to consumers.

----

For the second alternative, I think the idea of functions being phase-independent as the default could be interesting. But I feel like it might be more confusing since a lot of the Wing language is currently designed around the concept that "the top level of your code is preflight" -- for example, creating a bucket at the top-level scope of a Wing file happens in preflight. And the default constructor of a resource is its preflight constructor, which matches up with that. 

Put another way, I think preflight-only functions will be more common than phase-independent functions, so I think there should be more friction for the phase-independent use case than the preflight code use case. But this is definitely something we can experiment with.

Chriscbr:
Another idea/suggestion is to designate these functions with the keyword `pure`

hasanaburayyan:
Im on board, I just dont much care for the prefixed `?` maybe a keyword instead like pure but no strong opinions here.

ShaiBer:
I agree that letting the compiler infer whether a function is phase-independent is not the right way to go.
I also agree that making functions phase-independent should not be the default because it sends the wrong message about which type of function we expect to be more prevalent if writing correct wing code.

Like @hasanaburayyan, I'm also not crazy about the `?inflight` keyword. I like Rust's approach with `maybe-inflight` better, but it also doesn't feel like the best solution to me. I hope someone will come up with a better alternative everyone will like.

I'm also OK with experimenting with having to always put a phase modifier on functions, and then we have `preflight`, `inflight` and `maybe-inflight` (or `?inflight` or something better someone else will come up with)

ekeren:
@Chriscbr , great doc
Can you also add the reasoning and need for phase-independent methods, have you came across any concrete use cases?

marciocadev:
Perhaps an alternative is to create functions that are neither `inflight` nor `preflight`, some `flight` functions

and add the prefix when assigning to a variable

something like
```ts (wing)
let f = flight(...) { ... };
let p = pre f; // create a preflight function
let i = in f; // create a inflight function
```

ShaiBer:
> @Chriscbr , great doc Can you also add the reasoning and need for phase-independent methods, have you came across any concrete use cases?

Maybe I'm missing something, but I think any project has general purpose utility functions that are used throughout the code and have no business being confined to only one phase. For example: sorting an array with different algorithms

Chriscbr:
Yes, I think @ShaiBer 's answer is correct, the main use case that comes to my mind is utility functions that do not need to be restricted to one phase or another. If I think of any more use cases I'll update the proposal.

BTW, my initial guess is we don't need this feature for beta, or at least we should dogfood more before considering it.

@marciocadev that's a very interesting idea! From an implementation angle, I think that could be a good option if we want to simplify modeling phase-independent functions in the compiler (perhaps we would only need our [`Phase` enum](https://github.com/winglang/wing/blob/5b1c4ab3c0d27bcdf10eaff5c40d2f59ddcc80c6/libs/wingc/src/ast.rs#L65-L69) to have two choices instead of three).

But I'm a little cautious about templating based solutions / not sure I understand the benefits. In your example, suppose I use a preflight function inside `f` (this means it can only be used in preflight). Would I only get a compiler error once I write `let i = in f`? Or would I get an error on the `let f =` line?

marciocadev:
I was thinking in gives an error if `f` is declare and never used to create a inflight or preflight function

Some errors
```ts (wing)
bring cloud;

let f = flight ( ... ) { ... } // ERROR, something like f variable never used
```
```ts (wing)
bring cloud;

let f = flight ( ... ) { ... }

new cloud.Function(pre f); // ERROR, can not pass a preflight to cloud.Function
```
This works
```ts (wing)
bring cloud;

let f = flight ( ... ) { ... }

new cloud.Function(in f); // WORKS
```

Somehow, I think of a `preflight` function as a function that will only be performed during the stack deploy. It is something that can be executed during the deploy phase, either before building the stack, during its construction, or after it has been built.

```ts (wing)
bring cloud;

let f = flight ( ... ) { ... } // verify some data before deploying the stack
let p = pre f;
if (p.invoke() == false) {
   ...
}

let i = in f;

new cloud.Function() {
   ...
  // verify if data change
  if (i.invoke() == true) {
     ...
  }
  ...
}
```

Chriscbr:
Another naming suggestion from @MarkMcCulloh is "unphased"

skyrpex:
I'd expect Wing to **infer** the phase of the function most of the time, and also cast a function to inflight if necessary. In the same way that TypeScript can infer that the following function returns a number:

```ts
export const sum = (a: number, b: number) => {
  return a + b;
};
```

Regarding Chris' point: 

> When using a Wing library, consumers should be able to rely on the fact that the method always works in that phase in the future, even if it the implementation changes

I like Mark's `unphased` keyword to enforce that a function works in both preflight and inflight phases.

Here's how I'd like it to work in different scenarios:

### Compiler infers that the function is inflight

```wing
bring cloud;

new cloud.Function(() => {
  return "hello";
});
```

### Compiler can cast phase-independent functions to inflight (if they don't interact with resources nor capture mutable state)

```wing
bring cloud;

let sum = (a: num, b: num) => {
  return a + b;
}

log(sum(1, 1));

new cloud.Function(() => {
  return sum(2, 2);
});
```

### Compiler infers the function is inflight so it has access to bucket's inflight API

```wing
bring cloud;

new cloud.Function(() => {
  bucket.put("hello", "world");
});
```

### Compiler infers that "put" is preflight because it accesses the bucket

```wing
bring cloud;

let putFile = (key: str, value: str) => {
  bucket.put(key, value);
  // ^ Error: "bucket.put" is an inflight method but "putFile" is a preflight method. Try adding the inflight keyword to "putFile".
};
```

#### Compiler infers that "append" is preflight because it captures mutable state

```wing
bring cloud;

let mut state = "";

let append = (value: str) => {
  state = "${state} ${value}";
};

new cloud.Function(() => {
  append("hello");
  // ^ Error: Can't use "append" because it's a preflight function that captures mutable state.
});
```

Chriscbr:
@skyrpex Feel free to add a +1 to the issue about inferring function phases here: https://github.com/winglang/wing/issues/3755 

The intuition in your examples is correct - I think unphased functions should be possible to use both in places that expect preflight functions and places that expect inflight functions.

docs/contributing/999-rfcs/2023-06-12-language-spec.md
```diff
@@ -776,15 +777,121 @@ let handler2 = inflight() => {
...
+#### 1.3.1 Phase-independent code
+
+> **Note**: Phase-independent functions are not yet implemented. Subscribe to [issue #435](https://github.com/winglang/wing/issues/435) for updates.
+
+Code that is not dependent on the phase of execution can be designated as phase-independent using the `unphased` modifier.
+
+Using this modifier means that the function can be called from either preflight or inflight contexts.
+
+```TS
+let odd_numbers = unphased (arr: Array<num>): Array<num> => {
```
eladb, line 798: I like the suggestion but I prefer `inflight?` instead of introducing another keyword to the language.

I think it reads nicely:

```js
let foo = inflight? () => {
  return 34;
}
```

hasanaburayyan: With just `inflight`, `preflight`, `unphased` perspective I think it reads nice, but my concern here is the first thing I think of when I see `inflight?` is an optional inflight.

(Thinking outloud)
The RFC mentions:
```
However, a preflight or inflight function cannot be passed to a function that expects an unphased function.
```

How would an expected unphased function look in a method signature if we use `inflight?`

Chriscbr: I'm a little worried there could be a confusion or mental overloading with "?" since "foo?" looks a lot like an optional type. (It's also handy that "unphased" is fewer characters). But I'm not sure - let's keep the option open, it's probably the easiest part of the design to change 👍

eladb: I think that “optional inflight” is actually conveying the idea of “unphased” pretty well.

It implies that this is a preflight function that can also be used inflight.

We should be very careful to add additional keywords to the language. Every new keyword is a huge cognitive overload.

hasanaburayyan: I agree that thinking of the function as "optionally inflight" means it could be inflight or preflight.

However, we define optionality in the language as  potentially having a "lack of value" so T? is [T || nil], which makes `inflight?` being [inflight || preflight] weird in my opinion.

docs/contributing/999-rfcs/2023-06-12-language-spec.md
```diff
@@ -776,15 +777,121 @@ let handler2 = inflight() => {
...
+    return `s3://${this.name}/${key}`;
+  }
+}
+```
+
+Phase-independent methods take on the additional restriction that they
+cannot mutate fields of the resource. For example, the following is disallowed:
+
+```TS
+resource Bucket {
```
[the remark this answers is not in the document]
Chriscbr: Fixed

docs/contributing/999-rfcs/2023-06-12-language-spec.md
```diff
@@ -776,15 +777,121 @@ let handler2 = inflight() => {
...
+
+However, a preflight or inflight function cannot be passed to a function that expects an unphased function.
+An exception to this rule is that if a function is unphased, then we can automatically assume any functions passed to it or returned by it have a matching phase.
+
+You can imagine that when a function is unphased, then "preflight" and "inflight" versions of it are generated at compile-time, and unphased-function types in parameters or return types are automatically converted to the appropriate phase.
+
+For example, `Array<T>.map` is modeled like the following pseudocode:
+
+```js
+native class Array<T> {
```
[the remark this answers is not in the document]
Chriscbr: OK - yeah this is probably confusing. Removed

Chriscbr:
I've added an addendum with a proposed mechanism for how unphased extern functions can be implemented (it's not a high priority to support this immediately, but this will help us build a path forward for rewriting various parts of the `util` and `fs` modules in Wing.

skyrpex:
Random thought: `unphased` vs `universal`?

docs/api/05-language-reference.md
```diff
@@ -1983,6 +2086,36 @@ matching name (without any case conversion).
 
 Extern methods do not support access to class's members through `this`, so they must be declared `static`.
 
+If an extern function is declared as `unphased`, a preflight and inflight implementation must be provided, by providing two separate functions in the JavaScript module.
```
skyrpex, line 2089: Maybe we could also still accept a function, if both implementations are the same. I'm just trying to avoid clutter.

So it could be:

```js
exports.randomHex = async function (length) {
    const buffer = Buffer.alloc(length);
    await randomFillAsync(buffer);
    return buffer.toString("hex");
}
```

but also:

```js
exports.randomHex = {
  preflight: function (length) {
    const buffer = Buffer.alloc(length);
    crypto.randomFillSync(buffer);
    return buffer.toString("hex");
  },
  inflight: async function (length) {
    const buffer = Buffer.alloc(length);
    await randomFillAsync(buffer);
    return buffer.toString("hex");
  }
```

A simple check could be used to detect both cases: `exports.randomHex instanceof Function ? "single implementation" : "dual implementation"`.

skyrpex: Of course, this assumes externs can only be functions.

Chriscbr: Sounds ok to me as an extra enhancement (I probably wouldn't implement it for the MVP but maybe later).

Note: if you only provide a single function, it will need to be sync (not async), otherwise it can't be safely called in preflight.

Chriscbr: Also FWIW possible to reuse a single implementation twice in the JavaScript (so I'm not sure it's worth slowing down performance by checking if externs are `instanceof Function` each time an extern is called).

```
function randomHex(length) {
  const buffer = Buffer.alloc(length);
  await randomFillAsync(buffer);
  return buffer.toString("hex");
}
exports.randomHex = { preflight: randomHex, inflight: randomHex };
```

Issue #1497, opened 2023-02-09
cloud.Bucket.put_json()


hasanaburayyan:
https://docs.winglang.io/reference/wingsdk-spec#bucket

hasanaburayyan:
Ooops jumped the gun a little early on making this issue, was playing with the Github subtasks 😅 updating it with details now

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.91](https://github.com/winglang/wing/releases/tag/v0.5.91).

PR #1712, opened 2023-03-03
feat(sdk): support bucket put json

hasanaburayyan:
Adds the `get_json` and `put_json` api to `cloud.Bucket`

Closes: https://github.com/winglang/wing/issues/1497

Screenshot:
image

### DIFF
--- a/examples/tests/valid/json_bucket.w
+++ b/examples/tests/valid/json_bucket.w
@@ -19,7 +19,7 @@ let j = Json {
 
 let get_json = new cloud.Function(inflight(msg:str): str => {
   let x = b.get_json(file_name);
-  assert("failure" == x.get("persons").get_at(0).get("fears").get_at(1));
+  assert(x.get("persons").get_at(0).get("fears").get_at(1) == "failure");
 });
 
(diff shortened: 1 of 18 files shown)


examples/tests/valid/json_bucket.w
```diff
@@ -0,0 +1,29 @@
...
+        "failure"
+      ]
+    }
+  ]
+};
+
+
+let get_json = new cloud.Function(inflight(msg:str): str => {
+  let x = b.get_json(file_name);
+  assert("failure" == x.get("persons").get_at(0).get("fears").get_at(1));
```
eladb, line 22: How does the type checker know that this is a string?

eladb (approved):

examples/tests/valid/json_bucket.w
hasanaburayyan: It doesnt really, its because right now everything is a subtype of Json which is probably not ideal. It was just a quick way to get implicit casting from native to Json. 
See: https://github.com/winglang/wing/issues/1717 

I talked with @yoav-steinberg about this earlier as he pointed it out and looking to start some conversation in the above issue.

PR #1721, opened 2023-03-03
fix: compiling cloud function mangles terminal output

Chriscbr:
Adds a workaround for an issue where compiling cloud functions (which calls esbuild) results in mangled terminal output.

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -325,15 +325,14 @@ impl<'a> JSifier<'a> {
 					None
 				};
 
-				if !is_resource || fqn.is_none() {
-					format!("new {}({})", ctor, args)
-				} else {
-					let fqn = fqn.expect("expecting fqn to be defined");
+				if let (true, Some(fqn)) = (is_resource, fqn) {
 					if is_abstract {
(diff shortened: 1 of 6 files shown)


libs/wingc/src/jsify.rs
```diff
@@ -325,15 +325,14 @@ impl<'a> JSifier<'a> {
 					None
 				};
 
-				if !is_resource || fqn.is_none() {
-					format!("new {}({})", ctor, args)
-				} else {
-					let fqn = fqn.expect("expecting fqn to be defined");
+				if let (true, Some(fqn)) = (is_resource, fqn) {
```
yoav-steinberg, line 328: Is this related to this PR? What's the fix here?

yoav-steinberg (approved):
This seems fine, can you explain in the top comment the details of the workaround?

libs/wingc/src/jsify.rs
Chriscbr: Just a refactoring. The way it's currently written, the second branch has to call "expect" to unwrap the FQN even though logically we _know_ the fqn is a `Some(String)`.

eladb (commented):
🫡

PR #1727, opened 2023-03-04
feat(sdk): adding table resource (sim/tf-aws)

### LINKED ISSUE #1313: cloud.Table - Simulator implementation

### LINKED ISSUE #1314: cloud.Table - AWS implementation

marciocadev:
Adding table resource to AWS Terraform provider and sim

Closes #1313
Closes #1314


staycoolcall911:
This is awesome @marciocadev! We'll give it a thorough review very soon.

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Inflight interface for `Table`.
+ */
+export interface ITableClient {
+  /**
+   * Insert a row into the table.
+   * @param row data to be inserted.
+   * @inflight
+   */
+  insert(row: Json): void;
```
hasanaburayyan, line 91: Curious why row is Json type here. In the spec we have `insert(row: Map<str, Json>)` which I assume is supposed to be a column to Json content mapping. IE.
```ts
let row = {
  "id": Json "some_id",
  "headers": Json {
    some_header: "value",
    numbers: [1, 2, 3]
  }
};

table.insert(row);
```

Maybe the DX is nicer with just Json, but wanted to call out the spec difference.

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Update a row in the table.
+   * @param row data to be updated.
+   * @inflight
+   */
+  update(row: Json): void;
```
hasanaburayyan, line 97: Same question as above

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+   * @inflight
+   */
+  delete(key: any): void;
+  /**
+   * Increments the counter atomically by a certain amount and returns the previous value.
+   * @param key primary key to search.
+   * @returns get the row from table.
+   * @inflight
+   */
+  get(key: any): any;
```
hasanaburayyan, line 110:
```suggestion
  get(key: string): Map<string, Json>;
```

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Increments the counter atomically by a certain amount and returns the previous value.
+   * @returns list all row.
+   * @inflight
+   */
+  list(): any;
```
hasanaburayyan, line 116: Lets add a TODO: comment here to change the return type to `Iterator<Map<string, Json>>` as well as create an issue for tracking.

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+   */
+  update(row: Json): void;
+  /**
+   * Delete a row from the table.
+   * @param key primary key to delete the row.
+   * @inflight
+   */
+  delete(key: any): void;
+  /**
+   * Increments the counter atomically by a certain amount and returns the previous value.
```
hasanaburayyan, line 105: This description is for counter resource

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+   * @param key primary key to search.
+   * @returns get the row from table.
+   * @inflight
+   */
+  get(key: any): any;
+  /**
+   * Increments the counter atomically by a certain amount and returns the previous value.
```
hasanaburayyan, line 112: This description is for counter resource

libs/wingsdk/src/target-tf-aws/table.ts
```diff
@@ -0,0 +1,110 @@
+import { DynamodbTable } from "@cdktf/provider-aws/lib/dynamodb-table";
+import { Construct } from "constructs";
+import { Function } from "./function";
+import * as cloud from "../cloud";
+import * as core from "../core";
+import { NameOptions, ResourceNames } from "../utils/resource-names";
+
+/**
+ * Counter (Table) names must be between 3 and 255 characters.
```
hasanaburayyan, line 9:
```suggestion
 * Table names must be between 3 and 255 characters.
```

libs/wingsdk/src/target-tf-aws/table.ts
```diff
@@ -0,0 +1,110 @@
...
+ * You can use alphanumeric characters, dot (.), dash (-), and underscores (_).
```
hasanaburayyan, line 10: I know we dont do this yet for other resources but perhaps we should start adding links to reference where the naming convention rules came from in documentation. Since these are target specific it might just make it easier for future development.

libs/wingsdk/src/target-tf-aws/table.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * AWS implementation of `cloud.Table`.
+ *
+ * @inflight `@winglang/sdk.cloud.ITableClient`
+ */
+export class Table extends cloud.Table {
+  private readonly table: DynamodbTable;
+
+  constructor(scope: Construct, id: string, props: cloud.TableProps) {
```
hasanaburayyan, line 25:
```suggestion
  constructor(scope: Construct, id: string, props: cloud.TableProps = {}) {
```

libs/wingsdk/src/target-tf-aws/table.ts
```diff
@@ -0,0 +1,110 @@
...
+    super(scope, id, props);
+
+    if (this.columns[this.primaryKey] === undefined) {
```
hasanaburayyan, line 28: Yea if we use a Map (see comments in cloud.table.ts) we could change this to be someting like
```suggestion
    if (this.columns.has(this.primaryKey)) {
```

libs/wingsdk/src/target-tf-aws/table.ts
```diff
@@ -0,0 +1,110 @@
...
+      throw new Error(`${this.primaryKey} not found on columns`);
```
hasanaburayyan, line 29:
```suggestion
      throw new Error(`${this.primaryKey} not found in columns`);
```

hasanaburayyan (changes requested):
:rocket: Looking strong, I did not completely finish review yet but will pick it back up in the morning. For now though I have a few questions and change requests.

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+   * @param row data to be updated.
+   * @inflight
+   */
+  update(row: Json): void;
+  /**
+   * Delete a row from the table.
+   * @param key primary key to delete the row.
+   * @inflight
+   */
+  delete(key: any): void;
```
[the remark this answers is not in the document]
marciocadev: I had some doubts about this, will the `primary key` always be a string?

DynamoDB accepts string, number or binary types for primary key

libs/wingsdk/src/cloud/table.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Table primary key name
+   */
+  public readonly primaryKey: string;
+  /**
+   * Table columns
+   */
+  public readonly columns: { [key: string]: ColumnType };
```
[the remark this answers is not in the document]
marciocadev: I tryed this before but gives an JSii error in `TableProps`, so I changed to this object `{ [key: string]: ColumnType }`

`src/cloud/table.ts:35:12 - error JSII3001: Type "Map" cannot be used as the property type because it is private or @internal`

libs/wingsdk/src/cloud/table.ts
marciocadev: same error

`src/cloud/table.ts:111:15 - error JSII3001: Type "Map" cannot be used as the parameter type because it is private or @internal`

libs/wingsdk/src/cloud/table.ts
marciocadev: Change to Map give this error

```bash
src/cloud/table.ts:111:10 - error JSII9002: Unable to resolve type "Map". It may be @internal or not exported from the module's entry point (as configured in "package.json" as "main").

       111   insert(row: Map<string, Json>): void;
                    ~~~~~~~~~~~~~~~~~~~~~~
       src/cloud/table.ts:111:15 - error JSII3001: Type "Map" cannot be used as the parameter type because it is private or @internal

       111   insert(row: Map<string, Json>): void;
                         ~~~~~~~~~~~~~~~~~

         node_modules/jsii/node_modules/typescript/lib/lib.es2015.collection.d.ts:36:13
           36 declare var Map: MapConstructor;
                          ~~~~~~~~~~~~~~~~~~~
           The referenced type is declared here
```

libs/wingsdk/src/cloud/table.ts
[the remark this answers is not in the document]
Chriscbr: Yeah -- unfortunately JSII doesn't allow public APIs using a many built-in JavaScript types like `Set`, `Map`, `Date`, `Error`, etc. so I think a plain object map will have to do. 😕 (I think the reason is that many of these types don't have equivalents in the other JSII target languages)

libs/wingsdk/src/cloud/table.ts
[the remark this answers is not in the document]
Chriscbr: I would stay with strings for now (we don't want to over-design to the specifics of DynamoDB or a single cloud provider)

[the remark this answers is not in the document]
Chriscbr: I feel like dynamically adding new fields to the table after you create it might be an extra feature that we don't need to support in the API right now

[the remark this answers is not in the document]
marciocadev: Nice, I'll make some changes to ensure that

libs/wingsdk/src/cloud/table.ts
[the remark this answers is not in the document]
hasanaburayyan: Aw thanks JSii! lol yea @marciocadev feel free to disregard all my blabbering about Map types 😁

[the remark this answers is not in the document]
marciocadev: Don't worry man

I made the same mistake when I wrote these classes


JSii hurt my feelings too 😂😂😂

hasanaburayyan:
👋 @marciocadev is this ready for re-review? If so can you hit the re-request review and Ill make sure to get to this in the morning

libs/wingsdk/src/target-tf-aws/table.inflight.ts
```diff
@@ -0,0 +1,166 @@
...
+          break;
+        case ColumnType.JSON:
+          items[key] = this.unmarshallJson(row[key].M);
+          break;
+      }
+    }
+    return items;
+  }
+
+  private marshallJson(item: any) {
```
hasanaburayyan, line 122: Cant we just use the AWS SDK's util package for dynamodb to do this?
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_util_dynamodb.html

hasanaburayyan (commented):
Thanks for all the work and explanation with previous PR questions. Everything is looking awesome. Have a question about the marshaling methods.

libs/wingsdk/src/target-tf-aws/table.inflight.ts
marciocadev: Sure, I was previously using the util-dynamo

I bring them back again

marciocadev:
> Thanks for all the work and explanation with previous PR questions. Everything is looking awesome. Have a question about the marshaling methods.

@hasanaburayyan I have a question too, on update method I think will be better if we pass the id and the Json with what to update

Something like `update(id: str, row: Json);`

I think it's make more sense since we update a item, but never update the id (that will be crazy 🙃)

On insert it make sense pass only a Json with id as part of it, but on update I think this approach it wrong

What do you think ?

@staycoolcall911

hasanaburayyan (approved):
Awesome Job @marciocadev I think lets go ahead and take this resource in and see if we decide to make any changes to the API in the future but this gives us a tremendous place to start from!

PR #1729, opened 2023-03-04
fix(compiler): support std json in struct fields

hasanaburayyan:
Moves the replacement of STD lib types to be more encompassing rather than just on function signatures. Also adds replacement of std types from symbol lookups to builtin if needed.

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -1118,15 +1118,7 @@ impl<'a> TypeChecker<'a> {
 					};
 				}
 
-				// This replaces function signatures with valid TypeRefs if necessary. This step is also done in
-				// the hydration process for generics where we map std lib types like ImmutableMap to
-				// Type::Map(some_type). However for types like Json and MutJson which do
-				// not require hydration, we still need to Map the std `Json` type to type-checker's Json type
-				if let Some(t) = self.replace_with_builtin_type(func_sig.return_type.to_string().as_str()) {
-					t
-				} else {
-					func_sig.return_type
-				}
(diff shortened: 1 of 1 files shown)


MarkMcCulloh (commented):
Needs a test, otherwise looks good

hasanaburayyan:
> Needs a test, otherwise looks good

@MarkMcCulloh I cant think of a good way to add a test without code from https://github.com/winglang/wing/pull/1285 OR just adding a dummy interface to core lib in the sdk. 

Actually this introduces a problem with also our container types. I think @staycoolcall911 is facing this potentially as well. If we define an interface in the SDK like:

```ts 
export interface Stuff {
	x: ImmutableArray,
	y: Json
}
```

and we have some wingcode like
```js
bring cloud;

inflight (msg:str): cloud.Stuff {
	return cloud.Stuff {
		x: ["a", "b"],
		y: Json "Cool"
	};
};
```

Today we will get:
Expected type to be "ImmutableArray", but got "Array<str>" instead
Expected type to be "Json", but got "Json" instead

My PR here only fixes "Json" to "Json" mappings. 

So basically I think this is a temp fix for a problem that needs a more holistic approached solution. @Chriscbr @MarkMcCulloh maybe we can take some time to huddle and chat about this tomorrow? 

In the interim I know this unblocks https://github.com/winglang/wing/pull/1285 so Im happy to let this code be included in that PR instead whatever we would prefer.

hasanaburayyan:
Closing this PR in favor for better complete solution: https://github.com/winglang/wing/pull/1741

PR #1731, opened 2023-03-04
chore: safely assume snapshot tests don't produce colored output (see #1723)

yoav-steinberg:
@MarkMcCulloh please take a look and either approve or explain why not. Until then marking as `do-not-merge`.


MarkMcCulloh (approved):
Yeah this should be ok (assuming the tests pass here). This was originally added when we did have colors in tests. The custom serializer was pretty nice.

But then I think there were issues with cross-platform snapshots not matching with color, so they were disabled, but the special serializer was never removed.

PR #1732, opened 2023-03-05
chore: update rust toolchain and several misc changes

### LINKED ISSUE #813: Run rustfmt in workflow
### Summary

We should enforce rustfmt runs for consistent formatting

### Feature Spec

_No response_

### Use Cases

_No response_

### Implementation Notes

_No response_

### Component

Contributor Experience

MarkMcCulloh:
Misc changes while working on other stuff

- Update rust to 1.67.1 and made change to .cargo/config.toml to get it to work
- Changed hangar so that it stops spitting out `undefined` when the cli errors
- Updated wasi-sdk to the latest version, and changed the install scripts that it installs a tool within the repo itself (gitignored)
  - This avoids using sudo in the install
  - It is also now a versioned dir, this should make it easier to make sure everyone stays in sync whenever it changes
- Added WingPoint, and made better use of `Into`, `From`, and `Default` so that and WingSpan for some hopefully more idiomatic rust experiences
- Added "lint" as a cached nx operation. This should be okay since it relies on changes to src files
- Made sure clippy runs with release and a couple of other flags when needed, to avoid unnecessary compilations in CI
- Run rustfmt in "lint"
- Add "--fix" arg to clippy

Fixes #813

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,4 @@
 [env]
-CC_wasm32_wasi = "/opt/wasi-sdk/bin/clang"
-AR_wasm32_wasi = "/opt/wasi-sdk/bin/ar"
+WASI_SDK = { value = ".cargo/wasi-sdk-19.0", relative = true }
+CC_wasm32_wasi = { value = ".cargo/wasi-sdk-19.0/bin/clang", relative = true }
+AR_wasm32_wasi = { value = ".cargo/wasi-sdk-19.0/bin/ar", relative = true }
--- a/docs/06-contributors/020-development.md
+++ b/docs/06-contributors/020-development.md
@@ -105,7 +105,7 @@ separately).
 
 :::note
 
-The first time you run `npm install` you may be asked to enter your system password, this is because
+The first time you run `npm install` it may take a long time, this is because
 it's taking care of installing the [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) for you.
 
 If you wish to install it manually, you may do so by running `scripts/setup_wasi.sh`
--- a/docs/06-contributors/020-development.md
+++ b/docs/06-contributors/020-development.md
@@ -105,8 +105,8 @@ separately).
 
 :::note
 
-The first time you run `npm install` it may take a long time, this is because
-it's taking care of installing the [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) for you.
+The first time you run `npm install` it may take extra time to install the
+ [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) for you. This is needed to compile Wing for WASM.
 
 If you wish to install it manually, you may do so by running `scripts/setup_wasi.sh`
 
--- a/libs/tree-sitter-wing/bindings/rust/build.rs
+++ b/libs/tree-sitter-wing/bindings/rust/build.rs
@@ -8,48 +8,26 @@ fn main() {
 		.expect("Generating parser");
 
 	let mut c_config = cc::Build::new();
-	c_config.include(src_dir);
 	c_config
+		.include(src_dir)
+		.flag("-Wno-unused-parameter")
+		.flag("-Wno-unused-but-set-variable")
+		.flag("-Wno-trigraphs")
+		.file(&src_dir.join("parser.c"));
+
+	if cfg!(target_arch = "wasm32") {
 		// This parser is used in a WASI context, so it needs to be compiled with
 		// the sysroot and other tools needed for WASI-compatible C
-		.flag(
+		c_config.flag(
 			format!(
 				"--sysroot={}/share/wasi-sysroot",
 				env!("WASI_SDK", "WASI_SDK env not set")
 			)
 			.as_str(),
-		)
-		.flag("-Wno-unused-parameter")
-		.flag("-Wno-unused-but-set-variable")
-		.flag("-Wno-trigraphs");
-	let parser_path = src_dir.join("parser.c");
-	c_config.file(&parser_path);
-
-	// If your language uses an external scanner written in C,
-	// then include this block of code:
-
-	/*
-	let scanner_path = src_dir.join("scanner.c");
-	c_config.file(&scanner_path);
-	println!("cargo:rerun-if-changed={}", scanner_path.to_str().unwrap());
-	*/
+		);
+	}
 
 	c_config.compile("parser");
+
 	println!("cargo:rerun-if-changed=grammar.js");
-
-	// If your language uses an external scanner written in C++,
-	// then include this block of code:
-
-	/*
-	let mut cpp_config = cc::Build::new();
-	cpp_config.cpp(true);
-	cpp_config.include(&src_dir);
-	cpp_config
-			.flag_if_supported("-Wno-unused-parameter")
-			.flag_if_supported("-Wno-unused-but-set-variable");
-	let scanner_path = src_dir.join("scanner.cc");
-	cpp_config.file(&scanner_path);
-	cpp_config.compile("scanner");
-	println!("cargo:rerun-if-changed={}", scanner_path.to_str().unwrap());
-	*/
 }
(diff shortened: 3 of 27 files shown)


.cargo/config.toml
```diff
@@ -1,3 +1,4 @@
 [env]
-CC_wasm32_wasi = "/opt/wasi-sdk/bin/clang"
-AR_wasm32_wasi = "/opt/wasi-sdk/bin/ar"
+WASI_SDK = { value = ".cargo/wasi-sdk-19.0", relative = true }
```
MarkMcCulloh, line 2: ℹ️ This was the fix needed for the rust update.
This environment variable helps the linker find the proper `sysroot` when handling C dependencies. Not entirely sure why it worked before.

docs/06-contributors/020-development.md
```diff
@@ -105,7 +105,7 @@ separately).
 
 :::note
 
-The first time you run `npm install` you may be asked to enter your system password, this is because
+The first time you run `npm install` it may take a long time, this is because
```
eladb, line 108: How long?

libs/wingc/src/diagnostic.rs
```diff
@@ -1,65 +1,107 @@
...
+/// tree-sitter-based Point => WingPoint
+impl From<Point> for WingPoint {
+	fn from(point: Point) -> Self {
 		Self {
-			start: Point { row: 0, column: 0 },
-			end: Point { row: 0, column: 0 },
-			start_byte: 0,
-			end_byte: 0,
-			file_id: String::from(""),
+			line: point.row as u32,
```
eladb, line 24: Elegant!

libs/wingc/src/diagnostic.rs
```diff
@@ -71,7 +113,13 @@ impl WingSpan {
...
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		write!(f, "{}:{}:{}", self.file_id, self.start.row + 1, self.start.column + 1)
+		write!(
+			f,
+			"{}:{}:{}",
+			self.file_id,
+			self.start.line + 1,
+			self.start.character + 1
+		)
 	}
```
eladb, line 123: Maybe add a few tests for this?

tools/hangar/src/invalid.test.ts
```diff
@@ -19,7 +19,7 @@ describe.each(invalidWingFiles)("%s", (wingFile) => {
 
       const out = await runWingCommand(testDir, relativeWingFile, args, false);
 
-      const stderr = out.all ?? out.stderr;
+      const stderr = out.stderr;
```
eladb, line 22: Why not "all"?

tools/hangar/src/test.test.ts
```diff
@@ -19,7 +19,7 @@ describe.each(validWingFiles)("%s", (wingFile) => {
         true
       );
 
-      expect(out.all).toMatchSnapshot("stdout");
+      expect(out.stdout).toMatchSnapshot("stdout");
```
eladb, line 22: Why?

.cargo/config.toml
MarkMcCulloh: Ok I guess this isn't enough, I'll actually have to set the cflags here. Should be good after that though

docs/06-contributors/020-development.md
MarkMcCulloh: Depends on network conditions, it took a minute in my hotel room

tools/hangar/src/invalid.test.ts
MarkMcCulloh: It wasn't really behaving as expected. It would often be empty even though there was definitely something in stderr at least. Maybe it didn't like when stdout was empty.

libs/wingc/src/diagnostic.rs
```diff
@@ -1,65 +1,107 @@
...
-#[derive(Debug, PartialEq, Eq, Hash, Clone)]
-pub struct WingSpan {
-	pub start: CharacterLocation,
-	pub end: CharacterLocation,
-	pub start_byte: ByteIndex,
-	pub end_byte: ByteIndex,
-	pub file_id: FileId,
+/// Line and character location in a UTF8 Wing source file
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct WingPoint {
```
Chriscbr, line 15: WDYT of a name like `Location` or `Position`?

libs/wingc/src/diagnostic.rs
```diff
@@ -1,65 +1,107 @@
...
+	pub line: u32,
+	pub character: u32,
```
Chriscbr, line 17: I would rename to `col` or `column`

libs/wingc/src/diagnostic.rs
```diff
@@ -1,65 +1,107 @@
...
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		write!(f, "{}:{}", self.line, self.character)
+	}
+}
+
+#[derive(Debug, Default, PartialEq, Eq, Hash, Clone)]
+pub struct WingSpan {
+	pub start: WingPoint,
+	pub end: WingPoint,
+	pub file_id: String,
```
Chriscbr, line 70: What is file_id? Is that the path, or something else? (maybe we can add a docstring)

libs/wingc/src/type_check.rs
```diff
@@ -764,7 +764,7 @@ impl<'a> TypeChecker<'a> {
 			vec![WINGSDK_STD_MODULE.to_string()],
 			&Symbol {
 				name: WINGSDK_STD_MODULE.to_string(),
-				span: WingSpan::global(),
+				span: Default::default(),
 			},
```
Chriscbr, line 768: (optional) For these cases we could consider adding a helper method perhaps:
```rust
Symbol::global(WINGSDK_STD_MODULE)
```

implemented as something like
```rust
fn global<S: Into<String>>(name: S) -> Self {
	Self { name: name.into(), span: Default::default() }
}
```

libs/tree-sitter-wing/bindings/rust/build.rs
```diff
@@ -7,12 +7,16 @@ fn main() {
 	generate_parser_in_directory(&PathBuf::from("."), None, tree_sitter::LANGUAGE_VERSION, false, None)
 		.expect("Generating parser");
 
+	// get WASI_SDK env
+	let wasi_sdk = std::env::var("WASI_SDK").expect("WASI_SDK env not set");
```
Chriscbr, line 11: Do we know this value at compile time? Maybe we can compile it inline https://doc.rust-lang.org/std/macro.env.html

Chriscbr (approved):

libs/wingc/src/diagnostic.rs
MarkMcCulloh: I felt a similar dilemma with `Span`, where the type is used in several places. tree-sitter has one, lsp-types has one, and I expect us to run into another one whenever you use a new crate for beautiful diagnostics.

Normally I wouldn't name types with Wing___, but this one felt helpful to avoid ambiguity (or being forced to put namespaces)

libs/tree-sitter-wing/bindings/rust/build.rs
MarkMcCulloh: Nice, wasn't 100% sure it'd work due to this being in a build.rs but it looks like it's working

libs/wingc/src/type_check.rs
MarkMcCulloh: Noice, yeah this looks good I went ahead and added it

.cargo/config.toml
MarkMcCulloh: Ok there was also an issue with the allocator function we expose where it did not properly handle mem alignment

Issue #114, opened 2022-09-20
Break, Continue Statements


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#22-break
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#23-continue

staycoolcall911:
This was closed by @flyingImer in #1734 and #1770

PR #1734, opened 2023-03-06
feat(compiler): add break statement support

flyingImer:
Implements Break statement in https://github.com/winglang/wing/issues/114


libs/wingc/src/parser.rs
```diff
@@ -252,6 +253,17 @@ impl Parser<'_> {
 		})
 	}
 
+	fn build_break_statement(&self, statement_node: &Node) -> DiagnosticResult<StmtKind> {
+		let mut cursor = statement_node.walk();
+		while let Some(parent) = cursor.node().parent() {
```
flyingImer, line 258: for some reason, `cursor.goto_parent()` is always `false`, walking around with this ugly approach. LMK what did I miss

flyingImer: After playing around with this `goto_parent` API, it works only for the cursor previously went through, e.g., in this use case, it returns `true` IIF we do `cursor.goto_first_child()` first. In other word, it considers the `walk()` TreeNode as the root of the tree. However, I cannot find any doc about this behavior, LMK

staycoolcall911:
Looks great @flyingImer  - I asked @yoav-steinberg to take a look.

examples/tests/valid/while.w
```diff
@@ -6,3 +6,15 @@ let y = 123;
...
 }
+
+let var z = 0;
+while true {
+  z = z + 1;
+  if (z > 2) {
+    break;
+  }
+}
+
```
eladb, line 17:
```suggestion

assert(z == 3);
```

libs/wingc/src/type_check.rs
```diff
@@ -1542,6 +1542,14 @@ impl<'a> TypeChecker<'a> {
 
 				self.inner_scopes.push(statements);
 			}
+			StmtKind::Break { is_in_loop } => {
+				if !*is_in_loop {
```
yoav-steinberg, line 1546: I think this is redundant if we manage to handle the error reporting in _parser.rs_.

yoav-steinberg: We should probably also add an issue for reporting dead code:
```
while true {
  break;
  break; // Should fail on "unreachable code" (out of the scope of this PR)
}
```

flyingImer:
> This, or even if goto_parent would have worked as expected isn't that efficient: each time we encounter a break we'll need to run up the tree. Then if we encounter another break we'll need to do it again. Ideally we'll just have a in_loop: bool in our state (self).
Can you try an approach where you add some mutable state (in_loop) to the Parser class and set it to true when entering a while or for scope. To avoid the complexity of making self mutable lets just make ti a RefCell for now.

This makes a lot of sense, let me take a look

libs/wingc/src/type_check.rs
flyingImer: good callout on the dead code! wonder should it fail or just a warning? (I can create a follow-up issue on it) I'm inclined to the latter one - dead code does not hurt anything but it's "dead"

libs/wingc/src/parser.rs
flyingImer: This makes a lot of sense, let me take a look

libs/wingc/src/parser.rs
```diff
@@ -252,6 +253,17 @@ impl Parser<'_> {
...
+			if parent.kind() == "while_statement" || parent.kind() == "for_in_loop" {
+				return Ok(StmtKind::Break { is_in_loop: true });
+			}
+			cursor.reset(parent);
+		}
+		return Ok(StmtKind::Break { is_in_loop: false });
```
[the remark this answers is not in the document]
yoav-steinberg: Currently the rule of thumb is to report stuff early on if possible. The actual grammar parsing is done by tree-sitter anyway. _parser.rs_ converts tree-sitter's AST to a strongly typed AST and when doing so does some error checking that the grammar couldn't do on itself (like enum variant uniqueness). Anything we can do without symbol environments and typying I think we should do here.

libs/wingc/src/type_check.rs
yoav-steinberg: +1 for an issue on dead-code detection.
This is probably only one of a million other cases we can benefit from this.
We currently don't have warnings in wing, we're trying to see how long this approach will hold water..

flyingImer: dead-code detection tracking: https://github.com/winglang/wing/issues/1767

libs/wingc/src/ast.rs
```diff
@@ -252,9 +252,7 @@ pub enum StmtKind {
 		elif_statements: Vec<ElifBlock>,
 		else_statements: Option<Scope>,
 	},
-	Break {
-		is_in_loop: bool,
-	},
+	Break(),
```
yoav-steinberg, line 255:
```suggestion
	Break,
```

yoav-steinberg (approved):
excellent!

PR #1735, opened 2023-03-06
feat: import interfaces - attempt 2

Chriscbr:
Adding the interface importing support from #1684, this time without some of the extra changes made to SDK and the compiler's assembly loading logic. (#1684 had to be reverted because the changes made to the SDK caused the playground to stop working, and caused the getting started example to stop working in some environments https://winglang.slack.com/archives/C04BBDQUWQP/p1677784725607179).


libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -173,6 +174,14 @@ impl<'a> JsiiImporter<'a> {
 
 		self.setup_namespaces_for(&type_fqn);
 
+		// Hack: if the type is "constructs.IConstruct", we import it manually
+		// this is done so we can avoid loading the constructs module
+		if type_fqn.as_str() == CONSTRUCT_BASE_INTERFACE {
+			let symbol = Symbol::global(type_fqn.type_name());
+			self.register_jsii_type(&type_fqn, &symbol, self.wing_types.anything());
+			return;
+		}
```
Chriscbr, line 183: The hack we are using to avoid loading `constructs.Construct` didn't work out of the box because we have functions in the SDK with parameters accepting `constructs.IConstruct` directly i.e. the type is not just used in `implements`/`extends` clauses.

It _would_ work if I changed all of instances of `IConstruct` in the SDK to `IResource` instead, but that felt like it was worth avoiding.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -194,10 +203,7 @@ impl<'a> JsiiImporter<'a> {
 			return;
 		}
 
-		debug!(
-			"Type {} is unsupported or not found in the type system, skipping",
-			type_fqn
-		);
+		panic!("Type {} was not found in the type system, skipping", type_fqn);
```
hasanaburayyan, line 206: No longer skipping its blowing up 😁 
```suggestion
		panic!("Type {} was not found in the type system", type_fqn);
```

hasanaburayyan (commented):
LGTM Ill let someone else give the final say

staycoolcall911 (approved):
LGTM as well

MarkMcCulloh (approved):

Issue #611, opened 2022-11-15
external.Redis - Simulator implementation


Chriscbr:

staycoolcall911:
Keep

Chriscbr:
Potential resource - https://www.npmjs.com/package/redis-mock

yoav-steinberg:
done in #1742

PR #1742, opened 2023-03-07
feat(sdk): redis sim support

yoav-steinberg:
See #611

This adds a `Rdis` resource under the `redis` namespace in the sdk. Currently only implemented for sim.
```wing
bring cloud;
bring redis;

let r = new redis.Redis();
let r2 = new redis.Redis() as "r2";

new cloud.Function(inflight (s:str):str => {
   print("${r.url()}");
   print("${r2.url()}");
   let connection = r.ioredis();
   let connection2 = r.ioredis();
   connection.set("wing", "does redis");
   let value = connection.get("wing");
   print("${value}");
   connection2.set("wing", "does redis again");
   let value2 = connection2.get("wing");
   print("${value2}");
}) as "test";
```


libs/wingsdk/test/target-sim/__snapshots__/queue.test.ts.snap
```diff
@@ -127,9 +127,6 @@ exports[`messages are requeued if the function fails 1`] = `
   "Sending messages (messages=[\\"BAD MESSAGE\\"], subscriber=sim-1).",
   "Invoke (payload=\\"\\"{\\\\\\"messages\\\\\\":[\\\\\\"BAD MESSAGE\\\\\\"]}\\"\\").",
   "Subscriber error - returning 1 messages to queue.",
-  "Sending messages (messages=[\\"BAD MESSAGE\\"], subscriber=sim-1).",
```
staycoolcall911, line 130: Why was that deleted?

yoav-steinberg: don't know, fixed in later commit.

yoav-steinberg:
@MarkMcCulloh hi, this is currently failing the redis test in the sdk because it looks like the containers started by the redis-sim aren't accessibly from the github actions host (notice the `ECONNREFUSED`):
image

It seems like just running a docker image (which is what the redis simulator does) isn't good here because there's networking access between the host and the containers. I think I can't [the official](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) way to start the container (which handles network mapping) because we need to start the containers programmatically for each instantiated resource.
Any suggestions?

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  public async init(): Promise<void> {
+    this.docker = new Dockerode();
+
+    // Create a redis container
+    try {
+      let container = await this.docker.createContainer({
+        Image: "redis",
+        name: this.container_name,
+      });
```
MarkMcCulloh, line 38:
```suggestion
      let container = await this.docker.createContainer({
        Image: "redis",
        name: this.container_name,
        ExposedPorts: { "6379/tcp": {} },
      });
```

This is my guess of what could help make sure the container is accessible.

hasanaburayyan: hmm it looks like if you are using a github action environment for your actions then any started containers need to be accessed using `localhost:<port>` or `127.0.0.1:<port>` but when your actions are run within a container they have access to the user defined docker bridge network.

yoav-steinberg: I don't think explicitly exposing the ports will help, 6379/tcp is already exposed in the dockerfile of the redis image so I doubt it'll make a change. Nonetheless I'll give it a try.

> when your actions are run within a container they have access to the user defined docker bridge network.

Any chance we can change our actions to be run in a container to check this theory?

yoav-steinberg: > I don't think explicitly exposing the ports will help, 6379/tcp is already exposed in the dockerfile of the redis image so I doubt it'll make a change. Nonetheless I'll give it a try.

Tried, same issue, reverted.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,79 @@
+import Dockerode from "dockerode";
+import { ISimulatorResourceInstance } from "./resource";
+import { RedisSchema } from "./schema-resources";
+import { IRedisClient } from "../redis";
+import { ISimulatorContext } from "../testing/simulator";
+
+// Issue using types from ioredis with JSII
+//import { Redis as IoRedis } from "ioredis";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const IoRedis = require("ioredis");
```
Chriscbr, line 10: Yeah, this is a tricky issue to deal with - I haven't been able to come up with too many great solutions for this. 😞  I just encountered it again, trying to add the `express` dependency to the simulator for use in the `Api` resource.

If you're feeling ambitious, one other workaround you can try is to install [@types/ioredis](https://www.npmjs.com/package/@types/ioredis?activeTab=versions) with an older version of the type definitions (tagged with TS 3.7 or something) into wingsdk and see if JSII will use that instead of the newer types that it can't understand. Or even to use an older version of ioredis, like the 4.x version.

libs/wingsdk/src/target-sim/redis.inflight.ts
hasanaburayyan: > Any chance we can change our actions to be run in a container to check this theory?

It doesn't look too bad to try. Im not GHA expert, it looks simple enough. Ill test in separate junk PR.

yoav-steinberg: :pray: thanks.

hasanaburayyan: shame on me for thinking itd be so simple, keep running into tiny hitches with permissions and such but making progress, shooting to get it working this evening.

hasanaburayyan: @yoav-steinberg @MarkMcCulloh I finally built a docker container with everything needed to get to the build and tests going. Im not sure if the results are promising or not, I am failing for a separate reason.

https://github.com/winglang/wing/actions/runs/4452587048/jobs/7820349387?pr=1842

image

Gonna look into it more this weekend

ekeren:
@yoav-steinberg , I've notice you chose a different API pattern then the one that is used in our other resource.
We don't do `bucket.get_connection().put(f, s);` we do `bucket.put(f,s)`


consider these two api:
```ts (wing) 
bring cloud;
bring redis;

let redis = new redis.Redis();

// This Proposal 
new cloud.Function(inflight (s:str):str => {
   let connection = redis.ioredis();
   connection.set("wing", "does redis");
   let value = connection.get("wing");
   print("${value}");
}) as "test:1";


// cloud.* style
new cloud.Function(inflight (s:str):str => {
   redis.set("wing", "does redis");
   let value = redis.get("wing");
   print("${value}");
}) as "test:2";
```

Is this because of a technical issue or a deliberate design decision

libs/wingsdk/src/redis/redis.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Inflight interface for `Redis`.
+ */
+export interface IRedisClient {
+  /**
+   * Get underlying redis client.
+   * @inflight
+   */
+  ioredis(): Promise<any>;
```
ekeren, line 42: @yoav-steinberg , I've notice you chose a different API pattern then the one that is used in our other resource.
We don't do `bucket.get_connection().put(f, s);` we do `bucket.put(f,s)`


consider these two api:
```ts (wing) 
bring cloud;
bring redis;

let redis = new redis.Redis();

// This Proposal 
new cloud.Function(inflight (s:str):str => {
   let connection = redis.ioredis();
   connection.set("wing", "does redis");
   let value = connection.get("wing");
   print("${value}");
}) as "test:1";


// cloud.* style
new cloud.Function(inflight (s:str):str => {
   redis.set("wing", "does redis");
   let value = redis.get("wing");
   print("${value}");
}) as "test:2";
```

Is this because of a technical issue or a deliberate design decision?

ekeren, line 42: @yoav-steinberg , I've notice you chose a different API pattern then the one that is used in our other resource.
We don't do bucket.get_connection().put(f, s); we do bucket.put(f,s)

consider these two api:

bring cloud;
bring redis;

let redis = new redis.Redis();

// This Proposal 
new cloud.Function(inflight (s:str):str => {
   let connection = redis.ioredis();
   connection.set("wing", "does redis");
   let value = connection.get("wing");
   print("${value}");
}) as "test:1";


// cloud.* style
new cloud.Function(inflight (s:str):str => {
   redis.set("wing", "does redis");
   let value = redis.get("wing");
   print("${value}");
}) as "test:2";
Is this because of a technical issue or a deliberate design decision?

libs/wingc/src/type_check.rs
```diff
@@ -1783,7 +1783,7 @@ impl<'a> TypeChecker<'a> {
 						// we use the module name as the identifier.
 						// For example, `bring fs` will import the `fs` namespace from @winglang/sdk and assign it
 						// to an identifier named `fs`.
-						WINGSDK_CLOUD_MODULE | WINGSDK_FS_MODULE => {
+						WINGSDK_CLOUD_MODULE | WINGSDK_FS_MODULE | WINGSDK_REDIS_MODULE => {
```
eladb, line 1786: Can we move this to some kind of list or a utility function at this point?

libs/wingsdk/src/redis/redis.ts
```diff
@@ -0,0 +1,49 @@
...
+export abstract class Redis extends Resource {
+  /**
+   * Create a new redis.
+   * @internal
+   */
+  public static _newRedis(scope: Construct, id: string): Redis {
+    return App.of(scope).newAbstract(REDIS_FQN, scope, id);
+  }
+
+  public readonly stateful = false; // TODO: redis persistence
```
eladb, line 24: why isn't this `true`? A redis cluster is a persistent store, no?

libs/wingsdk/src/redis/redis.ts
eladb, line 42: - Do we really need to specify this is ioredis?
- Also, I would use some semantic indication that this is "raw" or something like that.

```suggestion
  raw_client(): Promise<any>;
```

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,79 @@
...
+  private connection_url?: string = undefined;
+  private docker?: Dockerode = undefined;
+  private connection?: any;
+
+  public constructor(_props: RedisSchema["props"], context: ISimulatorContext) {
+    this.context = context;
+    this.container_name = `wing-sim-redis-${this.context.resourcePath.replace(
+      "/",
+      "."
+    )}-${Redis.uid}`;
```
eladb, line 26: Perhaps use a `uuid()`. Generally trying to avoid static mutable state.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,79 @@
...
+      let container = await this.docker.createContainer({
+        Image: "redis",
+        name: this.container_name,
+      });
+
+      // Start the redis container
+      await container.start();
+
+      // Generate the redis url based on the container ip address
+      let container_spec = await container.inspect();
```
eladb, line 44:
```suggestion
      const container_spec = await container.inspect();
```

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,79 @@
...
+    )}-${Redis.uid}`;
+    Redis.uid++;
+  }
+
+  public async init(): Promise<void> {
+    this.docker = new Dockerode();
+
+    // Create a redis container
+    try {
+      let container = await this.docker.createContainer({
```
eladb, line 35: Any reason to use `let`?

```suggestion
      const container = await this.docker.createContainer({
```

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public async ioredis(): Promise<any> {
+    if (this.connection != undefined) {
+      return this.connection;
+    } else if (this.connection_url != undefined) {
+      return (this.connection = new IoRedis(this.connection_url));
+    } else {
+      throw new Error("Redis server not initialized");
+    }
```
eladb, line 69: Style: 
- prefer early bailout, `else` is not needed if you `return`
- never use `!=` or `==`, only `===` or `!==`, and for undefined checks, just `if (x)`

```suggestion
    if (this.connection) {
      return this.connection;
    }
    
    if (this.connection_url) {
      this.connection = new IoRedis(this.connection_url);
      return this.connection;
    }
    
    throw new Error("Redis server not initialized");
```

eladb (approved):
Very cool

libs/wingsdk/src/redis/redis.ts
yoav-steinberg: It can be either depends on how you configure your redis. I think by default ElastiCache and RedisLabs disable persistence (the name ElasitCahce kind of hints this).

libs/wingsdk/src/redis/redis.ts
yoav-steinberg: I like this. Lets go with `raw_client` although we'll need to document that the current "raw" client is an `ioredis` client and not some other client. Makes sense?

libs/wingsdk/src/target-sim/redis.inflight.ts
yoav-steinberg: This is just a sim hack and I didn't want to add more dependencies here and also make the docker container name very long. Make sense?

libs/wingsdk/src/target-sim/redis.inflight.ts
yoav-steinberg: I'm learning t/js :)

libs/wingsdk/src/target-sim/redis.inflight.ts
hasanaburayyan: So after an idea to use port mappings and local hosts I got this working (not in contianer)
https://github.com/winglang/wing/actions/runs/4491143694/jobs/7899192318

libs/wingsdk/src/target-sim/redis.inflight.ts
yoav-steinberg: @hasanaburayyan maybe check this static field and just use the generated port number in the name that that we've got that.

hasanaburayyan: Yupp, Ill remove the uid entirely using the port instead

hasanaburayyan: I ran into issues with portfinder, I opted for a different solution using port `0` wildcard and then inspecting running container to retrieve port (for url). This means I cannot provide the port number in the container name. So I used `uuid()`

libs/wingsdk/src/redis/redis.ts
hasanaburayyan: renamed

MarkMcCulloh:
The contributor docs should be updated to mention that docker is required now to run tests

libs/wingc/src/jsify.rs
```diff
@@ -378,9 +378,8 @@ impl<'a> JSifier<'a> {
 			ExprKind::Reference(_ref) => self.jsify_reference(&_ref, None, context),
 			ExprKind::Call { function, arg_list } => {
 				let function_type = function.evaluated_type.borrow().unwrap();
-				let function_sig = function_type
-					.as_function_sig()
-					.expect("Expected expression to be callable");
+				let function_sig = function_type.as_function_sig();
+				assert!(function_sig.is_some() || function_type.is_anything());
```
Chriscbr, line 382: I think it might be possible to include a message with `assert`
```suggestion
				assert!(function_sig.is_some() || function_type.is_anything(), "Expected expression to be callable");
```

libs/wingsdk/src/redis/redis.ts
Chriscbr: TIL

libs/wingsdk/test/target-sim/redis.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  // THEN
+  const s = await app.startSimulator();
+  try {
+    const client = s.getResource("/my_redis") as redis.IRedisClient;
+    const value = await client.smembers(key);
+    expect(value).toEqual([]);
+  } finally {
+    await s.stop();
+  }
```
Chriscbr, line 145: (random idea) We might be able to reduce boilerplate with something like this
```suggestion
  await app.withSimulator(async (s) => {
    const client = s.getResource("/my_redis") as redis.IRedisClient;
    const value = await client.smembers(key);
    expect(value).toEqual([]);
  });
```

libs/wingsdk/test/target-sim/redis.test.ts
```diff
@@ -0,0 +1,147 @@
...
+  try {
+    const client = s.getResource("/my_redis") as redis.IRedisClient;
+    expect((await client.url()).startsWith("redis://")).toBeTruthy();
+    const redisClient = (await client.rawClient()) as IoRedis;
+    await redisClient.set("foo", "bar");
+    expect(await redisClient.get("foo")).toEqual("bar");
+  } finally {
+    await s.stop();
+  }
+  expect(app.snapshot()).toMatchSnapshot();
```
Chriscbr, line 44: If the preflight config is the same each time we can probably skip all of these snapshots after the first test

libs/wingsdk/src/target-sim/redis.ts
```diff
@@ -0,0 +1,51 @@
...
+    bindSimulatorResource("redis", this, host);
+    super._bind(host, ops);
+  }
+
+  /** @internal */
+  public _toInflight(): core.Code {
+    return makeSimulatorJsClient("redis", this);
+  }
+
+  // TODO: kill container on sim shutdown
```
Chriscbr, line 40: Looks like this was handled in the inflight client?

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,131 @@
...
+        "6379/tcp"
+      ][0].HostPort;
+
+      // redis url based on host port
+      this.connection_url = `redis://0.0.0.0:${hostPort!}`;
+      return {};
+    } catch (e) {
+      throw Error(`Error setting up Redis resource simulation (${e})
+      - Make sure you have docker installed and running
+      - Make sure you have the redis image installed ('docker pull redis')`);
```
Chriscbr, line 60: Love this error message!

Chriscbr (approved):
Looks great!

libs/wingsdk/src/target-sim/redis.inflight.ts
MarkMcCulloh, line 60: Why not pull the image as part of execution of the resource?
Since we have access to the docker API, I don't think it's super useful to ask the user to do it.

Additionally, the image should be pinned to a very specific hash. This improves cacheability and reduces possible supply chain issues. I don't think people need the flexibility of choosing their own redis image in the sim.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -76,7 +70,7 @@ export class Redis implements IRedisClient, ISimulatorResourceInstance {
     await container?.remove();
   }
 
-  public async ioredis(): Promise<any> {
+  public async rawClient(): Promise<any> {
```
yoav-steinberg, line 73: Seem to me all methods from here downwards need to be implemented in some abstract base class that all redis implementations (be they sim/tf-x/.. need to inherit from).
Perhaps just add a comment here and later when we add ElastiCache, for example, create that abstract class.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -96,4 +90,42 @@ export class Redis implements IRedisClient, ISimulatorResourceInstance {
       throw new Error("Redis server not initialized");
     }
   }
+
+  public async set(key: string, value: string): Promise<void> {
+    let redis = await this.rawClient();
+    await redis.set(key, value);
+  }
+
+  public async get(key: string): Promise<string> {
```
yoav-steinberg, line 99: I'd say that the (wing) return value here needs to be a `str?`, I think this translates to a JSII `Promise<string | null>`.
(to handle the case of getting a non-existent key)

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -96,4 +90,42 @@ export class Redis implements IRedisClient, ISimulatorResourceInstance {
...
+    let redis = await this.rawClient();
+    let result = await redis.get(key);
+    return result ?? "";
+  }
+
+  public async hset(key: string, field: string, value: string): Promise<void> {
```
yoav-steinberg, line 105: note that the (perhaps unexpected) return type here needs to be a wing `num` representing the number of new fields added to the hash as a result of this command.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -96,4 +90,42 @@ export class Redis implements IRedisClient, ISimulatorResourceInstance {
...
+    await redis.hset(key, field, value);
+  }
+
+  public async hget(key: string, field: string): Promise<string> {
+    let redis = await this.rawClient();
+    let result = await redis.hget(key, field);
+    return result ?? "";
+  }
+
+  public async sadd(key: string, value: string): Promise<void> {
```
yoav-steinberg, line 116: see `hset` return type. same here.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -96,4 +90,42 @@ export class Redis implements IRedisClient, ISimulatorResourceInstance {
...
+    await redis.sadd(key, value);
+  }
+
+  public async smembers(key: string): Promise<string[]> {
+    let redis = await this.rawClient();
+    let result = await redis.smembers(key);
+    return result ?? [];
+  }
+
+  public async del(key: string): Promise<void> {
```
yoav-steinberg, line 127: return a wing `num`, the number of keys that were deleted.

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -25,13 +26,17 @@ export class Redis implements IRedisClient, ISimulatorResourceInstance {
...
     )}-${uuidv4()}`;
   }
-
+  
   public async init(): Promise<RedisAttributes> {
     this.docker = new Dockerode();
+
+    // Pull docker image
+    const stream = await this.docker.pull(this.REDIS_IMAGE);
+    await new Promise(res => this.docker?.modem.followProgress(stream, res));
```
yoav-steinberg, line 35: should these go into the `try`? I guess we also need to fix the error message to only suggest installing docker since we're taking care of the image.

libs/wingsdk/src/target-sim/redis.inflight.ts
hasanaburayyan: hmm, im not sure how we pass back an optional from the SDK. @Chriscbr do we do this somewhere else? JSII doesnt seem to be happy with me when I try to do <string | null>

yoav-steinberg: I think JSII has some equivalent of an optional value, the jsii_importer has code to support this, see `optional_type_to_wing_type`.

yoav-steinberg: Maybe a `Promise<string> | Promise<undefined>`?

hasanaburayyan: Was able to make this work by patching JSii basically copied what @Chriscbr did for his `try_get` apis on bucket (it just had not finished PR) https://github.com/winglang/wing/pull/1622

staycoolcall911:
Congrats @yoav-steinberg and @hasanaburayyan (and @Chriscbr and @MarkMcCulloh for assisting), this is a major one!

libs/wingsdk/patches/jsii+1.74.0.patch
```diff
@@ -1,3 +1,19 @@
...
++++ b/node_modules/jsii/lib/assembler.js
+@@ -1557,9 +1557,8 @@ class Assembler {
+                 this._diagnostics.push(jsii_diagnostic_1.JsiiDiagnostic.JSII_1002_UNSPECIFIED_PROMISE.create(declaration));
+                 return { type: spec.CANONICAL_ANY };
+             }
+-            return {
+-                type: this._typeReference(typeRef.typeArguments[0], declaration, purpose),
+-            };
++            // (very very hacky) allow optionals to be returned from async methods
++            return this._optionalValue(typeRef.typeArguments[0], declaration, purpose);
```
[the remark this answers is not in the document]
Chriscbr: it's hacky I guess because I'm not 100% sure why jsii disallows this behavior / if this limitation is intentional or not. I can definitely open an issue in their repo to see if it can be upstreamed

Issue #76, opened 2022-09-08
Mechanism for explicitly qualifying lifting


yoav-steinberg:
## Workaround

Since it is possible to lift inflight closures (because a closure's "operation" is always "call me"), then, it is possible to work around the limitation we currently have by wrapping the operation in a closure and lifting this closure.

In the following example, `selectBucket` returns a `cloud.Bucket` and then the `.put()` operation cannot be qualified.

```js
bring cloud;

let b1 = new cloud.Bucket() as "b1";
let b2 = new cloud.Bucket() as "b2";

let selectBucket = inflight (i: num) => {
  if i % 2 == 0 {
    return b1;
  } else {
    return b2;
  }
};

new cloud.Function(inflight () => {
  let bucket = selectBucket(1);
  bucket.put("hello.txt", "world");
// ^^^^ can't qualify
});
```

To overcome this limitation, we can change `selectBucket` to actually perform the operation:

```js
bring cloud;

let b1 = new cloud.Bucket() as "b1";
let b2 = new cloud.Bucket() as "b2";

let putBucket = inflight (i: num, k: str, v: str) => {
  if i % 2 == 0 {
    b1.put(k, v);
  } else {
    b2.put(k, v);
  }
};

new cloud.Function(inflight () => {
  putBucket(1, "hello.txt", "world");
});
```

---

We need a syntax and implementation for the user to explicitly define rules describing how inflight code is going to use a resource. 

The rules can be passed to the resource's capture implementation. 

The most basic rule is obviously what inflight methods of the resource are being used. This can be used to set up permissions, for example. But we should also keep in mind that there might be information we want to pass describing possible ranges of arguments to these methods or regexes for valid argument values etc. 

Example:

```ts
let b = cloud.Bucket();
cloud.Function((event) ~> { 
   // We should have a way to explicitly allow only write access to `b` and only access to the `"filename.dat"` key in `b`.
  b.upload("filename.dat", event.data);
});
```

In the future we should also have a way to automatically generate these rules based on the code.

ekeren:
@staycoolcall911 , @eladb and @yoav-steinberg 

As discussed, lets complete the explicit permission syntax RFC for sprint 13

monadabot:
Congrats! :rocket: This was released in [Wing 0.61.17](https://github.com/winglang/wing/releases/tag/v0.61.17).

PR #1743, opened 2023-03-07
feat: identify invalid captures (and other improvements)

eladb:
Following up on #1682 to identify cases where we cannot infer which operations are performed on a captured resource.

See positive/negative tests for examples.

Rewrite the algorithm which analyzes the expressions captured by inflight methods so that it is able to identify more cases and emit errors when captures cannot be qualified (i.e. a resource is captured but we cannot determine which operations are performed on it without static analysis).

For each method, we identify all expressions that start with `this.xxx` and break them down into parts (using nested references). Then, we traverse the list of parts and split the expression into *preflight* and *inflight*. The preflight part is what we are capturing and the first inflight component qualifies which operations are performed on the captured object.

Reorganized capture tests into `resource_captures` (both under valid and invalid).

This does not address #76 but it explicitly identifies these cases. We will follow up at some point with a way to allow users to explicitly qualify the reference.


examples/tests/valid/inflight_ref_resource_sub_field.w
```diff
@@ -0,0 +1,48 @@
+bring cloud;
```
eladb, line 1: I've tested this also on AWS

libs/wingc/src/capture.rs
```diff
@@ -235,14 +235,6 @@ fn scan_captures_in_expression(
 							return res;
 						}
 					};
-
```
eladb, line 238: This is already covered in type-checking

libs/wingsdk/src/core/resource.ts
```diff
@@ -46,7 +46,7 @@ export interface IResource extends IInspectable, IConstruct {
    *
    * @internal
    */
-  _registerBind(host: IInflightHost, ops: string[]): void;
+  _registerBind(host: IInflightHost, ops: string[], context?: string[]): void;
```
Chriscbr, line 49: Where is this used?

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,147 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
 
-		self.path.clear();
+			return true;
+		} else {
+			// method call
+
+			// cannot call preflight methods (supposed to be covered during type checking)
+			assert!(variable.flight != Phase::Preflight);
+
+			// cannot call static methods
```
Chriscbr, line 1342: should this be a TODO? or did we decide inflight static methods aren't a thing for now?

Chriscbr (approved):
Looks good to me (added a few non-blocking questions)

libs/wingsdk/src/core/resource.ts
eladb: Oops. Leftover

libs/wingc/src/jsify.rs
eladb: I don't see a reason why this won't be a thing. I'll add a todo.

examples/tests/invalid/inflight_ref_explicit_ops.w
```diff
@@ -0,0 +1,60 @@
...
+    let x = this.b;
+//               ^ Cannot qualify which operations are performed on resource
+    x.put("hello", "world");
+    assert(this.just_str == "hello");
+    this.just_bucket.put("hello", "world");
+  }
+
+  inflight test2() {
+    let q = this.another.inflight_returns_resource();
+    q.push("push!");
```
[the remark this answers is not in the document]
eladb: Yes, that's a good point. But can likely wait until we actually support this scenario. At the moment, methods like `inflight_returns_resource` are illegal.

examples/tests/invalid/inflight_ref_explicit_ops.w
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  inflight test3() {
+    let b = this.array.at(1);
+    assert(b.list().length == 0);
```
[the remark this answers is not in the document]
eladb: It shouldn't pass. You are right!

examples/tests/invalid/inflight_ref_resource_sub_method.w
```diff
@@ -11,6 +11,7 @@ resource Another {
 
   inflight inflight_returns_resource(): cloud.Queue {
     return this.my_queue;
+//              ^^^^^^^^ Ambigious reference to "this.my_queue" when referenced by "inflight_returns_resource"
```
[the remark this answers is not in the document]
eladb: done

examples/tests/valid/inflight_ref_resource_collection.w
```diff
@@ -7,6 +7,7 @@ resource Test {
 
   init() {
     this.array = [new cloud.Bucket()];
+    assert(this.array.length == 1);
```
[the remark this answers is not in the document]
eladb: Regardless, I think this test is supposed to fail. Looking...

examples/tests/valid/inflight_ref_resource_sub_field.w
```diff
@@ -0,0 +1,48 @@
...
+  init() {
+    this.another = new Another();
+    this.my_str = "hello";
+    this.my_queue = new cloud.Queue();
+  }
+
+  inflight test() {
+    let arr = [1,2,3];
+    assert(arr.length == 3);
+    print("array.len=${arr.length}");
```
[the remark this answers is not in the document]
eladb: I don't mind keeping them. They are harmless. WDYT?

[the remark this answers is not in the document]
staycoolcall911: I agree - it's ok to keep them IMO.

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
+		let expr_type = &*object
+			.evaluated_type
+			.borrow()
+			.expect("Object should have a type in this stage");
+
+		let object_env = match expr_type {
+			Type::Class(x) => &x.env,
+			Type::Resource(x) => &x.env,
 			_ => return false,
 		};
```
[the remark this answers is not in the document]
eladb: Done

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
 
-		let s = match r {
-			Reference::Identifier(s) => s,
-			_ => {
-				self.path.insert(0, property.name.clone());
-				return false;
-			}
+		let property_kind = match object_env.lookup(property, None) {
+			Ok(prop) => prop,
+			Err(_) => return false,
```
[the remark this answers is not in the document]
eladb: Done

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
-			.as_variable()
-			.expect("'this' to be a variable")
-			.type_;
-
-		let r = match &*resource_type {
-			Type::Resource(r) => r,
-			_ => panic!("'this' is not a resource"),
+		let variable = match property_kind.as_variable() {
+			Some(v) => v,
+			None => return false,
```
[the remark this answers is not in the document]
eladb: Done

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
 
-		// lookup our field in the class environment
-		let field_kind = r
-			.env
-			.lookup(property, None)
-			.expect("unable to find field in resource env")
-			.as_variable()
-			.expect("field is not a variable");
+		// now distinguish between two cases: fields and methods:
+		// - for fields, we only allow we only allow referencing inflight fields and non-reassignable,
```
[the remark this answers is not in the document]
eladb: fixed

examples/tests/valid/inflight_ref_resource_collection.w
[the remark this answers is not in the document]
yoav-steinberg: The rational was to add something specific to the fact that we have a **resource** in the array. Verifying the id is as expected is one way. Not critical.

libs/wingc/src/type_check.rs
```diff
@@ -624,6 +624,17 @@ impl TypeRef {
 			_ => false,
 		}
 	}
+
+	/// Returns `true` if the type is a resource or a collection of resources
+	pub fn contains_resources(&self) -> bool {
```
Chriscbr, line 629: Should structs or interfaces be included here? (I'm not sure)

libs/wingc/src/diagnostic.rs
```diff
@@ -111,6 +111,22 @@ impl WingSpan {
 			false
 		}
 	}
+
+	/// Returns the source code text of the span (reads the file from disk and extracts the text)
+	/// (currently only handles single line spans)
+	pub fn code(&self) -> String {
+		let Ok(contents) = fs::read_to_string(&self.file_id) else {
+			return format!("<cannot read {}", self.file_id);
```
Chriscbr, line 119:
```suggestion
			return format!("<cannot read {}>", self.file_id); 
```

libs/wingc/src/diagnostic.rs
```diff
@@ -111,6 +111,22 @@ impl WingSpan {
 			false
 		}
 	}
+
+	/// Returns the source code text of the span (reads the file from disk and extracts the text)
+	/// (currently only handles single line spans)
+	pub fn code(&self) -> String {
+		let Ok(contents) = fs::read_to_string(&self.file_id) else {
```
Chriscbr, line 118: I think we might we can improve the performance of this in the future by storing a byte offset into the original file inside `Position`, so that we don't have to read the entire file when we are obtaining the span. Add a TODO / issue?

(We should probably avoid loading entire source files into memory multiple times while compiling source code)

libs/wingc/src/jsify.rs
```diff
@@ -1232,135 +1221,305 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+		// check if the expression starts with "this"
+		if !parts.is_empty() {
+			if let Some(first) = parts.get(0) {
+				if first.text == "this" {
+					// split the expression into two parts - capturable and non-capturable
+					let mut capturable = vec![];
+					let mut qualification = vec![];
+
+					for p in parts.iter() {
+						let (flight, reassginable): (Option<Phase>, Option<bool>) = if let Some(v) = &p.variable {
```
Chriscbr, line 1236:
```suggestion
						let (flight, reassignable): (Option<Phase>, Option<bool>) = if let Some(v) = &p.variable {
```

examples/tests/valid/inflight_ref_resource_collection.w
```diff
@@ -1,16 +1,19 @@
...
+    this.array.at(0).add_object("foo.txt", "bar");
+    this.bucket = new cloud.Bucket() as "b2";
   }
 
   inflight test() {
-    assert(this.array.at(0).list().length == 0);
+    assert(this.array.at(0).list().length == 1);
+    assert(this.array.at(0).get("foo.txt") == "bar");
```
Chriscbr, line 16: I don't think we can compile this example at the moment, right?

It looks like we are just generating an annotation like this
```js
Test._annotateInflight("test", {"this.array": { ops: ["at"] }});
```
so the test will fail inflight 😞 

I reckon we need a bit more machinery on both sides. First, we need a way to represent information that says "here's a list of operations used on elements of this collection" in our inflight annotation data structure. For example:

```js
Test._annotateInflight("test", {"this.array": { elem_ops: ["list", "get"] }});
```

This says that we need "list" and "get" operation permissions from all of the buckets INSIDE the collection. From a bindings/permissions perspective, we don't care about what operations are called on the array itself (like `at`).

Second, the compiler needs to be able to obtain that information through analysis. I'm thinking that some operations that access elements of collections, like `map.get()` or `array.at()` need to be special cased in our compiler's binding checker. Note that we don't care about the actual index passed, or the key of the map, etc. -- `Map`, `Array`, etc. are all homogenous collections so we shouldn't treat the elements as pets / try to generate different permissions for them.

Anyways, these are just my initial thoughts of one direction I'd go in (probably beyond the scope of this PR) - curious if any parts of it resonate to you or not. :)

Chriscbr: We could also extend the approach I mention above to structs. For example, I'd like to write this in Wing:

```js
struct MyBuckets {
  east: cloud.Bucket;
  west: cloud.Bucket;
}
```

and based on usage, have specific annotations generated like:

```js
Test._annotateInflight("test", {"this.my_buckets": { key_ops: { east: ["list"], west: ["get"] } }});
```

But we can probably live without this for beta since a struct with resources isn't too different from a composite resource. (However, there is an important difference which is that a struct can be instantiated inflight while a resource cannot! ☝️)

libs/wingc/src/type_check.rs
```diff
@@ -708,6 +708,18 @@ impl TypeRef {
...
+		match **self {
+			Type::Array(t) => Some(t),
+			Type::MutArray(t) => Some(t),
+			Type::Map(t) => Some(t),
+			Type::MutMap(t) => Some(t),
+			Type::Set(t) => Some(t),
+			Type::MutSet(t) => Some(t),
+			_ => None,
+		}
+	}
```
Chriscbr, line 721: What about adding `Type::Optional(t)` as a case? Philosophically, you can think of an optional as a collection with at most one element. (In Rust, the `Option` type is iterable as an example)

examples/tests/invalid/inflight_ref_explicit_ops.w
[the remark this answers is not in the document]
eladb: Fixed

examples/tests/valid/inflight_ref_resource_collection.w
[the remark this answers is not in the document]
eladb: Fixed

libs/wingc/src/diagnostic.rs
eladb: Yes, agreed. I suspect this will be primarily used for debugging purposes. I am wondering if there's a way to make this clearer from the name or some Rust annotation?

examples/tests/valid/inflight_ref_resource_collection.w
eladb: This is an error (for now).

libs/wingc/src/type_check.rs
eladb: I am not sure it makes sense for me to treat `Type::Optional` as a collection.

eladb:
@yoav-steinberg @Chriscbr can you take another look here? I think this is done.

libs/wingc/src/diagnostic.rs
Chriscbr: You could try using `pub(crate)` to say it's only meant to be used internally, or just add a comment

libs/wingc/src/type_check.rs
eladb: Getting rid of this.

libs/wingc/src/jsify.rs
```diff
@@ -1233,136 +1223,188 @@ impl<'a> FieldReferenceVisitor<'a> {
...
-					.get_method(&op)
-					.filter(|v| v.flight == Phase::Inflight && !v.is_static)
-					.is_none()
-				{
+			// okay, so we have a non-reassignable, capturable type.
+			// one more special case is collections. we currently only support
+			// collections which do not include resources because we cannot
+			// qualify the capture.
+			if let Some(inner_type) = variable.type_.collection_item_type() {
+				if inner_type.as_resource().is_some() {
```
Chriscbr, line 1292: This doesn't handle collections of collections of resources - maybe we want some kind of `contains_resources()` function?

(not that important)

libs/wingc/src/diagnostic.rs
```diff
@@ -111,6 +111,22 @@ impl WingSpan {
 			false
 		}
 	}
+
+	/// Returns the source code text of the span (reads the file from disk and extracts the text)
+	/// (currently only handles single line spans)
+	pub fn code(&self) -> String {
```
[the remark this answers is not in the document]
eladb: I am seeing a warning `unused import: io::Write`. Any idea?

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
 }
 
 impl<'a> FieldReferenceVisitor<'a> {
-	/// If node is a reference to a field of "this", check that it references a legal field and add it
-	/// to the list of references. Returns true if no additional processing is needed or false if we
-	/// need to visit the node's children.
+	/// this function is called for every reference in the AST. We only care about references that are
+	/// to fields of "this" (e.g. `this.my_field` or `this.another.first.my_resource.list`).
+	///
+	/// NOTE: if the code has something like `this.foo.bar().zoo` then the reference is just `this.foo.bar`.
```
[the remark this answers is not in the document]
eladb: This is addressed.

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
 	fn collect_inflight_refs(&mut self, node: &Reference) -> bool {
+		// TODO: this is a hack to identify that the reference is to a field of "this".
+		if !node.to_string().starts_with("this.") {
```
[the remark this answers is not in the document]
eladb: Done

libs/wingc/src/jsify.rs
```diff
@@ -1242,125 +1229,136 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'_> {
...
+			return false;
+		}
+
+		// destructure the node into "object" and "property". In the example above, object is
```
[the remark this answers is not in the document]
eladb: Not relevant anymore

Chriscbr (approved):

libs/wingc/src/jsify.rs
eladb: https://github.com/winglang/wing/issues/1830

PR #1744, opened 2023-03-07
feat(compiler): allow expressions as json elements

hasanaburayyan:
Allows use of variable reference for Json element:

```js
let x = 5;
let j = Json x;
```

Closes: https://github.com/winglang/wing/issues/1745


examples/tests/valid/json.w
```diff
@@ -21,6 +21,9 @@ assert(json_mut_obj.get("hello") == message);
 // Assignment from natives
 let some_number: num = 999;
 
+let jj = Json some_number;
+assert(jj == some_number);
```
eladb, line 25: didn't we just say that this shouldn't work?

hasanaburayyan: hmm I thought we said this should work 😅. Are you referencing using the variable as the root element? Or the equality check?

eladb: Can we also add:

```js
let jj1 = Json { foo: some_number };
let jj2 = Json [ some_number, { bar: some_number } ];
```

eladb (changes requested):

examples/tests/valid/json.w
eladb: > hmm I thought we said this should work 😅. Are you referencing using the variable as the root element? Or the equality check?

Since `some_number` and `jj` have different types, they should not be equal.

hasanaburayyan: Oh yes the equality still needs to be fixed (I planned on separate PR), I just wanted to use this PR to update grammar. To support the identifier as root element in Json.

examples/tests/valid/json.w
```diff
@@ -21,6 +21,10 @@ assert(json_mut_obj.get("hello") == message);
 // Assignment from natives
 let some_number: num = 999;
 
+let jj = Json some_number;
+let jj1 = Json {foo: some_number};
+let jj2 = Json [ some_number, {bar: some_number} ];
```
Chriscbr, line 26: What happens if someone uses some other kind of expression? e.g.

```
let jj3 = Json returns_number();
let jj4 = Json foo.bar;
```

Part of me feels like either all of these should work or none of these should work (but it's definitely not a hill I'm going to die on)

hasanaburayyan: Yea im not sure where the line is, if there is one. Maybe we do just allow it all, and change the json element to be any valid expression.

I dont have any super strong opinions here. Maybe just go ahead and open up the flood gates and see where that takes us?

yoav-steinberg: I'm for changing this to any valid expression, but forcing the type checker to validate the expression's type evaluates to either a Json or a Json-convertable value  (a primitive).

staycoolcall911: I'm with @yoav-steinberg's suggestion

examples/tests/valid/json.w
```diff
@@ -21,6 +21,28 @@ assert(json_mut_obj.get("hello") == message);
...
+  return "hello";
+};
+
+let jj3 = Json get_str();
+assert(jj3 == Json "hello");
+
+resource Foo {
+  _sum_str: str;
+  init() {
+    this._sum_str = "wow!";
```
eladb, line 38: Wow indeed!

eladb (approved):

PR #1751, opened 2023-03-08
feat(sdk): standard array api

### LINKED ISSUE #1234: Container types API
As a user of wing I would like to be able to interact with the container types object (Set, MutSet, Array, MutArray, Map, MutMap)

## Definition of done

- [x] Version one of the api for the following types
  - [x] Set
  - [x] MutSet
  - [x] Map
  - [x] MutMap
  - [x] Array
  - [x] MutArray
- [x] Add conversion from immutable container types to mutable `.copy_mut()` and vice versa `.copy()`
- [ ] Do not forget to support `.len` (currently `.size` for Set, `.length` for Array))
- [ ] Reference doc
- [ ] Update the [wing language spec](https://docs.winglang.io/reference/spec) with a link to the reference doc (e.g. in the Arrays section, for instance)

## Implementation note

We want to start with an iteration approach (quick and dirty) for adding the required interfaces on the `Iterable` object.
I would go over the current Set, Map, etc. APIs of Javascript and quickly implement the required bridges.

staycoolcall911:
Adding the api to the standard library's `array` object (`map` and `set` were already implemented). 

This partially fixes #1234 - still need to fix the reference docs - I'll do it in a subsequent PR.


libs/wingsdk/src/std/array.ts
```diff
@@ -28,6 +28,30 @@ export class ImmutableArray {
     throw new Error("Abstract");
   }
 
+  /**
+   * Merge arr to the end of this array
+   * @param arr array to merge
+   *
+   * @returns a new ImmutableArray with the values of this array followed by the values of arr
+   */
+  public concat(arr: ImmutableArray): ImmutableArray {
```
hasanaburayyan, line 37: What should happen if someone tries to concat two mutable arrays? Is the resulting array also mutable?

examples/tests/valid/std_containers.w
```diff
@@ -2,28 +2,35 @@ let s_array = ["one", "two"];
...
+assert(immut_array.length == 3);
+let s_array2 = ["if", "you", "build", "it"];
+let s_array3 = ["he", "will", "come", "for", "you"];
+let merged_array = s_array2.concat(s_array3);
+assert(merged_array.contains("build"));
+assert(!merged_array.contains("bring"));
+assert(merged_array.index_of("you") == 1);
+assert(merged_array.join(" ") == "if you build it he will come for you");
+assert(merged_array.join() == "if,you,build,it,he,will,come,for,you");
+assert(merged_array.last_index_of("you") == 8);
```
hasanaburayyan, line 19: Maybe add an example for mutable arrays being concatenated

libs/wingsdk/src/std/string.ts
```diff
@@ -45,7 +45,7 @@ export class String {
    */
   public contains(searchString: string): boolean {
     searchString;
-    throw new Error("Macro");
+    throw new Error("Abstract");
```
hasanaburayyan, line 48: Shouldnt this still throw "Macro"? I think we are using the Error("Macro") anywhere where we use macro overrides.

hasanaburayyan (commented):
Looks good, I have a question about MutArrays.

libs/wingsdk/src/std/array.ts
staycoolcall911: I'd say since wing promotes immutability by default it's ok to always return an immutable array and if the user wants it mutable they can call `.copy()` on the result. Does that make sense?

staycoolcall911: Also noticed that the current implementation requires converting a mutable array to immutable if it is passed as an argument to `.concat()`, which might not be the best DX, but may still be legit:
```ts (wing)
let some_mut_arr   = MutArray<num> [1, 2, 3];
let some_mut_arr2 = MutArray<num> [4, 5, 6];
some_mut_arr.concat(some_mut_arr2.copy());
```

hasanaburayyan: I think it may be strange that two mutable things create something immutable, but maybe from design perspective its just a safer choice.

What I would be concerned about then is that this api alone will probably not allow you to pass in a Mutable array without overriding the method in the subclass and changing the arg type to `MutableArray`

staycoolcall911: > What I would be concerned about then is that this api alone will probably not allow you to pass in a Mutable array without overriding the method in the subclass and changing the arg type to MutableArray

Yeah, that's ugly 😄 I will avoid overriding the `.concat()` method in `std.MutableArray`.
I think I'll need to find a way to make our type_checker treat `MutArray` as a subtype of `Array`.

staycoolcall911: Fixed with help from @yoav-steinberg. `MutArray` is considered a subtype of `Array` if their values are subtypes. I'll add the same logic to `Map`/`MutMap` and `Set`/`MutSet`.
Commit: [`73e66ca` (#1751)](https://github.com/winglang/wing/pull/1751/commits/73e66ca89b23d2cd05858ff88f5e9fe2bf9e1425)

hasanaburayyan (approved):
💪

PR #1753, opened 2023-03-08
feat(vscode): symbol and property completions

MarkMcCulloh:
### Parser Changes
- More surgical usage of tree-sitter errors. We no longer throw away entire expressions just because they contain an error
  - This often replaces `Unknown parser error` with something more useful

### Grammar changes
- Formated grammar.js
- The "property" part of nested identifiers is now optional (!)
  - **Note**, the rust parser will still fail on this. This change was made to improve the error recovery of incomplete references. We can provide a better error message than tree-sitter can. Plus, the lsp handles this better
  - Better error handling for this would require us to write our own additional tree-sitter scanner in C, which I decided to not do (cause I don't wanna)

### Misc compiler change
- Made `parent` on SymbolEnv public
  - the usual "with_ancestry" was not good enough on iter() and it didn't seem valuable to change or make new iterators

### Extension changes
- Forced the `wordPattern` languageConfiguration to be set for Wing in vscode, this ensures autocomplete gets triggered when typing words (without having to use a trigger character)

### Corpus changes
- Due to grammar/parser changes, updated invalid examples with more focused errors

### LSP changes
- When compiling for the LSP, disabled capture visiting. It seems very fragile with incomplete code so it needs to be revisited.

### Completion Support for Symbols/Properties

https://user-images.githubusercontent.com/1237390/223821693-357989c3-b5e4-4a2e-8b69-e1ed0a1da5a4.mp4

#### What's missing?
Lots of stuff! but I'm tired of working on this now so...
- Can't handle incomplete nested expressions, only nested identifiers
  - This is most notable for calls like `a.b().`
  - (sidenote) `resolve_reference()` really needs to be exposed outside of TypeChecker
- props/methods of generic classes/builtins are not reified (shows T1)
- Many details of various types completion types are missing
- No ordering on the locality of suggestions


libs/wingc/src/ast.rs
```diff
@@ -399,6 +399,7 @@ pub enum InterpolatedStringPart {
 #[derivative(Debug)]
 pub struct Scope {
 	pub statements: Vec<Stmt>,
+	pub span: WingSpan,
```
eladb, line 402: Nice! I needed that!

libs/wingc/src/parser.rs
```diff
@@ -96,7 +96,7 @@ impl Parser<'_> {
 	}
 
 	fn check_error<'a>(&'a self, node: Node<'a>, expected: &str) -> DiagnosticResult<Node> {
-		if node.has_error() {
+		if node.is_error() {
```
eladb, line 99: What are the implications if there errors in child nodes?

libs/wingc/src/parser.rs
```diff
@@ -985,7 +995,15 @@ impl Parser<'_> {
 		for node in iter {
 			if !self.error_nodes.borrow().contains(&node.id()) {
 				if node.is_error() {
-					_ = self.add_error::<()>(String::from("Unknown parser error."), &node);
+					if node.named_child_count() == 0 {
+						_ = self.add_error::<()>(String::from("Unknown parser error."), &node);
```
eladb, line 999: Why do we need to assign to _?

Curious...

libs/wingc/src/lsp/completions.rs
```diff
@@ -103,28 +23,387 @@ pub unsafe extern "C" fn wingc_on_completion(ptr: u32, len: u32) -> u64 {
...
+				.root_node()
+				.descendant_for_point_range(point, point)
+				.expect("There is always at-least one tree-sitter node");
+		}
+		let file = if let Some(file) = root_scope.statements.first().and_then(|s| Some(s.span.file_id.clone())) {
+			file
+		} else {
+			return vec![];
+		};
+		// let position = Position {
```
eladb, line 61: Delete?

eladb (commented):
Can't wait for this to ship!

libs/wingc/src/parser.rs
MarkMcCulloh: I don't think there is any. Any errors in children will either be handled directly or via the later phase where we check all errors in the tree.

This was added before we had that later phase, so it made a little more sense then.

libs/wingc/src/parser.rs
MarkMcCulloh: `self.add_error` returns a `Result`, rust doesn't like you let Results hang out without using them somehow

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -137,19 +137,15 @@ Error at ../../../../examples/tests/invalid/mut_container_types.w:25:23 | Expect
...
 exports[`optional_invoke.w > wing test (--target sim) - invalid > stderr 1`] = `
-"Compilation failed with 5 error(s)
-Error at ../../../../examples/tests/invalid/optional_invoke.w:3:1 | Unknown parser error.
-Error at ../../../../examples/tests/invalid/optional_invoke.w:3:10 | Unknown parser error.
-Error at ../../../../examples/tests/invalid/optional_invoke.w:3:21 | Expected expression
-Error at ../../../../examples/tests/invalid/optional_invoke.w:7:1 | Unknown parser error.
-Error at ../../../../examples/tests/invalid/optional_invoke.w:8:1 | Unknown symbol \\"f\\""
+"Compilation failed with 2 error(s)
+Error at ../../../../examples/tests/invalid/optional_invoke.w:7:1 | Cannot call into inflight phase while preflight
+Error at ../../../../examples/tests/invalid/optional_invoke.w:8:1 | Cannot call into inflight phase while preflight"
```
yoav-steinberg, line 142: Something's wrong in this test, it's not producing anything related to it's name: "optional_invoke". We need to fix it, that's why err snapshots are important.

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -137,19 +137,15 @@ Error at ../../../../examples/tests/invalid/mut_container_types.w:25:23 | Expect
...
 `;
 
 exports[`optionals.w > wing test (--target sim) - invalid > stderr 1`] = `
-"Compilation failed with 3 error(s)
-Error at ../../../../examples/tests/invalid/optionals.w:1:1 | Unknown parser error.
-Error at ../../../../examples/tests/invalid/optionals.w:1:10 | Unknown parser error.
-Error at ../../../../examples/tests/invalid/optionals.w:6:1 | Unknown symbol \\"f\\""
+"Compilation failed with 2 error(s)
+Error at ../../../../examples/tests/invalid/optionals.w:6:1 | Cannot call into inflight phase while preflight
```
yoav-steinberg, line 147: Something's wrong in this test, it's not producing anything related to it's name: "optionals". We need to fix it, that's why err snapshots are important.

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -196,13 +192,12 @@ Error at ../../../../examples/tests/invalid/statements_if.w:8:5 | Expected type
...
 Error at ../../../../examples/tests/invalid/struct_expansion.w:3:15 | \\"bublic\\" is not a field of \\"BucketProps\\"
-Error at ../../../../examples/tests/invalid/struct_expansion.w:4:15 | Expected between 0 and 1 arguments but got 2 when instantiating \\"Bucket\\"
-Error at ../../../../examples/tests/invalid/struct_expansion.w:4:32 | Expected type to be \\"BucketProps?\\", but got \\"num\\" instead
-Error at ../../../../examples/tests/invalid/struct_expansion.w:6:15 | Expected expression
-Error at ../../../../examples/tests/invalid/struct_expansion.w:7:15 | Unknown parser error.
-Error at ../../../../examples/tests/invalid/struct_expansion.w:11:3 | Unknown symbol \\"handler\\""
+Error at ../../../../examples/tests/invalid/struct_expansion.w:7:15 | Expected between 0 and 1 arguments but got 2 when instantiating \\"Bucket\\"
+Error at ../../../../examples/tests/invalid/struct_expansion.w:7:32 | Expected type to be \\"BucketProps?\\", but got \\"num\\" instead
+Error at ../../../../examples/tests/invalid/struct_expansion.w:11:3 | Expected 2 arguments but got 1
+Error at ../../../../examples/tests/invalid/struct_expansion.w:11:15 | Unexpected 'keyword_argument'."
```
yoav-steinberg, line 200: I don't get this error: `Unexpected 'keyword_argument'."` the previous snapshot made more sense for line 11.

yoav-steinberg (commented):
:sparkles: amazing work.
Didn't go deeply into the lsp code cause it's a lot and I'm not too familiar with it.
I added lots of tiny nitpicks but some might be important so I'm not approving yet.

libs/wingc/src/lsp/completions.rs
```diff
@@ -102,29 +24,427 @@ pub unsafe extern "C" fn wingc_on_completion(ptr: u32, len: u32) -> u64 {
...
+
+			for (i, statement) in node.statements.iter().enumerate() {
+				if statement.span <= self.location {
+					self.visit_stmt(&statement);
+				} else if self.found_stmt_index.is_none() {
+					self.found_stmt_index = Some(max(i as i64 - 1, 0) as usize);
+				}
+			}
+			if self.found_stmt_index.is_none() {
+				self.found_stmt_index = Some(node.statements.len());
```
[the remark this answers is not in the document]
MarkMcCulloh: It's not necessarily a marker, I do technically use the value. Will add comment.

tools/hangar/src/__snapshots__/invalid.test.ts.snap
MarkMcCulloh: tweaked it

tools/hangar/src/__snapshots__/invalid.test.ts.snap
MarkMcCulloh: I went ahead and removed this file, I didn't see what benefit it was adding. optionals.w covers the same thing

examples/tests/invalid/optional_invoke.w
```diff
@@ -1,8 +1,8 @@
 let a: num? = 5;
 
-inflight f(x: num?) {
+let f = inflight (x: num?) => {
   // do something
-}
+};
 
 f(a);
 f();
```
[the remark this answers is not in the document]
MarkMcCulloh: I deleted this file now instead

libs/wingc/src/lsp/completions.rs
```diff
@@ -102,29 +24,427 @@ pub unsafe extern "C" fn wingc_on_completion(ptr: u32, len: u32) -> u64 {
...
+					if !text.contains(".") {
+						// Currently, we can only handle references that are just a single identifier
+						// In this code path, we are inside an error state so the type checker never resolved full references
+						let found_scope = scope_visitor.found_scope.expect("Should have found a scope");
+						let found_env = found_scope.env.borrow();
+						let found_env = found_env.as_ref().expect("Scope should have an env");
+						let found_phase = Some(found_env.flight);
+						let found_symbol = root_env
+							.try_lookup(text.as_str(), None)
+							.or_else(|| found_env.try_lookup(text.as_str(), None));
```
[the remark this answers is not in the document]
yoav-steinberg: I'd like to have a session about #1644, I'll try catch you about it, worst case it'll be next Wed.

tools/hangar/src/__snapshots__/invalid.test.ts.snap
yoav-steinberg: ok, that's cool, but is leaky. Maybe we can improve here?

MarkMcCulloh: https://github.com/winglang/wing/issues/1766

libs/wingc/src/lsp/completions.rs
```diff
@@ -102,29 +24,428 @@ pub unsafe extern "C" fn wingc_on_completion(ptr: u32, len: u32) -> u64 {
...
+		| Type::Array(_)
+		| Type::MutArray(_)
+		| Type::Map(_)
+		| Type::MutMap(_)
+		| Type::Set(_)
+		| Type::MutSet(_) => {
+			// This section of the code is hacky
+			// This is needed because our builtin types have no API
+			// So we have to get the API from the std lib
+			// But the std lib sometimes doesn't have the same names as the builtin types
```
Chriscbr, line 245: What changes would be needed from the compiler in order to avoid this hacky code? Can we create or link to an issue?

libs/wingc/src/lsp/completions.rs
```diff
@@ -102,29 +24,428 @@ pub unsafe extern "C" fn wingc_on_completion(ptr: u32, len: u32) -> u64 {
 		WASM_RETURN_ERROR
 	}
 }
+
 pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse {
-	FILES.with(|files| {
+	CompletionResponse::Array(FILES.with(|files| {
```
Chriscbr, line 29: We need a way to test these LSP functions -- is this something we could add in the PR (or create an issue for)?

For hovering for example, perhaps we could create a test where we have some Wing fixture file (that we can toss as many edge cases and scenarios into), and write a series of unit tests that are each "given X cursor position in the file, here is a snapshot of the completions that would be generated". WDYT?

libs/wingc/src/lsp/completions.rs
```diff
@@ -102,29 +24,428 @@ pub unsafe extern "C" fn wingc_on_completion(ptr: u32, len: u32) -> u64 {
 		WASM_RETURN_ERROR
 	}
 }
+
 pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse {
```
Chriscbr, line 28: (for my curiosity) are there any other interesting params / will any of them be relevant to us in the future?

Chriscbr (commented):
Really looking forward to shipping this. I added some more feedback, the main one is about tests.

I also think it could be really helpful to create a tracking issue for the known gaps in the different LSP features so that we can try and identify what parts of the LSP / extension our contributors can help with, which could be good first issues, etc.

libs/tree-sitter-wing/grammar.js
```diff
@@ -29,11 +29,9 @@ module.exports = grammar({
     [$.map_literal, $.set_literal],
   ],
 
-  supertypes: ($) => [$.expression, $._literal],
+  conflicts: ($) => [[$.reference, $.custom_type]],
 
-  conflicts: $ => [
-    [$.reference, $.custom_type],
-  ],
+  supertypes: ($) => [$.expression, $._literal],
```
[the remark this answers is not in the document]
MarkMcCulloh: https://github.com/winglang/wing/issues/1777

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: Adding a separate issue https://github.com/winglang/wing/issues/1779

I'm in full agreement of course, but the fixture loading is a whole can of worms that I think deserves its own PR. With the completion PR merged, I think it should be reasonable to pause feature work and focus on LSP testing.

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: Not sure what you mean. For completion this is the whole thing, or are you talking about the rest of the LSP stuff? If so, yeah there are tons. The tough part is prioritization.

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: Honestly, I'm not sure exactly what needs to happen. Here's what I could think of https://github.com/winglang/wing/issues/1780

I'll add it to the comment

Chriscbr (approved):

PR #1754, opened 2023-03-08
feat(sdk): add api resource sim

### LINKED ISSUE #623: cloud.Api - Simulator implementation
See current spec in https://github.com/winglang/wing/pull/858

RaphaelManke:
Adds the simulator implementation if the `cloud.api` resource

closes #623

### DIFF
--- a/libs/wingsdk/src/target-sim/api.onrequest.inflight.ts
+++ b/libs/wingsdk/src/target-sim/api.onrequest.inflight.ts
@@ -1,0 +1,16 @@
+import {
+  ApiRequest,
+  ApiResponse,
+  IApiEndpointHandlerClient,
+} from "../cloud/api";
+
+export class ApiOnRequestHandlerClient {
+  private readonly handler: IApiEndpointHandlerClient;
+  constructor({ handler }: { handler: IApiEndpointHandlerClient }) {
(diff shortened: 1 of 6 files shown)


RaphaelManke:
I started the work on this but not sure which direction it will lead 😅 
How should we interact with the api? 
- Expose a real api using something like express 
- Add a helper function on the simulator api that can be invoked to simulate a api call

Chriscbr:
@RaphaelManke My thought was to start with creating a real API using express - this way users expecting to use tools like Postman to invoke their Wing-created APIs can interact with their endpoints in a similar way whether it's deployed to AWS or the simulator.

RaphaelManke:
Ok
I will check the simulator and look how to start the server when the simulator starts.

Chriscbr:
@ainvoner Since the list of available routes/endpoints (and if they support GET, POST, etc.) is determined in preflight, the information should be mostly static, so it might be possible to get the information using the `getResourceConfig()` API on the simulator. Of course the endpoint URL may need be dynamically chosen if we have to generate a port number on your machine, so that could be a complication - I'll keep you posted

RaphaelManke:
Superseded by https://github.com/winglang/wing/pull/1852

PR #1755, opened 2023-03-08
feat: resources can implement interfaces

### LINKED ISSUE #1677: Implement function subtyping
Examples of function subtyping in action:
- any function that accepts a parameter of type T can be assigned to a variable typed as a function that accepts a parameter of type U as long as T is a subtype of U
- any function that returns type T can assigned to variable typed as a function that returns type U as long as U is a subtype of T
- any function that is phase-independent can be assigned to a variable typed as an inflight function as long as the rest of the signatures match
- any function that is phase-independent can be assigned to a variable typed as a preflight function as long as the rest of the signatures match

The subtyping logic is currently located here:

https://github.com/winglang/wing/blob/c3a953493b5d7374f28a1550889b6aea0419ae8e/libs/wingc/src/type_check.rs#L249

### LINKED ISSUE #1697: Type checking for interfaces

Chriscbr:
* Adds parsing and type checking for `impl Interface` syntax, currently limited to resources. A resource that implements an interface must contain implementations of all of the interface's members (fields or methods) in its environment. Closes #1697
* Implement a first pass at function subtyping - closes #1677.
* renamed `flight` to `phase` in `FunctionSignature`


libs/wingc/src/type_check.rs
```diff
@@ -299,6 +302,27 @@ trait Subtype {
...
+		// But the following pseudocode is not valid:
+		// > let x: independent fn = <preflight function>;
+		// (a preflight function is not a subtype of an inflight function)
+		if self == &Phase::Independent {
+			true
+		} else {
+			self == other
+		}
+	}
+}
```
MarkMcCulloh, line 324: I'm not convinced that `independent = preflight & inflight`, but I guess it's true enough currently so this is useful.

tools/hangar/src/__snapshots__/invalid.test.ts.snap
```diff
@@ -83,6 +83,12 @@ exports[`immutable_container_types.w > wing test (--target sim) - invalid > stde
 Error at ../../../../examples/tests/invalid/immutable_container_types.w:3:4 | Unknown symbol \\"set\\""
 `;
 
+exports[`impl_interface.w > wing test (--target sim) - invalid > stderr 1`] = `
+"Compilation failed with 2 error(s)
+Error at ../../../../examples/tests/invalid/impl_interface.w:3:10 | Resource \\"A (at ../../../../examples/tests/invalid/impl_interface.w:3:10)\\" does not implement method \\"handle\\" of interface \\"cloud\\"
```
MarkMcCulloh, line 88: > interface \\"cloud\\"

Looks like it's missing the full type

libs/tree-sitter-wing/grammar.js
```diff
@@ -94,7 +95,7 @@ module.exports = grammar({
       seq(
         "struct",
         field("name", $.identifier),
-        optional(seq("extends", commaSep($.identifier))),
+        optional(seq("extends", commaSepZeroOrMore($.identifier, false))),
```
MarkMcCulloh, line 98: Why not allow trailing commas here (and a couple other places)? I'm not against this change, just curious about the reasoning since we allow it in most places.

libs/wingc/src/type_check.rs
Chriscbr: Care to elaborate? (just curious - if it's a hunch, that's okay too :)) It might be relevant to the discussion on  #1711

tools/hangar/src/__snapshots__/invalid.test.ts.snap
Chriscbr: Looking into it

libs/tree-sitter-wing/grammar.js
Chriscbr: Ah, yeah I guess I don't see a strong reason to keep the rule different in different places (also it didn't occur to me until now that the `resource A impl B, C, ... Y, Z` clause can probably extend multiple lines), so I'll revert this change

libs/wingc/src/type_check.rs
MarkMcCulloh: independent as an intersection type implies it has the functionality of both preflight and inflight, but it actually has neither. It's not "capturable" like preflight and it can't "capture" preflight things like inflight.

The fact that these aren't actually types makes the usage of subtyping feel off too. 
It kind feels similar to saying `non-reassignable` is a subtype of `reassignable`.

Chriscbr: I see. From a behavior perspective (what the code actually does) maybe it makes more sense to model this as union types (where given a type `T | U`, the only operations we can safely perform on it are ones that make sense for _both_ `T` and `U`)... Maybe I've just confused myself with the name "intersection type"?  😅

MarkMcCulloh (approved):

PR #1769, opened 2023-03-10
chore: e2e test codegen, ci sharding, and being friends with windows again

MarkMcCulloh:
- Added a label to run the full e2e suite on PR
  - The label does not work ideally (e.g. you'll have to manually retrigger) but it's better than nothing
- Added sharding for hangar tests in CI
- Enabled windows in the full e2e matrix
- hangar now generates a `.test.ts` file for each valid example, containing the `compile` and `test` case from before
  - Now each .w file from the examples has its own snapshot
  - Separate files brings several benefits:
    - Allows you to easily run hangar for a specific wing file
    - More concurrency
    - Opens up the possibility to generate assertions for specific tests (like diagnostics 👀)
  - have not yet moved invalid test cases. Their current snapshots are small so it isn't really hard to see the output in one file.
- Separated the nx commands in CI and stopped using the gh action for it
  - I didn't really see the benefit of the dependency on that gh action
  - Having all of the commands together caused too much CPU thrashing and hurt the readability of error outputs
- Updated several actions in gh workflow, due to lack of maintenance and/or warnings during build about deprecations

---

Running all tests (including windows and mac) takes ~10 minutes, which is at least 3 minutes faster than before (and it should slow down less per added test)

image


.github/workflows/build.yml
```diff
@@ -103,55 +104,58 @@ jobs:
...
         run: npx nx workspace-lint
 
-      - name: Build, Test, and Package
-        uses: MansaGroup/nrwl-nx-action@v2
-        with:
-          targets: "build,test,package"
-          all: true
-          # If pushing, exclude hangar because we're going to run it in a separate job
-          args: "${{ github.event_name == 'push' && '--exclude=hangar ' || '' }}--configuration=release --output-style=stream --verbose"
+      - name: Build
```
eladb, line 125: Can we make each one of these an npm script at the root `package.json` and just call that script here?

It's useful to be able to run those on a dev environment

eladb (approved):
NICE!!

.github/workflows/build.yml
MarkMcCulloh: There already is a script that does all this (`npm run build`, added by you I believe), but I wouldn't put it in the workflow. The separation is useful because CI excludes hangar here and it uses "release" with other configurations that are appropriate for CI.

eladb: I meant to add a script for each of these steps separately

eladb (approved):

yoav-steinberg:
> Allows you to easily run hangar for a specific wing file

How would you go about doing that?

MarkMcCulloh:
> @yoav-steinberg  How would you go about doing that?

For any valid example file, there will be a ts file generated. You can use this file as an argument to `nx test hangar` to run only that test.

`npx nx test hangar -- src/test_corpus/valid/while.w.test.ts`:

image

Issue #1733, opened 2023-03-06
queue handler and code should be hidden from the console


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

The following wing program:
```ts (wing)
bring cloud;

let bucket = new cloud.Bucket();
let queue = new cloud.Queue();
let counter = new cloud.Counter();

queue.on_message(inflight (m: str) => {
  bucket.put("file-${counter.inc()}.txt", m);
});
```
 produces the following console view:
<img width="866" alt="image" src="https://user-images.githubusercontent.com/1727147/223159234-79b3d06c-882f-40e8-a163-aade56944946.png">

### I expected this:

Not to see `cloud.Queue-OnMessageHandler-de95ec56` and the `Code` inside `cloud.Queue-OnMessage-de95ec56`

### Instead, this happened:

These resources are not hidden:true inside the tree.json

### Component:

SDK

### Wing Version:

0.5.81

### Platform(s):

MacOS

### Anything else?

We need to display hidden these artifacts

monadabot:
Congrats! :rocket: This was released in [Wing 0.8.3](https://github.com/winglang/wing/releases/tag/v0.8.3).

PR #1774, opened 2023-03-10
fix: remove the "code" node from functions

eladb:
Get rid of the pesky `Code` node from `cloud.Function`s in the simulator.

While at it, get rid of the file constructs for now. They were not really plugging in very well into our architecture. We will introduce them later.


ekeren:
There is another node to remove , check out #1733

Chriscbr (approved):
LGTM 🙏

PR #1775, opened 2023-03-10
feat(compiler): check serializability of json elements

### LINKED ISSUE #1683: Validate Json Types are legal
When creating a Json in Wing the types allowed should be limited. IE right now we can do something like this:
```js
let b = new cloud.Bucket();
let arr = Json [b,b,b];
```

however we should get an error compiling this like:
```
let arr = Json [b,b,b];
//              ^ type cloud.Bucket is not serializable 
Only serializable types should be allowed in Json objects. Maybe we can solve this using interface system.

### PR DESCRIPTION
**Pull request by YarivLevy81**
Should close #1683.

Added validation for serializability of Json elements.

### CODE REVIEW

**Comment by YarivLevy81** [comment]
Missing some more tests, will add them

File: examples/tests/invalid/json.w [hunk 1]
```diff
@@ -20,7 +20,10 @@ let forever_json = Json {a: "hello"};
...
-// let bkt = new cloud.Bucket();
-// let j_arr = Json [bkt];
-//                   ^^^ type cloud.bucket is not serializable 
\ No newline at end of file
+let bkt = new cloud.Bucket();
+let j_arr = Json [bkt];
+//                ^^^ Expected "Json" elements to be serializable, but got "Bucket" which is not serializable
+
+let json_obj = Json { boom: bkt };
+//                ^^^ Expected "Json" elements to be serializable, but got "Bucket" which is not serializable
```
hasanaburayyan, line 28: The tics should be under `bkt`

libs/wingc/src/type_check.rs
```diff
@@ -636,6 +636,28 @@ impl TypeRef {
 			_ => false,
 		}
 	}
+
+	pub fn is_serializable(&self) -> bool {
+		match **self {
+			Type::Resource(_) => false,
+			Type::Enum(_) => true,
```
hasanaburayyan, line 643: For now lets leave Enum as false
```suggestion
			Type::Enum(_) => false,
```

libs/wingc/src/type_check.rs
```diff
@@ -636,6 +636,28 @@ impl TypeRef {
...
+			Type::Json => true,
+			Type::Array(v) => v.is_serializable(),
+			Type::Map(v) => v.is_serializable(),
+			Type::Set(v) => v.is_serializable(),
+			Type::Struct(_) => self.as_struct().unwrap().env.iter(true).all(|(_, v, _)| {
+				v.as_variable()
+					.expect("Expected struct field to be a variable in the struct env")
+					.type_
+					.is_serializable()
+			}),
```
hasanaburayyan, line 657: As well lets keep structs false for now
```suggestion
			Type::Struct(_) => false,
```

libs/wingc/src/type_check.rs
```diff
@@ -1376,6 +1392,31 @@ impl<'a> TypeChecker<'a> {
...
+				self.expr_error(
+					exp,
+					format!(
+						"Expected \"Json\" elements to be serializable, but got \"{}\" which is not serializable",
+						actual_type
+					),
+				);
+			}
+
+			return actual_type;
```
hasanaburayyan, line 1414: We can get rid of a level of nesting with this re-org. Score us some never-nesting points with @Chriscbr :grin:
```suggestion
		if !context.in_json {
			return self.validate_type(actual_type, expected_type, exp);
		}

		if !actual_type.is_serializable() {
			self.expr_error(
				exp,
				format!(
					"Expected \"Json\" elements to be serializable, but got \"{}\" which is not serializable",
					actual_type
				),
			);
		}
		actual_type
```

apps/jsii-docgen/test/docgen/view/__snapshots__/markdown.test.ts.snap
```diff
@@ -21,49 +21,49 @@ new GreeterBucket(props?: BucketProps)
 
 | **Name** | **Type** | **Description** |
 | --- | --- | --- |
-| <code><a href=\\"#custom-construct-library.GreeterBucket.Initializer.parameter.props\\">props</a></code> | <code><a href=\\"#custom-@aws-cdk/aws-s3.BucketProps\\">BucketProps</a></code> | *No description.* |
+| <code><a href=\\"/packages/construct-library/v/0.0.0/api/GreeterBucket.Initializer.parameter.props?lang=python\\">props</a></code> | <code><a href=\\"/packages/@aws-cdk/aws-s3/v/99.99.99/api/BucketProps?lang=python\\">BucketProps</a></code> | *No description.* |
```
hasanaburayyan, line 24: hmm, these changes dont seem intentional. Im not sure why there is all this version 0.0.0 and `lang=python` updates. @Chriscbr or @MarkMcCulloh any ideas?

hasanaburayyan (changes requested):
💪 🚀 Awesome stuff, couple minor change requests. Let me know what you think.

libs/wingc/src/type_check.rs
```diff
@@ -636,6 +636,28 @@ impl TypeRef {
...
 		}
 	}
+
+	pub fn is_serializable(&self) -> bool {
+		match **self {
+			Type::Resource(_) => false,
+			Type::Enum(_) => true,
+			Type::Number => true,
+			Type::String => true,
+			Type::Duration => true,
```
ekeren, line 646: duration should not be supported 
```suggestion
			Type::Duration => false,
```

examples/tests/invalid/json.w
ekeren: We do not support any serializable value in Json we only support str,num, bool and arrays with these values (or another json)  - https://docs.winglang.io/reference/spec#114-json-type

I would change the error message and remove serializable 

//                ^^^ Expected "Json" elements to be Json Value (https://www.json.org/json-en.html)

libs/wingc/src/type_check.rs
```diff
@@ -636,6 +636,28 @@ impl TypeRef {
...
+			Type::Boolean => true,
+			Type::Json => true,
+			Type::Array(v) => v.is_serializable(),
+			Type::Map(v) => v.is_serializable(),
```
ekeren, line 650: I think that Map should not be valid. 

and maybe `is_serializable` is the wrong term to use. Enum is serializable but is not a json_value
maybe something like `is_legal_json_value`

apps/jsii-docgen/test/docgen/view/__snapshots__/markdown.test.ts.snap
Chriscbr: Not sure off the top of my head - let's try to just remove it from the branch for now

libs/wingc/src/type_check.rs
hasanaburayyan: We need to leave map as valid. As for the name change I think thats fine.

hasanaburayyan (commented):
@YarivLevy81 thanks for getting those changes in! 🎉 

Lets uncommit the changes to `apps/jsii-docgen/test/docgen/view/__snapshots__/markdown.test.ts.snap` and I had 1 other suggestion. 

Happy to help if you need any

libs/wingc/src/type_check.rs
```diff
@@ -637,6 +637,23 @@ impl TypeRef {
...
+			Type::Boolean => true,
+			Type::Json => true,
+			Type::Array(v) => v.is_json_legal_value(),
+			Type::Map(v) => v.is_json_legal_value(),
+			Type::Set(v) => v.is_json_legal_value(),
+			Type::Struct(_) => false,
+			_ => false,
+		}
+	}
 }
```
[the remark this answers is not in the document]
YarivLevy81: I think we're good now 😃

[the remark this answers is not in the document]
staycoolcall911: Love this change!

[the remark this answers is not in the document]
YarivLevy81: Merge? @hasanaburayyan

YarivLevy81:
> Thank you for contributing! Your pull request will be updated from main and then merged automatically (do not update manually, and be sure to [allow changes to be pushed to your fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).

I think the bot won't embark the change @hasanaburayyan 😒

PR #1781, opened 2023-03-11
feat(compiler): adding for loop with sequence

### LINKED ISSUE #118: For Statement

marciocadev:
Fixes #118


examples/tests/valid/for_loop.w
```diff
@@ -16,4 +16,16 @@ for word in words {
...
   }
+}
+
+for x in 0..5 {
+  assert(x >= 0);
+  assert(x < 5);
+  print("increment => 0 <= ${x} < 5");
+}
+
+for x in 10..5 {
```
eladb, line 27: Can you add tests for using variables and expressions instead of literal values:

E.g.

```js
for x in y..(z*2) {
  // ...
}
```

libs/tree-sitter-wing/grammar.js
```diff
@@ -200,6 +201,22 @@ module.exports = grammar({
         field("block", $.block)
       ),
 
+    num_sequence: ($) =>
+      seq(
+        field("start", $.number),
```
eladb, line 206: Support expressions

eladb (changes requested):

examples/tests/valid/for_loop.w
```diff
@@ -16,4 +16,16 @@ for word in words {
...
+  assert(x <= 100);
```
staycoolcall911, line 28:
```suggestion
  assert(x <= 10);
```

libs/tree-sitter-wing/grammar.js
marciocadev: done 😄

examples/tests/valid/for_loop.w
marciocadev: good catch

examples/tests/valid/for_loop.w
marciocadev: @eladb, the `for n in 0..3` statement gives `n` the values 0, 1, and 2, similar to how it works in Rust.

However, in Rust, if we use the syntax `for n in 0..=3`, it gives `n` the values 0, 1, 2, and 3.

Do you think it would be a good idea to make the Wing for loop behave the same way as Rust in this case?

libs/wingc/src/jsify.rs
```diff
@@ -577,12 +582,35 @@ impl<'a> JSifier<'a> {
...
+			}
+			StmtKind::ForSequence {
+				iterator,
+				sequence,
+				statements,
+			} => {
+				format!(
+					"{{{}\n{}\n{}}}",
+					format!(
+						"function* iterator(start, end) {{\n  {}\n {}\n {}\n}}",
```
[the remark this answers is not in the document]
marciocadev: I follow the for loop example

https://docs.winglang.io/reference/spec#27-for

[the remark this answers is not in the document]
marciocadev: And using a generator function makes easy to do loops with variables like this

```ts (wing)
let b = 10;

for x in 0..b { ... }

for y in (b*2)..-10 { ... }
```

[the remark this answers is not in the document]
marciocadev: But I can try to create an array and remove the iterator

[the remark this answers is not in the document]
staycoolcall911: I'm ok with the generator - just wanted to ask for the reasoning behind it.
@MarkMcCulloh, @Chriscbr, @hasanaburayyan may have a different opinion.

libs/wingc/src/ast.rs
```diff
@@ -267,6 +267,11 @@ pub enum StmtKind {
 		iterable: Expr,
 		statements: Scope,
 	},
+	ForSequence {
+		iterator: Symbol,
+		sequence: Expr,
+		statements: Scope,
+	},
```
Chriscbr, line 274: This looks off to me, I don't think we should use different AST nodes to distinguish between `for x in <collection>` and `for x in <range>`. "sequence" in `ForSequence` should be the same as "iterable" in `ForLoop`.

libs/wingc/src/type_check.rs
```diff
@@ -1590,6 +1602,51 @@ impl<'a> TypeChecker<'a> {
...
+
+			// TODO: Handle non-builtin iterables
+			t => {
+				self.type_error(TypeError {
+					message: format!("Unable to iterate over \"{}\"", t),
+					span: expr.span.clone(),
+				});
+				self.types.anything()
+			}
+		};
```
Chriscbr, line 1632: Let's extract this into a function like `exp_type.is_iterable()`, which returns `true` for collections and `NumberSequence` and `false` for everything else

marciocadev: I created the is_iterable() function, but I need this variable iterator_type to get the type of variable received by the array.

Chriscbr (approved):
Nice work!

libs/wingc/src/type_check.rs
```diff
@@ -724,6 +724,14 @@ impl TypeRef {
 		}
 	}
 
+	pub fn is_iterable(&self) -> bool {
+		if let Type::Array(_) | Type::Set(_) | Type::MutArray(_) | Type::MutSet(_) = **self {
```
staycoolcall911, line 728: @hasanaburayyan - maybe not part of this PR, but what about `Json`? Maybe open an issue to support iterating over a json-array?

staycoolcall911:
@marciocadev - just FYI:
1. I added "Fixes #118" to this PR description, so it closes the issues when this PR is merged.
2. Also asked @eladb to re-review (since he requested changes) so we could finalize this PR and merge it in.

tools/hangar/__snapshots__/test_corpus/valid/for_loop_sequence.w.ts.snap
```diff
@@ -0,0 +1,334 @@
...
+    while (i > limit) yield i--;
+  }
+  const iter = iterator(0,x,false);
+  for (const i of iter) {
+    {((cond) => {if (!cond) throw new Error(\`assertion failed: '(i >= 0)'\`)})((i >= 0))};
+    {((cond) => {if (!cond) throw new Error(\`assertion failed: '(i < 2)'\`)})((i < 2))};
+    {console.log(\`\${i}\`)};
+  }};
+  {console.log(\\"---\\\\nfor i in 0..=x { ... } <=> x = 2\\")};
+  {function* iterator(start, end, inclusive) {
```
eladb, line 189: Instead of generating all of this code, would it make more sense to introduce an SDK type (e.g. `std.Range`) which will implement the [the_iterable_protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol). Then, basically the JSification of `a..b` will become `Range.of(a, b)` or something like that.

marciocadev: I can use the macro with `let a = std.Range.of(0, 10);` in a Wing file because it generates the macro during compilation, but in the jsification process, I couldn't reference the macro.

but I did a refactoring that I believe improved the generated code, and I kept the `std.Range` class.

staycoolcall911: @marciocadev when you're ready for re-review please click the button next to EladB's name in the Reviewers section of this page (on top right)

libs/wingsdk/src/std/range.ts
```diff
@@ -0,0 +1,26 @@
...
+   * @param start lower bound of the range.
+   * @param end upper bound of the range.
+   * @param inclusive if true, the upper bound value is included.
+   * @returns a iterator.
+   */
+  public static of(
+    start: number,
+    end: number,
+    inclusive: boolean = false
+  ): Array<number> {
```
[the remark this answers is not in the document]
Chriscbr: It should be doable -- but since generator functions aren't a native part of jsii, I think it will require some kind of trick, like using docstring annotations, or perhaps adding an `IIterable<T>` type to std. But I think returning an array here is a good enough to get this feature on the ground 👍

revitalbarletz:
@eladb  any additional comments?

tools/hangar/__snapshots__/test_corpus/valid/for_loop.w.ts.snap
```diff
@@ -87,6 +87,119 @@ constructor() {
     {((cond) => {if (!cond) throw new Error(\`assertion failed: '(post_continue_hits === 0)'\`)})((post_continue_hits === 0))};
   }
   {((cond) => {if (!cond) throw new Error(\`assertion failed: '(j === 3)'\`)})((j === 3))};
+  {console.log(\\"---\\\\nfor x in 0..0 { ... }\\")};
+  for (const x of ((s,e,i) => { function* iterator(start,end,inclusive) { let i = start; let limit = inclusive ? ((end < start) ? end - 1 : end + 1) : end; while (i < limit) yield i++; while (i > limit) yield i--; }; return iterator(s,e,i); })(0,0,false)) {
```
eladb, line 91: I expected this to be a call to an `std` function instead of inline.

Something like:

```suggestion
  for (const x of std.range(s,e,i)) {
```

eladb (changes requested):

tools/hangar/__snapshots__/test_corpus/valid/for_loop.w.ts.snap
marciocadev: In the preflight, the Wing library is imported, so it's possible to generate the loop with `std.Range` when compiling, but in the inflight, the lambda doesn't have the Wing SDK imported, so I had to leave the function inline.

I added a loop inside a function to generate a snapshot in the existing test.

eladb: That's a good point about preflight versus inflight.

Kind of feels like we need to import at least a portion of our stdlib into inflight but that's not for now. Can you open an issue so we can track?

Ok let's keep this as is at this point.

eladb:
LGTM. @MarkMcCulloh @Chriscbr @hasanaburayyan can one of you please review as well?

libs/wingc/src/parser.rs
```diff
@@ -939,7 +939,21 @@ impl<'s> Parser<'s> {
...
+				let inclusive = if expression_node.child_by_field_name("inclusive").is_some() {
+					Some(true)
+				} else {
+					Some(false)
+				};
+				Ok(Expr::new(
+					ExprKind::Range {
+						start: Box::new(self.build_expression(&expression_node.child_by_field_name("start").unwrap())?),
+						inclusive: inclusive,
+						end: Box::new(self.build_expression(&expression_node.child_by_field_name("end").unwrap())?),
```
hasanaburayyan, line 952: nit: we started to try and use more `expect` instead of `unwrap` in the compiler so maybe we can just swap it out with:

```suggestion
						start: Box::new(self.build_expression(&expression_node.child_by_field_name("start").expect("range expression should always include start"))?),
						inclusive: inclusive,
						end: Box::new(self.build_expression(&expression_node.child_by_field_name("end").expect("range expression should always include end"))?),
```

hasanaburayyan (approved):
🚀 Looks great 🔥

eladb (approved):

PR #1782, opened 2023-03-11
chore(repo): revert to default number of columns for the contributors section

jakepage91:
Hey there, after committing I realised that contributor avatars didn't increase in size when the column number was decreased, leaving a huge gap on the right hand size (as seen in the screenshot below). I've reverted to the default number of columns (12) which will fill up the readme file from left to right. 

Screenshot 2023-03-11 at 11 40 42


staycoolcall911 (approved):
Nice!

eladb:
@jakepage91 Thanks for this!

PR #1799, opened 2023-03-13
feat(cli): code snippets in error messages

Chriscbr:
Uses the codespan-reporting crate from Rust to print error diagnostics that include the original code snippets.

Screenshot 2023-03-13 at 6 01 47 PM

Error formatting logic is handled within our CLI to keep it outside of the responsibility of wingc. wingc is just responsible for returning a stringified JSON with all of the error messages.


Chriscbr:
@skyrpex @polamoros I'm not familiar with how the Wing console reads the compile errors or what assumptions it's making (we haven't had a chance to formalize that contract). Can you provide instructions for testing if this is a breaking change?

apps/wing/src/commands/compile.ts
```diff
@@ -59,6 +61,23 @@ function resolveSynthDir(outDir: string, entrypoint: string, target: Target) {
...
+      col: number;
+    };
+    end: {
+      line: number;
+      col: number;
+    };
+    file_id: string;
+  };
+  level: "Error" | "Warning" | "Note";
+}
```
MarkMcCulloh, line 79: This type should go in src/wingc.tc to help convey that it's sourced from there. Also it should probably be called `Diagnostic` or something

apps/wing/src/commands/compile.ts
```diff
@@ -227,3 +280,13 @@ function resolvePluginPaths(plugins: string[]): string[] {
   }
   return resolvedPluginPaths;
 }
+
+function offsetFromLineAndColumn(source: string, line: number, column: number) {
```
MarkMcCulloh, line 284: Too bad I recently removed the byte offsets from Wingspan because we weren't using them, that probably would've been a good optimization here.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1,255 +1,887 @@
...
+  |
+1 | let e0 = \\"2\\" + \\"2\\";
+  |          ^^^ Expected type to be \\"num\\", but got \\"str\\" instead
+
+
+error: Expected type to be \\"num\\", but got \\"str\\" instead
+  --> ../../../examples/tests/invalid/types_strings_arithmetic.w:1:16
+  |
+1 | let e0 = \\"2\\" + \\"2\\";
+  |                ^^^ Expected type to be \\"num\\", but got \\"str\\" instead
```
MarkMcCulloh, line 750: Ah, this exposes a pretty unfortunate problem with snapshots (character offsets not matching due to escape characters).

Not worth blocking this great feature, but we probably need to figure out something to see diagnostics as-is

MarkMcCulloh (approved):
Looks amazing, thanks for taking that on. LGTM pending any issues with the Console.

Next step: syntax highlighting in the diagnostics 😁

apps/wing/src/commands/compile.ts
eladb: Why not add them back?

eladb (commented):
Does that impact the console somehow?

ainvoner:
> Does that impact the console somehow?

this is how the change looks like without any change in the console code:

Screenshot 2023-03-14 at 10 15 38

eladb:
@ainvoner Any way we can somehow add support for ansi colors in the blue screen?

apps/wing/src/commands/compile.ts
Chriscbr: Done

apps/wing/src/commands/compile.ts
Chriscbr: #1815

PR #1810, opened 2023-03-14
feat(playground) adding local storage to the playground

tsuf239:
- keeps the example code the same upon refreshing
- avoiding the jump to the first file on the list, when the page re-render (which can be annoying on dev mode)
- most of the changes in the file were made by prettier (I only added lines 16-17, 64, 90-94)

https://www.loom.com/share/75aaf1f2cfa140b7b9446cd5edfc46f3


MarkMcCulloh (approved):
Nice, good idea

Issue #1809, opened 2023-03-14
Cant use `from_json()` methods within `assert()`


hasanaburayyan:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:
```js
let json_of_many = Json {a: 123, b: "hello", c: true};
assert(str.from_json(json_of_many.get("b")) == "hello");
assert(num.from_json(json_of_many.get("a")) == 123);
assert(bool.from_json(json_of_many.get("c")));
```
### I expected this:

Successful compile

### Instead, this happened:

```
preflight error: missing ) after argument list
```

### Component:

SDK

### Anything else?

This is because the assert's backticks are ended prematurely due to fact that the macros used by the `from_json()` method also use backticks 
<img width="1526" alt="image" src="https://user-images.githubusercontent.com/45375125/225015705-b6ecd1f6-a88c-4e09-9180-1046362c7939.png">

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.130](https://github.com/winglang/wing/releases/tag/v0.5.130).

PR #1812, opened 2023-03-14
fix(sdk): use from_json methods in asserts

hasanaburayyan:
Removed the use of backticks from macros to avoid prematurely ending backtick in assert preflight js. 

Closes: https://github.com/winglang/wing/issues/1809


libs/wingsdk/src/std/bool.ts
```diff
@@ -7,7 +7,7 @@ export class Boolean {
   /**
    * Parse a boolean from Json.
    *
-   * @macro ((args) => { if (typeof args !== "boolean") {throw new Error(`unable to parse ${typeof args} "${args}" as a boolean`)}; return JSON.parse(JSON.stringify(args)) })($args$)
+   * @macro ((args) => { if (typeof args !== "boolean") {throw new Error("unable to parse" + " " + typeof args + " " + args + " as a boolean")}; return JSON.parse(JSON.stringify(args)) })($args$)
```
staycoolcall911, line 10: I think you could just convert to single quotes and avoid string concatenation via + sign:
```suggestion
   * @macro ((args) => { if (typeof args !== "boolean") {throw new Error('unable to parse ${typeof args} ${args} as a boolean')}; return JSON.parse(JSON.stringify(args)) })($args$)
```

staycoolcall911 (approved):

libs/wingsdk/src/std/bool.ts
hasanaburayyan: Id love to get rid of the concatenation with `+` but this would not do string interpolation outside of an assert method though.
IE: 

```js
let j = Json {a: "cool"};

assert(bool.from_json(j.get("a")));  // <==== this would give: preflight error: unable to parse string cool as a boolean
```
```js
let j = Json {a: "cool"};

bool.from_json(j.get("a"));  // <==== this would give: preflight error: unable to parse ${typeof args} ${args} as a boolean
```

staycoolcall911: Got it, thanks for clarifying.

MarkMcCulloh (approved):

Issue #1448, opened 2023-02-06
existing closure mechanism is causing permissive permissions 


staycoolcall911:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec
Generate custom resources from inflight closures

```ts
let my_closure = (x: num) => {
  print("${x}");  
};
```
We would like the compiler to generate a custom resource with a single method from this closure. Modeling closures as resources will allow us to correctly handle all use cases of closures and jsify it correctly.

### End consumer Impact

The way we calculate permissions is based on the resource (`this.*`) mechanism and this ticket is required in order to make sure that we do not give permissive permissions to resources when capturing them in this way:

```ts (wing)
let b = new cloud.Bucket();
new cloud.Function(inflight () => {
   b.put("aa", "aaa");
});
```

will give all permissions for the bucket, while:
```ts (wing)
bring cloud;
resource A impl cloud.IFunctionHandler {
  b: cloud.Bucket;
  init(b: cloud.Bucket) { this.b = b;}

  inflight handle() {
    this.b.put("aa", "aaa");
  }
}
new cloud.Function(new A(new cloud.Bucket()));
``` 

will give minimum permission

### Use Cases

For example #1447 for an inflight closure instead of a resource method will work out of the box once #1447 is fixed.

### Component

Compiler

ekeren:
@staycoolcall911 , I am turning this into P1  (see End consumer Impact section in description for details)

Chriscbr:
I started working on this last week, though I haven't begun writing the main implementation (only #2007 done so far)

After looking at this I realized there's at least a little more complexity to the approach I had in my mind than when I first started, so I'm interested in getting some feedback on overall direction if possible.

The high level goal is to add a step to the compiler that transforms inflight closures defined in preflight into resources that implement a single inflight method, so that we can reuse all of the compiler logic for analyzing captures in resource.

> Open question: What are the alternatives to this approach?
>
>  The existing logic for analyzing captures (defined [here](https://github.com/winglang/wing/blob/0a10999b72419bae593596fe10778c6559c8439b/libs/wingc/src/jsify.rs#L1194)) can't seem to be reused for inflight closures right now because it currently assumes that the only resources that need to be captured are those belonging to `this.xxx` fields. But perhaps if we could change the inflight analyzer to avoid that assumption (per #2006) that would resolve this issue instead?

[...]

Issue #1678, opened 2023-02-27
chore(compiler): remove `this` first arg from type system and implement some other instance binding mechanism


yoav-steinberg:
Wing's type system adds `this` as a first arg to all instance methods. I originally designed it that way almost automatically just because it fest obvious to me that method instances need some way of knowing what `this` is. An implication of this approach is that you can also treat instance methods as static methods with another argument:
```
a.do(1);
A.do(a,1); // Another way to do the same
```
But this approach of "how to tell a method what its instance is" makes it complex to pass an instance method as an argument expecting a function signature:
```
call_a_callback(callback: (x:num)) { callback(5); }
let f = print_num(x: num) => { print(x); }
call_a_callback(f); // This will work and print 5, no instance no cry
call_a_callback(a.do); // Does this compile? if the signature of `do` has a `this` then this won't compile
``
Another problem with this approach is although `A.do(a,1)` technically works, no one ever uses this.

I suggest removing `this` as first arg from the type system. We then have two options how to handle passing methods as an argument expecting a function signature: 

1. Now that we have `static` support we can easily distinguish between functions and methods and just fail if we try to pass a method to a function signature. Doing so is usually bad design because if we want some callback on an object, we can make the function expect an interface instead and make sure our object implements that interface.
2. We can implement something like python's callable model where there's a hidden instance optionally attached to function signature argument. In TypeScript this is called binding an object to a function. In TypeScript you can do something like `call_a_callback(a.do.bind(a));`. We can make sure when instance methods are called we compile to that and maybe we'll also need to add something to our internal type system (`FunctionSignature`) to indicate whether this is a bound function (method) or not.

**Comment by Chriscbr** [comment]
#1942 addresses this partially by remodeling the "this" type as a separate field on the compiler's `FunctionSignature` struct instead of as a parameter. However this type information isn't used yet, so it's possible to encounter errors in generated JS code. For example, in Wing 0.8.26 this example fails to compile:

```js
resource A {
  var x: num;
  init() {
    this.x = 5;
  }

  f(inner: (): str) {
    let prev_x = this.x;
    inner();
    assert(this.x == prev_x);
  }

  g(): str {
    this.x = this.x + 1;
    return "done";
  }
}

let a = new A();
let b = a.g;
a.f(b);
```

Error message:

```
preflight error: Cannot read properties of undefined (reading 'x')

  note: intermediate javascript code:
      g()  {
        {
>>     this.x = (this.x + 1);
       return "done";
     }
```

PR #1817, opened 2023-03-14
feat: resources with `handle` method can substitute for function literals

### LINKED ISSUE #1803: `api.get` accepts any inflight closures
### Community Note

### I tried this:

```
bring cloud;

let api = new cloud.Api();
api.get("/hello", inflight (x: num): bool => {
  print("hi");
});
```

### I expected this:

An error saying that the expected type is `inflight (cloud.ApiRequest): cloud.ApiResponse` and not `inflight (num): bool`.

### Instead, this happened:

No compiler error.

### Component:

Compiler, SDK

### Wing Version:

0.5.126

### Node.js Version:

18.12.1

### Platform(s):

_No response_

### Anything else?

This issue applies to all preflight APIs that expect an inflight, like `api.get`, `queue.on_message`, `new cloud.Function(..)` and so on.

This happens right now because to make the compiler work before we had interfaces, we made a simplifying assumption that any preflight method/constructor that requires an inflight will have the parameter typed as "inflight (any): any" instead of a more specific inflight type.

The solution is probably to use interfaces:

https://github.com/winglang/wing/blob/9d9e15f45969f203d0801d2d2dcd0fd123596854/libs/wingsdk/src/target-tf-aws/queue.ts#L43-L46

### LINKED ISSUE #1816: Class instance cannot be assigned to interface
### Community Note

### I tried this:

```
bring cloud;

resource Foo impl cloud.IFunctionHandler {
  init() {}
  inflight handle(message: str): str {
    return "hello world!";
  }
}

let x: cloud.IFunctionHandler = new Foo();
```

### I expected this:

It compiles

### Instead, this happened:

```
Expected type to be "IFunctionHandler", but got "Foo" instead
```

### Component:

Compiler

### Wing Version:

0.5.126

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

_No response_

Chriscbr:
With this PR, it's now possible to use resources in place of inflight literals as long as they implement an interface with an inflight method named "handle" with a matching type. For example:

```js
bring cloud;

resource Foo impl cloud.IFunctionHandler {
  init() {}
  inflight handle(message: str): str {
    return "hello world!";
  }
}

let fn = new cloud.Function(new Foo());

  assert(fn.invoke("test") == "hello world!");
}) as "test";
```

This PR doesn't implement any changes to how code generation is performed, but this change makes progress towards removing instances of `core.Inflight` from the SDK (prereq for #1448).

Misc:
- Unify the naming around "phase" in the compiler

Fixes #1816
Fixes #1803


libs/wingc/src/type_check.rs
```diff
@@ -354,14 +365,11 @@ impl Subtype for Type {
 
 				// In this section, we check if the parameter types are not subtypes of each other, then this is not a subtype.
 
-				// Check that this function has at least as many required parameters as the other function requires
-				if l0.min_parameters() < r0.min_parameters() {
-					return false;
-				}
```
Chriscbr, line 360: I removed this for now since it was causing issues with type checking against interfaces... for now I've left type checking for matching # of params as a future TODO. Related: #1678

libs/wingc/src/type_check.rs
```diff
@@ -384,7 +392,7 @@ impl Subtype for Type {
 					// (Cat) => void is not a subtype of (Animal) => void
 					// but (Animal) => void is a subtype of (Cat) => void
 					// see https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
-					if l.is_strict_subtype_of(&r) {
+					if !r.is_subtype_of(l) {
 						return false;
 					}
```
Chriscbr, line 397: Write unit tests for this

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -138,14 +138,25 @@ exports[`capture_reassignable.w > stderr 1`] = `
 `;
 
 exports[`cloud_function_expects_inflight.w > stderr 1`] = `
-"error: Expected type to be \\"inflight (any): any\\", but got \\"preflight (str): str\\" instead
+"error: Expected type to be \\"IFunctionHandler\\", but got \\"preflight (str): str\\" instead
```
Chriscbr, line 141: This is a regression in error messages - not sure the best way to handle this yet

libs/wingc/src/type_check.rs
```diff
@@ -341,6 +341,17 @@ impl Subtype for Type {
 				// TODO: Hack to make anything's compatible with all other types, specifically useful for handling core.Inflight handlers
 				true
 			}
+			(Self::Function(_), Self::Interface(r0)) => {
+				// TODO: Hack to make functions compatible with interfaces
+				// Remove this after https://github.com/winglang/wing/issues/1448
+				// Compare the function to a method on the interface named "handle" if it exists
+				if let Some(method) = r0.get_env().try_lookup("handle", None) {
+					let method = method.as_variable().unwrap();
+					self.is_subtype_of(&*method.type_)
```
Chriscbr, line 350: Check that the method is an inflight method (otherwise return false)

libs/wingc/src/type_check.rs
```diff
@@ -2922,4 +2951,126 @@ mod tests {
...
+			return_type: void,
+			phase: Phase::Inflight,
+			js_override: None,
+		});
+		let str_fn = Type::Function(FunctionSignature {
+			parameters: vec![string],
+			return_type: void,
+			phase: Phase::Inflight,
+			js_override: None,
+		});
```
yoav-steinberg, line 3011: something like this might be clearer (also for all the other tests)
```suggestion
        let sig = FunctionSignature {
			parameters: vec![num],
			return_type: void,
			phase: Phase::Inflight,
			js_override: None,
		};
		let num_fn = Type::Function(sig);
		let str_fn = Type::Function(FunctionSignature {
			parameters: vec![string], ..sig.clone()
		});
```

libs/wingc/src/type_check.rs
```diff
@@ -2922,4 +2951,126 @@ mod tests {
...
+
+		// functions of incompatible return types are not subtypes of each other
+		assert!(!returns_num.is_subtype_of(&returns_str));
+		assert!(!returns_str.is_subtype_of(&returns_num));
+
+		// functions with specific return types are subtypes of functions with void return type
+		assert!(returns_num.is_subtype_of(&returns_void));
+		assert!(returns_str.is_subtype_of(&returns_void));
```
yoav-steinberg, line 3048: I'm not sure I approve of this relationship needed :)  Why is this needed?

libs/wingc/src/type_check.rs
yoav-steinberg, line 350: Since this is probably a temp hack, I'd do something along the lines of this to hack the `this` first arg and not change the function comparison below.
```suggestion
					let temp_fn = Self::Function(FunctionSignature { parameters: (vec![other]).concat(l0.parameters.clone()), ..l0.clone()});
					temp_fn.is_subtype_of(&*method.type_)
```

libs/wingc/src/type_check.rs
yoav-steinberg: lets not do this change. We'll fix this to something correct in #1678, but I don't want to hurt function comparison for this. See comment above.

libs/wingc/src/type_check.rs
```diff
@@ -727,6 +735,7 @@ impl TypeRef {
 	pub fn is_capturable(&self) -> bool {
 		match **self {
 			Type::Resource(_) => true,
+			Type::Interface(_) => true,
```
yoav-steinberg, line 738: Why do we need this? It's not strictly correct.

libs/wingc/src/type_check.rs
Chriscbr: Looks like yes

examples/tests/invalid/impl_interface.w
```diff
@@ -7,8 +7,12 @@ resource A impl cloud.IQueueOnMessageHandler {
...
   init() {}
-  inflight handle() {
-    // Error: Expected type to be "inflight (str): void", but got "inflight (): void" instead
+  inflight handle(x: num) {
+    // Error: Expected type to be "inflight (str): void", but got "inflight (num): void" instead
     return;
   }
 }
+
+resource C impl cloud.Bucket {
```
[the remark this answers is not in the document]
Chriscbr: Added

libs/wingc/src/type_check.rs
Chriscbr: The reason I have this in place is so that both of these kinds of existing usages in tests will work:

```js
let fn1 = new cloud.Function(inflight (m: str) {
  print("im not returning anything");
});

let fn2 = new cloud.Function(inflight (m: str): str {
  return "Hello, ${m}!";
});

let usage = inflight () => {
  fn1.invoke("");
  let result = fn2.invoke("");
};
```

We don't have generics, so we're kind of in a tough place as far as offering flexible parameter and return types, since `fn.invoke()` needs to always return the SAME type.

The alternative to this (if we don't want this special return-type subtyping rule) is that every `cloud.Function` must have the same return type, which would have to be `str` or `Json`. We could change it to `str` or `Json` but it would be a breaking change at the moment.

Let's open a separate issue about this because I don't think it's core to this PR

Chriscbr: Related: https://github.com/winglang/wing/issues/1843

libs/wingc/src/type_check.rs
Chriscbr: Done

libs/wingc/src/type_check.rs
```diff
@@ -2103,7 +2125,13 @@ impl<'a> TypeChecker<'a> {
 				for interface in implements.iter() {
 					let interface_type =
 						resolve_user_defined_type(interface, env, stmt.idx).unwrap_or_else(|e| self.type_error(e));
```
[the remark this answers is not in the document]
Chriscbr: 👍

libs/wingc/src/type_check.rs
Chriscbr: Ok, added it back

libs/wingc/src/type_check.rs
Chriscbr: This is necessary so that the resources typed as interfaces can be used inflight. Here's an example:

```js
bring cloud;

resource A {
  fn: cloud.IFunctionHandler;
  init(fn: cloud.IFunctionHandler) {
    this.fn = fn;
  }
  inflight do_stuff(message: str): str {
    this.fn.handle(message);
  }
}

resource Greeter impl cloud.IFunctionHandler {
  init() {}
  inflight handle(message: str): str {
    return "Hello ${message}";
  }
}

let fn = new A(new Greeter());
```

Perhaps we need some way to distinguish between interfaces on resources and interfaces on non-resources?

Chriscbr: OK, I've added a check. It's hacky since I couldn't see an easy way to check for type identity but it should be good enough for now

Chriscbr:
@yoav-steinberg Ready for another pass

libs/wingc/src/type_check.rs
```diff
@@ -341,6 +341,39 @@ impl Subtype for Type {
...
+				}
+
+				// Next, compare the function to a method on the interface named "handle" if it exists
+				if let Some(method) = r0.get_env().try_lookup("handle", None) {
+					let method = method.as_variable().unwrap();
+					if method.phase != Phase::Inflight {
+						return false;
+					}
+
+					// hallucinate a "this" parameter for the function, since the
```
eladb, line 360: Very zeitgeist

libs/wingc/src/type_check.rs
```diff
@@ -354,8 +387,9 @@ impl Subtype for Type {
 
 				// In this section, we check if the parameter types are not subtypes of each other, then this is not a subtype.
 
-				// Check that this function has at least as many required parameters as the other function requires
-				if l0.min_parameters() < r0.min_parameters() {
+				// Check that this function has at most as many required parameters as the other function requires
+				// if it doesn't, we know it's not a subtype
+				if l0.min_parameters() > r0.min_parameters() {
```
eladb, line 392: Shouldn't this be part of how we evaluate subtypes for functions?

libs/wingc/src/type_check.rs
```diff
@@ -738,8 +772,23 @@ impl TypeRef {
 	}
 
 	pub fn is_capturable(&self) -> bool {
-		match **self {
+		match &**self {
 			Type::Resource(_) => true,
+			Type::Interface(iface) => {
+				// TODO: This should check that the interface extends `IResource` from
+				// the SDK, not just any interface with the name `IResource`
+				if iface.name.name == "IResource"
```
eladb, line 780: Maybe extract to a helper like `iface.is_resource()`

libs/wingc/src/type_check.rs
Chriscbr: That's what this code is

eladb: Okay oops. Nice I see!

Awesome!

libs/wingc/src/type_check.rs
```diff
@@ -207,6 +207,24 @@ pub struct Interface {
 	pub env: SymbolEnv,
 }
 
+impl Interface {
+	fn is_resource(&self) -> bool {
+		// TODO: This should check that the interface extends `IResource` from
+		// the SDK, not just any interface with the name `IResource`
```
yoav-steinberg, line 213: The way to do this it to lookup the `cloud.IResource` name in the std lib namespace. And then compare the pointers (since there should be only one such interface imported). But I'm fine with making this a future task.

libs/wingc/src/type_check.rs
```diff
@@ -2988,6 +3030,15 @@ mod tests {
 		assert!(!Phase::Inflight.is_subtype_of(&Phase::Preflight));
 	}
 
+	fn make_function(params: Vec<TypeRef>, ret: TypeRef, phase: Phase) -> Type {
```
yoav-steinberg, line 3033: probably need this to avoid a dead code warning
```suggestion
	#[cfg(test)]
	fn make_function(params: Vec<TypeRef>, ret: TypeRef, phase: Phase) -> Type {
```

yoav-steinberg (approved):
really nice work

libs/wingc/src/type_check.rs
Chriscbr: `#[cfg(test)]` is already applied to the module it's in (`mod tests`) so I haven't gotten any dead code warnings

libs/wingc/src/type_check.rs
Chriscbr: Yeah, I wasn't sure how to do a lookup from here. One option might be to move this method so it exists on the `Types` struct instead, but that feels a bit weird, so I was also thinking of simply adding `is_resource` as a `bool` field on `Interface`

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: OK, figured out a way to get a reasonable-ish experience -- if there is an inflight method named "handle" on the interface, it will be displayed in parentheses

PR #1824, opened 2023-03-15
feat(examples): background task manager

Chriscbr:
Adds a proposed Wing example that lets you create background tasks asynchronously, and check their status at any point in time.

Ordinarily, the state of cloud functions (like AWS Lambda) isn't something you can query from the cloud provider. The Wing SDK spec for `cloud.Function` defines an `invoke_async` (not implemented yet) which lets you fire-and-forget a call to the cloud function, but there's no way to check the result unless it uploads it to a bucket or some other stateful cloud storage resource.

The example can also be modified to add retries to background tasks etc.


examples/proposed/task-manager.w
```diff
@@ -0,0 +1,107 @@
...
+resource TaskManager {
+  fn: cloud.Function;
+  state: cloud.Bucket;
+
+  init(handler: inflight (str): void) {
+    this.bucket = new cloud.Bucket();
+
+    let wrapper = inflight (id: str, input: str) => {
+      try {
+        let result = handler.invoke(message);
```
ekeren, line 25: where is message coming from?

Chriscbr: Whoops, that should be `input`

examples/proposed/task-manager.w
```diff
@@ -0,0 +1,107 @@
...
+    this.fn = new cloud.Function(wrapper);
+  }
+
+  /**
+   * Starts a background task with the given input, and returns an id
+   * that can be used to check the status of the task.
+   */
+  inflight start(input: str) {
+    let id = Utils.uuid();
+    this.fn.invoke_async(id, input);
```
ekeren, line 43: Kinda unrelated, but @Chriscbr you are still the guy to ask

why do we have `fn.invoke_async()` vs `defer fn.invoke()`

examples/proposed/task-manager.w
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+    return running.copy();
+  }
+}
+
+let import_contacts = new TaskManager(inflight () => {
+  // pretend to do some work
+  Utils.sleep(5s);
+}, retries: 3);
```
ekeren, line 90: notice that your TaskManager init function accepts only one argument

examples/proposed/task-manager.w
```diff
@@ -0,0 +1,107 @@
...
+  Utils.sleep(5s);
+  assert(import_contacts.check(key) == "done");
+  assert(import_contacts.get_result(key) == "hello");
+}) as "test:import_contacts_work"
+
+// Exercises for the curious reader:
+// - Add a timeout to the background task.
+// - Add a way to cancel a background task.
+// - Add a way to configure the number of retries.
+// - Return a Task object from start() that can be used to check the status of the task.
```
ekeren, line 107: Love this section

examples/proposed/task-manager.w
Chriscbr: You can use `defer fn.invoke()` as long as you `await` it later. (If you do not call `await` on the variable after you `defer` it, I think Wing should give you a compile error).

However for many applications you might not want to `await` it in your main loop -- for example, suppose you want a "GET /hello" request to start a job on the backend but return 200 as long as the job successfully started. In that case, I think `fn.invoke_async()` is a better option. `fn.invoke_async()` just waits for the cloud provider that says "Success, I started the function invocation" so that if it fails due to scaling/throttling etc. you can retry it.

It's also nice in that first time Wing users can use `fn.invoke_async()` without needing to learn about `defer` syntax (less cognitive load).

examples/proposed/task-manager.w
Chriscbr: Oops, good catch

PR #1825, opened 2023-03-15
feat(compiler): optional test operator

yoav-steinberg:
Support the `?` operator for checking if an optional is assigned a value or not. see #436.


libs/wingc/src/jsify.rs
```diff
@@ -532,6 +532,9 @@ impl<'a> JSifier<'a> {
 					context,
 				),
 			},
+			ExprKind::OptionalTest { optional } => {
+				format!("(({}) !== null)", self.jsify_expression(optional, context))
```
Chriscbr, line 536: So this is an interesting question, do we want to check against `null` or against `undefined`, or against both? (These are distinct values in JavaScript because, hey why not 🙃).

My slight inclination would be to check against `undefined` OR to have our language ignore the distinction between `null` and `undefined` in JS and just treat them the same. I'll list some thoughts below but there isn't really a clear/best option to me so I'm also fine leaving this as is.

```suggestion
				format!("((val) => val !== null && val !== undefined)({})", self.jsify_expression(optional, context))
```

Thoughts:
1. When function arguments are omitted in JS, they are treated as `undefined` when you try accessing them, which works out for us
2. JSII libraries don't distinguish between `null` and `undefined` (https://aws.github.io/jsii/specification/2-type-system/#null-and-undefined)
3. More flexible if someone wants to write an API in JS that uses optionals, as their API can return `null` or `undefined` and both are fine. See the example below:

```ts
// helpers.ts
export function func1(): string | undefined {
  return Math.random() > 0.5 ? "hello" : undefined;
}
export function func2(): string | null {
  return Math.random() > 0.5 ? "hello" : null;
}

// wing
resource Utils {
  // do we care to treat these differently? probably not
  extern "./helpers.ts" static inflight func1(): str?;
  extern "./helpers.ts" static inflight func2(): str?;
}
```

examples/tests/valid/optionality_test.w
```diff
@@ -0,0 +1,4 @@
+let x: num? = 4;
+assert(x? == true);
+assert(!x? == false);
+// TODO: add a `x? == false` case when we implement some way to null initialize an optional (https://github.com/winglang/wing/pull/1495)
```
eladb, line 4: We should be able to do this:

```js
let var y: num?;
assert(!y?);
```

libs/tree-sitter-wing/test/corpus/expressions.txt
```diff
@@ -455,3 +455,51 @@ let a = A { f1: 3, f2: true };
         )
     )
 )
+
+================================
+Optional test (and precedence)
+================================
+let a: num? = 1;
+a?;
+!a?;
```
eladb, line 464: While we are at it: `(a.b.c)?` `(a.b)?` `(a)?`

libs/wingc/src/jsify.rs
eladb: For JSII compatibility, I would consider both JS `null` and `undefined` as `false`.

eladb (approved):

libs/wingc/src/jsify.rs
yoav-steinberg: ok

examples/tests/valid/optionality_test.w
yoav-steinberg: We currently have no way to assign a "nothing" value to an optional. I don't really like this:
```wing
let x: num?; // I default to nothing
```
Once the spec is merged I'll align to it and add the test.

PR #1836, opened 2023-03-16
feat(compiler): optional unwrap with default operator (`??`)

yoav-steinberg:
See #436


libs/wingc/src/type_check.rs
```diff
@@ -1073,6 +1073,17 @@ impl<'a> TypeChecker<'a> {
 				} else if op.numerical_args() {
 					self.validate_type(ltype, self.types.number(), left);
 					self.validate_type(rtype, self.types.number(), right);
+				} else if matches!(op, crate::ast::BinaryOperator::UnwrapOr) {
+					// Left argument must be an optional type
+					if !ltype.is_option() {
+						self.expr_error(left, "Left argument of ?? must be an optional type".to_string());
```
MarkMcCulloh, line 1079: I don't think this error needs to mention the visible context. Maybe something along these lines:

```suggestion
						self.expr_error(left, format!("Expected optional type, found \"{}\"", ltype));
```

MarkMcCulloh (approved):
LGTM, only small nit

libs/wingc/src/type_check.rs
```diff
@@ -1073,6 +1073,17 @@ impl<'a> TypeChecker<'a> {
...
+				} else if matches!(op, crate::ast::BinaryOperator::UnwrapOr) {
+					// Left argument must be an optional type
+					if !ltype.is_option() {
+						self.expr_error(left, format!("Expected optional type, found \"{}\"", ltype));
+						return ltype;
+					} else {
+						// Right argument must be a subtype of the inner type of the left argument
+						let inner_type = ltype.maybe_unwrap_option();
+						self.validate_type(rtype, inner_type, right);
+						return inner_type;
```
Chriscbr, line 1085: I'm not sure this handles all cases - I created an issue about a potential bug (#1845)

yoav-steinberg: I don't see how this is a bug. I mean we're unwrapping so how would the result of an unwrap be a wrapped value?

Chriscbr: @yoav-steinberg Sorry for the confusion - I see this is an issue with the current optionality draft spec so I'll leave a comment there instead.

PR #1837, opened 2023-03-16
feat: extern js functions

### LINKED ISSUE #478: Module System - extern

Import external javascript libraries to Wing (such as axios - for making http calls).
```js
bring axios;
let {data} = await axios.get(some_url);
```

This needs to be further discussed.

MarkMcCulloh:
I consider this a first pass at extern, it should at least satisfy the needs of story 15.

This includes the functionality to load node modules in that extern js file

Closes #478, per spec

The primary non-obvious limitation is that the js file referenced is moved, so users cannot use the assumption for its behavior (e.g. relative path references).

Controversial changes needed to support this:
- Removed the "_inflight" postfix from generated inflight clients.
  - AFAIK, this isn't needed for any functionality
  - This was mostly done as a hack to give inflight methods a way to reference static inflight within the same class
- Gave the infight sim function VM access to the parent's `require`
- Pass the entire process object to the synth context and add an additional env to use there
  - passing the env without the context wasn't living long enough and idk why


apps/wing/src/commands/compile.ts
```diff
@@ -76,6 +76,10 @@ export async function compile(entrypoint: string, options: ICompileOptions) {
   const workDir = resolve(synthDir, ".wing");
   log("work dir: %s", workDir);
 
+  process.env["WINGSDK_SYNTH_DIR"] = synthDir;
```
eladb, line 79: Perhaps normalize:

```suggestion
  process.env["WING_SYNTH_DIR"] = synthDir;
```

libs/wingc/src/ast.rs
```diff
@@ -188,11 +188,15 @@ pub struct FunctionDefinition {
...
 	/// The function implementation (body).
-	pub statements: Scope,
+	pub statements: Option<Scope>,
 	/// The function signature, including the return type.
 	pub signature: FunctionSignature,
 	/// Whether this function is static or not. In case of a closure, this is always true.
 	pub is_static: bool,
+	/// The `extern` modifier value, pointing to an external implementation file
+	pub external_implementation: Option<String>,
+
```
eladb, line 198: A thought- It's interesting to connect the concept of `extern` and @macro. One could argue that extern is a sort of macro...

We (still) don't support macros in the Wing syntax but perhaps if we did then maybe we could have implemented extern simply using a `require` macro.

```js
@macro("require('./helper.js').uuid()")
inflight uuid(): str;
```

But maybe this is generalizing too quickly...

libs/wingc/src/ast.rs
```diff
@@ -188,11 +188,15 @@ pub struct FunctionDefinition {
 	/// List of names of function parameters and whether they are reassignable (`var`) or not.
 	pub parameters: Vec<(Symbol, bool)>, // TODO: move into FunctionSignature and make optional
 	/// The function implementation (body).
-	pub statements: Scope,
+	pub statements: Option<Scope>,
```
eladb, line 191: From a modeling perspective, since 
 `statements` and `external_implementation` are be mutually exclusive, I'd model them as an enumeration:

```rust
body: FunctionBody;


// and then
enum FunctionBody {
  External(...),
  Statements(...)
}
```

libs/wingc/src/jsify.rs
```diff
@@ -726,13 +721,19 @@ impl<'a> JSifier<'a> {
...
+		let block = if let Some(scope) = &func_def.statements {
+			self.jsify_scope(
+				scope,
+				&JSifyContext {
+					in_json: context.in_json.clone(),
+					phase: Phase::Inflight,
+				},
+			)
+		} else {
+			"".to_string()
```
eladb, line 734: Add a comment explaining how can this happen

libs/wingc/src/ast.rs
MarkMcCulloh: I like this idea, but I don't think we need to go that far yet. I agree with the direction though

libs/wingc/src/jsify.rs
```diff
@@ -796,17 +791,86 @@ impl<'a> JSifier<'a> {
...
+
+		formatdoc!(
+			"
+			{modifiers} {name}({parameters}) {arrow} {{
 				{body}
 			}}"
 		)
 	}
 
+	/// Retrieves the contents of the extern file and copies it to the intermediate directory.
```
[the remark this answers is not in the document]
MarkMcCulloh: We could even create a file that simply require()s the external file. But that solution and symlinking share the same problem to overcome: We don't have enough info to get the relative path between the out_dir and extern file. AFAIK, I only have enough info to load the extern file.

I think it's possible to optimize further and get the info needed and I can create an issue for that.

libs/wingc/src/ast.rs
eladb: It's good to have a North Star.

libs/wingc/src/capture.rs
```diff
@@ -84,18 +84,24 @@ impl Visit<'_> for CaptureVisitor {
 	}
 
 	fn visit_function_definition(&mut self, func_def: &FunctionDefinition) {
+		let func_scope = if let Some(scope) = &func_def.statements {
```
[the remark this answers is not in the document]
MarkMcCulloh: Updated to use an enum for the function body, which makes this section self-documenting (imo)

libs/wingc/src/capture.rs
```diff
@@ -84,18 +84,24 @@ impl Visit<'_> for CaptureVisitor {
 	}
 
 	fn visit_function_definition(&mut self, func_def: &FunctionDefinition) {
+		let func_scope = if let FunctionBody::Statements(scope) = &func_def.body {
+			scope
+		} else {
+			return;
+		};
```
eladb, line 91: Rust curious: wouldn't it be a bit nicer to use `let/else` here?

```suggestion
		let FunctionBody::Statements(func_scope) = &func_def.body else {
			return;
		};
```

libs/wingc/src/capture.rs
```diff
@@ -301,10 +307,16 @@ fn scan_captures_in_expression(
 			// Can't define preflight stuff in inflight context
 			assert!(func_def.signature.flight != Phase::Preflight);
 			if let Phase::Inflight = func_def.signature.flight {
+				let func_scope = if let FunctionBody::Statements(scope) = &func_def.body {
```
eladb, line 310: Let/else?

libs/wingc/src/capture.rs
```diff
@@ -378,7 +390,12 @@ fn scan_captures_in_inflight_scope(scope: &Scope, diagnostics: &mut Diagnostics)
 			}) => {
 				res.extend(scan_captures_in_inflight_scope(&constructor.statements, diagnostics));
 				for (_, m) in methods.iter() {
-					res.extend(scan_captures_in_inflight_scope(&m.statements, diagnostics))
+					let func_scope = if let FunctionBody::Statements(scope) = &m.body {
```
eladb, line 393: Let/else

libs/wingc/src/jsify.rs
```diff
@@ -722,17 +717,20 @@ impl<'a> JSifier<'a> {
...
+
+		let block = match &func_def.body {
+			FunctionBody::Statements(scope) => self.jsify_scope(
+				scope,
+				&JSifyContext {
+					in_json: context.in_json.clone(),
+					phase: Phase::Inflight,
+				},
+			),
+			// TODO extern usage with closures is not supported
```
eladb, line 730: Why is that a TODO? Do we expect this to be used for something in the future?

eladb (approved):
LgTM!

libs/wingc/src/capture.rs
MarkMcCulloh: I did not know this is valid syntax, neat.
`cargo fmt` does seem to struggle with it though, but it looks nicer so I'll go through and add it.

libs/wingc/src/jsify.rs
MarkMcCulloh: I think it should be reasonable/possible to use extern for local closures. We don't need it right now, but it'd be interesting for later.

I can remove the "TODO" if you'd like, but I think the comment itself is a useful marker.

PR #1838, opened 2023-03-16
fix(sdk): bucket put_json permissions for aws

hasanaburayyan:
fixed issue where I was not checking for inflight methods in binding for put_json and get_json


hasanaburayyan:
@staycoolcall911 @tsuf239 Okay so taking everyone for a roller coaster ride here 😁 turns out I was wrong (nothing new) we do want the inflight enums to stay camel case. 

**TLDR** This was a bug from the original implementation of `put_json` and `get_json` I goofed and didnt add the checks to the bind functions, but because our binding process adds all ops for functions when it captures It never failed my checks when I deployed to AWS. (resources are another story)

The reason this was not caught initially is because I never tested Json put/get methods within a resource inflight (which does not just add all ops to binding)

i.e.
```js
let my_b = new cloud.Bucket() as "my_bucket";

let f = new cloud.Function(inflight (msg: str): str => {
  my_b.put_json("cool", Json "sup");
}) as "my_function";
```

will produce a a preflight binding of
```js
bindings: {
      my_b: {
        obj: my_b,
        ops: ["delete","get","get_json","list","public_url","put","put_json"]
      },
    }
```

but when in a resource
```js
resource TaskList {
  my_bucket: cloud.Bucket;
  init() {
    this.my_bucket = new cloud.Bucket() as "my_bucket";
  }

  inflight put_json(key: str, value: Json) {
    this.my_bucket.put_json(key, value);
  }
}

let task_list = new TaskList();
```

the binding that gets produced only contains `put_json`
```js
TaskList._annotateInflight("put_json", {"this.my_bucket": { ops: ["put_json"] }});
```

@staycoolcall911 I think Im going to punt on additional tests here for now I think we have a broader discussion here on how we test to make sure if things work locally they work in cloud.

staycoolcall911:
Thanks for the thorough check @hasanaburayyan. This means @marciocadev and @tsuf239 were right 😄 Good for them.

Your change makes perfect sense. I'll let @Chriscbr review as well.

Chriscbr (approved):
Thanks for fixing this @hasanaburayyan :)

PR #1844, opened 2023-03-17
fix: incomplete identifiers cause cascading errors at the end of a scope

MarkMcCulloh:
### Huh?
In the grammar, a "reference" by itself is also valid expression. However, a lot of expressions can also start with a reference (or reference-like, in the case of custom types). References can also contain an expression themselves.
Long story short, these properties together made references very greedy and they often gobbled up many tokens like "}" and ")" trying to find the next semicolon when the reference is incomplete. Not sure if this is a tree-sitter bug or just a badly written grammar.

#### Before
image

### What changed?
The primary change here was to remove references as an expression, but still allow identifiers or nested identifiers. This should keep the same expressivity but removes an extra layer of recursion that was hurting parseability.

### How about now?

Incomplete identifiers now have more localized errors. This mostly matters for the vscode extension, but it should be useful for compilation errors in general

#### After
image


libs/tree-sitter-wing/grammar.js
```diff
@@ -56,14 +57,18 @@ module.exports = grammar({
...
       prec(
         PREC.MEMBER,
         seq(
-          field("object", $.expression),
+          field("object", 
+            choice(
+              $.expression, 
+              // This is required because of ambiguity with using Json keyword for both instantiation of Json
+              // and Identifier for static methods.
+              $.json_container_type
```
yoav-steinberg, line 69: In our ast's `Reference` we recognize 3 types:
1. `Identifier`.
2. `InstanceMember`.
3. `TypeMember` (static members).

But until now our grammar only recognized the first 2. And the type checker converted 2's to 3's in `resolve_reference` (see expr_maybe_type). But now if we are creating a `nested_identifier` node for `json_container_type` (or other types) we can make our _parser.rs_ detect this can create a `TypeMember` reference automatically.
Also we should probably not call this `object` anymore. Maybe `object_or_type`.

libs/wingc/src/parser.rs
```diff
@@ -681,8 +681,18 @@ impl<'s> Parser<'s> {
...
+			let object_expr = if object_expr.kind() == "json_container_type" {
+				Expr {
+					kind: ExprKind::Reference(Reference::Identifier(self.node_symbol(&object_expr)?)),
+					span: self.node_span(&object_expr),
+					evaluated_type: RefCell::new(None),
+				}
+			} else {
+				self.build_expression(&nested_node.child_by_field_name("object").unwrap())?
+			};
 			Ok(Reference::InstanceMember {
```
yoav-steinberg, line 694: See comment above about creating a `Reference::TypeMember` in case of an `xxx_container_type`.

yoav-steinberg (approved):
Cool.

libs/tree-sitter-wing/grammar.js
MarkMcCulloh: This is more or a very specific case for Json, because in our grammar we allow `Json (expression)`. I would say our grammar still only really recognizes the first two expect for this very specific case right now.

I considered going directly to TypeMember but I didn't know if the type checker would have trouble with that be done before expected. I'll try it out.

libs/tree-sitter-wing/grammar.js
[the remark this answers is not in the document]
MarkMcCulloh: I think the main issue is our struct/json literals. Since they start with a reference, that's where all of this gets messed up. I think we'd still want to allow literals in references so I think it would be similarly complicated. We could remove closure literals, I don't think there's a grammar issue there.

libs/tree-sitter-wing/grammar.js
MarkMcCulloh: I did get the parser working to generate `TypeMember` appropriately now for the Json.xxx case, will update

eladb:
Very cool!

Issue #1851, opened 2023-03-17
cloud.Api - path parameters in simulator


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Support path parameters for `cloud.Api` in the simulator.

Example:
```js
bring cloud;

let api = new cloud.Api();

let handler = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
  let user_name = req.vars.get("name");
  return cloud.ApiResponse {
    body: "Hello, ${user_name}!",
    status: 200,
    headers: {
      "Content-Type": "text/plain",
    },
  };
};

api.get("/users/{name}", handler);
```

### Use Cases

Create REST APIs that follow typical RESTful patterns

### Component

SDK

monadabot:
Congrats! :rocket: This was released in [Wing 0.9.7](https://github.com/winglang/wing/releases/tag/v0.9.7).

Issue #1850, opened 2023-03-17
cloud.Api - `request` inflight method


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

https://docs.winglang.io/reference/wingsdk-spec#api

This inflight method makes a HTTP request to the endpoint associated with the Api. Similar to JavaScript's [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) function except that the base URL is always the URL of the API endpoint.

### Use Cases

Unit testing `API`s, or calling your own APIs

### Component

SDK

staycoolcall911:
Reducing the priority for this and removing assignee because of #2169

staycoolcall911:
We are implementing fetch - reducing priority again

Chriscbr:
Since the `http` module has been added, I'm not sure this is needed. Closing for now

Issue #1849, opened 2023-03-17
`Api.url` preflight property


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

https://docs.winglang.io/reference/wingsdk-spec#api

This property that returns a string token that resolves to the API URL.

Note that the exact URL is not known until the app is deployed (since the API endpoint may be a random internet URL when you deploy to AWS, or be located on a random port when you run your Wing app using the simulator), but that is okay since all usages of the token will be substituted with the correct value at runtime through the token resolution mechanism.

### Use Cases

Write unit tests that call endpoints on an API. Use the URL of the API in the definitions of the rest of your cloud infrastructure (e.g. as environment variables).

### Implementation Notes

For more info about tokens, check out [this section on Opaque Tokens](https://docs.winglang.io/contributors/rfcs/2022-05-28-winglang-reqs#opaque-primitives-tokens), or read about tokens from CDK: https://docs.aws.amazon.com/cdk/v2/guide/tokens.html

### Component

SDK

Chriscbr:
My initial feeling is that adding a token resolution mechanism for this single use case falls below the threshold of priority for beta (it could be more complicated / require extra time) -- for now, we can support `url` as an inflight property, or support an inflight `request` method for calling on the `Api` endpoints directly, so that folks can write unit tests. @eladb @staycoolcall911 WDYT?

staycoolcall911:
I agree that a token resolution mechanism is not a high priority based on this use case (can we think of other use cases for that mechanism?).

Api.url as an inflight property should be both easy to implement and functionally sufficient for most use cases. We can quickly add it and revisit this later.

eladb:
cdktf already has a token system, as well as the simulator. What's missing?

Chriscbr:
> cdktf already has a token system, as well as the simulator. What's missing?

Perhaps you're right... I was thinking of an example like this:

```js
bring cloud;

let api = new cloud.Api();

new cloud.Function(inflight() => {
  print(api.url);
});
```

If the compiler lifts the entire Api resource (with `ops = []` since no inflight capabilities are used), I'm imagining...

1. The API's URL needs to get added as an environment variable to the cloud function (so the value resolved at deploy time is available to the inflight code)
2. The inflight client could have a getter corresponding to the preflight getter, which just returns the value injected by the environment variable.

[...]

Issue #1848, opened 2023-03-17
cloud.Api - `any` preflight method


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

https://docs.winglang.io/reference/wingsdk-spec#api

This method allows the user to specify an inflight that handles all requests to a particular route (irrespective of the method used).

### Use Cases

Similar to `app.all()` in expressjs (https://expressjs.com/en/4x/api.html#app.all)

### Component

SDK

Chriscbr:
I wonder if `api.all("/path", ...)` would be a better name for this method? 🤷‍♂️

PR #1852, opened 2023-03-17
feat(sdk): api resource on sim target

### LINKED ISSUE #623: cloud.Api - Simulator implementation

Chriscbr:
Closes #623

The API resource is simulated using the ExpressJS library. The base URL of the API is exposed through a runtime attribute in the simulator (as the exact value isn't known until the simulation starts):

```js
const simulator = new sdk.testing.Simulator({ simfile : "./target/hello.wsim"});
await simulator.start(); // start the simulator 

const apiUrl = simulator.getResourceConfig(path).attrs.url;
```

Misc changes:
- Changed ApiRequest.query to be a Map<str> instead of a string (AWS and ExpressJS both provide query parameters natively in this format)

Future work / features not supported yet:
- Expose `url` in preflight through some token mechanism(?) - #1849
- Support routes that handle all HTTP methods (`api.any("/hello", inflight)`) - #1848
- `request` method for unit testing the Api resource - #1850
- Support path variables in the simulator (like `/users/{id}`) - #1851


libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -0,0 +1,169 @@
...
+  );
+}
+
+function transformRequest(req: express.Request): ApiRequest {
+  return {
+    headers: sanitizeParamLikeObject(req.headers),
+    body: req.body,
+    method: parseHttpMethod(req.method),
+    path: req.path,
+    query: sanitizeParamLikeObject(req.query as any),
```
Chriscbr, line 166: Todo - I need to add a unit test for `query` and `path`

libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -0,0 +1,169 @@
...
+  IApiClient,
+  IFunctionClient,
+  parseHttpMethod,
+  sanitizeParamLikeObject,
+} from "../cloud";
+import { ISimulatorContext, TraceType } from "../testing/simulator";
+
+const LOCALHOST_ADDRESS = "127.0.0.1";
+
+export class Api implements IApiClient, ISimulatorResourceInstance {
```
Chriscbr, line 23: Missing docstring

libs/wingsdk/src/target-sim/api.ts
```diff
@@ -0,0 +1,203 @@
...
+import { ApiSchema, API_TYPE } from "./schema-resources";
+import {
+  bindSimulatorResource,
+  makeSimulatorJsClient,
+  simulatorHandleToken,
+} from "./util";
+import * as cloud from "../cloud";
+import * as core from "../core";
+
+export class Api extends cloud.Api implements ISimulatorResource {
```
Chriscbr, line 14: Missing docstring

libs/wingsdk/src/cloud/api.ts
```diff
@@ -255,6 +255,11 @@ export interface ApiHeadProps {}
  */
 export interface ApiConnectProps {}
 
+/**
+ * Inflight methods and members of `Bucket`.
```
Chriscbr, line 259:
```suggestion
 * Inflight methods and members of `Api`.
```

docs/04-reference/wingsdk-spec.md
```diff
@@ -876,7 +876,7 @@ struct ApiRequest {
   /** The request's body. */
   body: Json?;
   /** The request's headers. */
-  headers: Map<str, str>;
+  headers: Map<str, str>?;
```
eladb, line 879: Maybe?

```suggestion
  headers: Map<str, str> = {};
```

libs/wingsdk/src/cloud/api.ts
```diff
@@ -270,22 +275,13 @@ export enum HttpMethod {
...
   /** Options */
   OPTIONS = "OPTIONS",
   /** Patch */
   PATCH = "PATCH",
 }
 
-// /**
-//  * Json type representation.
-//  */
-// // TODO: this should come from a shared package
```
eladb, line 283: Delete?

libs/wingsdk/src/cloud/api.ts
```diff
@@ -326,3 +322,51 @@ export interface IApiEndpointHandlerClient {
...
 }
+
+/**
+ * Parse an HTTP method string to an HttpMethod enum
+ * @param method HTTP method string
+ * @returns HttpMethod enum
+ * @throws Error if the method is not supported
+ */
+export function parseHttpMethod(method: string): HttpMethod {
+  switch (method) {
```
eladb, line 333: Wouldn't this work?

```js
HttpMethod[method]
```

libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -0,0 +1,169 @@
...
+            throw new Error(
+              "Path variables are not yet supported in the simulator."
+            );
+          }
+
+          try {
+            await fnClient
+              // TODO: clean up once cloud.Function is typed as `inflight (Json): Json`
+              .invoke(apiRequest as unknown as string)
+              .then((response) => {
```
eladb, line 84: Why use 'then' instead of a normal async statement?

```js
const response = await fnClient.invoke(..);

// ...
```

libs/wingsdk/src/target-sim/api.ts
```diff
@@ -0,0 +1,203 @@
...
+  private _routes: ApiSchema["props"]["routes"] = [];
+  constructor(
```
eladb, line 16: Empty ctor is not needed

eladb (approved):
Nice job!

docs/04-reference/wingsdk-spec.md
Chriscbr: Interesting idea, maybe add to the issue about structs?

libs/wingsdk/src/cloud/api.ts
Chriscbr: It was commented out, I figured we can always add it back later if we need it

libs/wingsdk/src/cloud/api.ts
Chriscbr: That would work, the only downside is that if it tries parsing an invalid method, it would produce `undefined`. I feel like there could be some value in bubbling that up as an error

docs/04-reference/wingsdk-spec.md
eladb: I believe I've added this to the optionals spec.

PR #1853, opened 2023-03-19
rfc(docs): target specific resources implementation

ekeren:
Still work in progress, not ready for viewing


docs/06-contributors/999-rfcs/2023-05-19-target-specific-resource-implementation.md
```diff
@@ -0,0 +1,118 @@
...
+  } 
+  
+  on_upload(inflight fn: (str, str)): void {
+    // Add tf-aws impl
+  }
+  
+  inflight put(k: str, v:str): void {
+    // Add tf-aws impl
+  }
+}
```
skinny85, line 71: I think a decorator would be a great way to satisfy this usecase:

```js
@Target("tf-aws")
resource TfAwsKvStore extends KvStore {
  // ...
```

I already proposed [using it for `extern`](https://github.com/winglang/wing/issues/478#issuecomment-1465270334), which [Elad didn't like](https://github.com/winglang/wing/issues/478#issuecomment-1465272051). But I think it's a really great fit here.

Chriscbr: 2 cents - so I've come around to the decorators idea, partly after realizing Rust's macros are similar in a way. (For example, in Rust, `#[cfg(test)]` indicates a piece of code should only get compiled when you are testing.) I think an annotation system could be a cheap way to handle these use cases where we want to include extra information for the compiler about resources/fields/functions. 👍 

I only wonder if we should consider different syntaxes, since some developers may associate `@decorator` syntax with how decorators work in Python and TypeScript, where they wrap one function in another function (and I'm not sure if that's the implementation we need to use).

ekeren: @skinny85 , in your example `TfAwsKvStore` is  actually a dummy name that is never used in the code, right?

skinny85: It's never used by the customer code, but I assume Wing’s built-in dependency injection will use it to instantiate the correct class when compiling the code. So, it will be present in the outputted JavaScript.

ekeren: BTW, 
@eladb suggested another syntax option, something like:

```ts (wing)
resource KvStore for "tf-aws"  {
}
```

skinny85: I’m not sure this feature is worth adding new syntax to the language.

skinny85: Basically, I'm worried that there will be more metadata needed to be added to resources in the future, and using something like `for "tf-aws"` won't be able to accommodate those requirements easily.

For example, if you wanted to signify a given resource requires a certain minimum Wing version, with a decorator, it could be something as simple as:

```js
@Target("tf-aws", minVersion = "0.3.4")
```

While with `for "tf-aws"`, I don't see a great option for adding `minVersion`.

Just my 2 cents!

Issue #808, opened 2022-11-29
Wing version bumping


Chriscbr:
### Summary

Support multiple types of version bumps

### Feature Spec

Success criteria: any commit made to the main branch with "!" after the conventional commit type (i.e. "feat!" or "fix(scope)!:") or "BREAKING CHANGE:" in its commit body should result in a major version bump when Wing is released.

When Wing is pre-1.0, this should result in a change to the minor version number. For example, from 0.3.x to 0.4.0.

When Wing is post 1.0, this should result in a change to the major version number. For example, from 1.2.3 to 2.0.0.

MarkMcCulloh:
I believe this issue was solved in https://github.com/winglang/wing/pull/1989

PR #1859, opened 2023-03-20
feat!: improved performance for sim compilation

eladb:
Improve performance from *O(inflight)* to *O(1)* when compiling and testing on the `sim` target by skipping cloud function bundling. As a benchmark, compiling a program with a single cloud function took 6s and now takes 1s. A program with 5 cloud functions took 24s and now also takes 1s.

The `.wsim` format has changed. It is now a directory instead of a tarball and includes the generated javascript files without bundling any dependencies. The output directory is now fixed to be `target/xxx.wsim` relative to the location of the Wing entrypoint file. This allows any `require()` statements to resolve against the correct `node_modules`. In the future we will consider some kind of self-contained bundle format, but for now, we optimize for the local iteration case.

To avoid mistakes, we removed support for specifying the output directory `--out` in the compiler. We can introduce later when we implement bundling for the simulator (or forbid only for the simulator).

- **Compiler**: Changed `js_resolve_path` to produce relative paths instead of absolute paths.
- **SDK**: Moved the bundling logic into a utility function under `bundling.ts`

BREAKING CHANGE: **cli**: The `--outdir` switch has been removed and the default location has been changed so that compilation artifacts are always placed under `target/xxx.ttt` (where `xxx` is the base name of the entrypoint and `ttt` is the target - for example, `target/hello.tfaws`, `target/hello.wsim`).
* **sim**: The `.wsim` output is now a directory and not a file (tarball), and must reside in a location where node module resolution can find any external dependencies.

### DIFF
--- a/libs/wingsdk/src/target-sim/function.inflight.ts
+++ b/libs/wingsdk/src/target-sim/function.inflight.ts
@@ -1,5 +1,6 @@
 import * as fs from "fs";
 import * as path_ from "path";
+import { dirname } from "path";
 import * as process from "process";
 import * as vm from "vm";
 import {
@@ -21,7 +22,7 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
     if (props.sourceCodeLanguage !== "javascript") {
       throw new Error("Only JavaScript is supported");
     }
-    this.filename = path_.resolve(context.assetsDir, props.sourceCodeFile);
+    this.filename = path_.resolve(context.simdir, props.sourceCodeFile);
     this.env = props.environmentVariables ?? {};
     this.context = context;
     this.timeout = props.timeout;
@@ -45,6 +46,15 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
       `exports.handler(${JSON.stringify(payload)});`,
     ].join("\n");
 
+    const resolvePaths = [dirname(this.filename)];
+    const requireResolve = (p: string) =>
+      require.resolve(p, { paths: resolvePaths });
+    const inflightRequire = (p: string) => {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      return require(requireResolve(p));
+    };
+    inflightRequire.resolve = requireResolve;
+
     const context = vm.createContext({
       // TODO: include all NodeJS globals?
       // https://nodejs.org/api/globals.html#global-objects
@@ -53,6 +63,7 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
       path: path_,
       process: {
         ...process,
+
         // override process.exit to throw an exception instead of exiting the process
         exit: (code: number) => {
           throw new Error("process.exit() was called with exit code " + code);
@@ -68,7 +79,7 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
         [ENV_WING_SIM_INFLIGHT_RESOURCE_TYPE]: FUNCTION_TYPE,
       },
 
-      require,
+      require: inflightRequire,
       __dirname: path_.dirname(this.filename),
 
       // Make the global simulator available to user code so that they can find
--- a/libs/wingsdk/src/testing/sim-app.ts
+++ b/libs/wingsdk/src/testing/sim-app.ts
@@ -23,7 +23,7 @@ export class SimApp extends sim.App {
   public async startSimulator(): Promise<Simulator> {
     this.synthIfNeeded();
     const simfile = this.synth();
-    const s = new Simulator({ simfile });
+    const s = new Simulator({ simdir: simfile });
     await s.start();
     return s;
   }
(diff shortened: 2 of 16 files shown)


MarkMcCulloh:
Would it not be useful for the simulator functions to be self-contained as well? I worry this creates an unnecessary behavior difference between sim and non-sim targets.

eladb:
> Would it not be useful for the simulator functions to be self-contained as well?

We can talk about created self-contained `.wsim` file (in which case maybe the right approach would be to simply bundle the npm dependencies instead of bundling each function separately). But that's not P1 right now. We are obsessively optimizing for dev iteration time.

> I worry this creates an unnecessary behavior difference between sim and non-sim targets.

There's a huge difference between sim and non-sim targets in almost every aspect. Same as with every other aspect of our SDK, we will need good test coverage to guarantee that our resources behave the same across the targets. As the project matures, we will get to the point where we have a good approximation. 

Does that make sense?

Chriscbr:
Added a label so this PR gets a minor semver bump

eladb:
> Added a label so this PR gets a minor semver bump

Shouldn't that happen automatically when we specify `BREAKING CHANGE`?

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -45,6 +46,15 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
     ].join("\n");
 
+    const resolvePaths = [dirname(this.filename)];
+    const requireResolve = (p: string) =>
+      require.resolve(p, { paths: resolvePaths });
+    const inflightRequire = (p: string) => {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      return require(requireResolve(p));
+    };
+    inflightRequire.resolve = requireResolve;
```
Chriscbr, line 56: Add a comment on why this is needed

libs/wingsdk/src/testing/simulator.ts
```diff
@@ -10,14 +9,14 @@ import { DefaultSimulatorFactory } from "../target-sim/factory.inflight";
...
 
 /**
  * Props for `Simulator`.
  */
 export interface SimulatorProps {
   /**
-   * Path to a Wing simulator file (.wsim).
+   * Path to a Wing simulator output directory (.wsim).
    */
   readonly simfile: string;
```
Chriscbr, line 21: I would fully commit to the breaking change, we are pre-beta
```suggestion
   * Path to a Wing simulator output directory (.wsim).
   */
  readonly simdir: string;
```

Chriscbr (approved):

Chriscbr:
> > Added a label so this PR gets a minor semver bump
> 
> Shouldn't that happen automatically when we specify `BREAKING CHANGE`?

It should but I don't think it works right now (#808)

Chriscbr:
> I worry this creates an unnecessary behavior difference between sim and non-sim targets.

Yeah, I agree this change does potentially increase the gap between targets (specifically, things like module resolution are always a little bit sensitive / error prone), but I think it's probably outweighed by the improvement this makes for iteration loops while working with the Wing simulator.

Personally I also will miss our short-lived `.wsim` file format 😢 - but I think packaging simulated Wing apps is always something we can add later (as a separate feature) if our users really want it.

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: OK

libs/wingsdk/src/testing/sim-app.ts
```diff
@@ -23,7 +23,7 @@ export class SimApp extends sim.App {
   public async startSimulator(): Promise<Simulator> {
     this.synthIfNeeded();
     const simfile = this.synth();
-    const s = new Simulator({ simfile });
+    const s = new Simulator({ simfile: simfile });
```
eladb, line 26:
```suggestion
    const s = new Simulator({ simfile });
```

libs/wingsdk/src/testing/simulator.ts
eladb: I'll keep it for now.

eladb:
> Personally I also will miss our short-lived .wsim file format

Me too. It will come back with a vengeance!

libs/wingsdk/test/global.setup.ts
```diff
@@ -0,0 +1,6 @@
+import { exec, execSync } from "child_process";
+
+export default async function () {
+  execSync("npx tsc -p tsconfig.test.json");
```
eladb, line 4:
```suggestion
  // compile src/**/*.on*.inflight.ts to .js because these are going to be
  // injected into our javascript vm and cannot be resolved via vitest
  execSync("npx tsc -p tsconfig.test.json");
```

libs/wingsdk/tsconfig.test.json
```diff
@@ -0,0 +1,34 @@
+// ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
```
eladb, line 1:
```suggestion
```

eladb:
Added `do-not-merge`, waiting to sync with @ainvoner tomorrow so we can tandem-ship the console.

PR #1861, opened 2023-03-20
fix(compiler): resource without an initializer

### LINKED ISSUE #1705: Unable to declare a class/resource without an initializer
### Community Note

### I tried this:

```js
resource Foo { }
```

### I expected this:

To work

### Instead, this happened:

```
Error at hello.w:1:1 | No constructor defined in resource {Node resource_definition (0, 0) - (1, 1)}
```

### Component:

Compiler

### Wing Version:

0.5.81

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

marciocadev:
Closes #1705 


libs/wingc/src/capture.rs
```diff
@@ -384,7 +384,12 @@ fn scan_captures_in_inflight_scope(scope: &Scope, diagnostics: &mut Diagnostics)
 			StmtKind::Class(Class {
 				methods, constructor, ..
 			}) => {
-				res.extend(scan_captures_in_inflight_scope(&constructor.statements, diagnostics));
+				if constructor.is_some() {
+					res.extend(scan_captures_in_inflight_scope(
+						&constructor.as_ref().unwrap().statements,
+						diagnostics,
+					));
+				}
```
MarkMcCulloh, line 392: `if let` may be more idiomatic here, to avoid unwrap

```suggestion
				if let Some(constructor) = constructor {
					res.extend(scan_captures_in_inflight_scope(
						&constructor.statements,
						diagnostics,
					));
				}
```

libs/wingc/src/jsify.rs
```diff
@@ -975,7 +975,11 @@ impl<'a> JSifier<'a> {
 			} else {
 				format!(" extends {}.{}", STDLIB, WINGSDK_RESOURCE)
 			},
-			self.jsify_resource_constructor(&class.constructor, class.parent.is_none(), context),
+			if class.constructor.is_some() {
+				self.jsify_resource_constructor(&class.constructor.as_ref().unwrap(), class.parent.is_none(), context)
```
MarkMcCulloh, line 979: similar to above

```suggestion
			if let Some(constructor) = class.constructor {
				self.jsify_resource_constructor(&class.constructor, class.parent.is_none(), context)
```

libs/wingc/src/visit.rs
```diff
@@ -186,7 +186,9 @@ where
 {
 	v.visit_symbol(&node.name);
 
-	v.visit_constructor(&node.constructor);
+	if node.constructor.is_some() {
+		v.visit_constructor(&node.constructor.as_ref().unwrap());
+	}
```
MarkMcCulloh, line 191:
```suggestion
	if let Some(constructor) = node.constructor {
		v.visit_constructor(&constructor);
	}
```

libs/wingc/src/type_check.rs
```diff
@@ -2034,64 +2034,70 @@ impl<'a> TypeChecker<'a> {
...
-				let class_env = &class_type.as_class_or_resource().unwrap().env;
-
-				// Type check constructor
-				let constructor_sig = if let Type::Function(ref s) = *constructor_type {
-					s
-				} else {
-					panic!(
-						"Constructor of {} isn't defined as a function in the class environment",
-						name
+				if constructor.is_some() {
```
MarkMcCulloh, line 2037:
```suggestion
				if let Some(constructor) = constructor {
```

MarkMcCulloh: (will need other changes)

examples/tests/valid/resource.w
```diff
@@ -47,3 +47,5 @@ new cloud.Function(inflight () => {
   assert(s == "counter is: 1");
   assert(bucket.list().length == 1);
 }) as "test";
+
+resource Empty {}
```
MarkMcCulloh, line 51: Could you add some test cases covering if this works with inheritance and classes?

```ts
class T1 {
 init(stuff: string) { }
}

class T2 extends T1 {
}

let x = new T2();
          //^ this should be a type error
let y = new T2("yay");
```

Chriscbr: Depending on how tricky this is, we could start by only allowing `init() {}` to be omitted when the class/resource doesn't `extends` another class/resource

libs/wingc/src/type_check.rs
```diff
@@ -2122,6 +2118,7 @@ impl<'a> TypeChecker<'a> {
 					};
 				}
 
+				if let Some(constructor) = constructor {
```
yoav-steinberg, line 2121: something with the indentation here is wrong, it also seems like there's code duplication here (the comment " Type check constructor" appears twice). Can you take a look and either explain or fix this so it'll be easier to review..

libs/wingc/src/type_check.rs
yoav-steinberg, line 1: What' I'd expect here are the following:
* If there's no constructor defined then make sure `New` expressions are called with no arguments. In that case there's not need to type check the arguments and return value of the constructor.
* If a class/resource has any (instance) fields then they must be initialized in the constructor. In that case we need to fail type checking with an error if there's no constructor: (Type C, which container instance fields a,b,c.. requires an initializer) or something similar.

libs/wingc/src/parser.rs
```diff
@@ -552,7 +542,7 @@ impl<'s> Parser<'s> {
 			methods,
 			parent,
 			implements,
-			constructor: constructor.unwrap(),
+			constructor: constructor,
```
yoav-steinberg, line 545: nit
```suggestion
			constructor,
```

examples/tests/valid/class_without_an_initializer.w
yoav-steinberg, line 1: We need to add some **invalid** tests to verify:
* constructor return type is valid.
* arguments passed to constructor in `new` expression are valid.
* can't define a class with no constructor and instance fields.

yoav-steinberg (changes requested):
Thanks,
added some comments.

revitalbarletz:
@yoav-steinberg can you please review?

eladb:
@marciocadev seems like you need to re-request a review from @Chriscbr as well

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -137,6 +137,23 @@ exports[`capture_reassignable.w > stderr 1`] = `
 "
 `;
 
+exports[`classes_and_resources.w > stderr 1`] = `
+"error: Unknown symbol \\"init\\"
+  --> ../../../examples/tests/invalid/classes_and_resources.w:1:7
+  |
+1 | class C4 { x:num; } // class with field and no init
+  |       ^^ Unknown symbol \\"init\\"
```
Chriscbr, line 145: I'm not sure this error message makes sense to me - `init` is not written anywhere so it's not clear why the compiler is giving this error

libs/wingc/src/ast.rs
```diff
@@ -259,7 +259,7 @@ pub struct Class {
 	pub name: Symbol,
 	pub fields: Vec<ClassField>,
 	pub methods: Vec<(Symbol, FunctionDefinition)>,
-	pub constructor: Constructor,
+	pub constructor: Option<Constructor>,
```
Chriscbr, line 262: Is it possible we can create a default `Constructor` value when parsing (if the user doesn't provide an `init()` function) instead of making this optional? I'd expect the `Constructor` to have an empty list of parameters and an empty body. This will reduce the number of edge cases other parts of the compiler need to handle.

libs/wingc/src/type_check.rs
```diff
@@ -1269,58 +1273,74 @@ impl<'a> TypeChecker<'a> {
...
-				self.validate_type(constructor_sig.return_type, type_, exp);
-
-				if !arg_list.named_args.is_empty() {
-					let last_arg = constructor_sig.parameters.last().unwrap().maybe_unwrap_option();
-					self.validate_structural_type(&arg_list.named_args, &arg_list_types.named_args, &last_arg, exp);
+				if !has_fields && !arg_list.pos_args.is_empty() {
+					self.expr_error(
+						&arg_list.pos_args[0],
+						format!(
+							"{} should be called with no arguments.",
```
Chriscbr, line 1280: The error message here should be the same as if you specified an `init()` with zero arguments - e.g. `Expected 0 arguments but got 1 when instantiating "A"`

Chriscbr (changes requested):

libs/wingc/src/ast.rs
Chriscbr, line 262: nit: I wonder if we should rename this to `initializer`? (Aim for consistent phrasing across our docs / codebase)

examples/tests/invalid/classes_and_resources.w
```diff
@@ -0,0 +1,7 @@
+class C4 { x:num; } // class with field and no init
+//    ^^ Unknown symbol "init"
```
yoav-steinberg, line 2: This should be more explicit
```suggestion
//    ^^ Missing class/resource initializer (init)
```

libs/wingc/src/type_check.rs
```diff
@@ -1245,11 +1245,15 @@ impl<'a> TypeChecker<'a> {
...
+				let (class_env, class_symbol, has_fields) = match &*type_ {
+					Type::Class(ref class) => {
+						let fields = class.fields(true).into_iter().count();
+						(&class.env, &class.name, if fields > 0 { true } else { false })
+					}
 					Type::Resource(ref class) => {
 						if matches!(env.phase, Phase::Preflight) {
-							(&class.env, &class.name)
+							let fields = class.fields(true).into_iter().count();
+							(&class.env, &class.name, if fields > 0 { true } else { false })
```
yoav-steinberg, line 1256: perhaps better
```suggestion
							let fields = class.fields(true).into_iter().count();
							(&class.env, &class.name, fields > 0)
```
or even
```suggestion
							(&class.env, &class.name, class.fields(true).peekable().peek().is_some())
```

libs/wingc/src/type_check.rs
```diff
@@ -2204,58 +2231,19 @@ impl<'a> TypeChecker<'a> {
...
-				class_type.as_mut_class_or_resource().unwrap().env = class_env;
-				let class_env = &class_type.as_class_or_resource().unwrap().env;
-
-				// Type check constructor
-				let constructor_sig = if let Type::Function(ref s) = *constructor_type {
-					s
-				} else {
-					panic!(
-						"Constructor of {} isn't defined as a function in the class environment",
-						name
```
yoav-steinberg, line 2243: Still not clear why this was removed.

marciocadev:
Now, when creating a class with fields but without an initializer, no error is generated, which is problematic because it should be reported that the field was not initialized.

I will need to validate the initialization of the field in the constructor.

example
```ts
class T {
   x: num;
}
```

Chriscbr:
I see... maybe we need some kind of static analysis here? pinging @yoav-steinberg

marciocadev:
> I see... maybe we need some kind of static analysis here? pinging @yoav-steinberg

I fix this, but I don't know if its the right way to do it

marciocadev:
> I see... maybe we need some kind of static analysis here? pinging @yoav-steinberg

I fixed this, but I'm not sure if it's the correct way to do it.

@yoav-steinberg, could you please check if this is correct?

I listed the assignments inside the initializer, and if one of the class fields is missing from the list, I raise an error that points to the missing field.

But I came across an unusual case in the `resource_captures.w` test where there's an `inflight var inflight_field: num;`.
This field was not initialized in init because inflight initializations are not supported.
The `inflight_fields` now gives an error because all class fields need to be initialized in the constructor.

If a field is created with `var`, should this rule be ignored?

yoav-steinberg:
> > I see... maybe we need some kind of static analysis here? pinging @yoav-steinberg
> 
> @Chriscbr I fixed this, but I'm not sure if it's the correct way to do it.
> 
> @yoav-steinberg, could you please check if this is correct?
> 
> I listed the assignments inside the initializer, and if one of the class fields is missing from the list, I raise an error that points to the missing field.

Just making sure there's an assignment for each fields isn't good enough. We need some static analysis to get this right, and even then I'm not sure we can catch all cases. I was thinking that for now we just ignore this and hope the user doesn't leave any field uninitialized. The same goes for making sure `super` is called when we have a parent.

Supporting a resource/class without an initializer should only be valid for classes with no fields and no parent.

eladb:
@yoav-steinberg any updates on this?

marciocadev:
Guys, this branch is very outdated and I think it's better to create a new branch with this implementation. I will ask @Chriscbr, @yoav-steinberg, and @eladb to review the new branch

I'm going to close this branch.

staycoolcall911:
@marciocadev - sorry this one got left behind. We’ll work with you to get the next branch on the fast lane. Thank you for your patience.

marciocadev:
Don't worry @staycoolcall911 

I have some health issues last month so I can't work on that

staycoolcall911:
Oh, hoping you're feeling better now, @marciocadev !

marciocadev:
I'm fine now, back to work 😁

staycoolcall911:
👏

Issue #1866, opened 2023-03-21
Cannot use spaces in function `as` name


hasanaburayyan:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

```js
new cloud.Function(inflight (s: str): str => {
  tasks.add_task("clean the dishes");
  tasks.add_task("buy dishwasher soap");
  tasks.remove_tasks(tasks.find_tasks_with("clean the").at(0));
  let result = tasks.find_tasks_with("clean the dish");
  assert(result.length == 0);
}) as "test:add, remove and find task";
```

`wing compile -t tf-aws app.w`

then I tried to deploy with `tf apply`

### I expected this:

Successful deployment

### Instead, this happened:

Unable to find assets zip
<img width="1139" alt="image" src="https://user-images.githubusercontent.com/45375125/226513633-f26aae10-0129-4f18-8e74-194cb9a5a403.png">

### Component:

SDK

PR #1867, opened 2023-03-21
fix(sdk): spaces in function names cause problems with terraform assets

hasanaburayyan:
Disallows spaces from asset relative directories as they cause issues when creating TerraformAssets. Fixes issue where we could not find asset zip folder when deploying terraform

Closes: https://github.com/winglang/wing/issues/1866

### DIFF
--- a/libs/wingsdk/src/cloud/function.ts
+++ b/libs/wingsdk/src/cloud/function.ts
@@ -108,12 +108,15 @@ export abstract class Function extends Resource implements IInflightHost {
       implicit: true,
     });
 
-    const assetName = ResourceNames.generateName(this, {
-      // Avoid characters that may cause path issues
-      disallowedRegex: /[><:"/\\|?*]/g,
-      case: CaseConventions.LOWERCASE,
-      sep: "_",
-    });
+    const assetName = join(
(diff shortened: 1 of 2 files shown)


hasanaburayyan:
I forgot to add a test for this, will do in the morning

libs/wingsdk/src/cloud/function.ts
```diff
@@ -110,7 +110,7 @@ export abstract class Function extends Resource implements IInflightHost {
 
     const assetName = ResourceNames.generateName(this, {
       // Avoid characters that may cause path issues
-      disallowedRegex: /[><:"/\\|?*]/g,
+      disallowedRegex: /[><:"/\\|?* ]/g,
```
Chriscbr, line 113: tip: We can disallow all whitespace characters using a character class
```suggestion
      disallowedRegex: /[><:"/\\|?*\s]/g,
```

Chriscbr (approved):

libs/wingsdk/src/cloud/function.ts
hasanaburayyan: Great callout!

PR #1879, opened 2023-03-21
fix: `print`s are not working from inside a resource in simulator

### LINKED ISSUE #1871: `log`s are not working from inside a resource (regression)
### Community Note

### I tried this:

```js
bring cloud;

resource Test {
  init() { }
  inflight test() {
    print("i am inside a resource yey!");
  }
}

let t = new Test();
new cloud.Function(inflight () => { t.test(); }) as "test";
```

### I expected this:

The print to be included in the test:

```
pass ─ print_in_resource.w » root/test
     --> i am inside a resource yey!
```

### Instead, this happened:

Print was emitted outside of the test:

```
i am inside a resource yey!
pass ─ print_in_resource.w » root/test
```

### Component:

SDK

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

This is a regression caused by #1859

eladb:
Fixes https://github.com/winglang/wing/issues/1871 by replacing our hand rolled sandboxing solution with [vm2](https://www.npmjs.com/package/vm2) which takes care of everything for us.

The previous implementation allowed modules that were required from within the VM to escape the VM's context. This resulted in the `console.log` hijack to not carry over to other modules (such as inflight client modules of resources).

## Notes

* Function timeouts are not working. This is a limitation of `vm2` (and generally a limitation of the fact that we are running our inflight functions inside the same process). I don't think this is actually a huge problem right now, but here's an issue to track:
* Stopped hijacking `console.log()` inside inflight functions. It was a hack in the first place, and we decided to completely get rid of `Logger` in a subsequent pull request.


Chriscbr (approved):

MarkMcCulloh:
@eladb Looks like this change causes the sdk unit test time to skyrocket in CI. Are you able to replicate that locally?

eladb:
Yes. I think we need a simple module cache... on it

eladb:
Looking for a less Jacky solution damn!

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -116,31 +112,38 @@ async function runInSandbox(code: string, payload: any, opts: RunCodeOptions) {
...
+    const customRequire = (request: string) => {
+      if (Module.isBuiltin(request)) {
+        return require(request);
+      }
+
+      const filename = customResolve(request);
+      const code = fs.readFileSync(filename, "utf-8");
+      
+      // wrap the module based on https://nodejs.org/api/modules.html#the-module-wrapper
+      const wrapped = `(function(exports, require, module, __filename, __dirname) {${code}});`
```
eladb, line 132: Can we just emit `require(${filename})` so that the entrypoint will behave like any other module?

tools/hangar/__snapshots__/test_corpus/valid/resource_captures.w.ts.snap
```diff
@@ -581,8 +581,8 @@ exports[`wing compile -t tf-aws > proc.ec2c78b051aaf1991ae27a24d33eb54a2c0485ae7
 exports[`wing test -t sim > stdout 1`] = `
 "- Compiling to sim...
 ✔ Compiling to sim...
-array.len=3
-field=hello!
-this.another.first.my_resource:my_string
-pass ─ resource_captures.wsim » root/env0/test"
+pass ┌ resource_captures.wsim » root/env0/test
+     │ array.len=3
```
eladb, line 585: This is basically showing the bug is now resolved

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -41,111 +38,32 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
-            // TODO: Object.freeze this?
+        const vm = new NodeVM({
+          console: "redirect", // we hijack `console.xxx` calls inside the vm
+          require: {
+            external: true,
+            builtin: ["*"], // allow using all node modules
+            context: "sandbox", // require inside the sandbox (addresses #1871)
+          },
+          sandbox: {
+            fetch, // allow using fetch
```
eladb, line 52: @tsuf239 fyi

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -41,111 +38,32 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
-
-  // create a copy of all the globals from our current context.
-  for (const k of Object.getOwnPropertyNames(global)) {
-    try {
-      ctx[k] = (global as any)[k];
-    } catch {
-      // ignore unresolvable globals
-      // see https://github.com/winglang/wing/pull/1923
-    }
-  }
```
MarkMcCulloh, line 114: Does vm2's vm automatically include globals?

PR #1880, opened 2023-03-22
chore(docs): story 15 syntax amended

staycoolcall911:
Some syntax changes hand-compiled by @Chriscbr and myself.


docs/04-reference/wingsdk-spec.md
```diff
@@ -869,8 +869,8 @@ struct ApiRequest {
   method: HttpMethod;
   /** The request's path. */
   path: str;
-  /** The request's query string. */
-  query: str?;
+  /** The request's query parameters. */
+  query: Map<str, str>?;
   /** The path variables. */
   vars: Map<str, str>?;
```
Chriscbr, line 875: Thanks for fixing this. I also noticed there's a mistake since all maps in Wing have `str` keys right now, so it should just be `Map<str>` (my error from when I drafted the spec)

Chriscbr (approved):
Thanks for the fixes - I caught one more if you have a chance to take a look :)

docs/04-reference/wingsdk-spec.md
staycoolcall911: Sure thing - I'll change that across the SDK spec, there are a few places.

PR #1888, opened 2023-03-22
fix(compiler): type checker does not assign all nodes evaluated_type

Chriscbr:
I noticed an issue where the type checker did not assign all nodes an `evaluated_type`. The root cause was that whenever there was a `return` statement inside `type_check_exp`, it would return out of the function immediately, causing the last line (that sets the `evaluated_type`) to be skipped. I fixed this by extracting out the main `match expr.kind` clause to its own helper function.

Misc:
- Refactored how binary expressions are type checked to make it a little clearer that each binary operator case is handled. When new binary operators are added to the `BinaryOp` enum, it forces a compiler error.
- Fixed some bugs that were surfaced where argument types for function calls and constructors weren't always being evaluated.


libs/wingc/src/type_check_assert.rs
```diff
@@ -0,0 +1,13 @@
...
+pub struct TypeCheckAssert;
+
+impl Visit<'_> for TypeCheckAssert {
+	fn visit_expr(&mut self, expr: &Expr) {
+		assert!(
+			expr.evaluated_type.borrow().is_some(),
+			"Expr was not type checked: {:?}",
+			expr
+		);
+	}
```
MarkMcCulloh, line 12: Do you need to also call the global visit_expr to make sure this recurses properly?

MarkMcCulloh (approved):
Nice, this will probably prevent a lot of panics (especially in the language server)

libs/wingc/src/lib.rs
```diff
@@ -273,6 +275,12 @@ pub fn compile(source_path: &Path, out_dir: Option<&Path>) -> Result<CompilerOut
 		Diagnostics::new()
 	};
 
+	// Validate that every Expr has an evaluated_type
+	if cfg!(debug_assertions) {
```
eladb, line 279: I hope this is always on

eladb (commented):
This is going to address so many panics. 

Is there a way to make evaluated_type a non-optional type? So consumers won't need to blindly unwrap it?

libs/wingc/src/type_check_assert.rs
Chriscbr: Took a little bit of work to add this because it surfaced some more bugs where types weren't being evaluated. 😅

libs/wingc/src/lib.rs
Chriscbr: Right now it's only on during dev (local and CI builds). Should we also surface these in our release build?

libs/wingc/src/type_check.rs
```diff
@@ -1124,7 +1124,19 @@ impl<'a> TypeChecker<'a> {
 		statement_idx: usize,
 		context: &TypeCheckerContext,
 	) -> TypeRef {
-		let t = match &exp.kind {
+		let t = self.type_check_exp_helper(&exp, env, statement_idx, context);
+		exp.evaluated_type.replace(Some(t));
+		t
+	}
+
+	fn type_check_exp_helper(
```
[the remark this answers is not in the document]
Chriscbr: I think the `type_check_exp_helper` name emphasizes that nobody else should be calling it except for `type_check_exp`. I'll add a comment to clarify. (It could have also been an inner function / closure, but I wanted to avoid the extra nesting)

libs/wingc/src/lib.rs
eladb: I'd add to release builds too

PR #1890, opened 2023-03-22
fix: hierarchal symbol lookups in language server sometimes fail

### LINKED ISSUE #1644: Memory corruption in SymbolEnv during post-compilation traversal

MarkMcCulloh:
Full credit to @yoav-steinberg for finding the root cause/solution.

While there is a more systemic issue at play here (the unsafe pointers), I would call this change a workaround rather than a hack (which is what I was doing before).

Fixes #1644

### DIFF
--- a/libs/wingc/src/lsp/sync.rs
+++ b/libs/wingc/src/lsp/sync.rs
@@ -20,7 +20,7 @@ pub struct FileData {
 	/// The diagnostics returned by wingc
 	pub diagnostics: Diagnostics,
 	/// The top scope of the file
-	pub scope: Scope,
+	pub scope: Box<Scope>,
 	/// The universal type collection for the scope. This is saved to ensure references live long enough.
 	pub types: Types,
 }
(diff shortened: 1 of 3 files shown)


yoav-steinberg (approved):

libs/wingc/src/lsp/sync.rs
```diff
@@ -20,7 +20,7 @@ pub struct FileData {
 	/// The diagnostics returned by wingc
 	pub diagnostics: Diagnostics,
 	/// The top scope of the file
-	pub scope: Scope,
+	pub scope: Box<Scope>,
```
Chriscbr, line 23: IIUC this might be a possible use case for `Pin<Box<Scope>>`

Chriscbr (approved):

libs/wingc/src/lsp/sync.rs
yoav-steinberg: I think you are right, but I suspect all our `UnsafeRef`s are a use case for `Pin<>` and I'm also not sure it'll compile since I don't think you can pin stuff in a `RefCell`. I'd prefer to just refactor everything to get rid of `UnsafeRef`, but not sure we have time for that now.

PR #1898, opened 2023-03-23
feat!: rename `print` to `log`

Chriscbr:
Logs are a first class concern in building distributed cloud applications, so we would like to rename the global function `print` to `log` to make the intent of the API clearer (in both preflight and inflight phases). See https://github.com/winglang/wing/discussions/1772 for more details.

This PR updates the Language spec as the compiler to use the new function name.

BREAKING CHANGE: `print()` is now named `log()`


tools/hangar/__snapshots__/plugins.ts.snap
```diff
@@ -171,7 +171,7 @@ exports[`Plugin examples > AWS target plugins > permission-boundary.js 1`] = `
   "terraform": {
     "backend": {
       "local": {
-        "path": "/Users/hasan/repos/monada/wing/tools/hangar/tmp/terraform.root.tfstate",
+        "path": "/Users/chrisr/dev/wing3/tools/hangar/tmp/terraform.root.tfstate",
```
MarkMcCulloh, line 174: whoopsie, didn't catch this in the last pr

@hasanaburayyan

MarkMcCulloh (approved):

MarkMcCulloh:
Probably need to double-check the [docsite](https://github.com/winglang/docsite) to make sure there is no invalid examples floating around. Same for the console

tools/hangar/__snapshots__/plugins.ts.snap
Chriscbr: This should be caught by CI, since the GitHub runner would probably generate a snapshot with a different value, no?

MarkMcCulloh: 😢 https://github.com/winglang/wing/blob/ecd00497bae6d6c6cb2ac5c88a79b67388cafbf8/tools/hangar/vitest.config.ts#L8

I'll do a cleanup PR, it's fine to merge this one

PR #1900, opened 2023-03-23
chore: snapshot updates not failing ci

MarkMcCulloh:
Several whoopsies at once:

- Vitest config accidentally had `update: true`, so snapshots always updated instead of failing
- The recent starts_with and ends_with change didn't work, but this was considered a snapshot change so it was unnoticed. Added additional code to hangar to always fail if there's stderr when unexpected
- Snapshots had an unsanitized path for the new plugin tests, causing constant change
- Snapshots we're actually being overly-sanitized in many cases, causing chunks to be deleted from the terraform output
- During `main` builds, things were being published before e2e tests were run

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -240,7 +240,7 @@ jobs:
     name: Publish
     if: ${{ github.event_name == 'push' && github.repository == 'winglang/wing' }}
     needs:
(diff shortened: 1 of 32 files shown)


Chriscbr (approved):
🙏 🙏

PR #1901, opened 2023-03-23
feat(spec): even more cloud bucket apis

Chriscbr:
Proposing some more APIs for making cloud buckets even easier to work with. Based on my googling I think it should be possible to support all of these capabilities across the major cloud providers.

One API worth highlighting is `metadata`, which returns information about an individual file/object, like it's size, last modification date, and so on. It turns out there's a lot of possible custom metadata fields stored by cloud providers that we could show. I cherry-picked a few fields that seemed most useful / least hard to simulate but I imagine we can support more in the future.

References:
- Blob Properties docs for Azure: https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=azure-ad
- GCP object metadata docs: https://cloud.google.com/storage/docs/viewing-editing-metadata#storage-view-object-metadata-nodejs
- AWS S3 HeadObject docs: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_ResponseSyntax

### DIFF
--- a/docs/04-reference/wingsdk-spec.md
+++ b/docs/04-reference/wingsdk-spec.md
@@ -292,13 +292,53 @@ resource Bucket {
 
   /**
    * Get a file from the bucket.
+   *
+   * @throws Will throw if the file doesn't exist.
    */
   inflight get(key: str): Blob;
 
   /**
+   * Get a file from the bucket if it exists.
+   */
+  inflight try_get(key: str): Blob?;
+
+  /**
    * Delete a file from the bucket.
+   *
+   * @throws Will throw if the file doesn't exist.
    */
   inflight delete(key: str): void;
+
+  /**
+   * Delete a file from the bucket if it exists.
+   */
+  inflight try_delete(key: str): bool;
+
+  /**
+   * Check if a file exists in the bucket.
+   */
+  inflight exists(key: str): bool;
+
+  /**
+   * Get the metadata of a file in the bucket.
+   */
+  inflight stats(key: str): FileMetadata;
+
+  /**
+   * Move a file to a new location in the bucket. If the destination file already
+   * exists, it will be overwritten.
+   *
+   * @throws Will throw if the `src` file doesn't exist.
+   */
+  inflight move(src: str, dst: str): void;
+
+  /**
+   * Copy a file to a new location in the bucket. If the destination file already
+   * exists, it will be overwritten.
+   *
+   * @throws Will throw if the `src` file doesn't exist.
+   */
+  inflight copy(src: str, dst: str): void;
 
   /**
    * List all files in the bucket with the given prefix.
@@ -306,8 +346,9 @@ resource Bucket {
   inflight list(prefix: str?): Iterator<str>;
 
   /**
-   * Returns a url to the given file.
-   * @throws Will throw if the file is not public.
+   * Returns a url to the given file key. Does not check if the file exists.
+   *
+   * @throws Will throw if the bucket is not public.
    */
   inflight public_url(key: str): str;
 
@@ -316,6 +357,15 @@ resource Bucket {
    * access the file until the link expires (defaults to 24 hours).
    */
   inflight signed_url(key: str, duration?: duration): str;
+}
+
+struct FileMetadata {
+  /** The size of the file in bytes. */
+  size: Size;
+  /** The time the file was last modified. */
+  last_modified: Date; // or an ISO timestamp `str` until we have Date API support
+  /** The content type of the file, if it is known. */
+  content_type: str?;
 }
 
 struct BucketOnUploadProps { /* elided */ }


docs/04-reference/wingsdk-spec.md
```diff
@@ -292,22 +292,63 @@ resource Bucket {
...
+
+  /**
+   * Check if a file exists in the bucket.
+   */
+  inflight exists(key: str): bool;
+
+  /**
+   * Get the metadata of a file in the bucket.
+   */
+  inflight stats(key: str): FileMetadata;
```
Chriscbr, line 325: I thought `stat`, `stats`, `metadata` could all be good names, I'm curious which sounds best to other peoples' ears?

docs/04-reference/wingsdk-spec.md
```diff
@@ -292,22 +292,63 @@ resource Bucket {
 
   /**
    * Get a file from the bucket.
+   *
+   * @throws Will throw if the file doesn't exist.
```
Chriscbr, line 296: It's worth noting, even though these methods specifically throw for functional reasons, **methods without `@throws` can still throw because they're networked APIs**, and pretty much every API call made on the internet is a best effort attempt. Just something to think about for the future

docs/04-reference/wingsdk-spec.md
```diff
@@ -318,6 +359,15 @@ resource Bucket {
   inflight signed_url(key: str, duration?: duration): str;
 }
 
+struct FileMetadata {
+  /** The size of the file in bytes. */
+  size: Size;
```
Chriscbr, line 364: I was imagining `Size` as a standard type, similar to the one in the CDK ([link](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.Size.html)). I wonder if this should be a primitive type like `duration`? 🤷

MarkMcCulloh: I definitely think we should have size literals, would be real nice

hasanaburayyan: Oh that would be cool

docs/04-reference/wingsdk-spec.md
```diff
@@ -318,6 +359,15 @@ resource Bucket {
...
+  /** The time the file was last modified. */
+  last_modified: Date; // or an ISO timestamp `str` until we have Date API support
+  /** The content type of the file, if it is known. */
+  content_type: str?;
```
hasanaburayyan, line 368: We could also return version number if versioning is enabled

docs/04-reference/wingsdk-spec.md
```diff
@@ -292,22 +292,63 @@ resource Bucket {
...
+   */
+  inflight stats(key: str): FileMetadata;
+
+  /**
+   * Move a file to a new location in the bucket. If the destination file already
+   * exists, it will be overwritten.
+   *
+   * @throws Will throw if the `src` file doesn't exist.
+   */
+  inflight move(src: str, dst: str): void;
```
hasanaburayyan, line 333: What about calling this rename? When I think of move api it makes me think that I could specify the destination as a directory. move(“some_key”, “a/“)

docs/04-reference/wingsdk-spec.md
hasanaburayyan: I personally like metadata

docs/04-reference/wingsdk-spec.md
ekeren: We should consider having semantic versions as a primitive in the language (`v12.0.1`)

docs/04-reference/wingsdk-spec.md
ekeren: Hard question.

Coming from programming in linux, [stats](https://linux.die.net/man/2/stat) feels natural for me.
But I don't think it will be the background of the 99% 

From some reason AWS is using head-object api

docs/04-reference/wingsdk-spec.md
Chriscbr: Good ideas. Since the spec doesn't mention versioning I'll leave it out for now but definitely sounds like a natural place to fit it in the future 👍

docs/04-reference/wingsdk-spec.md
eladb: Paging @benkehoe 👋 I think `stat` sends a message that buckets are file systems and that's probably not the [best mental model](https://twitter.com/ben11kehoe/status/1635703838865870850?s=46&t=vQsWtya6IrJ2YERDZL4eYA).

I would go with `metadata` or something from that family.

eladb (approved):
Good stuff

docs/04-reference/wingsdk-spec.md
Chriscbr: Yes I was also thinking about updating some of the messaging here to switch out references to "files" with "objects"

Issue #120, opened 2022-09-20
Struct Declaration


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#31-structs

### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Declare struct in wing language that can contain only immutable values from types that are legal Json: `bool, str, num, Array<T>, Json, Struct` and their optionals where `T` can be all of the above 

```ts (wing)
struct Point {
   x: num;
   y: num; 
}

struct MyStruct {
 n: num;
 s: str?;
 ar_of_string: Array<str>?;
 b: bool?;
// struct can contain another struct
 another_struct: point?;
 arrays_of_points: Array<Array<Point>>?;
// can also contain a json
 j: Json?;
} 
``` 

struct can be created using the following:
```ts (wing)
// Initialising fields 
let p1 = Point { x: 0.2, y:-0.2 } ; 
// using spread
let p2 = Point {x: 0.2, ...p1 };
// specifying only non optional fields
let my_struct = MyStruct { b: true };
// Initializing another struct inline
let my_struct2 = MyStruct { another_struct: Point { x:1 , y:-1 }, ...my_struct }; 
```

struct can be converted with the following API: 
```ts (wing)
let p = Point { x: 0.2, y:-0.2 } ; 
// you can implicitly cast a struct into a Json 
let j: Json = p;
// you can use the Json.stringify + implicitly conversion tom stringify a struct
let  s: str = Json.stringify(p);  
// you can parse a using the from_json static method  T.from_json = (Json):T
let p_from_json = Point.from_json(j); // p_from_json equals p
// you can also parse a string into a struct instance using the T.parse = (str):T function
let p_from_str1 = Person.parse(""{\"x\":0.2,\"y\": -0.2}"); 
```

### Use Cases

a strongly typed json

eladb:
I suspect this is P1 for beta

staycoolcall911:
@eladb - ok

ekeren:
Because Json and MutJson are P1, we can manage this as P2

staycoolcall911:
We decided to move this to p1 - without `struct`s it is impossible to define resource props for user-defined resources.

monadabot:
Congrats! :rocket: This was released in [Wing 0.8.0](https://github.com/winglang/wing/releases/tag/v0.8.0).

[...]

PR #1904, opened 2023-03-24
feat(compiler): struct declarations

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/120


libs/wingc/src/parser.rs
```diff
@@ -332,6 +332,43 @@ impl<'s> Parser<'s> {
...
+					.expect("struct definition fields should always have a name"),
+			)?;
+			let type_ = field_node
+				.child_by_field_name("type")
+				.expect("struct definition fields should always have explicit type");
+			let f = ClassField {
+				name: identifier,
+				member_type: self.build_type_annotation(&type_)?,
+				reassignable: false,
+				phase: Phase::Preflight,
```
MarkMcCulloh, line 358: Is it preflight? I feel like the fields themselves may be phase independent, not really sure though.

libs/wingc/src/type_check.rs
```diff
@@ -836,6 +836,20 @@ impl TypeRef {
...
+	pub fn is_deep_mutable(&self) -> bool {
+		match &**self {
+			Type::MutArray(_) => true,
+			Type::MutMap(_) => true,
+			Type::MutSet(_) => true,
+			Type::MutJson => true,
+			Type::Array(v) => v.is_deep_mutable(),
+			Type::Map(v) => v.is_deep_mutable(),
+			Type::Set(v) => v.is_deep_mutable(),
+			_ => false,
```
MarkMcCulloh, line 849: Optionals?

Maybe it's worthwhile for this to be exhaustive anyways

libs/tree-sitter-wing/grammar.js
```diff
@@ -108,9 +108,9 @@ module.exports = grammar({
       seq(
         "struct",
         field("name", $.identifier),
-        optional(seq("extends", commaSep($.identifier))),
+        optional(seq("extends", commaSep(field("extend", $.identifier)))),
```
MarkMcCulloh, line 111: nit: I feel like this should be plural, both to match the keyword and also to better signify that there can be many of them.

```suggestion
        optional(seq("extends", commaSep(field("extends", $.identifier)))),
```

libs/wingc/src/parser.rs
```diff
@@ -332,6 +332,43 @@ impl<'s> Parser<'s> {
 		})
 	}
 
+	fn build_struct_definition_statement(&self, statement_node: &Node) -> DiagnosticResult<StmtKind> {
+		let name = self.node_symbol(
+			&statement_node
+				.child_by_field_name("name")
+				.expect("struct definitions should always have a name"),
```
MarkMcCulloh, line 339: nit: There's a helper method that can help avoid the expect

```suggestion
			&self.get_child_field(&statement_node, "name")?
```

MarkMcCulloh (approved):
I'm slightly concerned about which phase the fields should be, so maybe that needs to be double checked, but I otherwise think this looks great!

libs/wingc/src/type_check.rs
hasanaburayyan: Great catch!

libs/wingc/src/parser.rs
hasanaburayyan: That is fucking wonderful! Thanks for the nit!

Chriscbr:
Sweet!! 😄

PR #1902, opened 2023-03-24
chore(repo): generate actual release notes

MarkMcCulloh:
The current release notes suck for various reasons, let's improve/simplify them.

- Properly filters out bots from the contributor list
- Handles breaking changes without dealing with labels (since we have a 0 major version number, it bumps minor)
- Show proper types for changes
- Simplifies the workflow (removed the `prepare` job), opening up options to improve our github actions even more
- Provides a skeleton to create completely custom changelogs

Uses https://github.com/unjs/changelogen which is a really nice lib in a great ecosystem. See [here](https://github.com/unjs/changelogen/releases) for example changelogs.

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Monada Contribution License](https://docs.winglang.io/terms-and-policies/contribution-license.html)*.

PR #1906, opened 2023-03-24
chore: feats should only increment patch version by default

MarkMcCulloh:
Regression from https://github.com/winglang/wing/pull/1902 (thanks @Chriscbr)

If it's a breaking change, then it'll increment minor

### DIFF
--- a/scripts/changelog.mjs
+++ b/scripts/changelog.mjs
@@ -36,6 +36,10 @@ async function getData() {
         title: "🚀 Features",
         semver: "patch",
       },
+      revert: {
+        title: "⏪ Reverts",
+        semver: "patch",
+      },
     },
   });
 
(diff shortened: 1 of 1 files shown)


scripts/changelog.mjs
```diff
@@ -32,6 +32,10 @@ async function getData() {
         title: "👓 RFC",
         semver: "patch",
       },
+      feat: {
+        title: "🚀 Features",
+        semver: "patch",
+      },
```
Chriscbr, line 38: Maybe we should add the other types from https://github.com/winglang/wing/blob/main/.github/workflows/pull-request-lint.yml while we're at it?

Chriscbr (approved):

scripts/changelog.mjs
MarkMcCulloh: The rest are built-in, other than revert which I went ahead and added

PR #1916, opened 2023-03-27
fix: playground bundling issue

eladcon:
This invokes the playground release workflow on releases from this repo. This also resolves a small issue making sure webpack is not compiling the SDK as part of the CLI (the SDK is compiled separately).


.github/workflows/build.yml
```diff
@@ -121,6 +121,39 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: winglang-webpack
+          path: apps/wing/winglang-webpack.tgz
+
+      - name: Upload Wing SDK Webpack
+        uses: actions/upload-artifact@v3
+        with:
+          name: winglang-sdk-webpack
+          path: libs/wingsdk/winglang-sdk-webpack.tgz
```
MarkMcCulloh, line 155: I think this re-packaging process belongs on the consumer side rather than in this repo.

The playground project should instead pull down from npmjs and do this process during its own build.

MarkMcCulloh (changes requested):

.github/workflows/build.yml
eladcon: Changed the release trigger to be based on NPM

MarkMcCulloh (approved):

PR #1923, opened 2023-03-28
fix: unavailable global property crashes simulator

### LINKED ISSUE #1912: testing programmatically doesn't work
### Community Note

### I tried this:

Following instructions https://docs.winglang.io/getting-started/simulator

### I expected this:

to work

### Instead, this happened:

the function didn't run when the queue message got sent 

### Component:

Compiler

### Wing Version:

0.8.3

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

probably related to latest method wrapping solution

MarkMcCulloh:
When copying global for the new sim context, accessing `global.wasi` triggered a `require` that fails. This only seems to happens in a REPL 🤷 

In any case, a corrupted global obj probably isn't worth crashing over. I'm not even sure it's worth logging, feel free to let me know if you do.

Fixes #1912

### DIFF
--- a/libs/wingsdk/src/target-sim/function.inflight.ts
+++ b/libs/wingsdk/src/target-sim/function.inflight.ts
@@ -109,6 +109,7 @@ async function runInSandbox(code: string, payload: any, opts: RunCodeOptions) {
       ctx[k] = (global as any)[k];
     } catch {
       // ignore unresolvable globals
+      // see https://github.com/winglang/wing/pull/1923
     }
   }
 
--- a/libs/wingsdk/src/target-sim/function.inflight.ts
+++ b/libs/wingsdk/src/target-sim/function.inflight.ts
@@ -105,7 +105,11 @@ async function runInSandbox(code: string, payload: any, opts: RunCodeOptions) {
 
   // create a copy of all the globals from our current context.
   for (const k of Object.getOwnPropertyNames(global)) {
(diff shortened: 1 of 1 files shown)


ekeren:
Mark, can you explain how you debugged it?

Also (and related), is there anything we can do that will generate visible errors when something like that happens?

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -105,7 +105,11 @@ async function runInSandbox(code: string, payload: any, opts: RunCodeOptions) {
 
   // create a copy of all the globals from our current context.
   for (const k of Object.getOwnPropertyNames(global)) {
-    ctx[k] = (global as any)[k];
+    try {
+      ctx[k] = (global as any)[k];
+    } catch {
+      // ignore unresolvable globals
```
eladb, line 111:
```suggestion
      // ignore unresolvable globals
      // see https://github.com/winglang/wing/pull/1923
```

eladb (approved):

MarkMcCulloh:
@ekeren  I looked at the simulator traces (`simulator.listTraces()`) which showed the error.

We would probably need to add something that listens for traces and prints them to make them more visible. You can add your own listener by doing:

```js
  simulator.onTrace({
    callback: (trace: Trace) => {
      console.log(trace);
    },
  });
```

PR #1924, opened 2023-03-28
fix(cli): compile command does not write to target directory in an atomic way

### LINKED ISSUE #1919: .tfaws folder disappears?
### Community Note

### I tried this:

1. I wrote a .w file in `/examples/tests/valid` 
2. I compiled using `npx nx wing -- compile -t tf-aws ../../examples/tests/valid/bucket_events.w`
3. I run terraform `init` and `apply` in `/examples/tests/valid`
4. then compiled again using the command above

### I expected this:

terraform apply to work

### Instead, this happened:

image

**it did work again when getting out and into the folder, tried both in vs code terminal and in the default mac terminal, BUT I had to use `terraform init` first**

image

### Component:

Compiler, SDK

### Wing Version:

15.6.3

### Node.js Version:

v19.7.0

### Platform(s):

MacOS

### Anything else?

_No response_

ainvoner:
Fixes #1919


Chriscbr (commented):
This looks good to me, but copying @MarkMcCulloh to review as well just in case there's anything I'm missing

Chriscbr:
@ainvoner Updated the description so it closes #1919 (I think this would fix it?) cc @hasanaburayyan

apps/wing/src/commands/compile.ts
```diff
@@ -249,6 +251,14 @@ export async function compile(entrypoint: string, options: ICompileOptions): Pro
     return process.exit(1);
   }
 
+  // clean up before
+  await Promise.all([
+    rm(synthDir, { recursive: true, force: true }),
+    mkdir(synthDir, { recursive: true }),
+  ]);
```
MarkMcCulloh, line 258: Is there a race condition here? maybe it's impossible, but what if mkdir runs before rm?

```suggestion
  // clean up before
  await rm(synthDir, { recursive: true, force: true });
  await mkdir(synthDir, { recursive: true });
```

MarkMcCulloh (approved):

apps/wing/src/commands/compile.ts
```diff
@@ -67,39 +68,40 @@ function resolveSynthDir(outDir: string, entrypoint: string, target: Target) {
...
  */
 export async function compile(entrypoint: string, options: ICompileOptions): Promise<string> {
+  // create a unique temporary directory for the compilation
+  const tmpTargetdir = await mkdtemp(join(os.tmpdir(), "wing-"));
   const targetdir = join(dirname(entrypoint), "target");
   const wingFile = entrypoint;
   log("wing file: %s", wingFile);
   const wingDir = dirname(wingFile);
   log("wing dir: %s", wingDir);
+  const tmpSynthDir = resolveSynthDir(tmpTargetdir, wingFile, options.target);
```
eladb, line 78: Intuitively I feel that perhaps we would want to create the temporary directory still under `target/xxx` (maybe `target/xxx.wsim.tmp`), so that npm dependencies will resolve relative to the entrypoint. @Chriscbr @ainvoner WDYT?

Chriscbr: Sounds like a good callout, definitely might prevent some nasty bugs in the future. 👍

Chriscbr: Looks like this will be fixed in #1929

PR #1926, opened 2023-03-28
feat: `wing test -t tf-aws`

### LINKED ISSUE #1368: `wing test --target` (tracking issue)
### Community Note

### Feature Spec

The Wing `test` command now features a way to run integration tests live in the cloud, through `wing test --target`. The `--target` flag lets you sanity check that the same tests that you have written to validate your app works through the Wing simulator also pass when they're compiled and executed on the cloud.

Currently, `wing test`(without any flags) will run a sequence of tests based on your application code using the Wing simulator. It starts up the Wing simulator, looks for `cloud.Function` instances that have been labeled with `test:` and , validating that the function does not throw and all assertions are passing.

Here is an example output of the `wing test --target`
```sh
% wing test --target tf-aws --verbose a.w

Info: Compiling to tf-aws
Info: found 25 tests
Info: Setting up 25 isolated environments 
Info: terraform init
Info: terraform apply 
Info: Establishing connection to test runner 
Info: Opening websocket with test reporter
Info: Running tests
pass ┌ a.w » root/test: a
     └ before push counter is 0
pass ┌ a.w » root/test: b
     │ doing something
     │ Lorem ipsum dolor sit amet
     │ consectetur adipiscing elit, 
     └ sed do eiusmod tempor incididunt ut
Info: Cleanup
```

As you can notice in the above output, the `wing test` command takes care of the following: 
- compiling
- deploying
- isolated environment setup (see Test isolation [section](/#Isolation))
- deploying  (for tf-aws it is terraform init and apply) 
- running tests
- gathering and displaying tests results
- cleanup

## Isolation 
Currently the `wing test` on sim implementation creates a simulator for every test, this creates an optimal situation for 
test isolation. On the cloud this means that the `wing test --target tf-aws` will mimic the same situation by creating isolated environment by duplicating them, which might be time consuming, surprising to end users and expensive (every test duplicates the entire infrastructure) 

Although test isolation is important, it is more important that:
<h3 align="center" >The only reason for a test to have a different result on different target is a bug in the SDK</h3>

### Use Cases

Letting SDK contributors and library authors validate that "if it works on my machine, it also works in the cloud"

Letting Wing users run integration tests for local development and in their CI/CD, and test applications that use resources without a corresponding implementation for the Wing simulator.

### Implementation Notes

_No response_

### Component

_No response_

Chriscbr:
Adds support for testing apps against live cloud applications by running `wing test -t tf-aws`.

https://user-images.githubusercontent.com/5008987/229242303-53f65413-65bd-452c-b542-a56db0aa33a6.mov

In order to support this, a new `cloud.TestRunner` resource was added to the SDK. This resource doesn't represent any physical cloud resources itself, but it contains an inflight API for listing and running tests in the current app. The `cloud.TestRunner` resource is automatically added to the root of each app, just like `cloud.Logger`.

In order to ensure the tests behave the same in the cloud as in the simulator, when Wing source code is compiled as part of `wing test`, an isolated copy of the user's app is generated for each cloud test. (This is the same as how the simulator currently operates today). In each isolated environment, one test gets executed. The artifacts of this test variant of the application are synthesized into `target/test/`.

This model basically assumes the root of your application functions as the "setup" for each test. In the future, we can add support for sharing resources between tests through new extensions to the language or SDK.

Known issues / future work:
- `tf-gcp`, `tf-azure`, and `awscdk` targets are not supported
- Traces are not supported on the `tf-aws` target, so any logs generated by tests will not be emitted to the CLI if you test with this target.
- Loading spinner doesn't spin while compiling the app because compilation uses the main JavaScript thread
- Tests on AWS are currently invoked by making AWS API calls directly from your machine -- this means running `wing test` requires having a profile set up not only with permissions to deploy your app to AWS, but also to call any AWS Lambda functions
- We currently auto-approve all Terraform changes, and anything logged by Terraform is not emitted. Users may want to configure these.
- Test durations are not recorded.
- We've added `force_destroy` to all buckets so that integration tests can be cleaned up -- we will remove this once we have a mechanism to tell the SDK whether we are compiling in "test" mode or not.
- We currently create each test function in every test environment, even though we are only going to run one test per environment.

Closes #1368


apps/wing/src/commands/test.ts
ekeren, line 1: @Chriscbr , please plan with @polamoros how we don't break console<>wing compatibility on this

Chriscbr: I've made some progress with the PR, for now I'm aiming to make the PR fully backwards compatible with respect to the Simulator API in the SDK that I reckon is used in the console. The main breaking change however is that `wing test` command in the CLI now requires a target to be passed - I don't think we can avoid this one.

@skyrpex @polamoros does the Wing console invoke `wing test` directly? If so we might need to prepare a change in the console before we release this

libs/wingsdk/test/testing/__snapshots__/simulator.test.ts.snap
```diff
@@ -192,62 +155,16 @@ exports[`run single test > happy path 1`] = `
...
-    at Handler.handle (evalmachine.<anonymous>:20:7)
-    at Object.exports.handler (evalmachine.<anonymous>:24:5)
-    at evalmachine.<anonymous>:26:9
-    at Script.runInContext (node:vm:141:12)
-    at Object.runInContext (node:vm:297:6)
-<sanitized>
-    at new Promise (<anonymous>)
-<sanitized>
-<sanitized>
-<sanitized>",
```
Chriscbr, line 229: Possible regression

apps/wing/src/cli.ts
```diff
@@ -64,6 +64,12 @@ async function main() {
       "Compiles a Wing program and runs all functions with the word 'test' or start with 'test:' in their resource identifiers"
     )
     .argument("<entrypoint...>", "all entrypoints to test")
+    .addOption(
+      new Option("-t, --target <target>", "Target platform")
+        .choices(["tf-aws", "sim"])
+        .makeOptionMandatory()
```
eladb, line 70: I think defaulting to sim still makes sense here.

apps/wing/src/commands/compile.ts
```diff
@@ -42,24 +42,33 @@ const DEFAULT_SYNTH_DIR_SUFFIX: Record<Target, string | undefined> = {
...
+function resolveSynthDir(outDir: string, entrypoint: string, target: Target, testing: boolean = false, tmp: boolean = false) {
   const targetDirSuffix = DEFAULT_SYNTH_DIR_SUFFIX[target];
   if (!targetDirSuffix) {
     throw new Error(`unsupported target ${target}`);
   }
   const entrypointName = basename(entrypoint, ".w");
   const tmpSuffix = tmp ? `.${Date.now().toString().slice(-6)}.tmp` : "";
-  return join(outDir, `${entrypointName}.${targetDirSuffix}${tmpSuffix}`);
+  if (testing) {
+    return join(outDir, "test", `${entrypointName}.${targetDirSuffix}${tmpSuffix}`);
```
eladb, line 68: DRY the last component of the path

apps/wing/src/commands/compile.ts
```diff
@@ -68,23 +77,26 @@ function resolveSynthDir(outDir: string, entrypoint: string, target: Target, tmp
...
-  const synthDir = resolveSynthDir(targetdir, wingFile, options.target);
+  const synthDir = resolveSynthDir(targetdir, wingFile, options.target, testing);
   log("synth dir: %s", synthDir);
   const workDir = resolve(tmpSynthDir, ".wing");
   log("work dir: %s", workDir);
 
   process.env["WING_SYNTH_DIR"] = tmpSynthDir;
   process.env["WING_NODE_MODULES"] = resolve(join(wingDir, "node_modules"));
   process.env["WING_TARGET"] = options.target;
+  process.env["WING_TEST"] = testing.toString();
```
eladb, line 99:
```suggestion
  process.env["WING_IS_TEST"] = testing.toString();
```

apps/wing/src/commands/test.ts
```diff
@@ -91,12 +119,155 @@ async function testOne(entrypoint: string) {
...
+  }
+
+  await withSpinner("terraform init", () => terraformInit(synthDir));
+
+  await checkTerraformStateIsEmpty(synthDir);
+
+  await withSpinner("terraform apply", () => terraformApply(synthDir));
+
+  const [testRunner, tests] = await withSpinner("Setting up test runner...", async () => {
+    const testArns = await terraformOutput(synthDir, "WING_TEST_RUNNER_FUNCTION_ARNS");
```
eladb, line 161: Put in const

apps/wing/src/commands/test.ts
```diff
@@ -105,3 +276,39 @@ function sortTests(a: sdk.testing.TestResult, b: sdk.testing.TestResult) {
...
+    cwd: options.cwd,
+    maxBuffer: MAX_BUFFER,
+  });
+  if (stderr) {
+    throw new Error(stderr);
+  }
+  return stdout;
+}
+
+async function withSpinner<T>(
```
eladb, line 298: Maybe put in util.ts?

libs/wingc/src/jsify.rs
```diff
@@ -135,26 +149,47 @@ impl<'a> JSifier<'a> {
...
-				format!(
-					"super({{ outdir: $outdir, name: \"{}\", plugins: $plugins }});\n",
-					self.app_name
-				),
-			);
-			output.push(format!(
-				"class MyApp extends {} {{\nconstructor() {}\n}}",
-				TARGET_APP,
-				JSifier::render_block(js)
+			let mut root_class = CodeMaker::default();
```
eladb, line 160: Thank god!

libs/wingc/src/jsify.rs
```diff
@@ -135,26 +149,47 @@ impl<'a> JSifier<'a> {
...
+			app_wrapper.open(format!("class {} extends {} {{", APP_CLASS, APP_BASE_CLASS));
+			app_wrapper.open("constructor() {");
+			app_wrapper.line(format!(
+				"super({{ outdir: {}, name: \"{}\", plugins: $plugins }});",
+				OUTDIR_VAR, self.app_name
 			));
-			output.push(format!("new MyApp().synth();"));
+			app_wrapper.open(format!("if ({}) {{", ENV_WING_TEST));
+			app_wrapper.line(format!("new {}(this, \"env0\");", ROOT_CLASS));
+			app_wrapper.line("const $test_runner = this.testRunner;".to_string());
```
eladb, line 177: Perhaps there's a way to move more of this logic to the sdk instead of generate it per app?

libs/wingc/src/jsify/code.rs
```diff
@@ -0,0 +1,126 @@
+#[derive(Default)]
+pub struct CodeMaker {
+	lines: Vec<(usize, String)>,
+	indent: usize,
+}
+
+impl CodeMaker {
+	pub fn open<S: Into<String>>(&mut self, line: S) {
```
eladb, line 8: It's useful to also support open_block() and close_block()

libs/wingc/src/jsify/code.rs
```diff
@@ -0,0 +1,126 @@
...
+	pub fn indent(&mut self) {
+		self.indent += 1;
+	}
+}
+
+impl ToString for CodeMaker {
+	fn to_string(&self) -> String {
+		let mut code = String::new();
+		for (indent, line) in &self.lines {
+			code.push_str(&"\t".repeat(*indent));
```
eladb, line 50: Use two spaces instead of tab, which is the common style in JavaScript

libs/wingc/src/jsify/code.rs
```diff
@@ -0,0 +1,126 @@
+#[derive(Default)]
```
eladb, line 1: Maybe call this `codemaker.rs`

libs/wingsdk/src/cloud/test-runner.ts
```diff
@@ -0,0 +1,174 @@
...
+  ): TestRunner {
+    return App.of(scope).newAbstract(TEST_RUNNER_FQN, scope, id, props);
+  }
+
+  /**
+   * Returns whether a construct represents a runnable test.
+   * @param c A construct.
+   * @returns Whether the construct is a test.
+   */
+  public static isTest(c: IConstruct): boolean {
```
eladb, line 40: Maybe in a subsequent PR, but I'd like us to change how tests are modeled in apps.

Instead of cloud.Functions with special IDs, let's create a cloud.Test resource and only materialize it into a Function in the environment in which it will be executed.

This means that only a single test cloud function will be defined in each environment.

libs/wingsdk/src/cloud/test-runner.ts
```diff
@@ -0,0 +1,174 @@
...
+   */
+  LOG = "log",
+}
+
+/**
+ * List of inflight operations available for `TestRunner`.
+ * @internal
+ */
+export enum TestRunnerInflightMethods {
+  /** `TestRunner.runTest` */
```
eladb, line 170: Annotations?

libs/wingsdk/src/target-sim/test-runner.inflight.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  public async cleanup(): Promise<void> {
+    return;
+  }
+
+  public async listTests(): Promise<string[]> {
+    return Array.from(this.tests.keys());
+  }
+
+  public async runTest(path: string): Promise<TestResult> {
```
eladb, line 30: Do we really need a different implementation for this between aws and the simulator? Intuitively the test runner should be able to simply use cloud APIs and be implemented once for all targets. No?

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -76,6 +76,7 @@ export class Bucket extends cloud.Bucket {
 
     this.bucket = new S3Bucket(this, "Default", {
       bucketPrefix,
+      forceDestroy: true, // TODO: this should only be true when compiled in test mode
```
eladb, line 79: `App.of(this).isTestEnvironment`?

eladb (changes requested):
This is very exciting! Initial comments.

apps/wing/src/cli.ts
ekeren: Me too, and while we are on it. Let's also make sim the default for wing compile

apps/wing/src/commands/test.ts
ekeren: @skyrpex and @polamoros , wait with your change. We might decide to keep the `--target sim` the default behavior.

apps/wing/src/cli.ts
eladb: +1 feels like it makes sense to make sim "first amongst equals"

libs/wingsdk/src/target-sim/test-runner.inflight.ts
Chriscbr: Mmm... it sounds possible? Though I'm not sure I see a clear path to doing it right now, at least two issues come to mind:
- aws function client and simulator function client constructors expecting different kinds of values (one expects simulator handles and other simulator context, the other expects AWS arns)
- logic for obtaining logs/traces for the tests might be very different between the two targets

(Totally open to adding a base class in the future)

(Some WET code is also probably okay for now)

libs/wingsdk/src/cloud/test-runner.ts
Chriscbr: I like the idea

apps/wing/src/cli.ts
Chriscbr: Updated to make sim the default for `compile` and `test` commands

libs/wingsdk/src/target-sim/test-runner.inflight.ts
eladb: Something feels off... Let me challenge you like this: what would it take to implement `TestRunner` in Wing (just using `cloud.Xxx` resources)?

For what it's worth, it's a good dogfooding exercise...

libs/wingc/src/jsify.rs
Chriscbr: Sounds possible? For context, here's the high-level preflight code this PR generates at the moment:

```js
class $Root extends $stdlib.core.Resource {
	constructor(scope, id) {
		super(scope, id);
		// user resources
	}
}

class $App extends $AppBase { // $AppBase is one of the SDK's app classes
	constructor() {
		super({ outdir: $outdir, name: "test_bucket", plugins: $plugins });
		if ($wing_test) {
			new $Root(this, "env0");
			const $test_runner = this.testRunner;
			const $tests = $test_runner.findTests();
			for (let $i = 1; $i < $tests.length; $i++) {
				new $Root(this, "env" + $i);
			}
		} else {
			new $Root(this, "Default");
		}
	}
}
```

We could try something like this:

```js
class $App extends $AppBase { // $AppBase is one of the SDK's app classes
	constructor() {
		super({ outdir: $outdir, name: "test_bucket", plugins: $plugins, root: $Root, isTest: $wing_test });
	}
}
```

But I wasn't sure it made sense to make all of those pieces part of the `App`'s responsibilities. Definitely something we can play around with in the future though

Chriscbr: There's probably some clever-er solution I just can't think of it 😅

eladb: I think passing the `$Root` type is legitimate (albeit not JSIIable).

libs/wingsdk/src/cloud/test-runner.ts
Chriscbr: ?

libs/wingc/src/jsify/code.rs
Chriscbr: OK

hasanaburayyan (commented):
This is soo cool! Not completely through it yet, but I wonder should we document somewhere about known limitations? IE. AWS default quotas in accounts?

apps/wing/src/commands/test.ts
```diff
@@ -91,12 +119,155 @@ async function testOne(entrypoint: string) {
...
+    const results = new Array<sdk.cloud.TestResult>();
+    for (const path of tests) {
+      results.push(await testRunner.runTest(path));
+    }
+    return results;
+  });
+
+  const hasFailures = printTestReport(synthDir, results);
+
   if (hasFailures) {
```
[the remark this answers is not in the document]
Chriscbr: > How do they cleanup the resources in the interim? Is the experience that we have them run `terraform destroy` in different directories? (until we add the `--destroy` option)

Yeah, that would the experience at the moment. cd into `target/test/<your-app-path>` and then `terraform destroy`.

> Since the default way to manage TF state is local state files, what if the dev runs into a scenario where their local state file gets blown to bits or something. Does this just orphan resources in their account that they need to cleanup?

If you deleted your `target/` directory or otherwise tampered with the contents, then I think resources would be orphaned, yes. (I think losing state files is a limitation/risk inherent to terraform if I understand correctly). But as long as you're just using Wing CLI commands or the Wing Console that shouldn't happen. I think it'll probably be possible to use have Wing use remote terraform state via a Wing plugin.

> I wonder if we should by default always cleanup and make an optional flag --retain-on-fail or something so that test resources are not accidentally left over in an account, to avoid anyone getting a suprise AWS bill. I think my mental model of "test" resources are that they should be always be ephemeral unless someone presses the "I know what Im doing button"

I'm empathetic to the idea of cleaning up by default because of the high costs of resources in the cloud, but I guess I also just worry about what the developer experience would be like if a test failed and it automatically started cleaning it up before it even asked you (because that means you might have to way for an entire deployment cycle in order to reproduce the situation). Maybe we could have it so as long as you're running `wing test` interactively, we can prompt you whether you want to retain or not, and if it's not interactive the CLI will clean up by default (or retain resources if `--retain-on-fail` was passed)?

libs/wingsdk/src/target-tf-aws/test-runner.inflight.ts
```diff
@@ -0,0 +1,43 @@
...
+      await client.invoke("");
+      pass = true;
+    } catch (e) {
+      error = (e as any).message;
+    }
+    return {
+      path,
+      pass,
+      error,
+      traces: [], // TODO
```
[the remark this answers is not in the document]
Chriscbr: Tracking issue for now probably, thanks for reminding me

Chriscbr:
> I wonder should we document somewhere about known limitations? IE. AWS default quotas in accounts?

👍 Maybe this is could be something that can include in per-resource SDK docs? Like supposing we have a conceptual docs page about each SDK resource, that explains its purpose and usage, there could also be an "Operational details" section on each resource, highlighting any caveats or cloud-specific details that we deem are important.

apps/wing/src/commands/test.ts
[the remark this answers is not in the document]
hasanaburayyan: Yea I mean I think this maybe a philosophical choice. "To be cheap or to be convenient" 😂 I totally see the reasoning for both arguments. I know a lot of people get frustrated with cloud formation deployments since the default behavior is "If it failed roll it back" but personally I prefer this. 

When it comes to a developer experience, I feel like the time cost of having to re-run with a `--retain-on-failure` is preferable to "oops I left 8 copies of the app running last month" kinda thing.

hasanaburayyan:
> 👍 Maybe this is could be something that can include in per-resource SDK docs? Like supposing we have a conceptual docs page about each SDK resource, that explains its purpose and usage, there could also be an "Operational details" section on each resource, highlighting any caveats or cloud-specific details that we deem are important.

I like that, I wonder if this is something (later) the testrunner can manage. Like a validation step per resource to query existing quotas and ensure that given "x" number of tests that a sufficient quota is available. (thinking outloud)

apps/wing/src/commands/test.ts
Chriscbr: I've retained `--target sim` as the default, so resolving this thread

libs/wingc/src/jsify/code.rs
Chriscbr: Let's keep the API minimal for now 🙏

libs/wingsdk/src/cloud/test-runner.ts
Chriscbr: #1972

libs/wingsdk/src/target-sim/test-runner.inflight.ts
Chriscbr: @eladb I think to implement a `TestRunner` in Wing we are missing at least a few more capabilities, like construct tree inspection, logging, and an API for errors in the Wing language. Here's an idea of what I have sketched out:

```js
bring cloud;

struct TestResult {
  path: str;
  pass: bool;
  error: str?;
  logs: Array<str>;
}

resource TestRunner {
  fn: cloud.Function;
  init(fn: cloud.Function) {
    this.fn = fn;
    // TODO
    // this.fns = nodeof(this).find_all().filter(Test.is_test);
  }
  inflight run_test(): TestResult {
    let var pass = false;
    let var error: str?;
    try {
      this.fn.invoke("");
      pass = true;
    } catch e {
      error = e.message; // TODO: no error API in Wing yet
    }
    let logs = this.fn.logs(tail: 100); // TODO: no logs API in Wing yet
    return TestResult {
      path: nodeof(this.fn).path, // TODO: no `nodeof` API in Wing yet
      pass: true,
      error: error,
      logs: Array<str>[],
    };
  }
}
```

thoughts? Do you think any of these capabilities should be pulled forward into the current sprint?

apps/wing/src/commands/test.ts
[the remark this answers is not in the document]
Chriscbr: I've switched it clean up by default, resolving for now (let's be sure to dogfood this to make sure this default feels right)

libs/wingsdk/src/target-tf-aws/bucket.ts
Chriscbr: Added

libs/wingc/src/jsify.rs
Chriscbr: I agree with the point about testability, but I also still don't feel great about this `root` field in the `App` class / it doesn't seem right to me. :/ I also feel like some of this design might be clearer once we make some of the other modeling changes so that functions aren't materialized into every environment etc.

eladb: No problem. Just some thoughts...

libs/wingsdk/src/cloud/test-runner.ts
eladb: I was asking if we needed Inflight annotations for this class

libs/wingsdk/src/target-sim/test-runner.inflight.ts
eladb: The original question was whether we can implement this once (in typescript) for all targets and avoid target specific code (and logic duplication).

Seems that for typescript the main missing thing is a way to collect logs, correct?

Let's make sure we have issues for all the missing pieces.

eladb (approved):

libs/wingsdk/test/testing/__snapshots__/simulator.test.ts.snap
Chriscbr: Fixed

libs/wingsdk/src/cloud/test-runner.ts
Chriscbr: Not here I don't think

Chriscbr: Maybe we'll refactor these from `BucketInflightMethods` -> `BucketOperations` and then yeah it would makes sense. But I'm not sure if we need our binding system to track preflight operations yet

libs/wingsdk/src/target-sim/test-runner.inflight.ts
Chriscbr: https://github.com/winglang/wing/issues/1987

PR #1929, opened 2023-03-28
fix: unable to compile on windows when compile on non-system drive

MarkMcCulloh:

### DIFF
--- a/apps/wing/src/commands/compile.ts
+++ b/apps/wing/src/commands/compile.ts
@@ -1,6 +1,6 @@
 import * as vm from "vm";
 
-import { mkdir, readFile, mkdtemp, rename, rm } from "fs/promises";
+import { mkdir, readFile, mkdtemp, rename, rm, copyFile } from "fs/promises";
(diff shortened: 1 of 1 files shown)


Chriscbr (approved):

apps/wing/src/commands/compile.ts
```diff
@@ -1,14 +1,14 @@
 import * as vm from "vm";
 
-import { mkdir, readFile, mkdtemp, rename, rm } from "fs/promises";
+import { readFile, rmSync, mkdirp, move, mkdirpSync, copySync } from "fs-extra";
```
eladb, line 3: Do we really need this extra dependency (pun intended)? All of these are available in `fs` in the node versions we use.

MarkMcCulloh: `move` is not. It's similar to rename, but handles additional cases. `copy` also isn't (it's `copyFile` in `fs`), it also handles additional cases more gracefully.

Feel free to remove the dep if you'd like.

PR #1938, opened 2023-03-29
feat(compiler): interface declaration

### LINKED ISSUE #123: Interface Declaration
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#34-interfaces

staycoolcall911:
This PR adds support for `interface` according to the [spec](https://docs.winglang.io/reference/spec#34-interfaces).

Also in this PR:
- [x] Removed interface fields from the spec and our codebase ([slack discussion](https://winglang.slack.com/archives/C048QCN2XLJ/p1680077416200329)).
- [x] Removed access modifiers from interface methods (to comply with the spec).
- [x] (Very) small refactoring in our grammar.js to rename method return type from `type` to `return_type`.
- [x] Removed unused parameter `is_class` from SymbolEnv::new

Fixes #123


libs/wingc/src/parser.rs
```diff
@@ -584,6 +584,93 @@ impl<'s> Parser<'s> {
...
+
+			// ignore commas
+			if !extend.is_named() {
+				continue;
+			}
+
+			let interface_type = self.build_udt_annotation(&extend)?;
+			match interface_type {
+				TypeAnnotation::UserDefined(interface_type) => extends.push(interface_type),
+				_ => {
```
yoav-steinberg, line 646: `build_udt_annotation` alaways returns a udt or an error, so there's no need to handle this, we should just panic.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -400,6 +421,46 @@ exports[`inflight_ref_unknown_op.w > stderr 1`] = `
...
 
+exports[`interface.w > stderr 1`] = `
+"error: Unknown symbol \\"IB\\"
+  --> ../../../examples/tests/invalid/interface.w:2:22
+  |
+2 | interface IA extends IB {
+  |                      ^^ Unknown symbol \\"IB\\"
+
+
+error: Interface \\"IA (at ../../../examples/tests/invalid/interface.w:2:11)\\" extends \\"any\\", which is not an interface
```
yoav-steinberg, line 432: This seems like some bug in our type checker, the error should be that `IB` doesn't exist not that `any` is not an interface.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -400,6 +421,46 @@ exports[`inflight_ref_unknown_op.w > stderr 1`] = `
...
+
+
+error: Unknown symbol \\"IDontExist\\"
+   --> ../../../examples/tests/invalid/interface.w:10:26
+   |
+10 | interface IExist extends IDontExist {
+   |                          ^^^^^^^^^^ Unknown symbol \\"IDontExist\\"
+
+
+error: Interface \\"IExist (at ../../../examples/tests/invalid/interface.w:10:11)\\" extends \\"any\\", which is not an interface
```
yoav-steinberg, line 442: Same issue `IDontExist` doesn't exist, but `any` has nothing to do with it.

examples/tests/valid/interface.w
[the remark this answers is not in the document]
staycoolcall911: 1. FYI @yoav-steinberg - @Chriscbr decided to support impl `interface`s only by `resource`s at the moment. Not sure if we have an issue for supporting `class`es too

tools/hangar/__snapshots__/invalid.ts.snap
staycoolcall911: We actually show 2 errors:
```sh
error: Unknown symbol "IB"
  --> ../../examples/tests/invalid/interface.w:2:22
  |
2 | interface IA extends IB {
  |                      ^^ Unknown symbol "IB"
error: Interface "IA (at ../../examples/tests/invalid/interface.w:2:11)" extends "any", which is not an interface
```
The first one is the one you expected to have.
Let me see if the second one is something we want to remove.

examples/tests/invalid/impl_interface.w
```diff
@@ -17,3 +17,19 @@ resource C impl cloud.Bucket {
...
+interface I2 extends I1 {
+  inflight method_2(x: str): str;
+}
+
+interface I3 extends I2 {
+  method_3(x: Array<num>): Array<num>;
+}
+
+resource r impl I3 {
+  init() {}
```
Chriscbr, line 34: Can we add a comment saying what the error here is?

staycoolcall911: Yes - good one. Added

yoav-steinberg (approved):
I think the invalid snapshot tests still need to be updated (not to include the `any` errors).

tools/hangar/__snapshots__/invalid.ts.snap
staycoolcall911: Done (removed the 2nd error message)

PR #1942, opened 2023-03-29
fix(compiler): expects incorrect argument counts

### LINKED ISSUE #1566: Incorrect argument counts when inflight is assigned to resource field
### Community Note

### I tried this:

```
bring cloud;

resource Doubler {
  func: inflight (str): str;
  init(func: inflight (str): str) {
    this.func = func;
  }
  inflight invoke(message: str): str {
    let result1 = this.func(message);
    let result2 = this.func(message);
    return result2;
  }
}

let fn = new Doubler(inflight (m: str): str => {
  return "Hello ${m}!";
});

```

### I expected this:

It compiles

### Instead, this happened:

```
Error at hello.w:9:19 | Expected 0 arguments but got 1
Error at hello.w:10:19 | Expected 0 arguments but got 1
```

### Component:

Compiler

### Wing Version:

0.5.48

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

_No response_

Chriscbr:
Fixes #1566 by removing "this" as an implicit parameter in functions, instead storing this information in a `this_type` field for potential future use.

Related to #1678

Misc:
- Fixed an issue where "any" types were not stringable
- Fixed an issue where "any" types could not be iterated over
- Fixed an issue where capturable interfaces did not have their components extracted properly during the capturing/binding algorithm


libs/wingc/src/jsify.rs
```diff
@@ -1434,7 +1435,16 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+				} else if let Some(iface) = obj_type.as_interface() {
+					Some(
+						iface
+							.env
+							.lookup(&property, None)
+							.expect("covered by type checking")
+							.as_variable()
+							.unwrap(),
+					)
+				} else if let Some(s) = obj_type.as_struct() {
```
Chriscbr, line 1447: This was needed to fix a separate bug in the `Doubler` example where the compiler wouldn't allow `this.func` to be used inflight because it's an interface.

libs/wingc/src/type_check.rs
```diff
@@ -2167,20 +2134,17 @@ impl<'a> TypeChecker<'a> {
...
+					let mut method_type = self.resolve_type_annotation(
+						&TypeAnnotation::FunctionSignature(method_def.signature.clone()),
+						env,
+						stmt.idx,
+					);
+					// use the class type as the function's "this" type
+					if let Type::Function(ref mut f) = *method_type {
+						f.this_type = Some(class_type.clone());
+					} else {
+						panic!("Expected method type to be a function");
```
yoav-steinberg, line 2146: I think this should be done only if `!method_def.is_static`

libs/wingc/src/type_check.rs
```diff
@@ -2194,11 +2158,17 @@ impl<'a> TypeChecker<'a> {
...
 					&TypeAnnotation::FunctionSignature(constructor.signature.clone()),
 					env,
 					stmt.idx,
 				);
+				// use the class type as the constructor's "this" type
+				if let Type::Function(ref mut f) = *constructor_type {
+					f.this_type = Some(class_type.clone());
+				} else {
+					panic!("Expected constructor type to be a function");
+				}
```
yoav-steinberg, line 2171: I'd like to treat the constructor is a static with `this` being something special that's available in it's env, but isn't part of the sig. Lets remove this.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -475,11 +472,12 @@ impl<'a> JsiiImporter<'a> {
 					}
 
 					for param in params {
-						arg_types.push(self.parameter_to_wing_type(&param));
+						param_types.push(self.parameter_to_wing_type(&param));
 					}
 				}
 				let method_sig = self.wing_types.add_type(Type::Function(FunctionSignature {
-					parameters: arg_types,
+					this_type: Some(wing_type),
```
yoav-steinberg, line 479: condition based on `!is_static`

yoav-steinberg (commented):
Nice, this is more or less what I envisioned. The only thing is that I think `this_type` needs to be an indicator of whether the function is a static or not. In other words is is bound to an instance or not. And it it is, `this_type` is the type of that instance.
Initializers are a special case, they have a `this` in their env, so we can access it but they aren't bound to an instance yet so `this_type` should be `None`.

yoav-steinberg (approved):

PR #1945, opened 2023-03-30
chore(compiler): remove `make_instance_variable`

yoav-steinberg:
In too many cases we just called `make_variable` on a symbol env without considering if it were static or not. Now we'll have to do so. This should reduce bugs and make code cleaner.


staycoolcall911 (approved):
Nice!

PR #1951, opened 2023-03-30
feat(sdk): redis sim support

hasanaburayyan:
Take 2 at redis sim support. We had to revert the original PR due to an issue when publishing (see [slack thread](https://winglang.slack.com/archives/C048QCN2XLJ/p1680161898679079)). dockerode install hooks were probably causing us issues with hard links which are not allowed to be published. 

This is the same PR but with `dockerode` removed. Since our use cases are very simple for docker. (pull, run, inspect, delete) and none of those require any fancy mounting this just goes in favor of using `execFile` to invoke the docker cli.


libs/wingsdk/src/util.ts
```diff
@@ -161,3 +162,16 @@ export function sanitizeCodeText(code: string): string {
 export function sanitizeCode(code: Code): string {
   return sanitizeCodeText(code.text);
 }
+
+export async function runCommand(cmd: string): Promise<any> {
+  const raw = await new Promise((resolve, reject) => {
+    exec(cmd, (error, stdout, stderr) => {
```
yoav-steinberg, line 168: I suspect this might be a security hole, also considering we're passing the `REDIS_IMAGE` env var as an arg. Suggest switching to `execFile` and changing the api of `runCommand` accordingly (`(cmd: string, args: string[])`).

hasanaburayyan: I switched this to `execFile` but curious @yoav-steinberg what is the security hole concerns?

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,137 @@
...
+// Issue using types from ioredis with JSII
+//import { Redis as IoRedis } from "ioredis";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const IoRedis = require("ioredis");
+
+export class Redis implements IRedisClient, ISimulatorResourceInstance {
+  private container_name: string;
+  private readonly REDIS_IMAGE =
+    process.env.REDIS_IMAGE ??
+    "redis@sha256:e50c7e23f79ae81351beacb20e004720d4bed657415e68c2b1a2b5557c075ce0";
```
Chriscbr, line 17: What version of redis is this? Maybe we could add a comment

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// Issue using types from ioredis with JSII
+//import { Redis as IoRedis } from "ioredis";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const IoRedis = require("ioredis");
+
+export class Redis implements IRedisClient, ISimulatorResourceInstance {
+  private container_name: string;
+  private readonly REDIS_IMAGE =
+    process.env.REDIS_IMAGE ??
```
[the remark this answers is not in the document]
hasanaburayyan: Thats a good callout 👍

libs/wingsdk/src/util.ts
MarkMcCulloh: afaict, `REDIS_IMAGE` is not an environment variable, so there shouldn't be a concern with shell injection

hasanaburayyan: @MarkMcCulloh I actually just changed it to allow it as an environment variable. But even at that I think we should be fine, this runCommand() helper would only ever be run during local compilation. With that said I still think going with args[] is good.

PR #1967, opened 2023-04-03
chore(compiler): cleanups in type checker

yoav-steinberg:
* don't pass statement index all around the type checker, instead store it in the TypeChecker sate.
* Store `in_json` in the type checker state instead of adding the extra context struct.
* removed duplicate code for method/init type checking 

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -989,6 +989,7 @@ pub struct TypeChecker<'a> {
 	pub diagnostics: RefCell<Diagnostics>,
 
 	in_json: bool,
+	statement_idx: usize,
 	// generated_resource_counter: usize,
 }
 
@@ -1000,6 +1001,7 @@ impl<'a> TypeChecker<'a> {
 			source_path,
 			diagnostics: RefCell::new(Diagnostics::new()),
 			in_json: false,
+			statement_idx: 0,
 		}
 	}
 
@@ -1107,8 +1109,8 @@ impl<'a> TypeChecker<'a> {
 	}
 
 	// Validates types in the expression make sense and returns the expression's inferred type
-	fn type_check_exp(&mut self, exp: &Expr, env: &SymbolEnv, statement_idx: usize) -> TypeRef {
-		let t = self.type_check_exp_helper(&exp, env, statement_idx);
+	fn type_check_exp(&mut self, exp: &Expr, env: &SymbolEnv) -> TypeRef {
+		let t = self.type_check_exp_helper(&exp, env);
 		exp.evaluated_type.replace(Some(t));
 		t
 	}
@@ -1117,14 +1119,14 @@ impl<'a> TypeChecker<'a> {
 	/// and break early, while still setting the evaluated type on the expression.
 	///
 	/// Do not use this function directly, use `type_check_exp` instead.
-	fn type_check_exp_helper(&mut self, exp: &Expr, env: &SymbolEnv, statement_idx: usize) -> TypeRef {
+	fn type_check_exp_helper(&mut self, exp: &Expr, env: &SymbolEnv) -> TypeRef {
 		match &exp.kind {
 			ExprKind::Literal(lit) => match lit {
 				Literal::String(_) => self.types.string(),
 				Literal::InterpolatedString(s) => {
 					s.parts.iter().for_each(|part| {
 						if let InterpolatedStringPart::Expr(interpolated_expr) = part {
-							let exp_type = self.type_check_exp(interpolated_expr, env, statement_idx);
+							let exp_type = self.type_check_exp(interpolated_expr, env);
 							self.validate_type_in(exp_type, &self.types.stringables(), interpolated_expr);
 						}
 					});
@@ -1135,8 +1137,8 @@ impl<'a> TypeChecker<'a> {
 				Literal::Boolean(_) => self.types.bool(),
 			},
 			ExprKind::Binary { op, left, right } => {
-				let ltype = self.type_check_exp(left, env, statement_idx);
-				let rtype = self.type_check_exp(right, env, statement_idx);
+				let ltype = self.type_check_exp(left, env);
+				let rtype = self.type_check_exp(right, env);
 
 				match op {
 					BinaryOperator::LogicalAnd | BinaryOperator::LogicalOr => {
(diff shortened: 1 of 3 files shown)


Chriscbr (approved):

libs/wingc/src/type_check.rs
```diff
@@ -991,6 +995,9 @@ pub struct TypeChecker<'a> {
 	source_path: &'a Path,
 
 	pub diagnostics: RefCell<Diagnostics>,
+
+	in_json: bool,
```
hasanaburayyan, line 999: Originally this was my intention to have the Typechecker have a mutable field like: `is_json` though the discussion that arises here: https://github.com/winglang/wing/pull/1524#discussion_r1119555854

is that it removes clutter but is not as maintainable as passing context.

libs/wingc/src/type_check.rs
```diff
@@ -1460,15 +1457,17 @@ impl<'a> TypeChecker<'a> {
...
 					}
 				}
 
 				struct_type
 			}
 			ExprKind::JsonLiteral { is_mut, element } => {
-				self.type_check_exp(&element, env, statement_idx, &TypeCheckerContext { in_json: true });
+				self.in_json = true;
+				self.type_check_exp(&element, env);
+				self.in_json = false;
```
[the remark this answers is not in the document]
yoav-steinberg: Almost, if I do a json literal within a json literal I think I'll need a counter
```rust
self.in_json += 1;
self.type_check_expt();
self.i_json -=1;
```
I didn't do this because I was under the impression we don't allow an explicit `Json` within a `Json`. But I guess we do. So I'll fix.

libs/wingc/src/type_check.rs
yoav-steinberg: I'd consider the `TypeChecker` class as the mutable state used for context by design. It's identical to `self` in the `Visit` pattern. This still means the AST itself should remain immutable. But we need some mutable state during visiting (look at all the existing fields in `TypeChecker`). So I think you're original intention was spot on.

libs/wingc/src/type_check.rs
```diff
@@ -1803,16 +1793,21 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
-	fn type_check_statement(&mut self, stmt: &Stmt, env: &mut SymbolEnv, context: &TypeCheckerContext) {
+	fn type_check_statement(&mut self, stmt: &Stmt, env: &mut SymbolEnv) {
+		// Set the current statement index for symbol lookup checks. We can safely assume we're
+		// not overwriting the current statement index because `type_check_statement` is never
+		// recursively called (we use a breadth-first traversal of the AST statements).
+		self.statement_idx = stmt.idx;
```
Chriscbr, line 1800: hmm, maybe I still don't totally grasp how the breadth-first traversal works. How does a case like this get handled?

```
let a = 1;
let do_stuff = () => {
  let x = 2;
  let adder = (): num => {
    return a + b + x + y; // HERE
  };
  let y = 3;
  adder();
}
let b = 4;
```

I guess initially I thought this would require calling `type_check_statement` within another `type_check_statement` call, but is this comment saying that is not what happens?

(FWIW it looks like this example produces the correct errors in our current version of Wing so that is good news 👍)

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
Chriscbr: I think if you use a boolean it ought to still work since we are using the call stack to track the history of `in_json` values, but either should be OK

libs/wingc/src/type_check.rs
Chriscbr: I'm ok with either. I think it's useful to know at least that we can refactor out some of the state/context if necessary. (FWIW sometimes, depending on what you are trying to do in Rust, refactoring out structs into smaller pieces is actually needed in order to satisfy various borrow checking requirements -- check this article for an example: https://smallcultfollowing.com/babysteps/blog/2018/11/01/after-nll-interprocedural-conflicts/)

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
yoav-steinberg: once `in_json` is in (mut) `self` we're not using the call stack anymore.

libs/wingc/src/type_check.rs
hasanaburayyan: Yea as well Im not strongly opinionated here so either way is fine with me.

yoav-steinberg: nice read, i see how we can decide to factor out some `Context` struct and still store it in `TypeChecker` if we need to in the future, without passing it as an arg all over the place.

libs/wingc/src/type_check.rs
yoav-steinberg: 1. Traverse statements in top level and create this env

| symb | idx |
| --- | --- |
| a | 0 |
| do_stuff | 1 |
| b | 2 |

during traversal we store the inner scope of `do_stuff` for future work

2. Traverse the `do_stuff` scope and create this env with env.staement_idx=1 because it was created in the top level scope's statement idx 1

| symb | idx |
| --- | --- |
| x | 0 |
| addr | 1 |
| y | 2 |
| addr | 3 |

during traversal we store the inner scope of `addr` for future work

3.  Traverse the `addr` scope and create this env eith env.satemt_idx = 1 (becase it was created in `do_stuff`'s second statement)

| symb | idx |
| --- | --- |

(^nothing defined here)

When we handle `a + b + x + y` we know that we're in index 0 of the `adder` scope. And we do the following lookups:
`a`: it's not in our scope, go up, now we're in `do_stuff` env so our index is set to 1 (`addr` env's index), we still can't find `a`, go up, now we're the top env so our index is set to 1 (`do_stuff` env's index), we find `a` with index 0, its index is less than 1 (our index) so the lookup succeeds.

`b`: it's not in our scope, go up, now we're in `do_stuff` env so our index is set to 1 (`addr` env's index), we still can't find `b`, go up, now we're the top env so our index is set to 1 (`do_stuff` env's index), we find `b` with index 2, its index is more than 1 (our index) so the lookup fails.

`x`: it's not in our scope, go up, now we're in `do_stuff` env so our index is set to 1 (`addr` env's index), we find `x` with index 0, its index is less than 1 (our index) so the lookup succeeds.

`y`: it's not in our scope, go up, now we're in `do_stuff` env so our index is set to 1 (`addr` env's index), we find `y` with index 2, its index more than 1 (our index) so the lookup fails.

Hope that was followable :)

Chriscbr: This explanation is super helpful!! Yeah I think I was a little confused on how the statement indices are stored and how the lookups are performed, but now it makes sense to me. Thanks for writing this up 😄

staycoolcall911: Amazing, thanks for the detailed explanation

PR #1982, opened 2023-04-04
fix: large jsii imports unusable in vscode

MarkMcCulloh:
https://user-images.githubusercontent.com/1237390/229843544-59d1f96c-b179-482a-93d0-cd925ec719c4.mp4

### Changes

- Lazily load types from jsii as needed
  - All top-level submodules are imported immediately, but none of the types within them are
- Avoid double de-serializing jsii manifest
- In LSP, the JSII manifests loaded into memory are kept around for subsequent compilations
- Fixed an existing bug where vscode completion was doubling available symbols

### Results
- In LSP, after the first compilation, files with `bring "@cdktf/provider"` compile almost instantly.
- Compiling a file with `bring "@cdktf/provider"` now takes about **4 seconds** for me (down from 12). 
  - Most of this time is the File -> Object process, rather than our own code

### Follow-ups
 - Different files across an LSP context can likely share the exact same JSII manifest. We can reduce the load by sharing this as long as we guarantee the "import" is truly the same content.
 - We are double-importing the SDK JSII. This bug was already present.
 - Look into a faster way of reading JSON, even if it means losing static types
   - serde's Json `Value` type is about as performant to import as using actual types, so it'd likely be a different library
- If we continue to use serde_json, see if it would be reasonable to do zero-copy deserialization for the string-heavy data-structure


libs/wingc/src/lsp/completions.rs
```diff
@@ -178,8 +178,10 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse
 			completions.push(format_symbol_kind_as_completion(symbol_data.0, symbol_kind));
 		}
 
-		for data in found_env.iter(true) {
-			completions.push(format_symbol_kind_as_completion(&data.0, &data.1));
+		if let Some(parent) = found_env.parent {
```
yoav-steinberg, line 181: where we adding symbols from the current env twice up until now?
I suspect we need to use the `found_env.statemet_idx` and compare it against our parent's satement indexes and follow this all the way up the env tree. We might need to extend env's iter to support this.

MarkMcCulloh: Yep, it was an existing bug.

> I suspect we need to use the found_env.statemet_idx and compare it against our parent's satement indexes and follow this all the way up the env tree. We might need to extend env's iter to support this.

Yeah I agree, although I think it would be okay to keep that for another issue/PR. I only fixed the other bug since it was a big low-hanging fruit I ran into.

yoav-steinberg: can you verify this really happens and open and issue?

MarkMcCulloh: https://github.com/winglang/wing/issues/1984

libs/wingc/src/type_check.rs
```diff
@@ -3027,6 +3059,38 @@ impl<'a> TypeChecker<'a> {
...
+		statement_idx: usize,
+	) -> Result<TypeRef, TypeError> {
+		// Attempt to resolve the type from the current environment
+		// otherwise, attempt to resolve the type from any jsii libs
+		let res = resolve_user_defined_type(user_defined_type, env, statement_idx);
+		if res.is_ok() {
+			return res;
+		}
+
+		for jsii in &*self.jsii_imports {
```
Chriscbr, line 3076: If we have N jsii libraries imported in a Wing source file, would this `for` loop mean that we are doing an O(n) operation each time the compiler tries to resolve a user defined type? curious if we can avoid this, or provably bound the number of times this function is called in some way

libs/wingc/src/type_check.rs
```diff
@@ -2483,59 +2495,77 @@ impl<'a> TypeChecker<'a> {
...
+				namespace_filter,
+				alias: alias.clone(),
+				import_statement_idx: stmt.map(|s| s.idx).unwrap_or(0),
+			});
+
+			self
+				.jsii_imports
+				.iter()
+				.find(|j| j.assembly_name == assembly_name && j.alias.name == alias.name)
+				.unwrap()
```
Chriscbr, line 2560: `expect()`

libs/wingc/src/type_check.rs
```diff
@@ -2483,59 +2495,77 @@ impl<'a> TypeChecker<'a> {
...
-				}
-			};
-
-			assembly_name
+		let jsii = if let Some(jsii) = self
+			.jsii_imports
+			.iter()
+			.find(|j| j.assembly_name == library_name && j.alias.name == alias.name)
+		{
+			jsii
```
Chriscbr, line 2503: Checking my understanding - so in this first branch, we've found a module from `self.jsii_imports` that we've already imported it before, but because it might have only been a partial import, we still need to call `JsiiImporter` and try importing it again, yeah?

libs/wingc/src/type_check.rs
```diff
@@ -2211,7 +2221,9 @@ impl<'a> TypeChecker<'a> {
 				let extend_interfaces = extends
 					.iter()
 					.filter_map(|i| {
-						let t = resolve_user_defined_type(i, env, stmt.idx).unwrap_or_else(|e| self.type_error(e));
+						let t = self
+							.resolve_user_defined_type(i, env, stmt.idx)
+							.unwrap_or_else(|e| self.type_error(e));
```
Chriscbr, line 2226: I wonder if there's any way to make sure that the method version of `self.resolve_user_defined_type` always gets used instead of the free-floating function

libs/wingc/src/type_check.rs
```diff
@@ -994,6 +996,9 @@ pub struct TypeChecker<'a> {
 	/// The path to the source file being type checked.
 	source_path: &'a Path,
 
+	/// JSII Manifest descriptions to be imported
+	jsii_imports: &'a mut Vec<JsiiImportSpec>,
```
Chriscbr, line 1000:
```suggestion
	/// JSII Manifest descriptions to be imported.
	/// May be reused between compilations
	jsii_imports: &'a mut Vec<JsiiImportSpec>,
```

Chriscbr, line 1000: (random thoughts) It would be nice if there was some way to model that we want to allow this vec to be extended, but the individual `JsiiImportSpec` values are basically immutable / should not be touched. Not sure what the Rusty way to do this is though

Chriscbr (commented):
❤️  the improvements to wingii and the jsii importer

libs/wingc/src/type_check.rs
MarkMcCulloh: It's probably worth a benchmark. The alternative here would be to use a hashmap.
However, even though hashmaps have constant big O lookups, for small `n` the performance for plain old vectors may still be better. "small" here meaning like < 100. 
I don't think it'll be super common to have a large number of jsii imports.

Also, before this forloop it attempts to lookup the type in a O(1) way anyways. So this will not happen on every lookup.

eladb:
Very cool stuff!

libs/wingc/src/type_check.rs
MarkMcCulloh: they actually both serve different purposes now.
The method version not also imports the type from JSII as needed, whereas the function version will fail if any types are not already imported.

I'll add docstrings to clarify

libs/wingc/src/type_check.rs
MarkMcCulloh: AFAIK the only way to do it would be to make the JsiiImportSpec fields private and add getters. or whatever this is doing https://github.com/dtolnay

Not sure i'm conviced it's worthwhile, but it's not that weird to do

libs/wingc/src/type_check.rs
MarkMcCulloh: Wait, I remember why this is here now: For the LSP case where we carry forward the vector of import data. In that case, we need to update the current symbol environment with the same info again.

The double-import thing is still an issue though

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -172,47 +147,54 @@ impl<'a> JsiiImporter<'a> {
...
 		// Check if this is a JSII enum and import it if it is
-		let jsii_enum = self.jsii_types.find_enum(type_fqn);
+		let jsii_enum = self.jsii_spec.type_system.find_enum(type_fqn);
 		if let Some(jsii_enum) = jsii_enum {
 			self.import_enum(jsii_enum);
 			return;
 		}
 
-		panic!("Type {} was not found in the type system", type_fqn);
+		debug!("Type {} was not found in the type system", type_fqn);
```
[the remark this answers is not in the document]
MarkMcCulloh: Not sure about this one. This function is really just responsible for updating the internal collection of types. The real error that surfaces to the user later is that the type isn't found. I don't feel like adding an additional diagnostic here gives users any more info.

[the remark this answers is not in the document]
MarkMcCulloh: > in theory aren't we always importing known types?

Not anymore, if the user tries `cloud.butt` it will try to find it in the assembly via this method.

I can convert this method to have a boolean return type (Result would be overkill IMO) to make the intention more clear.

Chriscbr (approved):
LGTM

eladb:
Excited by this!

PR #1986, opened 2023-04-04
feat(sdk): `api.url`

### LINKED ISSUE #1849: `Api.url` preflight property

Chriscbr:
Closes #1849

Misc:
- Refactored some AWS-specific code from `core.CdktfApp` to `tfaws.App`
- Moved some simulator-specific logic out of `cloud.Bucket` and into the sim-specific `Bucket` class
- Fixed a bug where `ApiResponse.body` is not stringified, resulting in apps working differently locally as in the cloud


libs/wingsdk/src/shared-aws/api.inflight.ts
```diff
@@ -0,0 +1,6 @@
+import { IApiClient } from "../cloud";
+
+export class ApiClient implements IApiClient {
+  // TODO: does `url` belong inside the inflight class?
+  public constructor(public readonly url: string) {}
+}
```
Chriscbr, line 6: In order to get `api.url` accesses to work on AWS, I needed to copy the field from preflight into inflight. Something smells off to me about this, but I think it's probably reasonable for now since this is also a kind of duplication the compiler does when jsify-ing user-defined classes I think

eladb: Wouldn't it make sense to pass it though and environment variable like we pass other preflight data to inflight clients?

I feel I might be missing something...

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -35,6 +45,12 @@ export class Api extends cloud.Api {
     });
   }
 
+  public get url(): string {
+    const region = (App.of(this) as App).region;
+    const awsSuffix = "amazonaws.com"; // TODO: use dns_suffix from https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition
+    return `https://${this.api.api.id}.execute-api.${region}.${awsSuffix}/${STAGE_NAME}`;
```
eladb, line 51: Are you telling me there's no attribute for the url of the api gateway?

eladb (approved):

libs/wingsdk/src/target-tf-aws/api.ts
Chriscbr: You know what -- I thought there wasn't because I was looking at the [ApiGatewayRestApi resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/api_gateway_rest_api) docs, but after giving a second look, it looks like the [ApiGatewayDeployment resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_deployment) has an attribute with the invoke url, so we should be able to use that instead. Good callout

libs/wingsdk/src/target-sim/api.ts
```diff
@@ -18,6 +15,10 @@ import * as core from "../core";
 export class Api extends cloud.Api implements ISimulatorResource {
   private _routes: ApiSchema["props"]["routes"] = [];
 
+  public get url(): string {
```
ekeren, line 18: @Chriscbr 

Can we make sure that the console is able to access this field somehow? (maybe it is already possible, but I am not sure) 

AFAIK, the console (@polamoros ) is using an internal detail of the sim implementation in order to get the url value. The console should only use the inflight api of a resource, because we want to be able to use the console for the tf-aws (and friends) target as well 

 @ainvoner &  @eladb  - FYI

Chriscbr: @ekeren It should be possible to access the URL through the simulator's public methods:

```js
const simulator = new sdk.testing.Simulator({ simfile : "./target/api_simple.wsim"});
await simulator.start();

const queue = simulator.getResourceConfig("root/cloud.Api");
const routes = queue.props.routes;
console.log("routes are:", JSON.stringify(routes));
const url = queue.attrs.url;
console.log("url is:", url);
```

For now I think it is OK to treat resource config as a public API (breaking changes will be communicated with console team and should require semver bumps), as long as it's being accessed through the SDK's classes/methods. cc @ainvoner @polamoros

Issue #1872, opened 2023-03-21
unused resource breaks the code


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

The following code
```ts (wing)
bring cloud;

resource DeclareAndNotUse {
  c: cloud.Counter;
  init() {
    this.c = new cloud.Counter();
  }
  inflight hello(): str {
    return "hello world";
  }
}

let a = new DeclareAndNotUse();
new cloud.Function(inflight () => {
  print(a.hello());
}) as "test";
```

```sh
wing test a.w
```

### I expected this:

to compile and work

### Instead, this happened:

```sh
wingly % wing test aaa.w
fail ┌ aaa.w » root/test
     │ Error: Missing environment variable: COUNTER_HANDLE_3356e028
     │     at evalmachine.<anonymous>:27:17
     │     at exports.handler (evalmachine.<anonymous>:30:9)
     │     at evalmachine.<anonymous>:33:9
     │     at Script.runInContext (node:vm:141:12)
     │     at Object.runInContext (node:vm:297:6)
     │     at /Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:127:12
     │     at new Promise (<anonymous>)
     │     at runInSandbox (/Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:108:12)
     │     at Object.activity (/Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:55:24)
     │     at Object.withTrace (/Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:82:50)
     │     at Function.invoke (/Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:52:29)
     │     at Simulator.runTest (/Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:270:22)
[...]

PR #1988, opened 2023-04-05
fix(compiler): allow unused sim resources

hasanaburayyan:
Added no-ops bindings for unused fields in resources.

Thanks @Chriscbr and @yoav-steinberg for the pow-wows on this.

Closes: https://github.com/winglang/wing/issues/1872

**Misc**
- Cleaned up unneeded code in `get_captures()` where we no longer needed to return vec of tuples and can just use vec of string

### DIFF
--- a/examples/tests/valid/resource_captures.w
+++ b/examples/tests/valid/resource_captures.w
@@ -36,9 +36,11 @@ resource MyResource {
   set_of_str: Set<str>;
   another: Another;
   my_queue: cloud.Queue;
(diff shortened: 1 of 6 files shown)


Chriscbr (commented):
Can we add a test?

hasanaburayyan:
> Can we add a test?

😅 great call out, I just added one that would have failed previously.

Chriscbr (approved):

PR #1990, opened 2023-04-05
fix(docs): refers to inflight by incorrect name

davidnx:
The doc text was referring to an inflight by name, but the inflight no longer has a name after https://github.com/winglang/wing/commit/62b13c6c9e6760ef9423506da366b702c686a62e


Chriscbr (approved):
Thanks for the fix @davidnx! 🙏

PR #1994, opened 2023-04-06
fix: extern has inconsistant module semantics

MarkMcCulloh:
The string used with extern has several limitations
- It must be a file
- The file must be a CJS module
- The file itself cannot have references to paths relative to itself (because it gets moved)

But if it's a CJS module, what's the point of the first restriction? A CJS module can take many forms as long as it can be imported by `require()`. This PR makes extern more aligned with that, which also fixes the last issue mentioned (since we are no longer copying the file).


tools/hangar/__snapshots__/test_corpus/valid/extern_implementation.w.ts.snap
```diff
@@ -7,36 +7,28 @@ constructor({  }) {
...
-exports[`wing compile -t tf-aws > extern/8b481fc062bdfc1c35bbcbbab32dcf0805e52e773caa1c61121b09e2db8a0b5b.js 1`] = `
-"exports.get_greeting = function(name) {
-  return \`Hello, \${name}!\`;
-};
-
-exports.regex_inflight = async function(pattern, text) {
-  const regex = new RegExp(pattern);
-  return regex.test(text);
-};
+exports[`wing compile -t tf-aws > extern/09e024cf966860b34adf9cbfd8a494763760f871b1f5c00bbc98853b9989078e.js 1`] = `"module.exports = require(require.resolve(\`./external_js.js\`, {paths: [process.env.WING_FILE_DIR]}));"`;
```
Chriscbr, line 29: Maybe we can avoid this level of indirection some way...

MarkMcCulloh: Done. Good callout, I was too busy changing a small piece of the functionality I didn't consider just deleting 80% of it!

apps/wing/src/commands/compile.ts
```diff
@@ -95,6 +95,7 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
   log("work dir: %s", workDir);
 
   process.env["WING_SYNTH_DIR"] = tmpSynthDir;
+  process.env["WING_FILE_DIR"] = resolve(wingDir);
```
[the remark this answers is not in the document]
MarkMcCulloh: I'm good with that. I thought a little bit about this and initially avoided it because we don't really have the concept of a project (yet), but file_dir is pretty silly.

examples/tests/valid/extern_implementation.w
```diff
@@ -1,20 +1,29 @@
 bring cloud;
 
+interface IUuidModule {
+    inflight v4(): str;
+}
```
Chriscbr, line 5: It doesn't look like this is used

Chriscbr (approved):

examples/tests/valid/extern_implementation.w
```diff
@@ -4,17 +4,22 @@ resource Foo {
   extern "./external_js.js" static get_greeting(name: str): str;
   extern "./external_js.js" static inflight regex_inflight(pattern: str, text: str): bool;
   extern "./external_js.js" static inflight get_uuid(): str;
+  extern "./external_js.js" static inflight get_data(): str;
+  extern "uuid" static v4(): str;
```
eladb, line 8: slick!

eladb (commented):
Love it!

PR #1997, opened 2023-04-06
feat: save local terraform state per app

Chriscbr:
Today, if you try compiling your app to `tf-aws` and check the terraform JSON file that is created inside `target/`, you will find the value `terraform.backend.local.path` is set to a location at the root of your project directory (like `/path/to/my-project/terraform.root.tfstate`). This path specifies where `terraform apply` will create a local terraform state file to track what resources have already been created etc.

That works fine if you are working with a single Wing program. However, if you want to deploy or test multiple Wing apps in the cloud, they will all try to store their state in the same file, which means resources may get destroyed unexpectedly.

This PR fixes that so that the terraform state is saved in the per-target directory (for example, if you compiled `main.w` to AWS, your terraform state will be saved at `target/main.tfaws/terraform.tfstate`). This also applies to `wing test` (for example, `wing test -t tf-aws main.w` will save any terraform state to `target/test/main.tfaws/terraform.tfstate`).

### DIFF
--- a/libs/wingsdk/src/core/app.ts
+++ b/libs/wingsdk/src/core/app.ts
@@ -189,6 +189,7 @@ export abstract class CdktfApp extends App {
     super(cdktfStack, "Default");
 
     // TODO: allow the user to specify custom backends
+    // https://github.com/winglang/wing/issues/2003
     new cdktf.LocalBackend(cdktfStack, {
       path: "./terraform.tfstate",
     });
--- a/libs/wingsdk/src/core/app.ts
+++ b/libs/wingsdk/src/core/app.ts
@@ -188,6 +188,11 @@ export abstract class CdktfApp extends App {
 
     super(cdktfStack, "Default");
 
+    // TODO: allow the user to specify custom backends
+    new cdktf.LocalBackend(cdktfStack, {
+      path: "./terraform.tfstate",
+    });
+
     this.outdir = outdir;
     this.isTestEnvironment = props.isTestEnvironment ?? false;
 


libs/wingsdk/src/core/app.ts
```diff
@@ -188,6 +188,11 @@ export abstract class CdktfApp extends App {
 
     super(cdktfStack, "Default");
 
+    // TODO: allow the user to specify custom backends
+    new cdktf.LocalBackend(cdktfStack, {
+      path: "./terraform.tfstate",
```
Chriscbr, line 193: Crossing my fingers this works on Windows? 🤞 Terraform CLI is built with Go so it probably should...

I wanted to use `path.join(outdir, "terraform.tfstate")` here, but that resulted in a path like `/path/to/project-dir/target/test/main.tfaws.036281.tmp/terraform.tfstate` getting generated 😕

Chriscbr: (To test this in CI we would actually have to run `terraform apply` in our tests, and we don't have the tooling set up for that yet...)

MarkMcCulloh: Might be interesting to add https://developer.hashicorp.com/terraform/cli/commands/validate to hangar (and/or some part of wing's toolchain)

MarkMcCulloh (approved):

libs/wingsdk/src/core/app.ts
```diff
@@ -188,6 +188,11 @@ export abstract class CdktfApp extends App {
 
     super(cdktfStack, "Default");
 
+    // TODO: allow the user to specify custom backends
```
hasanaburayyan, line 191: Is this something on our issues radar? I dont see anything myself.

hasanaburayyan (approved):

libs/wingsdk/src/core/app.ts
Chriscbr: I think at some point we ran validate as part of the SDK tests, the problem was just that it's very slow (not sure what they're doing under the hood)

libs/wingsdk/src/core/app.ts
Chriscbr: Makes sense to me https://github.com/winglang/wing/issues/2003

eladb (commented):
nice touch

PR #2007, opened 2023-04-07
chore(compiler): add fold trait (compiler infrastructure)

Chriscbr:
This PR introduces the `Fold` trait, the sibling of `Visit`. This is a tool that lets us perform type-safe AST transformations. Here's an example of an AST transformation that changes each `new <resource>(<args>)` expression so that it's as if the user wrote `new <resource>(<args>) as "HelloWing"`:

```rust
use crate::{
	ast::{Expr, ExprKind},
	fold::{self, Fold},
};

pub struct ConstructIdTransform;

impl ConstructIdTransform {
	pub fn new() -> Self {
		Self
	}
}

impl Fold for ConstructIdTransform {
	fn fold_expr(&mut self, expr: Expr) -> Expr {
		let kind = match expr.kind {
			ExprKind::New {
				class,
				obj_id: _,
				obj_scope,
				arg_list,
			} => ExprKind::New {
				class: self.fold_type_annotation(class),
				obj_id: Some("HelloWing".to_string()),
				obj_scope: obj_scope.map(|s| Box::new(self.fold_expr(*s))),
				arg_list: self.fold_args(arg_list),
			},
			_ => return fold::fold_expr(self, expr),
		};

		Expr {
			kind,
			span: expr.span,
			evaluated_type: expr.evaluated_type,
		}
	}
}
```

The implementation of `Fold` has a lot of boilerplate (just like `Visit`) that could probably be code generated with some Advanced Rust Macros™️ but I wasn't sure how much effort that would require, and maintaining these by hand isn't too bad since they only need to change when our AST data structure changes.

This was written originally for #1448 but figured it's enough code to just put into its own PR.

Misc:
- Fixes our grammar/AST so that the struct `extends` clause expects a `$.custom_type` / `UserDefinedType` instead of just plain identifiers
- Renames `Struct.members` to `Struct.fields` (the word "members" usually refers to fields or methods, but structs do not have methods)


libs/wingc/src/type_check.rs
```diff
@@ -2334,19 +2334,14 @@ impl<'a> TypeChecker<'a> {
...
-						Err(type_error) => {
-							self.type_error(type_error);
+					.filter_map(|ext| {
+						let t = self
+							.resolve_user_defined_type(ext, env, stmt.idx)
+							.unwrap_or_else(|e| self.type_error(e));
+						if t.as_struct().is_some() {
+							Some(t)
+						} else {
+							self.general_type_error(format!("Expected a struct, instead found type \"{}\"", t));
```
MarkMcCulloh, line 2344: Random thought on error conventions: Is the word "instead" needed?

```suggestion
							self.general_type_error(format!("Expected a struct, found type \"{}\"", t));
```

libs/wingc/src/fold.rs
```diff
@@ -0,0 +1,438 @@
+use crate::ast::{
+	ArgList, CatchBlock, Class, ClassField, Constructor, ElifBlock, Expr, ExprKind, FunctionBody, FunctionDefinition,
+	FunctionParameter, FunctionSignature, FunctionTypeAnnotation, Interface, InterpolatedString, InterpolatedStringPart,
+	Literal, Reference, Scope, Stmt, StmtKind, Symbol, TypeAnnotation, UserDefinedType,
+};
+
+/// Similar to the `visit` module in `wingc` except each method takes ownership of an
+/// AST node instance of a reference to it, and returns a new AST node instance.
+/// This trait can be useful for AST transformations.
+pub trait Fold {
```
MarkMcCulloh, line 10: "Fold" is not a term I have heard for this, but I see it's a rust thing and I can't think of a better name 😅 

Maybe ASTTransformer or something, but I guess that kinda implies it's also changing the type (which this isn't).

libs/wingc/src/fold.rs
```diff
@@ -0,0 +1,438 @@
+use crate::ast::{
+	ArgList, CatchBlock, Class, ClassField, Constructor, ElifBlock, Expr, ExprKind, FunctionBody, FunctionDefinition,
+	FunctionParameter, FunctionSignature, FunctionTypeAnnotation, Interface, InterpolatedString, InterpolatedStringPart,
+	Literal, Reference, Scope, Stmt, StmtKind, Symbol, TypeAnnotation, UserDefinedType,
+};
+
+/// Similar to the `visit` module in `wingc` except each method takes ownership of an
+/// AST node instance of a reference to it, and returns a new AST node instance.
+/// This trait can be useful for AST transformations.
```
MarkMcCulloh, line 9:
```suggestion
/// Similar to the `visit` module in `wingc` except each method takes ownership of an
/// AST node instead of a reference to it, and returns a new AST node instance.
/// This trait can be useful for AST transformations.
```

MarkMcCulloh (approved):
Neato

yoav-steinberg (approved):
nice, can't wait to see how you're going to use this.

libs/wingc/src/fold.rs
Chriscbr: yeah, the use of the word "fold" actually goes pretty far back as a piece of functional programming jargon: https://en.wikipedia.org/wiki/Fold_(higher-order_function)

Besides being historically neat, the fact that "fold" is only four characters will save us a lot of keystrokes here, and it should be OK since it's compiler internal. I can probably improve the description here

eladb: +1 on the name

libs/wingc/src/fold.rs
```diff
@@ -0,0 +1,438 @@
...
+		fields: node.fields.into_iter().map(|s| f.fold_symbol(s)).collect(),
+	}
+}
+
+pub fn fold_symbol<F>(_f: &mut F, node: Symbol) -> Symbol
+where
+	F: Fold + ?Sized,
+{
+	node
+}
```
eladb, line 438: This is cool.

My understanding is that this is our mechanism to support syntactic sugar. Correct?

Practically, we will mostly use this to fold expressions to other expressions. Correct?

libs/wingc/src/fold.rs
```diff
@@ -0,0 +1,438 @@
...
+		signature: f.fold_function_signature(node.signature),
+		statements: f.fold_scope(node.statements),
+	}
+}
+
+pub fn fold_expr<F>(f: &mut F, node: Expr) -> Expr
+where
+	F: Fold + ?Sized,
+{
+	let kind = match node.kind {
```
eladb, line 234: I am a bit concerned about maintaining this... Do we have a way to enforce that when we update the model it will requiring updating this code?

I guess Rust algebraic types help...

eladb (approved):

libs/wingc/src/fold.rs
Chriscbr: Yep. Statements can probably be folded too if there's a use case for it

libs/wingc/src/fold.rs
Chriscbr: yeah, Rust's type help a lot, but it's still possible to miss things. It's actually a bigger issue for `visit` since it's possible for someone to change e.g. add a new field to `ast::Class`, and forget to update the `Visit` trait so that the field is recursively visited. Ideally we would like to safely code-generate these traits with Rust's macro system.

libs/wingc/src/type_check.rs
Chriscbr: Interesting, yeah I think to my ears it still makes sense without "instead" 👍

libs/wingc/src/fold.rs
MarkMcCulloh: When I said it's a rust thing I moreso meant that it's described here as being used in this pattern and being used in the rust compiler https://rust-unofficial.github.io/patterns/patterns/creational/fold.html, my bad.

I hadn't heard the term fold but I'm familiar with reducers. This is the first time I've seen a reducer that doesn't actually "reduce" anything, so it looked more like a map. From reading more I think I understand the distinction, but this was definitely not obvious to me on the name alone. I'll just take that as a personal issue of something I should understand.

Issue #1596, opened 2023-02-20
Resource Declaration: inflight fields


ekeren:

ekeren:
This starts with P2, once we work for 10 minutes on the dogfood we will find a real use case for inflight fields

monadabot:
Congrats! :rocket: This was released in [Wing 0.10.3](https://github.com/winglang/wing/releases/tag/v0.10.3).

Issue #1567, opened 2023-02-16
Inflight initializer


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

https://docs.winglang.io/reference/spec#33-resources

```
bring cloud;

resource A {
  init() {}
  inflight init () {}
}

let a = new A();
```

### I expected this:

It compiles

### Instead, this happened:

```
Error at hello.w:3:10 | Symbol "init" already defined in this scope
Error at hello.w:8:9 | Expected type to be "A", but got "void" instead
Error at hello.w:8:9 | Expected 1 arguments but got 0 when instantiating "A"
```

### Component:

Compiler

### Wing Version:

0.5.48

### Node.js Version:

18.12.1

### Platform(s):

MacOS

staycoolcall911:
Duplicates #67

Issue #67, opened 2022-09-01
Implicit inflight client initialization


Chriscbr:
In the "deny list" example, we have this interesting behavior where the inflight code implicitly initializes the inflight client of Bucket. The Bucket class implementation would have its own `~init()` function which might initialize the AWS SDK, for example. To the user, they're just accessing the `Bucket`'s inflight/runtime operations, and they aren't aware there's an AWS SDK under the hood.

I've abbreviated the example below:

```ts
struct DenyListRule { ... }  

resource DenyList {
  _bucket: cloud.Bucket;
  _object_key: str;

  init(props: DenyListProps) {
    this._bucket = cloud.Bucket();
    this._object_key = "deny-list.json";
  }

  ~ rules: mut_map<DenyListRule>?;

  ~ init() {
    // this._bucket is already initialized by the capture mechanic!
    this.rules = this._bucket.get(this._object_key) ?? mut_map<DenyListRule>();
  }

  ~ lookup(name: str, version: str): DenyListRule? {
    return this.rules[name] ?? this.rules["${name}/v${version}"];
  }

  ~ add_rule(rule: DenyListRule) { ... }
}
```

  But what if `Bucket`'s `~init` function had required parameters? In this case, the compiler would not have enough information to automatically initialize the client for `this._bucket` inside the `DenyList` client.

For MVP, I propose we start with the restriction that inflight init functions cannot have any parameters. I think this small restriction is worth it for the end user experience of using resources in their code directly inflight (without manual initialization). Later, if users need more flexibility or this turns out to be a confusing user experience, we could also add an alternative syntax to customize capture client initialization on a per-resource or per-inflight-scope basis, or change the language to make resource client initialization required.

ekeren:
@yoav-steinberg , I think that the TLDR here is the `inflight init` is a no param function.

Chriscbr:
I wonder how Wing should handle the case of performing async logic in inflight constructors, given our model that inflight is async by default. For example, in the following code

```js
resource MyResource {
  b: cloud.Bucket;
  init(b: cloud.Bucket) {
    this.b = b;
  }
  inflight init() {
    this.b.put("data.json", "{}");
  }
  inflight do_stuff(
    let data = this.b.get("data.json");
    // do stuff
  ) 
}
```

Maybe the inflight client could look something like

```js
class MyResourceClient {
  init({ b }) {
    this.$init_promises = [];
    this.b = b;
    $init_promises.push(this.b.put("data.json", "{}"));
[...]

Issue #1123, opened 2023-01-11
Class Declaration: `super`


eladb:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#32-classes

Add support for calling `super(...)` in inheriting classes' init

verify that:
- [ ]  user defined resource can inherit from about user defined resource.
- [ ] child resource's client class has access to parent's client fields.
- [ ] inflight init in child is called after inflight init in parent is called.

monadabot:
Congrats! :rocket: This was released in [Wing 0.21.23](https://github.com/winglang/wing/releases/tag/v0.21.23).

PR #2021, opened 2023-04-09
feat(compiler): inflight initilizers and inflight fields

yoav-steinberg:
see #1567 #1596 #67

```wing
bring cloud;

resource Foo {
  inflight var _tmp: num;
  init() { }
  inflight init() {
    this._tmp = 0;
  }
  inflight inc_tmp() {
    this._tmp = this._tmp + 1;
  }
}

let a = new Foo();
  a.inc_tmp();
  log("hey ${a._tmp}");
}) as "test";
```

- [x] missing tests


examples/tests/invalid/reassign_to_nonreassignable.w
```diff
@@ -5,13 +5,23 @@ x = x + 1;
 // Assign to non-reassignable field
 resource R {
   f: num;
+  inflight_f: num;
```
eladb, line 8: Can you please add the errors as comments? It's very helpful

eladb, line 8:
```suggestion
  not_inflight_f: num;
```

examples/tests/invalid/resource_init.w
```diff
@@ -0,0 +1,11 @@
+resource R {
+  init() {}
+  init() {}
+//^-- can't have multiple inits
+  inflight init() {}
+  inflight init() {}
+        // ^-- can't have multiple inflight inits
+
+  inflight init(x: num) {}
+             // ^-- can't have inflight init with args
```
eladb, line 10: nice

libs/wingc/src/jsify.rs
```diff
@@ -1108,7 +1127,31 @@ impl<'a> JSifier<'a> {
...
+					def,
+					&JSifyContext {
+						in_json: context.in_json.clone(),
+						phase: def.signature.phase,
+					},
+				)
+			})
+			.collect_vec();
+
+		if let Some(inflight_init) = inflight_init {
```
eladb, line 1144:
```suggestion
                // add the inflight init method to the list of client methods
		if let Some(inflight_init) = inflight_init {
```

libs/wingc/src/jsify.rs
```diff
@@ -1066,23 +1077,31 @@ impl<'a> JSifier<'a> {
...
 			.join(", ");
-		formatdoc!("
+		formatdoc!(
+			"
 			_toInflight() {{
 				{inner_clients}
 				const self_client_path = {client_path};
-				return {STDLIB}.core.NodeJsCode.fromInline(`(new (require(\"${{self_client_path}}\")).{resource_name}({{{captured_fields}}}))`);
+				return {STDLIB}.core.NodeJsCode.fromInline(`(
+					await (async () => {{
```
eladb, line 1086: Interesting... This implies that `_toInflight()` is called from an async scope. Is that the case? 

Feels like this is an assumption we should model somehow into the API. Otherwise, someone will assume in the future that it's a sync call.

@Chriscbr curious what you think...

libs/wingc/src/jsify.rs
```diff
@@ -1066,23 +1077,31 @@ impl<'a> JSifier<'a> {
...
+		formatdoc!(
+			"
 			_toInflight() {{
 				{inner_clients}
 				const self_client_path = {client_path};
-				return {STDLIB}.core.NodeJsCode.fromInline(`(new (require(\"${{self_client_path}}\")).{resource_name}({{{captured_fields}}}))`);
+				return {STDLIB}.core.NodeJsCode.fromInline(`(
+					await (async () => {{ 
+						const tmp = new (require(\"${{self_client_path}}\")).{resource_name}({{{captured_fields}}}); 
+						if (tmp.{WING_INFLIGHT_INIT_NAME} !== undefined) await tmp.{WING_INFLIGHT_INIT_NAME}();
```
eladb, line 1088:
```suggestion
						if (tmp.{WING_INFLIGHT_INIT_NAME}) { await tmp.{WING_INFLIGHT_INIT_NAME}(); }
```

examples/tests/invalid/reassign_to_nonreassignable.w
yoav-steinberg: fixed

libs/wingc/src/type_check.rs
```diff
@@ -2070,6 +2071,7 @@ impl<'a> TypeChecker<'a> {
 				implements,
 				constructor,
```
staycoolcall911, line 2072: Consider renaming:
```suggestion
				initializer,
```

staycoolcall911 (commented):
Looks great - just one small comment from me.

libs/wingc/src/type_check.rs
```diff
@@ -159,7 +159,8 @@ pub enum Type {
 	Enum(Enum),
 }
 
-const WING_CONSTRUCTOR_NAME: &'static str = "init";
+const WING_INIT_NAME: &'static str = "init";
+pub const WING_INFLIGHT_INIT_NAME: &'static str = "$inflight_init";
```
Chriscbr, line 163: (optional)
```suggestion
const CLASS_INIT_NAME: &'static str = "init";
pub const CLASS_INFLIGHT_INIT_NAME: &'static str = "$inflight_init";
```

libs/wingc/src/type_check.rs
```diff
@@ -2068,8 +2069,9 @@ impl<'a> TypeChecker<'a> {
 				methods,
 				parent,
 				implements,
-				constructor,
+				initializer: constructor,
```
Chriscbr, line 2072: style: rename local variable to match (in VS code you can right click on "constructor" in VS Code and click "rename symbol" and type out "initializer" and it should automatically fix it)
```suggestion
				initializer,
```

libs/wingc/src/type_check.rs
```diff
@@ -2172,18 +2174,45 @@ impl<'a> TypeChecker<'a> {
 
 				// Add the constructor to the class env
 				let init_symb = Symbol {
-					name: WING_CONSTRUCTOR_NAME.into(),
-					span: name.span.clone(),
+					name: WING_INIT_NAME.into(),
+					span: name.span.clone(), // TODO: change to init's span
```
Chriscbr, line 2178: Opening a thread in case this needs to be fixed before merge

libs/wingc/src/type_check.rs
```diff
@@ -2172,18 +2174,45 @@ impl<'a> TypeChecker<'a> {
...
 				};
 				self.add_method_to_class_env(&constructor.signature, env, None, &mut class_env, &init_symb);
 
+				// Add the inflight initializer to the class env
+				let inflight_init_symb = Symbol {
+					name: WING_INFLIGHT_INIT_NAME.into(),
+					span: name.span.clone(), // TODO: change to inflight init's span
```
Chriscbr, line 2185: Fix

libs/wingc/src/visit.rs
```diff
@@ -206,7 +206,7 @@ where
 {
 	v.visit_symbol(&node.name);
 
-	v.visit_constructor(&node.constructor);
+	v.visit_constructor(&node.initializer);
```
Chriscbr, line 209: We also need to visit the inflight initializer?

(I see this is something we should automate, I've created an issue for it for now https://github.com/winglang/wing/issues/2043)

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -838,6 +838,44 @@ exports[`resource_inflight.w > stderr 1`] = `
...
+  |
+6 |   inflight init() {}
+  |   ^^^^^^^^^^^^^^^^^^ Multiple inflight initializers defined in class R
+
+
+error: Unexpected 'ERROR'.
+  --> ../../../examples/tests/invalid/resource_init.w:9:12
+  |
+9 |   inflight init(x: num) {}
+  |            ^^^^^ Unexpected 'ERROR'.
```
Chriscbr, line 860: Can we improve this error message (or create an issue if it seems out of scope of this PR)

libs/wingc/src/jsify.rs
Chriscbr: `_toInflight()` is still sync -- do you mean to say the code fragment returned by `toInflight()` is called from an async scope? Yes, I agree it feels like should that should be documented it in the `_toInflight()` API since it means you can't call initialize one of the clients from an ordinary JavaScript constructor for example.

Technically I guess this isn't a big change since that is already how we are invoking some inflight code right now (see below), but something to keep our eyes on if we ever need to initialize inflight code from plain javascript constructors for some reason.

```js
// this is the structure of the code that the SDK bundles into an AWS Lambda function
exports.handler = async function(event) {
  return await (< inflight._toInflight().text >).handle(event);
};
```

examples/tests/valid/resource.w
```diff
@@ -3,11 +3,16 @@ bring cloud;
 // User defined resource
```
Chriscbr, line 3: Can we add an example where a user defined resource has an inflight init function, and it has a child resource that also has an inflight init function (and validate that it's initialized properly etc.)? It's important to check that these behaviors work with more than one layer of composition

examples/tests/invalid/reassign_to_nonreassignable.w
```diff
@@ -5,17 +5,30 @@ x = x + 1;
...
   init() {
     this.f = 1;
   }
+
+  inflight init() {
+    this.inflight_f = 1;
+  }
   
   inc() {
     this.f = this.f + 1;
```
Chriscbr, line 19: Maybe we could also test a case like
`this.child_resource.f = 1 // this.child_resource.f is not reassignable`

Chriscbr (commented):
Looks really promising Yoav :)

libs/wingc/src/ast.rs
```diff
@@ -326,7 +337,8 @@ pub struct Class {
 	pub name: Symbol,
 	pub fields: Vec<ClassField>,
 	pub methods: Vec<(Symbol, FunctionDefinition)>,
-	pub constructor: Constructor,
+	pub initializer: Constructor,
+	pub inflight_initializer: Option<FunctionDefinition>,
```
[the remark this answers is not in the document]
eladb: +1 on always having an inflight initializer (even if it's empty). I agree that if we can reduce the amount of `Option<T>`s in our code, by definition we will have less code branches to handle downstream.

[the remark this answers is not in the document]
MarkMcCulloh: > Can we avoid the Option indirection?

The Option is technically accurate though, no? I understand that the end result is the same (for now), but wouldn't we rather the model itself be correct? Being forced to handle this case could even be a little micro-optimization, at the cost of an extra line or two of code.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -838,6 +838,44 @@ exports[`resource_inflight.w > stderr 1`] = `
...
+
+
+error: Unexpected 'identifier'.
+  --> ../../../examples/tests/invalid/resource_init.w:9:17
+  |
+9 |   inflight init(x: num) {}
+  |                 ^ Unexpected 'identifier'.
```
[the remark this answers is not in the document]
yoav-steinberg: Yes, good idea.

examples/tests/valid/resource.w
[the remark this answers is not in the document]
yoav-steinberg: it works, i'll add it to the test.

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: fixed by adding inflight init parameters as valid grammar and error reporting in _parser.rs_

examples/tests/valid/resource.w
yoav-steinberg: I agree, but we currently don't support a user defined resource inheriting from another user defined resource because of no `super()` support: #1123
I'll add a comment about that in the relevant issue.

Chriscbr: I think inheritance may be a separate use case than having one resource containing another resource

yoav-steinberg: > I think inheritance may be a separate use case than having one resource containing another resource

I think that case is covered in the _resource.w_ test.

Chriscbr: You're right, I missed it

eladb:
Cool!

Issue #1632, opened 2023-02-22
Struct with camelCase fields are not jsified properly


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

```js
// wing
bring "@cdktf/provider-aws" as aws;

new aws.s3Bucket.S3Bucket(
  versioning: aws.s3Bucket.S3BucketVersioning {
    enabled: true,
    mfa_delete: true,
  },
) as "Bucket";
```

### I expected this:

Preflight JS code should look like:
```js
new aws.s3Bucket.S3Bucket(this, "Bucket", { versioning: {
  "enabled": true,
  "mfaDelete": true
}});
```

### Instead, this happened:

Preflight JS code looks like:
```js
new aws.s3Bucket.S3Bucket(this, "Bucket", { versioning: {
  "enabled": true,
  "mfa_delete": true
}});
```

### Component:

Compiler

marciocadev:
@Chriscbr, I did some debugging and found something:

`{ enabled: true, mfa_delete: true, }`

This part of the code was interpreted as `ExprKind::StructLiteral`. From what I saw, the key parts always have to be in camel case, but that's not the case. The same problem happened with `ExprKind::MapLiteral` as well.

If this is correct, it should be easy to fix. I will open a PR.

marciocadev:
@Chriscbr I think it's a problem in jsify on `ExprKind::StructLiteral` and maybe on `ExprKind::MapLiteral` too

I will open a PR to `StructLiteral` now

staycoolcall911:
Related #1049

monadabot:
Congrats! :rocket: This was released in [Wing 0.9.2](https://github.com/winglang/wing/releases/tag/v0.9.2).

Issue #1049, opened 2023-01-05
Case-convert in JSII importing should not be two-way


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

Let's say I have a JSII library with the following API:

```ts
class Host {
  getIPAddress(): string;
}
```

When this is imported to Wing, it will be case-converted to: `get_ip_address()` which is dandy, so I can use it like so:

```js
bring my_lib;

let host = new my_lib.Host();
let ip = host.get_ip_address();
```

### I expected this:

The JSification should look like this (preserve the original casing):

```
const ip = host.getIPAddress();
```

### Instead, this happened:

I *suspect* (didn't check) that this won't work because when we JSify this code it will turn into something like this:

```js
const ip = host.getIpAddress(); // <-- notice "Ip" instead of "IP"
```

### Component:

Compiler

### Wing Version:

0.4.47

### Anything else?

This is a well-known issue in JSII implementations. Case conversion is by definition a lossy transformation.

The only *correct* solution is to persist metadata with the original symbol names inside the imported type-system and then use them in the JSification instead of case converting back.

yoav-steinberg:
We can, instead of storing a bool `should_case_convert` just, optionally, store the original name and use it. But there's also a theoretical edge case where the JSII lib has
```ts
class Host {
  getIPAddress(): string;
  get_ip_address(): IpAddress; 
}
```
And then we'll have a duplicate name in our env when importing the lib. So in general I'm against all of this an will gladly remove all case conversion stuff..
Let me know what you think..

eladb:
> We can, instead of storing a bool `should_case_convert` just, optionally, store the original name and use it during jsification. But there's also a theoretical edge case where the JSII lib has
> 
> ```ts
> 
> class Host {
> 
>   getIPAddress(): string;
> 
>   get_ip_address(): IpAddress; 
> 
> }
> 
> ```

Yes this is what I suggested above.

> 
> And then we'll have a duplicate name in our env when importing the lib. So in general I'm against all of this an will gladly remove all case conversion stuff..
> 
> Let me know what you think..

Wing is not special.🫶

The edge case is hypothetical because jsii depends on these case normalization for all other languages.

---

I am wondering if perhaps we can basically treat the original name as a `@macro`.

eladb:
@MarkMcCulloh ☝️

monadabot:
[...]

PR #2034, opened 2023-04-11
fix(compiler): jsify struct fields with case conversions

hasanaburayyan:
Added a `should_case_convert_jsii` to structs during jsii_importing (same as we do with classes).

**note:** ideally we will end up doing something like was suggested in https://github.com/winglang/wing/issues/1049 to store something like `orginal_name` but I think this requires a bit more work especially for fields in a structs. I think this works for now.

Closes: https://github.com/winglang/wing/issues/1632


yoav-steinberg (approved):
looks good.
I think we were gearing towards just removing all of this case conversion and allowing anycase because of the lossyness of the process. @staycoolcall911 lets add this as a topic for the next team meeting?

PR #2042, opened 2023-04-11
feat(sdk): cloud secret

jogold:
Add the `cloud.Secret` resource.

Currently, the secret value must be "manually" set using the CLI or
console of the targeted cloud provider.

It can then be used in a `cloud.Function` like this:

```js
bring cloud;

let secret = new cloud.Secret();

  let mySecretApiToken = secret.valueJson().get('token');
  // use the secret to call an api
});
```

The secret value is cached by default.


libs/wingsdk/src/shared-aws/permissions.ts
```diff
@@ -235,3 +235,36 @@ export function calculateBucketPermissions(
...
+        policy = {
+          effect: "Allow",
+          actions: ["secretsmanager:GetSecretValue"],
+          resources: [arn],
+        };
+        break;
+      case AwsTarget.TF_AWS:
+        policy = {
+          effect: "Allow",
+          action: ["secretsmanager:GetSecretValue"],
```
eladb, line 261: I am wondering if TF can accept `actions` and `resources` (plural) and then we can avoid this silly duplication everywhere (copying @marciocadev)

eladb (commented):
@jogold it's great to see you here! I missed your style :-)

libs/wingsdk/src/shared-aws/permissions.ts
marciocadev: I did this test, unfortunately the object used in CDK is different from TF. If you use the same one that is used in CDK, the policy will look like this when compiled to TF.
`"policy": "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":\"none:null\",\"Resource\":\"*\"}]}",`

With `Resource: *` and `Action: none:null`

If we use the TF object to generate the CDK, a compilation error occurs.
```bash
~/code/wing/apps/wing/bin/wing compile -t awscdk cdk.w
preflight error: Validation failed with the following errors:
  [stack/Default/Default/wing-function/Default/ServiceRole/DefaultPolicy] A PolicyStatement must specify at least one 'action' or 'notAction'.
  [stack/Default/Default/wing-function/Default/ServiceRole/DefaultPolicy] A PolicyStatement used in an identity-based policy must specify at least one resource.
```

jogold: Trying to solve this in #2048

libs/wingsdk/src/target-tf-aws/secret.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @inflight `@winglang/sdk.cloud.ISecretClient`
+ */
+export class Secret extends cloud.Secret {
+  private readonly secret: SecretsmanagerSecret;
+
+  constructor(scope: Construct, id: string, props: cloud.SecretProps = {}) {
+    super(scope, id, props);
+
+    this.secret = new SecretsmanagerSecret(this, "Default", {
+      name: ResourceNames.generateName(this, NAME_OPTS),
```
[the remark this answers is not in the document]
jogold: Interesting discussion 😄 

> Is this the common use case? This implementation means that the secret resource is provisioned by the Wing app and then someone needs to go and find it and put the value in it.
> 
> In my experience, a more common workflow is to go and manually create the secret (either through AWS Console or CLI) and then only reference it from your cloud application.

I've used both approaches. When the secret is created by IaC it has the advantages that its configuration (e.g. rotation or replica regions for AWS) can also be managed in the code.

Maybe we can start by not creating any cloud resource and have the following API:

```
bring cloud;

let secret = new cloud.Secret('my-secret-id');

  let mySecretApiToken = secret.valueJson().get('token');
  // use the secret to call an api
});
```

With this API there aren't differences between targets and it is the responsibility of the target to "lookup" the secret id?

eladb (approved):

PR #2045, opened 2023-04-11
fix(compiler): indentation in javascript output is a mess

eladb:
Migrate all relevant jsification code to `CodeMaker` so that now javascript output is nicely indented and much more human readable.

### Testing

I've manually reviewed snapshots and verified that only formatting changes were introduced. Use "hide whitespace" in GitHub to review.

### Implementation Notes

Any `jsify_xxx` function that is expected to return a block of code (statement, code block, etc) now returns a `CodeMaker` object instead of a string. Then, we can compose these objects together using `add_code()`.

The `jsify_expression()` function, and any other function that returns a part of a sentence still returns a string, but we have updated `CodeMaker` to handle multiline strings properly by breaking them down and adding each line to preserve the indentation of the parent.

When we have chained blocks (`if/elseif/else` or `try/catch/finally`), we do this Instead of putting the `else` in the same line as the closing brackets:

    if (cond1) {
      // code
    }
    else if (cond2) {
      // code
    }
    else {
      // code
    }

I think that's okay.

### Misc

This change uncovered a latent bug in how `try/finally` works. Previously we used to always generate a `catch` block, **even if the code did not have one**. This means that if we had only a `try/finally` block, then the exception will not be thrown. Nice!

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -2,7 +2,6 @@ mod codemaker;
 
 use aho_corasick::AhoCorasick;
 use const_format::formatcp;
-use indoc::formatdoc;
 use itertools::Itertools;
 
 use std::{
@@ -98,25 +97,25 @@ impl<'a> JSifier<'a> {
 		format!("\"./{}\".replace(/\\\\/g, \"/\")", path_name)
 	}
 
-	fn render_block(statements: impl IntoIterator<Item = impl core::fmt::Display>) -> String {
-		let mut lines = vec![];
-		lines.push("{".to_string());
+	// fn render_block(statements: impl IntoIterator<Item = impl core::fmt::Display>) -> String {
+	// 	let mut lines = vec![];
+	// 	lines.push("{".to_string());
 
-		for statement in statements {
-			let statement_str = format!("{}", statement);
-			let result = statement_str.split("\n");
-			for l in result {
-				lines.push(format!("  {}", l));
-			}
-		}
+	// 	for statement in statements {
+	// 		let statement_str = format!("{}", statement);
+	// 		let result = statement_str.split("\n");
+	// 		for l in result {
+	// 			lines.push(format!("  {}", l));
+	// 		}
+	// 	}
 
-		lines.push("}".to_string());
-		lines.join("\n")
-	}
+	// 	lines.push("}".to_string());
+	// 	lines.join("\n")
+	// }
 
 	pub fn jsify(&mut self, scope: &Scope) -> String {
-		let mut js = vec![];
-		let mut imports = vec![];
+		let mut js = CodeMaker::default();
+		let mut imports = CodeMaker::default();
 
 		for statement in scope.statements.iter().sorted_by(|a, b| match (&a.kind, &b.kind) {
 			// Put type definitions first so JS won't complain of unknown types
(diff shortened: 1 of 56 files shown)


tools/hangar/__snapshots__/test_corpus/valid/bring_jsii.w.ts.snap
```diff
@@ -177,14 +177,18 @@ class $App extends $AppBase {
...
-"async handle(m) { const { greeting } = this; {
-  {((cond) => {if (!cond) throw new Error(\`assertion failed: '(greeting === \\"Hello, wingnuts\\")'\`)})((greeting === \\"Hello, wingnuts\\"))};
-} };"
+exports[`wing compile -t tf-aws > proc.f9fd45e979c44914dac14967e8ad3a3b4c5beb85154a204a119e273326d0f3c0/index.js 1`] = `
+"async handle(m) {
+  const { greeting } = this;
+  {
+    {((cond) => {if (!cond) throw new Error(\`assertion failed: '(greeting === \\"Hello, wingnuts\\")'\`)})((greeting === \\"Hello, wingnuts\\"))};
+  }
+};
```
eladb, line 190: @Chriscbr do we really need this `;`? Feels like it's not supposed to be there.

tools/hangar/__snapshots__/test_corpus/valid/bucket_events.w.ts.snap
```diff
@@ -1412,49 +1417,73 @@ class $App extends $AppBase {
...
 `;
 
-exports[`wing compile -t tf-aws > proc.4c20a426ba35d2bc9530ee9673d6111365502af4b9943b5993189fef4c9cbb24/index.js 1`] = `
-"async handle(key) { const {  } = this; {
-  {console.log(\`updated \${key}\`)};
-} };"
+exports[`wing compile -t tf-aws > proc.0b6500d3229de22066b0d8ef0d85b80986dabd1f05d4adce56d1a2f16b46a58e/index.js 1`] = `
+"async handle(key) {
+  const { other } = this;
+  {
```
eladb, line 1427: @Chriscbr any idea why we are surrounding the body of handlers in an an inner block? I don't think that's providing any value.

tools/hangar/__snapshots__/test_corpus/valid/bring_jsii.w.ts.snap
Chriscbr: I don't think the last semicolon in this snapshot is needed, we omit similar ones when writing `async handle(m)` snippets in the SDK tests.

It sounds plausible that the semicolon at the end of the `assert` macro line could also be omitted, feel free to change as you see fit

tools/hangar/__snapshots__/test_corpus/valid/bucket_events.w.ts.snap
Chriscbr: Agree, I think it's just an implementation bug / shortcoming, I reckon we're using jsify_scope to produce the code for the body but we should be able to add a boolean param to turn off generating the surrounding braces, or something to that effect

libs/wingc/src/jsify.rs
```diff
@@ -98,25 +97,25 @@ impl<'a> JSifier<'a> {
 		format!("\"./{}\".replace(/\\\\/g, \"/\")", path_name)
 	}
 
-	fn render_block(statements: impl IntoIterator<Item = impl core::fmt::Display>) -> String {
-		let mut lines = vec![];
-		lines.push("{".to_string());
+	// fn render_block(statements: impl IntoIterator<Item = impl core::fmt::Display>) -> String {
```
Chriscbr, line 100: Guess we can remove this

Chriscbr (approved):
Nice! I had a hunch making `CodeMaker` composable would lead to something cool 😄. I wonder if we could do something similar for pretty-ifying the SDK code that's emitted for the simulator

libs/wingc/src/jsify.rs
```diff
@@ -188,31 +184,30 @@ impl<'a> JSifier<'a> {
...
+		output.to_string()
 	}
 
-	fn jsify_scope(&mut self, scope: &Scope, context: &JSifyContext) -> String {
-		let mut lines = vec![];
-		lines.push("{".to_string());
+	fn jsify_scope_inner(&mut self, scope: &Scope, context: &JSifyContext) -> CodeMaker {
+		let mut code = CodeMaker::default();
+
+		// code.open("{");
```
hasanaburayyan, line 201: can remove?

hasanaburayyan (approved):
💪 Hell yea!!

libs/wingc/src/jsify.rs
eladb: Yes!

tools/hangar/__snapshots__/test_corpus/valid/bucket_events.w.ts.snap
eladb: I've actually changed `jsii_scope` to `jsii_scope_body` so it's just the body now.

Issue #1213, opened 2023-01-22
add queue max retries / retention period


ainvoner:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Add max retries or retention period (or both) to a Queue

### Use Cases

When a queue handler function fails, the message keeps getting pushed back to the queue (infinitely).

We should provide a way to limit the amount of times a function handler is being pushed back to queue.

**Why is it an big issue for us now:**
Running the simulator in the Console with a queue having a faulty handler can cause the Console to crash since the function is being executed repeatedly (and logs are being sent) without the ability to stop it.  

### Component

SDK, Wing Console

ainvoner:
@Chriscbr any comments?

Chriscbr:
This is an important use case, feels like something we need for beta. 👍 I think max retries is the most straightforward solution. Off the top of my head I'm not sure if this needs to be a configurable property on the queue or on `QueueOnMessageProps` - I'd propose starting with the latter, and implementing it on AWS first to check how it works on their cloud.

> Running the simulator in the Console with a queue having a faulty handler can cause the Console to crash since the function is being executed repeatedly (and logs are being sent) without the ability to stop it.

Is it the number of logs that crashes the console? Or is it some kind of stack overflow within the simulator itself?

revitalbarletz:
Fynn is working on it! @staycoolcall911 can you send him an invite - here is his alias: fynnfluegge

fynnfluegge:
@ainvoner I suggest we need to add `retention period` and a `visibility timeout`. I would prefer adding default values to both. Any recommendations about the value? E.g. 1 day retention and 30s visibility is decent. However, I am not sure what the cause of the simulator crash is right now (smells like a stack overflow of the infinite loop) but the `visibility timeout` will very likely prevent the crash.

monadabot:
Congrats! :rocket: This was released in [Wing 0.14.1](https://github.com/winglang/wing/releases/tag/v0.14.1).

PR #2046, opened 2023-04-11
feat(sdk): add timeout handler to simulator queue

### LINKED ISSUE #165: Simulate queue timeout property
There is a "timeout" property that can be passed to the props of a `Queue` resource, but it has no affect on the current simulation.

A timeout of 5s means that if a consumer processes a message from a queue and takes more than 5 seconds before returning a successful response, then the response should get ignored and the message is added back to the queue.

fynnfluegge:
Closes #165.

Console crash mentioned in #1213 should also be fixed with this PR, but I let this issue open because `retentionPeriod` still needs to be added to terraform queue implementations.

To be consistent I also changed the default timeout period in the simulator queue to 10s, since `@default Duration.fromSeconds(10)` is documented in base class `cloud/queue.ts`.


libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -102,12 +104,27 @@ export class Queue implements IQueueClient, ISimulatorResourceInstance {
...
           });
-          this.messages.push(...messages);
+          this.pushMessagesBackToQueue(messages);
         });
         processedMessages = true;
       }
     } while (processedMessages);
   }
+
+  public async pushMessagesBackToQueue(messages: any): Promise<void> {
```
Chriscbr, line 114: Can we add a more specific type for `messages`?

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -102,12 +104,27 @@ export class Queue implements IQueueClient, ISimulatorResourceInstance {
...
+      this.context.addTrace({
+        data: {
+          message: `${messages.length} messages pushed back to queue after timeout.`,
+        },
+        sourcePath: this.context.resourcePath,
+        sourceType: QUEUE_TYPE,
+        type: TraceType.RESOURCE,
+        timestamp: new Date().toISOString(),
+      });
+    }, this.timeout * 1000);
```
Chriscbr, line 126: If a function takes 5 seconds to process a message and fail, and the queue's timeout is set for 10 seconds, should we still wait the full 10 seconds before reprocessing it, or should we only wait for the "remaining" time of 5 seconds, so to speak? (I haven't played with this in AWS SQS so I'm not sure off the top of my head how it works there, but figured I'd ask)

Chriscbr (commented):
Wow, looks great (also thanks for the small fixes along the way)! I had one small question, but otherwise I think everything looks good.

libs/wingsdk/src/target-sim/queue.inflight.ts
fynnfluegge: Good point, I was also thinking about this 😅
 Usually the behaviour is, that the subscriber will delete the message after successfull processing it. During the timeout period, after a subscriber fetched the message from the queue, it remains in the queue but cannot be requested from other subscribers. [This](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) explanation is really good. To 100% simulate this behaviour the simulator queue might need a rework. This is not trivial I think.
But yes, my fix would currently have a timeout period of 15s if the subscriber throws an error after 5s and and the timeout is 10s. Is it fine or should we simulate this more precise, what do you think?

libs/wingsdk/src/target-sim/queue.inflight.ts
fynnfluegge: Sure ✅

Chriscbr (approved):
🙏

libs/wingsdk/src/target-sim/queue.inflight.ts
fynnfluegge: I agree, it should be more accurate in the future, I will create a follow-up feature issue for this ✅

Issue #1700, opened 2023-03-02
Unexpected dependency cycle error


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

```js
bring cloud;

resource MyProcessor {
  b: cloud.Bucket;
  q: cloud.Queue;

  init() {
    this.q = new cloud.Queue();
    this.b = new cloud.Bucket();

    this.q.on_message(inflight (message: str) => {
      print("processing ${message}..");
      this.b.put("file.txt", message);
    });
  }

  inflight publish(s: str) {
    this.q.push(s);
  }
}

new MyProcessor();
```

### I expected this:

To compile

### Instead, this happened:

```
preflight error: Dependency cycle detected: root/p1/cloud.Queue => root/p1/cloud.Queue-OnMessage-dac81b4d => root/p1/cloud.Queue

  note: intermediate javascript code:
   }
   }
>> new MyApp().synth();
--------------------------------- STACK TRACE ---------------------------------
/Users/eladb/code/wing4/libs/wingsdk/lib/core/dependency.js:139
            throw new Error(`Dependency cycle detected: ${cycle
            ^

Error: Dependency cycle detected: root/p1/cloud.Queue => root/p1/cloud.Queue-OnMessage-dac81b4d => root/p1/cloud.Queue
    at DependencyVertex.addChild (/Users/eladb/code/wing4/libs/wingsdk/lib/core/dependency.js:139:19)
    at new DependencyGraph (/Users/eladb/code/wing4/libs/wingsdk/lib/core/dependency.js:51:27)
    at MyApp.synthSimulatorFile (/Users/eladb/code/wing4/libs/wingsdk/lib/target-sim/app.js:106:27)
    at MyApp.synth (/Users/eladb/code/wing4/libs/wingsdk/lib/target-sim/app.js:89:14)
    at evalmachine.<anonymous>:56:13
    at Script.runInContext (node:vm:141:12)
    at Object.runInContext (node:vm:297:6)
    at compile (/Users/eladb/code/wing4/apps/wing/dist/commands/compile.js:139:12)
```

### Component:

Compiler, SDK

### Wing Version:

0.5.81

Chriscbr:
When I compile the example to the `tf-aws` target I get this preflight code (excerpt):

```
class MyProcessor extends $stdlib.core.Resource {
  constructor(scope, id, ) {
  super(scope, id);
{
[...]

PR #2054, opened 2023-04-12
feat(sdk): sim event mappings

hasanaburayyan:
Updates to simulator that enables the use of event mappings. Similar to how AWS events are handled 
(see: [sns_topic_subscription](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sns_topic_subscription)  and [lambda_event_source_mapping](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_event_source_mapping))

This overhaul was needed to resolve the dependency cycle issues in the sim due to the way dependencies of child resources are implemented. (Closes: https://github.com/winglang/wing/issues/1700)

**note:**
After resolving the dependency issue it unmasked another issue (TODO: add issue after I create it)

- [x] Test this in the console (Ensure we dont display the event maps)
- [x] Create followup issue for `const { this } = this;` discovered problem
- [x] Im sure I could use some more comments


polamoros:
Should be implemented by the simulator? 🤔

libs/wingsdk/src/target-sim/queue.ts
```diff
@@ -71,13 +72,13 @@ export class Queue extends cloud.Queue implements ISimulatorResource {
...
-    this.node.addDependency(fn);
-
-    this.subscribers.push({
-      functionHandle: simulatorHandleToken(fn),
-      batchSize: props.batchSize ?? 1,
+    new EventMapping(this, `${this.node.id}-QueueEventMapping-${hash}`, {
+      subscriber: fn,
+      publisher: this,
+      eventSubscription: {
+        functionHandle: simulatorHandleToken(fn),
```
Chriscbr, line 79: Can we remove this field `functionHandle` since we already pass `fn` as a subscriber?

hasanaburayyan:
> Should be implemented by the simulator? 🤔

Is there an alternative?

libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -72,11 +74,26 @@ export interface QueueSchema extends BaseResourceSchema {
   };
 }
 
+export interface EventSubscription {}
```
Chriscbr, line 77: (a little bit above)

I think we can remove `subscribers` from `QueueSchema.props` since we're tracking all of the subscribers using the event subscriptions (this will be a breaking change but I think it's appropriate here) / they're no longer need to be provided by the Queue's `toSimulator()` method

libs/wingsdk/src/target-sim/event-mapping.ts
```diff
@@ -0,0 +1,82 @@
...
+import { bindSimulatorResource, makeSimulatorJsClient } from "./util";
+import { fqnForType } from "../constants";
+import * as core from "../core";
+import { Resource, App } from "../core";
+
+export const EVENT_MAP_FQN = fqnForType("sim.EventMapping");
+
+export interface EventProps {}
+
+export abstract class EventMappingBase extends Resource {
```
Chriscbr, line 19: Maybe we can remove this base class

libs/wingsdk/src/target-sim/event-mapping.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  constructor(scope: Construct, id: string, props: EventProps = {}) {
+    super(scope, id);
+    props;
+  }
+}
+
+export interface IEventPublisher {
+  addEventSubscription: (payload: any) => Promise<void>;
+}
```
Chriscbr, line 37: Let's make sure we convey somehow that this needs to be implemented by inflight clients, not preflight classes, eg
```suggestion
export interface IEventPublisher extends ISimulatorResourceInstance {
  addEventSubscription: (payload: any) => Promise<void>;
}
```

libs/wingsdk/src/target-sim/queue.ts
Chriscbr, line 1: I believe we can remove `this.subscribers` from this file, and similar for the topic

libs/wingsdk/test/target-sim/__snapshots__/file-counter.test.ts.snap
```diff
@@ -95,16 +95,24 @@ bucket: (function(env) {
...
+      {
+        "attrs": {},
+        "path": "root/HelloWorld/Queue/Queue-QueueEventMapping-401ee792",
+        "props": {
+          "eventSubscription": {
+            "batchSize": 1,
+            "functionHandle": "\${root/HelloWorld/Queue-AddConsumer-401ee792#attrs.handle}",
+          },
+          "publisher": "\${root/HelloWorld/Queue#attrs.handle}",
+          "subscriber": "\${root/HelloWorld/Queue#attrs.handle}",
```
Chriscbr, line 112: Bug?

hasanaburayyan:
> Should be implemented by the simulator?

@polamoros oh jeez nevermind me I cant read 😅

libs/wingsdk/src/target-sim/queue.ts
hasanaburayyan: I dont know if we can without making the assumption that every subscription will have the `functionHandle` property. Which for now is always true but what Im thinking of is like event mapping between a topic and a queue (which does not require an intermediate function)

Chriscbr: I wouldn't overdesign for other kinds of event mappings yet. If it's a concern, something you could consider is adding a field to the `eventSubscription` to indicate what the type of the mapping is or what the types of the resources are.

Since every event mapping knows who the subscriber is, the inflight code should be able to pass it along to all subscribers I think? eg
```ts
const client = this.context.findInstance(this.publisher) as IEventPublisher;
await client.addEventSubscription(this.subscriber, this.eventSubscription);
```

hasanaburayyan: I think Im on the same page here, made some updates let me know if they align.

libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -68,18 +70,31 @@ export interface QueueSchema extends BaseResourceSchema {
...
+
+/** Schema for sim.EventMapping */
+export interface EventMappingSchema extends BaseResourceSchema {
+  readonly type: typeof EVENT_MAPPING_TYPE;
+  readonly props: {
+    subscriber: FunctionHandle;
+    publisher: PublisherHandle;
+    subscriptionProps: EventSubscription;
```
Chriscbr, line 86: Docstrings

libs/wingsdk/src/target-sim/event-mapping.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class EventMapping extends Resource implements ISimulatorResource {
+  public static _newEventMapping(
+    scope: Construct,
+    id: string,
+    props: EventMappingProps
+  ): Event {
+    return App.of(scope).newAbstract(EVENT_MAP_FQN, scope, id, props);
+  }
```
Chriscbr, line 41: We can probably omit this static method

libs/wingsdk/src/target-sim/event-mapping.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const EVENT_MAP_FQN = fqnForType("sim.EventMapping");
+
+export interface EventMappingProps {
+  subscriber: core.IResource;
+  publisher: core.IResource;
+  subscriptionProps: EventSubscription;
+}
+
+export class EventMapping extends Resource implements ISimulatorResource {
```
Chriscbr, line 34: Docstring

@eladb It looks like one side effect of #2076 is since the target-specific modules aren't exported by JSII, our API checker no longer catches docstrings missing in them - thoughts?

libs/wingsdk/src/target-sim/event-mapping.ts
```diff
@@ -0,0 +1,77 @@
...
+import { fqnForType } from "../constants";
+import * as core from "../core";
+import { Resource, App } from "../core";
+import {
+  BaseResourceSchema,
+  ISimulatorResourceInstance,
+} from "../testing/simulator";
+
+export interface IEventPublisher extends ISimulatorResourceInstance {
+  addEventSubscription: (
```
Chriscbr, line 20: Docstrings

Chriscbr (approved):
LGTM

Issue #2059, opened 2023-04-13
Cannot use `await import()` to import ESM-only modules in inflight


eladb:
### I tried this

Install `nanoid` and write an extern that uses it:

```js
const { nanoid } = require('nanoid');

exports.make_id = async function() {
  return nanoid();
};
```

### I expected this:

To work

### Instead, this happened

```
require() of ES Module /Users/eladb/playground/nano-id/node_modules/nanoid/index.js from /Users/eladb/playground/nano-id/test.js not supported.
Instead change the require of index.js in /Users/eladb/playground/nano-id/test.js to a dynamic import() which is available in all CommonJS modules.
```

### Is there a workaround?

~Yes, you can use `async import()` inside your extern function to import ESM modules:~

Not working!

```js
exports.foo = async function() {
  let n = await import("nanoid");
  return n.nanoid();
};
```

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

MarkMcCulloh:
Inflight bundling solves this and would allow us to mostly ignore the ESM/CJS kerfuffle

eladb:
The workaround is not working here!

monadabot:
Congrats! :rocket: This was released in [Wing 0.17.26](https://github.com/winglang/wing/releases/tag/v0.17.26).

Issue #1388, opened 2023-01-31
cloud.Logger - add log levels


staycoolcall911:

staycoolcall911:
We're removing the Logger resource in #2060. Closing this as not planned.

Issue #619, opened 2022-11-15
cloud.Logger - GCP implementation


Chriscbr:

staycoolcall911:
Keep

staycoolcall911:
We're removing the Logger resource in #2060. Closing this as not planned.

PR #2060, opened 2023-04-13
feat(sdk): remove logger

eladb:
The `Logger` resource basically doesn't add value, only complexity. It is a great example of redundant future proofing, so let's just get rid of it and make the word a better place! We can always introduce it later when use cases are clearer.

Let me explain: the only way to emit a log right now in Wing is using the `log()` builtin function, which is jsified as a `console.log()` statement (both preflight and inflight). Now, we hijacked `console.log` in our runtime environments and mapped it to the standard logger of that environment.

In the cloud (e.g. AWS Lambda), the "standard logger" is basically `console.log()`, so basically the implementation of the logger for these environments is a "no-op".

In the simulator, since https://github.com/winglang/wing/pull/1879, we can listen to `console.xxx` events that come from the VM, and map them directly to the simulator trace system, so there's no need to hijack `console.log` here as well.

All in all, doctor says it's a classic (and severe) case of future proofing, where we get the exact same behavior with much less complexity and code to maintain (over 700 lines of code are not needed) ***== GOODNESS**.

## Misc

`node-fetch@3` is an ESM-only module and cannot be imported into our VM (damn!). In the meantime, I've downgraded it to 2.x and opened a bug: https://github.com/winglang/wing/issues/2059


staycoolcall911 (approved):
Looks good (1 minor comment). I went ahead and closed #618, #619, #1388.

libs/wingsdk/src/target-sim/app.ts
```diff
@@ -71,9 +67,6 @@ export class App extends core.App {
       case BUCKET_FQN:
         return new Bucket(scope, id, args[0]);
 
-      case LOGGER_FQN:
-        return new Logger(scope, id);
-
       case QUEUE_FQN:
```
ainvoner, line 70: not related to your change, but there are two `case TOPIC_FQN:` here (line 73 and 82)

eladb: Nice catch

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -39,7 +39,7 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
       message: `Invoke (payload=${JSON.stringify(payload)}).`,
       activity: async () => {
         const vm = new NodeVM({
-          console: "redirect", // we hijack `console.xxx` in `cloud/function.ts`
+          console: "redirect", // we hijack `console.xxx` in `cloud/function.ts`"
```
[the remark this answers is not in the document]
eladb: Nope

PR #2067, opened 2023-04-13
chore(sdk): remove compiled js files after tests run

hasanaburayyan:
I was pretty exhausted of parsing all the generated js and document files after running vitests so I thought it would be nice to clean those up in a teardown step. 

Recursively searches the src folder to blast those pesky files away.


libs/wingsdk/test/global.setup.ts
```diff
@@ -1,8 +1,86 @@
...
+export async function setup() {
   // compile src/**/*.on*.inflight.ts to .js because these are going to be
   // injected into our javascript vm and cannot be resolved via vitest
   execSync("npx tsc -p tsconfig.test.json");
   return () => {};
 }
+
+// remove all the compiled .js and .d.ts files after the tests are done
+// so that we don't have a ocean of files to sift through
+export async function teardown() {
```
MarkMcCulloh, line 14: I worry this adds a little bit of fragility/wasted time. I think it would be worth exploring a way to compile these files to a different directory and making sure vitest resolves it from there.

hasanaburayyan: @MarkMcCulloh I get the concern, I played around with vitest resolve aliases for a bit last night but did not have any success getting tests to pass. I assume it has to do with the way the inflight client requires do not get the the replacement applied. (at least that is what my testing seemed to reveal)

Im fine passing on this or bringing it in as a temp solution. I just think the generated mess right now is cumbersome.

MarkMcCulloh (approved):

Issue #2075, opened 2023-04-14
Exporting all target types in the SDK causes `bring cloud` to be slow


eladb:
### I tried this

```js
bring cloud;
```

### I expected this:

Under the hood, to only `require` the `cloud` submodules.

I think we need the primary `index` file of the SDK to only export the standard library modules and not the target submodules.

Instead of this:

```ts
export * as cloud from "./cloud";
export * as core from "./core";
export * as redis from "./redis";
export * as aws from "./shared-aws";
export * as std from "./std";
export * as awscdk from "./target-awscdk";
export * as sim from "./target-sim";
export * as tfaws from "./target-tf-aws";
export * as tfazure from "./target-tf-azure";
export * as tfgcp from "./target-tf-gcp";
export * as testing from "./testing";
```

It should be:

```ts
export * as cloud from "./cloud";
export * as core from "./core";
export * as redis from "./redis";
export * as std from "./std";
export * as testing from "./testing";
```

This implies that the code we are emitting at the app level should either move into the SDK (preferable) or explicitly "barrel-require" the relevant target:

```js

function __app(target) {
  switch (target) {
    case "sim":
      return $stdlib.sim.App;
    case "tfaws":
    case "tf-aws":
      return require("@winglang/sdk/lib/target-tf-aws/app").App;
    case "tf-gcp":
      return $stdlib.tfgcp.App;
    case "tf-azure":
      return $stdlib.tfazure.App;
    case "awscdk":
      return $stdlib.awscdk.App;
    default:
      throw new Error(`Unknown WING_TARGET value: "${process.env.WING_TARGET ?? ""}"`);
  }
}
```

If we move this code into the SDK, we will still wait to use runtime `require()` statements (or `async import()`) to avoid preloading the entire submodule if not needed.

### Instead, this happened

Since we require `@winglang/sdk`, and the primary `index` file exports all targets, we end up loading CDKTF and AWSCDK every time.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

eladb:
@Chriscbr what do you think?

eladb:
I'll pick this up and see if I can also find a good way to lazy-install the dependency to resolve #2035 along the way

monadabot:
Congrats! :rocket: This was released in [Wing 0.11.0](https://github.com/winglang/wing/releases/tag/v0.11.0).

PR #2076, opened 2023-04-14
fix(sdk)!: `bring cloud` is slow due to excessive exports

eladb:
By only exporting the standard library submodules () from the root `index` file of the SDK, we can avoid loading target-specific modules such as `cdktf` and `aws-cdk-lib`. This shaves off a couple 100ms of compile time and dramatically reduces the surface area of the SDK since it doesn't export any non-Wing types.

This fixes https://github.com/winglang/wing/issues/2075.

This change required a few tweaks:

1. The code emitted by the compiler to select the base `App` class has been moved to the SDK under `App.for(target)`, which dynamically `require`s the right submodule.

BREAKING CHANGE: The `sim`, `awscdk`, `tfaws`, `tfazure` and `tfgcp` submodules are no longer part of the Wing SDK API. It is technically possible to import them through a barrel import (e.g. `@winglang/sdk/lib/target-tf-gcp`).
* The `testing.SimApp` class (a test utility) which used to be part of the SDK has been moved to the `test` directory and is no longer available in the SDK API.
* The types `ISimulatorResourceInstance`, `WingSimulatorSchema`, `BaseResourceSchema` and `BaseResourceAttributes` are now available under the `testing.` submodule (instead of under `sim`, because `sim` is no longer part of the public API).


tools/hangar/__snapshots__/test_corpus/json.w/compile/tf-aws/clients/Foo.inflight.js
```diff
@@ -1,4 +1,4 @@
-class Foo  {
+class  Foo {
```
MarkMcCulloh, line 1: Since we're bothering to pretty-print stuff, should we avoid extra whitespace?

libs/wingsdk/src/index.ts
```diff
@@ -1,11 +1,6 @@
+// this file must only export the standard library types and not target-specific types.
 export * as cloud from "./cloud";
 export * as core from "./core";
 export * as redis from "./redis";
```
MarkMcCulloh, line 4: Not a blocker for this PR, but redis probably shouldn't be considered part of our "standard library". It's a specific implementation.

libs/wingsdk/src/core/app.ts
```diff
@@ -64,6 +64,20 @@ export abstract class App extends Construct {
...
 
+  /**
+   * Loads the `App` class for the given target.
+   * @param target one of the supported targets
+   * @returns an `App` class constructor
+   */
+  public static for(target: string): any {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      return require(`../target-${target}/app`).App;
```
MarkMcCulloh, line 75: To be clear, the purpose of this PR is just to reduce the time of `require('@winglang/sdk')` right? This doesn't effect any JSII loading later in wingc (`.jsii` still considered all the peer deps to be submodules)?

libs/wingsdk/src/index.ts
MarkMcCulloh, line 1: Based on [this](https://github.com/search?q=repo%3Awinglang%2Fconsole%20%40winglang%2Fsdk&type=code), it doesn't seem like this is a breaking change for the console, nice

tools/hangar/__snapshots__/test_corpus/resource.w/compile/tf-aws/proc.09db675f603249912771c7aaaae83de02fce6b9d53d66116afd98e7156363b46/index.js
MarkMcCulloh, line 1: It's weird that this file was suddenly created. Not an issue with this PR, something I can look into.

libs/wingsdk/test/global.setup.ts
```diff
@@ -3,6 +3,6 @@ import { exec, execSync } from "child_process";
 export default async function () {
   // compile src/**/*.on*.inflight.ts to .js because these are going to be
   // injected into our javascript vm and cannot be resolved via vitest
-  execSync("npx tsc -p tsconfig.test.json");
+  execSync("npx tsc -p tsconfig.test.json", { stdio: "inherit" });
```
MarkMcCulloh, line 6: 👍

libs/wingsdk/test/core/app.test.ts
```diff
@@ -32,6 +37,23 @@ test("newAbstract() throws if there is no implementation", () => {
...
 });
 
+describe("appForTarget", () => {
+  const map = {
+    sim: SimApp,
+    "tf-aws": TfAwsApp,
+    "tf-azure": TfAzureApp,
+    "tf-gcp": TfGcpApp,
+    awscdk: AwsCdkApp,
+  };
```
MarkMcCulloh, line 47: nit: probably a good use-case for https://vitest.dev/api/#test-each

libs/wingc/src/jsify.rs
```diff
@@ -41,26 +41,7 @@ const STDLIB_MODULE: &str = WINGSDK_ASSEMBLY_NAME;
...
-			return $stdlib.tfazure.App;
-		case "awscdk":
-			return $stdlib.awscdk.App;
-		default:
-			throw new Error(`Unknown WING_TARGET value: "${process.env.WING_TARGET ?? ""}"`);
-	}
-}
-const $AppBase = __app(process.env.WING_TARGET);
-"#;
+const TARGET_CODE: &str = "const $AppBase = $stdlib.core.App.for(process.env.WING_TARGET);";
```
MarkMcCulloh, line 44: Random thought: Why do we put `process.env.WING_TARGET` here? Isn't the compiler artifact expected to be target specific? 
wingc knows the target so it could explicitly put it here.

MarkMcCulloh (approved):

tools/hangar/__snapshots__/test_corpus/json.w/compile/tf-aws/clients/Foo.inflight.js
eladb: Yes, not sure where this came from... Any idea?

libs/wingsdk/src/index.ts
eladb: We intend to extract it completely to a 3rd party library in the future but for now we decided that it will be shipped with the SDK under the `redis` submodule.

libs/wingsdk/src/core/app.ts
eladb: yes, not yet :-)

libs/wingsdk/src/index.ts
eladb: Yes, I verified this.

libs/wingc/src/jsify.rs
eladb: yes, good point. I can look into that in a subsequent PR.

eladb: https://github.com/winglang/wing/issues/2077

libs/wingsdk/test/core/app.test.ts
eladb: I tried... Since this is a map, the resulting code is much less readable in IMHO.

tools/hangar/__snapshots__/test_corpus/resource.w/compile/tf-aws/proc.09db675f603249912771c7aaaae83de02fce6b9d53d66116afd98e7156363b46/index.js
eladb: 10x

libs/wingsdk/test/core/app.test.ts
MarkMcCulloh: You shouldn't need a map anymore, how does this look to you?

```ts
test.each([
  ["sim", SimApp],
  ["tf-aws", TfAwsApp],
  ["tf-azure", TfAzureApp],
  ["tf-gcp", TfGcpApp],
  ["awscdk", AwsCdkApp],
])("App.for(%s) resolves to correct type", (target, app) => {
  expect(App.for(target).constructor).equal(app.constructor);
});
```

hasanaburayyan:
>The types ISimulatorResourceInstance, WingSimulatorSchema, BaseResourceSchema and BaseResourceAttributes are now available under the testing. submodule (instead of under sim, because sim is no longer part of the public API).

@eladb curious this should have an impact on the console no?

Issue #2078, opened 2023-04-14
Api simulator does not support urlencoded body


RaphaelManke:
### I tried this

```ts
bring cloud;
let api = new cloud.Api();
let EMPTY_JSON = Json { empty: "https://github.com/winglang/wing/issues/1947" };

api.post("/", inflight (request : cloud.ApiRequest): cloud.ApiResponse => {
    let body = request.body ?? EMPTY_JSON;

    let resp = cloud.ApiResponse{
        status: 200,
        body: body,
    };
    return resp;
});
```

### I expected this:

When calling this api post endpoint the response body is a json object of the input. 

### Instead, this happened

When posting an api call to the post endpoint with an `x-urlencoded-form` body the response is an empty body `{}`

### Is there a workaround?

no

### Component

SDK

### Wing Version

0.10.1

### Node.js Version

v18.13.0

### Platform(s)

MacOS

### Anything else?

Can be fixed by add the urlencoded middleware to the express server 
`this.app.use(express.urlencoded({ extended: true }));`

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.11.1](https://github.com/winglang/wing/releases/tag/v0.11.1).

PR #2079, opened 2023-04-14
feat(sdk): add url-encoded body support  for api

RaphaelManke:
Add support for api calls that use the the `x-www-form-urlencoded` format in the body. 

The form is normalized to a json object so that the handler code does not need manual parsing.

fixes: #2078


libs/wingsdk/src/shared-aws/api.onrequest.inflight.ts
```diff
@@ -50,12 +51,31 @@ function mapApigatewayEventToCloudApiRequest(
     ...request.queryStringParameters,
     ...request.multiValueQueryStringParameters,
   };
+
+  const body = parseBody(request);
   return {
     path: request.path,
-    body: request.body ? JSON.parse(request.body) : "",
+    body: body,
```
MarkMcCulloh, line 58: nit:

```suggestion
    body: parseBody(request),
```

libs/wingsdk/src/shared-aws/api.onrequest.inflight.ts
```diff
@@ -50,12 +51,31 @@ function mapApigatewayEventToCloudApiRequest(
...
+ * Parse body to JSON or empty string
+ * @param body body
+ * @returns JSON body
+ */
+function parseBody(request: APIGatewayProxyEvent) {
+  if (!request.body) return "";
+
+  const contentType = Object.entries(request.headers)
+    .filter(([key, _]) => key.toLowerCase() === "content-type")
+    .map(([_, value]) => value?.toLowerCase())[0];
```
MarkMcCulloh, line 76: filter isn't needed since we don't need to go through all the headers

```suggestion
  const contentType = Object.entries(request.headers)
    .find(([key, _]) => key.toLowerCase() === "content-type")?.[1];
```

libs/wingsdk/src/shared-aws/api.onrequest.inflight.ts
```diff
@@ -1,3 +1,4 @@
+import querystring from "querystring";
```
MarkMcCulloh, line 1: Should we use `URLSearchParams` instead?

RaphaelManke: makes sense since it seems to be the new way of doing it 😃

libs/wingsdk/src/shared-aws/api.onrequest.inflight.ts
MarkMcCulloh: We should probably normalize headers at some point. The local sim probably already toLower()s them via express since headers are not supposed to be case sensitive.

MarkMcCulloh (approved):

Chriscbr:
neat, TIL about this field

Issue #1833, opened 2023-03-16
assert doesn't work with string interpolation 


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

```ts (wing) 
let s = "hello";
if "${s}" == "${s}" {  } // compiles and works
assert("" == ""); //compiles and works
assert(s == s); //compiles and works
assert("${s}" == "${s}"); // doens't compile due to backticks I assume
```

```sh
wing compile -t  sim  a.w
```

### I expected this:

to compile and work 

### Instead, this happened:

```sh
wingly % wing compile -t  sim  story13.w
preflight error: missing ) after argument list

  note: intermediate javascript code (/private/tmp/wingly/target/.wing/preflight.js):
     {((cond) => {if (!cond) throw new Error(`assertion failed: '("" === "")'`)})(("" === ""))};
     {((cond) => {if (!cond) throw new Error(`assertion failed: '(s === s)'`)})((s === s))};
>>   {((cond) => {if (!cond) throw new Error(`assertion failed: '(`${s}` === `${s}`)'`)})((`${s}` === `${s}`))};
   }
   }
  note: run with `NODE_STACKTRACE=1` environment variable to display a stack trace
```

### Component:

Compiler

### Wing Version:

0.5.134

### Anything else?

It is probably related to the backticks in assert implementation. 

Maybe we should lint backticks out from our implementation or solve this issue from the ground up

Chriscbr:
I think change the `assert` macro so it doesn't use backticks should probably fix it.

We should also add tests for other cases (like interpolated strings literals containing other interpolated string literals) for completeness

ekeren:
My problem with changing this only for the assert code is that this is the second time this happen to us. And I can see how it will happen again when someone (external or internal) creates another macro

Chriscbr:
We could check that no backticks are used in macros to prevent these kinds of errors (compile time error would say e.g. "backticks don't compose well, change your macro to join strings a different way")

neilfarmer:
Hey! wanted to try tackling an issue. I just put up a PR for this but something to note if its accepted is that not using backticks makes some issues for multiline things in js

example:
```js
assert((():num => { try {return 3;} finally {}})() == 3);
``` 
has to be changed to 
```js
let return_from_try = (():num => { try {return 3;} finally {}})();
assert(return_from_try == 3);
```

[...]

PR #2091, opened 2023-04-15
fix(compiler): allow string interpolation asserts

neilfarmer:
Changed built in assertion function to use single quotes instead of backticks.

**Note**: If this PR is accepted, I wanted to mention that I had to refactor the `try_catch.w` tests to create a variable for each closure because multiline expansion within an assertion wont work with single quotes and must use backticks
I made additional comments here: https://github.com/winglang/wing/issues/1833#issuecomment-1509928472

Closes: https://github.com/winglang/wing/issues/1833


eladb (commented):
LGTM. @Chriscbr i will let you chime in

Chriscbr (changes requested):
Using single quotes to fix the string interpolation looks ok, except it seems to introduce another implementation bug -- see this example:

```
assert("'" == "'");
```

will compile to this invalid JS:

```js
{((cond) => {if (!cond) throw new Error('assertion failed: "("'" === "'")"')})(("'" === "'"))};
```

libs/wingc/src/lib.rs
```diff
@@ -199,7 +199,9 @@ pub fn type_check(
 			parameters: vec![types.bool()],
 			return_type: types.void(),
 			phase: Phase::Independent,
-			js_override: Some("{((cond) => {if (!cond) throw new Error(`assertion failed: '$args$'`)})($args$)}".to_string()),
+			js_override: Some(
+				"{((cond) => {if (!cond) throw new Error('assertion failed: \"$args$\"')})($args$)}".to_string()
+			),
```
Chriscbr, line 204: Try this perhaps?
```suggestion
			js_override: Some(
				r#"{((cond) => {if (!cond) throw new Error("assertion failed: \"" + $args$ + "\"")})($args$)}"#.to_string()
			),
```

eladb: I think we need to escape the internal string.

A nice trick is to use `JSON.stringify()`. It will return a quoted string that you can just plug into the other string.

neilfarmer: Thanks guys! We have tried both of those options (JSON.stringify and raw strings) but they seem to break some other tests. Ive been pairing with @hasanaburayyan on a fix for this. We are still trying to come up with a robust solution for this.

hasanaburayyan:
spoke with @neilfarmer @staycoolcall911 and @Chriscbr closing this PR for now in favor of pulling another issue. Right now it seems that all solutions proposed break something else. Maybe we need to revisit the assert implementation for a more robust solution.

Issue #2035, opened 2023-04-11
Wing installation is not a good experience


eladb:
### I tried this

```sh
npm install winglang
```

### I expected this:

Since winglang is currently is a pretty big package (and I am not sure we will really have time to take care of this before beta), I think we should make a simple installer that gives people a better indication of what's going on and maybe even has some ASCII entertainment in it.

Something like:

```sh
$ /bin/bash $(curl http://get.winglang.io)
```

And then design a cute little ASCII experience, colorful, emojis, progress bars, spinners, that stuff.

It will also be nice to automatically install node if needed, and other stuffs like that.

### Instead, this happened

It took about 5 hours to finish without any entertainment

### Component

CLI

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

staycoolcall911:
The better solution IMO is handling #1800

Chriscbr:
After [reducing the SDK's package size](https://github.com/winglang/wing/pull/2306), the install time can still take over 2 minutes (I tried a fresh install on a base-level GitHub codespace with 2 cores, 4GB of RAM, 32GB storage). A dedicated installer sounds like it could still be a good idea.

eladb:
I believe another low hanging fruit is to remove aws-cdk-lib from the package and require users to install it only if they use this target.

eladb:
There we go: https://github.com/winglang/wing/pull/2354

monadabot:
Congrats! :rocket: This was released in [Wing 0.15.5](https://github.com/winglang/wing/releases/tag/v0.15.5).

Issue #1780, opened 2023-03-10
Consolidate builtin and `std` types


MarkMcCulloh:
Several builtin types are represented as classes in the stdlib as well as precreated-builtin-types in wingc. The latter is useful for optimizations in the compilers (easy matches and special treatment) while the former is useful for defining APIs and macros.

This discrepancy has caused the need for several workaround to make sure we flip between both contexts as needed (look for any stringy references to std types to find them). We should find a way to simplify this situation.

PR #2095, opened 2023-04-15
feat!: expose base resource and construct apis

### LINKED ISSUE #1595: Cannot use `Resource` APIs in Wing
### I tried this:

For example, I want to be able to update the [`display`](https://github.com/winglang/wing/blob/5332d6ecdb16362ced14b709df916f16e0cb36cd/libs/wingsdk/src/core/resource.ts#L30) properties of the resource:

```js
let b = new cloud.Bucket();
b.display.title = "MyBucket";
```

### I expected this:

To work

### Instead, this happened:

Unknown symbol.

### Component:

Compiler

### Wing Version:

0.5.37

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

@yoav-steinberg said:

> The JSII importer respects base class methods, but `core.Resource` is a special case. It's actually never imported.

Related: https://github.com/winglang/wing/issues/261

### Community Note

---

### LINKED ISSUE #2028: Unable to reference `constructs.Construct`
### Community Note

### I tried this:

Install `constructs`:

```sh
npm i constructs
```

```js

let get_id = (c: constructs.Construct) => {
  return c.node.id;
};
```

### I expected this:

To work

### Instead, this happened:

```
error: Unknown symbol "Construct" in namespace "constructs"
  --> hello.w:3:29
  |
3 | let get_id = (c: constructs.Construct): str => {
  |                             ^^^^^^^^^ Unknown symbol "Construct" in namespace "constructs"
```

### Is there a workaround?

Not that I'm aware of

### Component:

Compiler

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

eladb:
Moved the `Resource` class from `core.Resource` (which is a hidden namespace) to `std.Resource` (which is for visible built-in types).

Removed much of the special-casing related to short-circuiting `Construct` (resolves #2028) and `Resource` during the jsii import and type checking.

To resolve #1595, if a `resource` is defined without a base class, implicitly associate `std.Resource` as a parent.

Moved the JSII `TypeSystem` object to be shared across JSII imports. This object is designed to be used as a mutable global cache for imported JSII types. Previously, every time a module was imported under a different alias, we reloaded the JSII typesystem, which is not needed. This also means that we don't need to persist `jsii_imports` because they are now basically just a set of specs and there isn't much value in persisting them across `TypeChecker` instantiations (@MarkMcCulloh can you validate this assumption please?).

As a follow up for #2076 in order to simplify the jsii importer so we can simply import all dependencies of the SDK, we have converted all the target-specific SDK dependencies to bundled dependencies instead of normal deps. This is fine because none of these types are exposed in the SDK's API, so JSII is happy. This dramatically decreases the public dependency surface area of the SDK and later on we can follow up with installing these dependencies just in time to reduce the package size (related to #2035).

Simplified wingii APIs by removing the concept of `roots` (not used) and the option to not load `deps`. This is not needed now because our SDK is only dependent on `constructs` which we want to load (because we want to allow folks to access its API).

### Example

```js
bring cloud;
bring "constructs" as cx;
bring "aws-cdk-lib" as aws;

resource WingResource {
  init() {
    // access construct node
    log("my id is ${this.node.id}"); 
  } 
}

// reference Construct
let get_path = (c: cx.Construct): str => {
  return c.node.path; 
};

// reference std.Resource
let get_display_name = (r: std.Resource): str? => {
  // reference resource API
  return r.display.title;
};

let q = new aws.aws_sqs.Queue();
let wr = new WingResource();

// wr is-a std.Resource
let another_resource: std.Resource = wr;

// queue and wr are Constructs
log("path of sqs.queue: ${get_path(q)}");
log("path of wing resource: ${get_path(wr)}");

let title = get_display_name(wr) ?? "no display name"; log("display name of wing resource: ${title}");
```

### Follow ups

- #2097
- #2099
- #2098

### Breaking Changes

BREAKING CHANGE: `core.Resource` is now `std.Resource` (as well as other types that were in that file).


libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -83,7 +88,7 @@ impl<'a> JsiiImporter<'a> {
 				PrimitiveType::Boolean => self.wing_types.bool(),
 				PrimitiveType::Any => self.wing_types.anything(),
 				PrimitiveType::Json => self.wing_types.json(),
-				PrimitiveType::Date => todo!(),
+				PrimitiveType::Date => self.wing_types.anything(), // TODO!!!!!!!!!!!!!!!!!!
```
eladb, line 91: Without this we can't really import `aws-cdk-lib`, so I guess `anything()` is what we need right now, but we need to remember this :-) I'll open an issue.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -101,9 +106,6 @@ impl<'a> JsiiImporter<'a> {
 					self.wing_types.json()
 				} else if type_fqn == &format!("{}.{}", WINGSDK_ASSEMBLY_NAME, WINGSDK_MUT_JSON) {
 					self.wing_types.mut_json()
```
eladb, line 108: Can we remove the special casing here for all the std types?

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -148,43 +150,32 @@ impl<'a> JsiiImporter<'a> {
 	}
 
 	pub fn import_type(&mut self, type_fqn: &FQN) -> bool {
-		// Hack: if the class name is a construct base then we treat this class as a resource and don't need to define it
-		if is_construct_base(&type_fqn) {
-			return true;
-		}
-
+		println!("import_type {}", type_fqn.as_str());
```
eladb, line 153: Remove

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -771,9 +753,9 @@ impl<'a> JsiiImporter<'a> {
 
 	/// Imports submodules of the assembly, preparing each as an available namespace
 	pub fn import_submodules_to_env(&mut self, env: &mut SymbolEnv) {
+		println!("Importing submodules... for {} (alias {})", self.jsii_spec.assembly_name, self.jsii_spec.alias);
```
eladb, line 756: Remove

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -794,20 +776,22 @@ impl<'a> JsiiImporter<'a> {
...
-			self
-				.wing_types
-				.libraries
-				.define(
-					&Symbol::global(assembly.name.clone()),
-					SymbolKind::Namespace(ns),
-					StatementIdx::Top,
-				)
-				.expect("Failed to define jsii root namespace");
+			// println!("defining root namespace for {}", assembly.name);
```
eladb, line 779: Seems like this is just creating trouble when I am trying to import `constructs` as a different alias.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -836,6 +820,11 @@ impl<'a> JsiiImporter<'a> {
 	}
 
 	fn register_jsii_type(&mut self, fqn: &FQN, symbol: &Symbol, type_ref: TypeRef) {
+		// skip if this type is already registered
```
eladb, line 823: We have multiple code paths that get here

libs/wingii/src/lib.rs
```diff
@@ -162,95 +144,60 @@ pub mod type_system {
...
-			if Path::new(file_or_directory).is_dir() {
-				self.load_module(file_or_directory, &opts)
-			} else {
-				// load_file always loads a single manifest and never recurses into dependencies
-				self.load_file(file_or_directory, Some(true))
-			}
-		}
+		fn load_assembly(&self, path: &str) -> Result<Assembly> {
+			// start a timer
+			let start = Instant::now();
```
eladb, line 149: remove instrumentation

libs/wingc/src/type_check.rs
```diff
@@ -2092,6 +2093,17 @@ impl<'a> TypeChecker<'a> {
 							(None, None)
 						}
 					}
+				} else if *is_resource {
+					let resource_fqn = format!("{}.{}", WINGSDK_ASSEMBLY_NAME, WINGSDK_RESOURCE);
```
eladb, line 2097: Cache this in `self.types` (like other standard types)

libs/wingc/src/type_check/jsii_importer.rs
eladb, line 91:
```suggestion
				PrimitiveType::Date => self.wing_types.anything(), // TODO: https://github.com/winglang/wing/issues/2102
```

Chriscbr (commented):
Looks good to me, leaving unapproved so @MarkMcCulloh and others can also take a look

random note: I feel like the distinction between `core` and `std` is blurring a bit / I'm not sure how meaningful it is to keep them separated. Maybe we could just combine them, and if there are types from `core`/`std` we do not want public, we can just mark them with `@internal`

MarkMcCulloh (approved):
:shipit: 

> This also means that we don't need to persist jsii_imports because they are now basically just a set of specs and there isn't much value in persisting them across TypeChecker instantiations (@MarkMcCulloh can you validate this assumption please?).

I don't think there is an issue with this. I noted on https://github.com/winglang/wing/pull/1982 for this to be a followup change but didn't make an issue for it (shame on me)

> I feel like the distinction between core and std is blurring a bit

Fully agree. I feel like when std was added 100 years ago there was a useful reason for the distinction but now I don't quite see it. I think there is a solution somewhere in here that can also relate to https://github.com/winglang/wing/issues/1780.

PR #2100, opened 2023-04-16
rfc(sdk): website

tsuf239:
Creating an RFC for the website resource.

Since this resource is included in the spec, and the implementation is fairly easy, I skipped some of the sections in the template. If any essential information is missing, please comment below and I'll add some further information for the missing section :) 


docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+Author(s): Git tag for PR authors and contributors
+Submission Date: Fill in with date of initial submission, YYYY-MM-DD
+Stage: Fill in with current stage in the RFC lifecycle
+Stage Date: Fill in with date of last stage change
+-->
+
+> Implementing the website resource following the [SDK spec](../../04-reference/wingsdk-spec.md#website).
+
+<!--
+This RFCs have 3 sections: Requirements -> Design -> Implementation. We intentionally start with *Design* since it
```
staycoolcall911, line 25: I'd expect to see these 3 sections. If the SDK spec contains most of the Requirements and Design then I would at least say so. It does feel like this RFC jumps too quicly into the details without providing any introduction

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+
+Following a few AWS guides on the internet (brought in the end of the section), it seems that the common way to serve a static website is by using a bucket.
+
+After creating a public bucket, we'll upload the folder from the relative path.
+AWS requires a index document name (to serve on the "/" path).
+We'll start by hardcoding it to be "index.html", and can add it later to the website props upon request.
+
+The last thing will be a registration to the chosen domain (if exists).
+
+The website resource contains preflight code only.
```
staycoolcall911, line 41: Maybe change to 
```suggestion
The website resource supports preflight code only.
```

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Implementation for aws
+
+Following a few AWS guides on the internet (brought in the end of the section), it seems that the common way to serve a static website is by using a bucket.
+
+After creating a public bucket, we'll upload the folder from the relative path.
+AWS requires a index document name (to serve on the "/" path).
+We'll start by hardcoding it to be "index.html", and can add it later to the website props upon request.
+
+The last thing will be a registration to the chosen domain (if exists).
```
staycoolcall911, line 39: What about a CDN? It appears on the sdk spec as a requirement, so maybe we should do something similar to what we did when we added a VPC for the redis resource

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+const app = Express();
+
+app.use(express.static("<resolved path to the website files folder>"));
+
+app.listen(
+  "<port, ideally will be specified by the user, otherwise generated>"
+);
+```
+
+Since the website is served on the local machine, we cannot specify a domain while deploying to the simulator.
```
staycoolcall911, line 62:
```suggestion
Since the website is served on the local machine, we cannot specify a domain, nor can we add a CDN while deploying to the simulator.
```

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+
+After creating a public bucket, we'll upload the folder from the relative path.
+AWS requires a index document name (to serve on the "/" path).
+We'll start by hardcoding it to be "index.html", and can add it later to the website props upon request.
+
+The last thing will be a registration to the chosen domain (if exists).
+
+The website resource contains preflight code only.
+
+#### resources:
```
staycoolcall911, line 43: resources is an overloaded word, maybe use "References" instead
```suggestion
#### References:
```

staycoolcall911 (commented):
I love this initiative @tsuf239 - thanks for picking it up. Left a few comments after an initial review

ShaiBer (commented):
Nice!
Personally I like to avoid .env files whenever I can (one of the benefits of the Wing compiler and capture mechanism is to avoid having to pass resource ids through env vars :) ).
So I would accept the domain in the constructor and leave it up to the dev whether to use an env var to know what to pass to the ctor or use something else

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+<!--
+This RFCs have 3 sections: Requirements -> Design -> Implementation. We intentionally start with *Design* since it
+is a great way to introduice the feature to readers.
+-->
+
+## Implementation
+
+### Implementation for aws
+
+Following a few AWS guides on the internet (brought in the end of the section), it seems that the common way to serve a static website is by using a bucket.
```
eladb, line 33: It's usually a bucket and a cloud front distribution

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+
+After creating a public bucket, we'll upload the folder from the relative path.
+AWS requires a index document name (to serve on the "/" path).
+We'll start by hardcoding it to be "index.html", and can add it later to the website props upon request.
+
+The last thing will be a registration to the chosen domain (if exists).
+
+The website resource supports preflight code only.
```
eladb, line 41: I think being able to perform simple `get` requests to retrieve static files from the website could be a useful inflight api as well.

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+
+- Website props are brought here for reference:
+
+```ts
+struct WebsiteProps {
+  /**
+   * Local path to the website's static files, relative to the Wing source file.
+   * @example "./dist"
+   */
+  path: str;
```
eladb, line 102: Thought: ideally the path should be relative to the source file (and not to the current working directory).

What I mean is that if I pass `./site` I actually want the sdk to receive the path `__dirname + "./site"`. Ideally this should be implicit, so we might need some language support for this.

@Chriscbr what do you think?

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * The website's custom domain name.
+   * @example "example.com"
+   * @default - a domain is generated by the cloud provider
+   */
+  domain: str?;
```
eladb, line 109: I feel we need to model domains through a `Domain` resource. For example I'd want to be able to connect an `API` to a domain...

eladb (commented):
Cool stuff!

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
```diff
@@ -0,0 +1,111 @@
...
+BO_WEBSITE_DOMAIN= "localhost:4000"
+```
+
+This, of course, rises more questions then answers-
+how can we support different development environments? How do we tell wing which env file to use? Can we share those files between targets? (for example, running the same file for tf-gcp and tf-azure, and a different one for the sim).
+
+- Website props are brought here for reference:
+
+```ts
+struct WebsiteProps {
```
eladb, line 97: We will likely also want to support 

- CORS
- deploying dynamic data into the website

See a real world example of how to setup a website using the aws cdk: https://github.com/cdklabs/construct-hub/blob/main/src/webapp/index.ts

A construct library: https://constructs.dev/packages/cdk-website/v/0.0.5?lang=typescript

eladb (commented):
Added some useful references.

It's not trivial to set up a website on aws!

tsuf239:
I looked into CloudFront, we can specify another bucket to see logs, should I add it to the websiteProps/ open an issue to add it later if needed? 
@eladb @ShaiBer @staycoolcall911

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
tsuf239: can you elaborate more on "deploying dynamic data into the website"? 
there are more configuration option that we may want to let the user decide on (regarding the cdn)
image
should I add some of them to the props too?

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
tsuf239: I totally agree!! 
would it be something like that? 
```
let d = new cloud.Domain("my-domain.com", {options});
let website = new cloud.Website({path: "../build", domain: d.domain });
let API = new cloud.Api({domain: "${d.domain}/api/" });
```

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
tsuf239: I started coding this to see if any unexpected issues are popping up, and one of the first things I notice is that I can't use relative paths to the .w files yet, since the SDK is located in the node_modules (maybe even the global ones) and "has no idea" where is the .w located, I might need some help with that :)
(currently using an absolute path to bypass it, but it isn't nice...)

tsuf239:
> Nice! Personally I like to avoid .env files whenever I can (one of the benefits of the Wing compiler and capture mechanism is to avoid having to pass resource ids through env vars :) ). So I would accept the domain in the constructor and leave it up to the dev whether to use an env var to know what to pass to the ctor or use something else

does wing support env vars?

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
tsuf239: In a "fetch" like operation? or "bucket.get" like?  if we'll reach it via "bucket.get" won't we lose the benefits of the caching via the CDN? 
It seems that an inflight fetch that is native to the language can be nice anyways :) 
(for limiting a common use of extern, and for this [issue](https://github.com/winglang/wing/issues/1850) as well )

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
Chriscbr: Another option could be to pass the domain directly, so the internal implementation can decide what fields to use / how to glue things together

```
let d = new cloud.Domain("my-domain.com", {options});
let website = new cloud.Website(path: "../build", custom_domain: d);
let API = new cloud.Api(custom_domain: d, custom_domain_path: "v1");
```

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
Chriscbr: I think Elad's thought is what I tried to include in the SDK spec 😅 sorry if the wording wasn't clear.

One idea for how to implement this could be to set a `WING_SOURCE_DIR` environment variable inside the `wing` CLI package [here](https://github.com/winglang/wing/blob/9db4d03c13f5689dca056599bdf368583235fc7e/apps/wing/src/commands/compile.ts#L97)-ish, and assign to it the absolute path to the directory, and then use the environment variable within the SDK code through `process.env` to create the full path name.

tsuf239: Works perfectly, thanks @Chriscbr!!

docs/06-contributors/999-rfcs/2023-04-16-website-resource.md
staycoolcall911: The requirements are the "contract" of the feature you're developing - "what does it do?" (as opposed to "how does it do it" - the implementation). The requirements usually specify use cases as well as edge case scenarios and the desired behavior of the software described.
Practically speaking, you can take a look at other resources in the SDK spec - some good examples are the first paragraphs of the [Function resource](https://docs.winglang.io/reference/wingsdk-spec#function) and the [Queue resource](https://docs.winglang.io/reference/wingsdk-spec#queue).

tsuf239: cool, I added your explanation to the template :)

tsuf239:
@eladb Do you think we can add anything else to the RFC? or we can merge it?

eladb (approved):

PR #2106, opened 2023-04-16
feat: concat (`++`) operator

Chriscbr:
Introduces a concatenation operator to the language. Using string interpolation or calling `x.concat(y)` can be cumbersome for simple use cases, and `++` offers a simple way to join strings together. The syntax is inspired by languages like Haskell and Grain, and makes it unambiguous that the result will be a string and not a number.

The concat operator does not implicitly convert types into strings. If you want to concatenate a string with a number, you need to first convert the number to a string by calling `x.to_str()` for example.

### DIFF
--- a/docs/04-reference/winglang-spec.md
+++ b/docs/04-reference/winglang-spec.md
@@ -2540,15 +2540,30 @@ Ternary or conditional operators are not supported.
 
 ---
 
-#### 6.3.4 Operator Precedence
+#### 6.3.4 Other Operators
+
+| Operator | Description       | Example  |
+| -------- | ----------------- | -------- |
+| `?`      | Optional test     | `a?`     |
+| `??`     | Unwrap-or-default | `a ?? b` |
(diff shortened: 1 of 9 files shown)


Chriscbr:
Right now I'm a bit indifferent to the choice between string concatenation with `++` or `+`, here are the main pros/cons I thought of so far:  

- If we use "++", then because we know the output type is a string, then we can safely support implicit casting / implicit stringifying, just like string interpolation. For example, `"colors: " ++ colors_array`. 
- If we use "+", then we need to require both operands to be numbers or both operands to be strings to avoid type checking ambiguities.
- If we use "+" then we can also support `x += y` syntax for concatenating to (reassignable) strings if we wanted to add that to the language at some point

Thoughts?

eladb:
> Thoughts?

I actually leaning towards `+`. Since we already have interpolated strings (which is where we can offer implicit casting), I think `+` more intuitively covers the simple cases where you'd just want to concat two strings. Not sure the additional cognitive load is justified.

PR #2108, opened 2023-04-16
chore(sdk): add retention to queue

### LINKED ISSUE #1213: add queue max retries / retention period

fynnfluegge:
Closes #1213 

Visibilty timeout is still an approximation in the queue simulator. E.g. if timeout is 30s and the function throws an error after 1s, the message will be pushed back to the queue after 31s overall. I think this is fine at the moment, but could be improved in a follow up issue at some time in the future.


libs/wingsdk/src/cloud/queue.ts
```diff
@@ -19,6 +19,12 @@ export interface QueueProps {
    */
   readonly timeout?: Duration;
 
+  /**
+   * How long a queue retains a message.
+   * @default Duration.fromHours(1)
```
revitalbarletz, line 24: @fynnfluegge why did you choose 1 hour and not 30 seconds?

fynnfluegge: AWS SQS for example has minimum 1 minute retention period and maximum of 14 days. To me 1 hour is always a good default for retention. We have timeout 30s default, if we choose for example 1 min retention then the message is lost if the function throws an error once. Maybe at least in the simulator we can choose several minutes for default? Running the simulator for hours is not a realistic scenario 😁

fynnfluegge: Default value for AWS SQS is 4 days retention 😁

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -88,47 +95,55 @@ export class Queue implements IQueueClient, ISimulatorResourceInstance {
...
     } while (processedMessages);
   }
 
-  public async pushMessagesBackToQueue(messages: Array<string>): Promise<void> {
+  public async pushMessagesBackToQueue(
+    messages: Array<QueueMessage>
+  ): Promise<void> {
     setTimeout(() => {
-      this.messages.push(...messages);
+      // Don't push back messages with retention timeout
```
hasanaburayyan, line 139:
```suggestion
      // Don't push back messages that retention timeout has expired
```

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -139,6 +154,18 @@ export class Queue implements IQueueClient, ISimulatorResourceInstance {
   }
 }
 
+class QueueMessage {
+  retentionTimeout: Date;
+  payload: string;
+
+  constructor(retentionPeriod: number, message: string) {
+    const currentTime = new Date();
+    currentTime.setSeconds(retentionPeriod);
```
hasanaburayyan, line 163: I believe if we dont capture the current seconds from time when you call `.setSeconds()` there is a bug here. IE if the current time was 
`13:29:56.245Z` and the retention period came in as two minutes (120) then the time will be set to ``13:31:00.245Z` instead of `13:31:56.245Z` so its more like the retention period is 64 seconds.
```suggestion
    currentTime.setSeconds(retentionPeriod + currentTime.getSeconds());
```

libs/wingsdk/src/cloud/queue.ts
hasanaburayyan: This default is sounds pretty sensible to me but its only used for the sim so I dont think we want to annotate this on the `cloud.queue`

libs/wingsdk/src/target-sim/queue.ts
```diff
@@ -18,12 +18,14 @@ import { convertBetweenHandlers } from "../utils/convert";
...
 export class Queue extends cloud.Queue implements ISimulatorResource {
   private readonly timeout: std.Duration;
+  private readonly retentionPeriod: std.Duration;
   private readonly subscribers: QueueSubscriber[];
   private readonly initialMessages: string[] = [];
   constructor(scope: Construct, id: string, props: cloud.QueueProps = {}) {
     super(scope, id, props);
 
     this.timeout = props.timeout ?? std.Duration.fromSeconds(10);
+    this.retentionPeriod = props.retentionPeriod ?? std.Duration.fromHours(1);
```
hasanaburayyan, line 28: Should we throw an exception here if the retain period is less than timeout?

libs/wingsdk/test/target-sim/queue.test.ts
```diff
@@ -179,6 +180,43 @@ test("messages are not requeued if the function fails before timeout", async ()
   ]);
 });
 
+test("messages are not requeued if the function fails after retention timeout", async () => {
```
hasanaburayyan, line 183: Cant comment on the other file 😁 but can we also add a test for the `tf-aws` and `aws-cdk` queue resources.

hasanaburayyan (changes requested):
🔥 🔥  Sorry for the delay @fynnfluegge but this is some awesome stuff! I Have a few minor remarks just smack the re-request review when you had a chance to review them.

libs/wingsdk/src/cloud/queue.ts
fynnfluegge: I agree, but we should also remove the annotation from timeout, right? 😁

hasanaburayyan (approved):
🚀

PR #2113, opened 2023-04-17
fix: updates are always checked on cli startup

MarkMcCulloh:
update-notifier is configured to check for updates once a day. We've been checking it on every startup, which means it makes a network call out to npm. On my machine, this reduced the time for arbitrary `wing` commands by ~300ms.

Simplified the update system. There is an argument to skip it and it no longer has any code to do the installation for you.


apps/wing/src/commands/upgrade.ts
```diff
@@ -24,8 +24,10 @@ export async function upgrade(options: IUpdateOptions) {
     pkg: PJSON,
   });
 
-  log("forcing update info...");
-  await notifier.fetchInfo();
+  if (force) {
```
Chriscbr, line 27: How is this configured?

MarkMcCulloh: Doesn't look like there is a way. Personally, I think an upgrade subcommand makes the most sense to force it, so I can add that

apps/wing/src/commands/upgrade.ts
```diff
@@ -24,8 +24,10 @@ export async function upgrade(options: IUpdateOptions) {
...
+    log("forcing update info...");
+    await notifier.fetchInfo();
```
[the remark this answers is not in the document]
MarkMcCulloh: If you're forcing it to check for updates, I think it makes sense that you want to await here

apps/wing/src/commands/upgrade.ts
Chriscbr: I don't think an upgrade command necessarily makes sense since people might have the Wing toolchain installed in different ways (it might not be via npm -g ...)

MarkMcCulloh: > Can we also add a way to disable the update check?

Done

> Can we disable update checks when running without a connected TTY?

Done

> Can we disable update checks when running with CI environment variable set?

The library we use already does this

> I don't think an upgrade command necessarily makes sense since people might have the Wing toolchain installed in different ways (it might not be via npm -g ...)

So the written code already kinda handles this and can actually run an install command for you. However, I'm not really a fan of this anyways (I install things with volta, for instance).

I went ahead and deleted most of the updating code and kept it simplified to just to updating and notifying.

eladb (approved):
Looks good

@ainvoner can you please make sure to pass this flag when calling the cli from the console?

PR #2139, opened 2023-04-18
chore: add benchmarking suite and reporting

MarkMcCulloh:
- Added some benchmark wing files to test against using hangar. These are separate from the normal e2e examples and are only run in benchmarks.
- Added a meta-comment system for test wing files, used to allow tests to be skipped in CI as well as a way to specify how long a benchmark should take
- Benchmarks publish a table of results into the github action, as well as a json file with raw data. This raw data will be available on releases as well

There is a functionality change in here as well:
I noticed it was not possible to `bring` a node module if two things were true:
1. `wing` was being run from a different directory than the wing file
2. The node_modules folder with the expected package is not in the same directory

This was due to missing WASI preopens for node_modules. We were only adding node_modules preopens relative to the CWD, not the wind file. So I added automatic preopens for any resolvable node_modules with the parent tree.

---

- [x] Verify this all works in CI
- [x] Add more benchmarks if we can avoid increasing CI time
- [x] Benchmarking should not add more time to ci (should only take ~3 minutes)
- [x] Switch to using CLI arg instead of ENV variable to skip wing update checking
- [x] Attach benchmarks to releases in some way (just the json, or md too?)
- [x] Fail tests if they break a certain threshold
- [x] Add test metadata thing and contributor docs for it
- [x] Cleanup

Generated report example https://github.com/winglang/wing/actions/runs/4734205215/attempts/1#summary-12837953378


.github/workflows/build.yml
```diff
@@ -116,6 +116,47 @@ jobs:
...
+        run: npm run bench
+
+      - name: Upload Report JSON
+        uses: actions/upload-artifact@v3
+        with:
+          name: benchmarks
+          path: tools/hangar/results/report.json
+
+      - name: Create Markdown report
+        run: node scripts/benchmark_json_to_table.mjs
```
eladb, line 158: Can you share an example of how this looks?

tools/hangar/src/cli.bench.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const targets = ["sim", "tf-aws"];
+const files: string[] = [];
+for await (const wingFile of walkdir(benchmarksTestDir)) {
+  if (wingFile.endsWith(".w")) {
+    files.push(wingFile);
+  }
+}
+
+describe("compile", async () => {
```
eladb, line 14: Do we have a way to record thresholds per test and fail if the test exceeds the threashold?

.github/workflows/build.yml
MarkMcCulloh: https://github.com/winglang/wing/actions/runs/4734205215/attempts/1#summary-12837953378

.github/workflows/build.yml
```diff
@@ -116,6 +116,47 @@ jobs:
           github-comment: false
           vercel-args: "--prod "
 
+  benchmarks:
```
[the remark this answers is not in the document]
MarkMcCulloh: It is present in PR builds, although [not as a comment](https://github.com/winglang/wing/actions/runs/4734205215/attempts/1#summary-12837953378). I think comment should wait until we have comparative metrics (e.g. diff of perf between branches), WDYT?

tools/hangar/src/cli.bench.ts
MarkMcCulloh: Not yet, would have to be a post-process. Should that be a blocker for this PR?

eladb: Perhaps we can just configure the timeout for each test as the threshold and simply have a map in the test code that maps test files to their timeouts.

.github/workflows/build.yml
eladb: Very cool!

apps/wing/src/wingc.ts
```diff
@@ -93,6 +93,21 @@ export async function load(options: WingCompilerLoadOptions) {
...
+    const resolved = resolve(dots);
+    iDots += 1;
+    if (existsSync(resolved)) {
+      preopens[dots] = resolved;
+    }
+    
+    if (resolved === "/" || resolved.match(/^[A-Z]:\\$/i)) {
+      break;
+    }
+  }
```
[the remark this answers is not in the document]
eladb: Since we are now preopening all `node_modules` paths [here](https://github.com/winglang/wing/pull/2136/files#diff-e51e2fcb0abcddbff1739190a153f55fa3e991cae1a7e64020de3235c862f180), I don't think this is needed anymore.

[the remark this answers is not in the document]
MarkMcCulloh: Those are for the cwd and current node process, which isn't necessarily the same as node_modules that a wing file might try to resolve. Either way, this change didn't really work either so I had to do another approach.

I'm really thinking the final solution for this is going to be calling `require.resolve` from rust itself

tools/hangar/src/cli.bench.ts
MarkMcCulloh: Ok I added a system where you can add a comment to the wing file to designate this info and it'll fail the test based on that.

examples/tests/valid/benchmarks/empty.w
```diff
@@ -0,0 +1,7 @@
+/*\
+cases:
+  - target: sim
+    maxMeanTime: 900
+  - target: tf-aws
+    maxMeanTime: 1000
+\*/
```
MarkMcCulloh, line 7: This comment style is a callback to one of the original versions of the spec, so I thought it was a fun homage. I'm definitely open to feedback on both the comment style and the data structure what's inside it.

.github/workflows/build.yml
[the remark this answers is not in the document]
eladb: Let's create an issue and tag with "good first". Could be a nice community follow up

@staycoolcall911

eladb (approved):

docs/06-contributors/020-development.md
```diff
@@ -124,6 +124,49 @@ To run the tests (and update snapshots), run the following command from anywhere
...
+```
+
+Currently, the only supported meta-comment for regular tests is `skipPlatforms`.
+This will skip the test on the given platforms when when running on CI. The current supported platforms are `win32`, `darwin`, and `linux`.
+This is useful if, for example, the test requires docker. In our CI only linux supports docker.
+
+### Benchmarks
+
+Benchmark files are located in `examples/tests/valid/benchmarks`. To run the benchmarks, run the following command from anywhere in the monorepo:
+
```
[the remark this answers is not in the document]
MarkMcCulloh: So I was initially going to add this, but I realize that vitest is hiding the benchmark stdout when run through nx. This make it very confusing when you run "build" and it suddenly starts to hand in the end.

Will create a follow-up issue to figure out why the output is hidden and then add it to the "build" command.

apps/wing/src/wingc.ts
```diff
@@ -101,11 +101,20 @@ export async function load(options: WingCompilerLoadOptions) {
     }
   }
 
-  // for each provided preopen, add resolved paths in case any absolute paths are used
+  // for each provided preopens, add resolved paths in case any absolute paths are used
   for (const [key, value] of Object.entries(preopens)) {
     preopens[normalPath(resolve(key))] = value;
   }
 
+  // for each provided preopens, add relative paths in case any relative paths are used
```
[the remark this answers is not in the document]
MarkMcCulloh: When doing the node module resolution from within wasm, if it needs to search for the node_modules in a parent dir it ends up using `..`.

docs/06-contributors/020-development.md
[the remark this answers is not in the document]
MarkMcCulloh: https://github.com/winglang/wing/issues/2201

PR #2140, opened 2023-04-18
chore: remove file hashes from intermediate file paths

### LINKED ISSUE #2117: Remove file hashes from compiler generated intermediate inflight code files
Currently the compiler generates an intermediate file for each inflight with a file path containing a long hash -- see for example this snapshot:

https://github.com/winglang/wing/blob/main/tools/hangar/__snapshots__/test_corpus/std_string.w/compile/tf-aws/proc.9a4846d755bc1e93d1377b973079b3fb455a5196a4309ea68138302e1a6f3f65/index.js

This hashed file path results in more complicated PR diffs, and it's not strictly necessary since the files are only consumed by the SDK.

Chriscbr:
Simplifying the compiler output a bit... fixes #2117

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -15,8 +15,6 @@ use std::{
 	vec,
 };
 
-use sha2::{Digest, Sha256};
-
 use crate::{
 	ast::{
 		ArgList, BinaryOperator, Class as AstClass, ClassField, Expr, ExprKind, FunctionBody, FunctionBodyRef,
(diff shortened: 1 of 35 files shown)


MarkMcCulloh:
The dot needs to be removed from here for snapshots to see the new path:

https://github.com/winglang/wing/blob/main/tools/hangar/src/generated_test_targets.ts#L40

Chriscbr:
@MarkMcCulloh Fixed - I ran `git rm -r "**/proc.*/index.js"` to remove the old files from source control

MarkMcCulloh (approved):
Noice

libs/wingc/src/jsify.rs
```diff
@@ -15,8 +15,6 @@ use std::{
 	vec,
 };
 
-use sha2::{Digest, Sha256};
```
Chriscbr, line 18: Let me check if I can remove this dep

staycoolcall911:
Awesome fix, this will reduce the amount of files changed in many future PRs - thanks!

Issue #1669, opened 2023-02-26
Access `static` resource members from inflight


yoav-steinberg:
We need some mechanism to "capture" types so we can access their static members inflight. 
This might also be related to #390 when the static member is a static inflight method.

```wing
bring cloud;

resource Foo {
  init() {}

  static inflight static_inflight_method(): num {
    return "static inflight method";
  }

  static static_method(): num {
    return "static method";
  }

  static inflight static_inflight_field: str = "static inflight field";
  static static_field: str = "static field";
}

new cloud.Function(inflight (s:str): str => {
  // All these fail because we don't know what `Foo` is inflight
  assert(Foo.static_inflight_method() == "static inflight method");
  assert(Foo.static_method() == "static method");
  assert(Foo.static_inflight_field == "static inflight field");
  assert(Foo.static_field == "static field"); 
}) as "test";
```

hasanaburayyan:
@yoav-steinberg I think this is also related to https://github.com/winglang/wing/issues/1169 where we are unable to capture `SomeEnum.Whatever`?

yoav-steinberg:
> @yoav-steinberg I think this is also related to https://github.com/winglang/wing/issues/1169 where we are unable to capture `SomeEnum.Whatever`?

I think you're right

Chriscbr:
@yoav-steinberg What if the statics are accessed within a resource? Is the issue the same, or does that fix it?

```
bring cloud;

resource Foo {
  init() {}

  static inflight static_inflight_method(): num {
    return "static inflight method";
  }

  static static_method(): num {
    return "static method";
  }

  static inflight static_inflight_field: str = "static inflight field";
  static static_field: str = "static field";
}

resource A {
  init() {}
  inflight handle() {
    assert(Foo.static_inflight_method() == "static inflight method");
    assert(Foo.static_method() == "static method");
    assert(Foo.static_inflight_field == "static inflight field");
    assert(Foo.static_field == "static field"); 
  }
}

let a = new A();
new cloud.Function(inflight () => { a.handle() }) as "test";
```

yoav-steinberg:
> @yoav-steinberg What if the statics are accessed within a resource? Is the issue the same, or does that fix it?

It's the same issue. Remember there's no one currently bundling `Foo` into the inflight code so it's becomes an unknown symbol. Even if `handle` would access an instance of `Foo` it'll be accessing a `Foo_client` instance, so `Foo` itself is still inaccessible.

monadabot:
[...]

PR #2158, opened 2023-04-19
feat(compiler): reference other types from inflight code

yoav-steinberg:
See #1669, #1169

Todo:
- [x] tests
- [ ] specifically test something under a namespace (for now this is only relevant when testing an SDK resource)
- [ ] specifically test something form the sdk (any idea how to do this? I need and SDK resource with a static method)
- [x] enum support
- [ ] currently this doesn't support inflight closures, assuming they are going to be converted to resources we probably don't need to -> decide.

```wing
bring cloud;

resource Bar {
  init(){}
  static inflight do_static() {
    log("Bar's static!");
  }
}

enum MyEnum {
  VariantA,
  VariantB,
  VariantC,
}

resource Foo {
  init(){}
  inflight do1(e: MyEnum) {
    let s = Foo.do_static();
    if (e == MyEnum.VariantA) {
      log("VariantA");
    } elif (e == MyEnum.VariantB) {
      log("VariantB");
    } elif (e == MyEnum.VariantC) {
      log("VariantC");
    }
    log("instance inflight, got \"${s}\" from static inflight!");
    Bar.do_static();
  }

  inflight do () {
    this.do1(MyEnum.VariantA);
  }

  static inflight do_static(): str {
    return "static inflight!";
  }
}

let f = new Foo();

  //log(Foo.do_static());
  f.do();
}) as "test";

```

## Other stuff
- This PR does a major cleanup of symbol_env.rs:
  - Remove duplicate code for mut/non-mut lookups using the `duplicate` crate.
  - Naming cleanups.
  - `TypeErrors` aren't handled in symbol_env anymore, and there's a helper function in type_check to convert a failed `LookupResult` error to a `TypeError`.
  - Added a bunch of unit tests for symbol_env.
- Added a simple `Spanned` trait for getting a span of an AST node. Currently only implemented on `UserDefinedType` and `Symbol` but can used in other places too.
- Make `Symbol`'s `Display` only print the name and not the span info. This was starting to be more of an inconvenience than being useful. We can always explicitly also print the span.
- Removed `FreeVariableScanner` and its usage. We can simply get the free var info from the existing capture phase. Note that even if we do decide to scrap the capture compilation phase and implement such a scanner, I think it should use the existing logic from that phase (namely being environment aware when checking if a symbol is defined preflight).
- Split the generated `_toInflight` method into a `_toInflightType` and a `_toInflight` that uses it. This way we reuse the logic for generating the inflight type in the following two cases:
  - Inside `_toInflight` when we need to instantiate a client.
  - Inside `_toInflightType` itself when we need to add other types to our namespace because our type accessing their static inflight members (or enums).
 


libs/wingc/src/jsify.rs
```diff
@@ -1534,3 +1562,58 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+	pub fn find_preflight_type_refs(mut self) -> IndexMap<String, TypeRef> {
+		self.visit_scope(self.function_scope);
+		self.references
+	}
+}
+
+impl<'ast> Visit<'ast> for PreflightTypeRefVisitor<'ast> {
+	fn visit_scope(&mut self, node: &'ast Scope) {
+		self.env = node.env.borrow().as_ref().unwrap().get_ref();
+		visit::visit_scope(self, node);
```
Chriscbr, line 1597: Should we "restore" `self.env` to the previous env after we're done visiting a child scope? Or add a comment maybe...

libs/wingc/src/jsify.rs
```diff
@@ -1534,3 +1562,58 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+				if let SymbolKind::Type(t) = kind {
+					// If this user type was defined preflight then store it
+					if info.phase == Phase::Preflight {
+						self.references.insert(type_name, *t);
+					}
+				} else {
+					panic!("Expected {type_name} to be a type");
+				}
+			} else {
+				panic!("Unknown symbol: {type_name}");
```
Chriscbr, line 1615: Maybe add a comment that this should be handled by type checking

Chriscbr (commented):
Cool! I think the high level approach makes sense. Would like to take a second pass after the branch is updated with some of the snapshot changes

libs/wingc/src/ast.rs
```diff
@@ -132,6 +132,23 @@ pub struct FunctionTypeAnnotation {
...
+	fn span(&self) -> &WingSpan {
+		&self.span
+	}
+}
+
+impl Locatable for Symbol {
+	fn span(&self) -> &WingSpan {
+		&self.span
+	}
 }
```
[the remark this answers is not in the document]
yoav-steinberg: :+1:

libs/wingc/src/ast.rs
```diff
@@ -657,31 +664,31 @@ impl Display for Reference {
...
 
 /// Represents any type that has a span.
-pub trait ToSpan {
+pub trait Locatable {
```
MarkMcCulloh, line 667: A [very common](https://github.com/search?q=Spanned+lang%3ARust+&type=code) name for is trait is `Spanned`

libs/wingc/src/jsify.rs
yoav-steinberg: good catch, fixed.

libs/wingc/src/ast.rs
yoav-steinberg: done

examples/tests/valid/resource_captures_globals.w
```diff
@@ -67,6 +71,7 @@ class MyResource {
     assert(global_another.my_field == "hello!");
     global_another.first.my_resource.put("key", "value");
     assert(global_another.my_method() > 0);
+    //assert(this.local_counter.peek() > 0); // TODO: this fails, why?
```
eladb, line 74: @Chriscbr any idea?

Chriscbr (approved):

eladb (approved):

examples/tests/valid/resource_captures_globals.w
yoav-steinberg: :+1:  https://github.com/winglang/wing/issues/2472

Issue #1293, opened 2023-01-29
cloud.Website - Simulator implementation


staycoolcall911:
https://docs.winglang.io/reference/wingsdk-spec#website

monadabot:
Congrats! :rocket: This was released in [Wing 0.15.26](https://github.com/winglang/wing/releases/tag/v0.15.26).

PR #2164, opened 2023-04-20
feat(sdk): website

tsuf239:
creating a static website resource! 

fixes https://github.com/winglang/wing/issues/1294
fixes https://github.com/winglang/wing/issues/1293

The website resource is implemented by a bucket connected to CloudFront distribution for aws, and an express server for the sim. 
It serves a static folder, and supports adding dynamic content via the `add_json` method.

demo:
https://www.loom.com/share/30a8cf02694440549a0d1d1fd6a0c7d2


apps/wing/src/commands/compile.test.ts
```diff
@@ -1,14 +1,23 @@
-import { Target, compile } from "./compile";
-import { readdir, stat, writeFile } from "fs/promises";
-import { describe, test, expect } from "vitest";
-import { join, resolve } from "path";
+import {Target, compile} from "./compile";
```
staycoolcall911, line 1: Looks like you have a formatting issue in your local env. All these spaces before/after import names should remain

examples/tests/valid/website_with_api.w
```diff
@@ -0,0 +1,63 @@
...
+    if (body.get("name") == "" || body.get("age")  == "" || body.get("id")  == "") {
+        return cloud.ApiResponse {
+            body: {error: "incomplete details"},       
+            status: 500
+          };
+    }
+    users_table.insert(body);
+    return cloud.ApiResponse {
+      body: {user: body.get("id")},
+      status: 200
```
staycoolcall911, line 37: Another nit: in REST successful post returns 201, not 200
```suggestion
      status: 201
```

examples/tests/valid/website_with_api.w
```diff
@@ -0,0 +1,63 @@
...
+    };
+  };
+
+let post_handler = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
+    let body: Json = req.body ?? {name: "", age: "", id: ""};
+    if (body.get("name") == "" || body.get("age")  == "" || body.get("id")  == "") {
+        return cloud.ApiResponse {
+            body: {error: "incomplete details"},       
+            status: 500
```
staycoolcall911, line 31: nit: http code for bad user input is 400 (bad request). 
In general 400-s are for user error, 500-s are for server error
```suggestion
            status: 400
```

examples/tests/valid/website_with_api.w
```diff
@@ -0,0 +1,63 @@
...
+
+  // responsible for the CORS - https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html
+let options_handler = inflight(req: cloud.ApiRequest): cloud.ApiResponse => {
+    return cloud.ApiResponse {
+        headers: {
+                "Access-Control-Allow-Headers" : "Content-Type",
+                "Access-Control-Allow-Origin": "*",
+                "Access-Control-Allow-Methods": "OPTIONS,POST,GET"
+                },
+        status: 200
```
staycoolcall911, line 49: nit: I believe it should be a 204
```suggestion
        status: 204
```

staycoolcall911 (commented):
Very cool resource @tsuf239 - I love it that it already supports CORS and has a CDN. I added a few comments, but this needs more review from others on our team

libs/wingsdk/test/target-tf-aws/__snapshots__/website.test.ts.snap
```diff
@@ -0,0 +1,540 @@
...
+                  "wing:resource:stateful": false,
+                },
+                "children": {
+                  "Default": {
+                    "constructInfo": {
+                      "fqn": "@cdktf/provider-aws.s3Bucket.S3Bucket",
+                      "version": "12.0.2",
+                    },
+                    "id": "Default",
+                    "path": "root/Default/Website/Default",
```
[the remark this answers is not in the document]
tsuf239: it's the same with the bucket: https://github.com/winglang/wing/blob/b8168d8486ce0948cd9c3c04240702414bb2f207/libs/wingsdk/test/target-tf-aws/__snapshots__/bucket.test.ts.snap#L94

tools/hangar/__snapshots__/test_corpus/redis.w_test_sim.md
```diff
@@ -1,9 +0,0 @@
-# [redis.w](../../../../examples/tests/valid/redis.w) | test | sim
```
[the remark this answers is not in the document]
MarkMcCulloh: Redis tests are actually skipped in ci, but I am surprised it skipped and deleted it locally. I would say undo this change for now

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(scope: Construct, id: string, props: WebsiteProps) {
+    super(scope, id);
+
+    this.display.title = "Website";
+    this.display.description = "A static website";
+
+    this._path = isAbsolute(props.path)
+      ? props.path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", props.path);
```
MarkMcCulloh, line 45: This should probably throw instead of "", if this is meant to be the correct path

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,93 @@
...
+    this._domain = props.domain;
+  }
+
+  /**
+   * Local path to the website's static files, relative to the Wing source file.
```
MarkMcCulloh, line 50: Based on the constructor, this is absolute

```suggestion
   * Absolute local path to the website's static files.
```

libs/wingsdk/test/target-tf-aws/__snapshots__/website.test.ts.snap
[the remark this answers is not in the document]
tsuf239: I change the names anyway now, to be other than their namespaces, in my last commit

tools/hangar/__snapshots__/test_corpus/website.w_compile_tf-aws.md
```diff
@@ -0,0 +1,209 @@
...
+    "aws": [
+      {}
+    ]
+  },
+  "resource": {
+    "aws_cloudfront_distribution": {
+      "root_cloudWebsite_distribution_CC27FD7B": {
+        "//": {
+          "metadata": {
+            "path": "root/Default/Default/cloud.Website/distribution",
```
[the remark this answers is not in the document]
Chriscbr: If you're referring to the two "Default" names, this is a side effect of how the SDK layers resources together internally right now to make `cloud.App` behave the same for all targets. For the `tf-aws` target, "root" corresponds to `cdktf.TerraformStack`, one of the "Default" components is corresponds to `CdktfApp` (which we cheated to make it a child of TerraformStack), and the other "Default" component corresponds to the `$Root` construct generated in preflight.js. There might be some ways to remove either/both "Default" layers, but I don't think it's a critical issue

apps/wing/src/commands/compile.test.ts
```diff
@@ -1,14 +1,21 @@
...
 
 describe(
   "compile command tests",
   () => {
+    beforeAll(() => {
+      // deleting the artifactDir
+      const artifactDir = resolve("../../examples/tests/valid/target/captures.tfaws");
+      rmdirSync(artifactDir, { recursive: true });
+    });
+
```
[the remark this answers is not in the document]
Chriscbr: Makes sense to me 👍  I know when folks are using the Wing CLI, we have to be careful about removing files as to not delete people's terraform state files, but I don't think we need to provide the same protection for unit tests so this should be fine

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+ */
+export class Website extends cloud.Website {
+  private readonly bucket: S3Bucket;
+
+  constructor(scope: Construct, id: string, props: cloud.WebsiteProps) {
+    super(scope, id, props);
+
+    this.bucket = createEncryptedBucket(this, true, "website-bucket");
+
+    new S3BucketWebsiteConfiguration(this, "bucket-website-configuration", {
```
Chriscbr, line 30:
```suggestion
    new S3BucketWebsiteConfiguration(this, "BucketWebsiteConfiguration", {
```

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+      bucket: this.bucket.bucket,
+      indexDocument: { suffix: INDEX_FILE },
+    });
+
+    this.uploadFiles(this.path);
+
+    // create a cloudFront distribution
+    const distribution = new CloudfrontDistribution(this, "distribution", {
```
Chriscbr, line 38:
```suggestion
    const distribution = new CloudfrontDistribution(this, "Distribution", {
```

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public addJson(filePath: string, obj: Json): string {
+    if (!filePath.endsWith(".json")) {
+      throw new Error(
+        `key must have a .json suffix: ${filePath.split(".").pop()}`
+      );
+    }
+
+    new S3Object(this, `aws_s3_bucket_object_${filePath}`, {
```
Chriscbr, line 83: I don't think including the name of the terraform resource is needed / usually that info is redundant
```suggestion
    new S3Object(this, `File-${filePath}`, {
```

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+      bucket: this.bucket.bucket,
+      contentType: "application/json",
+      key: filePath,
+    });
+
+    return `${this.url}/${filePath}`;
+  }
+
+  private uploadFile(filePath: string) {
+    new S3Object(this, `aws_s3_bucket_object_${path.basename(filePath)}`, {
```
Chriscbr, line 95: I'm not sure if we can only use the basename, because then:
* /foo/hello.txt
* /bar/hello.txt

would end up with the same construct ID. Maybe we can use the full file path, or add a hash of the full path, e.g.
```suggestion
    new S3Object(this, `File-${basename}-${hash(filePath).slice(-8)}`, {
```

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,93 @@
...
+  get url(): string {
+    return this._url;
+  }
+
+  /**
+   * used for adding dynamic content to the website after deployment
+   * @param filePath the bucket key to add
+   * @param obj the object to write to the key
+   */
+  abstract addJson(filePath: string, obj: Json): string;
```
Chriscbr, line 69: Let's avoid the hairiness of case conversion by naming the parameter `path` for now
```suggestion
  /**
   * Add a JSON file with custom values during the website's deployment.
   * If the path conflicts with file path from the website's static assets, an error will be thrown.
   * @param path the file path it will be uploaded as
   * @param data the data to write to the file
   */
  abstract addJson(path: string, data: Json): string;
```

Chriscbr (commented):
Looks great so far! 😄

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    this._path = isAbsolute(props.path)
+      ? props.path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", props.path);
+    this._domain = props.domain;
+  }
+
+  /**
+   * Absolute local path to the website's static files.
+   * @example "./dist"
```
hasanaburayyan, line 51: The example is a relative path, should that be updated or is relative acceptable and the comment be updated.

libs/wingsdk/src/target-sim/website.inflight.ts
```diff
@@ -0,0 +1,78 @@
...
+      });
+    }
+  }
+
+  public async init(): Promise<ApiAttributes> {
+    // `server.address()` returns `null` until the server is listening
+    // on a port. We use a promise to wait for the server to start
+    // listening before returning the URL.
+    const addrInfo: AddressInfo = await new Promise((resolve, reject) => {
+      this.server = this.app.listen(0, LOCALHOST_ADDRESS, () => {
```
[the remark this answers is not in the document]
hasanaburayyan: (thinking outloud) I think staying away from a fixed port is probably better for now, otherwise we might need to consider how to increment ports based on number of sites, and if port is in use do we fail or keep incrementing ports to find a valid `next` port. Also we may end up in some race conditions if we start to get opinionated on ports. 

Thought I do agree the port number changing on sim restarts could be frustrating.

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @inflight `@winglang/sdk.cloud.IWebsiteClient`
+ */
+export class Website extends cloud.Website {
+  private readonly bucket: S3Bucket;
+
+  constructor(scope: Construct, id: string, props: cloud.WebsiteProps) {
+    super(scope, id, props);
+
+    this.bucket = createEncryptedBucket(this, true, "website-bucket");
```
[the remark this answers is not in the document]
hasanaburayyan: nit: I think Chris meant PascalCase rather than cameCase :smile: and I agree thats probably a good style choice to enforce

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+      bucket: this.bucket.bucket,
+      indexDocument: { suffix: INDEX_FILE },
+    });
+
+    this.uploadFiles(this.path);
+
+    // create a cloudFront distribution
+    const distribution = new CloudfrontDistribution(this, "distribution", {
+      enabled: true,
+      ...(this._domain && { aliases: [this._domain] }),
```
hasanaburayyan, line 40: hmm excuse my typescript confusion its pretty late here :smile: but Im not sure I understand this line. Is it the same as saying 

```js
const additionalCloudFrontProps = this._domain? { aliases: [this._domain] } : {};
const distribution = new CloudfrontDistribution(this, "distribution", {
  enabled: true,
  ...additionalCloudFrontProps
```

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,125 @@
...
+        geoRestriction: {
+          locations: [],
+          restrictionType: "none",
+        },
+      },
+      priceClass: "PriceClass_100",
+      viewerCertificate: { cloudfrontDefaultCertificate: true },
+    });
+
+    // //TODO: this is currently a cdk token. we need to see if we can resolve those tokens somehow
```
hasanaburayyan, line 72: Do we need a tracking issue?

hasanaburayyan (commented):
Super cool stuff 🔥 I added a few comments but did not get to finish, will pick back up in the morning.

tools/hangar/__snapshots__/test_corpus/website.w_compile_tf-aws.md
```diff
@@ -0,0 +1,209 @@
...
+            "uniqueId": "root_cloudWebsite_awss3bucketobjectindexhtml_3A649306"
+          }
+        },
+        "bucket": "${aws_s3_bucket.root_cloudWebsite_B2013695.bucket}",
+        "content_type": "text/html; charset=utf-8",
+        "depends_on": [
+          "aws_s3_bucket.root_cloudWebsite_B2013695"
+        ],
+        "key": "/index.html",
+        "source": "/Users/tsuf/Documents/wing/examples/tests/valid/website/index.html"
```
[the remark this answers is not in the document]
tsuf239: fixed :)

libs/wingsdk/src/target-sim/website.inflight.ts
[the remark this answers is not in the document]
tsuf239: I think the user should choose the port (similar to when they make one from scratch) - do we know what target we are running? 
(so we can do something like `let w = new website({path: "...", domain: target == "sim" ? "localhost:8080" :  "real-domain.com" })`)

tools/hangar/__snapshots__/test_corpus/redis.w_test_sim.md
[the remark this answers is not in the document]
tsuf239: image
seems to be skipped on local too, I'll take a look :)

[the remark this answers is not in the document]
tsuf239: This code is responsible for that:
https://github.com/winglang/wing/blob/746c0f17a4362b77b9c1095f7f28bc3a8ee238b7/tools/hangar/src/generate_tests.ts#L16-L31
here is the meta-comment:
image
and my machine is the one to skip:
image

+ if a contributor will run it they might have the other skipped platform, and it'll happen again...
What do you think will be the best way to solve it @MarkMcCulloh?

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,93 @@
...
+    id: string,
+    props: WebsiteProps
+  ): Website {
+    return App.of(scope).newAbstract(WEBSITE_FQN, scope, id, props);
+  }
+
+  public readonly stateful = false;
+  private readonly _path: string;
+  /**@internal */
+  protected _url!: string;
```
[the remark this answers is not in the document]
tsuf239: This property exists only on the tf-aws implementation now, for the sim we just override the URL getter

libs/wingsdk/src/cloud/website.ts
tsuf239: good catch!

libs/wingsdk/src/target-tf-aws/website.ts
tsuf239: it's actually the default behaviour, I wrote it to myself while coding, removing it then, thanks!

libs/wingsdk/src/target-tf-aws/website.ts
tsuf239: exactly!

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,98 @@
...
+  constructor(scope: Construct, id: string, props: WebsiteProps) {
+    super(scope, id);
+
+    this.display.title = "Website";
+    this.display.description = "A static website";
+
+    if (isAbsolute(props.path)) {
+      this._path = props.path;
+    } else {
+      if (!process.env.WING_SOURCE_DIR) {
```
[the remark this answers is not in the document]
tsuf239: Sure!! thanks!

examples/tests/valid/website_with_api.w
```diff
@@ -0,0 +1,63 @@
...
+    }
+    users_table.insert(body);
+    return cloud.ApiResponse {
+      body: {user: body.get("id")},
+      status: 201
+    };
+  };
+
+  // responsible for the CORS - https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html
+let options_handler = inflight(req: cloud.ApiRequest): cloud.ApiResponse => {
```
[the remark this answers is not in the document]
tsuf239: opened: https://github.com/winglang/wing/issues/2289

libs/wingsdk/src/cloud/website.ts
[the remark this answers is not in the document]
tsuf239: https://github.com/winglang/wing/issues/2290 added

apps/wing/src/commands/compile.ts
```diff
@@ -288,7 +290,7 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
     rmSync(tmpSynthDir, { recursive: true, force: true });
   } else {
     // Move the temporary directory to the final target location in an atomic operation
-    copySync(tmpSynthDir, synthDir, { overwrite: true } );
```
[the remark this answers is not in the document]
tsuf239: I agree :) 
should I run the formatted on this (never formatted apparently) folder and create another PR after this one will be merged?

libs/wingsdk/src/cloud/website.ts
```diff
@@ -0,0 +1,95 @@
+import { isAbsolute, resolve } from "path";
```
staycoolcall911, line 1: Noticed another issue when working with the website resource on aws:
Need to invalidate cache on CloudFront after every change to the static website's code.
I think as part of `terraform apply` we should have some logic to invalidate some/all of the static files (need to leave an escape hatch for users to decide to remove specific files/patterns only)

staycoolcall911, line 1: Noticed another issue when working with the website resource on aws:
Even after we manually invalidated the CloudFront cache - `terraform apply` didn't publish index.html correctly.

tools/hangar/__snapshots__/test_corpus/redis.w_test_sim.md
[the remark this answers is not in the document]
MarkMcCulloh: Ah, I think I intended to only skip on CI. I think it would be sufficient to add a check only to respect skipPlatforms when `process.env.CI` is `true`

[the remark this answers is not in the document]
tsuf239: perfect! thanks :) 
we already have `process.env.CI` right?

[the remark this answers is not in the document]
tsuf239: Done :)

libs/wingsdk/src/cloud/website.ts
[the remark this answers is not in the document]
tsuf239: fixed :)

libs/wingsdk/src/cloud/website.ts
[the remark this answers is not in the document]
tsuf239: I think I could solve it when making the policy depends on the bucket

tsuf239: I cannot find a cdktf module to perform cache invalidation :( 
I wonder if I'm missing something
I could find one for cdk:
https://docs.aws.amazon.com/cdk/api/v1/docs/aws-s3-deployment-readme.html#cloudfront-invalidation

tsuf239: I think it takes time and depends on the invalidation method, I'll try implementing somehow the invalidation and we'll see if it persists

tsuf239: and see that it's not trivial in terraform :( 
https://stackoverflow.com/questions/69794727/how-can-i-invalidate-aws-cloudfront-distribution-cache-using-terraform

tsuf239: I think we should open an issue and think what can be the best way to achieve it based on the available tools... @staycoolcall911  WDYT?

staycoolcall911: Yes, we can open a separate issue and leave this to a following PR, but then it must be documented, because every user of the Website resource will run into that sooner or later. Perhaps @Chriscbr, @hasanaburayyan or @eladb will have an idea.

staycoolcall911: Ok, thanks, but please note that we (ShaiA and myself) waited for the cache invalidation to complete, then viewed index.html and it was still outdated.

Chriscbr: Ah right, cache invalidation - the hardest problem in computer science 😅 yes I remember now CDK had some dedicated code for handling this for users... it looks like it was written in Python as part of the BucketDeployment resource [here](https://github.com/aws/aws-cdk/blob/c81d115955dbb27ce873ed7c9d71cc0dc8eacf99/packages/aws-cdk-lib/aws-s3-deployment/lib/lambda/index.py#L130-L131)

I feel like we eventually want to handle this using the `OnDeploy` resource (basically, a `cloud.Function` that runs during deployment to call the AWS SDK and perform a cache invalidation with minimal permissions). But the approach mentioned on StackOverflow (creating a local_exec resource) sounds okay as a workaround - documenting as @staycoolcall911 mentioned also works 👍

tsuf239: opened an issue: https://github.com/winglang/wing/issues/2404

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -0,0 +1,130 @@
...
+    this._url = distribution.domainName;
+  }
+
+  public get url(): string {
+    return this._url;
+  }
+
+  public addJson(path: string, data: Json): string {
+    if (!path.endsWith(".json")) {
+      throw new Error(`key must have a .json suffix: ${path.split(".").pop()}`);
```
Chriscbr, line 82: nit: what if the user inputs a file path with no extension?

libs/wingsdk/test/target-sim/website.test.ts
```diff
@@ -0,0 +1,79 @@
...
+test("addJson throws an error for no json path", async () => {
+  expect(() => {
+    const jsonConfig = { version: "3.3.5" };
+    const jsonPath = "not a json Path.csv";
+    const app = new SimApp();
+    const website = cloud.Website._newWebsite(app, "website", {
+      path: resolve(__dirname, "website"),
+    });
+    website.addJson(jsonPath, Object(jsonConfig));
+  }).toThrowError("key must have a .json suffix: csv");
```
Chriscbr, line 78: nit: Wrap the `expect` only around the addJson() line to make it clear that is the one throwing the error

libs/wingsdk/test/util.ts
```diff
@@ -18,6 +18,26 @@ export function tfResourcesOfCount(
   return Object.values(JSON.parse(templateStr).resource[resourceId]).length;
 }
 
+export function tfResourcesOfWithProperty(
```
Chriscbr, line 21:
```suggestion
export function tfResourcesWithProperty(
```

libs/wingsdk/test/util.ts
```diff
@@ -18,6 +18,26 @@ export function tfResourcesOfCount(
...
+  templateStr: string,
+  resourceId: string,
+  properties: Record<string, string>
+) {
```
Chriscbr, line 25: style: Add return type

tools/hangar/__snapshots__/test_corpus/website.w_test_sim.md
```diff
@@ -0,0 +1,10 @@
+# [website.w](../../../../examples/tests/valid/website.w) | test | sim
+
+## stdout.log
```
Chriscbr, line 3: Can we also start snapshotting the simulator.json files? I think tracking these would help us detect breaking changes to the console

(maybe this should be a separate issue...)

Chriscbr (approved):
🎉 🎉 🎉

staycoolcall911 (approved):
Should be good to go - nice!

tools/hangar/__snapshots__/test_corpus/website.w_test_sim.md
tsuf239: added: #2427

Issue #1447, opened 2023-02-06
Resource Declaration: generate implicit constructors based on captures from outer scope


staycoolcall911:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Consider this use case where `global_bucket` is defined outside of the scope of `MyResource`, and is captured by one of `MyResource`'s inflight methods. We would like our JSified code to initialize MyResource and MyResourceClient with this captured global_bucket. 
This is the desired wing code and the js it compiles to:
```ts
// Wing
bring cloud;
​
let global_bucket = new cloud.Bucket();
​
resource MyResource {
  b2: cloud.Bucket;
  init(b2) {
    this.b2 = b2;
  }
​
  inflight my_put() {
    global_bucket.put('key', 'value');
    this.b2.put('key', 'value');
  }
}
​
let someb2 = new cloud.Bucket();
new MyResource(someb2);
​
// JSified preflight
let cloud = require('wingsdk/cloud');
let global_bucket = new cloud.Bucket(this, 'cloud.Bucket');
​
class MyResource {
  constructor(scope, id, b2, captured_global_bucket) {
    super(scope, id);
    this.b2 = b2;
    this.captured_global_bucket = captured_global_bucket;
  }
}
​
new MyResource(this, 'MyResource', b2, global_bucket);
​
// JSified client
class MyResourceClient {
  constructor(b2, captured_global_bucket) {
    this.b2 = b2;
    this.captured_global_bucket = captured_global_bucket;
  }
​
  myPut() {
    this.captured_global_bucket.put('key', 'value');
    this.b2.put('key', 'value');
  }
}
``` 

### Use Cases

See above.

### Component

Compiler

staycoolcall911:
Completed in #2300

PR #2205, opened 2023-04-20
chore(compiler): inline inflight clients in preflight.js

Chriscbr:
Reduce the number of intermediate files generated by the compiler by putting inflights client code inside `preflight.js`, adjacent to the classes they're defined for.

The main motivation for this is to make it easier in a later PR to inline global variable values in the resource's context without having to inject them as class fields (see #1447). See the screenshot below of how we might jsify a `global_bucket` being used by an inflight method:

Screenshot 2023-04-20 at 7 23 15 PM

(To get this working with permissions requires more work so I didn't include it in this PR).

This change might also have performance improvements since it reduces the number of `require` calls made by the simulator and the number of files the compiler has to emit. (Maybe the benchmark will indicate one way or another?)

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -37,8 +37,6 @@ const STDLIB: &str = "$stdlib";
 const STDLIB_CORE_RESOURCE: &str = formatcp!("{}.{}", STDLIB, WINGSDK_RESOURCE);
 const STDLIB_MODULE: &str = WINGSDK_ASSEMBLY_NAME;
 
-const INFLIGHT_CLIENTS_DIR: &str = "clients";
-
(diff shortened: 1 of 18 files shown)


eladb (approved):
I am not 100% sure this is the right direction. Concern is that `preflight.js` will blow up at some point, as well as if it will make it easier/harder to create source maps.

I like the fact that now all of the emitted javascript code is self-contained. It's much easier to follow and debug. 

But... I think this is totally reversible, so let's give this a try!

eladb:
What happens if there's a cyclic dependency? Resource A references an object of resource B and vice versa?

eladb (changes requested):
See comment about cycles

PR #2206, opened 2023-04-21
chore(sdk): remove binding symbol magic

Chriscbr:
This PR removes some of the "magic" in the SDK that we had to support binding annotations.

The conceptual idea behind the binding system is simple: the compiler annotates each resource with a set of facts that each say "if operation X is used on this resource, then this resource transitively needs to use operations Y on Z". Then, the SDK uses the facts to `bind` each resource to the inflight host (typically a `cloud.Function`) that needs to use it. The `bind` method gives each resource an opportunity to grant IAM permissions, environment variables, or other capabilities to the inflight host based on what operations are being used on it.

For example, when an AWS `cloud.Bucket` is told that an inflight host wants to call `put` on it, it can grant the inflight host `s3:PutObject` permissions, and set an environment variable with the bucket's ARN.

We had a clever way of registering these facts into the system, where the static method `Resource._annotateInflight(<fact>)` would add a fact to a hidden property of the class that can be accessed by all class instances. Unfortunately it requires the facts to always describe **resources or data that belong to the class**, like `this.data` or `this.child_bucket`.

This approach has held up well so far, but it doesn't extend for cases where a resource needs to delegate the inflight host to bind to resources that are NOT field members -- such as for binding to a global bucket (see #1447). So instead, this PR rewinds our approach a bit so that resources override the `_registerBind` methods directly. This extends the amount of code in `preflight.js` files but I think it's an ok trade-off.

An `_addInflightOps` method has been added to maintain the existing behavior where if an inflight host requests to use operations from a resource that the resource does not support, it will still throw an error. (This is aimed to prevent hard-to-uncover compiler bugs in the future).

### DIFF
--- a/libs/wingsdk/src/cloud/bucket.ts
+++ b/libs/wingsdk/src/cloud/bucket.ts
@@ -49,7 +49,7 @@ export abstract class Bucket extends Resource {
     this.display.title = "Bucket";
     this.display.description = "A cloud object store";
 
-    this._inflightOps.push(
+    this._addInflightOps(
       BucketInflightMethods.DELETE,
       BucketInflightMethods.GET,
       BucketInflightMethods.GET_JSON,
--- a/libs/wingsdk/src/cloud/bucket.ts
+++ b/libs/wingsdk/src/cloud/bucket.ts
@@ -49,6 +49,16 @@ export abstract class Bucket extends Resource {
     this.display.title = "Bucket";
     this.display.description = "A cloud object store";
 
+    this._inflightOps.push(
+      BucketInflightMethods.DELETE,
+      BucketInflightMethods.GET,
+      BucketInflightMethods.GET_JSON,
+      BucketInflightMethods.LIST,
+      BucketInflightMethods.PUT,
+      BucketInflightMethods.PUT_JSON,
+      BucketInflightMethods.PUBLIC_URL
+    );
+
     props;
   }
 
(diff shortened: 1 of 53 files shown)


tools/hangar/__snapshots__/test_corpus/construct-base.w_compile_tf-aws.md
```diff
@@ -59,6 +59,7 @@ class $Root extends $stdlib.std.Resource {
     class WingResource extends $stdlib.std.Resource {
       constructor(scope, id, ) {
         super(scope, id);
+        this._inflightOps.push();
```
Chriscbr, line 62: TODO - if the resource has no inflight ops, we can skip generating this line

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -49,6 +49,16 @@ export abstract class Bucket extends Resource {
     this.display.title = "Bucket";
     this.display.description = "A cloud object store";
 
+    this._inflightOps.push(
```
eladb, line 52: I *think* (?) that we should wrap this with an API instead of directly have everyone access `_inflightOps`. Something like `this._addInflightMethod()`, but we can change that later...

eladb (approved):
Much much better!

tools/hangar/__snapshots__/test_corpus/api.w_compile_tf-aws.md
```diff
@@ -288,6 +288,7 @@ class $Root extends $stdlib.std.Resource {
     class Foo extends $stdlib.std.Resource {
       constructor(scope, id, api) {
         super(scope, id);
+        this._inflightOps.push("handle");
```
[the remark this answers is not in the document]
Chriscbr: $inflight_init is added as a supported op to all resources in `std.Resource`

libs/wingsdk/src/cloud/bucket.ts
Chriscbr: Done

PR #2211, opened 2023-04-21
fix(compiler): resources cannot reference free variables inflight

### LINKED ISSUE #1447: Resource Declaration: generate implicit constructors based on captures from outer scope

Chriscbr:
Closes #1447

This pull request adds support for capturing [free variables](https://en.wikipedia.org/wiki/Free_variables_and_bound_variables) in resource classes, a feature that allows Wing to generate more expressive code for inflight clients. It implements a new visitor, modifies the code generation, and adds a new test case and a snapshot to verify the functionality.

The approach I've taken differs from the solution outlined in #1447. The original issue suggests that we transform free variables into resource fields. I saw some risks with this approach since all usages of the global value inside inflight constructors and inflight methods would also need to be transformed (for example, changing `global_bucket` to `this.global_bucket`), and tracking the list of these mappings so that they only apply to inflight code, and preventing name conflicts, could be tricky.

Instead, this PR augments jsification so that free variables are effectively also emitted as free variables in the JavaScript client code. 

I'm not sure this approach is any better, but it seems to work just as well for a number of cases I've tried.

### DIFF
--- a/examples/tests/valid/resource_captures_globals.w
+++ b/examples/tests/valid/resource_captures_globals.w
@@ -1,0 +1,62 @@
+bring cloud;
+
+let global_bucket = new cloud.Bucket();
+let global_counter = new cloud.Counter();
+let global_str = "hello";
+let global_bool = true;
+let global_num = 42;
+let global_array_of_str = ["hello", "world"];
+let global_map_of_num = Map<num>{ "a": -5, "b": 2 };
+let global_set_of_str = Set<str>{ "a", "b" };
+
+resource First {
+  my_resource: cloud.Bucket;
+
+  init() {
+    this.my_resource = new cloud.Bucket();
+  }
+}
+
+resource Another {
+  my_field: str;
+  first: First;
+
+  init () {
+    this.my_field = "hello!";
+    this.first = new First();
+  }
+
+  inflight init() {
+    assert(global_counter.peek() == 0);
+  }
+
+  inflight my_method(): num {
+    global_counter.inc();
+    return global_counter.peek();
+  }
+}
+
+let global_another = new Another();
+
+resource MyResource {
+  init() {}
+
+  inflight my_put() {
+    global_bucket.put("key", "value");
+    assert(global_str == "hello");
+    assert(global_bool == true);
+    assert(global_num == 42);
+    assert(global_array_of_str.at(0) == "hello");
+    assert(global_map_of_num.get("a") == -5);
+    assert(global_set_of_str.has("a"));
+    assert(global_another.my_field == "hello!");
+    global_another.first.my_resource.put("key", "value");
+    assert(global_another.my_method() > 0);
+  }
+}
+
+let res = new MyResource();
+
+new cloud.Function(inflight () => {
+  res.my_put();
+}) as "test";
--- a/libs/wingc/src/ast.rs
+++ b/libs/wingc/src/ast.rs
@@ -23,6 +23,12 @@ impl Symbol {
 			span: Default::default(),
 		}
 	}
+
+	/// Returns true if the symbols refer to the same name AND location in the source code.
+	/// Use `eq` to compare symbols only by name.
+	pub fn exact_eq(&self, other: &Self) -> bool {
+		self.name == other.name && self.span == other.span
+	}
 }
 
 impl Ord for Symbol {
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -7,7 +7,7 @@ use itertools::Itertools;
 use std::{
 	cell::RefCell,
 	cmp::Ordering,
-	collections::{BTreeMap, BTreeSet},
+	collections::{BTreeMap, BTreeSet, HashSet},
 	fmt::Display,
 	fs,
 	path::Path,
@@ -19,7 +19,7 @@ use crate::{
 	ast::{
 		ArgList, BinaryOperator, Class as AstClass, ClassField, Expr, ExprKind, FunctionBody, FunctionBodyRef,
 		FunctionDefinition, Initializer, InterpolatedStringPart, Literal, MethodLike, Phase, Reference, Scope, Stmt,
-		StmtKind, Symbol, TypeAnnotation, UnaryOperator, UserDefinedType,
+		StmtKind, Symbol, TypeAnnotation, UnaryOperator, UserDefinedType, UtilityFunctions,
 	},
 	debug,
 	diagnostic::{Diagnostic, DiagnosticLevel, Diagnostics},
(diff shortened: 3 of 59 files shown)


libs/wingc/src/jsify.rs
```diff
@@ -7,7 +7,7 @@ use itertools::Itertools;
 use std::{
 	cell::RefCell,
 	cmp::Ordering,
-	collections::{BTreeMap, BTreeSet},
+	collections::{BTreeMap, BTreeSet, HashSet},
```
Chriscbr, line 10: Fix

examples/tests/valid/resource_captures_globals.w
```diff
@@ -0,0 +1,62 @@
+bring cloud;
```
eladb, line 1: Beautiful!

libs/wingc/src/ast.rs
```diff
@@ -23,6 +23,12 @@ impl Symbol {
 			span: Default::default(),
 		}
 	}
+
+	/// Returns true if the symbols refer to the same name AND location in the source code.
+	/// Use `eq` to compare symbols only by name.
+	pub fn exact_eq(&self, other: &Self) -> bool {
```
eladb, line 29: Perhaps call it `same`?

libs/wingc/src/jsify.rs
```diff
@@ -1476,13 +1524,25 @@ impl Display for Component<'_> {
...
-	fn analyze_expr(node: &'a Expr) -> Vec<Component> {
+	fn analyze_expr(&self, node: &'a Expr) -> Vec<Component> {
 		match &node.kind {
 			ExprKind::Reference(Reference::Identifier(x)) => {
+				let scope = match &self.function_def.body {
+					FunctionBody::Statements(scope) => scope,
+					FunctionBody::External(_) => panic!("unexpected expression inside body of extern functions"),
+				};
+				let env = scope.env.borrow();
+				let env = env.as_ref().expect("scope should have an env");
```
eladb, line 1535: Wow this is interesting... overriding `env`... I guess that's legitimate.

libs/wingc/src/jsify.rs
```diff
@@ -1520,15 +1580,175 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+	fn visit_stmt(&mut self, stmt: &Stmt) {
+		match &stmt.kind {
+			// this statement introduces bound variables!
+			StmtKind::Bring {
+				module_name,
+				identifier,
+			} => {
+				self.visit_symbol(&module_name);
+
+				// bring cloud;
```
eladb, line 1640: Nit: Add `"` to indicate this is what we are matching (or some other way)

```suggestion
				// "bring cloud;"
```

libs/wingc/src/jsify.rs
```diff
@@ -1520,15 +1580,175 @@ impl<'a> FieldReferenceVisitor<'a> {
...
 				return [obj, prop].concat();
 			}
 
 			_ => vec![],
 		}
 	}
 }
+
+struct FreeVariableScanner {
+	bound_vars: Vec<Symbol>,
```
eladb, line 1598: Add a wee comment explaining each field

libs/wingc/src/jsify.rs
```diff
@@ -1520,15 +1580,175 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+				// bring "foo" as bar;
+				if let Some(ref id) = identifier {
+					self.visit_symbol(&id);
+
+					// add `bar` to the list of bound variables
+					self.bound_vars.push(id.clone());
+				}
+			}
+			// this statement introduces bound variables!
+			StmtKind::VariableDef {
```
eladb, line 1655:
```suggestion
			// "let x = y"
			StmtKind::VariableDef {
```

libs/wingc/src/jsify.rs
```diff
@@ -1520,15 +1580,175 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+				type_,
+			} => {
+				self.visit_symbol(&var_name);
+				self.visit_expr(&initial_value);
+				if let Some(type_) = type_ {
+					self.visit_type_annotation(&type_);
+				}
+				self.bound_vars.push(var_name.clone());
+			}
+			// invariant: adds net zero bound variables
```
eladb, line 1668: Is this legal in Wing?

```js
for x in y {
  resource Foo {
    init() {
      // access "x"
    }
    
  }
}
```

libs/wingc/src/jsify.rs
```diff
@@ -1520,15 +1580,175 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+			// check if there is already a bound variable with the same name
+			if !self.bound_vars.contains(x) {
+				self.free_vars.push(x.clone());
+			}
+		};
+
+		return visit::visit_reference(self, node);
+	}
+
+	// invariant: adds net zero bound variables
```
eladb, line 1623: Not 100% sure I understand what this means

tools/hangar/__snapshots__/test_corpus/resource_captures_globals.w_compile_tf-aws.md
```diff
@@ -0,0 +1,449 @@
+# [resource_captures_globals.w](../../../../examples/tests/valid/resource_captures_globals.w) | compile | tf-aws
```
eladb, line 1: I just love those markdown snapshots! (@MarkMcCulloh 👍 👍)

eladb (approved):
Looks great! Some minor comments.

libs/wingc/src/jsify.rs
Chriscbr: It probably shouldn't

libs/wingc/src/jsify.rs
Chriscbr: Ah yeah, this is a Rust trick. You can also do stuff like this:
```rust
let mut xs = vec![1, 2, 3];
xs.push(4);
let xs = xs; // no longer can be changed!
```
It's pretty neat™️

eladb: Why not. It's safe

PR #2212, opened 2023-04-22
feat(sdk): try parse static method for json

jogold:
Add `try_parse()` static method on `Json`.

### DIFF
--- a/libs/wingsdk/API.md
+++ b/libs/wingsdk/API.md
@@ -3513,6 +3513,7 @@ The index of the element in the Json Array to return.
 | <code><a href="#@winglang/sdk.std.Json.keys">keys</a></code> | Returns the keys from the Json object. |
 | <code><a href="#@winglang/sdk.std.Json.parse">parse</a></code> | Parse a string into a Json. |
 | <code><a href="#@winglang/sdk.std.Json.stringify">stringify</a></code> | Formats Json as string. |
+| <code><a href="#@winglang/sdk.std.Json.tryParse">try_parse</a></code> | Try to parse a string into a Json. |
 | <code><a href="#@winglang/sdk.std.Json.values">values</a></code> | Returns the values from the Json. |
 
 ---
@@ -3638,6 +3639,24 @@ to format as string.
 ###### `indent`<sup>Optional</sup> <a name="indent" id="@winglang/sdk.std.Json.stringify.parameter.indent"></a>
 
 - *Type:* num
+
+---
+
+##### `try_parse` <a name="try_parse" id="@winglang/sdk.std.Json.tryParse"></a>
+
+```wing
+bring std;
+
+std.Json.try_parse(str: str)
+```
+
+Try to parse a string into a Json.
+
+###### `str`<sup>Required</sup> <a name="str" id="@winglang/sdk.std.Json.tryParse.parameter.str"></a>
+
+- *Type:* str
+
+to parse as Json.
 
 ---
 
--- a/libs/wingsdk/src/std/json.ts
+++ b/libs/wingsdk/src/std/json.ts
@@ -98,6 +98,19 @@ export class Json {
   }
 
   /**
+   * Try to parse a string into a Json
+   *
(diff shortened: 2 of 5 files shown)


libs/wingsdk/API.md
```diff
@@ -3513,6 +3513,7 @@ The index of the element in the Json Array to return.
 | <code><a href="#@winglang/sdk.std.Json.keys">keys</a></code> | Returns the keys from the Json object. |
 | <code><a href="#@winglang/sdk.std.Json.parse">parse</a></code> | Parse a string into a Json. |
 | <code><a href="#@winglang/sdk.std.Json.stringify">stringify</a></code> | Formats Json as string. |
+| <code><a href="#@winglang/sdk.std.Json.tryParse">try_parse</a></code> | Try to parse a string into a Json. |
```
staycoolcall911, line 3516: Hey - just a small comment regarding auto generating the ref docs:
Please also update https://github.com/winglang/wing/blob/main/docs/04-reference/wingsdk-api.md
You don't need to make any of these doc updates manually. 
To generate these docs automatically after you made changes to the SDK code, please just run `npm run build` from the [wingsdk root folder](https://github.com/winglang/wing/tree/main/libs/wingsdk).

libs/wingsdk/src/std/json.ts
```diff
@@ -97,6 +97,19 @@ export class Json {
...
 
+  /**
+   * Try to parse a string into a Json
+   *
+   * @macro ((args) => { try { return JSON.parse(args); } catch (err) { return undefined; } })($args$)
+   *
+   * @param str to parse as Json
+   * @returns Json representation of the string or undefined if string is not parsable
+   */
+  public static tryParse(str: string): Json | undefined {
```
staycoolcall911, line 108: @Chriscbr - do you think we should return an optional json (`Json?`) here instead of this union type?
Similar to `Bucket.try_get()` (which appears in the [SDK spec](https://docs.winglang.io/reference/wingsdk-spec#bucket), but is not implemented yet).

jogold: `Json?` as return type is not valid typescript afaik.

libs/wingsdk/API.md
jogold: OK.

But this is generated doc, committed by the mutate job, why is the other doc not covered by the mutate job?

libs/wingsdk/src/std/json.ts
staycoolcall911: You are correct, of course. `Json?` Is the desired Wing return type for `try_get`. I'm just wondering what should be the typescript return type that corresponds to that (and it might be that what you've put there is entirely correct - I just wanna see if Chris has an other idea)

libs/wingsdk/API.md
staycoolcall911: Good question, I'll check

libs/wingsdk/src/std/json.ts
Chriscbr: Thanks for double checking! The way jogold has written this is all good, union with undefined is the best way to represent this in TypeScript I know of 👍

We might want to add some of these type correspondences to our contributing guide, since there are a couple of quirks like this (another example is inflight APIs should always return Promises)

staycoolcall911: Great, thank you both. Resolving my comment

eladb (approved):
Awesome

libs/wingsdk/API.md
jogold: Here we have `add .` and the working directory is `libs/wingsdk` but the doc is in `../../docs`.

https://github.com/winglang/wing/blob/7e8d309d1be79c35c35bde3fb33f468eb18aa31a/.github/workflows/wingsdk-mutation.yml#L64

I think it should be `git add -u`

jogold: #2219

PR #2219, opened 2023-04-23
fix(repo): not all changes picked up by mutation job

jogold:
Use `git add --all` otherwise changes in folders outside the working directory (like `docs`) are not picked up.

### DIFF
--- a/.github/workflows/wingsdk-mutation.yml
+++ b/.github/workflows/wingsdk-mutation.yml
@@ -61,7 +61,7 @@ jobs:
         if: steps.token-check.outputs.HAS_TOKEN == 'true'
         name: Find mutations
         run: |-
-          git add .
+          git add -u
           git diff --staged --patch --exit-code > /dev/null || echo "::set-output name=self_mutation_happened::true"
       - name: Push changes
         if: steps.token-check.outputs.HAS_TOKEN == 'true' && steps.self_mutation.outputs.self_mutation_happened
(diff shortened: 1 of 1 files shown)


.github/workflows/wingsdk-mutation.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         if: steps.token-check.outputs.HAS_TOKEN == 'true'
         name: Find mutations
         run: |-
-          git add .
+          git add -u
```
jogold, line 64: This doesn't cover new files, not sure it should.

jogold: Maybe `-all` is better

Chriscbr (approved):

.github/workflows/wingsdk-mutation.yml
Chriscbr: Introducing new snapshot files would be a valid use case for `--all`, let's try it

.github/workflows/wingsdk-mutation.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         if: steps.token-check.outputs.HAS_TOKEN == 'true'
         name: Find mutations
         run: |-
-          git add .
+          git add -all
```
jogold, line 64:
```suggestion
          git add --all
```

Issue #1438, opened 2023-02-05
[Tracking Issue] "readme" documentation for each SDK resource


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

The Wing docsite now includes a documentation topic for each SDK resource (we call it the "readme" of the resource). It is a human-written document that describes the resource and how it behaves, as well as target-specific details such as #1405.

### Use Cases

Give user detailed information about how to use resources and how they behave.

### Implementation Notes

The proposed implementation is to create an `.md` file next to each file [here](https://github.com/winglang/wing/tree/main/libs/wingsdk/src/cloud) which will include the readme for this resource. For example, `bucket.md`.

Each readme will contain basic information about how to use the resource, and any relevant details about how the resource is implemented on different cloud targets (like AWS, Azure, or the Wing simulator).

### Component

SDK

-----

Issues:

- [x] #2652
- [x] #2653
- [x] #2654
- [x] #2655
- [x] #2656
- [ ] #2657
- [x] #2658
- [x] #2659
- [x] #2660
- [x] #2661
- [x] #2662

eladb:
We need this!

ekeren:
done

PR #2255, opened 2023-04-24
feat(sdk): reference existing secrets

### LINKED ISSUE #2252: `name` of `cloud.Secret` should refer to an existing name
### I tried this

```js
bring cloud;

new cloud.Secret(name: "my-secret-name");
```

### I expected this:

Since `name` refers to a physical name, I think it should mean that it references an *existing* secret instead of trying to create a new one.

If I want a new secret for every app, I expect this behavior:

```js
new cloud.Secret() as "my-secret-id";
```

And then `terraform apply` should output the `aws secretsmanager` command for putting the secret value:

```sh
aws secretsmanager put-secret-value --secret-id "my-secret-id-129823unb29438723" --secret-string "STRING"
```

### Instead, this happened

Since this is a physical name, it means that it is impossible to deploy multiple instances of the app (e.g. with `wing test`), and generally it's a hassle.

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

jogold:
If a name is provided then the `cloud.Secret` references an existing secret
in the target cloud. If no name is provided then a new secret is provisioned
in the target cloud.

⚠️ It is no longer possible to provision a secret with a well-defined/well-known name.

Closes #2252 
 
### DIFF
--- a/libs/wingsdk/src/target-awscdk/secret.ts
+++ b/libs/wingsdk/src/target-awscdk/secret.ts
@@ -1,3 +1,4 @@
+import { CfnOutput } from "aws-cdk-lib";
 import {
   ISecret as ICdkSecret,
   Secret as CdkSecret,
@@ -16,13 +17,21 @@ import { IInflightHost } from "../std";
  */
 export class Secret extends cloud.Secret {
   private readonly secret: ICdkSecret;
+  private readonly arnForPolicies: string;
 
   constructor(scope: Construct, id: string, props: cloud.SecretProps = {}) {
     super(scope, id, props);
 
-    this.secret = props.name
-      ? CdkSecret.fromSecretNameV2(this, "Default", props.name)
-      : new CdkSecret(this, "Default");
+    if (props.name) {
+      this.secret = CdkSecret.fromSecretNameV2(this, "Default", props.name);
+      this.arnForPolicies = `${this.secret.secretArn}-??????`;
+    } else {
+      this.secret = new CdkSecret(this, "Default");
+
+      this.arnForPolicies = this.secret.secretArn;
+
+      new CfnOutput(this, "SecretArn", { value: this.secret.secretName });
+    }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


libs/wingsdk/src/target-awscdk/secret.ts
```diff
@@ -12,14 +15,14 @@ import { IInflightHost } from "../std";
...
 
   constructor(scope: Construct, id: string, props: cloud.SecretProps = {}) {
     super(scope, id, props);
 
-    this.secret = new CdkSecret(this, "Default", {
-      secretName: props.name,
-    });
+    this.secret = props.name
+      ? CdkSecret.fromSecretNameV2(this, "Default", props.name)
+      : new CdkSecret(this, "Default");
```
Chriscbr, line 25: (not blocking) I wonder if we this is something we should mention in the docstring for `cloud.Secret`, or if this is something we could document in a dedicated README / conceptual docs page about the `cloud.Secret` resource?

jogold:
In `sim` we could "mimic" the provisioning by creating an entry in `~/.wing/secrets.json` with `null` as the value?

libs/wingsdk/src/target-awscdk/secret.ts
Chriscbr: Ah I forgot we have an issue related to this https://github.com/winglang/wing/issues/1438 - @jogold is this something you'd like to try? (Only if it interests you of course 🙂)

Chriscbr:
> In `sim` we could "mimic" the provisioning by creating an entry in `~/.wing/secrets.json` with `null` as the value?

Sounds reasonable to me. Maybe we want to log something during preflight so the user knows that a file was added to their system where secrets should be inserted

libs/wingsdk/src/target-awscdk/secret.ts
jogold: Maybe... not sure 🙂  will let know and assign myself if I do.

Issue #2141, opened 2023-04-18
Sim cloud.Api dependency cycle error


hasanaburayyan:
### I tried this

```js
bring cloud;

resource A {
  api: cloud.Api;
  init() {
    this.api = new cloud.Api();
    this.api.get("/endpoint1", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
      let text = "${this.api.url}/endpoint2";
      return cloud.ApiResponse {
        status: 200,
        body: Json text,
      };
    });
  }
}

new A();

```
new A();	

### I expected this:

Success		

### Instead, this happened

error: Dependency cycle detected: root/Default/A/cloud.Api => root/Default/A/cloud.Api/cloud.Api-OnRequestHandler-72c9d4e4 => root/Default/A/cloud.Api

### Is there a workaround?

None to my knowledge	

### Component

Compiler

### Anything else?

This issue will be resolved by using the new EventMappings resource introduced in PR: https://github.com/winglang/wing/pull/2120 but I wanted to make a separate issue since the implementation for API resource requires a bit of nuance that merits its own PR.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

hasanaburayyan:
@staycoolcall911 just a heads up this is a dupe of https://github.com/winglang/wing/issues/1700 intentionally (see `Anything else?` section)

hasanaburayyan:
Forgot to close with: https://github.com/winglang/wing/pull/2120

hasanaburayyan:
Oops closed wrong issue 😓

monadabot:
Congrats! :rocket: This was released in [Wing 0.14.7](https://github.com/winglang/wing/releases/tag/v0.14.7).

PR #2259, opened 2023-04-24
fix(sdk): api use event maps

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/2141

- [ ] Update wing spec to use path vocab

**Misc:**
- refactored `route` vocabulary to `path` to fit the proper definitions of the terms and be more inline with popular framework vocabulary of the terms.

See: 
- https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
- https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-set-up-lambda-proxy-integration-on-proxy-resource
- https://expressjs.com/en/4x/api.html#router 

### DIFF
--- a/docs/04-reference/wingsdk-api.md
+++ b/docs/04-reference/wingsdk-api.md
@@ -37,30 +37,30 @@ new cloud.Api(props?: ApiProps)
 
 | **Name** | **Description** |
 | --- | --- |
-| <code><a href="#@winglang/sdk.cloud.Api.connect">connect</a></code> | Add a inflight handler to the api for CONNECT requests on the given route. |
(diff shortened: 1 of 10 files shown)


Chriscbr (approved):
Looks great 👍 

This might be a breaking change for the Wing Console - please check with @skyrpex @polamoros if we need to do a parallel release for this feature

Chriscbr:
BTW if possible we should add a test case to `examples/` showing that the use case from the original issue is resolved

hasanaburayyan:
> BTW if possible we should add a test case to `examples/` showing that the use case from the original issue is resolved

🤦 I knew I was forgetting something great callout @Chriscbr

PR #2270, opened 2023-04-26
feat!: goodbye `resource`

### LINKED ISSUE #2264: Replace `resource` with `class` (and use `inflight class` for "normal" classes)
### Feature Spec

Wing classes are user-defined types. They are similar to classes in other object-oriented languages like JavaScript, TypeScript, Java, C#, etc.

Similar to function closures, Wing classes are implicitly bound to the phase in which they are *declared*.

By default, Wing classes are bound to the *preflight phase*. This means that objects can be instantiated (or *defined*) in any preflight scope:

```js
// preflight class declaration
class MyClass { }

// preflight class definition
new MyClass();
```

Classes can have methods and fields which can be used to interact with object.

In preflight classes, all fields and methods are implicitly associated with the preflight phase. To denote an initializer, field or method as an inflight member, use the `inflight` modifier:

```js
class MyClass {
  // preflight field
  my_name: str;

  // preflight method
  set_name(n: str): void;

  // inflight initializer
  inflight init() {
    // ...
  }

  // inflight field
  inflight var state: num;

  // inflight method
  inflight upload_file(...): void;
}
```

From preflight context, it is possible to access only the preflight API of the object. From inflight context, it is possible to access the inflight API and also any *capturable* preflight fields (non reassignable, immutable, ...).

It is not allowed to define preflight objects from an inflight scope:

```js
inflight () => {
  new MyClass();
//^^^^^^^^^^^^ Cannot define an preflight object of type "Bucket" from an inflight scope
};
```

It is possible to declare a class as an `inflight class`, in which case it can be defined within inflight scopes:

```js
inflight class User {
  init(name: str, last: str) { ... }
}

inflight () => {
  let user = new User("boom", "bang");
  // ...
};
```

It is not allowed to define an inflight object from preflight scopes:

```js
new User();
//^^^^^^^^^^^^ Cannot define a inflight object of type "User" from a preflight scope
```

Preflight objects also have an API that can be used to inspect their relationships to other preflight objects:

```js
my_bucket.node.scope; // <-- the preflight object in which `my_bucket` was defined
```

This API is very powerful and needed in order to be able to synthesize desired state models for your application such as Terraform configurations. Under the hood, preflight classes use [constructs](https://github.com/aws/constructs) to allow interoperability with the CDK ecosystem.

If a class (or a function closure for that matter) is declared within an inflight scope, the `inflight` is implied:

```js
inflight () => {
  class Foo {} // <-- the "inflight" modifier is not needed here
  new Foo();
};
```

### Use Cases

Simplify the Wing mental model by eliminating the concept of "resources" in the language and aligning it to the concept of "inflight" (which is already a new concept).

### Implementation Notes

_No response_

### Component

Language Design

### Community Notes

eladb:
In order to reduce the number of new concepts in Wing and to improve the mental model behind the language, we decided to get rid of the concept of "resource" and instead simply associate a phase modifier with classes, similarly to what we have for functions.

The default phase is preflight, so basically when a class is declared within a preflight scope, it will be a "preflight class". These classes cannot be instantiated within inflight scopes. To declare a class that can be instantiated within an inflight scope, use `inflight class`.

Before             |After
-------------------|-----------------------
`resource Foo { }` | `class Foo { }`
`class Bar { }`    | `inflight Bar { } `

This change merely changes the language, but not the compiler implementation. Internally we still maintain the two concepts, but that's something we should clean up of course.

Resolves #2264

BREAKING CHANGE: use `class` instead of `resource` to declare preflight types and use `inflight class` instead of just `class` to declare inflight-only types.


docs/04-reference/winglang-spec.md
```diff
@@ -1248,8 +1253,8 @@ expected from a call and it is not being caught.
 Wing recommends the following formatting and naming conventions:
 
 - Interface names should start with capital letter "I"
-- Class, struct, interface, and resource names should be TitleCased
-- Members of classes, interfaces, and resources cannot share the same TitleCased
+- Class, struct and interface names should be TitleCased
+- Members of classes and interfaces cannot share the same TitleCased
```
MarkMcCulloh, line 1257: no oxford comma 🥲

```suggestion
- Class, struct, and interface names should be TitleCased
- Members of classes, and interfaces cannot share the same TitleCased
```

libs/wingc/src/type_check.rs
```diff
@@ -1281,7 +1281,7 @@ impl<'a> TypeChecker<'a> {
 							(&class.env, &class.name)
 						} else {
 							return self.general_type_error(format!(
-								"Cannot create the resource \"{}\" in inflight phase",
+								"Cannot create the preflight class \"{}\" in inflight phase",
```
MarkMcCulloh, line 1284:
```suggestion
								"Cannot create preflight class \"{}\" in inflight phase",
```

libs/tree-sitter-wing/grammar.js
```diff
@@ -186,7 +187,7 @@ module.exports = grammar({
 
     resource_definition: ($) =>
```
MarkMcCulloh, line 188: Followup to remove `resource_definition` to consolidate `class_definition`?

docs/04-reference/winglang-spec.md
```diff
@@ -2291,13 +2299,13 @@ acts like C `#include`s. Symbols collision is fatal in this style of imports.
...
-In inflight, the above excluding resources are importable.  
+In preflight, anything with `public` at block scope level is importable.
+This includes functions, classes, structs and interfaces.
+In inflight, the above excluding preflight classes are importable.
 Variables are not exportable.
 
-Resources are not usable in inflight functions. There is no synthesizer inside
-and no deployment system in the inflight body to synthesize inflight resources.
+Preflight classes are not usable in inflight functions. There is no synthesizer inside
+and no deployment system in the inflight body to provision.
```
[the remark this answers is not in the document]
MarkMcCulloh: Yeah "usable" seems too broad here. I assume this was the intended meaning?

```suggestion
Preflight classes cannot be instantiated in inflight functions. There is no synthesizer inside
and no deployment system in the inflight body to provision.
```

libs/tree-sitter-wing/grammar.js
eladb: But of course... but not in this PR

docs/04-reference/winglang-spec.md
```diff
@@ -1849,43 +1854,44 @@ resource Foo {
...
 support (normal strings as well as shell strings).  
-"scope" must be an expression of resource type.
+"scope" must be an expression that resolves to a preflight object.
 
-Resources can be captured into inflight functions and once that happens, inside
+Preflight objects can be captured into inflight scopes and once that happens, inside
 the capture block only the inflight members are available.
 
-Resources can extend other resources (but not structs) and implement interfaces.
+Preflight classes can extend other preflight classes (but not structs) and implement preflight interfaces.
```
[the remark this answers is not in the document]
eladb: Added to the interfaces section... We need to rewrite this doc, but for now it's okay I think.

docs/03-concepts/01-resources.md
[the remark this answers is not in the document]
eladb: Tweaked this a bit for now.

Chriscbr (approved):

PR #2288, opened 2023-04-26
chore(repo): remove instead word from bug label

ekeren:
The recent change in this template was to replace the order of the expected and the actual result. The "instead" word was relevant before but doesn't make sense now.


Chriscbr (approved):
Ah this was bothering me too 😅 thanks for the fix!

PR #2295, opened 2023-04-27
fix: lsp completion does not show namespaced types or static members

MarkMcCulloh:
### Type annotations in `new` expressions
image

### Statics members
image

### Changes

- The grammar now allows new expressions like `new cloud.Bucket`
  - Our parser will still yell at you about it
- Most of the changes here are due to the added TypeAnnotationKind enum and changed TypeAnnotation to a struct that also has a span

## Still does not work
- If the jsii type is not imported, it will not show up in completion
- type annotations in variable assignment/declaration don't work
- new expressions really should only show classes
- probably other stuff

### DIFF
--- a/libs/tree-sitter-wing/grammar.js
+++ b/libs/tree-sitter-wing/grammar.js
@@ -63,6 +63,7 @@ module.exports = grammar({
     custom_type: ($) =>
       prec.right(seq(
         field("object", $._type_identifier),
+        // While the final "fields" identifier is optional in this grammar, upstream parsing will fail if it is not present
         repeat(seq(".", optional(field("fields", $._type_identifier))))
       )),
 
@@ -80,6 +81,7 @@ module.exports = grammar({
             )
           ),
           choice(".", "?."),
+          // While the "property" identifier is optional in this grammar, upstream parsing will fail if it is not present
           optional(field("property", $._member_identifier))
         )
       ),
@@ -399,6 +401,7 @@ module.exports = grammar({
       prec.right(seq(
         "new",
         field("class", choice($.custom_type, $.mutable_container_type)),
+        // While "args" is optional in this grammar, upstream parsing will fail if it is not present
         field("args", optional($.argument_list)),
         field("id", optional($.new_object_id)),
         field("scope", optional($.new_object_scope))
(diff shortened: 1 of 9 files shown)


eladb:
@MarkMcCulloh do we have issues opened for each item in the "Still does not work" section?

libs/wingc/src/fold.rs
```diff
@@ -413,26 +413,71 @@ pub fn fold_type_annotation<F>(f: &mut F, node: TypeAnnotation) -> TypeAnnotatio
...
-			param_types: t.param_types.into_iter().map(|t| f.fold_type_annotation(t)).collect(),
-			return_type: t.return_type.map(|t| Box::new(f.fold_type_annotation(*t))),
-			phase: t.phase,
-		}),
-		TypeAnnotation::UserDefined(t) => TypeAnnotation::UserDefined(f.fold_user_defined_type(t)),
+	match node.kind {
+		TypeAnnotationKind::Number => TypeAnnotation {
+			kind: TypeAnnotationKind::Number,
+			span: node.span,
+		},
```
Chriscbr, line 420: Small simplification to avoid having to copy `span` all of those times
```rust
	let kind = match node.kind {
		TypeAnnotationKind::Number => TypeAnnotationKind::Number,
		TypeAnnotationKind::String => TypeAnnotationKind::String,
		...
	};
	TypeAnnotation {
	 	kind,
	 	span: node.span,
	 }
```

libs/tree-sitter-wing/grammar.js
```diff
@@ -396,13 +396,13 @@ module.exports = grammar({
       ),
 
     new_expression: ($) =>
-      seq(
+      prec.right(seq(
         "new",
         field("class", choice($.custom_type, $.mutable_container_type)),
-        field("args", $.argument_list),
+        field("args", optional($.argument_list)),
```
Chriscbr, line 402: does this mean `new cloud.Bucket;` is now valid in Wing? Feels like we should add an example Wing file or mention it in the spec if that's the case

If this is just an optimization for better parsing or better diagnostics, then let's add a comment here to clarify that arguments aren't actually optional

MarkMcCulloh: No, our parser will catch it. I mean technically it could work but there is currently an intentional error.

I can add a comment for this and other similar ones (e.g. `a.b.` is valid as well)

Chriscbr (approved):

PR #2300, opened 2023-04-27
fix(compiler): preflight objects cannot reference free variables inflight

Chriscbr:
Re-roll of #2211 with some modifications so the implementation keeps inflight clients in their own separate files. I went through a couple iterations to get to this point, here were the potential problems I found with them:
- as @eladb called out, inlining inflight clients into `preflight.js` as implemented in #2205 seems like it could be problematic if clients need to reference other inflight clients (for example, to access static methods)
- another idea I had was to put all inflight clients all in a single file (`preflight.js` and `inflight.js` - cute, right?) - the problem with this is that global variables could conflict with each other, and there might be scoping issues if/when the user starts defining resources at places besides the top level of a wing file

original description:

---

This pull request adds support for capturing [free variables](https://en.wikipedia.org/wiki/Free_variables_and_bound_variables) in resource classes, a feature that allows Wing to generate more expressive code for inflight clients. It implements a new visitor, modifies the code generation, and adds a new test case and a snapshot to verify the functionality.

The approach I've taken differs from the solution outlined in #1447. The original issue suggests that we transform free variables into resource fields. I saw some risks with this approach since all usages of the global value inside inflight constructors and inflight methods would also need to be transformed (for example, changing `global_bucket` to `this.global_bucket`), and tracking the list of these mappings so that they only apply to inflight code, and preventing name conflicts, could be tricky.

Instead, this PR augments jsification so that free variables are effectively also emitted as free variables in the JavaScript client code. 

I'm not sure this approach is any better, but it seems to work just as well for a number of cases I've tried.

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -1513,6 +1513,12 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'ast> {
 		let binding = parts.iter().collect::<Vec<_>>();
 		let qualification = binding.split_at(index).1.iter();
 
+		let fmt = |x: Iter<&Component>| x.map(|f| f.text.to_string()).collect_vec();
+		let mut key = fmt(capture.iter()).join(".");
+		if is_field_reference {
+			key = format!("this.{}", key);
+		}
+
 		// if our last captured component is a resource and we don't have
 		// a qualification for it, it's currently an error.
 		if let Some(c) = capture.last() {
@@ -1523,9 +1529,9 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'ast> {
 							level: DiagnosticLevel::Error,
 							message: format!(
 								"Unable to qualify which operations are performed on '{}' of type '{}'. This is not supported yet.",
-								c.text, v.type_,
+								key, v.type_,
 							),
-							span: Some(c.symbol.span.clone()),
+							span: Some(node.span.clone()),
 						});
 
 						return;
@@ -1534,11 +1540,6 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'ast> {
 			}
 		}
 
-		let fmt = |x: Iter<&Component>| x.map(|f| f.text.to_string()).collect_vec();
-		let mut key = fmt(capture.iter()).join(".");
-		if is_field_reference {
-			key = format!("this.{}", key);
-		}
 		let ops = fmt(qualification);
 
 		self.references.entry(key).or_default().extend(ops);
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -19,7 +19,7 @@ use crate::{
 	ast::{
 		ArgList, BinaryOperator, Class as AstClass, ClassField, Expr, ExprKind, FunctionBody, FunctionBodyRef,
 		FunctionDefinition, Initializer, InterpolatedStringPart, Literal, MethodLike, Phase, Reference, Scope, Stmt,
-		StmtKind, Symbol, TypeAnnotationKind, UnaryOperator, UserDefinedType,
+		StmtKind, Symbol, TypeAnnotationKind, UnaryOperator, UserDefinedType, UtilityFunctions,
 	},
 	debug,
 	diagnostic::{Diagnostic, DiagnosticLevel, Diagnostics},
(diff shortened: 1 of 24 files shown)


eladb:
By the way:

> another idea I had was to put all inflight clients all in a single file

Is also not idea because bundlers won't be able to tree shake this properly (most bundler tree shake at the file level afaik).

tools/hangar/__snapshots__/test_corpus/resource_captures_globals.w_compile_tf-aws.md
```diff
@@ -0,0 +1,500 @@
...
+      {((cond) => {if (!cond) throw new Error(`assertion failed: '((await global_array_of_str.at(0)) === "hello")'`)})(((await global_array_of_str.at(0)) === "hello"))};
+      {((cond) => {if (!cond) throw new Error(`assertion failed: '((global_map_of_num)["a"] === (-5))'`)})(((global_map_of_num)["a"] === (-5)))};
+      {((cond) => {if (!cond) throw new Error(`assertion failed: '(await global_set_of_str.has("a"))'`)})((await global_set_of_str.has("a")))};
+      {((cond) => {if (!cond) throw new Error(`assertion failed: '(global_another.my_field === "hello!")'`)})((global_another.my_field === "hello!"))};
+      (await global_another.first.my_resource.put("key","value"));
+      {((cond) => {if (!cond) throw new Error(`assertion failed: '((await global_another.my_method()) > 0)'`)})(((await global_another.my_method()) > 0))};
+    }
+  }
+}
+exports.MyResource = MyResource;
```
eladb, line 66: Perhaps I can suggest a different approach? (a bit more idiomatic to javascript):

Instead of a separate `setupGlobals` functions, why not just wrap the entire file in a function and use it to bind the global symbols:

```js
module.exports = function(globals) {
  class MyResource { ... /* simply references globals */ }
  return MyResource;
};
```

Then, the usage is:

```js
(await (async () => {
  const MyResource = require("${self_client_path}")({
    global_another: ${global_another_client},
    global_array_of_str: ${global_array_of_str_client},
    global_bool: ${global_bool_client},
    global_bucket: ${global_bucket_client},
    global_map_of_num: ${global_map_of_num_client},
    global_num: ${global_num_client},
    global_set_of_str: ${global_set_of_str_client},
    global_str: ${global_str_client},
  });
  const client = new MyResource({
    stateful: ${stateful_client},
  });
  if (client.$inflight_init) { await client.$inflight_init(); }
  return client;
})())
```

eladb (changes requested):

tools/hangar/__snapshots__/test_corpus/resource_captures_globals.w_compile_tf-aws.md
Chriscbr: Right on 👍

libs/wingc/src/jsify.rs
```diff
@@ -1468,7 +1531,10 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'ast> {
 		}
 
 		let fmt = |x: Iter<&Component>| x.map(|f| f.text.to_string()).collect_vec();
-		let key = format!("this.{}", fmt(capture.iter()).join("."));
+		let mut key = fmt(capture.iter()).join(".");
+		if is_field_reference {
```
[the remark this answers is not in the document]
Chriscbr: If I try removing this condition and always start from index 0, it seems to cause a regression, maybe we can investigate this later...

libs/wingc/src/jsify.rs
```diff
@@ -1490,13 +1557,25 @@ impl Display for Component<'_> {
 }
 
 impl<'a> FieldReferenceVisitor<'a> {
-	fn analyze_expr(node: &'a Expr) -> Vec<Component> {
+	fn analyze_expr(&self, node: &'a Expr) -> Vec<Component> {
 		match &node.kind {
 			ExprKind::Reference(Reference::Identifier(x)) => {
+				let scope = match &self.function_def.body {
```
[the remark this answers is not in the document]
Chriscbr: if this is a free variable, we still end up analyzing the expression. I wasn't sure what to rename this piece of our compiler to instead of `FieldReferenceVisitor`. My current idea is that this visitor looks inside functions for expressions that require bindings to be generated, and collects the operations that are used on them in `references` (or generates diagnostics if the expression shouldn't be possible to capture) - it's no longer limiting the analysis to fields anymore.

tools/hangar/__snapshots__/test_corpus/resource_captures_globals.w_compile_tf-aws.md
```diff
@@ -0,0 +1,677 @@
+# [resource_captures_globals.w](../../../../examples/tests/valid/resource_captures_globals.w) | compile | tf-aws
+
+## clients/Another.inflight.js
+```js
+module.exports = function($globals) {
+  const { global_counter } = $globals;
```
eladb, line 6: Nit: this could be:

```suggestion
module.exports = function({ global_counter }) {
```

eladb (approved):
Looks great! Ship it!

PR #2304, opened 2023-04-28
feat(sdk): `std.util.sleep()`

GuyDev1:

### DIFF
--- a/libs/wingsdk/src/std/duration.ts
+++ b/libs/wingsdk/src/std/duration.ts
@@ -9,7 +9,7 @@ export class Duration {
    * @returns a new `Duration` representing `amount` Minutes.
    */
   public static fromMinutes(amount: number): Duration {
-    return new Duration(amount * 60);
+    return new Duration(amount * 60 * 1000);
   }
 
   /**
@@ -19,7 +19,7 @@ export class Duration {
    * @returns a new `Duration` representing `amount` Hours.
    */
   public static fromHours(amount: number): Duration {
-    return new Duration(amount * 60 * 60);
+    return new Duration(amount * 60 * 60 * 1000);
   }
 
   /**
@@ -29,18 +29,28 @@ export class Duration {
    * @returns a new `Duration` representing `amount` Seconds.
    */
   public static fromSeconds(amount: number): Duration {
+    return new Duration(amount * 1000);
+  }
+
+  /**
+   * Create a Duration representing an amount of milliseconds
+   *
+   * @param amount the amount of milliseconds the `Duration` will represent.
+   * @returns a new `Duration` representing `amount` milliseconds.
+   */
+  public static fromMilliseconds(amount: number): Duration {
     return new Duration(amount);
   }
 
   /**
-   * Return the total number of seconds in this Duration
+   * Return the total number of milliseconds in this Duration
    *
-   * @returns the value of this `Duration` expressed in Seconds.
+   * @returns the value of this `Duration` expressed in Milliseconds.
    */
-  public readonly seconds: number;
+  public readonly milliseconds: number;
 
-  private constructor(seconds: number) {
-    this.seconds = seconds;
+  private constructor(milliseconds: number) {
+    this.milliseconds = milliseconds;
   }
 
   /**
@@ -59,5 +69,14 @@ export class Duration {
    */
   public get hours() {
     return this.minutes / 60;
+  }
+
+  /**
+   * Return the total number of milliseconds in this Duration
+   *
+   * @returns the value of this `Duration` expressed in Milliseconds.
+   */
+  public get seconds() {
+    return this.milliseconds / 1000;
   }
 }
--- a/libs/wingsdk/src/std/duration.ts
+++ b/libs/wingsdk/src/std/duration.ts
@@ -3,23 +3,13 @@
  */
 export class Duration {
   /**
-   * Create a Duration representing an amount of minutes
+   * Create a Duration representing an amount of milliseconds
    *
-   * @param amount the amount of Minutes the `Duration` will represent.
-   * @returns a new `Duration` representing `amount` Minutes.
+   * @param amount the amount of milliseconds the `Duration` will represent.
+   * @returns a new `Duration` representing `amount` milliseconds.
    */
-  public static fromMinutes(amount: number): Duration {
-    return new Duration(amount * 60 * 1000);
-  }
-
-  /**
-   * Create a Duration representing an amount of hours
-   *
-   * @param amount the amount of Hours the `Duration` will represent.
-   * @returns a new `Duration` representing `amount` Hours.
-   */
-  public static fromHours(amount: number): Duration {
-    return new Duration(amount * 60 * 60 * 1000);
+  public static fromMilliseconds(amount: number): Duration {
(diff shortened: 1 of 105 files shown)


libs/wingsdk/src/std/duration.ts
```diff
@@ -60,4 +70,13 @@ export class Duration {
   public get hours() {
     return this.minutes / 60;
   }
+
+  /**
+   * Return the total number of milliseconds in this Duration
```
revitalbarletz, line 75: you mean return the number seconds?

libs/wingsdk/src/std/duration.ts
```diff
@@ -9,7 +9,7 @@ export class Duration {
    * @returns a new `Duration` representing `amount` Minutes.
    */
   public static fromMinutes(amount: number): Duration {
-    return new Duration(amount * 60);
```
revitalbarletz, line 12: No tests were affected by this change?

libs/wingsdk/src/std/duration.ts
```diff
@@ -29,18 +29,28 @@ export class Duration {
...
 
   /**
-   * Return the total number of seconds in this Duration
+   * Return the total number of milliseconds in this Duration
    *
-   * @returns the value of this `Duration` expressed in Seconds.
+   * @returns the value of this `Duration` expressed in Milliseconds.
    */
-  public readonly seconds: number;
+  public readonly milliseconds: number;
```
ekeren, line 50: I would consider removing this member and exposing milliseconds similar to how seconds, minutes and hours are exposed. 


Also, should be add days?

eladb (changes requested):

libs/wingsdk/src/std/sleep.ts
```diff
@@ -0,0 +1,4 @@
+import { Duration } from "./duration";
+
+export const sleep = async (duration: Duration): Promise<void> =>
```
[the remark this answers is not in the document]
eladb: @GuyDev1 let's just put this under a `Util` class for now:

```js
export class Util {
  public static sleep(duration: Duration): Promise<void> { ... }
}
```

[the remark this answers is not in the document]
GuyDev1: @eladb Got it, will do.
I'm usually not in favor of Util classes as they tend to become garbage collectors (pun intended) of various unrelated functions. But it sounds like we don't have much choice with JSII and it's only cosmetic, so I'll do it :)

libs/wingsdk/src/std/duration.ts
GuyDev1: I didn't see any tests for std functions (other than the one I added to sleep)
And since the internals return the same underlying duration, we should be good

libs/wingsdk/src/std/duration.ts
GuyDev1: We can't because we need a base metric to work with (in the private constructor).
That's why seconds was configured in that way when seconds was the base metric.

libs/wingsdk/src/std/duration.ts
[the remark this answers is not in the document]
GuyDev1: I might take it later as well :)

libs/wingsdk/src/std/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { Duration } from "./duration";
+
+export class Utils {
+  public static async sleep(duration: Duration): Promise<void> {
```
eladb, line 4: Missing api docs here

libs/wingsdk/src/std/index.ts
```diff
@@ -9,3 +9,4 @@ export * from "./resource";
 export * from "./set";
 export * from "./string";
 export * from "./util";
```
eladb, line 11: Do we really need this export?

libs/wingsdk/src/std/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { Duration } from "./duration";
+
+export class Utils {
```
eladb, line 3: Let's move this from `std` to `util`. I think it makes more sense to use it like this:

```js
bring util:

util.sleep(...);
```

And until we support globals:

```js
bring util;

util.Util.sleep();
```

Then we can change compiler import logic to treat classes with the same name as the namespace as "global".

eladb (changes requested):

libs/wingsdk/src/std/utils.ts
GuyDev1: Sounds good, will do

libs/wingsdk/src/std/index.ts
GuyDev1: I assumed it's required for the compiler, we're moving it to a different namespace so I think we should be alright

libs/wingsdk/src/std/utils.ts
GuyDev1: @eladb Let me know if you want me to open an issue for the compiler import logic

libs/wingsdk/src/util/util.ts
```diff
@@ -0,0 +1,16 @@
+import { Duration } from "../std";
+
+/**
+ * A collection of utility functions
+ */
+export class Util {
+  /**
+   * Returns a Promise that resolves after the given Duration has passed.
```
eladb, line 8:
```suggestion
   * Suspends execution for a given duration.
```

libs/wingsdk/src/util/util.ts
```diff
@@ -0,0 +1,16 @@
...
+   *
+   * @param duration The duration that has to pass before the Promise is resolved.
+   * @returns a Promise that resolves after the given Duration has passed.
```
eladb, line 11: In async functions, no need to state the obvious (that a promise is returned). And then, this function basically doesn't return anything.

```suggestion
```

eladb (approved):

libs/wingsdk/src/util/util.ts
```diff
@@ -22,4 +24,13 @@ export class Util {
   public static tryEnv(name: string): string | undefined {
     return process.env[name];
   }
+
+  /**
+   * Suspends execution for a given duration.
+   *
+   * @param duration The duration that has to pass before the Promise is resolved.
+   */
```
eladb, line 32: I think you'll need to annotate this with:

```suggestion
   * @inflight
   */
```

eladb (changes requested):
Please add a wing test

GuyDev1:
image
Happened to @tsuf239 as well, seems utils is unreachable inside inflight functions (and sleep should run in an inflight context).

Changing to draft until it's fixed

staycoolcall911:
Hey @GuyDev1, we have an issue for that: #2576. I think you can proceed with `sleep` unrelated and we'll solve #2576 soon (it's high priority)

ShaiBer:
keep

ekeren:
@GuyDev1 , I needed this functionality for the examples in the docs, so I took the liberty to take the code from here and publish it https://github.com/winglang/wing/pull/2906

Chriscbr:
Closing as superseded by #2906

PR #2306, opened 2023-04-28
feat: reduce package size from 890mb to 181mb

### LINKED ISSUE #1800: winglang dependency closure is over 3GB
### Community Note

### Feature Spec

The largest dependencies are:

```
1.3G	@cdktf/provider-aws
992M	@cdktf/provider-azurerm
624M	@cdktf/provider-google
```

### Use Cases

Reduce the time it takes to install wing

### Implementation Notes

To avoid needing these as dependencies, we can import the CDKTF bindings we need directly using the `cdktf get` command as part of the build process (or just keep the generated code in source control)

### Component

SDK, CLI

Chriscbr:
Closes #1800

Reduces the size of the `@winglang/sdk` package from ~890MB (131MB compressed) to 181MB (71.8MB compressed). This is achieved by removing the `@cdktf/provider-xxx` libraries from our bundled dependencies, and instead generating the TypeScript bindings for various terraform resources as part of our build.

Since CDKTF's binding generator creates a lot of files (over 7000 in our case), it's not practical to save them in git, so we now generate them as a separate build step. I've added a note in our contributing guide, and I'll also post a notice in our #dev slack channel.

Thanks to how TypeScript's compilation works, only bindings for the resources we use get compiled into JavaScript in the new `libs/src/.gen` folder, so we didn't need a separate tree shaking process to get the bundle size improvement.


examples/plugins/permission-boundary.js
```diff
@@ -20,7 +20,7 @@ class PermissionBoundaryAspect {
   }
 
   visit(node) {
-    if (node instanceof iam_role.IamRole) {
+    if (node.terraformResourceType === "aws_iam_role") {
       node.permissionsBoundary = this.permissionBoundaryArn;
     }
```
Chriscbr, line 25: The constructs created by the SDK are no longer the same classes as those provided by `@cdktf/provider-aws` so `instanceof` will fail - hence this change

libs/wingsdk/.projenrc.ts
```diff
@@ -54,6 +58,7 @@ const project = new cdk.JsiiProject({
     "vm2",
   ],
   devDeps: [
+    `@cdktf/provider-aws@^12.0.1`, // only for testing Wing plugins
```
Chriscbr, line 61: I tried removing this as a dependency entirely, but then the Wing plugin E2E tests began failing because there was no copy of this dependency they could use. I tried fixing that by adding the dependency in the `examples/plugins` folder but it still couldn't find the dependency. So for now this dependency is just here

tools/hangar/__snapshots__/tree_json.ts.snap
```diff
@@ -10,8 +10,8 @@ exports[`tree.json for an app with many resources 1`] = `
             "children": {
               "Code": {
                 "constructInfo": {
-                  "fqn": "@cdktf/provider-aws.s3Bucket.S3Bucket",
-                  "version": "12.0.2",
+                  "fqn": "cdktf.TerraformResource",
+                  "version": "0.15.2",
                 },
```
Chriscbr, line 15: Looks like we lost some information here (the tree.json no longer shows concrete information about specific AWS resources). This is a side effect of how we're no longer using the `@cdktf/provider-aws` package. It's not great, but I'm not sure how we can avoid it with our solution. (The alternative might be to "dynamically" install these dependencies when the user tries compiling their Wing app to `tf-aws` for the first time, or breaking up the SDK into multiple packages like @eladb was investigating).

libs/wingsdk/.projenrc.ts
```diff
@@ -1,7 +1,15 @@
-import { cdk, javascript } from "projen";
+import { JsonFile, cdk, javascript } from "projen";
 import rootPackageJson from "../../package.json";
 
 const JSII_DEPS = ["constructs@~10.1.228"];
+const CDKTF_VERSION = "0.15.2";
+
+const CDKTF_BINDINGS = [
```
eladb, line 7:
```suggestion
const CDKTF_PROVIDERS = [
```

libs/wingsdk/.projenrc.ts
eladb: I think this is perfectly fine

tools/hangar/__snapshots__/tree_json.ts.snap
eladb: Is there a way to inject `terraformResourceType` into `tree.json` instead?

eladb (approved):

eladb:
Updated title

eladb:
I think the next step in this project is to remove `aws-cdk-lib` from the SDK and add code to the CLI that will ask the user to install `aws-cdk-lib` if they want to compile to the `awscdk` target.

tools/hangar/__snapshots__/tree_json.ts.snap
Chriscbr: #2347

PR #2307, opened 2023-04-28
fix(sdk): table insert and update now require the primary key value

marciocadev:
changing the `table.update(row: Json)` and `table.insert(row: Json` to `table.update(key: str, row: Json)` and `table.insert(key: str, row: Json)`

I also update the Table SDK Spec


docs/04-reference/wingsdk-spec.md
```diff
@@ -1297,12 +1297,12 @@ resource Table {
   /**
    * Insert a row into the table.
    */
-  inflight insert(row: Map<Json>): void;
+  inflight insert(key: str, row: Json): void;
```
Chriscbr, line 1300: Thanks for updating the spec 🙏

Chriscbr (approved):

docs/04-reference/wingsdk-spec.md
marciocadev: Yeap 😄 
 
I thought it would be important to reflect the changes in the current documentation to avoid confusion for those who try to use it.

PR #2309, opened 2023-04-30
feat(sdk): adding bucket event hooks to awscdk

marciocadev:


libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -41,6 +51,38 @@ export class Bucket extends cloud.Bucket {
...
+
+    // TODO: remove this constraint by adding generic permission APIs to cloud.Function
+    if (!(handler instanceof AWSTopic)) {
+      throw new Error("Topic only supports creating tfaws.Function right now");
+    }
+
+    this.bucket.addEventNotification(
+      EVENTS[actionType],
+      new SnsDestination(handler._topic)
+    );
```
Chriscbr, line 81: Is it possible to directly add the Lambda as a destination? I forget if this is something aws-cdk-lib handles for us - if that's possible we might be able to just override the `onCreate` from `cloud.Bucket` base class, and avoid creating these topics altogether

marciocadev: It's possible Chris

I didn't do this way because I wanted to follow what was developed in `target-tf-aws`, but of it's not necessary to do the same, I can send the events via DynamoDB Streams to Lambda, which is actually much simpler.

Chriscbr: How is the DynamoDB stream involved? Not sure I understand

marciocadev: Sorry, I was thinking about capture event from DynamoDB to Table resource

I'm working on this and made a confusion

But send a every from Bucket to Lambda is easy, we don't need the Topic, I only use because I follow the `target-tf-aws` implementation

Chriscbr: Gotcha, no worries :) yeah let's try the approach without topics first since it should result in fewer resources created for the user, and it might event be less code

Chriscbr:
@marciocadev Hey, wanted to just check in to see if you are still working on this. Let us know if there is anything we can help with

marciocadev:
I will come back soon @Chriscbr

Chriscbr (approved):
Nice work! Thanks @marciocadev 🙏

PR #2310, opened 2023-04-30
rfc: id for preflight instances

ekeren:
## Background

In the current implementation of Wing, when instantiating a new preflight class, the compiler provides a default ID for the instance. 
In cases where the default ID cannot be made unique, a compiler error occurs.

In the last wingathon, more than 70% of preflight instances in the Wing research repository are created with a custom ID. 

## Observed Patterns

When examining the code, the following patterns can be observed:

1. Improved readability of preflight instances, e.g.:

    [dogfooding/publications-etl](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741//app.w#L70):
     ```ts (wing)
     this.initialized = new cloud.Counter() as "initialized";
     ```
2. Name duplications - situations where the ID (or part of it) is also reflected in the constructor:

    [dogfooding/release-webhook](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/release-webhook/main.w#L207):
     ```ts (wing)
     let slack_token = new cloud.Secret(name: "slack-token") as "Slack Token";
     ```
     You can also consider situations where `cloud.Bucket` and `cloud.Test` where the ID of the instance and the name of bucket/test are connected 
3. Multiple instances of the same type in the same scope:

     [dogfooding/publications-etl](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/publications-etl/app.w#LL31-L33C58):
     ```ts (wing)
     this.inbox = new cloud.Bucket() as "inbox";
     this.transformations = new cloud.Bucket() as "transformations";
     this.catalogs = new cloud.Bucket() as "xslt catalog";
     ```

### DIFF
--- a/docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
+++ b/docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
@@ -18,23 +18,26 @@ In the last wingathon, more than 70% of preflight instances in the Wing research
 When examining the code, the following patterns can be observed:
 
 1. Improved readability of preflight instances, e.g.:
-  -  [dogfooding/publications-etl](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741//app.w#L70)
-  ```ts (wing)
-    this.initialized = new cloud.Counter() as "initialized";
-  ```
-1. Name duplications - situations where the ID (or part of it) is also reflected in the constructor:
-  - [dogfooding/release-webhoo](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/release-webhook/main.w#L207):
-  ```ts (wing)
+  
+    [dogfooding/publications-etl](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741//app.w#L70)
+     ```ts (wing)
+     this.initialized = new cloud.Counter() as "initialized";
+     ```
+2. Name duplications - situations where the ID (or part of it) is also reflected in the constructor:
+  
+    [dogfooding/release-webhoo](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/release-webhook/main.w#L207):
+     ```ts (wing)
      let slack_token = new cloud.Secret(name: "slack-token") as "Slack Token";
-  ```
-  - You can also consider situations where `cloud.Bucket` and `cloud.Test` where the name of the test and the name ofd 
-1. Multiple instances of the same type in the same scope:
-  - [dogfooding/publications-et](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/publications-etl/app.w#LL31-L33C58)
-  ```ts (wing)
-    this.inbox = new cloud.Bucket() as "inbox";
-    this.transformations = new cloud.Bucket() as "transformations";
-    this.catalogs = new cloud.Bucket() as "xslt catalog";
-  ```
+     ```
+     You can also consider situations where `cloud.Bucket` and `cloud.Test` where the ID of the instance and the name of bucket/test are connected 
+3. Multiple instances of the same type in the same scope:
+   
+     [dogfooding/publications-et](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/publications-etl/app.w#LL31-L33C58)
+     ```ts (wing)
+     this.inbox = new cloud.Bucket() as "inbox";
+     this.transformations = new cloud.Bucket() as "transformations";
+     this.catalogs = new cloud.Bucket() as "xslt catalog";
+     ```
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
```diff
@@ -0,0 +1,70 @@
...
+
+When examining the code, the following patterns can be observed:
+
+1. Improved readability of preflight instances, e.g.:
+  
+    [dogfooding/publications-etl](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741//app.w#L70):
+     ```ts (wing)
+     this.initialized = new cloud.Counter() as "initialized";
+     ```
+2. Name duplications - situations where the ID (or part of it) is also reflected in the constructor:
```
eladb, line 26: For this use case, I am wondering if the resource has a physical name (in this case `name: "slack-token"`), then maybe the resource class itself can suggest an identifier that's based on that physical name?

docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
```diff
@@ -0,0 +1,70 @@
...
+     [dogfooding/publications-etl](https://github.com/winglang/research/blob/926453f5c5abd2e3158b4c8c659f1e790c99d741/dogfooding/publications-etl/app.w#LL31-L33C58):
+     ```ts (wing)
+     this.inbox = new cloud.Bucket() as "inbox";
+     this.transformations = new cloud.Bucket() as "transformations";
+     this.catalogs = new cloud.Bucket() as "xslt catalog";
+     ```
+
+## Proposal
+
+Differentiate between stateful instance (cloud.Bucket, Redis, cloud.Counter, cloud.Table, cloud.Queue) that require a stable ID,
```
eladb, line 44: I think that's a good direction generally. For resources with persistent state it makes sense to always require an ID and then make it the first argument of the initializer.

docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
```diff
@@ -0,0 +1,70 @@
...
+  ```
+  
+- For stateless instances, default the instance name to the corresponding variable:
+  ```ts (wing)
+  // ./app.w
+  bring cloud; 
+  let my_func = new cloud.Function(...); // named "my_func"
+  new cloud.Function(...); // when no variable, default to named "cloud.Function$1"
+  new cloud.Function(...); // when no variable, default to named "cloud.Function$2"
+  ```
```
ekeren, line 61: @eladb , do you see a problem with not having a stable id for non persistent resources?

docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
ekeren: When thinking about it, I think like the term `has_persistent_state` more then `statefull`

eladb: I feel that we need to do a bit more design work around stateful/persistent resources. I am wondering if for beta we should just get rid of this property altogether

ekeren: Does it have any meaning now?

eladb: No, it's just a future proofing junk :-)

docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
eladb: Yes, stable IDs are important even for non-persistent resources, mainly due to the fact that infrastructure is (sometimes) hard to provision and there could be "in-flight" transactions that you don't want to disrupt.

Take a VPC for example - it's a non persistent resource but you definitely don't want it to be recreated if it's identity changes unexpectedly.

Even a Lambda function, you don't want to recreated it unnecessarily - it would take longer to deploy, logs will get deleted (sort of persistent things), etc.

If think the low hanging fruit would be the cosmetic/ergonomic change to use the variable name as display name (if defined).

We *could* explore the idea of requiring an `id` for persistent resources like `cloud.Bucket` and `cloud.Table` by simply adding a positional argument:

```js
class Bucket {
  constructor(scope: Construct, id: string, actualId: string, props: BucketProps) {
    super(scope, actualId);

    // ...
  }
}
```

This means that when you define a `cloud.Bucket` in Wing, you'll do it like this:

```js
new cloud.Bucket("my-bucket");
```

Note that if you use `as "foo"` it will be ignored (we can add some special casing to the compiler to disallow of course).

Maybe we can try it out for buckets and see how it feels and then make a decision.

docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
eladb: Open an issue?

docs/06-contributors/999-rfcs/2022-04-30-preflight-instance-id.md
ekeren: This is great feedback, I'll work to add this information to the RFC.

PR #2314, opened 2023-04-30
feat(sdk) run sdk tests via hangar

tsuf239:
## Description
copilot:summary 

##Walkthrough 
copilot:walkthrough


tools/hangar/src/meta_comment.ts
```diff
@@ -25,7 +25,14 @@ export interface MetaComment {
...
    */
-  skipPlatforms?: typeof process.platform[];
+  skipPlatforms?: (typeof process.platform)[];
+
+  /**
+   * Should the test be skipped at all platform, all environments
+   *
+   * @default - Will not be skipped
+   */
+  skipTest?: boolean;
```
MarkMcCulloh, line 35:
```suggestion
  skip?: boolean;
```

Also is this used in the PR? If not we can probably omit it, I'm not sure we want to include tests that are always skipped.

MarkMcCulloh (approved):

Chriscbr:
Thinking about how to avoid future confusion - maybe we can name these the "sdk spec tests" in conversation to avoid ambiguity with the tests inside the SDK? The tests inside the sdk could be called the "sdk unit tests"

Chriscbr (approved):

tools/hangar/src/meta_comment.ts
tsuf239: Cool, I'll change it in my next PR (which includes SDK spec tests to skip- for example a test for queue.pop- that isn't an existent method yet)

Issue #2183, opened 2023-04-20
Add FAQ topic: Can I import custom Terraform modules into Wing?


ShaiBer:
### Feature Spec

From this reddit comment: https://www.reddit.com/r/devops/comments/12n4sna/comment/jgeyf66/?utm_source=share&utm_medium=web2x&context=3

> It looks like, assuming cdktf has a provider, it should work.

> This is both good and bad.

> Good because a lot of providers are converted, bad because quite a few are missing.

### Use Cases

Developer wanting to understand how they can bring their Terraform modules into Wing

### Component

Documentation

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.15.21](https://github.com/winglang/wing/releases/tag/v0.15.21).

PR #2320, opened 2023-04-30
chore(docs): faq - importing custom terraform modules

ShaiBer:
Fixes: #2183


docs/07-faq/061-import-custom-terraform-modules.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Can I import custom Terraform Modules into Wing?
+id: import-custom-terraform-modules
+keywords: [faq, winglang, Wing programming language, Wing language, Terraform]
+---
+
+You can import custom Terraform modules into Wing if they have a CDKTF provider.
```
eladb, line 7: Mmmm.... I don't think terraform modules have pre-built CDKTF providers. I think users will need to use `cdktf get` to generate the TypeScript code and publish it as a JSII library. It's a bit of a mess. We need to document this.

eladb (changes requested):

docs/07-faq/061-import-custom-terraform-modules.md
```diff
@@ -0,0 +1,10 @@
...
+
+You can use any resource from the vast [Terraform Registry](https://registry.terraform.io/) in your Wing applications as explained [here](https://github.com/winglang/wing/issues/489).
```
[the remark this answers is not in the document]
ShaiBer: I prefer to have a single source of truth if possible. WDYT?

docs/07-faq/061-import-custom-terraform-modules.md
ShaiBer: Yes, they don't. This is why I wrote they can be used `if` they have a provider :) 
Can we say for now that we know it is not trivial to create the provider, but that if one was created it can be used?

PR #2322, opened 2023-04-30
fix(docs): improve readme "why a language"

### LINKED ISSUE #2181: Improve "why a language" explanation
### Feature Spec

From Reddit: https://www.reddit.com/r/devops/comments/12n4sna/comment/jgeloz2/?utm_source=share&utm_medium=web2x&context=3 

> "I would recommend something like showing how some code would look in a hypothetical Typescript or Python version of Wing and then show the same code being smaller or better or easier in your language. Pick maybe the top 3 examples in your mind that your language is worth the learning-time investment because it's just so much better. Because otherwise that's a very big ask for a team that already knows Python or Typescript or Go."

> "I probably wouldn't compare it to the compiled output since that's not usually a fair thing, compiled output is usually not intended to be the same as human written code. But just imagine if you were to try and make something like Wing without a custom language, what would it probably end up looking like, and then show why Wing fixes some specific problem there.

> Because otherwise it's a very hard sell. Existing languages have linters and CI workflows and dependency management and a million other little things that will need to be rewritten for Wing. And just showing me some nice Wing code and saying "see, isn't it nice" doesn't really give me enough to want to take on that massive pile of work to use your project."

> "the cloud is fundamentally different from legacy computing, and legacy programming paradigms are a continuing source of friction when solutioning for the cloud. approaching these problems with a new coding paradigm allows us to provide purpose built tools that will reduce operational overhead by directly addressing the needs of the cloud practitioner.

> then some bullets that help convince me this paragraph isnt a lie, probably about conceptual features. this list could be a pared down and tweaked version of the list titled "Wing is designed to address these challenges in the following ways:" in the "new programming paradigm" click through link.

> then you can do whatever you want with the rest of the page."

### Use Cases

Developers wanting to understand why this is a language and not a library or language extension

### Implementation Notes

This should go to both the FAQ version and possibly the main readme of the project

### Component

Documentation

### Community Notes

eladb:
Fixes #2181

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ Wing is built by [Elad Ben-Israel](https://github.com/eladb), the guy behind the
 Existing languages are unable to express the distributed and infrastructure-heavy nature of cloud 
 applications, which are fundamentally different from applications that run on a single machine. 
 They are distributed systems which rely on cloud infrastructure to achieve their goals. The primary
-motivation for a new langugae is to allow developers to express both cloud infrastructure *and*
-application through a unified programming model.
+motivation for a new language is to allow developers to express both cloud infrastructure *and*
+application logic through a unified programming model.
 
 To allow Winglang code to express both infrastructure and application, the language introduces the concept of *execution phases*. 
 The "preflight" phase is where you define your cloud infrastructure and the "inflight" phase is where you define your application logic.
@@ -68,7 +68,7 @@ Now comes the cool part: the code that runs inside the inflight function interac
 through the *inflight methods* of these objects (`counter.inc()` and `bucket.put()`). These methods can only be
 called from inflight scopes. In existing languages, where there is no way to distinguish between multiple execution phases,
 it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific
-execution phase.
+execution phase in a type-safe manner.
 
 > ### Note for cloud experts 🤓
 >
--- a/README.md
+++ b/README.md
@@ -31,14 +31,12 @@ Wing is built by [Elad Ben-Israel](https://github.com/eladb), the guy behind the
 
 ## Why do we think the cloud needs a programming language? 🤔
 
-Existing languages are unable to express the distributed and infrastructure-heavy nature of cloud 
-applications, which are fundamentally different from applications that run on a single machine. 
-They are distributed systems which rely on cloud infrastructure to achieve their goals. The primary
-motivation for a new language is to allow developers to express both cloud infrastructure *and*
-application logic through a unified programming model.
+Cloud applications are are fundamentally different from applications that run on a single machine - 
+they are distributed systems that rely on cloud infrastructure to achieve their goals.
 
-To allow Winglang code to express both infrastructure and application, the language introduces the concept of *execution phases*. 
-The "preflight" phase is where you define your cloud infrastructure and the "inflight" phase is where you define your application logic.
+In order to be able to express both infrastructure and application logic in a safe and unified programming model, 
+Winglang has two execution phases: *Preflight* for infrastructure definitions and *inflight* is for runtime code.
+
 Preflight code is executed *during compilation* and produces the infrastructure configuration for your app (e.g. **Terraform**, **CloudFormation**, etc).
 Inflight code is compiled to **JavaScript** and executed on within cloud compute platforms in Node.js environments.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,38 +29,46 @@ Our mission is to bring back your creative flow and close the gap between imagin
...
+  bucket.put("file-${i}.txt", message);
+});
+```
 
-// An `inflight` represents code that runs later, on other machines,
-// and can interact with any cloud resources
-let hello_world = inflight () => {
-  bucket.put("hello.txt", "Hello, World!");
-};
+The `cloud.Bucket`, `cloud.Counter` and `cloud.Bucket` are *preflight objects*. They represent
```
ShaiBer, line 60:
```suggestion
The `cloud.Queue`, `cloud.Counter` and `cloud.Bucket` are *preflight objects*. They represent
```

README.md
```diff
@@ -29,38 +29,46 @@ Our mission is to bring back your creative flow and close the gap between imagin
...
+cloud infrastructure resources. When compiled to AWS, a Terraform file will be produced with the AWS implementation
```
ShaiBer, line 61: I don't want it to sound like we are an AWS shop

```suggestion
cloud infrastructure resources. When compiled to a specific cloud provider, such as AWS, a Terraform file will be produced with the provider's implementation
```

ShaiBer (changes requested):
I think that it's a step in the right direction and makes the readme shorter and more exact.
It doesn't answer all the feedback that arose from the Reddit post though because:
1. There is no direct comparison of the same code in Wing and in other languages/frameworks.
2. There is no clear statement of exactly what we can do with Wing that cannot be done with a library/compiler extension, and why. BTW, I think this part may be long and too technical for most, so maybe best to put as an FAQ topic and link to it

eladb:
> There is no clear statement of exactly what we can do with Wing that cannot be done with a library/compiler extension

I thought this is that statement:

"In existing languages, where there is no way to distinguish between multiple execution phases, it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific
execution phase."

No?

ShaiBer:
> I thought this is that statement:
> 
> "In existing languages, where there is no way to distinguish between multiple execution phases, it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific execution phase."
> 
> No?

I don't think it will be clear enough to most readers. Also, it doesn't explain why it cannot be done by other means that are not a language (there are some attempts at similar things by others, like Pulumi and we don't explain where they hit a wall that we don't )

README.md
```diff
@@ -29,38 +29,46 @@ Our mission is to bring back your creative flow and close the gap between imagin
...
-new cloud.Function(hello_world);
-```
+The `queue.add_consumer()` method is a *preflight method* that configures the infrastructure to
+invoke a particular *inflight function* for each message in the queue.
+
+Now comes the cool part: the code that runs inside the inflight function interacts with the `counter` and the `bucket` objects
+through the *inflight methods* of these objects (`counter.inc()` and `bucket.put()`). These methods can only be
+called from inflight scopes. In existing languages, where there is no way to distinguish between multiple execution phases,
+it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific
+execution phase.
```
Chriscbr, line 71: (Optional)
```suggestion
called from inflight scopes. In existing languages, where there is no way to distinguish between multiple execution phases,
it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific
execution phase in a type-safe manner.
```

or

```suggestion
called from inflight scopes. In existing languages, where there is no way to distinguish between multiple execution phases in the type system,
it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific
execution phase.
```

Chriscbr:
> Also, it doesn't explain why it cannot be done by other means that are not a language (there are some attempts at similar things by others, like Pulumi and we don't explain where they hit a wall that we don't )

Maybe we can include a nod to how execution phases as a first-class citizen/concept means your IaC code becomes more composable and reusable, or how it enables resource to infer IAM policies automatically? I could also see some of this being left for the tutorial or getting started guide

ShaiBer:
> > There is no clear statement of exactly what we can do with Wing that cannot be done with a library/compiler extension
> 
> I thought this is that statement:
> 
> "In existing languages, where there is no way to distinguish between multiple execution phases, it is impossible to naturally represent this idea that an object has methods that can only be executed from within a specific execution phase."
> 
> No?

I like it

ShaiBer (approved):
Approve, and would like to see your suggestion for making the message even more clear added later as well

Issue #2184, opened 2023-04-20
Add to FAQ: why start with JS interop and not another language


ShaiBer:
### Feature Spec

Rough draft:

Since we compile to JavaScript it is easiest to add interoperability with it. But it doesn't exclude the addition of more languages to compile and / or interoperate with Winglang in the future.

The reason we chose to start with compilation to JavaScript and not some other popular cloud language is because JavaScript is currently one of the most used ones, and in addition, we are using [JSII](https://github.com/aws/jsii) to write our SDK in Typescript and then export it for consumption in other languages, such as Go, Python, Java and others.

BTW, even though you cannot directly call non Wing or JS code from Wing, you can still use other languages inside cloud functions and then invoke the functions with Wing. You will not get the full benefits of the compiler seeing everything in this way, but it will work.

### Use Cases

Answer this common question

### Component

Documentation

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.15.21](https://github.com/winglang/wing/releases/tag/v0.15.21).

PR #2334, opened 2023-05-01
chore(docs): faq - why start with js interop

ShaiBer:
Fixes: #2184


docs/07-faq/097-why-start-with-js-interop.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Why start with Javascript interop and not some other language?
```
Chriscbr, line 2: I wonder if it makes sense to rephrase this Q&A to make it more focused on how this feature provides people value. The goal of a FAQ should be to deliver important information about how Wing solves people's problems, not to justify architectural decisions.

(I also think the current explanation is a little confusing since JavaScript interoperability followed from the choice to make Wing compatible with CDK / use CDK under the hood for producing infra config, not the other way around)

Here's an idea of what I was thinking...

> Q: What languages does Wing interoperate with?
>
> A: Wing compiles [inflight] application code into JavaScript so that your cloud apps can leverage the world's largest package ecosystem via [npm]. JavaScript has been proven as a flexible, portable language for a variety of use cases from client-side scripting to running application servers, and Wing's support for natively importing JavaScript libraries allows you to seamlessly integrate with a vast range of existing tools and frameworks. Through JavaScript's WebAssembly API, it's even possible to call into compiled from other languages in Wing.
>
> For generating infrastructure configuration (like Terraform and Cloudformation) during the [preflight] execution phase, Wing also uses JavaScript so that you can import any resource providers available through the [AWS CDK] or [CDKTF] frameworks.
>
> While Wing currently emits JavaScript, it doesn't exclude the possibility of compiling and interoperating with other languages for inflight code in the future.

ShaiBer: I agree that the FAQ should deliver important information about how Wing can solve your problems, but it should also provide answers for frequent questions. I added this topic because I was asked on several occasions why we started with JS interop by developers who already knew that we interop with JS.
I'm uncertain if someone who already knows about our JS interoperability and is curious about the reasoning behind starting with it would know they can find the answer in your version. WDYT, Should we include your version alongside a clearer version of this, or should we opt for your version instead?

ShaiBer: I got this question a bunch of times from devs who understand why we're building wing, some of them also understand why we're compiling to JS. They are asking why we chose to have extern functions in JS first, and not started with extern for any other language.
I will add some of the points you've raised, but will direct them to read others elsewhere since I don't think they fit  exactly into the scope of this question.

shaiber01 (approved):

PR #2351, opened 2023-05-02
chore(plugins): update s3-backend to allow multiple backends

hasanaburayyan:
I thought it might be interesting to extend the s3-backend plugin to support multiple tf-backends. I think its cool, you be the judge 😄 


examples/plugins/tf-backend-configuration.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * Plugin: tf-backend-configuration
```
eladb, line 2:
```suggestion
 * Plugin: tf-backend
```

eladb (changes requested):
Rename the file as well

eladb (approved):

PR #2354, opened 2023-05-02
feat: reduce package size to from ~180mb to ~15mb

### LINKED ISSUE #2035: Wing installation is not a good experience

eladb:
By not bundling the `aws-cdk-lib` dependency into the Wing SDK we can reduce the package size dramatically. I believe this fixes #2035.

```
npm notice === Tarball Details ===
npm notice name:          @winglang/sdk
npm notice version:       0.0.0
npm notice filename:      @winglang/sdk-0.0.0.tgz
npm notice package size:  14.9 MB
npm notice unpacked size: 86.1 MB
npm notice shasum:        fb7d473c7a65811d154b4eb6e5cf7eb5f7c35530
npm notice integrity:     sha512-LeTGWZrVelYSw[...]MTU4EUUYQTP3Q==
npm notice bundled deps:  226
npm notice bundled files: 0
npm notice own files:     11675
npm notice total files:   11675
npm notice
```

To implement this, we patch `package.json` just before we pack it and remove the `aws-cdk-lib` from the dependency list (yes, hack).

When compiling to the `awscdk` target, if `aws-cdk-lib` is not installed, this will be the output:

```console
$ wing compile -t awscdk hello.w
Cannot find module 'aws-cdk-lib'. The target "awscdk" requires this module to be installed.
```

### Future improvements

* Verify that the `aws-cdk-lib` version is compatible with the one we compiled with.
* Automatically install aws-cdk-lib just in time

### Misc

* Moved the `CdktfApp` outside of the `core` module so it's not publicly exported.
* Reduce the verbosity for preflight errors in the CLI so that by default only the error
  will be displayed. To increase verbosity, users can pass `--debug` (or set `DEBUG=1`).


libs/wingsdk/src/core/app.ts
```diff
@@ -171,182 +163,6 @@ export abstract class App extends Construct {
   }
 }
 
-/**
- * An app that knows how to synthesize constructs into Terraform configuration
- * using cdktf. No polycon factory or Terraform providers are included.
- */
-export abstract class CdktfApp extends App {
```
eladb, line 178: I've moved this outside of `core.App` so that we won't have CDKTF APIs leaking out of the core module. It's not directly related but technically if we want to do the same trick for `cdktf` in the future, this will be required.

Also, it does't make sense at all that this class is here :-)

Chriscbr (approved):
Looking forward to trying this out!

I wonder if in the CLI if we should install aws-cdk-lib for them, or does it make sense to let them use whichever package manager they want?

(adding `do-not-merge` so we can put something in the CLI to give a nice info message)

staycoolcall911:
Before this PR: 44 sec for installation
image

After: 31s
image

On WeWork's fast connection
image

eladb:
I expected this to be a bit better. 

But I think that's what we have for now...

PR #2358, opened 2023-05-02
chore: ensure the console builds correctly

### LINKED ISSUE #2357: Ensure the Console builds correctly during the Wing CI
Add some steps to the Wing CI to ensure that the Console builds correctly before publishing a new Wing version.

skyrpex:
Fixes #2357.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -225,28 +225,28 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: wing
-          path: /wing
+          path: wing
 
-      - run: ls -al /home/runner/work/wing/wing
-      - run: ls -al /home/runner/work/wing/wing/wing
-      - run: ls -al /wing
+      # - run: ls -al /home/runner/work/wing/wing
+      # - run: ls -al /home/runner/work/wing/wing/wing
+      - run: ls -al ${{ github.workspace }}/wing
 
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
           node-version: 18
 
-      - name: Install Dependencies
-        uses: bahmutov/npm-install@v1
-        with:
-          working-directory: |
-            tools/hangar
-            examples/tests/valid
-            examples/tests/invalid
-          install-command: npm ci
+      # - name: Install Dependencies
+      #   uses: bahmutov/npm-install@v1
+      #   with:
+      #     working-directory: |
+      #       tools/hangar
+      #       examples/tests/valid
+      #       examples/tests/invalid
+      #     install-command: npm ci
 
       - name: Link Wing CLI
-        run: ln -s /wing/bin/wing /usr/local/bin/wing
+        run: ln -s ${{ github.workspace }}/bin/wing /usr/local/bin/wing
 
       - name: Check Wing Version
         run: wing --version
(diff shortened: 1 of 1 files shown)


skyrpex:
I'm getting this wing compilation error in the console:

image

Help would be appreciated :) @winglang/maintainers

staycoolcall911:
Happy to help @skyrpex. Let's move to #dev on slack. Also, can you share what code you were trying to run?

skyrpex:
The job works fine now:
- The build job uploads the dist files from the Wing CLI and the Wing SDK
- A new job downloads the dist files above, links them to the Console and builds it

It currently fails because the Console is not compatible with the latest Wing versions.

@ainvoner what are the next steps?

ainvoner:
@skyrpex we should provide a way (label) to not fail the wing build workflow even if the console breaks.
there will be situations were a breaking change must be introduced and there is no way to fix the console before the new wing version is published.

.github/workflows/build.yml
```diff
@@ -72,6 +72,16 @@ jobs:
...
 
+      - name: Upload Dist Files
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: |
+            apps/wing/dist
+            apps/wing/*.wasm
+            libs/wingsdk/lib
+            libs/wingsdk/.jsii
```
MarkMcCulloh, line 83: These files are already updated in the next steps, (some are technically in tgz files)

.github/workflows/build.yml
```diff
@@ -212,13 +222,74 @@ jobs:
           npm run test:generate
           npm run test --shard=${{ matrix.shard }}
 
+  console:
```
MarkMcCulloh, line 225: I think this flow could be optimized to use the tgz files directly:

1. Clone Console, setup node and pnmp
1. Download the wing and wingsdk tgz artifacts
1. pnpm link both tgz packages

.github/workflows/build.yml
```diff
@@ -212,13 +222,74 @@ jobs:
...
+      - name: Download Dist Files
+        uses: actions/download-artifact@v3
+        with:
+          name: dist
+
+      - name: Link Wing CLI
+        run: npm i -g ./apps/wing
+
+      - name: Check Wing Version
+        run: wing --version
```
MarkMcCulloh, line 251: Does the Console use the global version of the cli (when running the tests in `pnpm run build`)? Could it use the linked one instead?

skyrpex:
@MarkMcCulloh regarding your suggestions, I'm not sure if it's worth the hassle just to emulate real npm packaging when what we're really looking for is to check for console compatibility issues in advance. In my opinion, ensuring that the wing and wingsdk packages are correctly packaged, we should wait until you implement the local npm repository instead.

In order to use the current packaged artifacts, I'd need to:
- Download the artifacts of wing and wingsdk
- Untar the contents into separate directories
- Manipulate the package.json of both packages to remove devDependencies and/or link wing to the local wingsdk package (they reference non-existing wing and wingsdk versions)
- Run pnpm install on both of them since they don't come with all of the node_modules files
- Link those packages to the console packages

Whereas with the current method I only need to download the artifacts and link the packages to the console.

I'll go ahead if you're strongly opinionated about this. I just want a final confirmation before diving in.

PR #2391, opened 2023-05-03
fix(compiler): class without an initializer

### LINKED ISSUE #1705: Unable to declare a class/resource without an initializer

marciocadev:
This PR is similar to PR #1861.

This PR is similar to PR #1861. I added new tests and removed all unnecessary constructors from the classes of the old tests

Closes #1705


examples/tests/invalid/resource_captures.w
```diff
@@ -19,10 +18,6 @@ class Foo {
     log(this.mut_array.at(0));
 //        ^^^^^^^^^ Unable to reference "this.mut_array" from inflight method "test" because type MutArray<str> is not capturable
 
-    let b = this.bucket;
-//          ^^^^^^^^^^^ Unable to qualify which operations are performed on 'this.bucket' of type 'Bucket'. This is not supported yet.
-    b.put("hello", "world");
```
Chriscbr, line 24: I'm curious why this had to be removed

libs/wingc/src/parser.rs
```diff
@@ -625,7 +643,7 @@ impl<'s> Parser<'s> {
 			methods,
 			parent,
 			implements,
-			initializer: initializer.unwrap(),
+			initializer: initializer,
```
Chriscbr, line 646:
```suggestion
			initializer,
```

libs/wingc/src/type_check.rs
```diff
@@ -2483,6 +2486,79 @@ impl<'a> TypeChecker<'a> {
...
+						fields.append(&mut self.list_fields_on_init(&finally_statements));
+					}
+				}
+				_ => (),
+			}
+		}
+		return fields;
+	}
+
+	fn check_class_field_initialization(&mut self, statements: &Scope, fields: &Vec<ClassField>) {
```
Chriscbr, line 2545: nit
```suggestion
	fn check_class_field_initialization(&mut self, statements: &Scope, fields: &[ClassField]) {
```

Chriscbr, line 2545: Add a docstring for this method

libs/wingc/src/type_check.rs
```diff
@@ -2483,6 +2486,79 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	fn list_fields_on_init(&mut self, statements: &Scope) -> Vec<String> {
```
Chriscbr, line 2489: Add a docstring

libs/wingc/src/type_check.rs
```diff
@@ -2483,6 +2486,79 @@ impl<'a> TypeChecker<'a> {
...
+			if field.phase == Phase::Inflight || field.is_static {
+				continue;
+			}
+			if !init_fields.contains(&field.name.name) {
+				self.type_error(TypeError {
+					message: format!("\"{}\" is not initialized", field.name.name),
+					span: field.name.span.clone(),
+				});
+			} else {
+			}
```
Chriscbr, line 2558:
```suggestion
			}
```

libs/wingc/src/type_check.rs
```diff
@@ -2483,6 +2486,79 @@ impl<'a> TypeChecker<'a> {
...
+	fn list_fields_on_init(&mut self, statements: &Scope) -> Vec<String> {
+		let mut fields: Vec<String> = Vec::new();
+		for stmt in &statements.statements {
+			match &stmt.kind {
+				StmtKind::Assignment { variable, value: _ } => match &variable {
+					Reference::InstanceMember { property, object: _ } => {
+						fields.push(property.name.clone());
+					}
+					_ => (),
+				},
```
Chriscbr, line 2498: (optional) Another way to do this would be to create a visitor using the trait in `visit.rs`. Doing so might let you avoid having to write all of the cases for `If`/`While`/`ForLoop`/`TryCatch` since it will automatically do the recursion for those cases by default.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -144,6 +144,30 @@ exports[`capture_reassignable.w > stderr 1`] = `
 "
 `;
 
+exports[`class.w > stderr 1`] = `
+"error: \\"x\\" is not initialized
+  --> ../../../examples/tests/invalid/class.w:2:4
+  |
+2 |    x:num;
+  |    ^ \\"x\\" is not initialized
```
Chriscbr, line 152: I like how the error message is next to where the field is defined 👍

Chriscbr (commented):
Thanks for your patience @marciocadev, I know you picked up this issue quite a while back. :) I added a few comments and questions but the high level approach looks good to me

examples/tests/invalid/resource_captures.w
marciocadev: Because the bucket was not initialized, than the fail from initialization was blocking the other errors

tools/hangar/__snapshots__/invalid.ts.snap
marciocadev: Me too 😀

examples/tests/invalid/resource_captures.w
Chriscbr: Can we initialize the bucket in the `init()` section then? I think the error message here ("Unable to qualify which operations...") should still apply if the bucket is initialized

marciocadev: You're right, changed

libs/wingc/src/type_check.rs
marciocadev: Cool, I will try this

libs/wingc/src/type_check.rs
```diff
@@ -28,6 +29,41 @@ use wingii::type_system::TypeSystem;
...
+		Self { fields }
+	}
+}
+
+impl<'ast> Visit<'ast> for VisitClassInit<'ast> {
+	fn visit_scope(&mut self, node: &'ast Scope) {
+		for stmt in &node.statements {
+			visit::visit_stmt(self, stmt);
+		}
+	}
```
Chriscbr, line 48: I think you can remove this method and it will work the same, since this is the same as the method's default implementation [here](https://github.com/winglang/wing/blob/3e5534cc27d7d29f3d592025e19156a9984bcab6/libs/wingc/src/visit.rs#L78-L85)

Chriscbr (approved):
Nicely done!

libs/wingc/src/type_check.rs
marciocadev: It works, I can't believe it works

Chriscbr (approved):

eladb:
Hell yeah!

PR #2393, opened 2023-05-03
fix(compiler): panic when calling string methods inflight

### LINKED ISSUE #2390: Panic when calling methods on global string
### I tried this:

```js
bring cloud;

let s1 = "hello";
let s2 = " world";
class A {
  init() {}
  inflight foo() {
    log(s1.concat(s2));
  }
}

let r = new A();
```

### This happened:

```
thread '<unnamed>' panicked at 'unexpected - all components should have a variable at this point', libs/wingc/src/jsify.rs:1443:17
stack backtrace:
RuntimeError: unreachable
    at std::panicking::rust_panic_with_hook::h911ba172265603df (wasm://wasm/009a3f06:wasm-function[220]:0xca683)
    at core::panicking::panic_fmt::h6958314cbbe045b2 (wasm://wasm/009a3f06:wasm-function[1438]:0x172578)
    at <wingc::jsify::FieldReferenceVisitor as wingc::visit::Visit>::visit_expr::hccd000dcdd336ea9 (wasm://wasm/009a3f06:wasm-function[58]:0x7206a)
    at <wingc::jsify::FieldReferenceVisitor as wingc::visit::Visit>::visit_expr::hccd000dcdd336ea9 (wasm://wasm/009a3f06:wasm-function[58]:0x72bac)
    at <wingc::jsify::FieldReferenceVisitor as wingc::visit::Visit>::visit_expr::hccd000dcdd336ea9 (wasm://wasm/009a3f06:wasm-function[58]:0x72bd2)
    at wingc::visit::visit_stmt::hc47e69e6d9fcb939 (wasm://wasm/009a3f06:wasm-function[212]:0xc8320)
    at wingc::jsify::JSifier::jsify_statement::h217f18b8ad6ec044 (wasm://wasm/009a3f06:wasm-function[17]:0x8ffd)
    at wingc_compile.command_export (wasm://wasm/009a3f06:wasm-function[1583]:0x179510)
    at Object.invoke (/Users/chrisr/dev/wing-test/node_modules/winglang/dist/wingc.js:138:37)
    at compile (/Users/chrisr/dev/wing-test/node_modules/winglang/dist/commands/compile.js:135:38)
```

### I expected this:

It compiles

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

0.15.7

### Wing Console Version

_No response_

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

Chriscbr:
Fixes #2390

There's definitely something funky with this code (it seems to be duplicating some of the logic in the type checker's `resolve_reference` function) but it's hard to say what the best way to fix it is - for now, skipping analysis of methods on built-in types seems like an OK patch for now. Let's see if this solution holds water


libs/wingc/src/jsify.rs
```diff
@@ -1602,34 +1602,57 @@ impl<'a> FieldReferenceVisitor<'a> {
 
 			ExprKind::Reference(Reference::InstanceMember { object, property }) => {
 				let obj_type = object.evaluated_type.borrow().unwrap();
-				let var = if let Some(cls) = obj_type.as_class_or_resource() {
-					Some(
+				let var = match &*obj_type {
+					Type::Void => unreachable!("cannot reference a member of void"),
+					Type::Function(_) => unreachable!("cannot reference a member of a function"),
+					Type::Optional(_) => unreachable!("cannot reference a member of an optional"),
```
MarkMcCulloh, line 1608: Could we avoid these panics? Maybe the void one doesn't matter, but the other two seem like things that a user could type (lmk if I'm wrong) and we probably shouldn't crash the compiler for that.

libs/wingc/src/jsify.rs
```diff
@@ -1602,34 +1602,57 @@ impl<'a> FieldReferenceVisitor<'a> {
...
+					// all fields / methods / values of these types are phase-independent so we can skip them
+					Type::Anything => return vec![],
+					Type::Number => return vec![],
+					Type::String => return vec![],
+					Type::Duration => return vec![],
+					Type::Boolean => return vec![],
+					Type::Json => return vec![],
+					Type::MutJson => return vec![],
+					Type::Enum(_) => return vec![],
```
MarkMcCulloh, line 1617: nit: use `|` for slightly less code


(probably need to run fmt on this)
```suggestion
					Type::Anything |
					Type::Number |
					Type::String |
					Type::Duration |
					Type::Boolean |
					Type::Json => |
					Type::MutJson |
					Type::Enum(_) => return vec![],
```

libs/wingc/src/jsify.rs
Chriscbr: I don't think these are lines are reachable, at least from my time playing around on the branch a bit. Unless we can produce a case where these paths are possible, I don't think we should add a diagnostic since we wouldn't be able to test it

MarkMcCulloh: Yeah I was thinking about the case of an LSP where incomplete input could possibly lead here, but even if that were true we don't jsify if there's an error in the previous phase (or at all, in the LSP).

MarkMcCulloh (approved):

PR #2403, opened 2023-05-04
fix(docs): disable js minimization in docusaurus to display code samples properly

ShaiBer:

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -8,14 +8,21 @@ module.exports = {
   themeConfig: {
     // ...
   },
-  configureWebpack: {
-    module: {
-      rules: [
-        {
-          test: /\.(w|js|ts|tf)$/,
-          use: 'file-loader',
-        },
-      ],
-    },
+  configureWebpack: (config, isServer, utils) => {
+    const { getCacheLoader } = utils;
+    return {
+      module: {
+        rules: [
+          {
+            test: /\.(w|js|ts|tf)$/,
+            use: 'file-loader',
+          },
+        ],
+      },
+      optimization: {
+        ...config.optimization,
+        minimize: false, // Disables JavaScript minification
+      },
+    };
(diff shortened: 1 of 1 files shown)


staycoolcall911 (approved):

docusaurus.config.js
MarkMcCulloh, line 1: How is this config file used? the actual config is [here](https://github.com/winglang/docsite/blob/main/docusaurus.config.js) no?

MarkMcCulloh (changes requested):

docusaurus.config.js
```diff
@@ -8,14 +8,21 @@ module.exports = {
...
-      ],
-    },
+  configureWebpack: (config, isServer, utils) => {
+    const { getCacheLoader } = utils;
+    return {
+      module: {
+        rules: [
+          {
+            test: /\.(w|js|ts|tf)$/,
+            use: 'file-loader',
```
MarkMcCulloh, line 18: I believe webpack 5 does not need this anymore, you can use [asset/resource](https://webpack.js.org/guides/asset-modules/)

MarkMcCulloh: note this section as well https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

docusaurus.config.js
ShaiBer: That's a good question.. I assumed they somehow get merged.
The reason I didn't use the one in the docsite is that I wanted to test these things locally and when I work locally I don't even have the docsite repo.
Is there some other way to solve this?

MarkMcCulloh: afaik there is no current process for this file to be used in the build process.

Not 100% sure of the best approach right now, I think the workflow deserves some thought.

ShaiBer: Yes.. definitely

Issue #390, opened 2022-10-31
Function Declaration: call inflight closure from inflight code


staycoolcall911:
Our current compiler implementation wraps each inflight declaration in a file of its own. This poses a limitation and does not allow basic code reuse using functions.

Consider this simplistic code for example:
`wing test`
```ts
bring cloud;

let bucket = new cloud.Bucket();

let store_in_bucket = inflight (event: str, file: str): str => {
    bucket.put(file, event);
};

let handler1 = inflight (event: str): str => {
	store_in_bucket(event, "file1");
};     

let func1 = new cloud.Function(handler1) as "func1";

new cloud.Function(inflight () => {
    func1.invoke("hi1");
}) as "test";
```
Error message:
```sh
ReferenceError: store_in_bucket is not defined
     │     at Handler.handle (evalmachine.<anonymous>:20:3)
     │     at exports.handler (evalmachine.<anonymous>:25:5)
     │     at evalmachine.<anonymous>:27:9
     │     at Script.runInContext (node:vm:141:12)
     │     at Object.runInContext (node:vm:291:6)
     │     at /Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:133:12
     │     at new Promise (<anonymous>)
     │     at runInSandbox (/Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:114:12)
     │     at Object.activity (/Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/function.inflight.js:55:24)
     └     at Object.withTrace (/Users/urib/.nvm/versions/node/v19.3.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:82:50)
```

This code won't work currently, because the call to `store_in_bucket` inflight function is not recognized properly by the compiler.
A possible solution may be to change all inflight to be in the same file, each with a different entry.

staycoolcall911:
Should be resolved as part of #542

staycoolcall911:
@yoav-steinberg - #1448 is a duplicate of that right? It is actually defining how we are going to implement this issue if I understand correctly

yoav-steinberg:
[...]

PR #2411, opened 2023-05-04
chore(compiler): transform inflight closures into preflight classes

### LINKED ISSUE #1448: existing closure mechanism is causing permissive permissions

Chriscbr:
Fixes #1448

Adds some code in the compiler that inflight closures in the AST into resources. The main goal of this is to unify how we are analyzing resources and inflight closures so that they are treated the same for the purposes of emitting diagnostic errors and for code generation. It should also get us closer to addressing #390.

Highlights:
- Introduces a new "inflight transformer" phase to the compiler, which runs right after the initial parse tree is created, and before type checking. This is only possible because the transform is purely syntactical - it doesn't require any type information.
  - Side note: in earlier iterations of this PR I had this run after type checking, and it pretty much behaves the same, but it required re-type-checking everything the entire tree
- Removes `capture.rs` compiler phase from the compiler, as it was only being used for analyzing inflight closures in the "old way" (which granted all all permissions to all resources it saw). Now, the capture analysis code we have in jsify.rs (which only works on preflight class AST nodes) is used for everything (though some day we might move this into a separate compiler phase...)
- Removes the `core.Inflight` class from the SDK, since we don't need to code-generate it anymore.

Meta comments: there will probably be some regressions lurking here by virtue of behavior that isn't caught by our test suite. I'm also not sure this is the end-game solution for handling inflight closures in the compiler... but it seems like a step in the right direction.


tools/hangar/__snapshots__/test_corpus/bucket_events.w_compile_tf-aws.md
```diff
@@ -130,7 +237,7 @@
             "uniqueId": "root_b_boncreateOnMessage8588493f_IamRolePolicy_369E4FFE"
           }
         },
-        "policy": "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Action\":[\"s3:PutObject*\",\"s3:Abort*\"],\"Resource\":[\"${aws_s3_bucket.root_other_26932ECB.arn}\",\"${aws_s3_bucket.root_other_26932ECB.arn}/*\"],\"Effect\":\"Allow\"},{\"Action\":[\"s3:GetObject*\",\"s3:GetBucket*\",\"s3:List*\"],\"Resource\":[\"${aws_s3_bucket.root_other_26932ECB.arn}\",\"${aws_s3_bucket.root_other_26932ECB.arn}/*\"],\"Effect\":\"Allow\"},{\"Action\":[\"s3:DeleteObject*\",\"s3:DeleteObjectVersion*\",\"s3:PutLifecycleConfiguration*\"],\"Resource\":[\"${aws_s3_bucket.root_other_26932ECB.arn}\",\"${aws_s3_bucket.root_other_26932ECB.arn}/*\"],\"Effect\":\"Allow\"}]}",
```
Chriscbr, line 133: I believe most of these removed policies in snapshots are a result of no longer generating permissions for bucket operations that aren't being called (like `bucket.delete`).

examples/tests/valid/capture_containers_of_resources.w
Chriscbr, line 1: I removed this test because we don't actually support capturing collections of resources right now, the fact that this compiled was just a bug

examples/tests/valid/while_loop_await.w
```diff
@@ -2,12 +2,11 @@ bring cloud;
...
 
-let iterator = inflight (j: num): num => {
-    return j+1;
-};
-
 let handler = inflight (body: str): str => {
     let i = 0;
+    let iterator = inflight (j: num): num => {
+        return j+1;
+    };
```
Chriscbr, line 9: Need to investigate this

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+							span: expr.span.clone(),
+						},
+						fields: resource_fields,
+						implements: vec![],
+						parent: None,
+						methods: vec![(handle_name.clone(), new_func_def)],
+						inflight_initializer: None,
+					}),
+					idx: 1,
+					span: expr.span.clone(),
```
Chriscbr, line 162: There's an interesting implementation question here of, "ok, if we're introducing random new AST nodes to the tree (that could potentially get type checked and so on), what `span` do we assign them?" Should it just be a "global" span?

Here I'm trying the approach of "let's just inherit the span of the nearest parent AST node that we're adding ourselves to", which happens to be the span of the original inflight closure. I'm not sure if this is correct, but it seemed like a decent option for making sure any unexpected errors are localized near their source.

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+/// let f = ((): core.Resource => {
+///   class $Inflight1 {
+///     init() {}
+///     inflight handle(message: str) {
+///       b.put("file.txt", message);
+///     }
+///   }
+///   return new $Inflight1();
+/// })();
+pub struct InflightTransformer {
```
eladb, line 38:
```suggestion
pub struct ClosureTransformer {
```


The term "inflight" can also refer to classes and objects.

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+	// Whether the transformer is inside a preflight or inflight scope.
+	// Only inflight closures defined in preflight scopes need to be transformed.
```
eladb, line 40: I'd argue that maybe that's not necessarily going to be true in the long run but we'll discuss later.

libs/wingc/src/inflight_transform.rs
eladb: I think inheriting the original span makes sense to me

libs/wingc/src/lib.rs
```diff
@@ -281,10 +282,15 @@ pub fn compile(
 	let default_out_dir = PathBuf::from(format!("{}.out", file_name));
 	let out_dir = out_dir.unwrap_or(default_out_dir.as_ref());
 
+	// Build our AST
+	let (scope, parse_diagnostics) = parse(&source_path);
+
+	// Transform all inflight closures defined in preflight into single-method resources
```
eladb, line 288: This is our "desugaring" phase.

I wouldn't over design this right now but in the future there could be more that we would want to put.

Maybe just add a header comment that says this is where desugaring happens

libs/wingsdk/src/core/inflight.ts
```diff
@@ -86,59 +82,6 @@ export interface InflightProps {
...
-  /** @internal */
-  public _bind(_host: IInflightHost, _ops: string[]): void {
-    throw new Error("Method not implemented.");
-  }
-  /** @internal */
-  public _registerBind(_host: IInflightHost, _ops: string[]): void {
-    throw new Error("Method not implemented.");
-  }
-  /** @internal */
-  public _toInflight(): Code {
```
eladb, line 129: 🙌 yessss

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -97,49 +97,42 @@ error: Cannot find module \\"foobar\\" in source directory: Unable to load \\"fo
...
+   |          ^ Cannot capture field 'm' with non-capturable type 'MutMap<bool>'
 
 
-error: Cannot reference 'a_cloned' of type 'MutArray<str>' from an inflight context
+error: Cannot capture field 'a_cloned' with non-capturable type 'MutArray<str>'
    --> ../../../examples/tests/invalid/capture_mutables.w:14:10
    |
-14 |   assert(a_cloned.size == 1);
-   |          ^^^^^^^^ Cannot reference 'a_cloned' of type 'MutArray<str>' from an inflight context
+14 |   assert(a_cloned.length == 1);
```
eladb, line 124: Was this intentional (rename of size to length)?

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -178,17 +171,6 @@ exports[`cloud_function_expects_inflight.w > stderr 1`] = `
 5 | | });
   | \\\\-^ Expected type to be \\"IFunctionHandler (inflight (str): void)\\", but got \\"preflight (str): str\\" instead
 
-
-error: Expected type to be \\"IQueueAddConsumerHandler (inflight (str): void)\\", but got \\"inflight (num): void\\" instead
```
eladb, line 182: Regression?

tools/hangar/__snapshots__/plugins.ts.snap
```diff
@@ -613,7 +613,7 @@ exports[`Plugin examples > AWS target plugins > tf-s3-backend.js 1`] = `
             "uniqueId": "root_cloudQueueAddConsumere46e5cb7_IamRolePolicy_756548A7",
           },
         },
-        "policy": "{\\"Version\\":\\"2012-10-17\\",\\"Statement\\":[{\\"Action\\":[\\"sqs:ReceiveMessage\\",\\"sqs:ChangeMessageVisibility\\",\\"sqs:GetQueueUrl\\",\\"sqs:DeleteMessage\\",\\"sqs:GetQueueAttributes\\"],\\"Resource\\":[\\"\${aws_sqs_queue.root_cloudQueue_E3597F7A.arn}\\"],\\"Effect\\":\\"Allow\\"},{\\"Action\\":[\\"s3:PutObject*\\",\\"s3:Abort*\\"],\\"Resource\\":[\\"\${aws_s3_bucket.root_cloudBucket_4F3C4F53.arn}\\",\\"\${aws_s3_bucket.root_cloudBucket_4F3C4F53.arn}/*\\"],\\"Effect\\":\\"Allow\\"},{\\"Action\\":[\\"s3:GetObject*\\",\\"s3:GetBucket*\\",\\"s3:List*\\"],\\"Resource\\":[\\"\${aws_s3_bucket.root_cloudBucket_4F3C4F53.arn}\\",\\"\${aws_s3_bucket.root_cloudBucket_4F3C4F53.arn}/*\\"],\\"Effect\\":\\"Allow\\"},{\\"Action\\":[\\"s3:DeleteObject*\\",\\"s3:DeleteObjectVersion*\\",\\"s3:PutLifecycleConfiguration*\\"],\\"Resource\\":[\\"\${aws_s3_bucket.root_cloudBucket_4F3C4F53.arn}\\",\\"\${aws_s3_bucket.root_cloudBucket_4F3C4F53.arn}/*\\"],\\"Effect\\":\\"Allow\\"}]}",
```
eladb, line 616: Nice!

tools/hangar/__snapshots__/test_corpus/redis.w_compile_tf-aws.md
```diff
@@ -1,5 +1,28 @@
 # [redis.w](../../../../examples/tests/valid/redis.w) | compile | tf-aws
 
+## clients/$Inflight1.inflight.js
+```js
+module.exports = function({ r, r2 }) {
+  class  $Inflight1 {
+    constructor({  }) {
```
eladb, line 7: Would be nice to omit this ctor at some point

eladb (approved):

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+				let handle_name = Symbol {
+					name: "handle".to_string(),
+					span: expr.span.clone(),
+				};
+				let parent_this_name = Symbol {
+					name: PARENT_THIS_NAME.to_string(),
+					span: expr.span.clone(),
+				};
+				let this_name = Symbol {
+					name: "this".to_string(),
```
eladb, line 110: Might be nice to move all these strings to a central constants file at some point

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+						root: resource_name.clone(),
+						fields: vec![],
+					}),
+					span: expr.span.clone(),
+				};
+
+				let resource_fields: Vec<ClassField> = vec![];
+				let resource_init_params: Vec<FunctionParameter> = vec![];
+
+				let new_func_def = if self.is_inside_resource {
```
eladb, line 125: Add a comment explaining this use case. 


Also- what happens if we have multiple nesting? Parent_parent_this

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+					// transform all occurrences of "this" into "__parent_this"
+					let mut this_transform = ThisTransformer;
+					this_transform.fold_function_definition(func_def)
+				} else {
+					func_def
+				};
+
+				// resource_def :=
+				// ```
+				// resource {
```
eladb, line 135:
```suggestion
				// preflight class {
```

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+				let parent_this_name = Symbol {
+					name: PARENT_THIS_NAME.to_string(),
+					span: expr.span.clone(),
+				};
+				let this_name = Symbol {
+					name: "this".to_string(),
+					span: expr.span.clone(),
+				};
+
+				let resource_type_annotation = TypeAnnotation {
```
eladb, line 114:
```suggestion
				let class_type = TypeAnnotation {
```


Why do we call this TypeAnnotation??

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+				// ```
+				let return_resource_instance = Stmt {
+					kind: StmtKind::Return(Some(Expr::new(
+						ExprKind::New {
+							class: resource_type_annotation,
+							arg_list: ArgList {
+								named_args: IndexMap::new(),
+								pos_args: vec![],
+							},
+							obj_id: None,
```
eladb, line 177: Does this imply that the defaults will be applied?

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+				//   <return_resource_instance>
+				// }
+				// ```
+				let make_resource_body = if self.is_inside_resource {
+					// parent_this_def :=
+					// ```
+					// let __parent_this = this;
+					// ```
+					let parent_this_def = Stmt {
+						kind: StmtKind::VariableDef {
```
eladb, line 200:
```suggestion
						kind: StmtKind::Let {
```

We should rename this at some point

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+							reassignable: false,
+							var_name: parent_this_name,
+							initial_value: Expr::new(ExprKind::Reference(Reference::Identifier(this_name)), expr.span.clone()),
+							type_: None, // thank god we have type inference
```
eladb, line 204: 🤓

examples/tests/valid/while_loop_await.w
```diff
@@ -2,12 +2,11 @@ bring cloud;
 
 let queue = new cloud.Queue();
 
-let iterator = inflight (j: num): num => {
-    return j+1;
-};
-
 let handler = inflight (body: str): str => {
     let i = 0;
+    let iterator = inflight (j: num): num => {
```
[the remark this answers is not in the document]
eladb: https://github.com/winglang/wing/issues/2413

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+		let prev_phase = self.curr_phase;
+		self.curr_phase = node.signature.phase;
+		let new_node = fold::fold_function_definition(self, node);
+		self.curr_phase = prev_phase;
+		new_node
+	}
+
+	fn fold_class(&mut self, node: Class) -> Class {
+		let prev_is_inside_resource = self.is_inside_resource;
+		self.is_inside_resource = true;
```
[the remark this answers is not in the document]
Chriscbr: Let's see if more usages of this pattern comes up

libs/wingc/src/inflight_transform.rs
Chriscbr: Good question...

libs/wingc/src/inflight_transform.rs
Chriscbr: This is to help us distinguish them from the `Type` structs we generate and store during type checking. For example, when an expression node is type checked, its `evaluated_type` is assigned a value with the type it has.

Part of the reason this is needed in the current compiler architecture is because we need a way to represent code like `let x: foo.Bar = // snip` before we have started resolving symbols or doing type checking (steps we currently do together).

libs/wingc/src/inflight_transform.rs
Chriscbr: Right now yes

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+					};
+
+					Scope::new(
+						vec![parent_this_def, resource_def, return_resource_instance],
+						expr.span.clone(),
+					)
+				} else {
+					Scope::new(vec![resource_def, return_resource_instance], expr.span.clone())
+				};
+
```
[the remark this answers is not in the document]
Chriscbr: For sure. Macros might help with this in the future

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+	}
+}
+
+/// Transform usages of "this.<field>" into "__parent_this.<field>"
+struct ThisTransformer;
+
+impl Fold for ThisTransformer {
+	fn fold_reference(&mut self, node: Reference) -> Reference {
+		match node {
+			Reference::Identifier(ident) => {
```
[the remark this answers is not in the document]
Chriscbr: See my other comment about making it possible to code generate the `Fold` boilerplate (future work)

I think the clean way to achieve this would be to have the `Reference::Identifier` enum contain `Identifier` (a new struct that wraps a string) instead of `String`. like so:

```rs
pub enum Reference {
	Identifier(Identifier),
	InstanceMember(InstanceMember),
	TypeMember(TypeMember),
}

pub struct Identifier(String);
pub struct InstanceMember(/* snip */);
```

examples/tests/valid/while_loop_await.w
Chriscbr: This isn't really a great test, I'm removing it for now

examples/tests/valid/capture_containers_of_resources.w
```diff
@@ -1,25 +0,0 @@
-bring cloud;
-
-let arr = [
-  new cloud.Bucket() as "b1", 
-  new cloud.Bucket() as "b2"
-];
-
-let map = { "my_queue": new cloud.Queue() };
```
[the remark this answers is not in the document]
Chriscbr: We don't support capturing collections of resources right now

libs/wingc/src/inflight_transform.rs
```diff
@@ -0,0 +1,277 @@
...
+						vec![parent_this_def, resource_def, return_resource_instance],
+						expr.span.clone(),
+					)
+				} else {
+					Scope::new(vec![resource_def, return_resource_instance], expr.span.clone())
+				};
+
+				// make_resource_closure :=
+				// ```
+				// (): resource => { ...<make_resource_body> }
```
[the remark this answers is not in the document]
Chriscbr: I believe so

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: Fixed

libs/wingc/src/closure_transform.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+impl Fold for ClosureTransformer {
+	fn fold_scope(&mut self, node: Scope) -> Scope {
+		let mut statements = vec![];
+
+		// If we are inside a scope with "this", add define `let __parent_this = this` which can be
+		// used by the newly-created preflight classes
+		if self.inside_scope_with_this {
+			let parent_this_name = Symbol::global(PARENT_THIS_NAME); // TODO: can we use a span?
```
yoav-steinberg, line 73: Not that it solves the span thing but I recently added added a `From` trait to `Symbol` so this should be equivalent:
```suggestion
			let parent_this_name = PARENT_THIS_NAME.into(); // TODO: can we use a span?
```

yoav-steinberg, line 73: I agree that we should find a way to assign a span to everything here. We can can use the `Scope`'s span if there's nothing better.

libs/wingc/src/closure_transform.rs
```diff
@@ -0,0 +1,306 @@
...
+				.collect(),
+			methods: node
+				.methods
+				.into_iter()
+				.map(|(name, def)| {
+					let new_symbol = self.fold_symbol(name);
+
+					let new_def = if !def.is_static {
+						let prev_inside_scope_with_this = self.inside_scope_with_this;
+						self.inside_scope_with_this = true;
```
yoav-steinberg, line 141: I have a feeling we can move this logic into `fold_function_definition` which might be cleaner.
Same for `fold_constructor`.
If we do this, we might be able to keep the default implementation of `fold_class`.

libs/wingc/src/closure_transform.rs
```diff
@@ -0,0 +1,306 @@
...
+
+				let new_func_def = if self.inside_scope_with_this {
+					// If we are inside a class, we transform inflight closures with an extra
+					// `let __parent_this = this;` statement before the class definition, and replace all references
+					// of `this` with `__parent_this` so that they can access the parent class's fields.
+					let mut this_transform = RenameIdentifierTransformer {
+						from: "this",
+						to: PARENT_THIS_NAME,
+					};
+					this_transform.fold_function_definition(func_def)
```
yoav-steinberg, line 220: Just wondering if this handles the case of another inner class with it's own this:
```wing
class PreflightClass {
  preflight_method() {
    let inflight_colosure = () => {
      this.b.put("k","v");  // Here `this` is the parent class instance
      class InflightClass() {
        inflight method() {
          log(this.field); // Here `this` is the inflight class instance
        }
      }
    }
  }
}

File: libs/wingc/src/closure_transform.rs [hunk 28]
```diff
@@ -0,0 +1,306 @@
...
+							named_args: IndexMap::new(),
+							pos_args: vec![],
+						},
+						obj_id: None,
+						obj_scope: None,
+					},
+					expr.span.clone(),
+				);
+
+				self.class_statements.push(class_def);
```
yoav-steinberg, line 274: is `class_statements` ever more than a single statement? If not it might be clearer to get rid of this vec.

libs/wingc/src/closure_transform.rs
```diff
@@ -0,0 +1,306 @@
...
+					Reference::Identifier(Symbol {
+						name: self.to.to_string(),
+						span: ident.span,
+					})
+				} else {
+					Reference::Identifier(ident)
+				}
+			}
+			Reference::InstanceMember { .. } => fold::fold_reference(self, node),
+			Reference::TypeMember { .. } => fold::fold_reference(self, node),
```
yoav-steinberg, line 303:
```suggestion
			_ => fold::fold_reference(self, node),
```

libs/wingc/src/jsify.rs
```diff
@@ -561,14 +557,17 @@ impl<'a> JSifier<'a> {
 				}
 			}
 			ExprKind::FunctionClosure(func_def) => match func_def.signature.phase {
-				Phase::Inflight => self.jsify_inflight_function(func_def, context).to_string(),
+				Phase::Inflight => {
+					assert!(ctx.phase == Phase::Inflight, "inflight closures should have been transformed into resources by now");
```
yoav-steinberg, line 561:
```suggestion
					assert!(ctx.phase == Phase::Inflight, "inflight closures should have been transformed into preflight handler classes in the closure_transform compiler pahse");
```

libs/wingc/src/jsify.rs
```diff
@@ -1041,6 +963,15 @@ impl<'a> JSifier<'a> {
 		for (method_name, method_refs) in refs {
 			bind_method.open(format!("if (ops.includes(\"{method_name}\")) {{"));
 			for (field, ops) in method_refs {
+				// field might be a dotted accessor, like foo.bar.baz, in which case
+				// we first bind to `foo` as an op
+				if field.split(".").count() > 1 && field.split(".").next().unwrap() != "this" {
+					let first_part = field.split(".").next().unwrap();
```
yoav-steinberg, line 969: performance nit: we should call `field.split(".").next()` only once before the `if` (we call it 3 times here). Also consider using `splitn(1,".")` instead of `split`.

libs/wingc/src/jsify.rs
```diff
@@ -1041,6 +963,15 @@ impl<'a> JSifier<'a> {
 		for (method_name, method_refs) in refs {
 			bind_method.open(format!("if (ops.includes(\"{method_name}\")) {{"));
 			for (field, ops) in method_refs {
+				// field might be a dotted accessor, like foo.bar.baz, in which case
```
yoav-steinberg, line 966: Can you explain this fix? Is it related to this PR or something else? Can you give an example?

libs/wingc/src/lsp/sync.rs
```diff
@@ -107,14 +105,9 @@ fn partial_compile(source_file: &str, text: &[u8], jsii_types: &mut TypeSystem)
 
 	let type_diag = type_check(&mut scope, &mut types, &Path::new(source_file), jsii_types);
 
-	// Analyze inflight captures
-	let mut capture_visitor = CaptureVisitor::new();
-	capture_visitor.visit_scope(&scope);
```
yoav-steinberg, line 112: I guess we're missing some diagnostics now here now? Are we? Maybe open an issue and run this by @MarkMcCulloh?

libs/wingc/src/type_check.rs
```diff
@@ -450,12 +450,60 @@ impl Subtype for Type {
...
 					parent_type.is_subtype_of(other)
-				})
+				});
+
+				if implements_iface {
+					return true;
+				}
+
+				// To support flexible inflight closures, we say that any
+				// resource with an inflight method named "handle" is a subtype of
```
yoav-steinberg, line 465:
```suggestion
				// preflight class with an inflight method named "handle" is a subtype of
```

libs/wingc/src/type_check.rs
```diff
@@ -450,12 +450,60 @@ impl Subtype for Type {
...
+				// any single-method interface with a matching "handle" method type.
```
yoav-steinberg, line 466: Unless we want to annotate these interfaces in some way (`@handler`) and then have this logic "anded" with the fact that such an interface is also correctly annotated.

libs/wingc/src/type_check.rs
```diff
@@ -450,12 +450,60 @@ impl Subtype for Type {
...
+
+				// First, check if there is exactly one inflight method in the interface
```
yoav-steinberg, line 468: Note this might be clearer using the `ClassLike` trait implemented for interface. I think you can do something like:
```rust
  let inflight_methods = iface.methods().filter( /* is inflight and is called "handler" */);
  let handler_method = inflight_methods.next();
  if handler_method.is_none() || inflight_methods.next().is_some() { return false; }
```

libs/wingc/src/type_check.rs
```diff
@@ -450,12 +450,60 @@ impl Subtype for Type {
...
+						method.type_
+					} else {
+						return false;
+					}
+				} else {
+					return false;
+				};
+
+				// Then get the type of the resource's "handle" method if it has one
+				let res_handle_type = if let Some(method) = res.get_env().try_lookup("handle", None) {
```
yoav-steinberg, line 494: again, might be easier using the `ClassLike` trait:
`res.get_method("handle".into())`

yoav-steinberg (commented):
Very nice. Added some comments and questions.

libs/wingc/src/ast.rs
```diff
@@ -118,6 +117,7 @@ pub enum TypeAnnotationKind {
 	Duration,
 	Json,
 	MutJson,
+	Resource,
```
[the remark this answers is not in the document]
Chriscbr: Hm - so I couldn't do this before because when I transformed closures like so:
```js
let fn = ((): A => {
  class A {
    // snip
  }
  return new A();
})();
```
... then `A` would not be available outside, so `(): A` was invalid. But since we're now emitting the class in a separate statement, as in...

```
class A {
  // snip
}
let fn = new A();
```
... then I think we can remove this new annotation kind. Let me see what I can do

libs/wingc/src/closure_transform.rs
Chriscbr: I would rather make the cases explicit just so that if we add or remove choices to this enum in the future, someone is required to update this transformer

libs/wingc/src/closure_transform.rs
Chriscbr: I think the idea sounds reasonable... unfortunately when I tried this now, I encountered an issue because when `is_static: false` on a FunctionDefinition does not imply it is an instance method (for example, an anonymous preflight function like `() => {}` is not currently treated as static in the compiler). I wasn't sure if we should change that, so I'll leave this logic for now

libs/wingc/src/closure_transform.rs
Chriscbr: It's possible for it to contain more than one statement if we are calling `fold_statement` on a statement with more than one inflight expression, for example `foo(inflight () => {}, inflight () => {})`

libs/wingc/src/jsify.rs
Chriscbr: Good catch. also TIL about `splitn` 😮

libs/wingc/src/jsify.rs
Chriscbr: Let me remove this hack and create a separate issue - I have a feeling this isn't the right fix, but it might require more investigation.

libs/wingc/src/ast.rs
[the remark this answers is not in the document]
yoav-steinberg: cool

libs/wingc/src/jsify.rs
Chriscbr: But TL;DR some bindings aren't generated correctly when accessing members of a resource (but the only case we have of it happening is with `api.url`)

libs/wingc/src/closure_transform.rs
yoav-steinberg: I see. It actually seems like a bug, if we change `build_function_definition` in _parser.rs_ to correctly set `is_static` to `true` when called from `build_anonymous_closure` it might resolve this. You may try or leave it for a future improvement...

libs/wingc/src/closure_transform.rs
yoav-steinberg: :exploding_head: Any chance you can show the "wing" output of that fold so I better understand what's going on here?

Chriscbr: Sure, so some code like this:
```js
let b = foo(inflight (key: str): str => {
  counter.inc();
}, inflight (key: str): str => {
  counter.dec();
});
```

will get transformed into this:

```js
class $Inflight1 {
  inflight handle(key: str): str {
    counter.inc();
  }
}
class $Inflight2 {
  inflight handle(key: str): str {
    counter.dec();
  }
}
let b = foo(new $Inflight1(), new $Inflight2());
```

I'm having trouble adding this as a unit test though

libs/wingc/src/lsp/sync.rs
Chriscbr: I don't think there should be any missing diagnostics, or if they should they should appear in our snapshots.

libs/wingc/src/type_check.rs
Chriscbr: Nice, this worked 👍

libs/wingc/src/type_check.rs
Chriscbr: I think I'll leave this as a future improvement for now

libs/wingc/src/closure_transform.rs
Chriscbr: Done

libs/wingc/src/closure_transform.rs
Chriscbr: This use case didn't work out of the box, but I've updated the visitor so it skips renaming `this` inside any classes defined inside the inflight closure, so it works now. (And I've added a unit test)

libs/wingc/src/closure_transform.rs
Chriscbr: OK - updated

libs/wingc/src/jsify.rs
```diff
@@ -1819,7 +1752,8 @@ impl Visit<'_> for FreeVariableScanner {
 				// If this reference is a capturable, non-reassignable, preflight
 				// variable then it's a capture
 				if let SymbolKind::Variable(v) = lookup_result {
-					if v.type_.is_capturable() && !v.reassignable && v.phase == Phase::Preflight {
+					// TODO: if we see a reassignable or non-capturable variable, should we emit an error here or inside analyze_expr?
+					if v.phase == Phase::Preflight {
```
Chriscbr, line 1756: @yoav-steinberg note this change - if I didn't do this, I saw some diagnostics started to disappear (about disallowing capturing reassignable values)... not sure if this is correct

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -97,49 +97,42 @@ error: Cannot find module \\"foobar\\" in source directory: Unable to load \\"fo
...
 `;
 
 exports[`capture_reassignable.w > stderr 1`] = `
-"error: Cannot capture a reassignable variable \\"x\\"
+"error: Cannot capture reassignable field 'x'
   --> ../../../examples/tests/invalid/capture_reassignable.w:6:15
   |
 6 |     log(\\"x: \${x}\\");
-  |               ^ Cannot capture a reassignable variable \\"x\\"
+  |               ^ Cannot capture reassignable field 'x'
```
Chriscbr, line 135: This should say "variable" and not "field" (TODO: create issue)

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: Err- I recovered the error so it's still produced, but the error makes less sense since it references `$Inflight1` now.

libs/wingc/src/jsify.rs
yoav-steinberg: Ok, I guess `analyze_expr` created those diags. Since this is in jsification phase, it's a bit of an issue either way. We need these diags to be available in the lsp too..
@staycoolcall911 Lets have a discussion about this and see if this is P1, if we need an issue for this or what...

yoav-steinberg: https://github.com/winglang/wing/issues/2548

PR #2423, opened 2023-05-06
fix(sdk): pinned cdktf providers version

### LINKED ISSUE #2417: Pin the SDK's CDKTF provider versions to avoid changes in builds
The versions listed in https://github.com/winglang/wing/blob/main/libs/wingsdk/src/.gen/versions.json are getting automatically updated in contributors' PRs. These should probably be pinned so that the versions of bindings we are generating are stable and reproducible.

garysassano:
Closes #2417


Chriscbr (approved):
Thanks @garysassano!

PR #2426, opened 2023-05-07
chore(sdk): making `api.vars` mandatory and returning an empty map by default

### LINKED ISSUE #2144: `req.vars` should not be optional
### I tried this

Use `req.vars`

### I expected this:

If there are no vars in API requests, the `req.vars` map should not be `nil` but rather an empty map.

### Instead, this happened

It is optional, so it was harder to use

### Is there a workaround?

Yes, something like:

```js
let vars = req.vars ?? Map<str>{};
```

### Component

SDK

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

garysassano:
Closes #2144


tsuf239 (commented):
Great! Can you please:
1. change the code on `examples/tests/valid/api_path_vars.w`, 
2. run `npx nx test hangar` (to generate snapshots)
3. run `npm run build`

docs/04-reference/wingsdk-api.md
```diff
@@ -1557,39 +1557,39 @@ The request's query string values.
 
 ---
 
-##### `body`<sup>Optional</sup> <a name="body" id="@winglang/sdk.cloud.ApiRequest.property.body"></a>
+##### `vars`<sup>Required</sup> <a name="vars" id="@winglang/sdk.cloud.ApiRequest.property.vars"></a>
```
staycoolcall911, line 1560: Hey @garysassano we would be grateful if you could also update the relevant part of [Wing SDK Spec](https://github.com/winglang/wing/blob/main/docs/04-reference/wingsdk-spec.md#api) as part of your PR 🙏

examples/tests/valid/api_path_vars.w
```diff
@@ -7,7 +7,7 @@ class Fetch {
 }
 
 let handler = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
-  let vars = req.vars ?? {name: ""};
+  let vars = req.vars;
   return cloud.ApiResponse {
     body: {user: vars.get("name")},
```
tsuf239, line 12:
```suggestion
  return cloud.ApiResponse {
    body: {user:  req.vars.get("name")},
```

docs/04-reference/wingsdk-api.md
garysassano: Thanks a lot for your input, updated that as well.

tsuf239 (approved):
🏆

PR #2433, opened 2023-05-07
chore: don't run tests for pure docs changes

### LINKED ISSUE #2428: As a contributor I want to make changes to the documentation even faster by avoiding running all tests when merging to main
### Feature Spec

We would like to avoid running all tests when a PR is created against the documentation under https://github.com/winglang/wing/tree/main/docs

Followup for #1088. 

After the work done in PR #2346 for doc changes - we're skipping tests on the dev branch.
We want to also avoid unnecessary steps when the dev branch merges to `main`.

This allows very quick iterations on the docs.

CC: @MarkMcCulloh, @WeepingClown13, @ShaiBer 

### Use Cases

See #1088

### Implementation Notes

_No response_

### Component

Documentation, Contributor Experience

### Community Notes

WeepingClown13:
Adds a new workflow that'd only run for pure doc changes and doesn't run tests. As such also limits the current build flow to not run at all for pure doc changes.

Not sure if we can skip the dependency installation as changelog is generated with `npm run changelog`. If we can, we can speed it up a bit more.

closes #2428 if a new workflow file is an acceptable solution.

PS: This is not entirely new, just takes the necessary parts from the existing workflow.


.github/workflows/build.yml
```diff
@@ -12,6 +12,10 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - "docs/**"
+      - "logo/**"
+      - "*.md"
```
eladb, line 18: This might be tricky since we are going to be having md files within the sdk directory.

Let's remove this filter for now

.github/workflows/release-docs
```diff
@@ -0,0 +1,123 @@
+name: Release docs
+
+on:
+  push:
+    branches:
```
eladb, line 5: Is there a way to avoid the duplicated workflow?

eladb (changes requested):

.github/workflows/build.yml
WeepingClown13: This only filters the root level .md files, so it should be fine. The general query syntax should be **.md.

.github/workflows/release-docs
WeepingClown13: The best I could come up with was skip wing release, and still trigger docsite update but that'd require a logic change on even the docsite repo. Or we could skip both for pure doc changes. The only solution that allows wing release and also skips tests which I found was this. Ofc I'm open to suggestions.

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,88 @@
+name: Release docs
```
MarkMcCulloh, line 1: An idea to maybe simplify this and reduce some duplication:

What if this job only updates the existing release assets (specifically the docs tarball), and then triggers the docsite update directly like https://github.com/winglang/wing/blob/main/.github/workflows/update-docsite.yml

If the only change is pure docs, I think it's safe to avoid versioning/releasing as long as the documentation site is told to update.

WeepingClown13: This is already downloading existing assets and modifying only the tarball.
In regards to the last part, I believe that was my initial suggestion; skipping release. But at this point I'd lean towards not skipping the release because while the workflow is not entirely new, it gives a fast and proper solution that'd not affect the existing release logic. What do you say?

scripts/docs/package.json
MarkMcCulloh, line 1: It may now be the time for the scripts directory to get its own package.json. If you move this update, you can also avoid moving the changelog.mjs file.

scripts/docs/downloadAssets.ts
```diff
@@ -0,0 +1,54 @@
...
+      const filePath = join(dir, asset.name);
+      await fs.writeFile(filePath, Buffer.from(JSON.stringify(response.data)));
+      console.log(`Downloaded ${asset.name}.`);
+    }
+
+    console.log("Downloaded all assets successfully.");
+  } catch (error) {
+    console.error(error);
+  }
+})();
```
MarkMcCulloh, line 54: Can't use top-level await? I thought using tsx would help with that. You could change this to an mts file or change the package type to module.

.github/workflows/release-docs.yml
MarkMcCulloh: Ah my bad, I misread some of the changes. Nice!

The original reason to avoid skipping a release was just to ensure the docsite gets updated. I'm cool with this approach.

scripts/downloadAssets.mts
```diff
@@ -0,0 +1,52 @@
...
+    );
+    const filePath = join(dir, asset.name);
+    await fs.writeFile(filePath, Buffer.from(JSON.stringify(response.data)));
+    console.log(`Downloaded ${asset.name}.`);
+  }
+
+  console.log("Downloaded all assets successfully.");
+} catch (error) {
+  console.error(error);
+};
```
MarkMcCulloh, line 52: Not sure I see the benefit of this catch. If anything inside throws, the effect will be the same if we didn't catch except we lose the stacktrace.

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          working-directory: scripts/
+
+      - name: Generate changelong
+        id: changelog
+        env:
+          GENERATE_VERSION: false
+        run: cd scripts/ && npm run changelog
```
MarkMcCulloh, line 42: nit:

```suggestion
        id: changelog
        working-directory: scripts
        env:
          GENERATE_VERSION: false
        run: npm run changelog
```

scripts/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "docs",
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+      "changelog": "node changelog.mjs"
+    },
+    "author": "",
+    "license": "ISC",
```
MarkMcCulloh, line 10:
```suggestion
```

also maybe just minimize this package.json in general

scripts/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "docs",
```
MarkMcCulloh, line 2:
```suggestion
    "name": "scripts",
```

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,85 @@
...
+      - main
+    paths:
+      - "docs/**"
+      - "logo/**"
+      - "*.md"
+  workflow_dispatch: {}
+      
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.after || github.ref }}
+  cancel-in-progress: true
```
MarkMcCulloh, line 15: If a regular release build is running and then this job is triggered before it completes, we'll likely run into a problem because they could both try to tag the same version. We can't make their groups the same because if this cancels the other full build, it won't include the changes from it. A similar race condition happens if this build is running first and then the full build starts.

I was on board with this approach although I think this issue is a blocker. WDYT?

MarkMcCulloh (changes requested):

.github/workflows/release-docs.yml
WeepingClown13: It really is quite an issue. I did consider about the concurrency issue a lot and that's how I settled on `github.event.after` in the id so that every job will have a unique id. But it was an oversight on my end not thinking even if commits are different, release could be the same. Now the one remaining idea I have is to just add a job to check all the changes since last commit with a script, add this workflow as another job in the build and then conditionally trigger the jobs based on file check output, instead of filtering paths on the workflow level. Though it feels like this is all getting a bit complicated so I don't mind scrapping the pr and handing the issue over to someone else with a better idea. Do you think the idea could end up as a proper solution?

MarkMcCulloh: Yeah I think this approach will have to be scrapped unfortunately, can't really think of a way to overcome this issue when there is a separate workflow.

> Now the one remaining idea I have is to just add a job to check all the file changes since last commit with a script, add this workflow as another job in the build and then conditionally trigger the jobs based on file check output, instead of filtering paths on the workflow level.

I do think the idea would work (have both seen and done similar things) but it may be time to take a step back and evaluate the core pieces here. Notably: The build system (nx), the workflows, and the boundaries between them. It's a little messy right now which makes this PR difficult. Additionally, as Wing matures, versioning will likely change which complicates this more.

So for now we can close this, but I really love the work you've done on the workflows ❤️

PR #2435, opened 2023-05-08
fix(sdk): updating website permission

marciocadev:
I was testing cloud.Website() and there was a small issue with permissions.
I am uploading the fix, but you can already see the result at this link.

[static website example](https://d24lp7k5o7ycd7.cloudfront.net/index.html)

Another thing is that in AWS it is not possible to know the URL of the website at compile time since the URL is only created when the deployment is completed (and in this case, when CloudFront is finished activating).

### DIFF
--- a/libs/wingsdk/src/target-tf-aws/bucket.ts
+++ b/libs/wingsdk/src/target-tf-aws/bucket.ts
@@ -181,6 +181,13 @@ export function createEncryptedBucket(
   });
 
   if (isPublic) {
+    new S3BucketPublicAccessBlock(scope, "PublicAccessBlock", {
+      bucket: bucket.bucket,
+      blockPublicAcls: false,
+      blockPublicPolicy: false,
+      ignorePublicAcls: false,
+      restrictPublicBuckets: false,
+    });
     const policy = {
       Version: "2012-10-17",
       Statement: [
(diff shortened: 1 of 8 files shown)


libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -181,6 +181,13 @@ export function createEncryptedBucket(
   });
 
   if (isPublic) {
+    new S3BucketPublicAccessBlock(scope, "PublicAccessBlock", {
```
Chriscbr, line 184: Why do we need a resource named S3BucketPublicAccessBlock when the bucket is supposed to be public? Maybe we can add a comment to explain why its needed in the case of `cloud.Website`

Chriscbr (approved):

Chriscbr:
> Another thing is that in AWS it is not possible to know the URL of the website at compile time since the URL is only created when the deployment is completed (and in this case, when CloudFront is finished activating)

Can you create an issue for this? I wonder if we need to make `url` an inflight property, or change its type to `str?` to indicate it's not available in some circumstances...

libs/wingsdk/src/target-tf-aws/bucket.ts
marciocadev: When we create a bucket it's always blocked,  to give access we need to remove this blocked then create the access policy

If we not open the bucket the policy it's not created at all and break the deploy because it's not permitted give access to a blocked bucket

Chriscbr: Oh I see - sounds like it might be related to this https://aws.amazon.com/about-aws/whats-new/2022/12/amazon-s3-automatically-enable-block-public-access-disable-access-control-lists-buckets-april-2023/

PR #2436, opened 2023-05-08
feat: `util.env()` and `util.try_env()` (WIP)

### LINKED ISSUE #2432: `util.env(key)`
### Feature Spec

The `util.env(name: str): str` function can be used to retrieve the value of an environment variable. This function will throw an error if the variable is not defined, _or if it is an empty string_.

The `util.try_env(name: str): str?` variant will return an optional string which will be nil if the value is not defined or is an empty string.

### Use Cases

Get the value of an environment variable.

Related to #1853 

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

eladb:
Introduce utility functions for reading environment variables.

At the moment, the functions are available under the `Util` class of the new `util` module:

   bring util;
   log(util.Util.env("PATH");

In a subsequent PR we will "hoist" these static functions to become `util.env()`.

Additionally, we cannot currently test these inflight. I believe @yoav-steinberg is working to address this. Is that correct?

Resolves #2432

### Misc

Reorganize SDK non public utility functions. Most of them were actually only used by tests.


examples/tests/valid/util.env.w
```diff
@@ -0,0 +1,26 @@
...
+let var failed = false;
+try {
+  util.Util.env(random_name);
+} catch {
+  failed = true;
+}
+assert(failed);
+
+// TODO: this is not supported yet (P1)
+/*
```
eladb, line 17: @yoav-steinberg is this supposed to be resolved by https://github.com/winglang/wing/pull/2158?

libs/wingsdk/src/std/resource.ts
```diff
@@ -168,12 +167,6 @@ export abstract class Resource extends Construct implements IResource {
    * @internal
    */
   public _registerBind(host: IInflightHost, ops: string[]) {
-    log(
```
eladb, line 171: The only log in the village, so remove for now...

libs/wingsdk/src/target-sim/app.ts
```diff
@@ -32,7 +32,11 @@ import * as core from "../core";
 import { preSynthesizeAllConstructs } from "../core/app";
 import { REDIS_FQN } from "../redis";
 import { WingSimulatorSchema } from "../testing/simulator";
-import { SIMULATOR_FILE_PATH } from "../util";
+
+/**
+ * Path of the simulator configuration file in every .wsim tarball.
+ */
+export const SIMULATOR_FILE_PATH = "simulator.json";
```
eladb, line 39: Hoping no one reads this (@ainvoner ?)

examples/tests/valid/util.env.w
```diff
@@ -0,0 +1,26 @@
+bring cloud;
+bring util;
+
+let random_name = "RANDOM123412121212kjhkjskdjkj";
+
+assert(util.Util.env("PATH").length > 0);
+
+let var failed = false;
+try {
+  util.Util.env(random_name);
```
Chriscbr, line 10: Will this in the future be an `unphased` function? (naming tbd...)

libs/wingsdk/src/std/resource.ts
Chriscbr: I'd be lying if I said I haven't used this at least once or twice for debugging the binding system with Hasan and others. 🙂 How about we keep it and do this instead?

```js
if (process.env.DEBUG) {
  console.error(<message>);
}
```

Chriscbr (approved):

examples/tests/valid/util.env.w
eladb: Yes

PR #2443, opened 2023-05-09
chore: centralize wing installation script for ci and user workflows

MarkMcCulloh:
There are several places where it is essential to create a usable (and often temporary) wing toolchain installation from an arbitrary source. This source would be directories, tarballs, or even just plain-old versions from the npm registry.

The logic for this was exclusive to hangar, but is needed elsewhere. To make sure it's easily available, this PR extracts and simplifies the logic. A convenient symlinking ability is also added.


eladb (approved):
Shiny!

PR #2444, opened 2023-05-09
fix(compiler): friendly error message when defining preflight class inflight

### LINKED ISSUE #1710: Defining a resource in an inflight causes a panic
### Community Note

### I tried this:

```ts
bring cloud;

  class R {
    init() {}
  }

  log("hello world!");
};
```

### I expected this:

A compiler error saying that a resource cannot be defined in inflight.

A few reasons I might expect this:
1. It's not possible to instantiate the resource inflight (see https://docs.winglang.io/blog/2023/02/17/preflight-vs-inflight)
2. A resource can contain preflight methods and a preflight constructor, and introducing sections of preflight code inside of inflight code would not make much sense since they could never be executed.

Counterargument: it's possible a resource could just be defined by a user to group together static inflight methods, so I could also see it being reasonable if the above code didn't error.

### Instead, this happened:

```
thread '<unnamed>' panicked at 'assertion failed: !*is_resource || env.flight == Phase::Preflight', libs/wingc/src/type_check.rs:1738:17
```

### Component:

Compiler

### Wing Version:

0.5.85

### Node.js Version:

18.14.1

### Platform(s):

MacOS

### Anything else?

_No response_

staycoolcall911:
Fixes #1710.

Copilot summary wasn't great to say the least... (you can see it at the bottom).

This PR makes sure a friendly error message is shown when a preflight class (`class`) is defined in `inflight` scope.

-----------------------------------------------------------------------

### <samp>🤖 Generated by Copilot at fb5cbc4</samp>

This pull request fixes two bugs related to preflight and inflight code in `wingc`, the Wing compiler. It prevents the compiler from panicking or asserting when trying to capture variables or resources that are not available in preflight code, and reports an error when defining a preflight class in an inflight scope.

### DIFF
--- a/libs/wingc/src/capture.rs
+++ b/libs/wingc/src/capture.rs
@@ -328,8 +328,10 @@ fn scan_captures_in_inflight_scope(scope: &Scope, diagnostics: &mut Diagnostics)
 	let env_ref = scope.env.borrow();
 	let env = env_ref.as_ref().unwrap();
 
-	// Make sure we're looking for captures only in inflight code
-	assert!(matches!(env.phase, Phase::Inflight));
+	// Early bail-out if we're looking for captures in preflight code - an error will be reported in type checking.
+	if matches!(env.phase, Phase::Preflight) {
+		return res;
+	}
 
 	for s in scope.statements.iter() {
 		match &s.kind {
--- a/examples/tests/invalid/preflight_class_inflight.w
+++ b/examples/tests/invalid/preflight_class_inflight.w
@@ -1,0 +1,9 @@
+bring cloud;
+
+let handler = inflight () => {
+  class R {
+    init() {}
+  }
+//^ Cannot define a preflight class in inflight scope
+  log("hello world!");
+};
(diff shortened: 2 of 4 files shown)


libs/wingc/src/capture.rs
```diff
@@ -328,8 +328,10 @@ fn scan_captures_in_inflight_scope(scope: &Scope, diagnostics: &mut Diagnostics)
 	let env_ref = scope.env.borrow();
 	let env = env_ref.as_ref().unwrap();
 
-	// Make sure we're looking for captures only in inflight code
-	assert!(matches!(env.phase, Phase::Inflight));
+	// Early bail-out if we're looking for captures in preflight code - an error will be reported in type checking.
+	if matches!(env.phase, Phase::Preflight) {
```
staycoolcall911, line 332: After emitting an error instead of the assert in type_check.rs - this assert surfaced. Decided to bail early if calling the `scan_captures_in_inflight_scope` method in preflight. Happy to hear if we should further limit this if statement.

MarkMcCulloh (approved):

PR #2448, opened 2023-05-09
fix(cli): trying to fix wing the test method

tsuf239:
image


apps/wing/src/commands/test.ts
```diff
@@ -200,7 +200,11 @@ async function checkTerraformStateIsEmpty(synthDir: string) {
       );
     }
   } catch (err) {
```
MarkMcCulloh, line 202: maybe add these types to line 202, that way you can should be able to avoid casting as `any` (will still need to verify the types though, but TS should understand your intent)

```suggestion
  } catch(err: string | { stderr: string }) {
```

MarkMcCulloh: even better if we know the "real" type of the error with stderr (looks like an `exec` of some kind?)

tsuf239: Found the type!! it's just a regular `Error` :) 
@Chriscbr was there a reason it accessed `(err as any).stderr` earlier?

MarkMcCulloh: It's possible the real error is actually a subclass of `Error`, so it can have that additional property.

tsuf239: added a unit test to this method: checkTerraformStateIsEmpty

Chriscbr (approved):

Issue #1305, opened 2023-01-29
cloud.Service - Simulator implementation


staycoolcall911:
https://docs.winglang.io/reference/wingsdk-spec#service

monadabot:
Congrats! :rocket: This was released in [Wing 0.16.5](https://github.com/winglang/wing/releases/tag/v0.16.5).

PR #2450, opened 2023-05-09
feat(sdk): support service resource in sim

hasanaburayyan:
Proposed mvp for `cloud.Service` sim implementation

Closes: https://github.com/winglang/wing/issues/1305


libs/wingsdk/src/cloud/service.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Properties for `Service`.
+ */
+export interface ServiceProps {
+  /**
+   * Handler to run with the service starts.
+   */
+  readonly onStart: IServiceOnEventHandler;
+  /**
+   * Handler to run with the service stops.
+   * @default undefined
```
eladb, line 22:
```suggestion
   * @default - no special activity at shutdown
```

libs/wingsdk/src/target-sim/service.ts
```diff
@@ -0,0 +1,103 @@
...
+  private onStopHandlerToken?: string;
+
+  constructor(scope: Construct, id: string, props: cloud.ServiceProps) {
+    super(scope, id, props);
+    this.autoStart = props.autoStart ?? true;
+
+    // On Start Handler
+    const onStartHash = props.onStart.node.addr.slice(-8);
+
+    const onStartFunctionHandler = convertBetweenHandlers(
```
eladb, line 26: There's probably some reuse opportunity...

Extract the code from lines 26..48 into a method and reuse in lines 52..75.

examples/tests/valid/service.w
```diff
@@ -0,0 +1,28 @@
+// TODO: uncomment test with https://github.com/winglang/wing/issues/1306
```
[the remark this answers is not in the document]
hasanaburayyan: There is not a way right now to skip hangar from running both compile for tf-aws and sim and currently only sim will work.

PR #2467, opened 2023-05-09
chore(sdk): remove extra void

Chriscbr:

### DIFF
--- a/libs/wingsdk/src/target-sim/topic.inflight.ts
+++ b/libs/wingsdk/src/target-sim/topic.inflight.ts
@@ -50,7 +50,7 @@ export class Topic
         timestamp: new Date().toISOString(),
       });
 
-      void (await fnClient.invoke(message).catch((err) => {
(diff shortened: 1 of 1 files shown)


libs/wingsdk/src/target-sim/topic.inflight.ts
```diff
@@ -50,7 +50,7 @@ export class Topic
         timestamp: new Date().toISOString(),
       });
 
-      void (await fnClient.invoke(message).catch((err) => {
```
MarkMcCulloh, line 53: Was this caught by eslint or did you just notice it? eslint does have a rule for this.

eladb (approved):

libs/wingsdk/src/target-sim/topic.inflight.ts
Chriscbr: Just noticed it. Definitely open to adding eslint rules though

PR #2480, opened 2023-05-10
feat(sdk): adding try methods to bucket

garysassano:


libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -277,6 +277,36 @@ export interface IBucketClient {
...
+   * @returns `true` if the object was successfully stored, `false` if an error occurred.
+   */
+  tryPutJson(key: string, body: Json): Promise<boolean>;
+
+  /**
+   * Tries to retrieve an object from the bucket.
+   * @param key Key of the object.
+   * @returns `true` if the object was successfully retrieved, `false` if an error occurred.
+   */
+  tryGet(key: string): Promise<boolean>;
```
Chriscbr, line 302: Please take a look at our SDK spec for bucket [here](https://docs.winglang.io/reference/wingsdk-spec#bucket). If the object exists, we still want to obtain its contents, so it should look something like this:
```suggestion
  /**
   * Get an object from the bucket if it exists
   * @param key Key of the object.
   * @returns the contents of the object as a string if it exists, nil otherwise
   */
  tryGet(key: string): Promise<string | undefined>;
```

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -277,6 +277,36 @@ export interface IBucketClient {
...
+
+  /**
+   * Tries to retrieve a Json object from the bucket.
+   * @param key Key of the object.
+   * @returns `true` if the object was successfully retrieved, `false` if an error occurred.
+   */
+  tryGetJson(key: string): Promise<boolean>;
```
Chriscbr, line 309:
```suggestion
  /**
   * Gets an object from the bucket if it exists, parsing it as Json.
   * @param key Key of the object.
   * @returns the contents of the object as Json if it exists, nil otherwise
   */
  tryGetJson(key: string): Promise<Json | undefined>;
```

Chriscbr (changes requested):

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -373,10 +403,18 @@ export enum BucketInflightMethods {
...
+  /** `Bucket.putJson` */
   PUT_JSON = "put_json",
-  /** `Bucket.getJson */
+  /** `Bucket.getJson` */
   GET_JSON = "get_json",
-  /** `Bucket.publicUrl */
+  /** `Bucket.publicUrl` */
   PUBLIC_URL = "public_url",
+  /** `Bucket.tryPut` */
+  TRY_PUT = "try_put",
```
revitalbarletz, line 413: After your merge from `main`, you'll need to change these to camelCase:

```suggestion
  TRY_PUT = "tryPut",
```

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -277,6 +277,36 @@ export interface IBucketClient {
...
+   * @param body Content of the object we want to store into the bucket.
+   * @returns `true` if the object was successfully stored, `false` if an error occurred.
+   */
+  tryPut(key: string, body: string): Promise<boolean>;
+
+  /**
+   * Tries to put a Json object in the bucket.
+   * @param key Key of the object.
+   * @param body Json object that we want to store into the bucket.
+   * @returns `true` if the object was successfully stored, `false` if an error occurred.
```
revitalbarletz, line 293:
```suggestion
   * @returns `true` if the object was successfully stored, `false` if an an object already existed for the same key.
```

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -170,6 +170,42 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     });
   }
 
+  public async tryPut(key: string, value: string): Promise<boolean> {
+    try {
+      await this.put(key, value);
+      return true;
+    } catch (error) {
+      return false;
+    }
```
revitalbarletz, line 179: In the simulator, since we are single threaded, we can just check existence and return false:

```suggestion
    if (this.fileExists(key)) {
      return false;
    }
    
    return this.put(key, value); // return the promise (await not needed)
```

libs/wingsdk/src/cloud/bucket.ts
garysassano: @revitalbarletz thanks, I saw #2468 but I thought it wasn't applying there.

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -83,6 +83,40 @@ export class BucketClient implements IBucketClient {
...
+      return undefined;
+    }
+  }
+
+  public async tryGetJson(key: string): Promise<Json | undefined> {
+    try {
+      return await this.getJson(key);
+    } catch (error) {
+      return undefined;
+    }
```
tsuf239, line 117: Why do the targets behave differently? 
Are we checking only the case in which the key already exists? or any other case is counted too?

tsuf239 (changes requested):
Can you please add a test for each platform and an SDK spec test for all of the new methods?

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -277,6 +277,36 @@ export interface IBucketClient {
...
    */
   delete(key: string, opts?: BucketDeleteOptions): Promise<void>;
+
+  /**
+   * Tries to put an object in the bucket.
+   * @param key Key of the object.
+   * @param body Content of the object we want to store into the bucket.
+   * @returns `true` if the object was successfully stored, `false` if an error occurred.
+   */
+  tryPut(key: string, body: string): Promise<boolean>;
```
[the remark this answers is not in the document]
Chriscbr: I'm ok with keeping these (and you can add them to the spec document @garysassano) or leaving them out. I'm not sure if `tryPut` is as common of a use case as `tryGet`

revitalbarletz:
@tsuf239 can you elaborate what do you mean by "Can you please add a test for each platform and an SDK spec test for all of the new methods?" are these the generated files? if so what should we do to run the hanger tests? These details are missing

tsuf239:
> @tsuf239 can you elaborate what do you mean by "Can you please add a test for each platform and an SDK spec test for all of the new methods?" are these the generated files? if so what should we do to run the hanger tests? These details are missing

Sure!
The first tests to be added are the vitest unit tests, they should be added to those 3 files: 
- `libs/wingsdk/test/shared-aws/bucket.inflight.test.ts`
- `libs/wingsdk/test/target-sim/bucket.test.ts`
- `libs/wingsdk/test/target-tf-azure/bucket.inflight.test.ts`

The other kind of test is an SDK spec test:
- see here: https://github.com/winglang/wing/issues/2520
**this one was written when only tryGet was planned, please add one for each of the methods you've added**

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
garysassano: They behave differently because the `cloud.Bucket` implementation for `tf-azure` target misses an `exist` method compared to `aws-shared` and `sim`.
Regarding your second question, I think we are just checking if a key already exists.

garysassano: They behave differently because the `cloud.Bucket` implementation for `tf-azure` target lacks an `exist` method compared to `aws-shared` and `sim`.
Regarding your second question, I think we are just checking if a key already exists.

tsuf239: If so, let's add the `exist` method to azure too so they all behave the same :)
another option is to skip the implementation for azure for now, 

Another question that I have now is, if the `try` role is to check if the key exists, why do we need the `tryPut`? maybe can we omit it for now? what do you think?

garysassano: The Azure implementation is mandatory or a test would fail.
As for removing `tryPut`, I actually thought that was something you decided to add internally, since it was requested by @revitalbarletz on Slack. But if you changed your mind and that actually not needed, I'm fine with removing it.

garysassano: The Azure implementation is mandatory or a test would fail.

As for removing `tryPut`, I actually thought that was something you decided to add internally, since it was requested by @revitalbarletz on Slack. But if you changed your mind and that's actually not needed, I'm fine with removing it.

tsuf239: it'll be great, thanks! We apologize for the confusion :)

libs/wingsdk/test/shared-aws/bucket.inflight.test.ts
```diff
@@ -273,3 +273,121 @@ test("check that an object doesn't exist in the bucket", async () => {
...
+
+  // THEN
+  expect(objectTryGet).toEqual(undefined);
+});
+
+test("tryGetJson an existing object from the bucket", async () => {
+  // GIVEN
+  const BUCKET_NAME = "BUCKET_NAME";
+  const KEY = "KEY";
+  const VALUE = "VALUE";
```
tsuf239, line 321: can this value be a real object?

libs/wingsdk/test/shared-aws/bucket.inflight.test.ts
```diff
@@ -273,3 +273,121 @@ test("check that an object doesn't exist in the bucket", async () => {
...
+});
+
+test("tryGet a non-existent object from the bucket", async () => {
+  // GIVEN
+  const BUCKET_NAME = "BUCKET_NAME";
+  const KEY = "KEY";
+  const VALUE = "VALUE";
+  s3Mock
+    .on(GetObjectCommand, { Bucket: BUCKET_NAME, Key: KEY })
+    .resolves({ Body: createMockStream(VALUE) });
```
tsuf239, line 304: this should be no value, isn't it?

libs/wingsdk/test/shared-aws/bucket.inflight.test.ts
```diff
@@ -273,3 +273,121 @@ test("check that an object doesn't exist in the bucket", async () => {
...
+});
+
+test("tryGetJson a non-existent object from the bucket", async () => {
+  // GIVEN
+  const BUCKET_NAME = "BUCKET_NAME";
+  const KEY = "KEY";
+  const VALUE = "VALUE";
+  s3Mock
+    .on(GetObjectCommand, { Bucket: BUCKET_NAME, Key: KEY })
+    .resolves({ Body: createMockStream(JSON.stringify(VALUE)) });
```
tsuf239, line 344: this should be no value, isn't it?

libs/wingsdk/test/target-sim/bucket.test.ts
```diff
@@ -572,3 +572,76 @@ test("check if an object exists in the bucket", async () => {
...
+test("tryGetJson objects from bucket", async () => {
+  // GIVEN
+  const app = new SimApp();
+  cloud.Bucket._newBucket(app, "my_bucket");
+
+  const s = await app.startSimulator();
+  const client = s.getResource("/my_bucket") as cloud.IBucketClient;
+
+  const KEY = "file.json";
+  const VALUE = JSON.stringify({ msg: "Hello world!" });
```
tsuf239, line 607: should it be an object? or a string?

libs/wingsdk/test/target-tf-azure/bucket.inflight.test.ts
```diff
@@ -211,6 +303,8 @@ class MockBlobClient extends BlobClient {
       return Promise.resolve(false);
     }
   }
+
+  tryGetJson;
```
tsuf239, line 307: can be removed?

libs/wingsdk/test/target-tf-azure/bucket.inflight.test.ts
```diff
@@ -190,6 +190,98 @@ test("check that an object doesn't exist in the bucket", async () => {
   expect(objectExists).toEqual(false);
 });
 
+test("tryGet an existing object from the bucket", async () => {
```
tsuf239, line 193: can you please add one for tryGetJson too?

tsuf239 (commented):
Really great job Gary!

libs/wingsdk/test/shared-aws/bucket.inflight.test.ts
garysassano: fixed.

libs/wingsdk/test/shared-aws/bucket.inflight.test.ts
garysassano: fixed.

libs/wingsdk/test/shared-aws/bucket.inflight.test.ts
garysassano: fixed.

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -68,53 +91,83 @@ export class BucketClient implements IBucketClient {
...
-      await this.s3Client.send(command);
-    return region;
+  /**
+   * Get a Json object from the bucket if it exists
+   *
+   * @param key Key of the object
+   * @returns Json content of the object
+   */
+  public async tryGetJson(key: string): Promise<Json | undefined> {
+    if (await this.exists(key)) {
```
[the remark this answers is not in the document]
garysassano: Added edge case test `tryGetJson an existing non-Json object from the bucket` for all targets.

tsuf239 (approved):
🚀 🚀 🚀

Chriscbr (approved):

Issue #1385, opened 2023-01-31
cloud.Queue.pop()


staycoolcall911:
https://docs.winglang.io/reference/wingsdk-spec#queue

Bharadwajshivam28:
> https://docs.winglang.io/reference/wingsdk-spec#queue

Assign it to me sir I want to work on it

Chriscbr:
@Bharadwajshivam28 This issue is about implementing the `pop` inflight method described in the spec. Right now it says it "Pops (deletes and returns) a message from the queue". Here's an example of how it would be used in Wing:

```js
bring cloud;

let q = new cloud.Queue();

new cloud.Function(inflight () => {
  q.push("hello");
  q.push("world");
  let item = q.pop();
  // Right now we are not assuming the queue is FIFO
  assert(item == "hello" || item == "world");
}) as "test";
```

Here's an example of a PR that added two inflight methods whose code could be helpful to look at as a reference https://github.com/winglang/wing/pull/1712/files

Bharadwajshivam28:
okay

Bharadwajshivam28:
@Chriscbr can you please drop a short demo video .. like then it will be good for me work .. if you can then please

Chriscbr:
Hey @Bharadwajshivam28 - I highly recommend you first check out our [development guide](https://docs.winglang.io/contributors/development)! It will guide you on how to build our monorepo, which contains all of the code for the Wing compiler toolchain. We are still working on improving this guide, so if you run into errors or problems getting the project to build locally, please let us know since it's still changing. 🙂

Bharadwajshivam28:
> Hey @Bharadwajshivam28 - I highly recommend you first check out our [development guide](https://docs.winglang.io/contributors/development)! It will guide you on how to build our monorepo, which contains all of the code for the Wing compiler toolchain. We are still working on improving this guide, so if you run into errors or problems getting the project to build locally, please let us know since it's still changing. 🙂 

Okay thanks

monadabot:
Congrats! :rocket: This was released in [Wing 0.17.9](https://github.com/winglang/wing/releases/tag/v0.17.9).

Issue #2284, opened 2023-04-26
Return optional on Queue.pop


flyingImer:
### Feature Spec

In current spec, `inflight pop(): Json;`, however, this assumes messages exist in the queue, or it throws runtime exception otherwise. Cloud implies being distributed. Even `peek` ahead of `pop` does not guarantee the message existence. Hence, runtime exception handling is inevitable. 

From developer standpoint, it would be smoother to have `inflight pop(): Json?;`. Instead of dealing with `try...catch`, it can be as simple as:
```
if let msg = myQueue.pop() {
  // process my message
} else {
  log("no message");
}
```

### Use Cases

This allows blind queue polling without worrying out runtime exceptions on an empty queue.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

fynnfluegge:
@flyingImer Guess this can be closed right? Seems that github does not recognize Closes #issues concatenized with *and* in PR description.

PR #2506, opened 2023-05-13
feat(sdk): add queue pop operation

flyingImer:
This code change adds support for sim and aws targets. Besides, it updates the spec to return an optional instead as proposed in https://github.com/winglang/wing/issues/2284.

Closes https://github.com/winglang/wing/issues/1385 and https://github.com/winglang/wing/issues/2284


docs/04-reference/wingsdk-api.md
```diff
@@ -5514,6 +5515,16 @@ approxSize(): num
 
 Retrieve the approximate number of messages in the queue.
 
+##### `pop` <a name="pop" id="@winglang/sdk.cloud.IQueueClient.pop"></a>
+
+```wing
+pop(): str
```
flyingImer, line 5521: hmm, this is supposed to be `pop(): str?`, any clue what went wrong?

docs/04-reference/wingsdk-spec.md
```diff
@@ -454,9 +454,9 @@ resource Queue {
   inflight push(message: Json): void;
 
   /**
-   * Pops (deletes and returns) a message from the queue.
+   * Pops (deletes and returns) a message from the queue if any.
    */
-  inflight pop(): Json;
+  inflight pop(): Json?;
```
flyingImer, line 460: there is some discrepancy between the spec and the IQueueClient, where spec pushes and pops `Json` but IQueueClient goes with `string`. LMK which one we should fix

libs/wingsdk/API.md
```diff
@@ -5507,6 +5508,16 @@ approxSize(): num
 
 Retrieve the approximate number of messages in the queue.
 
+##### `pop` <a name="pop" id="@winglang/sdk.cloud.IQueueClient.pop"></a>
+
+```wing
+pop(): str
```
flyingImer, line 5514: same issue here, should be `pop(): str?`

docs/04-reference/wingsdk-api.md
Chriscbr: Looks like a bug in the doc generation tool. I think it's related how we had to hack to support `Promise<T | undefined>` (something jsii doesn't allow at the moment). Can you open an issue?

docs/04-reference/wingsdk-spec.md
Chriscbr: I think long term we want to use `Json` since it's closer to a future where we support queues of user-defined structs. But keeping the current implementation with strings is okay for now, since switching to Json will require a breaking change

libs/wingsdk/test/shared-aws/queue.inflight.test.ts
```diff
@@ -68,3 +69,31 @@ test("approxSize - happy path", async () => {
...
+        Body: MESSAGE,
+      },
+    ],
+  };
+  const NO_MSG_RESPONSE = {};
+
+  sqsMock
+    .on(ReceiveMessageCommand, { QueueUrl: QUEUE_URL })
+    .resolvesOnce(ONE_MSG_RESPONSE)
+    .resolves(NO_MSG_RESPONSE);
```
Chriscbr, line 89: Nice mocking

Chriscbr (approved):
🙏

revitalbarletz:
@flyingImer can you please also enable the pop tests? we should remove the skiping keyword at the top of the file in the examples folder

flyingImer:
@revitalbarletz not sure I follow. Assuming you are referring to [examples/tests/sdk_tests/queue/pop.w](https://github.com/winglang/wing/pull/2506/files#diff-632eb7c56abd724dd815a8169b7c4d9d648617dccf6d881315278d5e5153609c), what else did I miss here?

docs/04-reference/wingsdk-api.md
flyingImer: Done: https://github.com/winglang/wing/issues/2543

PR #2531, opened 2023-05-14
fix(docs): fix mismatching sdk spec for bucket

garysassano:
I noticed that `cloud.Bucket` methods differ from those specified in the SDK Spec so I propose some changes.


docs/04-reference/wingsdk-spec.md
```diff
@@ -253,105 +253,115 @@ resource Bucket {
   /**
    * Run an inflight whenever an object is uploaded to the bucket.
    */
-  onCreate(fn: inflight (key: str) => void, opts: BucketOnUploadProps?): void;
+  onCreate(fn: IBucketEventHandler, opts?: BucketOnUploadProps): void;
```
Chriscbr, line 256: I think the main reason reason we have `IBucketEventHandler` is because TypeScript doesn't have a way to represent inflight function types. I think we should probably avoid leaking this implementation detail in Wing if possible, to me it makes more sense to keep the existing type. The existing signature is also easier to read since it doesn't require looking up what `IBucketEventHandler` means. does that make sense?

Also curious if @eladb has any thoughts

docs/04-reference/wingsdk-spec.md
```diff
@@ -253,105 +253,115 @@ resource Bucket {
...
    * Add an object to the bucket that is uploaded when the app is deployed.
    */
-  addObject(key: str, value: Blob): void;
+  addObject(key: string, body: string): void;
 
   /**
    * Upload an object to the bucket.
    */
-  inflight put(key: str, value: Blob): void;
+  inflight put(key: string, body: string): Promise<void>;
```
Chriscbr, line 281: All inflight methods are async in Wing by default, so the return type does not need to be changed here. (if you mouse over the word "put" when you have some Wing code open in your favorite IDE, we want the return type to just show `void` and not `Promise<void>`

docs/04-reference/wingsdk-spec.md
```diff
@@ -253,105 +253,115 @@ resource Bucket {
...
    * Run an inflight whenever an object is uploaded, modified, or deleted from the bucket.
    */
-  onEvent(fn: inflight (event: BucketEvent) => void, opts: BucketOnEventProps?): void;
+  onEvent(fn: IBucketEventHandler, opts?: BucketOnEventProps?): void;
 
   /**
    * Add an object to the bucket that is uploaded when the app is deployed.
    */
-  addObject(key: str, value: Blob): void;
+  addObject(key: string, body: string): void;
```
Chriscbr, line 276: Since we don't have `Blob` type yet / adding it might require more design work, removing the uses of it makes sense to me 👍 
```suggestion
  addObject(key: str, body: str): void;
```

docs/04-reference/wingsdk-spec.md
```diff
@@ -98,7 +98,7 @@ For example, a `cloud.Bucket` can emit events whenever an object is uploaded, an
 ```ts
 // wing
-bucket.onCreate(inflight (file: str) => {
+bucket.onCreate(inflight (file: string) => {
```
Chriscbr, line 101: `str` is the name of the type in Wing, so let's keep it that way throughout the doc
```suggestion
bucket.onCreate(inflight (file: str) => {
```

docs/04-reference/wingsdk-spec.md
```diff
@@ -837,13 +847,13 @@ struct ApiCorsProps {
    * The list of allowed headers.
    * @example ["Content-Type"]
    */
-  headers: Array<str>;
+  headers: Record<string, string>?;
```
Chriscbr, line 850:
```suggestion
  headers: Map<str>;
```

Chriscbr (commented):
Thanks for helping update this @garysassano :) I think there might have been some confusion because the types in this file are intended to reflect how the code will look in Wing, not necessarily how it's implemented today in TypeScript. (A lot of this is a side effect of trying to build the standard library before we had the language fully implemented!) So for now, let's use `str`  throughout the doc, since that is the name of the type in Wing.

docs/04-reference/wingsdk-spec.md
garysassano: The reason why I changed `str` to `string` is actually because there is currently a mix of both as you can see [HERE](https://github.com/winglang/wing/blob/main/docs/04-reference/wingsdk-spec.md?plain=1#L583).

So you confirm that `str` should be preferred over `string` throughout the SDK Spec?

Chriscbr: > So you confirm that `str` should be preferred over `string` throughout the SDK Spec?

Yap

docs/04-reference/wingsdk-spec.md
```diff
@@ -426,7 +436,7 @@ struct QueueProps {
    * Initialize the queue with a set of messages.
    * @default []
    */
-   initialMessages?: string[];
+   initialMessages?: str[];
```
Chriscbr, line 439:
```suggestion
   initialMessages?: Array<str>;
```

docs/04-reference/wingsdk-spec.md
```diff
@@ -338,7 +348,7 @@ resource Bucket {
   /**
    * List all objects in the bucket with the given prefix.
    */
-  inflight list(prefix: str?): Iterator<str>;
+  inflight list(prefix?: str): str[];
```
Chriscbr, line 351:
```suggestion
  inflight list(prefix?: str): Array<str>; // return an iterator in the future
```

docs/04-reference/wingsdk-spec.md
garysassano: I reverted the change for now while we wait for @eladb opinion.

Chriscbr (approved):

docs/04-reference/wingsdk-spec.md
revitalbarletz: @eladb Can you take a look?

eladb: I agree with @Chriscbr 💅

PR #2534, opened 2023-05-15
feat(vscode): show docs on hover

### LINKED ISSUE #1569: VSCode hover for methods and closures should show parameter names
### Community Note

### Feature Spec

When hovering over a symbol representing a function of some kind (method or closure), it should show the names of parameters rather than just the types.

### Use Cases

Ugliness

### Implementation Notes

_No response_

### Component

IDE Extension

eladb:
## Description

Add initial support for hover documentation for classes, functions and symbols.

Screenshot 2023-06-09 at 16 24 17

Screenshot 2023-06-09 at 16 24 35

Screenshot 2023-06-09 at 16 25 34

## Implementation Notes

Propagate the JSII docs throughout the type system. This is currently done for function signatures and parameters as well as classes.

Created a `Documented` trait and implement it for various types in order to allow polymorphic doc generation.
Removed all the spans from the synthetic syntax produced by the `ClosureTransformer` since they shadowed the actual symbols in the source file and that caused the hover visitor to ignore any statements within the transformed closure.

Rewrite HoverVisitor to find the relevant element and use the new trait to render docs.

Added `name` to `FunctionParameter` so it will be possible to display the parameter names in docs.

Added the desugaring phase to the LSP so that closure will be transformed appropriately.

## Misc

* Removed special handling of the `core.Inflight` type which no longer exists.
* Add `full_path_str` to `UserDefinedType` to make it easier to produce a nested string for lookups.
* Extended visitor to also allow subscribing to symbol visits (low level).
* Removed the name of the interface when showing closure signatures. This fixes https://github.com/winglang/wing/issues/2786.
* Do not apply spans to default initializers of classes.
* Fix #1569 by adding parameter names to function signatures.


eladb:
@MarkMcCulloh any chance you can take a look at this? I have a couple of tests that are crashing with some memory issue (see "TODO" above).

MarkMcCulloh:
@eladb  Added several changes to the PR to address the test failures, as well as a couple other bugs:

- resolve latest merge conflict
- `invalid memory` was caused by the usage of `deref()` rather than (I assume) `.as_ref()`
- `inside_class_field` now shows info, I added back in the code that was there before to handle scope tracking in classes
- Added additional special hover cases (for loops, try/catch, etc.) to make sure lookups succeed
- When compiling in the lsp, the `scope` needs to be boxed prior to type checking. Without this, lookups would fail when vars were defined in the parent due to pointer issues.
- In fold_scope, scope span was being cleared but it's needed for most lsp functions
- In completions, the highlights for which argument is current was broken due to the docs
- For primitive types, the docs should be empty until we figure that out. Otherwise the redundancy was strange (it was just printing the type name twice)

eladb:
@MarkMcCulloh 🙏🙏🙏

libs/wingc/src/lsp/hover.rs
```diff
@@ -25,26 +28,24 @@ impl<'a> HoverVisitor<'a> {
...
 		self.visit_scope(self.current_scope);
 		self.found.clone()
 	}
 
 	/// Try to look up a full path of a symbol in the current scope and if found, render the docs
 	/// associated with the symbol kind. Returns `None` if not found.
 	fn lookup_docs(&mut self, nested_str: &str, property: Option<&Symbol>) -> Option<String> {
-		let b = self.current_scope.env.borrow();
+		let current_env = self.current_scope.env.borrow();
+		let current_env = current_env.as_ref()?;
```
eladb, line 46: TIL @MarkMcCulloh thanks!

libs/wingc/src/lsp/snapshots/hovers/static_method.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: libs/wingc/src/lsp/hover.rs
+---
+contents:
+  kind: markdown
+  value: "```wing\nmy: preflight (): str\n```\n---"
```
eladb, line 6: missing `static` here?

libs/wingc/src/lsp/hover.rs
```diff
@@ -106,40 +149,132 @@ impl<'a> Visit<'a> for HoverVisitor<'a> {
...
+				if sym.span.contains(&self.position) {
+					self.found = Some((sym.span.clone(), self.lookup_docs(&sym.name, None)));
+				}
+			}
+			Reference::InstanceMember {
+				object,
+				property,
+				optional_accessor: _,
+			} => {
+				if let Some(obj_type) = self.types.get_expr_type(object) {
```
eladb, line 246: what about the case where `object` is matched?

libs/wingc/src/lsp/snapshots/hovers/static_method.snap
eladb: added

libs/wingc/src/lsp/hover.rs
eladb: checked

eladb:
@MarkMcCulloh ready for review

libs/wingc/src/parser.rs
```diff
@@ -890,14 +889,22 @@ impl<'s> Parser<'s> {
...
-					.named_children(&mut cursor)
-					.filter_map(|param_type| self.build_type_annotation(&param_type, phase).ok())
-					.collect::<Vec<TypeAnnotation>>();
+
+				let mut parameters = vec![];
+				for param_type in param_type_list_node.named_children(&mut cursor) {
+					let t = self.build_type_annotation(&param_type, phase)?;
+
+					parameters.push(FunctionParameter {
+						name: "".into(),
```
Chriscbr, line 898: Usually when a value isn't always available, it's preferable to model it in Rust using the `Option` type, e.g. `Option<String>`

libs/wingc/src/ast.rs
```diff
@@ -140,19 +141,10 @@ pub enum TypeAnnotationKind {
 	MutMap(Box<TypeAnnotation>),
 	Set(Box<TypeAnnotation>),
 	MutSet(Box<TypeAnnotation>),
-	Function(FunctionTypeAnnotation),
```
Chriscbr, line 143: What does the VS code extension show you if you hover over a variable/field with a function type, like "fn" in the following example?

```js
class A {
  fn: inflight (num): str;
}
```

MarkMcCulloh (approved):
LGTM, really excited for this change.

I pushed up a fix for completions inside inflights/test statements (they were incorrect due to the span changes in the fold thing). 
Also added a small misc change to filter out some symbols from completions.

Added a do-not-merge label to give you a chance to counter-review my changes to your branch.

libs/wingc/src/ast.rs
eladb: It shows `p1` as the name of the argument. Not ideal, but good enough I think.

eladb:
> 

Looks good!

PR #2549, opened 2023-05-16
feat(sdk): implementing `bucket.exists`

garysassano:
I noticed that, while `bucket.exists` is part of the SDK Spec, it's not currently defined in the `IBucketClient` interface. Despite this, it actually got implemented anyway (sometimes with different name) for some of the targets.

### DIFF
--- a/libs/wingsdk/src/target-sim/bucket.inflight.ts
+++ b/libs/wingsdk/src/target-sim/bucket.inflight.ts
@@ -60,7 +60,7 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     return topicClient.publish(key);
   }
 
-  private async fileExists(key: string): Promise<boolean> {
+  private async exists(key: string): Promise<boolean> {
     return fs.promises
       .access(join(this.fileDir, key))
       .then(() => true)
@@ -80,7 +80,7 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     return this.context.withTrace({
       message: `Put Json (key=${key}).`,
       activity: async () => {
-        const actionType: BucketEventType = (await this.fileExists(key))
+        const actionType: BucketEventType = (await this.exists(key))
           ? BucketEventType.UPDATE
           : BucketEventType.CREATE;
 
--- a/libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
+++ b/libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
@@ -35,6 +35,15 @@ export class BucketClient implements IBucketClient {
   }
 
   /**
+   * Check if an object exists in the bucket
+   *
+   * @param key Key of the object
+   */
+  public async exists(key: string): Promise<boolean> {
+    // WIP
+  }
+
+  /**
    * Put object into bucket with given body contents
    *
    * @param key Key of the object
--- a/libs/wingsdk/src/cloud/bucket.ts
+++ b/libs/wingsdk/src/cloud/bucket.ts
@@ -222,6 +222,13 @@ export interface BucketDeleteOptions {
  */
 export interface IBucketClient {
   /**
+   * Check if an object exists in the bucket.
+   * @param key Key of the object.
+   * @inflight
+   */
+  exists(key: string): Promise<boolean>;
+
+  /**
    * Put an object in the bucket.
    * @param key Key of the object.
    * @param body Content of the object we want to store into the bucket.
--- a/libs/wingsdk/src/shared-aws/bucket.inflight.ts
+++ b/libs/wingsdk/src/shared-aws/bucket.inflight.ts
@@ -20,6 +20,16 @@ export class BucketClient implements IBucketClient {
     private readonly s3Client = new S3Client({})
   ) {}
 
+  private async exists(key: string): Promise<boolean> {
+    const command = new ListObjectsV2Command({
+      Bucket: this.bucketName,
+      Prefix: key,
+      MaxKeys: 1,
+    });
+    const resp: ListObjectsV2CommandOutput = await this.s3Client.send(command);
+    return !!resp.Contents && resp.Contents.length > 0;
+  }
+
   public async put(key: string, body: string): Promise<void> {
     const command = new PutObjectCommand({
       Bucket: this.bucketName,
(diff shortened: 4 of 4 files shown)


tsuf239 (commented):
Hi Gary, thanks for another significant contribution! 👏 👏 👏 
Can you please add the following:
1. a unit test for each target (checking if some key `exists`, then putting something, then checking `exists` again)
2. an SDK spec test

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -34,6 +34,30 @@ export class BucketClient implements IBucketClient {
...
+    for await (const container of containers) {
+      const containerClient = blobServiceClient.getContainerClient(
+        container.name
+      );
+      const blobClient = containerClient.getBlobClient(key);
+      if (await blobClient.exists()) {
+        return true;
+      }
+    }
+    return false;
```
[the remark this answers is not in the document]
garysassano: I don't have an active Azure Subscription for trying it out. I guess it wouldn't work since you would need to edit the connection string to include your Azure account.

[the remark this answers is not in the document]
tsuf239: It's OK, let's just omit the implementation for Azure for now :)
Can you please open an issue (for completing `Bucket.exists` method on Azure) 
so we'll know what is missing?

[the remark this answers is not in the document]
tsuf239: then we'll just have the unit test (for tf-aws and sim) and the SDK spec test, and we're good to go!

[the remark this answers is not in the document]
garysassano: Ticket for finishing `tf-azure` implementation 🡆 https://github.com/winglang/wing/issues/2575

I'll add tests for `tf-aws` and `sim` soon.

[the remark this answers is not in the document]
staycoolcall911: Thanks @garysassano! 
If the Azure code wasn't tested I would go ahead and delete it from this PR for now - it's not a lot of code and I believe testing will show us we need a few tweaks anyway.

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -60,7 +60,7 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     return topicClient.publish(key);
   }
 
-  private async fileExists(key: string): Promise<boolean> {
+  public async exists(key: string): Promise<boolean> {
     return fs.promises
```
tsuf239, line 64: please wrap it in the context function:
``` 
      message: `Exists (key=${key}).`,
      activity: async () => {...}})
``` 
we may want to keep fileExists private method in order to keep one calling to `this.context.withTrace` for each public method

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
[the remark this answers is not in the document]
garysassano: @staycoolcall911 If I remove the `exists` method from the `tf-azure` target, then the tests would fail because `BucketClient` needs to implement `IBucketClient` interface.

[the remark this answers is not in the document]
staycoolcall911: You can implement and throw unsupported operation exception or something of that sort. We should have a few places with similar implementation. LMK if you need assistance locating one

garysassano (commented):
Done.

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -34,6 +34,19 @@ export class BucketClient implements IBucketClient {
...
+   * Check if an object exists in the bucket
+   *
+   * @param key Key of the object
+   */
+  public async exists(key: string): Promise<boolean> {
+    key = "none";
+    throw new Error(
+      "exists method isn't implemented yet on the current target."
+    );
+    return false;
```
Chriscbr, line 47: I think this should fix the current build error:
```suggestion
    key;
    throw new Error(
      "exists method isn't implemented yet on the current target."
    );
```

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -67,6 +67,15 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
       .catch(() => false);
   }
 
+  public async exists(key: string): Promise<boolean> {
+    return this.context.withTrace({
+      message: `Exists (key=${key}).`,
+      activity: async () => {
+        return this.objectKeys.has(key);
```
tsuf239, line 74: Can you please change this to return `this.fileExists(key)` and fileExists to return `this.objectKeys.has(key)`? this way both will keep the same behavior

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -34,6 +34,16 @@ export class BucketClient implements IBucketClient {
...
 
+  /**
+   * Check if an object exists in the bucket
+   *
+   * @param key Key of the object
+   */
+  public async exists(key: string): Promise<boolean> {
+    const blobClient = this.containerClient.getBlobClient(key);
+    return blobClient.exists();
+  }
```
tsuf239, line 45: Can you please add a unit test? 
(I can test it on Azure before the PR will be merged)

examples/tests/sdk_tests/bucket/exists.w
```diff
@@ -0,0 +1,18 @@
...
+  assert(!b.exists("test2.txt"));
+
+  b.put("test2.txt", "Bar");
+
+  assert(b.exists("test2.txt"));
+
+  b.delete("test1.txt");
+
+  assert(!b.exists("test1.txt"));
+}
```
tsuf239, line 18: Great! 👏 👏

libs/wingsdk/src/target-sim/bucket.inflight.ts
garysassano: I decided to remove `this.fileExists(key)` because redundant. Instead, I replaced all calls to either `this.fileExists(key)` or `this.objectKeys.has(key)` with `this.exists(key)` so that everything is the same across the class and every call benefits from the `withTrace`.

libs/wingsdk/test/target-tf-azure/bucket.inflight.test.ts
```diff
@@ -1,9 +1,11 @@
 import { Readable } from "stream";
+import { promiseHooks } from "v8";
```
Chriscbr, line 2: Can this import be removed? (I don't see it used anywhere)

Chriscbr (approved):
Amazing work! Thanks for adding this feature @garysassano 💪

libs/wingsdk/test/target-tf-azure/bucket.inflight.test.ts
garysassano: Thanks for catching that, it was automatically added by a VS Code plugin.

PR #2577, opened 2023-05-17
chore: workflow and dep vulnerability update

MarkMcCulloh:
In response to a security advisory from github:

- The self-mutation workflow effectively allowed for shell injection via branch name, so it was updated to avoid that
- `npm audit fix` for all packages. This updated vm2 as well as projen. The projen update needed a change to make sure it handled our `.npmrc`. Other than that the projen update itself changed several files but things look good
- In the winglang package, had to add esbuild as an explicit dev dependency. In theory, vitest already handles this, but I noticed all the optional system-specific dependencies of esbuild were removed from the lockfile. Not sure what changed, but adding esbuild explicitly doesn't hurt our dependency closure anyways


apps/jsii-docgen/.projenrc.ts
```diff
@@ -65,4 +65,6 @@ project.addFields({
   volta: rootPackageJson.volta,
 });
 
+project.npmrc.addConfig("install-links", "false");
```
eladb, line 68: Add a comment explaining why this is needed

eladb (approved):

PR #2579, opened 2023-05-18
feat(sdk): add keys/values to map interfaces

### LINKED ISSUE #2376: Improvements to `Map<T>`
### Feature Spec

* List all the keys in a map: `my_map.keys()`
* List all the values: `my_map.values()`
* Iterate over keys/values in a map: 
  ```js
  for kv in my_map {
    log(kv.key);
    log(kv.value);
  }
  ```

### Use Cases

Work with maps

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

fynnfluegge:
This PR adds keys/values/iterator interface to Map<T>.

Closes #2376 


fynnfluegge:
Could anyone please give me some hint about adding the iterator to `map.ts`. I thought about this somehow  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator?retiredLocale=de 
but have no idea how to make it JSIIable. 😅
Thanks in advance! 🙌

Chriscbr:
> Could anyone please give me some hint about adding the iterator to `map.ts`. I thought about this somehow [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator?retiredLocale=de](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator?retiredLocale=de) but have no idea how to make it JSIIable. 😅 Thanks in advance! 🙌

I think for now it makes sense to just return immutable arrays as you have in the current version of your PR. Returning an iterator can be a future improvement 👍

examples/tests/valid/container_types.w
```diff
@@ -59,10 +59,20 @@ assert(m7.size() == 3);
...
+assert(m8.values().at(0) == "a1");
+assert(m8.values().at(1) == "b1");
+assert(m8.values().at(2) == "c1");
+let m9 = MutMap<str>{"a": "a1", "b": "b1", "c": "c1"};
+assert(m9.keys().at(0) == "a");
+assert(m9.keys().at(1) == "b");
+assert(m9.keys().at(2) == "c");
+assert(m9.values().at(0) == "a1");
+assert(m9.values().at(1) == "b1");
+assert(m9.values().at(2) == "c1");
```
Chriscbr, line 75: Can we also add a few lines to show it's possible to iterate over the return value? For example:

```
for val in m9.keys() {
  assert(!val.endsWith("1"));
}
for val in m9.values() {
  assert(val.endsWith("1"));
}
```

fynnfluegge: Sure, good hint! ✅

fynnfluegge:
> > Could anyone please give me some hint about adding the iterator to `map.ts`. I thought about this somehow [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator?retiredLocale=de](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator?retiredLocale=de) but have no idea how to make it JSIIable. 😅 Thanks in advance! 🙌
> 
> I think for now it makes sense to just return immutable arrays as you have in the current version of your PR. Returning an iterator can be a future improvement 👍

Alright, I just wanted to add the feature to iterate over key/value pairs like this:
```javascript

for kv in someMap {
    log(someMap.key);
    log(someMap.value);
}
```
But have no clue how to add the iterator to the interface 🤔

Chriscbr (approved):

Chriscbr:
> Alright, I just wanted to add the feature to iterate over key/value pairs like this:
> 
> ```js
> for kv in someMap {
>     log(someMap.key);
>     log(someMap.value);
> }
> ```
> 
> But have no clue how to add the iterator to the interface 🤔

I see, gotcha. I think for now it's okay to handle that address that use case separately. It introduces at least two challenges:
1) how does the compiler know `someMap` is iterable / what logic to run while iterating over it (maybe we can treat map types as a special case when the compiler emits code for for-loops?)
2) typing `kv` (is it struct type like `MapEntry<T>`? this will be easier to implement once Wing's compiler supports generics for user-defined types)

fynnfluegge:
> I see, gotcha. I think for now it's okay to handle that address that use case separately. It introduces at least two challenges:
> 
> 1. how does the compiler know `someMap` is iterable / what logic to run while iterating over it (maybe we can treat map types as a special case when the compiler emits code for for-loops?)
> 2. typing `kv` (is it struct type like `MapEntry<T>`? this will be easier to implement once Wing's compiler supports generics for user-defined types)

Alright, then let's create a separate issue with more details. Since #2376 is closed with this PR and it already mentions the feature to iterate over key value pairs. But how to implement an iterator requires more information how to tackle that feature and the challenges I think.

PR #2580, opened 2023-05-18
fix(compiler): cannot call inflights from other inflights

### LINKED ISSUE #390: Function Declaration: call inflight closure from inflight code

### LINKED ISSUE #2236: Unable to capture a variable that is an inflight closure
### I tried this

```js
bring cloud;

resource MyResource {
  closure: inflight (str): str;

  init() {
    this.closure = inflight (s: str): str => { return "hello"; };
  }

  inflight foo() {
    this.closure("hello");
  }
}
```

### I expected this:

To be possible

### Instead, this happened

```sh
error: Cannot capture field 'closure' with non-capturable type 'inflight (str): str'
   --> .[/test.w:11:5](vscode://file//test.w:11:5)
   |
11 |     this.closure("hello");
   |     ^^^^^^^^^^^^ Cannot capture field 'closure' with non-capturable type 'inflight (str): str'
```

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

### LINKED ISSUE #2242: Unable to capture an inflight closure
### I tried this

```js
bring cloud;

resource Schedule {  
  init() {}
  on_tick(handler: inflight (): str) {
      handler();
    });
  }
}

let s = new Schedule();
s.on_tick(inflight (): str => {
  return "foo";
});
```

### I expected this:

To work

### Instead, this happened

```
handler is not a function
```

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

Chriscbr:
Fixes a limitation that prevented users from calling inflight functions that were defined in preflight. The main thrust of the strategy is to leverage the fact that inflight functions defined in preflight are compiled into classes with a single inflight method named "handle" (#2411) - we can call these "handler classes".

To allow for the possibility that values typed as inflight functions could refer to handler classes OR to plain inflight functions (that were declared in an inflight scope, so they didn't need to get converted into a handler class), we desugar inflight call expressions into `typeof <fn> === "function" ? await <fn>() : await <fn>.handle()` to support either case.

Fixes #390
Fixes #2236
Fixes #2242


tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
```diff
@@ -8,7 +8,7 @@ module.exports = function({ c }) {
     }
     async handle(msg)  {
       {
-        (await c.inc());
+        (typeof c.inc === "function" ? await c.inc() : await c.inc.handle());
```
Chriscbr, line 11: Open to ideas/suggestions on how to make this less verbose

examples/tests/valid/inflights_calling_inflights.w
```diff
@@ -0,0 +1,41 @@
+bring cloud;
+
+let global_bucket = new cloud.Bucket();
```
eladb, line 3: Camel case maybe?

examples/tests/valid/schedule.w
```diff
@@ -0,0 +1,14 @@
...
+    new cloud.Function(inflight () => {
+      handler();
+    });
+  }
+}
+
+let s = new Schedule();
+s.onTick(inflight (): str => {
+  return "foo";
+});
```
eladb, line 14: Maybe add a `test`

libs/wingc/src/type_check.rs
```diff
@@ -773,6 +792,24 @@ impl TypeRef {
...
+			false
+		}
+	}
+
+	/// Returns whether the type is a preflight class with an inflight method named "handle"
+	pub fn is_handler_resource(&self) -> bool {
+		if let Type::Resource(ref class) = **self {
+			return class
+				.methods(true)
+				.any(|(name, type_)| name == "handle" && type_.is_inflight_function());
```
eladb: const

tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
eladb, line 11: Curious- do we really need a runtime check? don't we know at compile time if this is supposed to be a function or a handler?

eladb (changes requested):

tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
Chriscbr: Perhaps. In an expression like `a.b.c()`, if the last reference component (`c`) is a preflight member of its parent (`a.b`), then we know it's going to be a class instance in JS, while if the last variable component is an inflight member (for example, an inflight method of a class, or an inflight class member with a function type), then it will be plain old JS function. But I'm not sure there isn't some case I'm missing here. In any case I feel like we can leave this as a future optimization

examples/tests/valid/inflights_calling_inflights.w
Chriscbr: Oops. Old habits die hard 🙂

examples/tests/valid/schedule.w
Chriscbr: Testing this was less clear than expected because it's not possible to write the full SDK Schedule resource in Wing yet, so I rewrote this into a different example that still validates you can capture inflight closures in `doubler.w`

tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
eladb: Can you check if this is a low hanging fruit? Feels like this test should be easy to make at the jsification stage. No?

examples/tests/valid/doubler.w
yoav-steinberg, line 1: I'm missing some comments about what these tests are

libs/wingc/src/type_check.rs
```diff
@@ -166,6 +166,8 @@ pub enum Type {
 pub const CLASS_INIT_NAME: &'static str = "init";
 pub const CLASS_INFLIGHT_INIT_NAME: &'static str = "$inflight_init";
 
+pub const HANDLE_METHOD_NAME: &'static str = "handle";
```
yoav-steinberg, line 169: Should we make this `$handle` so users can't create their own "handler" classes? I'm worried this name will start to collide with stuff, "handler" is pretty common?

libs/wingc/src/type_check.rs
```diff
@@ -1440,7 +1471,27 @@ impl<'a> TypeChecker<'a> {
 
 				// Make sure this is a function signature type
 				let func_sig = if let Some(func_sig) = func_type.as_function_sig() {
-					func_sig
+					func_sig.clone()
+				} else if let Some(res) = func_type.as_resource() {
+					// return the signature of the "handle" method
+					let lookup_res = res.get_env().lookup_ext(&HANDLE_METHOD_NAME.into(), None);
```
yoav-steinberg, line 1477: might be able to make this cleaner with `res.get_method(...)`

libs/wingc/src/type_check.rs
Chriscbr: I feel like it's pretty common to use well-defined / common names for important language features. For example, consider how "self" is a special argument when it's the name of the first parameter of a method in Rust. We're basically treating "handle" with the same level of importance here.

examples/tests/valid/doubler.w
Chriscbr: Added

libs/wingc/src/type_check.rs
Chriscbr: Done

tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
Chriscbr: @eladb After some pondering, here's the counter example showing why we can't [always] statically determine whether it's a function or handler. I'll add it as a unit test:

```js
bring cloud;

class Foo {
  inflight1: inflight (): num;
  init() {
    // here is an inflight function created during preflight
    this.inflight1 = inflight (): num => {
      return 1;
    };
  }

  inflight inflight2: inflight (): num;
  inflight init() {
    // here is an inflight function created during inflight
    this.inflight2 = inflight (): num => {
      return 2;
    };
  }

  inflight makeFn(x: bool): inflight (): num {
    if (x == true) {
      return this.inflight1;
    } else {
      return this.inflight2;
    }
  }

  inflight callFn(x: bool): num {
    // partialFn could be an inflight function created during preflight or inflight,
    // so we have to code-generate this to work for both cases
    let partialFn = this.makeFn(x);
    return partialFn();
  }
}

let foo = new Foo();

test "calling inflight functions created in different phases" {
  assert(foo.callFn(true) == 1);
  assert(foo.callFn(false) == 2);
}
```

examples/tests/valid/calling_inflight_variants.w
yoav-steinberg, line 1: beautiful :sparkles:

libs/wingc/src/jsify.rs
```diff
@@ -1486,22 +1489,31 @@ impl<'ast> Visit<'ast> for FieldReferenceVisitor<'ast> {
...
+						),
+						span: Some(node.span.clone()),
+					});
+
+					return;
+				}
+
+				// if this reference refers to an inflight function or handler resource,
+				// we need to give permission to the "handle" operation
+				if v.type_.is_inflight_function() || v.type_.is_handler_resource() {
```
yoav-steinberg, line 1510: I think I understand why we need to add the "handle" operation in casse it's a handler resource. But if it's a plain old inflight function, then why add this permission?

libs/wingc/src/type_check.rs
yoav-steinberg: I think that if we currently name it something that the user can never create on their own (`$handle`) then we might be able to discuss/change this later. By keeping this `handle` we're kind of creating a language feature (like `self`).
I'd like some more opinions here: @eladb?

libs/wingc/src/type_check.rs
```diff
@@ -773,6 +794,24 @@ impl TypeRef {
 		}
 	}
 
+	pub fn is_resource(&self) -> bool {
+		if let Type::Resource(_) = **self {
+			true
+		} else {
+			false
+		}
```
yoav-steinberg, line 802: Perhaps change to this and also all the other "is_XXX" methods here.
```suggestion
		matches!(**self, Type::Resource(_))
```

yoav-steinberg (approved):

examples/tests/valid/calling_inflight_variants.w
```diff
@@ -0,0 +1,41 @@
...
+  inflight inflight2: inflight (): num;
+  inflight init() {
+    // here is an inflight function created during inflight
+    this.inflight2 = inflight (): num => {
+      return 2;
+    };
+  }
+
+  inflight makeFn(x: bool): inflight (): num {
+    if (x == true) {
```
Chriscbr, line 21:
```suggestion
    if x == true {
```

libs/wingc/src/jsify.rs
Chriscbr: We need to add the "handle" operation in both situations in case a handler resource is assigned to a field of an inflight closure type. For example, in this snippet:

```js
class MyResource {
  closure: inflight (str): str;

  init() {
    this.closure = inflight (s: str): str => {
      globalBucket.list();
      return "hello";
    };
  }

  inflight foo(): str {
    return this.closure("anything");
  }
}
```

If I change this line in the `jsify.rs` so we only add the "handle" operation if it's a handler resource, then the compiler no longer annotates `foo` with the "handle" permission, causing the test to fail.

tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
eladb: Thanks for the clarification. It might still be possible to optimize in cases we know for certain, but that's something we can do later on.

libs/wingc/src/type_check.rs
eladb: I am not 100% sure whether `handle` is the right now for the method name. In Java it could be any name and the generated class will take the name based on the interface it is implementing.

But, I'd like users to be able to simply pass an object that implements the interface as an alternative to passing a closure.

Bottom line, My preference is to leaving it `handle` for now.

eladb (approved):

eladb (approved):

libs/wingc/src/type_check.rs
Chriscbr: Sounds good 👍

PR #2584, opened 2023-05-18
fix: generated inflight classes are visible by default in Wing Console

Chriscbr:
Fixes a regression to the Wing Console experience that was brought to my attention by @ainvoner 🙂

#2411 changed our compiler so inflight closures defined in preflight scopes are turned into preflight classes (which are added to the construct tree)  -- however it doesn't make sense to show them in the console by default, since they represent ephemeral pieces of code, and not logical resources or services.

### DIFF
--- a/libs/wingc/src/closure_transform.rs
+++ b/libs/wingc/src/closure_transform.rs
@@ -5,7 +5,7 @@ use indexmap::IndexMap;
 use crate::{
 	ast::{
 		ArgList, Class, ClassField, Expr, ExprKind, FunctionDefinition, FunctionParameter, FunctionSignature, Initializer,
-		Phase, Reference, Scope, Stmt, StmtKind, Symbol, TypeAnnotation, TypeAnnotationKind, UserDefinedType,
+		Literal, Phase, Reference, Scope, Stmt, StmtKind, Symbol, TypeAnnotation, TypeAnnotationKind, UserDefinedType,
 	},
 	fold::{self, Fold},
 };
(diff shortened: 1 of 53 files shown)


tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
```diff
@@ -393,6 +393,7 @@ class $Root extends $stdlib.std.Resource {
       constructor(scope, id, ) {
         super(scope, id);
         this._addInflightOps("handle");
+        this.display.hidden = true;
```
Chriscbr, line 396: We can probably "shift right" some of this logic into the SDK if this gets too bloated

libs/wingc/src/closure_transform.rs
```diff
@@ -195,6 +195,33 @@ impl Fold for ClosureTransformer {
 				// we need to set this to false.
 				new_func_def.is_static = false;
 
+				let class_init_body = vec![Stmt {
```
yoav-steinberg, line 198: Comment of what we're doing here `this.display.hidden = true`?
Also kind of feels like we can benefit from calling tree-sitter+parser to generate this ast from a string. Probably felt weird coding like this...

tools/hangar/__snapshots__/test_corpus/add_consumer.w_compile_tf-aws.md
yoav-steinberg: We inherit from a `HiddenResource` for example.. But I guess this is ok.

yoav-steinberg (approved):

libs/wingc/src/closure_transform.rs
Chriscbr: It did feel quite weird, yeah :)

Hmm, it didn't occur to me that maybe we could generate this AST from a string directly inside this transform... let me take a look into this

Chriscbr: Added a comment.

I'll probably leave the other suggested changes out of this PR because I think it'll require refactoring code in the parser so that AST nodes are created with the correct spans. The other caveat is even if we could generate an AST from a string, it's not clear doing this in the closure transform makes sense considering it would require the compiler to do more work (lexing and parsing a string) when compared to creating a set of structs directly.

PR #2587, opened 2023-05-19
fix(sdk): aws funtion invoke returns the correct type

### LINKED ISSUE #2559: Function invoke test works on sim but not on tf-aws
### I tried this:

Run the invoke function test with tf-aws target breaks, but works on sim

### This happened:

image

### I expected this:

Works

### Is there a workaround?

I don't know

### Component

Compiler

### Wing Version

0.17.7

### Wing Console Version

latest

### Node.js Version

v18.12.1

### Platform(s)

Linux

### Anything else?

_No response_

### Community Notes

marciocadev:
Closes #2559 


libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -27,6 +27,6 @@ export class FunctionClient implements IFunctionClient {
     if (!response.Payload) {
       return "";
     }
-    return toUtf8(response.Payload);
+    return JSON.parse(toUtf8(response.Payload));
```
Chriscbr, line 30: If we just add this line (without changing the signature of `invoke` in `cloud/function.ts`) does the fix still work?

marciocadev: To be true, I made this change to give the possibility to return any type, not only string

As the documentation of cloud.Function it's declare as any

marciocadev: To be honest, I made this change after reviewing the documentation (where the invoke function returns a JSON), and this modification allows for the return of any type of value.

Chriscbr (approved):
🙏

PR #2591, opened 2023-05-19
fix(compiler): show compiler error if inflight fields are not initialized

### LINKED ISSUE #2582: Missing compiler error if inflight fields are not initialized
### I tried this:

```js
bring cloud;

class Foo {
  inflight x: str;
}

let foo = new Foo();

test "access x" {
  log(foo.x);
}
```

### This happened:

Test runs but it prints "undefined"

### I expected this:

Compile time error saying that "x" has not been initialized in the inflight constructor/initializer

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

0.17.7

### Wing Console Version

_No response_

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

marciocadev:
Closes #2582 


examples/tests/invalid/class.w
```diff
@@ -13,4 +13,11 @@ class C3 {
...
-}
\ No newline at end of file
+}
+
+class C4 {
+    inflight var x: num;
+//               ^ "x" is not initialized
+  inflight y: str;
+//         ^ "y" is not initialized
+}
```
Chriscbr, line 23: style: consistent spacing
```suggestion
  inflight var x: num;
//               ^ "x" is not initialized
  inflight y: str;
//         ^ "y" is not initialized
}
```

examples/tests/valid/resource.w
```diff
@@ -7,6 +7,7 @@ class Foo {
 
   init() {
     this.c = new cloud.Counter();
+    this.inflightField = 1;
```
Chriscbr, line 10: Remove this line (`inflightField` is initialized inside `inflight init()` already)

examples/tests/valid/resource_captures.w
```diff
@@ -42,11 +42,11 @@ class MyResource {
   extBucket: cloud.Bucket;
   extNum: num;
 
-
   // Needs to be var since we don't support inflight inits yet.
   inflight var inflightField: num;
 
   init(externalBucket: cloud.Bucket, externalNum: num) {
+    this.inflightField = 1;
```
Chriscbr, line 49:
```suggestion
  inflight inflightField: num;
  
  inflight init() {
    this.inflightfield = 1;
  }

  init(externalBucket: cloud.Bucket, externalNum: num) {
```

libs/wingc/src/type_check.rs
```diff
@@ -2544,8 +2544,8 @@ impl<'a> TypeChecker<'a> {
 		let initialized_fields = visit_init.fields;
 
 		for field in fields.iter() {
-			// inflight or static fields cannot be initialized in the initializer
-			if field.phase == Phase::Inflight || field.is_static {
+			// static fields cannot be initialized in the initializer
+			if field.is_static {
```
Chriscbr, line 2548: Undo

Chriscbr (changes requested):

examples/tests/valid/class.w
```diff
@@ -27,9 +27,22 @@ let c3 = new C3(1, 2);
...
-assert(C4.m() == 1);
\ No newline at end of file
+assert(C4.m() == 1);
+
+// class with infligth field
+class C5 {
+  inflight var x: num;
+  init(a: num) {
+    this.x = a;
+  }
```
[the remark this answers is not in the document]
Chriscbr: Oops, I didn't the change you made to the type checker. Let me know if my feedback about inflight fields having to be initialized in `inflight init() {}` makes sense (no need to open a new bug) 😅

libs/wingc/src/type_check.rs
marciocadev: I understand now, sorry, I don't notice the `inflight init` constructor 

I will make some changes to capture the specific phase initialization

marciocadev: @Chriscbr Question: inflight var fields needs to be initialized ?

marciocadev: @Chriscbr 

Do the `inflight var` fields need to be initialized?

Chriscbr: Yes 👍

examples/tests/invalid/class.w
```diff
@@ -13,4 +15,34 @@ class C3 {
...
+  
+  init() {
+    this.x = 1;
+//       ^ "x" cannot be initialized in the Preflight initializer
+  }
+
+  inflight init() {
+    this.y = 1;
+//       ^ "y" cannot be initialized in the Inflight initializer
+  }
```
Chriscbr, line 47: Nice!

examples/tests/valid/class.w
```diff
@@ -1,3 +1,5 @@
+bring cloud;
+
```
Chriscbr, line 2: Was this needed? (just curious if we have some hidden dependency on cloud somewhere)

examples/tests/valid/resource_captures.w
```diff
@@ -42,9 +42,10 @@ class MyResource {
   extBucket: cloud.Bucket;
   extNum: num;
 
-
-  // Needs to be var since we don't support inflight inits yet.
   inflight var inflightField: num;
```
Chriscbr, line 45:
```suggestion
  inflight inflightField: num;
```

libs/wingc/src/type_check.rs
```diff
@@ -2577,26 +2591,44 @@ impl<'a> TypeChecker<'a> {
...
+
 		for field in fields.iter() {
+			let result = initialized_fields.iter().find(|&s| &s.name == &field.name.name);
 			// inflight or static fields cannot be initialized in the initializer
-			if field.phase == Phase::Inflight || field.is_static {
+			if field.phase == forbidden_phase || field.is_static {
+				if let Some(result) = result {
+					self.type_error(TypeError {
+						message: format!(
+							"\"{}\" cannot be initialized in the {} initializer",
```
Chriscbr, line 2620: Note "inflight" and "preflight" should be lowercase in the middle of a sentence

libs/wingc/src/type_check.rs
```diff
@@ -2577,26 +2591,44 @@ impl<'a> TypeChecker<'a> {
...
+		} else {
+			("Preflight", Phase::Inflight)
+		};
+
 		for field in fields.iter() {
+			let result = initialized_fields.iter().find(|&s| &s.name == &field.name.name);
 			// inflight or static fields cannot be initialized in the initializer
-			if field.phase == Phase::Inflight || field.is_static {
+			if field.phase == forbidden_phase || field.is_static {
+				if let Some(result) = result {
```
Chriscbr, line 2617: I would shuffle this around a little bit to avoid some additional nesting:

```rust
let matching_field = initialized_fields.iter().find(|&s| &s.name == &field.name.name);
if field.phase != forbidden_phase && !field.is_static {
	continue;
}

if let Some(matching_field) = matching_field {
  self.type_error(...);
} else {
  self.type_error(...);
}
```

libs/wingc/src/type_check.rs
```diff
@@ -2577,26 +2591,44 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
-	/// Validate if the fields of a class are initialized in the constructor (init).
+	/// Validate if the fields of a class are initialized in the preflight constructor (init).
```
Chriscbr, line 2594:
```suggestion
	/// Validate if the fields of a class are initialized in the constructor (init) according to the given phase.
	/// For example, if the phase is preflight, then all non-static preflight fields must be initialized
	/// and if the phase is inflight, then all non-static inflight fields must be initialized.
```

Chriscbr (changes requested):
Looking good, I've got just a few more comments but I think we're almost there!

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -138,24 +138,73 @@ exports[`capture_reassignable.w > stderr 1`] = `
...
+   |
+45 |     this.y = 1;
+   |          ^ \\"y\\" cannot be initialized in the Inflight initializer
+
+
+error: \\"x\\" cannot be initialized in the Preflight initializer
+   --> ../../../examples/tests/invalid/class.w:40:10
+   |
+40 |     this.x = 1;
+   |          ^ \\"x\\" cannot be initialized in the Preflight initializer
```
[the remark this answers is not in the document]
marciocadev: I believe it's the way the Visitor patter work to catch error

Don't create errors in order

examples/tests/valid/class.w
marciocadev: no need

libs/wingc/src/type_check.rs
marciocadev: @Chriscbr the `{} cannot be initialized in the {} initializer` needs before the `continue`

What I can do is remove the `if let Some(matching_field) = matching_field { ... }` and create a function, then I can call when is needed

Chriscbr: Ah gotcha. I'm okay with leaving this as is just as well 👍

marciocadev: @Chriscbr to create the `{} cannot be initialized in the {} initializer` I need to create the validation if matching_field exist before the `continue`

This change breaks the validation

tools/hangar/__snapshots__/invalid.ts.snap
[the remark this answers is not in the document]
marciocadev: Issue open

https://github.com/winglang/wing/issues/2637

Chriscbr, line 194: "inflight" and "preflight" should be lowercase in these error messages

examples/tests/valid/nil.w
```diff
@@ -3,7 +3,7 @@ bring cloud;
 class Foo {
   inflight var optionalVar: str?;
   
-  init() {
+  inflight init() {
```
staycoolcall911, line 6: @marciocadev, @hasanaburayyan - curious why the `inflight` keyword was added? Correct me if I'm wrong, but I think it wasn't necessary (but not harmful too) since we can init inflight fields also  in preflight init, right?

marciocadev: @staycoolcall911 from what I understood from issue #2582, inflight fields should be initialized in inflight initializers, but my understanding may be wrong. However, if that's the case, I don't understand why the initializer could receive the "inflight" prefix.

staycoolcall911: I might be wrong as well :)
My understanding is that inflight fields could be initialized in either preflight or inflight init.
They must be already initialized in inflight after init ended.

yoav-steinberg: We aren't supposed to be able to access inflight members (fields or methods) in preflight. So I think @marciocadev is correct here.

PR #2610, opened 2023-05-21
chore(docs): updating contributing docs

garysassano:
Wing SDK stopped using prebuilt CDKTF providers from libraries like `@cdktf/provider-aws` and the classes are now generated inside the SDK's `src/.gen` folder. Because of this, you need to run the `cdktf get` command inside `libs/wingsdk` at least once, otherwise you may get import errors saying that the modules do not exist.

### DIFF
--- a/docs/06-contributors/020-development.md
+++ b/docs/06-contributors/020-development.md
@@ -28,6 +28,7 @@ Installation:
 git clone https://github.com/winglang/wing
 cd wing
 npm install
+(cd libs/wingsdk && cdktf get)
 ```
 
 :::note Nx Commands


docs/06-contributors/020-development.md
```diff
@@ -28,6 +28,7 @@ Installation:
 git clone https://github.com/winglang/wing
 cd wing
 npm install
+(cd libs/wingsdk && cdktf get)
```
MarkMcCulloh, line 31: `cdktf get` is run during the sdk build. I understand what you mean about the missing types though, so what do you think about suggesting to users that they run the full-repo build command as part of their setup: 

```suggestion
npm run build
```

This has the added benefit of ensuring they're ready to do development.

as an aside: `cdktf` also probably won't be available globally for most users anyways.

MarkMcCulloh (changes requested):

docs/06-contributors/020-development.md
garysassano: That's actually already specified [HERE](https://github.com/winglang/wing/blob/f4601823bf1a5ba464d31b8f97dfdbbaf236c1da/docs/06-contributors/020-development.md?plain=1#L63).

I get error when running `npm run build` if I didn't generate the CDKTF providers using `cdktf get` first, but I guess that's an issue on my side then.

PR #2612, opened 2023-05-21
feat(sdk): adding awscdk test command

marciocadev:
This pull request adds the `awscdk` target to the test command. (Everyone :heart:'s a long-running test, right?)

You can run a cdk test with this steps:
* set the CDK_STACK_NAME environment variable
* run tests with awscdk target

```bash
$ export CDK_STACK_NAME=test-my-cdk-stack
$ wing test -t awscdk app.w
```

Two things should be noted. In the ENV_WING_TEST_RUNNER_FUNCTION_ARNS variable, I had to remove the underscore because the CloudFormation output doesn't accept special characters. And to capture the ARN of the test function, I had to create an output.json file in the directory of the created object. But don't worry, I delete the file in the end when the stack is being removed.


apps/wing/src/commands/test.ts
```diff
@@ -153,6 +157,71 @@ async function testSimulator(synthDir: string) {
   }
 }
 
+async function testAwsCdk(synthDir: string): Promise<sdk.cloud.TestResult[]> {
+  const stackName = process.env.CDK_STACK_NAME
+  if (!stackName) {
+    throw new Error(
+      "In order to successfully deploy using AWS-CDK, it is essential to define the environment variable CDK_STACK_NAME."
+    );
+  }
```
Chriscbr, line 166: Is CDK_STACK_NAME needed for the actual deployment, or is it just needed for running `wing compile`?

I think `wing compile -t awscdk` already gives you an error if you do not specify CDK_STACK_NAME, so this check might be redundant.
```suggestion
```

apps/wing/src/commands/test.ts
```diff
@@ -153,6 +157,71 @@ async function testSimulator(synthDir: string) {
...
+
+  if (!isAwsCdkInstalled(synthDir)) {
+    throw new Error(
+      "AWS-CDK is not installed. Please install AWS-CDK to run tests in the cloud."
```
Chriscbr, line 170:
```suggestion
      "AWS-CDK is not installed. Please install AWS-CDK to run tests in the cloud (npm i aws-cdk)."
```

libs/wingsdk/src/target-awscdk/function.ts
```diff
@@ -52,6 +53,26 @@ export class Function extends cloud.Function {
     this.arn = this.function.functionArn;
   }
 
+  /** @internal */
+  public _bind(host: IInflightHost, ops: string[]): void {
+    if (!(host instanceof Function)) {
+      throw new Error("functions can only be bound by tfaws.Function for now");
```
Chriscbr, line 59:
```suggestion
      throw new Error("functions can only be bound by awscdk.Function for now");
```

libs/wingsdk/src/target-awscdk/test-runner.ts
```diff
@@ -0,0 +1,92 @@
+import { CfnOutput, Lazy } from "aws-cdk-lib";
+import { Construct } from "constructs";
+import { Function as AwsFunction } from "./function";
+import * as cloud from "../cloud";
+import * as core from "../core";
+import { IInflightHost } from "../std";
+
+const OUTPUT_TEST_RUNNER_FUNCTION_ARNS = "WINGTESTRUNNERFUNCTIONARNS";
```
Chriscbr, line 8: Let's make this name more readable
```suggestion
const OUTPUT_TEST_RUNNER_FUNCTION_ARNS = "WingTestRunnerFunctionArns";
```

apps/wing/src/commands/test.ts
```diff
@@ -153,6 +158,65 @@ async function testSimulator(synthDir: string) {
   }
 }
 
+async function testAwsCdk(synthDir: string): Promise<sdk.cloud.TestResult[]> {
+
+  if (!isAwsCdkInstalled(synthDir)) {
+    throw new Error(
+      "AWS-CDK is not installed. Please install AWS-CDK to run tests in the cloud (npm i aws-cdk)."
+    );
+  }
```
Chriscbr, line 167: When I tried this out (without CDK installed), I didn't see the error message and instead received:

```
✔ Compiling to awscdk...
⠋ cdk deploynode:internal/errors:867
  const err = new Error(message);
              ^

Error: Command failed: cdk version
/bin/sh: cdk: command not found

    at ChildProcess.exithandler (node:child_process:419:12)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1091:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:513:28)
    at Pipe.<anonymous> (node:net:322:12) {
  code: 127,
  killed: false,
  signal: null,
  cmd: 'cdk version',
  stdout: '',
  stderr: '/bin/sh: cdk: command not found\n'
}

Node.js v18.14.1
```

You might need to try/catch inside of `isAwsCdkInstalled` in order to show the error message correctly.

apps/wing/src/commands/test.ts
```diff
@@ -153,6 +158,65 @@ async function testSimulator(synthDir: string) {
   }
 }
 
+async function testAwsCdk(synthDir: string): Promise<sdk.cloud.TestResult[]> {
+
+  if (!isAwsCdkInstalled(synthDir)) {
+    throw new Error(
+      "AWS-CDK is not installed. Please install AWS-CDK to run tests in the cloud (npm i aws-cdk)."
```
Chriscbr, line 165: My bad -- I think it only works when it is globally installed (otherwise aws-cdk will not be on your PATH), so let's recommend that instead

```suggestion
      "AWS-CDK is not installed. Please install AWS-CDK to run tests in the cloud (npm i -g aws-cdk)."
```

Chriscbr (approved):

Chriscbr:
I tried out your PR, and the new test capability worked really well! The only real issue I had was that when I ran a test with buckets, after my tests passed, CloudFormation wouldn't delete the buckets since they weren't empty. 😅 We might want to add something for the AWSCDK target that is a bit like this piece of code:

https://github.com/winglang/wing/blob/64a78dac2460126989ed585d6b8a3a448a292c2d/libs/wingsdk/src/target-tf-aws/bucket.ts#L174

Otherwise looks good to me 👍

apps/wing/src/commands/test.ts
marciocadev: Done, thanks @Chriscbr

marciocadev:
> I tried out your PR, and the new test capability worked really well! The only real issue I had was that when I ran a test with buckets, after my tests passed, CloudFormation wouldn't delete the buckets since they weren't empty. 😅 We might want to add something for the AWSCDK target that is a bit like this piece of code:
> 
> https://github.com/winglang/wing/blob/64a78dac2460126989ed585d6b8a3a448a292c2d/libs/wingsdk/src/target-tf-aws/bucket.ts#L174
> 
> Otherwise looks good to me 👍

Good catch, I have fixed it now.

PR #2615, opened 2023-05-22
fix(sdk): `vm2` is not good enough

### LINKED ISSUE #2059: Cannot use `await import()` to import ESM-only modules in inflight

### LINKED ISSUE #2084: Cannot use `ngrok` in extern
### I tried this

`npm i ngrok`

```js
// repro.w
bring cloud;

resource Utils {
  init() {}
  extern "./repro.js" inflight start_ngrok(): str;
}

let utils = new Utils();

  utils.start_ngrok();
}) as "test";
```

```
// repro.js
exports.start_ngrok = async function() {
    const ngrok = require('ngrok');
    const url = await ngrok.connect(4567);
    console.log(`ngrok url: ${url}`);
    return url;
}
```

### I expected this:

A url to be printed

### Instead, this happened

```
wing test src/repro.w
✔ Compiling to sim...
fail ┌ repro.wsim » root/env0/test
     │ TypeError [ERR_INVALID_ARG_TYPE]: The "buf" argument must be an instance of ArrayBuffer or ArrayBufferView. Received an instance of Object
     │     at new NodeError (node:internal/errors:399:5)
     │     at Object.randomFillSync (node:internal/crypto/random:121:11)
     │     at ReadOnlyHandler.apply (/Users/chrisr/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/node_modules/vm2/lib/bridge.js:485:11)
     │     at rng (/Users/chrisr/dev/wing-webhook/node_modules/uuid/dist/rng.js:18:21)
     │     at Object.v4 (/Users/chrisr/dev/wing-webhook/node_modules/uuid/dist/v4.js:17:63)
     │     at connectRetry (/Users/chrisr/dev/wing-webhook/node_modules/ngrok/index.js:27:40)
     │     at Object.connect (/Users/chrisr/dev/wing-webhook/node_modules/ngrok/index.js:23:10)
     │     at async exports.start_ngrok (/Users/chrisr/dev/wing-webhook/src/repro.js:3:17)
     │     at async Handler.handle (/Users/chrisr/dev/wing-webhook/src/target/test/repro.wsim/.wing/test_c8c85f80.js:11:4)
     └     at async exports.handler (/Users/chrisr/dev/wing-webhook/src/target/test/repro.wsim/.wing/test_c8c85f80.js:2:10)
```

### Is there a workaround?

Shell out to ngrok by calling `process.execSync` in the extern code instead

### Component

Compiler

### Wing Version

0.11.2

### Wing Console Version

_No response_

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

If I try calling the extern code outside of the Wing simulator, it works fine

Screenshot 2023-04-14 at 4 55 02 PM

Based on the error message it looks like some kind of issue with vm2

### Community Notes

### LINKED ISSUE #2088: Global node package URLSearchParams not defined
### I tried this

```ts
bring cloud;
class TestApi {
  extern "./nodeHelpers.js" static inflight _encodeBody(): str;

  init() { 
  }
  inflight encodedBody(): str {
    return TestApi._encodeBody();
  }

   
}

let testApi = new TestApi();

  log("------ test: find ------");
    let response = testApi.encodedBody();
  assert(response == "helloWorld");
}) as "test: handle urlencoded body";
```
nodeHelpers.js
```js
exports._encodeBody = () => {
  const data = new URLSearchParams({
    wing: "fly",
  }).toString();
  return data;
};
```

### I expected this:

To get a Urlencoded version of the object.

### Instead, this happened

```bash
fail ┌ api.wsim » root/env0/test: handle urlencoded body
     │ ------ test: find ------
     │ ReferenceError: URLSearchParams is not defined
     │     at exports._encodeBody (<path>/nodeHelpers.js:15:16)
     │     at TestApi._encodeBody (<path>/target/test/api.wsim/.wing/clients/TestApi.inflight.js:5:130)
     │     at TestApi.encodedBody (<path>/target/test/api.wsim/.wing/clients/TestApi.inflight.js:9:29)
     │     at Handler.handle (<path>/target/test/api.wsim/.wing/test__handle_urlencoded_body_c89b2f3c.js:12:35)
     └     at exports.handler (<path>/target/test/api.wsim/.wing/test__handle_urlencoded_body_c89b2f3c.js:26:5)
```

### Is there a workaround?

```js
const url = require("url");
exports._encodeBody = () => {
  const data = new url.URLSearchParams({
    wing: "fly",
  }).toString();
  return data;
};
```

### Component

Compiler

### Wing Version

0.11.4

### Wing Console Version

_No response_

### Node.js Version

v18.13.0

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

eladb:
The `vm2` module which used to execute inflight functions within the simulator was not cutting it. The sandbox has limitations such as not being able to handle ESM imports, unable to map certain global types due to the vm2 proxy behavior, and more.

To address this, we use `esbuild` to bundle the function's code *at runtime* (JIT). This is done during the initialization of the simulator cloud function and then reused across invokes. It will likely increase the simulator app start time by a little but this is not something we optimize for at the moment.

### Implementation Notes

- esbuild does not bundle any node.js sdk modules, so we map `require` into the VM to allow that.
- Since we use `esbuild` through wasm, we needed the `dom` lib in `tsconfig.test.json`.
- To allow testing ESM imports, we symlink the SDK's `node_modules` directory into the temporary directory created for each test app.
- Unfortunately, modules that use `__dirname` and/or `__filename` cannot be bundled and therefore are not supported. We emit an explicit error when these are used.

### Misc

- Fixed a bug in serialization of null objects (`typeof(null)` in javascript is surprisingly an `"object"`).

### Related

- Resolves #2059
- Resolves #2088
- #2084 is not fully resolved because it cannot be bundled due to the fact that it uses `__dirname` internally to locate the ngrok binary.


libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -40,24 +58,24 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
-            context: "sandbox", // require inside the sandbox (addresses #1871)
-          },
-          sandbox: {
-            $simulator: this.context,
-          },
-          env: {
-            ...process.env,
-            ...this.env,
+        if (!this.bundle) {
+          throw new Error("unable to find bundle (function not initialized?)");
```
hasanaburayyan, line 62: Would it make sense to check the budled path actually exists rather than just checking the instance variable is set? Incase there was some wonky-sauce with the bundle during init()?

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -78,12 +96,94 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
+    // bundling there as well).
+    const forbidGlobal = (name: string) => {
+      Object.defineProperty(context, name, {
+        get: () => {
+          throw new Error(
+            `${name} cannot be used within bundled cloud functions`
+          );
+        },
+      });
+    };
```
hasanaburayyan, line 179: 😎

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -78,12 +96,94 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
+
+    forbidGlobal("__dirname");
+    forbidGlobal("__filename");
```
hasanaburayyan, line 182: Should we open a tracking issue for supporting these globals at somepoint?

hasanaburayyan (commented):
😎 LGTM ill let others take a look though.

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -18,6 +21,8 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
   private readonly context: ISimulatorContext;
   private readonly timeout: number;
 
+  private bundle?: string;
```
Chriscbr, line 24: Docstring would be nice here (name isn't self-descriptive)

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -40,24 +58,24 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
           },
-          timeout: this.timeout,
-        });
 
-        // see https://github.com/patriksimek/vm2/blob/master/lib/nodevm.js#L89
+          env: this.env,
+        };
+
+        const sandboxConsole: any = {};
         const levels = [
```
Chriscbr, line 79: Maybe consider removing "dir" and "trace" from this list (or add a TODO to handle them differently) -- it looks like we changing the behavior of the functions so that they no longer emit a list of properties or a stack trace respectively, which might be surprising

libs/wingsdk/src/target-sim/function.inflight.ts
Chriscbr, line 182: Could we have `__dirname` and `__filename` return the directory/filename of the final bundle path produced by esbuild (assuming we're bundling everything into a single file)?

Chriscbr, line 182: > Unfortunately, modules that use __dirname and/or __filename cannot be bundled and therefore are not supported. We emit an explicit error when these are used.

We're using __dirname / __filename a few places in the SDK, I'm curious to understand how those code paths are still supported (I'd have expected those usages might have to be removed)? Are those usages okay because they're only in preflight code?

libs/wingsdk/test/util.ts
```diff
@@ -88,6 +88,9 @@ export function directorySnapshot(initialRoot: string) {
   const visit = (root: string, subdir: string, prefix = "") => {
     const files = readdirSync(join(root, subdir));
     for (const f of files) {
+      if (f === "node_modules") {
+        continue;
+      }
```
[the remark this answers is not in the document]
eladb: Done

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: Perhaps it makes sense to set `__filename` to point to the bundle file, but I think it will likely be a bad idea to make `__dirname` point to the bundle file directory because in most cases `__dirname` is used to locate a sibling file in a module's node.js package so it's almost comes with some assumptions about the structure of the package itself.

For now, I rather just emit a clear error and we will see if this needs tweaking in the future if you guys are okay with that.

eladb: Yes, in preflight there's no problem because we are not bundling.

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: Done. Thanks!

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: Done

libs/wingsdk/src/target-sim/function.inflight.ts
Chriscbr: Totally cool with me 🙂

Chriscbr (approved):

PR #2632, opened 2023-05-23
revert: `vm2` is not good enough

MarkMcCulloh:
Reverts winglang/wing#2615

Caused regression on Mac+Node19 https://github.com/winglang/wing/actions/runs/5051767063/jobs/9064557874


MarkMcCulloh:
> Is this also an error you're able to reproduce locally?

I tried but was unable to, so likely something purely environmental or a race condition that is aggravated by the environment.

Chriscbr (approved):

PR #2634, opened 2023-05-23
fix(sdk): vm2 is not good enough (take two)

### LINKED ISSUE #2059: Cannot use `await import()` to import ESM-only modules in inflight

### LINKED ISSUE #2088: Global node package URLSearchParams not defined

eladb:
*(Take 2 of #2615 which was reverted by #2632 - we now use bundle synchronously)*

The `vm2` module which used to execute inflight functions within the simulator was not cutting it. The sandbox has limitations such as not being able to handle ESM imports, unable to map certain global types due to the vm2 proxy behavior, and more.

To address this, we use `esbuild` to bundle the function's code *at runtime* (JIT). This is done during the initialization of the simulator cloud function and then reused across invokes. It will likely increase the simulator app start time by a little but this is not something we optimize for at the moment.

### Implementation Notes

- esbuild does not bundle any node.js sdk modules, so we map `require` into the VM to allow that.
- Since we use `esbuild` through wasm, we needed the `dom` lib in `tsconfig.test.json`.
- To allow testing ESM imports, we symlink the SDK's `node_modules` directory into the temporary directory created for each test app.
- Unfortunately, modules that use `__dirname` and/or `__filename` cannot be bundled and therefore are not supported. We emit an explicit error when these are used.

### Misc

- Fixed a bug in serialization of null objects (`typeof(null)` in javascript is surprisingly an `"object"`).

### Related

- Resolves #2059
- Resolves #2088
- #2084 is not fully resolved because it cannot be bundled due to the fact that it uses `__dirname` internally to locate the ngrok binary.


libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -1,23 +1,31 @@
...
+import { Bundle, createBundle } from "../utils/bundling";
 
 export class Function implements IFunctionClient, ISimulatorResourceInstance {
   private readonly filename: string;
   private readonly env: Record<string, string>;
   private readonly context: ISimulatorContext;
   private readonly timeout: number;
 
+  /**
+   * The path to the bundled function code. Assigned in `init()`.
```
eladb, line 25:
```suggestion
   * The bundle information (assigned during `init()`).
```

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -29,6 +37,9 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
   }
 
   public async init(): Promise<FunctionAttributes> {
+    const workdir = await mkdtemp(path.join(tmpdir(), "wing-bundles-"));
+
```
eladb, line 41:
```suggestion
```

Chriscbr (approved):

Issue #2103, opened 2023-04-16
Optionality: advanced behavior


staycoolcall911:
https://github.com/winglang/wing/blob/main/docs/04-reference/winglang-spec.md#17-optionality

Most of the features described in the Optionality section are implemented as part of #436. The following features are extra (p3):
- [ ] #3857
- [ ] #3858
- [ ] #3859
- [ ] #3860
- [ ] #3861
- [ ] #3862

marciocadev:
The PR #3840 add the support to `if let var`

staycoolcall911:
Added #3857, #3858, #3859, #3860, #3861, #3862 for the remaining sub-tasks and closing this issue.

Issue #884, opened 2022-12-13
Type checker: treat unresolved types as `unresolved` instead of `anything`


staycoolcall911:
During compilation's type checking phase, in certain situations where we can't resolve/infer the type for an object, we currently do the following:
1. Add an error, so this will eventually fail.
2. Internally resolve the object type to `anything`, to allow our compilation to proceed to the next steps (e.g. jsification) and accumulate additional compilation errors.

Consider the following code from https://github.com/winglang/wing/blob/main/libs/wingc/src/type_check.rs - where an empty un-typed array (`let arr = []`) leads to the case described above where we cannot resolve the type for the array object:
```ts
self.expr_error(exp, "Cannot infer type of empty array".to_owned());
self.types.add_type(Type::Array(self.types.anything()))
```

Now, since `anything` is a **legal** internal type, we want to use a different type here, to indicate there's an error with the resolved type in this case.
To sum up, we would like to keep adding an error in the same way in these cases (# 1 above), but change the type we resolve to `unresolved` instead of `anything` (# 2 above).

eladb:
@MarkMcCulloh what should we call this pseudo-type?

* `unresolved`
* `unknown`
* `error`

Else?

MarkMcCulloh:
[I'm still a fan of unresolved](https://github.com/winglang/wing/issues/884#issuecomment-1351864661) 👍

Chriscbr:
Can anyone share more examples to illuminate the issue?

If I try the example from @yoav-steinberg in #2038, I think the errors Wing produce look reasonable to me
```
error: Expected type to be "num", but got "str" instead
  --> add.w:1:13
  |
1 | let x = 5 + "hello";
  |             ^^^^^^^ Expected type to be "num", but got "str" instead


error: Expected type to be "str", but got "num" instead
  --> add.w:2:5
  |
2 | log(x);
  |     ^ Expected type to be "str", but got "num" instead
```

(of course we can still improve this example by special-casing a message that recommends `.concat()` etc. - but I feel like it's a better developer experience for the compiler to assume `x` is a `num`)

MarkMcCulloh:
For me this issue is more about correctness, I'm sure we can find ways to improve the messages with this info though.

```rust
let x = jsii_module.method_returns_any();
  //^ `untyped`

log(x);
  //^ This should be allowed imo. Warning maybe?

let y = 2 + "a";
	      //^ error generated here
  //^ `unresolved`

log(y2);
[...]

PR #2636, opened 2023-05-23
feat(compiler): add `if let` statements

hasanaburayyan:
Adds support for `if let` statements to be used with optionals.

as described: https://github.com/winglang/wing/issues/436

- [ ] PR title matches [Winglang's style guide](https://docs.winglang.io/contributors/pull_requests#how-are-pull-request-titles-formatted)
- [ ] PR description explains motivation and solution
- [ ] Tests added
- [ ] Docs updated

### DIFF
--- a/examples/tests/invalid/optionals.w
+++ b/examples/tests/invalid/optionals.w
@@ -33,3 +33,8 @@ optionalSub1 ?? new Sub2();
 //               ^ error: Sub2 is not a subtype of Sub1
 optionalSub1 ?? new Super();
 //               ^ error: Super? is not a subtype of Sub1
+
+
+if let x = true {
+//         ^^^^ Expected type to be optional, but got "bool" instead
+}
--- a/examples/tests/valid/optionals.w
+++ b/examples/tests/valid/optionals.w
@@ -23,3 +23,56 @@ inflight class Sub1 extends Super {
 let optionalSup: Super? = new Super();
 let s = optionalSup ?? new Sub();
 assert(s.name == "Super");
+
+struct Name {
+  first: str;
+  last: str?;
+}
+
+let var name: Name? = Name {
+  first: "John",
+  last: "Doe",
+};
+
+// simple tests
+if let n = name {
+  assert(n.first == "John");
+}
+
+name = nil;
+
+if let n = name {
+  assert(false); // should not be reached
+} else {
+  assert(true);
+}
+
+let tryParseName = (fullName: str): Name? => {
+  let parts = fullName.split(" ");
+  if parts.length < 1 {
+    return nil;
+  }
+  return Name {
+    first: parts.at(0),
+    last: parts.at(1),
+  };
+};
+
+
+// Nested if lets
+if let parsedName = tryParseName("Good Name") {
+  assert(parsedName.first == "Good");
+  // check for optional last name
+  if let lastName = parsedName.last {
+    assert(lastName == "Name");
+  } else {
+    assert(false); // Something went wrong
+  }
+}
+
+if let parsedName = tryParseName("BadName") {
+  assert(parsedName.first == "BadName");
+  if let lastName = parsedName.last {
+    assert(false); // No last name should exist
+  }
+}
--- a/libs/tree-sitter-wing/grammar.js
+++ b/libs/tree-sitter-wing/grammar.js
@@ -104,6 +104,7 @@ module.exports = grammar({
         $.break_statement,
         $.continue_statement,
         $.if_statement,
+        $.if_let_statement,
         $.struct_definition,
         $.enum_definition,
         $.try_catch_statement,
@@ -266,6 +267,15 @@ module.exports = grammar({
     break_statement: ($) => seq("break", $._semicolon),
 
     continue_statement: ($) => seq("continue", $._semicolon),
+    
+    if_let_statement: ($) => seq(
+      "if let",
+      field("name", $.identifier),
+      "=",
+      field("value", $.expression),
+      field("block", $.block),
+      optional(seq("else", field("else_block", $.block)))
+    ),
 
     if_statement: ($) =>
       seq(
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -588,6 +588,26 @@ impl<'a> JSifier<'a> {
 			}
 			StmtKind::Break => CodeMaker::one_line("break;"),
 			StmtKind::Continue => CodeMaker::one_line("continue;"),
+      StmtKind::IfLet { 
+        condition, 
+        statements, 
+        var_name,
+        else_statements
+      } => {
+        let mut code = CodeMaker::default();
+        code.open(format!("if ({}) {{", self.jsify_expression(condition, ctx)));
+        code.line(format!("const {} = {};", self.jsify_symbol(var_name), self.jsify_expression(condition, ctx)));
+        code.add_code(self.jsify_scope_body(statements, ctx));
+        code.close("}");
+
+        if let Some(else_scope) = else_statements {
+					code.open("else {");
+					code.add_code(self.jsify_scope_body(else_scope, ctx));
+					code.close("}");
+				}
+        
+        code
+      }
 			StmtKind::If {
 				condition,
 				statements,
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -588,26 +588,30 @@ impl<'a> JSifier<'a> {
 			}
 			StmtKind::Break => CodeMaker::one_line("break;"),
 			StmtKind::Continue => CodeMaker::one_line("continue;"),
-      StmtKind::IfLet { 
-        condition, 
-        statements, 
-        var_name,
-        else_statements
-      } => {
-        let mut code = CodeMaker::default();
-        code.open(format!("if ({}) {{", self.jsify_expression(condition, ctx)));
-        code.line(format!("const {} = {};", self.jsify_symbol(var_name), self.jsify_expression(condition, ctx)));
-        code.add_code(self.jsify_scope_body(statements, ctx));
-        code.close("}");
+			StmtKind::IfLet {
+				condition,
+				statements,
+				var_name,
+				else_statements,
+			} => {
+				let mut code = CodeMaker::default();
+				code.open(format!("if ({}) {{", self.jsify_expression(condition, ctx)));
+				code.line(format!(
+					"const {} = {};",
+					self.jsify_symbol(var_name),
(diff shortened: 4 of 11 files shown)


libs/wingc/src/jsify.rs
```diff
@@ -591,6 +591,30 @@ impl<'a> JSifier<'a> {
...
 			StmtKind::Break => CodeMaker::one_line("break;"),
 			StmtKind::Continue => CodeMaker::one_line("continue;"),
+			StmtKind::IfLet {
+				condition,
+				statements,
+				var_name,
+				else_statements,
+			} => {
+				let mut code = CodeMaker::default();
+				code.open(format!("if ({}) {{", self.jsify_expression(condition, ctx)));
```
Chriscbr, line 601: I think we need to check explicitly against null here. Consider this example:

```js
// wing
let x: bool? = false;
if let y = x {
  assert(y == false);
}
```

The current code would jsify it as
```js
// javascript
const x = false;
if (x) {
  const y = x;
  // assert code
}
```

which would result in the code in the branch not being executed since "false" is [falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) in JavaScript. Instead we might need something like:

```js
// javascript
const x = false;
if (x != null) {
  const y = x;
  // assert code
}
```

libs/wingc/src/visit.rs
```diff
@@ -124,6 +124,19 @@ where
...
 		StmtKind::Break | StmtKind::Continue => {}
+		StmtKind::IfLet {
+			condition,
+			statements,
+			var_name,
+			else_statements,
+		} => {
+			v.visit_expr(condition);
+			v.visit_scope(statements);
+			v.visit_symbol(var_name);
```
Chriscbr, line 135: Reorder (need to visit AST in syntax order - I believe some code in the LSP relies on this property)
```suggestion
			v.visit_symbol(var_name);
			v.visit_expr(condition);
			v.visit_scope(statements);
```

libs/wingc/src/type_check.rs
```diff
@@ -2045,6 +2062,55 @@ impl<'a> TypeChecker<'a> {
...
+				// and above validate_type_is_optional method will attach a diagnostic error if it is not.
+				// However to avoid panics this will allow code to continue if the type is not an optional
+				// and complete the type checking process for additional errors.
+				let var_type = if cond_type.is_option() {
+					cond_type
+						.optional_type()
+						.expect("Optional types should always have a type")
+				} else {
+					cond_type
+				};
```
Chriscbr, line 2084: I think we can avoid making an unsafe `expect` / `unwrap` call here
```suggestion
				let var_type = match cond_type.optional_type() {
					Some(t) => t,
					None => cond_type,
				};
```

examples/tests/valid/optionals.w
Chriscbr, line 1: Let's also cover these cases:

1. The user can put an expression on the right side, e.g. `if let x = foo() { .. }`
2. The user can shadow the name of an existing variable, e.g. `if let n = n { .. }`
3. Return statements work inside if let statements, e.g.

```
let f = (a: str?): str => {
  if let y = a {
    return y;
  };
  return "default";
}
```

Chriscbr (commented):
Wow you're on fire with the optional PRs

libs/wingc/src/jsify.rs
hasanaburayyan: Oh this is a great catch!

examples/tests/valid/optionals.w
```diff
@@ -23,3 +23,56 @@ inflight class Sub1 extends Super {
...
+    assert(false); // Something went wrong
+  }
+}
+
+if let parsedName = tryParseName("BadName") {
+  assert(parsedName.first == "BadName");
+  if let lastName = parsedName.last {
+    assert(false); // No last name should exist
+  }
+}
```
eladb, line 78: Very coooool

libs/tree-sitter-wing/grammar.js
yoav-steinberg, line 1: Missing grammar test (libs/tree-sitter-wing/test/corpus/)

libs/wingc/src/type_check.rs
```diff
@@ -1843,6 +1851,15 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	fn validate_type_is_optional(&mut self, actual_type: TypeRef, span: &impl Spanned) {
```
yoav-steinberg, line 1854: style: Note sure we need this function, maybe inline it down below do be more explicit (it's only used once).

libs/wingc/src/type_check.rs
```diff
@@ -2045,6 +2062,55 @@ impl<'a> TypeChecker<'a> {
...
+				statements,
+				var_name,
+				else_statements,
+			} => {
+				let cond_type = self.type_check_exp(condition, env);
+				self.validate_type_is_optional(cond_type, condition);
+
+				// Technically we only allow if let statements to be used with optionals
+				// and above validate_type_is_optional method will attach a diagnostic error if it is not.
+				// However to avoid panics this will allow code to continue if the type is not an optional
```
yoav-steinberg, line 2076: Actually we're not avoiding panics, we could just return here after creating the diagnostic, but we continue anyway for the sake of verbose diagnostics (see #884).
```suggestion
				// However for the sake of verbose diagnostics we'll allow the code to continue if the type is not an optional
```

libs/wingc/src/type_check.rs
yoav-steinberg: Check out @MarkMcCulloh's neat helper:
```suggestion
				let var_type = if cond_type.maybe_unwrap_optional();
```
We can also delete the `optional_type()` method above as I think we don't really need it.

yoav-steinberg (commented):
Looks great
Do we have an issue for `let else` as well? I think we should have it as the natural companion for `if let`.

staycoolcall911:
> Do we have an issue for let else as well? I think we should have it as the natural companion for if let.

@yoav-steinberg, @hasanaburayyan - added as a sub-task to this issue:
[Optionality: advanced behavior#2103](https://github.com/winglang/wing/issues/2103)

libs/tree-sitter-wing/grammar.js
```diff
@@ -266,6 +267,15 @@ module.exports = grammar({
...
     continue_statement: ($) => seq("continue", $._semicolon),
+    
+    if_let_statement: ($) => seq(
+      "if let",
+      field("name", $.identifier),
+      "=",
+      field("value", $.expression),
+      field("block", $.block),
+      optional(seq("else", field("else_block", $.block)))
+    ),
```
Chriscbr, line 278: Let's add a grammar test as well

libs/wingc/src/jsify.rs
```diff
@@ -591,6 +591,59 @@ impl<'a> JSifier<'a> {
...
 			StmtKind::Break => CodeMaker::one_line("break;"),
 			StmtKind::Continue => CodeMaker::one_line("continue;"),
+			StmtKind::IfLet {
+				value,
+				statements,
+				var_name,
+				else_statements,
+			} => {
+				let mut code = CodeMaker::default();
+				// To enable shadowing variables in if let statements, the following does some scope trickery
```
hasanaburayyan, line 601: I took some time to think about how to handle shadowing here, I think this is the easiest way but let me know whatcha think.

libs/tree-sitter-wing/grammar.js
```diff
@@ -266,6 +267,15 @@ module.exports = grammar({
     break_statement: ($) => seq("break", $._semicolon),
 
     continue_statement: ($) => seq("continue", $._semicolon),
+    
+    if_let_statement: ($) => seq(
+      "if let",
```
Chriscbr, line 272: Optional idea for a future enhancement...
```suggestion
      // TODO: support "if let var"
      "if let",
```

examples/tests/invalid/optionals.w
```diff
@@ -33,3 +33,8 @@ optionalSub1 ?? new Sub2();
 //               ^ error: Sub2 is not a subtype of Sub1
 optionalSub1 ?? new Super();
 //               ^ error: Super? is not a subtype of Sub1
+
+
+if let x = true {
+//         ^^^^ Expected type to be optional, but got "bool" instead
+}
```
Chriscbr, line 40: Another case that might be worth checking
```
let var x: str? = "hi";
if let x = x {
  x = 5; // error - cannot reassign to 'x'
}
```

libs/wingc/src/jsify.rs
```diff
@@ -591,6 +591,59 @@ impl<'a> JSifier<'a> {
...
+				// To work around this, we can generate a temporary scope, then use an intermediate variable to carry the
+				// shadowed value, like so:
+				// const x = "hello"
+				// {
+				//  const $IF_LET_VALUE = x; <== intermediate variable that expires at the end of the scope
+				//  if (x != undefined) {
+				//    const x = $IF_LET_VALUE;
+				//    log(x);
+				//  }
+				// }
```
Chriscbr, line 625:
```suggestion
				// }
				// The temporary scope is created so that intermediate variables created by consecutive `if let` clauses
				// do not interfere with each other.
```

libs/wingc/src/jsify.rs
Chriscbr: Makes sense to me 👍 . Another way could be
```js
const x = "hello"
if (x != undefined) {
    (($IF_LET_VARIABLE) => {
        const x = $IF_LET_VARIABLE;
        console.log(x);
    })(x);
}
```
(I'm not sure if there's any functional difference though, so we can probably keep the existing way)

Chriscbr (approved):
🙌

libs/wingc/src/jsify.rs
hasanaburayyan: @Chriscbr I love that, wish I would have thought about it 😂  let me check if it works (which Im sure it does) Ill update PR if so. I dont believe there is a functional difference but as far as conciseness is concerned ❤️

hasanaburayyan: Ah actually this causes some issues with returning out of an `if let` statement. Ill keep the current impl for now

examples/tests/invalid/optionals.w
```diff
@@ -33,3 +33,14 @@ optionalSub1 ?? new Sub2();
...
+
+let var hi: str = "hi";
```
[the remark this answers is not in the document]
hasanaburayyan: Yes you are right I meant for it to be optional. Actually the compiler complained about that but I missed it. Also Ill add a comment now

eladb:
Yes!!

PR #2665, opened 2023-05-25
fix(compiler): diagnostics missing error messages

Chriscbr:
Fixes an issue where some type checking diagnostics are generated without spans showing where in your code the error is.

In the process I refactored some of the code for generating error messages. It didn't feel like it was as big of an improvement as I expected so I'm open to reverting if the changes seem counterproductive. There's also this `TypeError` struct I saw which doesn't seem to serve any purpose / it's basically the same as `Diagnostic`, so I was also thinking about removing that, but I wasn't sure so I left it as is.

Misc:
- removed some dead (unused) code

----

- [x] Tests added
- [x] Docs updated

### DIFF
--- a/tools/hangar/__snapshots__/invalid.ts.snap
+++ b/tools/hangar/__snapshots__/invalid.ts.snap
@@ -447,6 +447,10 @@ error: Expected type to be \\"inflight (str): void\\", but got \\"inflight (num)
 
 
 error: Expected an interface, instead found type \\"Bucket\\"
+   --> ../../../examples/tests/invalid/impl_interface.w:14:14
+   |
+14 | class C impl cloud.Bucket {
+   |              ^^^^^^^^^^^^ Expected an interface, instead found type \\"Bucket\\"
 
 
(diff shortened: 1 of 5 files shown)


tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1233,18 +1245,18 @@ error: \\"x\\" is not initialized
    |                  ^^^^^^^^^^ \\"x\\" is not initialized
 
 
-error: struct fields must be immutable got: MutArray<str>
+error: Struct fields must have immutable types
    --> ../../../examples/tests/invalid/structs.w:20:3
    |
 20 |   f: MutArray<str>;
-   |   ^ struct fields must be immutable got: MutArray<str>
```
Chriscbr, line 1240: The type annotation is required and is right next to the field name, so removing the duplication in the error message seemed like a net good, I think?

eladb: yes

eladb (approved):
LGTM!

PR #2666, opened 2023-05-26
fix: missing completions in if-statement condition

### LINKED ISSUE #2374: Invalid completion inside an if statement (vscode)
### I tried this:

See video

@ekeren additional info:
```ts (wing)
let a = MutMap<str> {};
let removeThisLineToFixIssue = MutMap<str> {"a": "b"};

if a.
// .. ^ autocomplete does not work because of line 2 
// (the fact that we have  {"a": "b"} seems to break stuff)
```

### This happened:

https://user-images.githubusercontent.com/598796/235889844-7b8a3bae-626d-4ea7-9fc6-b2e067c84d5c.mov

### I expected this:

To see the completion for the map api

### Is there a workaround?

_No response_

### Component

IDE Extension

### Wing Version

0.15.4

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
Fixes #2374

### Problem

tree-sitter is unable to recognize the following if-statement:

```
let a = MutMap<str> {};
if a.
```

due to missing at least a `{` to start off a block. This means the entire statement becomes an error, causing the reference inside the condition to get thrown away and unavailable to the compiler/lsp.

### Solution

The idea here is to have something similar to the automatic semicolon insertion, but for blocks instead. This means tree-sitter will automatically insert an empty block `{}` when one is expected but unavailable. Like the auto semicolon, this still results in a parse error but we still create a useful AST.

This not only solves the problem for if-statements, but other similar syntactic structures like for-loops.

- [x] Tests added
- [ ] Docs updated


libs/tree-sitter-wing/src/scanner.c
```diff
@@ -150,6 +151,42 @@ static bool scan_automatic_semicolon(TSLexer *lexer)
   return true;
 }
 
+/**
+ * Check if an automatic empty block ( `{}` ) could be inserted
+ *
+ * @return true if an automatic block was found (aka "inserted"), false otherwise
+ */
+static bool scan_automatic_block(TSLexer *lexer)
+{
```
Chriscbr, line 160: I have no idea how the tree sitter parser works but it looks dope. Would be cool to learn more about it in a team time or something 😄

Chriscbr (approved):
Very nice

PR #2669, opened 2023-05-26
fix(compiler): require `?.` for optional access

hasanaburayyan:
This PR is to address a bug found in https://github.com/winglang/wing/pull/2663 where we are not enforcing the use of `?.` when access properties on an optional type.

I wanted to break it into its own PR to keep the changes small in scope and separate before updating the other PR.

- [x] Tests added
- [x] Docs updated
- [x] Needs end-to-end tests (`pr/e2e-full` label)


eladb (approved):

ekeren:
@hasanaburayyan I am wondering if,  following this change, we should change .get and .getAt to return `Json?`? (Instead of `Json`)

hasanaburayyan:
> @hasanaburayyan I am wondering if, following this change, we should change .get and .getAt to return `Json?`? (Instead of `Json`)

@ekeren My initial thoughts are no because we are not trying to guarantee any sort of strong typing on Json until you convert it back to a wing type, so perhaps instead we should make the primitive static methods return optional types. IE `str.from_json()` should return optional `str?` 

Im also increasingly dissatisfied with the DX on Json and would love to get `.` and `[]` accessors implemented, to make it easier to work with.

ekeren:
I think that the return type of the `.` Operator is a `Json?`

Basically 
```
a.b?.c
```
Is equal to:
```
a.get("b")?.get("c")
```
Only if `get` returns an optional Json 

On the subject of Json Dx, I actually think that we should consider adding the conversion methods on the Json (instead of the current static functions)`asStr` `asBool` `asNum`


```
a.get("b")?.get("c")?.asStr()
```

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -899,6 +899,20 @@ error: Expected type to be optional, but got \\"bool\\" instead
    |            ^^^^ Expected type to be optional, but got \\"bool\\" instead
 
 
+error: Property access unsupported on type \\"A?\\"
+   --> ../../../examples/tests/invalid/optionals.w:61:9
+   |
+61 | let c = b.a.val;
+   |         ^^^ Property access unsupported on type \\"A?\\"
```
staycoolcall911, line 906: Is that error message intentional?

hasanaburayyan: @staycoolcall911 Actually this error message is current issue with optional types that gets resolved in the sister PR for optional chaining: https://github.com/winglang/wing/pull/2663

However I cannot produce my intended error without producing this one as well.

libs/tree-sitter-wing/grammar.js
```diff
@@ -83,12 +83,14 @@ module.exports = grammar({
...
             )
           ),
-          choice(".", "?."),
+          field("accessor_type", $.accessor),
           // While the "property" identifier is optional in this grammar, upstream parsing will fail if it is not present
           optional(field("property", $._member_identifier))
         )
       ),
 
+    accessor: ($) => choice(".", "?."),
```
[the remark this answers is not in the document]
hasanaburayyan: Thanks Mark, TIL 🚀

Issue #2413, opened 2023-05-05
Default phase based on scope


eladb:
### Feature Spec

The phase modifier for classes and closures is optional and defaults to the phase of the scope in which the declaration is made.

This implies, for example, that a class or a closure declared within the scope of an inflight method will be an inflight class:

```js
class Hello {}

inflight class Bye {}

let y = () => {};

let x = inflight () => {
  class Foo {}
  
  new Foo(); // ok!

  let f = (): str => {
    return “bar”;
  }; // ok!

  new Bye(); // ok!

  new Hello(); // error
};
```

In the above snippet:

- `Hello` is a preflight class
- `y` is a preflight closure because the entry phase is preflight
- `x` is an inflight closure (explicitly stated).
- `Foo` is an inflight class (declared inside an inflight scope)
- `f` is an inflight (declared within an inflight scope)

credits: @mark

### Use Cases

Normalize model

### Component

Language Design

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.18.11](https://github.com/winglang/wing/releases/tag/v0.18.11).

PR #2677, opened 2023-05-28
feat: inflight classes

eladb:
## Description

This PR adds support for inflight classes:

```js
// either declared in a preflight scope
inflight class User {
  name: str;
  last: str;
  
  fullName(): str { return "${name} ${last}"; }
}
```

Inflight classes can also be declared within inflight scopes:

```js
class Foo {
  inflight bar() {
    class MyClass { } // inflight is implicit
    new MyClass();
  }
}
```

The `inflight` modifier is inherited from the scope, so there is no need
to add it when a class is defined within an inflight scope. Similarly, it is
required to specify `inflight` for members of an inflight class. All of them are
implicitly inflight.

## Implementation Notes

* This is a pretty ambitious refactor to merge the resources and classes in our type system (and jsification) into a single concept of "class" associated with a phase.
* When a class is declared without an explicit phase, it will inherit the phase of its context. This means that a `class` within an inflight context is the same as `inflight class` from a preflight context. This required tracking the phase during parsing (which is mostly just annoying boilerplate).
* I've attempted to also unify the code paths within the jsification phase so basically every inflight class is emitted into a separate file (same as the inflight portion of preflight classes) and rendered within the same code path with a few differences:
* Inflight initializer of inflight classes is basically `constructor` (while the preflight constructor is not emitted).
* When we emit an inflight class within an inflight context, we still emit it into a separate file (with capturing and all) and `require` this file at the point of declaration.

## Misc

* Cleaned up a bunch of whitespace issues in emitted JavaScript.
* I've flattened all of the emitted javascript output (removed the `clients/` directory) and switched the inflight files to have an `inflight.` prefix (instead of a suffix). This results in a nice list of JavaScript files - one called `preflight.js` and the rest are `inflight.Xyz.js`. This makes it a little cleaner to `require()` modules of inflight classes and gets rid of the concept of "client".
* Got rid of `Initializer` and `MethodLike` (because now initializers are just functions).
* Merged `FreeVariableScanner` and `PreflightTypeRefVisitor` into `CaptureScanner` that takes care of both.
* This PR includes https://github.com/winglang/wing/pull/2737
* Added static `phase` tracking within the parser. This is required in order to determine the default implicit phase of declarations.
* Fixes https://github.com/winglang/wing/issues/2413

## Follow ups
* We have a a problem with our async programming model in inflight classes because there is no way to indicate that a method is async. For now, all methods are async (since they are tagged `inflight`), but that means we cannot call them from within the constructor. We need to decide how to deal with it.
* Merge the `FieldReferenceVisitor` into `CaptureScanner` as well.
* https://github.com/winglang/wing/issues/2764
* https://github.com/winglang/wing/issues/2763
* https://github.com/winglang/wing/issues/2757
* https://github.com/winglang/wing/issues/2753
* https://github.com/winglang/wing/issues/2730
* https://github.com/winglang/wing/issues/2765

## Checklist

- [x] Tests added (always)


examples/tests/valid/preflight_class_inflight.w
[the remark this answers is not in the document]
eladb: Yes, I'll rename

libs/wingc/src/type_check.rs
```diff
@@ -190,6 +190,9 @@ pub struct Class {
 	pub is_abstract: bool,
 	pub type_parameters: Option<Vec<TypeRef>>,
 	pub phase: Phase,
+
+	/// The phase in which this class was declared
+	pub declaration_phase: Phase,
```
yoav-steinberg, line 195: Note that this should be identical to the `phase` field of the `env` in which the lookup found the type. See `SymbolLookupInfo`. I tried to add some documentation in symbol_env.rs to clear this up, and hopefully this can be used instead of of adding this field here.

libs/wingc/src/visit.rs
```diff
@@ -254,22 +257,29 @@ where
...
 
+pub fn visit_expr_new<'ast, V>(v: &mut V, _node: &'ast Expr, class: &'ast TypeAnnotation, _obj_id: &'ast Option<String>, obj_scope: &'ast Option<Box<Expr>>, arg_list: &'ast ArgList)
+	where
+	V: Visit<'ast> + ?Sized,
+{
+	v.visit_type_annotation(class);
+	if let Some(scope) = obj_scope {
+		v.visit_expr(&scope);
+	}
+	v.visit_args(arg_list);
```
yoav-steinberg, line 268: Feels like the args should come before the `scope`.
```suggestion
	v.visit_args(arg_list);
	if let Some(scope) = obj_scope {
		v.visit_expr(&scope);
	}
```

examples/tests/valid/inflight_class_inside_inflight_closure.w
```diff
@@ -30,3 +30,14 @@ let f = p.preflight_method();
 test "it works" {
   f.invoke("text");
 }
+
+test "inflight class inside closure captures from closure" {
```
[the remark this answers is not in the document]
eladb: Very nice catch.

For outer inflight classes, the compiler detects the attempt to capture mutable/reassignable and emits an error.
For inner inflight classes, the compiler allows the capture and only the mutability works (but reassignablity doesn't because the inflight class is defined in a separate file).

I'll add an issue to track this (I think we shouold be able to merge all the capture visitors together and then we can run these checks there).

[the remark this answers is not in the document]
eladb: https://github.com/winglang/wing/issues/2729

libs/wingc/src/type_check.rs
eladb: I am removing this

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -167,8 +179,12 @@ impl SymbolEnv {
 		self: reference([Self]),
 		symbol: &Symbol,
 		not_after_stmt_idx: Option<usize>,
+		depth: usize,
```
yoav-steinberg, line 182: ugly, but given I don't have a quick and easy solution then, ok

libs/wingc/src/jsify.rs
```diff
@@ -1771,87 +1764,115 @@ impl<'a> FieldReferenceVisitor<'a> {
...
-			if let SymbolKind::Type(t) = kind {
-				// If this class was defined preflight then store it
-				let decl_phase = t.as_class().and_then(|f| Some(f.declaration_phase)).unwrap_or(Phase::Preflight);
-				if decl_phase == Phase::Preflight {
-					self.references.insert(type_name, *t);
+	fn consider_reference(&mut self, path: &[&Symbol]) {
+		if let LookupResult::Found(kind, info) = self.method_env.lookup_nested(path, Some(self.statement_index), 0) {
+			// if the symbol was found in our current enviroment, we move on. that's the essence of capturing:
+			// you reference something that defined in a parent environment.
+			if info.depth == 0 {
```
yoav-steinberg, line 1805: As we discussed this needs to account for the case of a nested scope inside the function.

libs/wingc/src/jsify.rs
```diff
@@ -1825,27 +1794,33 @@ impl<'ast> Visit<'ast> for PreflightTypeRefVisitor<'ast> {
...
+			arr.extend(u.fields.iter().map(|f| f.name.as_str()));
+
+			let path = arr.join(".");
+			self.add_ref(path);
+		}
+
+		visit::visit_type_annotation(self, node);
+	}
+
 	fn visit_reference(&mut self, node: &'ast Reference) {
```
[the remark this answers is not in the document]
yoav-steinberg: I think the visitor's trait's default implementation of `visit_reference` calls `visit_type_annotation` in case of a static method (in case of a `TypeMember` reference). So it should work.

yoav-steinberg (approved):
:rocket:

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
eladb: Yes but it also calls `visit_type_annotation` in many cases I don't want to account for.

Chriscbr:
> We have a a problem with our async programming model in inflight classes because there is no way to indicate that a method is async. For now, all methods are async (since they are tagged inflight), but that means we cannot call them from within the constructor. We need to decide how to deal with it.

If inflight scopes are always async, then couldn't we emit code like this each time an inflight class is constructed?

```js
let a = new A();
await a.$inflight_init(); // allowed to call async methods
```

libs/wingc/src/parser.rs
```diff
@@ -535,54 +539,62 @@ impl<'s> Parser<'s> {
 							)
 							.err();
 					}
-					if !is_resource && is_inflight {
+					if class_phase == Phase::Independent && is_inflight {
```
Chriscbr, line 542: Not sure I understand how the code could reach a point where the class phase is independent, since the grammar only supports defining preflight classes and inflight classes at the moment

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -94,6 +94,11 @@ pub struct SymbolLookupInfo {
 	pub phase: Phase,
 	/// Whether the symbol was defined in an `init`'s environment
 	pub init: bool,
+
+	/// How deep up the symbol environment tree did we look to find this symbol.
+	///
+	/// If this is 0 it means we found it in the entry environment, 1 means first parent 2 next, etc.
+	pub depth: usize,
```
Chriscbr, line 101: Nowhere in this file or in the PR description explains this architectural change / why it was introduced, could you shed some light on this?

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -250,12 +288,17 @@ impl SymbolEnv {
 	)]
 	/// Lookup a symbol in the environment, returning a `LookupResult`. The symbol name may be a
 	/// nested symbol (e.g. `foo.bar`).
-	pub fn lookup_nested_str(self: reference([Self]), nested_str: &str, statement_idx: Option<usize>) -> LookupResult {
+	pub fn lookup_nested_str(
+		self: reference([Self]),
+		nested_str: &str,
+		statement_idx: Option<usize>,
+		depth: usize,
```
Chriscbr, line 295: What does the `depth` parameter do in this method? (Please update docstrings)

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -135,6 +140,24 @@ impl SymbolEnv {
...
 		}
 
+		// if this is a variable, verify that it adheres to the phase rules (notice that "this" is bound
+		// to the phase of the class).
+		if let SymbolKind::Variable(v) = &kind {
+			if self.phase == Phase::Inflight
+				&& v.phase == Phase::Preflight
+				&& symbol.name != "this"
+				&& symbol.name != CLASS_INIT_NAME
+			{
```
Chriscbr, line 150: IMO checking for phase compatibility seems like more of a responsibility of the type checker, not of our symbol environment abstraction, so I think this logic might belong elsewhere - what do you think?

examples/tests/invalid/inflight_class_capture_mutable.w
Chriscbr, line 1: Can inflight classes be members of structs / can we add a unit test to snapshot this behavior one way or another?

libs/wingc/src/type_check.rs
```diff
@@ -432,22 +432,14 @@ impl Subtype for Type {
...
-			}
 			(Self::Interface(l0), Self::Interface(_)) => {
 				// If we extend from `other` then I'm a subtype of it (inheritance)
 				l0.extends.iter().any(|parent| {
 					let parent_type: &Type = parent;
 					parent_type.is_subtype_of(other)
 				})
 			}
-			(Self::Resource(res), Self::Interface(iface)) => {
+			(Self::Class(res), Self::Interface(iface)) => {
```
Chriscbr, line 442: Previously, the type checker had two separate types (Class and Resource) and now it only has one type (Class). Since this subtyping rule was only expected to work with the Resource type, we need to update the logic so that the subtype relationship does not hold if the class's phase is inflight.

For example, we should check that this is a compiler error:

```js
bring cloud;

inflight class A {
  handle(msg: str): str {
    return msg;
  }
}

// A does not automatically implement cloud.IFunctionHandler
// if A was a preflight class (and `handle` was an inflight method) then this would be ok
new cloud.Function(new A());
```

I think we just need to add something like:

```rust
if res.phase != Phase::Preflight {
	return false;
}
```

libs/wingc/src/type_check.rs
```diff
@@ -491,7 +483,7 @@ impl Subtype for Type {
 				// Finally check if they're subtypes
 				res_handle_type.is_subtype_of(&handler_method_type)
 			}
-			(Self::Resource(res), Self::Function(_)) => {
+			(Self::Class(res), Self::Function(_)) => {
```
Chriscbr, line 486: See previous comment

libs/wingc/src/type_check.rs
```diff
@@ -710,32 +701,29 @@ impl Display for FunctionSignature {
 unsafe impl Send for TypeRef {}
 
 impl TypeRef {
-	pub fn as_resource(&self) -> Option<&Class> {
-		if let Type::Resource(ref res) = **self {
-			Some(res)
-		} else {
-			None
+	pub fn as_class_preflight(&self) -> Option<&Class> {
```
Chriscbr, line 704: (optional)
```suggestion
	pub fn as_preflight_class(&self) -> Option<&Class> {
```

libs/wingc/src/type_check.rs
```diff
@@ -1875,7 +1876,7 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
-	pub fn type_check_scope(&mut self, scope: &Scope) {
+	pub fn type_check_scope(&mut self, scope: &'a Scope) {
```
Chriscbr, line 1879: I think the conventional Rust style is to omit lifetimes unless they can't be inferred by the compiler

libs/wingc/src/jsify.rs
```diff
@@ -87,10 +88,10 @@ impl<'a> JSifier<'a> {
 	}
 
 	fn js_resolve_path(path_name: &str) -> String {
-		format!("\"./{}\".replace(/\\\\/g, \"/\")", path_name)
+		format!("\"./{}\"", path_name.replace("\\", "/"))
 	}
 
-	pub fn jsify(&mut self, scope: &Scope) -> String {
+	pub fn jsify(&mut self, scope: &'a Scope) -> String {
```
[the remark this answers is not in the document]
Chriscbr: I don't think so

libs/wingc/src/parser.rs
```diff
@@ -502,24 +504,26 @@ impl<'s> Parser<'s> {
 						});
 					}
 
+					// if there is no "phase_modifier", then inherit from the class phase
+					// currently "phase_modifier" can only be "inflight".
+					let phase = match class_element.child_by_field_name("phase_modifier") {
+						None => class_phase,
+						Some(_) => Phase::Inflight,
+					};
```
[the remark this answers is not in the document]
yoav-steinberg: I think we don't support explicit `preflight` phase modifiers, so if there's a modifier it's an inflight field.

libs/wingc/src/type_check.rs
yoav-steinberg: nice catch

libs/wingc/src/parser.rs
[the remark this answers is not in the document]
Chriscbr: Ah, I see what you mean now (resolving)

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -132,6 +147,23 @@ impl SymbolEnv {
 		self.parent.is_none()
 	}
 
+	/// Checks if this environment is a parent of another environment.
+	pub fn is_parent_of(&self, other: &SymbolEnv) -> bool {
```
Chriscbr, line 151: 👍

examples/tests/invalid/capture_redefinition.w
```diff
@@ -0,0 +1,7 @@
+let y = "hello";
+
+inflight () => {
+  log(y);
+//    ^ Cannot capture symbol "y" because it is shadowed by another symbol with the same name
+  let y = "world";
+};
```
yoav-steinberg, line 7: We should probably also add a test for the simple (non-capture) case, or do we support it (because TS doesn't):
```ts
// TypeScript
let x = 1;
if (true) {
  console.log(x); // Block-scoped variable 'x' used before its declaration. ts(2448)
  let x = 2;
}
```

libs/wingc/src/jsify.rs
```diff
@@ -1655,8 +1650,14 @@ impl<'a> FieldReferenceVisitor<'a> {
...
 				let var = env
-					.lookup(&x, Some(self.statement_index), 0)
-					.expect("covered by type checking")
+					.lookup(&x, Some(self.statement_index))
+					.expect(
+						format!(
+							"unable to find symbol '{x}' in environment that contains: {}",
+							env.symbol_map.keys().join(", ")
+						)
+						.as_str(),
```
yoav-steinberg, line 1659: This will also work
```suggestion
						&format!(
							"unable to find symbol '{x}' in environment that contains: {}",
							env.symbol_map.keys().join(", ")
						),
```
But anyway, this is one heck of a verbose `expect`... maybe too verbose.

libs/wingc/src/jsify.rs
```diff
@@ -1801,31 +1811,53 @@ impl<'a> CaptureScanner<'a> {
...
-				return;
-			}
+		// look up the reference in the current environment
+		let LookupResult::Found(kind, symbol_info) = self.current_env.lookup_nested(path, self.current_index) else {
+			panic!("unable to find symbol in current environment");
+		};
+
+		// now, we need to determine if the environment this symbol is defined in is a parent of the
+		// method's environment. if it is, we need to capture it.
+		if !(symbol_info.env.is_parent_of(&self.method_env)) {
```
yoav-steinberg, line 1821: nice!

libs/wingc/src/jsify.rs
```diff
@@ -1801,31 +1811,53 @@ impl<'a> CaptureScanner<'a> {
...
+					}
 				}
-				SymbolKind::Variable(var) => {
-					// skip macro functions (like "log" and "assert")
-					if let Type::Function(f) = &*var.type_ {
-						if f.js_override.is_some() {
+
+				// if the symbol was found in another environment, make sure this scope doesn't have another
+				// variable defined with the same name. if it does, it's an error
+				if !self.current_env.is_same(&symbol_info.env) {
```
yoav-steinberg, line 1841: Isn't this part of the `if` redundant since if this is a capture it was by definition found in another env?

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -132,6 +147,23 @@ impl SymbolEnv {
...
+
+		if self.is_same(&parent) {
+			return true;
+		}
+
+		self.is_parent_of(&parent)
+	}
+
+	pub fn is_same(&self, other: &SymbolEnv) -> bool {
+		std::ptr::eq(&*other, &*self)
```
yoav-steinberg, line 164: Since we're not using a `SymbolEnvRef` here jsut a `&SymbolEnv` I suspect this will work too:
```suggestion
		std::ptr::eq(other, self)
```

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -179,71 +211,69 @@ impl SymbolEnv {
...
-					phase: self.phase,
-					init: self.is_init,
-					depth: depth,
-				},
-			)
-		} else if let Some(ref_annotation([parent_env])) = self.parent {
-			parent_env.lookup_ext(symbol, not_after_stmt_idx.map(|_| self.statement_idx), depth + 1)
+
+			// its defined later in the current scope, but we still need to check if it might be captured
+			// from the parent.
```
yoav-steinberg, line 252: Not sure this comment is understandable in this context.
Also I'm not sure what the changes you did in this function do, seems like the same logic as before. Noting this in case I missed something.

eladb: mmm.. actually, I think the previous implementation was the correct one. It returned "defined later" even if the symbol could be resolved from a parent environment. Originally I thought that's something we wanted to support, but we realized this cannot be implemented in JavaScript so basically I think we can revert this and remove the check I added in the capture scanner.

examples/tests/invalid/capture_redefinition.w
eladb: Added

libs/wingc/src/type_check/symbol_env.rs
eladb: Yap, indeed. Reverted the previous behavior (with slightly more readable code if that's okay).

examples/tests/valid/calling_inflight_variants.w
```diff
@@ -31,6 +35,14 @@ class Foo {
     let partialFn = this.makeFn(x);
     return partialFn();
   }
+
+  inflight callFn2(): num {
```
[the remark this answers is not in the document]
eladb: PASS!

examples/tests/valid/inflight_class_inner_capture_mutable.w
```diff
@@ -0,0 +1,23 @@
...
+  let y = MutArray<num>[1];
+  let var i = 10;
+
+  class Inner {
+    dang(): num {
+      y.push(2);
+
+      // TODO: this should be a compiler error (it doesn't)
+      // see https://github.com/winglang/wing/issues/2729
+      i = i + 1;
```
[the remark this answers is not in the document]
eladb: Yes, but that's really hard to implement (due to the fact that our inflight classes are evaluated in a separate javascript file), and honestly I don't think very useful.
For now, I feel emitting a compiler error is good enough.

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -135,6 +140,24 @@ impl SymbolEnv {
...
+				&& symbol.name != "this"
+				&& symbol.name != CLASS_INIT_NAME
+			{
+				return Err(TypeError {
+					span: symbol.span.clone(),
+					message: format!(
+						"Unable to define the preflight member \"{}\" within an inflight scope",
+						symbol.name
+					),
+				});
```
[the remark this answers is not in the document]
eladb: Will do

libs/wingc/src/type_check/symbol_env.rs
eladb: Mmmm... not 100% sure I agree with you on this. As I see this the symbol environment is part of our type system and it makes sense that it maintains its invariants. The alternative is to actually add this check every time I define a symbol within an environment which doesn't make sense to duplicate throughout the type checker.

If that's okay, I'll leave this here for now.

libs/wingc/src/parser.rs
```diff
@@ -159,43 +159,43 @@ impl<'s> Parser<'s> {
...
-				.filter_map(|(i, st_node)| self.build_statement(&st_node, i).ok())
+				.filter_map(|(i, st_node)| self.build_statement(&st_node, i, phase).ok())
 				.collect(),
 			env: RefCell::new(None), // env should be set later when scope is type-checked
 			span: self.node_span(scope_node),
 		}
 	}
 
-	fn build_statement(&self, statement_node: &Node, idx: usize) -> DiagnosticResult<Stmt> {
+	fn build_statement(&self, statement_node: &Node, idx: usize, phase: Phase) -> DiagnosticResult<Stmt> {
```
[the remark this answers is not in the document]
eladb: I also think it's useful to track the phase.

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
eladb: removed

libs/wingc/src/jsify.rs
eladb: done

libs/wingc/src/type_check/symbol_env.rs
eladb: @Chriscbr let me know if you are okay with this.

libs/wingc/src/visit.rs
```diff
@@ -260,12 +264,21 @@ where
...
-pub fn visit_constructor<'ast, V>(v: &mut V, node: &'ast Initializer)
-where
+pub fn visit_expr_new<'ast, V>(
+	v: &mut V,
+	_node: &'ast Expr,
+	class: &'ast TypeAnnotation,
+	_obj_id: &'ast Option<String>,
+	obj_scope: &'ast Option<Box<Expr>>,
+	arg_list: &'ast ArgList,
+) where
```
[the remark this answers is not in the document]
eladb: Seems redundant to need to refactor the AST every time we want to add sugar visit events. This decoupling will allow us to enrich the visitor with events that will help remove a lot of the duplication we have across the compiler.

apps/wing/test.w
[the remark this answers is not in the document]
eladb: cleaned up

libs/wingc/src/type_check/symbol_env.rs
eladb: Trying...

examples/tests/valid/shadowing.w
```diff
@@ -0,0 +1,27 @@
...
+    result.push(foo);
+  }
+
+  let foo = "bang";
+  result.push(foo);
+
+  return result.copy();
+};
+
+test "do not capture shadowed variable" {
```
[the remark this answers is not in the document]
eladb: the test happens in `fn()` - this is just the title of the test case

Chriscbr (approved):

examples/tests/valid/shadowing.w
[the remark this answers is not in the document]
eladb: yes, i redefied `bar` inside the `if` statement.

libs/wingc/src/jsify.rs
```diff
@@ -937,45 +941,58 @@ impl<'a> JSifier<'a> {
...
-		let preflight_methods = class
-			.methods
-			.iter()
-			.filter(|(_, m)| m.signature.phase != Phase::Inflight)
-			.collect_vec();
-
-		// default base class for preflight classes is `core.Resource`
-		let extends = if let Some(parent) = &class.parent {
-			format!(" extends {}", self.jsify_user_defined_type(parent))
+		if ctx.phase == Phase::Preflight {
```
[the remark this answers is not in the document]
eladb: yes, that's exactly what happens here. Let me add some commentray/refactor.

libs/wingc/src/parser.rs
```diff
@@ -535,54 +537,58 @@ impl<'s> Parser<'s> {
 							)
 							.err();
 					}
-					if !is_resource && is_inflight {
+					if class_phase == Phase::Independent && is_inflight {
```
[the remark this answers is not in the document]
eladb: added a test and removed this condition

libs/wingc/src/type_check.rs
eladb: So this currently fails with:

```
   |                    ^^^^^^^ Cannot create inflight class "A" in preflight phase
```

Which is a reasonable error, no?

But I guess I can add this if statement for completness (although impossible to test IMHO, no)?

eladb: OK, on second thought, I think we don't have to limit this to preflight classes. I don't see a good reason for this structural subtyping not to work for inflight classes as well.
I generally think interface subtyping should be structural, but that's for a future discussion. In the meantime, I've added a test to verify this:

```js
bring cloud;

interface IGoo {
  inflight handle(): num;
}

inflight class NotGoo {
  handle(): num {
    return 123;
  }
}

test "structure interface types for 'handle'" {
  inflight class YesGoo impl IGoo {
    handle(): num {
      return 456;
    }
  }

  // this works
  let y: IGoo = new YesGoo();
  assert(y.handle() == 456);

  // this works too because if the interface has a `handle` method we use structural typing
  let x: IGoo = new NotGoo();
  assert(x.handle() == 123);
}

```

I generally prefer to reduce to a minimum the amount of special casing (preflight/inflight handle/no-handle). I think interfaces should be structurally typed acroos the system.

examples/tests/valid/preflight_class_inflight.w
```diff
@@ -0,0 +1,6 @@
+bring cloud;
+
+let handler = inflight () => {
+  class R { }
+  new R();
+};
```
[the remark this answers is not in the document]
eladb: Done, check out  `inflight_preflight_class_definition.w`

libs/wingc/src/jsify.rs
```diff
@@ -915,13 +894,11 @@ impl<'a> JSifier<'a> {
 	///   }
 	/// }
 	/// ```
-	fn jsify_resource(&mut self, env: &SymbolEnv, class: &AstClass, ctx: &JSifyContext) -> CodeMaker {
-		assert!(ctx.phase == Phase::Preflight);
+	fn jsify_class(&mut self, env: &SymbolEnv, class: &'a AstClass, ctx: &JSifyContext) -> CodeMaker {
```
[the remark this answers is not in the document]
eladb: added a test

libs/wingc/src/type_check/symbol_env.rs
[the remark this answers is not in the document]
eladb: OK, I am realizing you were right and that basically this case can't happen because there is no way (there is no grammar for it)

examples/tests/valid/inflight_class_capture_preflight_resource.w
```diff
@@ -0,0 +1,15 @@
...
+
+inflight class Foo {
+  uploadToBucket(k: str, value: str) {
+    b.put(k, value);
+  }
+}
+
+test "inflight class captures preflight resource" {
+  let f = new Foo();
+  f.uploadToBucket("hello.txt", "world");
```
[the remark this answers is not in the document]
eladb: https://github.com/winglang/wing/issues/2730

libs/wingc/src/parser.rs
```diff
@@ -159,43 +159,45 @@ impl<'s> Parser<'s> {
...
+			"while_statement" => self.build_while_statement(statement_node, phase)?,
 			"break_statement" => self.build_break_statement(statement_node)?,
 			"continue_statement" => self.build_continue_statement(statement_node)?,
-			"return_statement" => self.build_return_statement(statement_node)?,
-			"class_definition" => self.build_class_statement(statement_node, false)?,
-			"resource_definition" => self.build_class_statement(statement_node, true)?,
-			"interface_definition" => self.build_interface_statement(statement_node)?,
+			"return_statement" => self.build_return_statement(statement_node, phase)?,
+			"class_definition" => self.build_class_statement(statement_node, Phase::Inflight)?, // `inflight class` is always "inflight"
+			"resource_definition" => self.build_class_statement(statement_node, phase)?, // `class` without a modifier inherits from scope
```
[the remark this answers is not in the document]
eladb: https://github.com/winglang/wing/issues/2763

examples/tests/invalid/inflight_class_capture_mutable.w
eladb: yes, works: inflight_class_as_struct_members.w

libs/wingc/src/type_check/symbol_env.rs
eladb: Works.

PR #2687, opened 2023-05-29
fix: build process cwd is wrong in the console

skyrpex:
The Wing Console fails when using the `@winglang/sdk` package (specifically, the Simulator) because it uses the `esbuild-wasm` package in ways that bundlers can't manage properly.

This change sets the cwd of the `esbuild-wasm` spawned process to `__dirname`, and removes the `require(require.resolve("esbuild-wasm"))` to a normal `require("esbuild-wasm")`. Since the process' cwd is now the same as `libs/wingsdk/src/shared/bundling.ts` directory, Node will naturally find `esbuild-wasm`, and our Console's dist files, which include the SDK, will work as expected as well. This is necessary because the cwd in the Console environment is the root directory (`/`).

## Checklist

- [ ] Tests added (always)


eladb (approved):
Nice solution.

ShaiBer:
reverting because of #2698

PR #2688, opened 2023-05-29
fix(sdk): aws test when invoke function that returns null

marciocadev:
The PR #2587 modification fixed tests that were returning strings when executed on AWS, but it created a problem when the function didn't return a value (null). This PR validates this return and returns an empty string in these cases.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/shared-aws/function.inflight.ts
+++ b/libs/wingsdk/src/shared-aws/function.inflight.ts
@@ -27,7 +27,10 @@ export class FunctionClient implements IFunctionClient {
     if (!response.Payload) {
       return "";
     }
-    const value = JSON.parse(toUtf8(response.Payload));
+    const value =
+      toUtf8(response.Payload) === "null"
+        ? ""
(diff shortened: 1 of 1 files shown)


libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -27,7 +27,10 @@ export class FunctionClient implements IFunctionClient {
     if (!response.Payload) {
       return "";
     }
-    const value = JSON.parse(toUtf8(response.Payload));
+    const value =
+      toUtf8(response.Payload) === "null"
+        ? ""
+        : JSON.parse(toUtf8(response.Payload));
```
Chriscbr, line 33: This will result in the result being be parsed into utf8 twice.

If the string is "null" then `JSON.parse("null")` will equal `null` in JS, so I think this might also work:

```suggestion
    const value = JSON.parse(toUtf8(response.Payload)) ?? "";
```

Chriscbr (approved):

PR #2703, opened 2023-05-30
fix: build process cwd is wrong in the console

skyrpex:
Adds the fix from #2687 but without the error that caused.

The fix to the introduced error is to resolve all relative paths to absolute paths (the outdir path and the entrypoint path).

This time, I built winglang and the sdk locally and installed them globally. Could reproduce the previous issue and it doesn't happen anymore, but I guess we should add a test to the CI that proves it as well.

### Additionally:

- Changes the Node GitHub Action from `volta-cli/action@v4.0.0` to `actions/setup-node@v3`
- Adds a sleep in a test that failed randomly


.github/workflows/build.yml
```diff
@@ -38,15 +38,10 @@ jobs:
...
-
-      # Forces volta to load node and npm upfront, reducing conflict when running in parallel later
-      - name: Preload Node/NPM
-        run: |
-          node -v
-          npm -v
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: package.json
```
MarkMcCulloh, line 44:
```suggestion
          node-version-file: ".node-version"
```

.github/workflows/build.yml
```diff
@@ -78,6 +73,11 @@ jobs:
       - name: Test
         run: npm run test:ci
 
+      - name: Test Global Wing Installation
+        run: |
+          npm install -g ./apps/wing
+          wing compile examples/tests/valid/hello.w -t tf-aws
```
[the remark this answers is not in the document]
MarkMcCulloh: Oh I see a difference, we do wing test rather than a tf-aws compile. We could add that to the canary test instead.

.github/workflows/build.yml
MarkMcCulloh: Since we're not using volta, an additional step is needed to install npm at the proper version as well.

MarkMcCulloh (changes requested):

.github/workflows/build.yml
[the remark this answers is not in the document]
MarkMcCulloh: https://github.com/winglang/wing/blob/main/.github/workflows/canary.yml#L44

.github/workflows/build.yml
skyrpex: Is it really important to pin npm to a specific version? Different users may have a wide range of npm versions 🤔

PR #2722, opened 2023-05-31
fix(compiler): calls and `if let` expressions evaluated twice

### LINKED ISSUE #2713: `If let` statements preflight code makes multiple calls to expression
### I tried this:

This wing example
```js
let getSomething = (b: bool): str? => {
  if b {
    return "something";
  } else {
    return nil;
  }
};

if let msg = getSomething(true) {
  log(msg);
} else {
  log("nothing");
}
``

### This happened:

Generated preflight code makes two calls to `getSomething()` rather than just using `$IF_LET_VALUE` be used in the if statement. This opens up code to logical errors due to some mutation between function invocations
```js
{
      const $IF_LET_VALUE = (getSomething(true));
      if ((getSomething(true)) != undefined) {               // <===== Second call 
        const msg = $IF_LET_VALUE;
        {console.log(msg)};
      }
      else {
        {console.log("nothing")};
      }
    }
```

### I expected this:

Generated preflight code to look like this
```js
{
      const $IF_LET_VALUE = (getSomething(true));
      if ($IF_LET_VALUE != undefined) {
        const msg = $IF_LET_VALUE;
        {console.log(msg)};
      }
      else {
        {console.log("nothing")};
      }
    }
```

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

eladb:
When jsifying call and `if let` expressions we generate code that uses the original expression multiple times. If this expression is e.g. a call or a `new` statement, there could be side effects to the fact that it gets called.

To fix the function call case, I've changed the strategy taken in #2580. Instead of testing whether an object is a function or an object with a `handle` method, we are now returning a "function proxy" from the constructor of the generated handle class inflight. This technique allows the returned object to behave both as a function and as an object:

```js
class MyHandler {
  constructor() {
    const $obj = (...args) => this.handle(...args);
    Object.setPrototypeOf($obj, this);
    return this;
  }

  handle() {
    console.log("hi");
  }
}

const f = new MyHandler();
f(); // prints "hi"
f.handle(); prints "hi"
```

This cleans up *a lot* of boilerplate from the call sites because now they don't care if an object a normal function or a handle function. Neat!

The `if let` fix was easier: basically just use the `$IF_LET_VALUE` variable inside the condition instead of replicating the expression. This resolves #2713

## Misc

Took the opportunity to clean the js output a little:
1. Remove redundant whitespace between `class` and it's name in inflight clients.
2. Remove the redundant `{`, `}` within class methods.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -407,10 +407,9 @@ impl<'a> JSifier<'a> {
 					}
 				}
 
-				match ctx.phase {
-					Phase::Inflight => format!("(typeof {} === \"function\" ? {}{}({}) : {}{}.handle({}))", expr_string, auto_await, expr_string, arg_string, auto_await, expr_string, arg_string),
-					Phase::Independent | Phase::Preflight => format!("({}{}({}))", auto_await, expr_string, arg_string),
-				}
+				// NOTE: if the expression is a "handle" class, the object itself is callable (see
+				// `jsify_class_inflight` below), so we can just call it as-is.
+				format!("({auto_await}{expr_string}({arg_string}))")
 			}
 			ExprKind::Unary { op, exp } => {
 				let js_exp = self.jsify_expression(exp, ctx);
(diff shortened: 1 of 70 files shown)


Chriscbr (approved):
Pretty clever 👀

staycoolcall911:
Reviewing just now and this is absolutely great!

Issue #2724, opened 2023-05-31
AWS bucket notification events don't differ create and update one.


tsuf239:
### I tried this:

```
let b = new cloud.Bucket();

b.onCreate(inflight () => {
    log ("CREATED");
});

b.onUpdate(inflight () => {
    log ("UPDATED");
});

test "test" {
    b.put("a", "1");
    b.put("a", "2");
}
```

### This happened:

"CREATED" is logged twice.

We are using the "putObjectCommand" both for uploading and updating a key. This will always result in a PUT event (https://stackoverflow.com/questions/53421520/objectcreatedpost-on-s3-console-upload). 
it seems like the only way to get a POST event is to do it via the browser, there is no equivalent command in @aws-sdk/client-s3, unfortunately.

I also looked at the notification payload, there is nothing that differentiates one from the other.

We may be able to conclude what kind of event is sent by looking at the object versioning.

On the other hand, if it's a necessary feature- we could find a workaround and gain a lot of users 😊  

### I expected this:

"CREATED" to be logged once, and "UPDATED" to be logged once.is 

### Component

SDK

### Wing Version

0.17.43

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

garysassano:
Closing this since I've just tested the code and it works as expected now.

tsuf239:
It may work on sim, but still doesn't work in AWS:
<img width="1013" alt="image" src="https://github.com/winglang/wing/assets/39455181/123276cb-0ad1-4a7e-842b-b9e84e701faa">
(To test this issue properly, please remove the condition [here](https://github.com/winglang/wing/blob/main/examples/tests/sdk_tests/bucket/events.test.w#L92-L94) and run the test **on aws**)
Thank you once again @garysassano, for making us write our issues in a clearer way, keep up your great work! 💪 🚀 🦸

PR #2725, opened 2023-05-31
feat(sdk): bucket onEvent second argument and more

tsuf239:
* Adding a second argument (event type) for bucket.onEvent, so now it's aligned to the spec (fixes https://github.com/winglang/wing/issues/192, but not systemically)
* Improving `bucket/events.w`  test
* Adding an example for functions that return functions
* Exposing a severe error in aws events- https://github.com/winglang/wing/issues/2724 

### <samp>🤖 Generated by Copilot at 7bfe8d9</samp>

This pull request adds and updates several examples and tests for Wing, especially related to bucket events and functions as values. It also modifies the `convertBetweenHandlers` function and the `BucketEventHandlerClient` class to support passing additional arguments to handler clients. It updates the snapshots to reflect the changes in the examples, tests, and compilation output.

___
## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/bucket.onevent.inflight.ts
```diff
@@ -1,11 +1,19 @@
...
+  }: {
+    handler: IBucketEventHandlerClient;
+    args: { eventType: BucketEventType };
+  }) {
     this.handler = handler;
+    this.eventType = args.eventType;
   }
   public async handle(event: string) {
-    return this.handler.handle(event);
+    return this.handler.handle(event, this.eventType);
```
staycoolcall911, line 17: Does this mean this will not be a breaking change for the Wing Console?

tsuf239: no breaking changes, since you can do: 
`b.onEvent(inflight (key: str, event: cloud.BucketEventType) => {});`

but also: (the way it was before)
`b.onEvent(inflight (key: str) => {});`

and even:
`b.onEvent(inflight () => {});`

How do bucket events are supported in the console right now? @ainvoner @skyrpex @polamoros

examples/tests/sdk_tests/bucket/events.w
```diff
@@ -1,60 +1,92 @@
...
+    operation: cloud.ColumnType.STRING,
+    source: cloud.ColumnType.NUMBER,
+    }
+  });
+
+  class Util {
+    extern "../external/sleep.js" static inflight sleep(milli: num);
+  }
+
+// the source of onDelete, onUpdate, onCreate are marked as: 1, and for onEvent it's marked as source 2.
```
Chriscbr, line 20: Maybe we can store the value as a string? I think it would make the test more readable

examples/tests/sdk_tests/bucket/events.w
```diff
@@ -1,60 +1,92 @@
...
+    return inflight (): bool => {
+      let var count = 0;
+      for u in table.list() {
+        if (u.get("key") == key && u.get("operation") == operation && num.fromJson(u.get("source")) == source) {
+          count = count + 1;
         }
+      }
+      return count == expectedVal;  
+      };
+    };
```
Chriscbr, line 70: Can we move these function definitions outside of `new cloud.Test`? (style: avoid nesting functions if it isn't needed)

```js
let f = (a: num) => {
  let b = ...
  let c = ...
  
  let c = () => {
    // ...code...
  };
  
  // ...
}
```

For example here, if `d` doesn't use `a`, `b`, or `c` in its body, then there isn't a need to nest it inside of `f`.

examples/tests/sdk_tests/bucket/events.w
```diff
@@ -1,60 +1,92 @@
...
+  b.put("a", "1");
+  b.put("b", "1");
+  b.put("c", "1");
+  b.put("b", "100");
+  b.delete("c");
 
-        extern "../external/sleep.js" static inflight sleep(ms: num);
+  assert(wait(checkHitCount("a", "CREATE", 1,  1)));
+  assert(wait(checkHitCount("b", "CREATE", 1,  1)));
+  assert(wait(checkHitCount("c", "CREATE", 1,  1)));
```
Chriscbr, line 80: The test is a little hard to understand at first glance, maybe we can add a comment here explaining what we are trying to assert:
```suggestion
  // assert that onCreate events about the "a", "b", and "c" objects were each produced exactly 1 time
  assert(wait(checkHitCount("a", "CREATE", 1,  1)));
  assert(wait(checkHitCount("b", "CREATE", 1,  1)));
  assert(wait(checkHitCount("c", "CREATE", 1,  1)));
```

examples/tests/sdk_tests/bucket/events.w
```diff
@@ -1,60 +1,92 @@
...
+    }
+
+    return false;
+  };
 
-    inflight class Predicate {
-        counterVal: num;
-        init(counterVal: num) {
-            this.counterVal = counterVal;
+  let checkHitCount = inflight (key: str, operation: str, source: num, expectedVal: num): inflight (): bool => {
```
Chriscbr, line 60: This function has several arguments, so when someone reads the usage, it might not be apparent what each field means:

```js
checkHitCount("a", "CREATE", 1, 1);
```

This would be a great use case for a trailing struct argument! (I'm not sure if that's the official name of the feature but it's close enough.) You can specify a struct as the last argument, and then values of the struct can be inlined as keyword arguments to the function call, so it's easier to read, and the arguments can go in any order.

```js
struct CheckHitCountOptions {
  type: str;
  key: str;
  source: str;
  count: num;
}

let checkHitCount = inflight (opts: CheckHitCountOptions) {
  // snip
};

checkHitCount(type: "CREATE", key: "a", source: "ON_EVENT", count: 1)
```

examples/tests/valid/bucket_events.w
```diff
@@ -15,16 +15,15 @@ b.onCreate(inflight (key:str) => {
...
+b.onEvent(inflight (key: str, event: cloud.BucketEventType) => {   
+    other.put("last_${event}_key", key); 
 });
 
 other.onEvent(inflight (key: str) => {
     log("other bucket event called!");
 });
 
-test "test" {
+test "test" {
```
Chriscbr, line 26: Can we add a description for this test?

examples/tests/valid/function_returns_function.w
```diff
@@ -0,0 +1,26 @@
...
+};
+
+let wingFn = fn("wing");
+let dingFn = fn("ding");
+
+
+assert(wingFn());
+assert(!dingFn());
+
+test "inflight test" {
```
Chriscbr, line 15:
```suggestion
test "inflight functions can return other inflight functions" {
```

examples/tests/valid/function_returns_function.w
Chriscbr, line 1: TIL this works! Thanks for adding the test

Chriscbr (approved):
Nice work - I've just left a few suggestions for cleaning up the unit test a bit 👍

libs/wingsdk/src/cloud/bucket.ts
Chriscbr, line 1: I think the signature of onEvent needs to be updated to include the new parameter.

Also any changes to APIs should also include an update to the docs (libs/src/cloud/bucket.md)

Issue #1656, opened 2023-02-24
Don't panic


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

Our compiler code currently panics in places where a compiler diagnostic should be used

(e.g. #1648, #1647, #1635, #1624 and probably more).

### I expected this:

Compiler errors instead of crash

### Instead, this happened:

Panics

### Component:

Compiler

### Anything else?

* As an initial quick thing to do we should replace all `panic`s/`upwrap`s with `expect`s, so at least there's a bit more context as to what wasn't working. @MarkMcCulloh suggested to use the rust linter to require `expect` use.
* Go over the entire compiler codebase and identify all the cases where we should convert these unwraps into a compiler diagnostic.
* Move to immutable data structures so that expectations are codified into the type system and not at runtime.

MarkMcCulloh:
clippy lint for this https://rust-lang.github.io/rust-clippy/master/#unwrap_used

yoav-steinberg:
We also need to make sure we have some context when panicking (basically what span in the code caused the panic). And use this context to generate a diagnostic.
- We need to keep track of the current span and update some global context. This can be done generically in the visitor/fold trait but will also need to be done manually in the parser phase, type-checking phase and jsification where we don't use the visitor trait (yet).
- We need to install a panic handler that'll generate a diagnostic.
- We need to extend the diagnostics API so we can have some callback back to JS so we can send diags to JS including the on panic.
- The former is also related to changing the LSP so it doesn't crash on a wasm crash so we can recover after panics.

monadabot:
Congrats! :rocket: This was released in [Wing 0.21.3](https://github.com/winglang/wing/releases/tag/v0.21.3).

PR #2734, opened 2023-05-31
feat(compiler): compilation context in panic messages

yoav-steinberg:
First step addressing issues in #1656
This currently crashes the compiler:
```wing
bring cloud;
bring util;

let shell = util.env("SHELL");
log("preflight shell ${shell}");

test "test" {
  let shell = util.env("SHELL");
  log("inflight shell ${shell}");
}
```
Now we get this to stderr (with optional backtrace if `RUST_BACKTRACE=1` env is set):
`Compiler bug when jsifying ../../temp/tst.w:7:13 | panicked at 'Unexpected type: "Util" referenced inflight', libs/wingc/src/jsify.rs:1081:22`

This PR also adds two new features:
- Support for adding custom env variables to specific tests via the existing meta-comments mechanism:
```wing
/*\
env:
  MY_VAR: some value
\*/
log(util.env("MY_VAR"));
```
- A wing language "easter eggish" kind of feature for explicitly panicking the compiler via a new `😱` expression:
```wing
let x = 1;
😱; // This line will panic the compiler if `WINGC_DEBUG_PANIC=1` env is set.
```
possible values for `WINGC_DEBUG_PANIC` are `1` or `true` to make wingc panic as soon as it encounters `😱`, or you may explicitly specify the compilation phase in which you want the compiler to panic (`parsing`, `type checking`, `jsifying` see `set_compilation_context`).


libs/wingc/src/lib.rs
```diff
@@ -253,6 +255,53 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
...
 
+struct CompilationContext {
+	/// Description of current compilation phase
+	phase: String,
+
+	/// Location in source we're currently processing
+	span: WingSpan,
+}
+
+static mut COMPILATION_CONTEXT: Option<CompilationContext> = None;
```
MarkMcCulloh, line 266: Any reason to not use the same pattern as the lsp? (`thread_local!` with RefCell)

yoav-steinberg: Not really, kind of looked at this and figured to go as simple as possible and build up from there:
image

examples/tests/valid/util.env.w
```diff
@@ -1,10 +1,15 @@
+/*\
+env:
+  MY_VAR: my value
+\*/
```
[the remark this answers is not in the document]
yoav-steinberg: Makes sense. @MarkMcCulloh what would be the procedure to manually add valid/invalid TS test files (where do I put them and how do I name them) to get them working with our current test framework and snapshots and all that?
I'm also fine with just noting this for the future and keeping the `env` feature, which is pretty useful and clear and has a simple implementation.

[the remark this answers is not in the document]
MarkMcCulloh: Might be worth a sync meeting. I'm open to making our e2e tests more explicit rather than having this system to generate tests, but I also feel like the current system makes tests pretty easy to make. I'm on the fence though, especially after the recent discussion about doing less snapshotting.

In any case, I'd vote that any changes like that should be in a separate PR.

libs/wingc/src/lib.rs
```diff
@@ -279,6 +328,21 @@ pub fn compile(
 	let default_out_dir = PathBuf::from(format!("{}.out", file_name));
 	let out_dir = out_dir.unwrap_or(default_out_dir.as_ref());
 
+	std::panic::set_hook(Box::new(|pi| {
+		eprintln!(
+			"Compiler bug when {} {} | {}",
```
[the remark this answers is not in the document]
yoav-steinberg: It's just that this isn't a regular compilation failure, it's actually a compiler bug. So I think we should make that clear (that it's not something the user didn't understand, but rather an issue they should report).

libs/tree-sitter-wing/test/corpus/expressions.txt
```diff
@@ -431,3 +431,15 @@ nil;
 (source
   (expression_statement
     (nil_value)))
+
+================================================================================
+Debug panic
+================================================================================
+
+😱;
```
eladb, line 439: Too cute

libs/wingc/src/jsify.rs
```diff
@@ -1215,7 +1225,11 @@ impl<'a> JSifier<'a> {
 
 		// if this class has a "handle" method, we are going to turn it into a callable function
 		// so that instances of this class can also be called like regular functions
-		if inflight_methods.iter().find(|(name, _)| name.name == HANDLE_METHOD_NAME).is_some() {
+		if inflight_methods
+			.iter()
+			.find(|(name, _)| name.name == HANDLE_METHOD_NAME)
```
eladb, line 1230: Nit: Use `any` instead?

libs/wingc/src/lib.rs
```diff
@@ -253,6 +256,56 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
 		.expect("Failed to add builtin");
 }
 
+pub struct CompilationContext {
+	/// Description of current compilation phase
+	pub phase: String,
```
eladb, line 261: Can we make this an enum instead of a string?

libs/wingc/src/lib.rs
```diff
@@ -253,6 +256,56 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
...
+}
+
+/// Set global information about the current compilation phase.
+/// This is used for diagnostics, specifically for custom panic messages.
+///
+/// # Arguments
+///
+/// * `phase` - Description of current compilation phase, should end with 'ing' (building, parsing, etc.)
+/// * `span` - Location in source we're currently processing.
+fn set_compilation_context(phase: &str, span: &WingSpan) {
```
eladb, line 281: Is it possible in rust to make these type scoped?

```rust
CompilationContext::set(..)
```

Instead of free floating?

libs/wingc/src/lib.rs
```diff
@@ -279,6 +332,21 @@ pub fn compile(
 	let default_out_dir = PathBuf::from(format!("{}.out", file_name));
 	let out_dir = out_dir.unwrap_or(default_out_dir.as_ref());
 
+	std::panic::set_hook(Box::new(|pi| {
+		eprintln!(
+			"Compiler bug when {} {} | {}",
+			compilation_phase(),
+			compilation_span(),
+			pi.to_string().bold().white()
```
eladb, line 340: Maybe add a link to opening issues or docs?

libs/wingc/src/jsify.rs
yoav-steinberg: this was cleaned up in the inflight classes PR so not relevant anymore

libs/wingc/src/lib.rs
```diff
@@ -254,6 +257,91 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
 		.expect("Failed to add builtin");
 }
 
+/// The different phases of compilation, used for tracking compilation context
+/// for diagnostic purposes. Feel free to add new phases as needed.
+/// Try to make these end with 'ing' (building, parsing, etc.) so they'll fit in
+/// in the context of a diagnostic message.
+/// See `CompilationContext::set` for more information.
+#[derive(EnumString, Display, PartialEq, Clone, Copy)]
+#[strum(serialize_all = "kebab-case")]
```
Chriscbr, line 266: I wonder if we can avoid adding a new dependency by just implementing the FromStr trait on this type (shouldn't be that many lines of code)?

libs/wingc/src/lib.rs
```diff
@@ -254,6 +257,91 @@ fn add_builtin(name: &str, typ: Type, scope: &mut Scope, types: &mut Types) {
 		.expect("Failed to add builtin");
 }
 
+/// The different phases of compilation, used for tracking compilation context
```
Chriscbr, line 260: (style) WDYT of putting these types and macros into a dedicated file/module?

Chriscbr (approved):

libs/wingc/src/lib.rs
yoav-steinberg: Yes, good idea. I'll do it.

tsuf239:
changes added:
* reading snapshots from stderr if the right env flag exists
* changing `shouldSucceed` flag to be called `expectStdErr`

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1323,6 +1323,24 @@ Duration <DURATION>
 "
 `;
 
+exports[`panic.w 1`] = `
+"RuntimeError: unreachable
+Internal error: An internal compiler error occurred. Please report this bug by creating an issue on GitHub (github.com/winglang/wing/issues) with your code and this trace.
```
staycoolcall911, line 1328:
```suggestion
Internal error: An internal compiler error occurred. Please report this bug by creating an issue on GitHub (https://github.com/winglang/wing/issues/new/choose) with your code and this trace.
```

libs/wingc/src/jsify.rs
```diff
@@ -89,6 +90,7 @@ impl<'a> JSifier<'a> {
 	}
 
 	pub fn jsify(&mut self, scope: &Scope) -> String {
+		CompilationContext::set(CompilationPhase::Jsifying, &scope.span);
```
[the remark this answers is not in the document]
yoav-steinberg: Adding this to `jsify_statement` and `jsify_expression` should be enough to make sure we get decent location info. I don't expect we'll need to add this to many other places. Ideally we'll move more things to some visitor trait and let the trait handle this for us, but that's out of the scope here.
I currently can't think of a better way to handle this with rust.

PR #2738, opened 2023-05-31
fix(vscode): language server cannot recover from panics

MarkMcCulloh:
Normally when the language server panics, the whole thing crashes and you need to restart your editor. This adds a built-in mechanism to avoid exiting completely. It instead lets you know and waits until the wing file is changed before trying again.

The method is pretty naive and effectively resets the entire wasm memory. I did this because the thread local storage has a poisoned refcell and I couldn't figure out how to clear it. I think a full reset is avoidable but this current solution is ok.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing/src/commands/lsp.ts
+++ b/apps/wing/src/commands/lsp.ts
@@ -3,6 +3,7 @@ import {
   InitializeParams,
   TextDocumentSyncKind,
   InitializeResult,
+  DiagnosticSeverity,
 } from "vscode-languageserver/node";
 
 import * as wingCompiler from "../wingc";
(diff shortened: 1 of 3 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -1213,6 +1213,7 @@ impl<'a> TypeChecker<'a> {
 				Literal::Boolean(_) => self.types.bool(),
 			},
 			ExprKind::Binary { op, left, right } => {
+				panic!("TEST");
```
yoav-steinberg, line 1216: :point_up: probably need to clean this up, but maybe we do want some debug capability to cause the compiler to panic at a certain point in the code. I propose the panic emoji (😱) and adding it to the grammar ;).

apps/wing/src/commands/lsp.ts
```diff
@@ -3,25 +3,47 @@ import {
...
+        if (result === 0) {
+          return null;
+        } else {
+          return result;
+        }
+      } else {
+        return JSON.parse(result);
+      }
+    } catch (e) {
+      connection.window.showErrorMessage(
```
yoav-steinberg, line 37: Maybe this is too "in your face". We might want to quietly log this and recover without bothering the user (this is under the assumption that some diagnostic was reported to the vscode extension with relevant span info (so the crash looks like any other syntax error and is visible to the user).

yoav-steinberg (approved):

libs/wingc/src/type_check.rs
yoav-steinberg: ok I'm on it as part of my pr

apps/wing/src/commands/lsp.ts
MarkMcCulloh: Since the compiler crashed there are no diagnostics available. So it is a little jarring to not be told why lsp features are suddenly not working.

I added a compromise though: Instead of a message a new diagnostic error is added with the same message. Less "in your face" but still pretty visible.

yoav-steinberg: That's good, i expect this'll change soon to a real diag with my panic fixes.

Issue #2169, opened 2023-04-20
fetch in wing standard library


ekeren:
[fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) should be part of wing standard library

Chriscbr:
I don't have a strong intuition of what is the best way to expose this capability, but here's some initial ideas / suggestions - curious what others think:

Q1: where is this API found?

1. Add it to the "std" namespace so the user can use it through "std.fetch" or "std.Globals.fetch"
2. Add it as a global function so user can just use "fetch"
3. Add it to a new / custom namespace like "node" so the user can use it through "node.fetch" or "node.Globals.fetch"

Q2: what is it's API / interface?

1. Try to mirror the types of node's fetch API as closely as possible without needing to introduce "any" types. To match the return type, we need to return an inflight class named "Response" with methods for returning data as a string, or json, or blob, etc. If we do this, then we can just jsify the code into "fetch" in JS and no extra work is needed.
2. ? Not sure if there are other options that don't require more complex jsification.

Q3: How is the type/phase modeled since fetch is async?

Our compiler does not yet support defining preflight functions that async. I think it's OK if we say fetch is inflight-only function/method for now.

ekeren:
Q1:

- Another option is supporting having the `fetch` as class, using it like this:
```ts (wing)
bring cloud;
let do_everything_with_fetch = inflight ()=> {
  fetch.get(/* */);
  fetch.post(/* */);
  fetch.put(/* */);
  fetch.options(/* */);
  fetch.patch(/* */);
  fetch.delete(/* */);
  fetch.head(/* */);
  fetch.raw(method: str, /* */);
});
```

Q2: not familiar enough with fetch api, but my above response does introduce some change 

Q3:  makes sense to me

eladb:
> Q1: where is this API found?

We need a proper high level design for all of our standard library (e.g. where are the filesystem operations? Where are the crypto operations? etc).

But... Since I think this is something we don't have much bandwidth right now. I'll just propose something that "feels right" for now.

Can we put this under an `http` namespace?

```js
bring http;

let response = http.get("https://foo.bar");
```

Since JSII doesn't support global functions, how about we decide that if we import a JSII module and it has a class named `Default` (or `Global`, or whatever), then it's static methods will be accessible like so.

> Q2: what is it's API / interface?

Here's my proposed API:

```js
enum RequestCache {
  DEFAULT,
  NO_STORE,
[...]

PR #2739, opened 2023-05-31
feat(examples): url shortener

Chriscbr:
Adds another example to the repo. (Maybe we should move these out of the repo soon?)

It can't be deployed to AWS at the moment since the generated terraform has some kind of dependency cycle (using Wing 0.18.0).

## Checklist

- [x] Tests added (always)


examples/README.md
```diff
@@ -46,6 +46,12 @@ _Add yours here!_
 * **Description:** A task manager that lets you create background tasks asynchronously, and check their status at any point in time.
 * **Author:** [@Chriscbr](https://github.com/Chriscbr)
 
+### URL Shortener
+
+* **Source code:** [./proposed/url-shortener.w](./proposed/url-shortener.w)
+* **Description:** A URL shortener consisting of two API endpoints, `/create` and `/u/:id`. IDs are managed using buckets, but could be swapped with a table or other stateful storage mechanism. Requires installing `node-fetch@2` from npm.
+* **Author:** [@Chriscbr](https://github.com/Chriscbr)
```
tsuf239, line 53: That's a sweet one! (we used to have a dedicated one in my former workplace) 
soon we'll have a native fetch [#2169](https://github.com/winglang/wing/issues/2169) 🙂

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,141 @@
...
+
+  init(shortener: UrlShortener) {
+    this.shortener = shortener;
+    this.api = new cloud.Api();
+
+    this.api.post("/create", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
+      let var requestUrl = "";
+      try {
+        // TODO: use `x ?? throw()` syntax
+        requestUrl = str.fromJson((req.body ?? EMPTY_JSON).get("url"));
```
staycoolcall911, line 59: Could you use `Json {}` instead of `Json "empty"`?

staycoolcall911 (approved):
Very nice example Chris. I'm wondering if we have an issue for the dependency cycle error you saw when trying to deploy to aws via terraform

examples/README.md
```diff
@@ -46,6 +46,12 @@ _Add yours here!_
 * **Description:** A task manager that lets you create background tasks asynchronously, and check their status at any point in time.
```
eladb, line 46: Let's move everything to:

https://github.com/winglang/examples

@skorfmann created an awesome framework there for maintaining these

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,141 @@
...
+
+    // map from ids to urls
+    this.idLookup = new cloud.Bucket() as "IdLookup";
+  }
+
+  // Returns an id for the given long url. Creates a new id if one does not
+  // already exist.
+  inflight getId(url: str): str {
+    if this.urlLookup.exists(url) {
+      return this.urlLookup.get(url);
```
eladb, line 27: Better use tryGet because it's atomic

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,141 @@
...
+    }
+  
+    let id = utils.makeId();
+    this.idLookup.put(id, url);
+    this.urlLookup.put(url, id);
```
eladb, line 32: How do you handle a broken transaction?

Likely on the read path...

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,141 @@
...
+    return id;
+  }
+
+  // Get the url for the given id. Returns nil if the url does not have a
+  // corresponding id.
+  inflight getUrl(id: str): str? {
+    if this.idLookup.exists(id) {
+      return this.idLookup.get(id);
```
eladb, line 40: Don't we have tryGet yet?

examples/proposed/url-shortener.w
eladb: Why not `if let`?

eladb (approved):
Very cool example!!

Feels like there are some distributed system concerns to handle

examples/README.md
skorfmann: yes, seems like a great addition for the examples over there :)

examples/proposed/url-shortener.w
Chriscbr: We should now I believe

examples/proposed/url-shortener.w
Chriscbr: Ah I wrote this a while before we had `if let`, let me try that

examples/proposed/url-shortener.w
Chriscbr: Got it. OK I tweaked this method and documented it to fix the broken transaction issue (I think)

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,163 @@
...
+    // (transaction 2)
+    this.idLookup.put(newId, url);
+
+    return newId;
+  }
+
+  // Get the url for the given id. Returns nil if the url does not have a
+  // corresponding id.
+  inflight getUrl(id: str): str? {
+    if this.idLookup.exists(id) {
```
eladb, line 61: Any reason not to use tryGet here as well?

eladb (approved):

examples/proposed/url-shortener.w
Chriscbr: Oops! thanks!

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,163 @@
...
+      return this.idLookup.get(id);
+    } else {
+      return nil;
+    }
```
Chriscbr, line 65:
```suggestion
  inflight getUrl(id: str): str? {
    return this.idLookup.tryGet(id);
```

examples/proposed/url-shortener.w
```diff
@@ -0,0 +1,159 @@
...
+  
+    let newId = utils.makeId();
+
+    // (transaction 1)
+    this.urlLookup.put(url, newId);
+
+    // (transaction 2)
+    this.idLookup.put(newId, url);
+
+    return newId;
```
ekeren, line 55: @Chriscbr I wonder if there is a race condition here

What happens if the code is called twice with the same url?

Can the result be that it will return a newId that doesn't have a url pointing to it (in the urlLookup table)

PR #2755, opened 2023-06-01
chore(build): publish vscode extension when changed

fynnfluegge:
Skips step `Publish Extension to Visual Studio Marketplace` in publish job if no changes made to VS Code extension. Adds [step to fetch commit SHA of last successful workflow](https://github.com/nrwl/nx-set-shas). The previous reverted solution missed to handle commits with failed jobs, in particular if there is new commit with cli changes once before the last one has finished and hence cancelled. That means it should always taken the diff against the lastest successful commit in account.

Closes #2672.

## Checklist


fynnfluegge:
Hey @MarkMcCulloh second try here, since I saw an issue in the initial approach.
I added a step to the workflow to fetch the commit SHA of the last successful commit. The initial approach just calculated git diff against the previous commit which is not always correct since the workflow of that commit was cancelled or failed. I tested this [here](https://github.com/winglang/wing/actions/runs/5163224083/jobs/9301421153?pr=2755) by switching the main branch to `the-wing-trinity` since there are several workflow failures and it gave me the correct SHA of the last successful commit which is 15ae9222707140eb8a8018f718de51a9391240bf.
I think this approach seems promising now.

MarkMcCulloh (approved):
Thanks for taking another crack at this Fynn, and sorry for the big delay. I think this makes sense so lets give it a try!

PR #2762, opened 2023-06-02
chore: separate wing cli and wing compiler

skyrpex:
Extract a new package `@winglang/compiler` from the `winglang` package.

The `@winglang/compiler` package:

- Exports a `compile` function that can be reused by the Console.
- Adds three error classes to denote the different kind of errors that might occur:
  - `InternalError`
  - `CompileError`
  - `PreflightError`
- Doesn't know anything about terminals and terminal colors. That knowledge belongs to the `winglang` package

Important things to note:

- Makes `nx` test task depend on `^build` since we can't test `winglang` until `@winglang/compiler` has been compiled


skyrpex:
Right now, the `compile.test.ts` is duplicated: appears in `apps/wing` and `apps/wing-compiler`. It's the same exact test.

I think it should stay in `apps/wing`, along with the other tests. We should add lib-specific tests in `apps/wing-compiler` instead.

skyrpex:
@MarkMcCulloh @Chriscbr I need help here. The `apps/wing` package now depends on `libs/wingcompiler` but can't make the compiler build before wing does, so winglang fails to build.

skyrpex:
I'm in the process of cleaning everything up and review my own changes, but we're mostly finished here.

libs/wingcompiler/project.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "compiler",
```
MarkMcCulloh, line 2: I'm concerned we'll have some naming confusion.

`wingc` is technically `wingcompiler`. I would argument it would be more accurate to say that `wingc` should be called `wing-frontend` or `wing-compiler-frontend`.

skyrpex: Right. I chose `compiler` because the `libs/wingsdk` package is named `sdk` in the `project.json` file.

So, should I rename `wingc` to `wing-compiler-frontend`?

tools/hangar/src/package.setup.ts
```diff
@@ -40,7 +41,9 @@ export default async function () {
     "install",
     "--no-package-lock",
     "--install-links=false",
+    "--ignore-scripts",
```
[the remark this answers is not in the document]
skyrpex: I'll give it another go, now that I'm more used to the tooling here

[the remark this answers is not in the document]
skyrpex: Removed the `--ignore-scripts` and bundled `wasi-js` into `@winglang/compiler`, so there's no need for patch-packaging it.

libs/wingcompiler/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    // Some tests use process.chdir which is not supported in vitest's multi-thread mode
+    threads: false,
```
MarkMcCulloh, line 6: I don't think this config is needed in the wingcompiler project (the tests this is referring to are in the winglang proj)

MarkMcCulloh (approved):

libs/wingcompiler/project.json
MarkMcCulloh: Discussed in https://winglang.slack.com/archives/C048QCN2XLJ/p1686137164071349

No rename needed

skyrpex:
@MarkMcCulloh @staycoolcall911 please, have a look 👍🏻

PR #2773, opened 2023-06-02
feat(compiler): improve diagnostics for duplicate construct ids

### LINKED ISSUE #2623: Confusing error when instantiating two resources of the same type
### I tried this:

Wrote this code:
```ts
bring cloud;

let b = new cloud.Bucket();
let b2 = new cloud.Bucket();
```

### This happened:

Got this error:
> There is already a Construct with name 'cloud.Function' in $Root [Default]

### I expected this:

The code to compile successfully. or at least get an error that helps me to resolve this.

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

There are different problems with this error:
1. A reasonable user will not expect to get it
2. It talks about Constructs
3. It doesn't explain how to resolve the issue (use the 'as' keyword)

We cannot implement the trivial fix of editing the error text because it is coming from the CDK.

So we need some other solution.

Some alternatives I can think of:
1. Make sure that auto-generated resource names are unique. Right now we use the type of the resource. We can add some number to it.
2. Use the variable names as the default resource names - this is also not perfect because we can create un-named resources, like so:
	 ```ts
	 bring cloud;
	 new cloud.Function(inflight () => {});
	 new cloud.Function(inflight () => {});
     ```
3. Catch these errors in the LSP too and add info there. Also not perfect because you still expect the compiler to give a good error.

I think the best solution is #2, with taking care of the edge case of nameless resources for which we'll implement #1.

### Community Notes

Chriscbr:
Improves the diagnostics for duplicate construct IDs by providing extra information when the error is thrown by the underlying `constructs` library.

Fixes #2623

Now, when you compile Wing code like this:

```
bring cloud;

let b = new cloud.Bucket();
let b2 = new cloud.Bucket();
```

Users will see an error like this:

Screenshot 2023-06-02 at 3 56 39 PM

## Checklist

- [x] Tests added (always)


apps/wing/src/commands/compile.ts
```diff
@@ -280,6 +284,28 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
   return synthDir;
 }
 
+function annotatePreflightError(e: any): any {
+  if (!(e instanceof Error)) {
+    return e;
+  }
+
+  if (e.message.startsWith("There is already a Construct with name")) {
```
eladb, line 292: Yak

I was wondering if perhaps we can deal with this at the compiler level but that's good enough for now...

eladb (changes requested):
Missing test...

apps/wing/src/commands/compile.ts
Chriscbr: Curious what you had in mind?

I couldn't think of any solutions based in wingc that would hold water after you start introducing dynamic IDs:

```
let x = 1;
new cloud.Bucket() as "bucket-${x}";
new cloud.Bucket() as "bucket-${x}";
```

and/or dynamic scopes:

```
new cloud.Bucket() in scope1;
new cloud.Bucket() as scope2;
```

There might be heuristics that catch some cases, but I think the risk is that having errors that only _sometimes_ appear as red underlines in VS Code would feel bad to use as a developer.

One idea I had for surfacing errors better was that if you haven't edited your code for 500ms or so, the VS Code extension might be able to compile your app to the sim target in the background and then highlight any preflight errors in your editor, but I think that might require source maps.

Chriscbr: I'll mark this conversation as resolved, but feel free to reopen or add a comment on the issue anything else comes to mind

MarkMcCulloh (approved):

apps/wing/src/commands/compile.ts
```diff
@@ -280,6 +284,28 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
+    const newMessage = [];
+    newMessage.push(e.message);
+    newMessage.push(
+      "hint: Every preflight class needs a unique id within its scope. You can assign an id to a class instance as shown:"
```
eladb, line 296:
```suggestion
      "hint: Every preflight object needs a unique identifier within its scope. You can assign one as shown:"
```

eladb (approved):

PR #2774, opened 2023-06-03
feat(sdk): adding memoization to awscdk and tf-aws targets

marciocadev:
## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/05-reference/wingsdk-api.md
+++ b/docs/05-reference/wingsdk-api.md
@@ -2223,6 +2223,7 @@ let FunctionProps = cloud.FunctionProps{ ... }
 | **Name** | **Type** | **Description** |
 | --- | --- | --- |
 | <code><a href="#@winglang/sdk.cloud.FunctionProps.property.env">env</a></code> | <code>MutMap&lt;str&gt;</code> | Environment variables to pass to the function. |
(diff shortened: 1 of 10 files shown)


eladb (commented):
Hey @marciocadev, can you elaborate on the use case for this feature? Not sure I fully understand when people would want to use it...

I am not sure this belongs in the standard library... maybe an example?

https://github.com/winglang/examples

marciocadev:
> Hey @marciocadev, can you elaborate on the use case for this feature? Not sure I fully understand when people would want to use it...
> 
> I am not sure this belongs in the standard library... maybe an example?
> 
> https://github.com/winglang/examples

Hey @eladb, this feature is widely used to avoid consecutive extensive processing, but the use case that led me to implement it is the cases where we shouldn't process duplicate payloads.

I work at a bank, and we use this approach to prevent duplicate money transfers when there is a failure in the processing steps. This way, even if a function receives the same payload, we ensure that it is executed only once.

In general, these are specific cases for using this functionality.

I liked the way we can implement it in Wing because it becomes extremely simple for the developer. The AWS Lambda powertools for Python has this functionality, but it requires creating DynamoDB and granting permission for the Lambda to use it. Then you have to configure the Lambda for usage.

In Wing, basically, we can add a parameter when creating a cloud.Function, and all the configuration steps are generated without the developer having to worry about the infrastructure.

I will try to create an example and put it in the repository you mentioned. I think an example explaining the motivation and the benefits in detail would be really cool.

PR #2809, opened 2023-06-06
chore: improve sdk doc generation performance by 8x

Chriscbr:
The main bottleneck while running `jsii-docgen` to generate the SDK API Reference today is time spent copying all of `node_modules` into a temporary directory. The reason it did this was for a process called "transliteration" where it would use `jsii-rosetta` to translate TypeScript code snippets inside docstrings into code snippets of other languages, like Python and Java, and `jsii-rosetta` happens to pollute your node_modules (not sure the exact reason of this).

Anyway, we don't need to do any transliteration to generate our docs, so we can just skip that step along with all of the file-copying associated with it. This improves the build time on my M1 Mac from around 1m5s to 7.5s.

No new tests since there shouldn't be any changes in behavior.

## Checklist

- [x] Tests added (always)


staycoolcall911 (approved):
Great performance improvement!

Issue #2731, opened 2023-05-31
Adding Api Gateway sdk tests


revitalbarletz:
### Feature Spec

The test will be written in Wing.
it will cover the following methods of the Ai Gateway resource:

- [x] options 
- [x] head
- [x] connect
- [x] HttpMethod.CONNECT
- [x] HttpMethod.OPTIONS
- [x] HttpMethod.HEAD
- [x] ApiResponse.headers

### Use Cases

testing

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.23.14](https://github.com/winglang/wing/releases/tag/v0.23.14).

PR #2826, opened 2023-06-06
feat(sdk): native fetch

tsuf239:
fixes https://github.com/winglang/wing/issues/2169

* adding a static `Http` class to `wingsdk.util`
* replacing the use of external fetch in `api_path_vars.w` file
* removing   `api_path_vars.js` -contains the js fetch

left todo:
- [x] adding more tests
- [ ] Updating docs (prefer to do it after the implementation is reviewed)

## Checklist

- [ ] Tests added (always)


examples/tests/valid/api_path_vars.w
```diff
@@ -1,10 +1,8 @@
 bring cloud;
+bring util;
```
eladb, line 2: This should be `bring http`, not `bring util`.

examples/tests/valid/api_path_vars.w
```diff
@@ -15,13 +13,13 @@ let handler = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
 
 api.get("/users/{name}", handler);
 
-let f = new Fetch();
 
 test "test" {
   let username = "tsuf";
-  // TODO: change f.get to static when possible
-  let res = f.get("${api.url}/users/${username}");
+  let res: util.Response = util.Http.get("${api.url}/users/${username}");
```
eladb, line 19:
```suggestion
  let res: util.Response = http.get("${api.url}/users/${username}");
```

libs/wingsdk/src/util/fetch.ts
```diff
@@ -0,0 +1,260 @@
+import { Code, InflightClient } from "../core";
```
eladb, line 1: This file should be under `src/http/fetch.ts` inside a class named `Util`:

`src/http/fetch.ts`:

```js
export class Util {
  public async get(...);
  public async fetch(...);
}
```

The `Util` class name is [special] and will automatically turn into a "global namespace" for the submodule.

[special]: https://github.com/winglang/wing/blob/a69af8e0451101770dda4d240d9521c4cc31191a/libs/wingc/src/type_check.rs#L3136

libs/wingsdk/src/util/fetch.ts
```diff
@@ -0,0 +1,260 @@
...
+  headers: {},
+  cache: RequestCache.DEFAULT,
+  redirect: RequestRedirect.FOLLOW,
+};
+/**
+ * the Http class is used for calling different HTTP methods and requesting and sending information online,
+ *  as well as testing public accessible resources
+ * @inflight
+ */
+export class Http {
```
eladb, line 150:
```suggestion
export class Util {
```

eladb (changes requested):

tsuf239:
It seems like the body needs to be parsed twice (JSON.parse(JSON.parse(response.body))) in order to get its value into Json, in tf-aws, will fix in the current or following PR 🙂

tsuf239:
> It seems like the body needs to be parsed twice (JSON.parse(JSON.parse(response.body))) in order to get its value into Json, in tf-aws, will fix in the current or following PR 🙂

Fixed. found the other stringifying method :)

MarkMcCulloh:
Something to note: we should raise the required node version in our packages/docs from just "18" to "8.13.0", since fetch was experimental until that version. Otherwise users will get a scary warning about it.

eladb:
> Something to note: we should raise the required node version in our packages/docs from just "18" to "8.13.0", since fetch was experimental until that version. Otherwise users will get a scary warning about it.

That's a great point. Maybe even raise to the latest LTS 18.16.1?

libs/wingsdk/test/http/http.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  Util as Http,
+  HttpMethods,
+  Response,
+  RequestCache,
+  RequestRedirect,
+} from "../../src/http/http";
+
+let defaultOptions = {
+  method: HttpMethods.GET,
+  headers: {},
```
eladb, line 12: Can we please write this test with wing instead of typescript?

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * The cache mode you want to use for the request.
+   */
+  readonly cache?: RequestCache;
+  /**
+   * he redirect mode to use: follow, error. The default is follow.
+   */
+  readonly redirect?: RequestRedirect;
+  /**
+   * A string specifying "no-referrer", client, or a URL. The default is "about:client".
```
eladb, line 107: Is that really the default?

eladb (changes requested):

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
+import { Code, InflightClient } from "../core";
+
+/**
+ * RequestCache-
+ * The cache read-only property that contains the cache mode of the request.
+ * It controls how the request will interact with the browser's HTTP cache.
+
```
Chriscbr, line 7:
```suggestion
 * The cache mode of the request.
 * It controls how a request will interact with the system's HTTP cache.
```

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+ */
+export enum RequestCache {
+  /**
+   * The browser looks for a matching request in its HTTP cache.
+   * If there is a match and it is fresh, it will be returned from the cache.
```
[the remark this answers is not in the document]
Chriscbr: Sure thing. I just did some testing, and for showing bullets in our docs, the list needs to (1) start with an empty line before it, and (2) each start with a `*` character. For example:

```js
/**
 * Here is my description, blah blah
 *
 * * This is my first bullet point. The sentence may go on
 * for multiple lines.
 * * This is my second bullet point. Our documentation can
 * go on forever!
 *
 * More content can follow if I want.
 */
```

If we follow these rules then our `jsii-docgen` tool seems to not get confused. :)

@tsuf239 If you try running `npm run compile` and `npm run docgen` from the `wingsdk` folder and check the output, I think that's the easiest way I think to see how docs / bulleted items are rendered.

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
+import { Code, InflightClient } from "../core";
+
+/**
+ * RequestCache-
+ * The cache read-only property that contains the cache mode of the request.
+ * It controls how the request will interact with the browser's HTTP cache.
+
+ */
+export enum RequestCache {
```
Chriscbr, line 9: The docs information here is great. One thing we should update is since we don't know whether the code will be running on a browser or on a cloud serverless function or elsewhere, let's avoid using the term "browser" and instead use "system" or "runtime environment" throughout the descriptions.

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * Follow all redirects incurred when fetching a resource.
+   */
+  FOLLOW = "follow",
+  /**
+   * Return a network error when a request is met with a redirect.
+   */
+  ERROR = "error",
+}
+/**
```
Chriscbr, line 56: Nit: add a little breathing room between each new class or enum (helps with readability, and in some editors like `vim` they make it easy for users to paginate through definitions if they are separated with newlines)
```suggestion
}

/**
```

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * The URL of the response.
+   */
+  readonly url: string;
+  /**
+   * A boolean indicating whether the response was successful (status in the range 200 – 299) or not.
+   */
+  readonly ok: boolean;
+  /**
+   * The Headers object associated with the response.
```
Chriscbr, line 128:
```suggestion
   * The map of header information associated with the response.
```

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+ */
+export class Util {
+  /**
+   * @internal
+   */
+  public static _toInflightType(): Code {
+    return InflightClient.forType(__filename, this.name);
+  }
+  /**
+   * Sends an http request of any method.
```
Chriscbr, line 158:
```suggestion
   * Sends an HTTP request to a URL, returning a fulfilled or failed response.
   * @throws Only throws if there is a networking error
```

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+      ...options,
+      method: HttpMethods.DELETE,
+    });
+  }
+
+  private static async _formatResponse(
+    response: globalThis.Response
+  ): Promise<Response> {
+    const headers: Record<string, string> = {};
+    response.headers?.forEach((val: string, key: string) => {
```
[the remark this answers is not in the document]
Chriscbr: I think this is needed because the Fetch API returns a custom Headers object - looks ok to me.

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,259 @@
...
+      headers[key] = val;
+    });
```
Chriscbr, line 249: style: Prefer for loop
```suggestion
    // convert Headers object into a plain JS object
    const headers: Record<string, string> = {};
    for (const [key, value] of response.headers.entries()) {
      headers[key] = val;
    }
```

tsuf239:
> > Something to note: we should raise the required node version in our packages/docs from just "18" to "8.13.0", since fetch was experimental until that version. Otherwise users will get a scary warning about it.
> 
> That's a great point. Maybe even raise to the latest LTS 18.16.1?

Cool! Where is it written in the code?

examples/tests/valid/api_path_vars.w
```diff
@@ -1,28 +1,26 @@
...
 
-class Fetch {
-  extern "./api_path_vars.js" inflight get(url: str): Json;
-}
 
 let handler = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
   return cloud.ApiResponse {
     body: Json.stringify({ user: req.vars.get("name") }),
-    headers: { "content-type": "application/json" },
+    headers: { "content-type": "text/plain" },
```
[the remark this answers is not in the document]
tsuf239: restoring now, it was part of the experiments to fix the double stringifying

libs/wingsdk/src/http/http.ts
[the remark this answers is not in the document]
tsuf239: exactly :)
https://developer.mozilla.org/en-US/docs/Web/API/Headers

libs/wingsdk/src/http/http.ts
tsuf239: it's all taken from here: 
https://developer.mozilla.org/en-US/docs/Web/API/Request/Request

libs/wingsdk/src/http/http.ts
tsuf239: I wish, but it seems like it doesn't have entries :(
image

libs/wingsdk/src/http/http.ts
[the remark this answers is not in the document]
tsuf239: I don't see it generated, I think the folder isn't included, should I do something regarding that? 
(This file is under `src/http`)
https://github.com/winglang/wing/blob/24d91917ea2fd9dd482548dc449936aa81d8854f/libs/wingsdk/.projenrc.ts#L203

libs/wingsdk/test/http/http.test.ts
tsuf239: yes, I'll add a wing test :) 
 also, since I'll have to use a real API resource- those can cover also the api's SDK spec tests! #2731

libs/wingsdk/src/http/http.ts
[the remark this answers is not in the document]
tsuf239: it's all taken from there :)
(and from the fetch spec: https://fetch.spec.whatwg.org/#fetch-method)
Is there something specific that needs to be changed?

libs/wingsdk/src/http/http.ts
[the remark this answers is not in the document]
Chriscbr: @tsuf239 ah, you'll need to add the module here (sorry its a bit hacky at the moment): https://github.com/winglang/wing/blob/main/apps/jsii-docgen/src/docgen/view/wing-filters.ts

examples/tests/sdk_tests/api/delete.w
```diff
@@ -0,0 +1,38 @@
...
+api.delete("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
+    assert(req.method == api_DELETE);
+    assert(req.query?.get("all") == "true");
+    assert(req.query?.get("page") == "6");
+    assert(req.path == "/path");
+
+    return cloud.ApiResponse {
+        status: 200,
+        body: req.query?.get("page")
+    };
```
Chriscbr, line 19: Should be indented 2 spaces for consistency with our other examples (someday we will have a Wing formatter 🤞)

examples/tests/sdk_tests/api/delete.w
```diff
@@ -0,0 +1,38 @@
+
+bring cloud;
```
Chriscbr, line 2: Start examples on first line
```suggestion
bring cloud;
```

libs/wingsdk/src/http/http.ts
```diff
@@ -0,0 +1,267 @@
...
+   */
+  readonly ok: boolean;
+  /**
+   * The map of header information associated with the response.
+   */
+  readonly headers: Record<string, string>;
+  /**
+   * A string represents the body contents.
+   */
+  readonly body: string;
```
Chriscbr, line 138: I wonder if this should be optional, to handle the case where some responses may not have a body (for example, HEAD requests)

Chriscbr (approved):

libs/wingsdk/src/http/http.ts
tsuf239: image

libs/wingsdk/src/http/http.ts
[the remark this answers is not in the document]
tsuf239: done, thanks @Chriscbr!

examples/tests/sdk_tests/api/delete.w
```diff
@@ -0,0 +1,37 @@
...
+bring http;
+
+let http_DELETE = http.HttpMethod.DELETE;
```
eladb, line 4: Why is this needed?

eladb (approved):

examples/tests/sdk_tests/api/delete.w
tsuf239: https://winglang.slack.com/archives/C048QCN2XLJ/p1687425563136599
Opened a bug: https://github.com/winglang/wing/issues/3049
I'll add the link to those files

PR #2835, opened 2023-06-07
feat(sdk): fixing public url for private aws bucket

### LINKED ISSUE #3112: tryGet is not asking for permissions
### I tried this:

```ts
bring cloud;
bring util;

let origin = new cloud.Bucket() as "origin";
let dest = new cloud.Bucket() as "dest";

origin.onCreate(inflight (key: str): void => {
  let s = origin.get(key);
  dest.put(key, s.uppercase());
});

test "main" {
  origin.put("file", "some");
  assert(util.waitUntil((): bool =>{
    return dest.tryGet("file") == "SOME";
  }));
}
```

### This happened:

The test failed on aws because of permissions, the iam rules where not granted for the lambda function because `tryGet` is not part of `calculateBucketPermissions` tf-aws bucket function

### I expected this:

to pass

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

The problem is that tryGet is not part of calculateBucketPermissions

 

### Community Notes

tsuf239:
* **fixing a bug supplying public url for private buckets on aws** - we were passing it as a string in the inflight constructor, and the better approach is to query the bucket itself via aws-sdk's `GetPublicAccessBlockCommand` (also better for a case we'll be able to change the access level of a bucket in runtime in the future...)
* adding fetch to verify the bucket public url
* UPDATE: added missing permissions for exists, tryGet, tryDelete, tryGetJson, and made sure the tests are passing 
* fixes #3112

## Checklist

- [x] Tests added (**updated**)


examples/tests/sdk_tests/bucket/public_url.w
```diff
@@ -9,13 +10,15 @@ test "publicUrl" {
   publicBucket.put("file1.txt", "Foo");
   privateBucket.put("file2.txt", "Bar");
 
-  assert(publicBucket.publicUrl("file1.txt") != "");
+  let publicUrl = publicBucket.publicUrl("file1.txt");
+  assert(publicUrl != "");
+  // TODO: works in aws, doesn't work in sime since publicUrl is returnning a path to the file, uncomment when #2833 is resolved.
```
eladb, line 15: Maybe instead of commenting out, check if `util.env("WING_TARGET") == "sim"`

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -200,15 +201,34 @@ export class BucketClient implements IBucketClient {
...
+    );
+  }
 
   /**
    * Returns a url to the given file.
    * @Throws if the file is not public or if object does not exist.
    */
   public async publicUrl(key: string): Promise<string> {
-    if (!this._public) {
+    if (!(await this.checkIfPublic())) {
```
eladb, line 228: I am wondering: instead of rendering the url in line 241, can we just return the url the api sends back?

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -200,15 +201,34 @@ export class BucketClient implements IBucketClient {
...
+      !resp.PublicAccessBlockConfiguration?.RestrictPublicBuckets &&
+      !resp.PublicAccessBlockConfiguration?.IgnorePublicAcls
+    );
+  }
 
   /**
    * Returns a url to the given file.
    * @Throws if the file is not public or if object does not exist.
    */
   public async publicUrl(key: string): Promise<string> {
```
eladb, line 227: Do we need to add permissions for this in the IAM policy?

eladb (changes requested):

libs/wingsdk/src/shared-aws/bucket.inflight.ts
tsuf239: I looked for it, and I couldn't see any direct reference to that online, only for a signed URL, do you have an idea how to do it?  
`GetPublicAccessBlockCommand` returns this type: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/PublicAccessBlockConfiguration/ 
which just contains booleans...

eladb: Got it. OK, so the current implementation is reasonable.

libs/wingsdk/src/shared-aws/bucket.inflight.ts
tsuf239: yes, I added those, and also fixed the permissions of the other methods (as the tests were failing)

libs/wingsdk/src/shared-aws/permissions.ts
```diff
@@ -101,10 +101,35 @@ export function calculateBucketPermissions(
...
+        "s3:GetBucketPublicAccessBlock",
+      ],
+      resources: [arn, `${arn}/*`],
+    });
+  }
+
+  if (ops.includes(cloud.BucketInflightMethods.TRY_DELETE)) {
+    policies.push({
+      actions: [
+        "s3:List*",
```
eladb, line 128: This shouldn't be required for `tryDelete`

libs/wingsdk/src/shared-aws/permissions.ts
```diff
@@ -101,10 +101,35 @@ export function calculateBucketPermissions(
...
+        "s3:DeleteObject*",
+        "s3:DeleteObjectVersion*",
+        "s3:PutLifecycleConfiguration*",
```
eladb, line 131: Ditto

libs/wingsdk/src/shared-aws/permissions.ts
```diff
@@ -101,10 +101,35 @@ export function calculateBucketPermissions(
...
+      actions: ["s3:GetObject*", "s3:GetBucket*", "s3:ListBucket"],
+      resources: [arn, `${arn}/*`],
+    });
+  }
+
+  if (ops.includes(cloud.BucketInflightMethods.PUBLIC_URL)) {
+    policies.push({
+      actions: [
+        "s3:GetBucket*",
+        "s3:ListBucket",
```
eladb, line 118: Why are these permissions required here?

eladb (changes requested):

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -138,6 +139,10 @@ export class BucketClient implements IBucketClient {
   public async delete(key: string, opts?: BucketDeleteOptions): Promise<void> {
     const mustExist = opts?.mustExist ?? false;
 
+    if (mustExist && !(await this.exists(key))) {
```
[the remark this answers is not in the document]
tsuf239: I tested it carefully and found out that it doesn't throw an error even if mustExist=true and the key doesn't exist on the bucket...

libs/wingsdk/src/shared-aws/permissions.ts
```diff
@@ -101,10 +101,35 @@ export function calculateBucketPermissions(
...
+        "s3:GetBucketPublicAccessBlock",
+      ],
+      resources: [arn, `${arn}/*`],
+    });
+  }
+
+  if (ops.includes(cloud.BucketInflightMethods.TRY_DELETE)) {
```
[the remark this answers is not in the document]
tsuf239: "try" functions are using the "exists" method permissions in addition to the delete permissions, it can be united (supplying inaccurate permissions to delete)

libs/wingsdk/src/shared-aws/permissions.ts
tsuf239: it required for calling `exists`

libs/wingsdk/src/shared-aws/permissions.ts
tsuf239: it's written here: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/command/DeleteObjectCommand/
I can try without it though

libs/wingsdk/src/shared-aws/permissions.ts
tsuf239: for checking if the key exists, will add comments :)

tsuf239:
refactored the permission calculation, tests are still working: 
image

libs/wingsdk/src/shared-aws/bucket.inflight.ts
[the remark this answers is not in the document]
eladb: Ok fair enough

eladb (approved):

PR #2838, opened 2023-06-07
fix: unable to send non-json api responses

### LINKED ISSUE #2840: SDK: API should return not only JSON
### I tried this:

Returning a plain text yaml file. 

### This happened:

The response is always stringified. 

```
"myyaml"
```

### I expected this:

An unmodified response body like

```myaml```

### Is there a workaround?

Doesn't look like it, headers don't have an effect since the response is always stringified anyway.

### Component

SDK

### Wing Version

0.19.2

### Wing Console Version

_No response_

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

### PR DESCRIPTION
**Pull request by eladb**
Change the signature of `body` in `ApiRequest` and `ApiResponse` to be `str` instead of `Json` to allow any value. JSON parse/stringify is now the responsibility of the handler.

To support this we basically tell express to parse requests as "raw".

## Resolves

* Fixes #2840

## Misc

* Added a test to verify that it is possible to respond without a body.
* Removed the `body` from the trace that's emitted by the simulator (could be huge).

## Checklist

- [x] Tests added (always)

### CODE REVIEW

**Review by staycoolcall911 — DISMISSED** [review]

File: libs/wingsdk/src/cloud/api.ts
Comment (ekeren) [inline libs/wingsdk/src/cloud/api.ts:1]: @eladb , we also need to change the req.body to be string (not just the response)

**Comment by eladb** [comment]
Yes this is very partial. Just started to sketch this

File: libs/wingsdk/src/cloud/api.ts
Comment (eladb) [reply libs/wingsdk/src/cloud/api.ts:1]: sure

File: examples/tests/valid/api.w [hunk 1]
```diff
@@ -13,7 +13,10 @@ let handler = inflight (request: cloud.ApiRequest): cloud.ApiResponse => {
   };
   
   let resp = cloud.ApiResponse {
-    body: bodyResponse,
+    body: Json.stringify(bodyResponse),
+    headers: {
+      "content-type": "application/json"
+    },
```
skorfmann, line 19: Probably a follow up, but couldn't there be something like `cloud.ApiJsonResponse` to cover the default json cases as before and have the generic `cloud.ApiResponse` to let the user fully control the response?

skorfmann (commented):
That was quick, looks good 👍

examples/tests/valid/api.w
eladb: Yes. We need to add some sugar APIs on top.

Chriscbr (approved):

PR #2848, opened 2023-06-07
chore(sdk): bucket.publicurl should return an url for target-sim

### LINKED ISSUE #2833: sim bucket.publicUrl should return a web server endpoint
### Feature Spec

Right now, on sim, bucket.publicUrl returns a path to the bucket file,
I think it can be good to align it to the cloud behavior of returning an actual URL.

### Use Cases

mainly for tests to act the same, but also for being able to access the file in the local network

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

garysassano:
Closes #2833


libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -17,21 +18,25 @@ import {
...
+    this._fileDir = fs.mkdtempSync(join(os.tmpdir(), "wing-sim-"));
     this.context = context;
     this.initialObjects = props.initialObjects ?? {};
     this._public = props.public ?? false;
     this.topicHandlers = props.topics;
   }
 
+  public get fileDir(): string {
+    return this._fileDir;
+  }
```
Chriscbr, line 38: I don't think this getter is used anywhere, so we can probably remove it?

Chriscbr: Oops, never mind I see

Chriscbr (approved):

libs/wingsdk/src/target-sim/bucket.inflight.ts
garysassano: Yeah, I added that getter because one of the tests was bypassing TypeScript checks and calling a private method, which I didn't like.

garysassano: Yeah, I added that getter because one of the tests was bypassing TypeScript checks and calling a private property, which I didn't like.

PR #2851, opened 2023-06-08
fix(compiler): panics not reported as diagnostics

yoav-steinberg:
See #1656

Now panics are reported as normal diagnostics:
`WINGC_DEBUG_PANIC=type-checking apps/wing/bin/wing test examples/tests/invalid/panic.w` will produce a standard wing diagnostic output.

image

The vscode extension also utilizes this now (you may test with `WINGC_DEBUG_PANIC=type-checking npx nx dev vscode-wing`).
 
image

## Checklist

- [ ] Tests added (always)


apps/wing/src/commands/lsp.ts
```diff
@@ -86,28 +103,43 @@ export async function run_server() {
...
-
+    raw_diagnostics.length = 0;
     callWing("wingc_on_did_open_text_document", params);
+    // purposely not awaiting this, notifications are fire-and-forget
+    void connection.sendDiagnostics({
+      uri: params.textDocument.uri,
+      diagnostics: raw_diagnostics.map((rd) => {
+        return Diagnostic.create(Range.create(rd.span.start.line, rd.span.start.col, rd.span.end.line, rd.span.end.col), rd.message)
+      })
+    });
```
MarkMcCulloh, line 120: Duplicated, worth extracting to separate function?

apps/wing/src/wingc.ts
```diff
@@ -172,14 +173,13 @@ export async function load(options: WingCompilerLoadOptions) {
     preopens,
   });
 
-  const importObject = {
+  const importObject = merge({
```
MarkMcCulloh, line 176: not sure it's worth introducing a package for this. Arguable the API should just be simpler here (we should accept an explicit function to set for new_diagnostic, not the entire import object).

libs/wingc/src/diagnostic.rs
```diff
@@ -172,6 +172,58 @@ impl PartialOrd for Diagnostic {
...
+/// Returns whether any errors were found during compilation
+pub fn found_errors() -> bool {
+	DIAGNOSTICS.with(|diagnostics| {
+		let diagnostics = diagnostics.borrow();
+		!diagnostics.is_empty()
+	})
+}
+
+/// Returns the list of diagnostics
+pub fn get_diagnostics() -> Vec<Diagnostic> {
```
MarkMcCulloh, line 211: nit: Possible to avoid cloning for our use cases?

libs/wingc/src/lsp/sync.rs
```diff
@@ -76,14 +76,18 @@ pub fn on_document_did_change(params: DidChangeTextDocumentParams) {
...
 		});
 	})
 }
 
 /// Runs several phases of the wing compile on a file, including: parsing, type checking, and capturing
 fn partial_compile(source_file: &str, text: &[u8], jsii_types: &mut TypeSystem) -> FileData {
+	// Reset diagnostics before new compilation (`partial_compile` can be called multiple)
+	reset_diagnostics();
+	// Set custom panic hook before compilation (TODO: this is redundant in subsequent calls to `partial_compile`)
+	set_custom_panic_hook();
```
MarkMcCulloh, line 89: It'd be worthwhile to expose an init function to js to invoke this

apps/wing/src/wingc.ts
yoav-steinberg: Originally I thought I might have more than one import function in the env. So I needed merge to avoid some ugliness here. But since I got rid of `send_notification` I think I can revert this.

libs/wingc/src/diagnostic.rs
yoav-steinberg: not sure how, I don't think I can "move out" of the thread local storage and the users here need a clone anyway (the diags are sorted or copied into FileData ...).

libs/wingcompiler/src/errors/compile-error.ts
```diff
@@ -1,7 +1,7 @@
 import { WingDiagnostic } from "../wingc";
 
 export class CompileError extends Error {
-  constructor(message: string, public readonly diagnostics: WingDiagnostic[]) {
-    super(message);
+  constructor(public readonly diagnostics: WingDiagnostic[]) {
+    super("comliation error");
```
Chriscbr, line 5:
```suggestion
    super("compilation error");
```

libs/wingc/src/jsify.rs
```diff
@@ -182,18 +180,9 @@ impl<'a> JSifier<'a> {
...
-		}
-	}
-
-	/// Write all files to the output directory
-	pub fn emit_files(&mut self, out_dir: &Path) {
-		match self.files.emit_files(out_dir) {
-			Ok(()) => {}
-			Err(err) => self.diagnostics.push(err.into()),
-		}
+==== BASE ====
```
staycoolcall911, line 183: Remove
```suggestion
```

libs/wingc/src/jsify.rs
```diff
@@ -182,18 +180,9 @@ impl<'a> JSifier<'a> {
...
+		output.to_string()
+==== BASE ====
```
staycoolcall911, line 185: Remove
```suggestion
```

apps/wing/src/commands/compile.ts
```diff
@@ -64,7 +53,7 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
         // file_id might be "" if the span is synthetic (see #2521)
         if (span !== null && span.file_id) {
           // `span` should only be null if source file couldn't be read etc.
-          const source = await fsPromise.readFile(span.file_id, "utf8");
+          const source = readFileSync(span.file_id, "utf8");
```
MarkMcCulloh, line 56: Why this change? I think we should actually prefer to keep things async here as much as possible because this compiler is possibly used in other interactive settings where we don't want to block the main thread.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1314,10 +1319,7 @@ Duration <DURATION>
 "
 `;
 
-exports[`panic.w 2`] = `
-"Compiler bug when type-checking ../../../examples/tests/invalid/panic.w:6:1 | panicked at 'User invoked panic', libs/wingc/src/type_check.rs:LINE:COL
-Please report this bug at https://docs.winglang.io/contributors/bugs"
-`;
+exports[`panic.w 2`] = `"Panicked, backtrace not captured: Unsupported"`;
```
MarkMcCulloh, line 1322: Is this a regression?

MarkMcCulloh (approved):

apps/wing/src/commands/compile.ts
yoav-steinberg: The change is because there was a version that this was in a callback called from within WASM (a wasm import). But after (the rather complex) merge with Cristian's wingcompiler lib PR it bounced back into being called from TS, so I'll revert.

libs/wingc/src/diagnostic.rs
```diff
@@ -172,6 +172,58 @@ impl PartialOrd for Diagnostic {
...
+}
+
+/// Report a compilation diagnostic
+pub fn report_diagnostic(diagnostic: Diagnostic) {
+	// Add the diagnostic to the list of diagnostics
+	DIAGNOSTICS.with(|diagnostics| {
+		diagnostics.borrow_mut().push(diagnostic.clone());
+	});
+
+	// If we're running in wasm32 then send the diagnostic to the client
```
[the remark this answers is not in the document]
yoav-steinberg: In all real life cases we're currently running in wasm32 (vscode extension, wing command line, console). In the single case where we aren't (the rust compile example) the diagnostics are are read directly with the `get_diagnostics` method in this class.

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: No, notice this is the stderr output. Now panics go out as normal diags which in the context of invalid tests are printed to stdout so this output now moved to stdout above (with a slight change in phrasing).
I added an automatic stack trace to stderr in case of panics for the benefit of reporting, but in practice these aren't supported in WASM (unless I'm missing something). So all we get in stderr is a notice stack traces aren't supported, if they ever will be, running the compiler with `RUST_BACKTRACE=1` should produce the trace here (it does when using native code).

Issue #2548, opened 2023-05-16
move all diagnostics out of jsification so they appear in the LSP


yoav-steinberg:
Move all diagnostics out of jsification so they appear in the LSP.
Currently the LSP (and I think rightfully so) doesn't perform the compiler's output (jsify) phase. So diagnostics added at this point aren't available in the vs code extension. These are some of the coolest diagnostics we have (can't reference a preflight mutable array inflight, for example). Ideally we should have a phase that "prepares" all the info we need for output before running the output phase (similar to the capture pahse we use to have). It'll also generate diags and we can run it in the LSP.

note here: https://github.com/winglang/wing/pull/2411#discussion_r1194155812

staycoolcall911:
I think this comes down to 1 diagnostic - FYI @yoav-steinberg

yoav-steinberg:
The capturing phase moved to be pre-jsify and also the jsify is now being run in the LSP as well. I think we can safely close this issue.

PR #2866, opened 2023-06-09
chore(compiler): jsify in memory

Chriscbr:
Progress towards #2548 - this PR refactors the jsifier stage so that errors can be emitted without having to generate files.

## Checklist

- [x] Tests added (always)


Chriscbr:
@MarkMcCulloh I wanted to add the jsify step to the LSP, but I ran into an issue since the main compiler expects an absolute project directory, but I wasn't sure if it's made available in the LSP. Do you know if there's a way to get this information?

eladb (approved):
LGTM, although I am not sure why emitting files is a Bad Thing™️

Chriscbr:
@eladb When the LSP recompiles your code in the background, there shouldn't be any need for it to emit intermediate JS files (it would just slow it down). I think it's actually a really nice/clean separation of concerns

MarkMcCulloh:
> there's also no harm in that

There is, fs operations have a cost that is effectively wasted in certain contexts. It also adds a new failure category (io error).

> at certain scale, keeping all of our output in-memory might actually be a problem and we will end up having to emit output files.

I think the true solution is that emitting shouldn't do any analysis and should only be responsible for emitting. That way this change wouldn't even have been needed in the first place because. I'm sure we'll get there at some point.

> compiler expects an absolute project directory, but I wasn't sure if it's made available in the LSP. Do you know if there's a way to get this information?

@Chriscbr iirc, the path of the wing file is absolute. So the "project directory" would be the directory of that file. We also have access to the root of the "workspace", the lsp provides that in the [initializer](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams) as well.

Issue #1375, opened 2023-01-31
counter increment named value 


ekeren:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

You can increment/decrement/peek value with `str` key, allowing you to dynamically define counters in `inflight`

The `inc`, `dec`, `peek` functions should accept a CounterKeyProp argument.
The CounterKeyProp currently has only one member `key: str;`

```ts (wing) 
bring cloud;
let counter = new cloud.Counter();
let counter_with_initial = new cloud.Counter(initial: 100);

new cloud.Function(inflight (s:str): str => {
  assert(counter.inc(1) == 0);
  assert(counter.inc(11, key: "k") == 0);
  assert(counter.inc(0, key: "k2") == 0);
  assert(counter.peek() == 1);
  assert(counter.peek(key: "k") == 11);
  assert(counter.dec(1) == 1);
  assert(counter.dec(11, key: "k") == 11);
  assert(counter.peek() == 0);
  assert(counter.peek(key: "k") == 0);

  assert(counter_with_initial.inc(1) == 100);
  assert(counter_with_initial.inc(11, key: "k") == 100);
  assert(counter.inc(0, key: "k2") == 100);
  assert(counter_with_initial.peek() == 101);
  assert(counter_with_initial.peek(key: "k") == 111);
  assert(counter_with_initial.dec(1) == 101);
  assert(counter_with_initial.dec(11, key: "k") == 111);
  assert(counter_with_initial.peek() == 100);
  assert(counter_with_initial.peek(key: "k") == 100);
}) as "test counter";
```

### Use Cases

You can use counter with dynamic keys, allowing you to count things that are not known at compilation time. 

For example:
```ts (wing)
let contributors = new cloud.Counter();

let github_handler = inflight (github_username: str): void => {
  let prev_contributions = contributors.inc(1, github_username: github_username);
  print("user ${github_username} has ${prev_contributions + 1} contributions");
}; 

//TODO connect github_handler to github webhook
```

### Component

SDK

Chriscbr:
Random thoughts / questions:
- What should the expected behavior be if you call `peek` or `inc` on a key that hasn't been used before?
- Should there be a way to list all keys?
- Should there be a way to `set` or `reset` a counter to a particular value (or to the counter's `initial_value`)?

ekeren:
1.
[...]

PR #2875, opened 2023-06-11
feat(sdk): adding key argument to counter

### LINKED ISSUE #1286: Tracking issue: work left todo to implement the SDK Spec

marciocadev:
Adding the key argument to the inc, dec, peek, and set methods of the Counter, in this way, the same counter can be used to control different values according to the desired key.

Closes #1286 

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-sim/counter.inflight.ts
+++ b/libs/wingsdk/src/target-sim/counter.inflight.ts
@@ -6,14 +6,14 @@ import {
 } from "../testing/simulator";
 
 export class Counter implements ICounterClient, ISimulatorResourceInstance {
-  private value: number;
+  private values: Map<string, number>;
   private readonly context: ISimulatorContext;
 
   public constructor(
     props: CounterSchema["props"],
     context: ISimulatorContext
   ) {
-    this.value = props.initial;
+    this.values = new Map().set("default", props.initial);
     this.context = context;
   }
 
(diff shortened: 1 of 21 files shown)


staycoolcall911:
Hey @marciocadev , I think this should close #1375, not #1286

marciocadev:
Wow, sorry @staycoolcall911 my mistake 🤦🏻‍♂️

libs/wingsdk/src/target-sim/counter.inflight.ts
```diff
@@ -23,42 +23,58 @@ export class Counter implements ICounterClient, ISimulatorResourceInstance {
 
   public async cleanup(): Promise<void> {}
 
-  public async inc(amount: number = 1): Promise<number> {
+  public async inc(
```
revitalbarletz, line 26: It feels more intuitive for me that the key will be the first parameter, what do you think? is it too complex to switch the order?

libs/wingsdk/test/shared-aws/counter.inflight.test.ts
```diff
@@ -55,6 +55,44 @@ test("inc(5)", async () => {
...
+    responseValue: prevValue + 1,
+    key: "my-key",
+  });
+
+  // WHEN
+  const client = new CounterClient(MOCK_TABLE_NAME);
+  const response = await client.inc(undefined, "my-key");
+
+  // THEN
+  expect(response).toEqual(prevValue); // returns previous value
```
revitalbarletz, line 74: Why prevValue and not prevValue + 1?

libs/wingsdk/src/target-sim/counter.inflight.ts
marciocadev: I understand what you said, Revital, but I had doubts about it because currently the standard way to add 1 to the counter is `counter.inc()`, which wouldn't be affected. However, if I wanted to add more than the default value of 1, currently it is `counter.inc(2)`, and it would become something like `counter.inc(nil, 2)`.

Where the default id wouldn't need to be part of the call, but since the id is the first parameter, I would have to pass the value `nil`.

libs/wingsdk/test/shared-aws/counter.inflight.test.ts
marciocadev: Because `counter.inc()` returns the previous value

The current value can be recovery with `counter.peek()`

If you run this code
```ts (wing)
bring cloud;

let c = new clount.Counter();

let var x = c.inc();
log("${x}"); // print 0
log("${c.peek()}"); //print 1
```

libs/wingsdk/test/shared-aws/counter.inflight.test.ts
```diff
@@ -55,6 +55,44 @@ test("inc(5)", async () => {
...
+    responseValue: 887 + 5,
+    key: "my-key",
+  });
+
+  // WHEN
+  const client = new CounterClient(MOCK_TABLE_NAME);
+  const response = await client.inc(5, "my-key");
+
+  // THEN
+  expect(response).toEqual(prevValue); // returns previous value
```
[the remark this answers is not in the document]
marciocadev: `inc` increment the counter and return the previous value

staycoolcall911 (approved):

PR #2893, opened 2023-06-13
chore: fix snapshot sanitization for timestamps

skyrpex:
Updates the regex that sanitizes snapshot timestamps so it includes negative timestamps, such as `-1m-0.123s`.

See an example of previous jobs failing: https://github.com/winglang/wing/actions/runs/5250503621/jobs/9484788227.


staycoolcall911 (approved):

staycoolcall911:
On second thought, test duration of -1m doesn't make sense, so we didn't want that to fail silently.
Instead we'll monitor if this happens again and address the root cause.
@tsuf239 will look into it next week.

PR #2908, opened 2023-06-13
chore(compiler): add 🗺️ expression for debugging symbol environments

Chriscbr:
This PR introduces another debugging utility for our compiler. Given code like this:

```js
bring cloud;

class A {
  f: num;
  init(b: cloud.Bucket) {
    this.f = 5;
    🗺️;

    let n = b.node.id;
  }
}

new A(new cloud.Bucket());
```

Compiling this will print to the terminal:

```
[symbol environment at example.w:7:5]
level 0: { __parent_this => A, b => Bucket, this => A }
level 1: { A => A [type], assert => (condition: bool): void, cloud => cloud [namespace], log => (message: str): void, panic => (message: str): void, std => std [namespace], throw => (message: str): void }
```

## Checklist

- [x] Tests added (always)


libs/tree-sitter-wing/grammar.js
```diff
@@ -370,6 +371,7 @@ module.exports = grammar({
       prec.right(PREC.OPTIONAL_TEST, seq($.expression, "?")),
 
     compiler_dbg_panic: ($) => "😱",
```
eladb, line 373: As we have more of these, I am wondering if we should put these under some env flag (`DEBUG=1`)...

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -27,6 +30,36 @@ pub struct SymbolEnv {
 	statement_idx: usize,
 }
 
+impl Display for SymbolEnv {
```
eladb, line 33: Useful generally!

eladb (approved):

tools/hangar/__snapshots__/test_corpus/valid/debug_env.w_test_sim.md
```diff
@@ -0,0 +1,18 @@
+# [debug_env.w](../../../../../examples/tests/valid/debug_env.w) | test | sim
+
+## stdout.log
+```log
+[symbol environment at /Users/chrisr/dev/wing3/examples/tests/valid/debug_env.w:7:5]
```
eladb, line 5: This will likely need to be redacted:

```suggestion
[symbol environment at <REDACTED>/debug_env.w:7:5]
```

eladb (commented):
Where did you dig this emoji from?

yoav-steinberg:
> Where did you dig this emoji from?

Don't know where Chris got it form it seems better than these:
image

I do however have the feeling that we'll need to provide some more conventional way to writing these debug commands and perhaps keep the special chars as aliases. If have a bunch of these they won't be descriptive enough. (The code might look like a screenshot out of my teenage daughter's messaging app).

libs/tree-sitter-wing/grammar.js
Chriscbr: Sounds like a sensible idea, I'll leave it as a future improvement

libs/tree-sitter-wing/grammar.js
```diff
@@ -320,6 +320,7 @@ module.exports = grammar({
         $.struct_literal,
         $.optional_test,
         $.compiler_dbg_panic,
+        $.compiler_dbg_env,
```
yoav-steinberg, line 323: note that in this case I'd go for a special statement and not a special expression. I wanted the expression option for `compiler_dbg_panic` so I can cause panics within an expression evaluation. But here I think we can simplify things if we go for a statement which doesn't need to evaluate to any type and can't be used within another expression.

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -27,6 +30,36 @@ pub struct SymbolEnv {
...
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
yoav-steinberg, line 34: Not critical for this PR, but would be very useful to add some colorization for this (we already have the colorized crate). I suspect it'll be much easier to look at with colors (which will be automatically emitted on non tty outputs, so you don't need to worry about that).

libs/tree-sitter-wing/grammar.js
Chriscbr: Mmmm yeah that makes more sense. Evaluating sub-expressions should never change the environment so there's not really a point nesting it deeply inside expressions

tools/hangar/__snapshots__/test_corpus/valid/debug_env.w_test_sim.md
Chriscbr: Fixed

PR #2909, opened 2023-06-13
feat: allow referencing properties of constructs

eladb:
It is now possible to reference properties of Terraform CDK constructs from inflight context. For example:

```js
bring "@cdktf/provider-aws" as aws;
let b = new aws.s3Bucket.S3Bucket();

inflight () => {
  log(b.arn);
};
```

In order to support this, we need the lifting and binding mechanism to have information about which properties are being accessed. To that end, I've extended this API to accept an additional `props` array with the names of properties being bound/lifted.

I changed how `FieldReferenceVisitor` so that if it finds a property reference (e.g. `this.bucket.arn` or `bucket.arn`), it will treat `arn` as the `op`, and we merge this information with the list of captured variables/fields.

Tweaked the `registerBind()` method to allow any `op` that matches a preflight property.

* Fixed a bug in how we lift tokens. CDKTF produces a new token every time a tokenized property is retrieved (e.g. `bucket.arn`). And since we used this token string every time. Since we use the token ID to produce an environment variable, this created a situation where `lift()` and `bind()` had a different environment variable name.
* Update the `Debug Wing Compiler` launch configuration to use the currently open file as the source for debugging.
* Remove the `self_client_path` from the generated `_toInflightType()` methods and just embed the path directly into the inline code.

## Checklist

- [x] Tests added (always)


tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/add_object.w_test_sim.md
```diff
@@ -2,13 +2,31 @@
 
 ## stdout.log
 ```log
-pass ─ add_object.wsim » root/env0/test:addObject
+fail ┌ add_object.wsim » root/env0/test:addObject
```
Chriscbr, line 5: Regression?

libs/wingsdk/src/shared-tf/tokens.ts
```diff
@@ -33,11 +38,16 @@ export class CdkTfTokens extends Tokens {
...
-    // the same token might be bound multiple times by different variables/inflight contexts
     if (host.env[envName] === undefined) {
       host.addEnvironment(envName, envValue);
     }
   }
+
+  private envNameForToken(value: any) {
+    const resolved = Tokenization.resolve(value, { scope: this.stack, resolver: new DefaultTokenResolver(new StringConcat()) });
+    return this.envName(resolved);
+  }
```
Chriscbr, line 51: Do you think there's a way to unit test this?

libs/wingsdk/src/std/resource.ts
```diff
@@ -224,7 +225,7 @@ export abstract class Resource extends Construct implements IResource {
     }
 
     throw new Error(
-      `unable to serialize immutable data object of type ${obj.constructor?.name}`
+      `objects of type '${obj.constructor?.name}' cannot be referenced from an inflight scope`
```
Chriscbr, line 228: In case it's not an object
```suggestion
      `objects of type '${obj.constructor?.name ?? typeof obj}' cannot be referenced from an inflight scope`
```

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/add_object.w_test_sim.md
eladb: yes, i have some work to do here.

libs/wingsdk/src/shared-tf/tokens.ts
eladb: yes, i'll look into this.

libs/wingsdk/src/std/resource.ts
eladb: 👍

Chriscbr:
Let me know when this is ready for another review

eladb:
Will do

eladb:
Superseded by #3125

Issue #2146, opened 2023-04-19
Type Inference: Deduce function return types automatically


skyrpex:
### Feature Spec

The Wing compiler can deduce the return type of functions, even if they aren't explicitly typed. 

The following functions are equivalent:

```ts
let create_number = inflight () => {
    return 1;
};
// let create_number: num = inflight () => {
//     return 1;
// };

let create_string = inflight () => {
    return "abc";
};
// let create_string: str = inflight () => {
//     return "abc";
// };
```

Additionally, the compiler may apply some constraints to the inferred types, such as when passing a function as an argument where the return type is already defined. For example:

```ts
declare needs_a_function_that_returns_a_number(callback: inflight () => num): void;

needs_a_function_that_returns_a_number(inflight () => {
    return "not a number";
           // ^ Error: expecting a number, not a string
});
```

### Use Cases

Improves the developer experience.

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.75](https://github.com/winglang/wing/releases/tag/v0.24.75).

PR #2910, opened 2023-06-13
[wip] arrow syntax proposal

Chriscbr:
The current syntax in Wing overloads ":" for both assigning a type to a name, and for specifying that type is a return type of a function. An example that highlights this:

```js
class Bloo { 
  x: (): cloud.Bucket; 
}
```

In the code above, x stores a function closure which accepts 0 arguments, and returns a `cloud.Bucket`. Notice also how two operators are needed today when defining closures -- a colon (":") and a arrow ("=>"):

```js
let add = inflight (x: num, y: num): num => { return x + y };
```

This PR considers an alternate syntax where ":" assigns types to names, and a new operator "->", relates pairs of types into functions. To compensate for the added operator to the grammar, it removes the "=>" arrow association with function closures, so that closures are created like so:

```js
let add = inflight (x: num, y: num) -> num { return x + y };
```

The previous example would look like:

```js
class Bloo { 
  x: () -> cloud.Bucket; 
}
```

---

The syntax might complement a "trailing closure" syntax like Swift provides, whereby if an inflight closure's type can be inferred, then the type signature can be entirely omitted:

```js
new cloud.Function(inflight {
  bucket.put("hello", $0);
});

// shorthand for

new cloud.Function(inflight ($0: str) -> void {
  bucket.put("hello", $0);
});
```

A trailing closure syntax would still be possible without this grammar change, though it's possible some users might instead try to write:

```js
new cloud.Function(inflight => {
  bucket.put("hello", $0);
});
```

## Checklist

- [ ] Tests added (always)


examples/tests/valid/bucket_events.w
```diff
@@ -3,24 +3,24 @@ bring cloud;
...
     log("updated ${key}");
 });
 
-b.onCreate(inflight (key:str) => {
+b.onCreate(inflight (key:str) -> void {
     log("created ${key}");
 });
 
-b.onEvent(inflight (key: str) => {   
+b.onEvent(inflight (key: str) -> void {
```
eladb, line 18: A major downside of this syntax is that if we want to support inference of return types (we already sort of have that with `void`), the syntax breaks:

```js
b.onEvent((key: str) {
  // ...
});
```

Right? That looks odd.

examples/tests/error/repeat_construct_id2.w
```diff
@@ -1,6 +1,6 @@
 bring cloud;
 
-let make_name = (): str => {
+let make_name = () -> str {
```
eladb, line 3: I am wondering if it makes sense to only use the `->` arrow syntax for function *declarations* but keep the *definitions* with `:`:

```js
struct Foo {
  predicate: (str) -> bool;
}

But then:

Foo {
  predicate: (x: str): bool => true;
}
```

This means that if I want to declare a function that returns a function it will look like this:

```js
let returnsPredicate = (): (str) -> bool => {
  return (x: str) => true;
};
```

eladb (changes requested):
See comments.

Also - I *love* the potential of trailing closures, but in your example you didn't go all the way. The beauty of a trailing closure is that it can be defined *outside*:

```js
new cloud.Function() {
  log("hello, function");
}
```

(this infers both phase and omits input/output params).

This is a very powerful tool.

examples/tests/valid/bucket_events.w
Chriscbr: I think this is one reason why inferring return types might not be desirable (I've also alluded to some other reasons [here](https://github.com/winglang/wing/issues/2146#issuecomment-1514905983))

examples/tests/error/repeat_construct_id2.w
Chriscbr: Interesting alternative 🤔 . You've also something else, which is whether we will have a shorter syntax for closures that just return a single expression. Ignoring the type annotation stuff, today in Wing you have to write

```js
let returnsPredicate = (): (str): bool => {
  return (x: str): bool => { return true; };
};
```

But it could be nice to support something like
```js
let returnsPredicate = (): (str): bool => {
  return (x: str): bool => true;
};

File: examples/tests/valid/bucket_events.w (hunk 1 above)
Comment (eladb) [reply examples/tests/valid/bucket_events.w:18]: I tend to agree that for *methods* we should require a return type annotation (for the reasons you state in your comment), but for *closures*, it's a huge ergonomic win.

So I think that a syntax that doesn't really allow us to add return type inference for closures is not ideal...

File: examples/tests/error/repeat_construct_id2.w (hunk 2 above)
Comment (eladb) [reply examples/tests/error/repeat_construct_id2.w:3]: yes, `true` instead of `{ return true; }` is definitely something we would want to introcuce.

File: examples/tests/valid/bucket_events.w (hunk 1 above)
Comment (Chriscbr) [reply examples/tests/valid/bucket_events.w:18]: One solution to this could be to add `fn` to the syntax for anonymous functions. For example, I think this might look reasonable:

```js
b.onEvent(fn(key) {
  // ...
});
```

With trailing closures, the `fn` could be omitted:

```js
b.onEvent { key in
  // ...
};
```

eladb:
Reopening

PR #2918, opened 2023-06-14
feat(sdk): util.wait

ekeren:
## Checklist

- [x] Tests added (always)


examples/tests/sdk_tests/util/wait.w
```diff
@@ -0,0 +1,57 @@
...
+test "returns true immediately" {
+  let start = JSHelper.getTime();
+  if util.wait((): bool => { return true; }) {
+    assert(JSHelper.getTime() - start < 1000);
+  } else {
+    assert(false);
+  }
+}
+
+test "returns false takes a minute" {
```
ekeren, line 18: This test is taking a minute, because it is testing the default behavior of timeout (set to 1 minute)

It is a long time to wait for developers on their box, should we change the test to a quicker timeout and not verify the 1m default timeout?

libs/wingsdk/src/util/util.ts
```diff
@@ -35,6 +70,32 @@ export class Util {
...
+  public static async wait(
+    predicate: IPredicateHandler,
+    props: WaitProps = {}
+  ): Promise<boolean> {
+    const timeout = props.timeout ?? Duration.fromMinutes(1);
+    const interval = props.interval ?? Duration.fromSeconds(0.1);
+    const f = predicate as any;
+    let elapsed = 0;
+    while (elapsed < timeout.seconds) {
+      if (await f()) {
```
ekeren, line 87: Should we surround the f invocation with try catch and treat exception as false?

examples/tests/sdk_tests/util/wait.w
staycoolcall911: Unless someone comes up with a way around this, I think it's ok to avoid testing the default 1m. For me, it's the slight chance of ever introducing an error in this simple code vs. the huge cost × number of runs...

libs/wingsdk/src/util/util.ts
```diff
@@ -35,6 +69,32 @@ export class Util {
     return new Promise((resolve) => setTimeout(resolve, delay.seconds * 1000));
   }
 
+  /**
+   * Waits until predicate is true or return false
```
Chriscbr, line 73:
```suggestion
   * Run a predicate repeatedly, waiting until it returns true (or throws)
```

Chriscbr, line 73: Please document what the return value means (`@returns your description`)

libs/wingsdk/src/util/util.ts
Chriscbr: I could see it going either way. Maybe propagating the error is fine? Let's document it either way, e.g.:

`@throws if the given predicate throws`

Chriscbr (approved):

examples/tests/sdk_tests/util/wait.w
```diff
@@ -0,0 +1,57 @@
+bring cloud;
+bring util;
+
+class JSHelper { 
+  extern "./sleep-helper.js" static inflight getTime(): num;
+}
+let oneHundredMiliseconds = 0.1s;
```
eladb, line 7: nit:

```suggestion

let oneHundredMiliseconds = 0.1s;
```

examples/tests/sdk_tests/util/wait.w
```diff
@@ -0,0 +1,57 @@
...
+}
+
+test "returns after some time waiting" {
+  let start = JSHelper.getTime();
+  let returnTrueAfter30Seconds = (): bool => { 
+    return JSHelper.getTime() - start > 30 * 1000; 
+  };
+  if util.wait(returnTrueAfter30Seconds) {
+    let diff = JSHelper.getTime() - start;
+      assert( diff > 30 * 1000 && diff < 40 * 1000);
```
eladb, line 34:
```suggestion
    assert( diff > 30 * 1000 && diff < 40 * 1000);
```

examples/tests/sdk_tests/util/wait.w
```diff
@@ -0,0 +1,57 @@
...
+let oneSecond = 1s;
+let tenSeconds = 10s;
+
+test "setting props" {
+  let start = JSHelper.getTime();
+  let returnFalse = (): bool => { 
+    invokeCounter.inc();
+    return false;
+  };
+  if util.wait(returnFalse, interval: oneSecond, timeout: tenSeconds ) {
```
eladb, line 50:
```suggestion

  if util.wait(returnFalse, interval: oneSecond, timeout: tenSeconds) {
```

libs/wingsdk/src/util/util.ts
```diff
@@ -1,5 +1,39 @@
...
+ * Represents a predicate with an inflight "handle" method that can be passed to
+ * `util.busyWait`.
+ * @inflight `@winglang/sdk.util.IPredicateHandlerClient`
+ */
+export interface IPredicateHandler extends IResource {}
+
+/**
+ * Inflight client for `IPredicateHandler`.
+ */
+export interface IPredicateHandlerClient {
```
eladb, line 30: Curious - I am wondering if it is possible to avoid `IPredicateHandler` and just use `IPredicateHandlerClient` directly on the method...

libs/wingsdk/src/util/util.ts
eladb: Use capital `Y` in the return doc:

```
@returns Your description
```

(@Chriscbr we need @return documentation to start with a capital case).

(maybe add to the linter?)

libs/wingsdk/src/util/util.ts
```diff
@@ -35,6 +69,32 @@ export class Util {
     return new Promise((resolve) => setTimeout(resolve, delay.seconds * 1000));
   }
 
+  /**
+   * Waits until predicate is true or return false
+   * @param predicate The function that will be evaluated
+   * @param props Timeout and interval values, default to one 1m timeout and 0.1sec interval
+   * @inflight
+   */
+  public static async wait(
```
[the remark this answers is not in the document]
eladb: I am okay with `waitUntil` (I didn't like `busyWait` which was the original name).

libs/wingsdk/src/util/util.ts
eladb: I think we should propagate the error. There's no "happy reason" for the predicate to throw.

eladb (approved):
Cool!

libs/wingsdk/src/util/util.ts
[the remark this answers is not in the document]
ekeren: `waitUntil` I like it

libs/wingsdk/src/util/util.ts
eladb: Not sure that's a super convincing example 🤷🏻‍♀️

I okay if we want to add an option:

```js
let p = () => { return b.get("key") == "value"; };
util.waitUntil(p, ignoreErrors: true);
```

This example also brings up an idea related to optionality: perhaps we should allow comparing `T?` and `T`? Seems like a safe thing to allow...

examples/tests/sdk_tests/util/wait.w
ekeren: fixed , thanks!

libs/wingsdk/src/util/util.ts
ekeren: > ignoreErrors

Not doing this for now, let see if it surface

ekeren: Right now, it throws  (added the `@throws Will throw if the given predicate throws.`)

libs/wingsdk/src/util/util.ts
ekeren: no

```
|       util.waitUntil((): bool => { 
   | /--------------------^
65 | |       invokeCounter.inc();
66 | |       throw("ERROR"); 
67 | |     });
   | \-----^ Expected type to be "IPredicateHandlerClient", but got "inflight (): bool" instead
```

libs/wingsdk/src/util/util.ts
Chriscbr: Yes, we need a style guide...

libs/wingsdk/src/util/util.ts
Chriscbr: It would be useful to find a way to collapse these in the our `@winglang/sdk` source code (it is quite boilerplatey to have to define two interfaces)

PR #2921, opened 2023-06-14
fix(vscode): missing completions in lazy-loaded jsii namespaces

### LINKED ISSUE #2639: namespace completions are missing unused types for jsii libraries
### I tried this:

```
bring "aws-cdk-lib" as awscdk;

new awscdk.aws_lambda.
```

### This happened:

no completions available

### I expected this:

to show some classes

### Is there a workaround?

The completions shown are ones that have been used in the current wing file.

e.g.

```
bring "aws-cdk-lib" as awscdk;

new awscdk.aws_lambda.DockerImageFunction(...);

new awscdk.aws_lambda.
		//^ now shows completion for `DockerImageFunction`
```

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

This is due to the lazy-jsii importer. The lsp needs a way to add additional imports on-demand.

### Community Notes

MarkMcCulloh:
Due to the jsii type lazy loading, completions only show types that were used in the current file. We now eagerly load namespaces when referenced (even if incomplete).

For example:

```
bring "aws-cdk-lib" as awscdk;

new awscdk.aws_lambda.
                    //^ completions will show here
```

https://github.com/winglang/wing/assets/1237390/d6edeee2-8c64-428d-a2ef-0aa51b5e3b96

## Notes
  - I couldn't think of a reasonable way to test this, the lsp tests need a good jsii fixture system
  - refactored some code into a `reference_to_udt` and as part of it, allowed a udt to basically refer to a namespace
  - We also now also eagerly load jsii types that don't belong to a namespace, just to cover that base

Fixes #2639

## Checklist

- [ ] Tests added (always)
  - See notes above 

### DIFF
--- a/libs/wingc/src/lsp/completions.rs
+++ b/libs/wingc/src/lsp/completions.rs
@@ -150,27 +150,26 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse 
 							return get_completions_from_type(&type_lookup, types, Some(found_env.phase), false);
 						} else {
 							// this is probably a namespace, let's look it up
-							let namespace = root_env
+							if let Some(namespace) = root_env
 								.lookup_nested_str(&udt.full_path_str(), scope_visitor.found_stmt_index)
-								.ok();
-							if let Some((namespace, _)) = namespace {
-								if let SymbolKind::Namespace(namespace) = namespace {
-									let completions = get_completions_from_namespace(namespace, Some(found_env.phase));
-									//for namespaces - return only classes and namespaces
-									if parent.parent().expect("custom_type must have a parent node").kind() == "new_expression" {
-										return completions
-											.iter()
-											.filter(|c| {
-												matches!(
-													c.kind,
-													Some(CompletionItemKind::CLASS) | Some(CompletionItemKind::MODULE)
-												)
-											})
-											.cloned()
-											.collect();
-									} else {
-										return completions;
-									}
+								.ok()
+								.and_then(|n| n.0.as_namespace_ref())
+							{
+								let completions = get_completions_from_namespace(&namespace, Some(found_env.phase));
+								//for namespaces - return only classes and namespaces
+								if parent.parent().expect("custom_type must have a parent node").kind() == "new_expression" {
+									return completions
+										.iter()
+										.filter(|c| {
+											matches!(
+												c.kind,
+												Some(CompletionItemKind::CLASS) | Some(CompletionItemKind::MODULE)
+											)
+										})
+										.cloned()
+										.collect();
+								} else {
+									return completions;
 								}
 							}
 
@@ -331,14 +330,13 @@ fn get_completions_from_namespace(
 		LookupResult::DefinedLater => vec![],
 		LookupResult::ExpectedNamespace(_) => vec![],
 	};
-	let l: Vec<CompletionItem> = namespace
+	namespace
 		.env
 		.symbol_map
 		.iter()
 		.flat_map(|(name, symbol)| format_symbol_kind_as_completion(name, &symbol.1))
 		.chain(util_completions.into_iter())
-		.collect();
-	l
+		.collect()
 }
 
 /// Gets accessible properties on a class as a list of CompletionItems
--- a/libs/wingc/src/lsp/completions.rs
+++ b/libs/wingc/src/lsp/completions.rs
@@ -149,10 +149,9 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse 
 						if let Ok(type_lookup) = type_lookup {
 							return get_completions_from_type(&type_lookup, types, Some(found_env.phase), false);
 						} else {
-							// this is probably a namespace
-							// `resolve_user_defined_type` will fail for namespaces, let's just look it up instead
+							// this is probably a namespace, let's look it up
 							let namespace = root_env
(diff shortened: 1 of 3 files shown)


libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -756,27 +757,56 @@ impl<'a> JsiiImporter<'a> {
...
 
 	fn parameter_to_wing_type(&mut self, parameter: &jsii::Parameter) -> TypeRef {
+		let mut param_type = self.type_ref_to_wing_type(&parameter.type_);
+
 		if parameter.variadic.unwrap_or(false) {
-			panic!("TODO: variadic parameters are unsupported - Give a +1 to this issue: https://github.com/winglang/wing/issues/397");
+			// TODO properly thought out variadic parameter support https://github.com/winglang/wing/issues/397
+			param_type = self.wing_types.add_type(Type::Array(param_type));
+			param_type = self.wing_types.add_type(Type::Optional(param_type));
 		}
```
MarkMcCulloh, line 766: We now eagerly load types in the root of imported jsii packages. In the root of "constructs", one of the types has a variadic param.

I feel like this approach is okay, but I didn't want to close the issue because I didn't really think that hard about it. If this change is sufficient then I'll close it.

libs/wingc/src/lsp/completions.rs
```diff
@@ -144,36 +144,37 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse
...
+						if let Ok(type_lookup) = type_lookup {
+							return get_completions_from_type(&type_lookup, types, Some(found_env.phase), false);
+						} else {
+							// this is probably a namespace, let's look it up
 							let namespace = root_env
-								.lookup_nested_str(&udt.root.name, scope_visitor.found_stmt_index)
+								.lookup_nested_str(&udt.full_path_str(), scope_visitor.found_stmt_index)
 								.ok();
 							if let Some((namespace, _)) = namespace {
 								if let SymbolKind::Namespace(namespace) = namespace {
```
Chriscbr, line 157: I think we have an `as_namespace()` method?

libs/wingc/src/lsp/completions.rs
```diff
@@ -144,36 +144,37 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse
...
+						let type_lookup = resolve_user_defined_type(udt, found_env, scope_visitor.found_stmt_index.unwrap());
+
+						if let Ok(type_lookup) = type_lookup {
+							return get_completions_from_type(&type_lookup, types, Some(found_env.phase), false);
+						} else {
+							// this is probably a namespace, let's look it up
 							let namespace = root_env
-								.lookup_nested_str(&udt.root.name, scope_visitor.found_stmt_index)
+								.lookup_nested_str(&udt.full_path_str(), scope_visitor.found_stmt_index)
 								.ok();
```
Chriscbr, line 155: (optional) There might be enough logic here to warrant moving this else branch into a separate function #nevernester

libs/wingc/src/lsp/completions.rs
```diff
@@ -330,13 +331,14 @@ fn get_completions_from_namespace(
...
-	namespace
+	let l: Vec<CompletionItem> = namespace
 		.env
 		.symbol_map
 		.iter()
 		.flat_map(|(name, symbol)| format_symbol_kind_as_completion(name, &symbol.1))
 		.chain(util_completions.into_iter())
-		.collect()
+		.collect();
+	l
```
Chriscbr, line 341: nit: A most rusty way to do this might be to call `collect_vec` (requires importing a trait from `itertools`) or specifying the collection type directly with `collect::<Vec<_>>()` so that the temporary variable isn't needed

libs/wingc/src/type_check/jsii_importer.rs
Chriscbr: I think the only sure-fire way to know if this works (now and going forward) is to add an e2e test where a variadic param is used. 🙂

The main thing we want to avoid is e.g. code-generating `variadicFunction([1, 2, 3])` if, say, the underlying JavaScript expects `variadicFunction(1, 2, 3)`.

(I'm totally cool with this workaround and leaving the issue open if removing this panic is needed for LSP stuff to work)

Chriscbr (approved):

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: ah yeah this was better before, I added the temp var for debugging. If I take it out and just return the collect it's cleaner

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: true, just wasn't public. I went ahead and exposed it.

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: I removed a layer of nesting but otherwise kept it in place

libs/wingc/src/type_check/jsii_importer.rs
MarkMcCulloh: Okay this still straight up doesn't work during preflight. I made it more clear in the comment but still kept the workaround in-place

PR #2923, opened 2023-06-14
feat(docs): equality semantics

### LINKED ISSUE #2756: Add equality semantics to language spec
### Feature Spec

Write down how equality (the `==` operator) is defined in the language. How should `==` behave for each type in the language?

The explanation can reference JavaScript for how equality works for foreign objects (like values typed as `any`) but should probably avoid leaking it as an implementation detail for built-in types of the language if possible.

### Use Cases

Explain why certain comparisons do or don't work

### Implementation Notes

_No response_

### Component

Language Design

### Community Notes

Chriscbr:
Initial spec for v0 of equality in the Wing language. Many of the relationships described already hold, though some of them are not implemented yet. All feedback is welcome.

[Rendered doc](https://github.com/winglang/wing/blob/rybickic/equality-spec/docs/03-language-guide/14-equality.md)

Closes #2756 

## Checklist

- [x] Tests added (always)


docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,131 @@
...
+description: A reference guide to equality in Wing.
+keywords: [Winglang equality, Wing equality, equals, identity]
 ---
 
-TODO (for now, see the [language spec RFC](../999-contributors/999-rfcs/2023-06-12-language-spec.md))
+Checking for equality is performed with the `==` operator in Wing. It returns `true` if the two values are equal, and `false` otherwise.
+
+The main difference between equality in JavaScript and Wing is that `==` in Wing is not allowed to compare values of different types. For example, `1 == "1"` is not allowed in Wing, and will result in a compile-time error.
+
+The execution phase ([preflight or inflight](../02-core-concepts/01-preflight-and-inflight.md)) that a value was created in does not affect its equality. For example, a value created in preflight can be equal to a value created in inflight.
```
MarkMcCulloh, line 12: This weirdly feels like it's too complicated to be correct, but I can't think of any issues, so 👍

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,131 @@
 ---
 title: Equality
 id: equality
+description: A reference guide to equality in Wing.
+keywords: [Winglang equality, Wing equality, equals, identity]
 ---
 
-TODO (for now, see the [language spec RFC](../999-contributors/999-rfcs/2023-06-12-language-spec.md))
+Checking for equality is performed with the `==` operator in Wing. It returns `true` if the two values are equal, and `false` otherwise.
```
MarkMcCulloh, line 8: nit: I feel like our reference docs should avoid mentioning "in Wing" unless it's ambiguous (e.g. when comparing to other languages)

```suggestion
Checking for equality is performed with the `==` operator. It returns `true` if the two values are equal, and `false` otherwise.
```

Contextually, all of these docs are referring to the world of Wing so it's redundant.

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,131 @@
...
+The execution phase ([preflight or inflight](../02-core-concepts/01-preflight-and-inflight.md)) that a value was created in does not affect its equality. For example, a value created in preflight can be equal to a value created in inflight.
+
+Equality in Wing is a symmetric and transitive relationship - that is, (1) if `a == b`, then `b == a`, and (2) if `a == b` and `b == c`, then `a == c`.
+
+The following is a set of rules for checking equality:
+
+## Basic types
+
+1. Two `str` values are equal if they have the same characters in the same order.
+2. Two `num` values are equal if they have the same floating-point value. `NaN` is not equal to any value, including itself.
```
MarkMcCulloh, line 21: Should we reference https://en.wikipedia.org/wiki/IEEE_754? Assuming we're gonna go by that (as js does).
That would include specifying that -0 == +0.

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,131 @@
 ---
 title: Equality
 id: equality
+description: A reference guide to equality in Wing.
+keywords: [Winglang equality, Wing equality, equals, identity]
```
MarkMcCulloh, line 5: I don't think we need to include `wing` in our keywords. Ideally, we should find a way to inject it for every page in the docs automatically. or maybe it's just not needed at all.

```suggestion
keywords: [equality, equals, identity]
```

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,131 @@
...
+
+1. Two `Json` values are equal if they are both `null`.
+2. Two `Json` values are equal if they are both `bool` values and are equal.
+3. Two `Json` values are equal if they are both `num` values and are equal.
+4. Two `Json` values are equal if they are both `str` values and are equal.
+5. Two `Json` values are equal if they are both `Array` values and are equal.
+6. Two `Json` values are equal if they are both `Map` values and are equal.
+
+```js
+assert(Json.null == Json.null); // TODO: https://github.com/winglang/wing/issues/1819
```
MarkMcCulloh, line 93: I know the syntax isn't actually finalized, but given we added nil I feel like this is also a possible alternative.

```suggestion
assert(Json nil == Json nil); // TODO: https://github.com/winglang/wing/issues/1819
```

MarkMcCulloh (approved):

docs/03-language-guide/14-equality.md
Chriscbr: Sure, let's try it

docs/03-language-guide/14-equality.md
Chriscbr: I agree

docs/03-language-guide/14-equality.md
Chriscbr: Yeah, it's a little optimistic, for sure :-)

One question I had was if you have two instances of a class created in preflight, and then you lift them into inflight, can you still safely check them for equality? In any case, I'm sure we'll figure out in due time

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
 title: Equality
 id: equality
+description: A reference guide to equality in Wing.
+keywords: [equality, equals, identity]
 ---
 
-TODO (for now, see the [language spec RFC](../999-contributors/999-rfcs/2023-06-12-language-spec.md))
+Checking for equality is performed with the `==` operator. It returns `true` if the two values are equal, and `false` otherwise.
+
+The main difference between equality in JavaScript and Wing is that `==` in Wing is not allowed to compare values of different types. For example, `1 == "1"` is not allowed in Wing, and will result in a compile-time error.
```
eladb, line 10:
```suggestion
> The main difference between equality in JavaScript and Wing is that `==` in Wing is not allowed to compare values of different types. For example, `1 == "1"` is not allowed in Wing, and will result in a compile-time error.
```

docs/03-language-guide/14-equality.md
eladb: I feel like this paragraph is a pretty advanced statement and should likely go at the bottom of this section (perhaps also the one about JavaScript).

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+The execution phase ([preflight or inflight](../02-core-concepts/01-preflight-and-inflight.md)) that a value was created in does not affect its equality. For example, a value created in preflight can be equal to a value created in inflight.
+
+Equality in Wing is a symmetric and transitive relationship - that is, (1) if `a == b`, then `b == a`, and (2) if `a == b` and `b == c`, then `a == c`.
+
+The following is a set of rules for checking equality:
+
+## Basic types
+
+1. Two `str` values are equal if they have the same characters in the same order.
+2. Two `num` values are equal if they have the same floating-point value. The [IEEE 754] standard is used for storing numbers, which means that for example `-0 == +0`. `NaN` is not equal to any value, including itself.
```
eladb, line 21: with floating point numbers, it is sometimes desirable to compare them with some possible accuracy level (but that's likely a `math.equal()` thing).

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+3. Two `bool` values are equal if they are both `true` or both `false`.
+4. Two `duration` values are equal if they have the same number of milliseconds.
+5. Two `T?` types (optional `T` values) are equal if they are both empty (`nil`) or both non-empty, and if they are both non-empty, their inner values are equal. A value of type `T?` is never equal to a value of type `T`.
+
+[IEEE 754]: https://en.wikipedia.org/wiki/IEEE_754
+
+## Collection types
+
+Wing contains six collection types: `Array`, `MutArray`, `Map`, `MutMap`, `Set`, and `MutSet`.The following rules apply to all of them:
```
eladb, line 30: 👍

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+
+1. Two collections are equal if they have the same number of elements, and if each element in the first collection is equal to the corresponding element in the second collection. (The order of elements only matters for `Array` and `MutArray`.)
+2. The mutability of a collection does not affect its equality. In other words, a `MutArray` is equal to an `Array` with the same elements, and a `MutMap` is equal to a `Map` with the same keys and values.
+3. Only collections of the same "kind" can be equal. For example, an `Array` cannot be equal to a `Map`, and a `MutArray` cannot be equal to a `MutMap`.
+
+## Function types
+
+Two functions are equal if they are both the same function. This means that two functions that have the same code are not necessarily equal, since they may have been defined in different places.
```
eladb, line 38:
```suggestion
Two functions are equal if they are both the same object (by reference). This means that two functions that have the same code are not necessarily equal, since they may have been defined in different places.
```

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+The main difference between equality in JavaScript and Wing is that `==` in Wing is not allowed to compare values of different types. For example, `1 == "1"` is not allowed in Wing, and will result in a compile-time error.
+
+The execution phase ([preflight or inflight](../02-core-concepts/01-preflight-and-inflight.md)) that a value was created in does not affect its equality. For example, a value created in preflight can be equal to a value created in inflight.
+
+Equality in Wing is a symmetric and transitive relationship - that is, (1) if `a == b`, then `b == a`, and (2) if `a == b` and `b == c`, then `a == c`.
+
+The following is a set of rules for checking equality:
+
+## Basic types
+
```
eladb, line 19:
```suggestion

Basic types are compared *by value*.

```

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+Checking for equality is performed with the `==` operator. It returns `true` if the two values are equal, and `false` otherwise.
+
+The main difference between equality in JavaScript and Wing is that `==` in Wing is not allowed to compare values of different types. For example, `1 == "1"` is not allowed in Wing, and will result in a compile-time error.
+
+The execution phase ([preflight or inflight](../02-core-concepts/01-preflight-and-inflight.md)) that a value was created in does not affect its equality. For example, a value created in preflight can be equal to a value created in inflight.
+
+Equality in Wing is a symmetric and transitive relationship - that is, (1) if `a == b`, then `b == a`, and (2) if `a == b` and `b == c`, then `a == c`.
+
+The following is a set of rules for checking equality:
+
```
eladb, line 17: I think it's worth defining the terms "by value" and "by reference" here and using the at the beginning of each section.

Also worth mentioning the concept of deep versus shallow comparison.

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+  PINEAPPLE,
+}
+
+assert(PizzaTopping.CHEESE == PizzaTopping.CHEESE);
+assert(PizzaTopping.CHEESE != PizzaTopping.PINEAPPLE);
+```
+
+## Classes and interfaces
+
+Two class instances or interface-satisfying objects are equal if they are the same instance. This means that two class instances, or interface-satisfying objects that have the same data are not necessarily equal, since they may have been created in different places.
```
eladb, line 65:
```suggestion
Two class instances or interface-satisfying objects are equal if they are the same instance (by reference). This means that two class instances, or interface-satisfying objects that have the same data are not necessarily equal, since they may have been created in different places.
```

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+let shop2 = new Shop(1);
+let shop3 = shop1;
+
+assert(shop1 != shop2);
+assert(shop1 == shop3);
+```
+
+## `Json`
+
+Two `Json` values are equal if they contain the same structure and values. The following rules apply:
```
eladb, line 85:
```suggestion
Two `Json` values are equal if they contain the same structure and values (deep by reference equality). Another way to think about it is the two `Json` values are equal if their stringified representation is equal.

The following rules apply:
```

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+[IEEE 754]: https://en.wikipedia.org/wiki/IEEE_754
+
+## Collection types
+
+Wing contains six collection types: `Array`, `MutArray`, `Map`, `MutMap`, `Set`, and `MutSet`.The following rules apply to all of them:
+
+1. Two collections are equal if they have the same number of elements, and if each element in the first collection is equal to the corresponding element in the second collection. (The order of elements only matters for `Array` and `MutArray`.)
+2. The mutability of a collection does not affect its equality. In other words, a `MutArray` is equal to an `Array` with the same elements, and a `MutMap` is equal to a `Map` with the same keys and values.
+3. Only collections of the same "kind" can be equal. For example, an `Array` cannot be equal to a `Map`, and a `MutArray` cannot be equal to a `MutMap`.
+
```
eladb, line 35: This is currently not the behavior correct? Maybe worth mentioning and adding a tracking issue.

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+assert(Json 1 == Json 1);
+assert(Json -0.42 == Json -0.42);
+assert(Json "foo" == Json "foo");
+assert(Json [1, 2, 3] == Json [1, 2, 3]);
+assert(Json { "foo": 1, "bar": 2 } == Json { "foo": 1, "bar": 2 });
+```
+
+## Structs
+
+Two struct values are equal if they have the same type and all of their fields are equal. The order in which fields are defined does not matter.
```
eladb, line 107:
```suggestion
Two struct values are equal if they have the same type and all of their fields are equal (based on rules of equality of their type). The order in which fields are defined does not matter.
```

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+
+```js
+struct Node {
```
eladb, line 110: That's a pretty esoteric example...

eladb (approved):
Great job!

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+
+The following is a set of rules for checking equality:
+
+## Basic types
+
+1. Two `str` values are equal if they have the same characters in the same order.
+2. Two `num` values are equal if they have the same floating-point value. The [IEEE 754] standard is used for storing numbers, which means that for example `-0 == +0`. `NaN` is not equal to any value, including itself.
+3. Two `bool` values are equal if they are both `true` or both `false`.
+4. Two `duration` values are equal if they have the same number of milliseconds.
+5. Two `T?` types (optional `T` values) are equal if they are both empty (`nil`) or both non-empty, and if they are both non-empty, their inner values are equal. A value of type `T?` is never equal to a value of type `T`.
```
eladb, line 24: Something that came up [here](https://github.com/winglang/wing/pull/2918#discussion_r1230673067) is that maybe we should allow comparing `T?` and `T`:

```js
let var x: str? = "hello";

if x == "hello" {
  log("x is hello");
}

x = nil;
if x != "hello" {
  log("x is not hello");
}
```

yoav-steinberg: I'm for staying explicit about this for now. But I also see the value (pun intended) of implicitly checking for nil.
If we take that course there's the question of being recursive about it. Can we compare a `T???` with a `T`? Can we compare a `T??` with a `T?`? I think we should allow all of these if we allow any of them.

docs/03-language-guide/14-equality.md
yoav-steinberg: Right, this is a special case of non-primitive equality (always by reference).

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+Two enum values are equal if they refer to the same enum case.
+
+```js
+enum PizzaTopping {
+  CHEESE,
+  PINEAPPLE,
+}
+
+assert(PizzaTopping.CHEESE == PizzaTopping.CHEESE);
+assert(PizzaTopping.CHEESE != PizzaTopping.PINEAPPLE);
```
yoav-steinberg, line 60:
```suggestion
assert(PizzaTopping.CHEESE == PizzaTopping.CHEESE);
assert(PizzaTopping.CHEESE != PizzaTopping.PINEAPPLE);
let weird_taste = PizzaTopping.PINEAPPLE;
assert(weird_taste == PizzaTopping.PINEAPPLE);
```

docs/03-language-guide/14-equality.md
eladb: > I'm for staying explicit about this for now.

What's the motivation?

docs/03-language-guide/14-equality.md
yoav-steinberg: why "deep reference equality"? just "deep equality". References seem not relevant for how we want to do `Json` compares.

docs/03-language-guide/14-equality.md
yoav-steinberg: The comment about the order sees a bit misleading to me. If they are the same type they are the same type.

docs/03-language-guide/14-equality.md
Chriscbr: > What's the motivation?

I think the idea is to require unwrapping optional values, no?

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+
+assert(shop1 != shop2);
+assert(shop1 == shop3);
+```
+
+## `Json`
+
+Two `Json` values are equal if they contain the same structure and values. The following rules apply:
+
+1. Two `Json` values are equal if they are both `null`.
```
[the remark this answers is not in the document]
Chriscbr: Sure

docs/03-language-guide/14-equality.md
Chriscbr: Good point, I think we will need a math utility for that use case. `==` for numbers is still sensible for integers of course

docs/03-language-guide/14-equality.md
Chriscbr: Good idea, I think adding "by value" / "by reference" improves the explanation 👍

the way I've defined the semantics hopefully doesn't require burdening users with understanding the difference between deep and shallow comparisons (the suggested direction is if we're checking if two arrays are equal, we perform `==` on every element recursively); but if that doesn't end up working, then we can add it back.

docs/03-language-guide/14-equality.md
Chriscbr: I guess the question how does someone know that `Foo { a: 1, b: 2 }` is the same as `Foo { b: 2, a: 1 }`? It sounds like you're saying that's already part of the definition of a struct, in which case I'm ok taking that sentence out.

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+
+Two functions are equal if they are both the same function. This means that two functions that have the same code are not necessarily equal, since they may have been defined in different places.
+
+```js
+let f1 = (x: num): num => x + 1;
+let f2 = (x: num): num => x + 1;
+let f3 = f1;
+
+assert(f1 != f2);
+assert(f1 == f3);
```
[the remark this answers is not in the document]
Chriscbr: Aha, copilot got me there... thanks for catching this

docs/03-language-guide/14-equality.md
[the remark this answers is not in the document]
Chriscbr: Good call, added.

docs/03-language-guide/14-equality.md
Chriscbr: Hey, what are you trying to say about pineapple pizza...? 😜

docs/03-language-guide/14-equality.md
```diff
@@ -1,6 +1,133 @@
...
+enum PizzaTopping {
+  CHEESE,
+  PINEAPPLE,
+}
+
+assert(PizzaTopping.CHEESE == PizzaTopping.CHEESE);
+assert(PizzaTopping.CHEESE != PizzaTopping.PINEAPPLE);
+```
+
+## Classes and interfaces
```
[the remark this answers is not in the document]
Chriscbr: Right I could imagine users might want a way to define / override equality for custom types. Let's open a separate issue for this

docs/03-language-guide/14-equality.md
Chriscbr: Changed it a bit

docs/03-language-guide/14-equality.md
Chriscbr: I've added links to issues covering aspects of equality that have not been implemented yet.

PR #2924, opened 2023-06-14
fix(compiler): type errors are resolved as `any` type by the type checker

yoav-steinberg:
See #884
Since type errors resolved as `any` type by the type checker we don't have a cleat way to distinguish unresolved types from valid `any` types. This can lead to bug downstream where a symbol with an unknown type is accessed as if it's an `any` without the compiler noticing it. In practice this usually isn't a problem since type errors lead to a diagnostic which leads to a compilation failure so we never get too far downstream with this. But correctness is important here:
 * There might be cases we aren't aware of.
 * A dedicated error-type can be used for additional diagnostics like location where the type resolving failed.
 * We are now more explicit in the code about ignore cascading errors as opposed to handing an any type.

- [x] ~~As of now I don't have a good test for this showing how using an `any` type creates an error where it's illegal compared to ignoring that error in case of a cascading type error.~~ But I'm working on finding such a case... **update:** The one case that's relevant that I could find is that unknown symbols are now "unresolved types" which means we don't need to check if they are reassignable in assignment statements, previously they were any types which lead to an additional redundant error (can't reassign to a non-reassignable). Updated snapshot test accordingly.

The PR also cleans up some type checking code and produces slightly better diags in some cases while eliminating diags which where redundant (see snapshot).

This PR also changes the LSP to eagerly load submodule types when needed in the completion logic instead of during the type checking phase. This is a revised fix of #2921.

## Checklist

- [ ] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -1527,7 +1547,7 @@ impl<'a> TypeChecker<'a> {
 						return self.types.error();
 					}
 				} else {
-					self.spanned_error(callee, "Expected a function or method");
+					self.spanned_error(callee, format!("Expected a function or method got {}", func_type));
```
MarkMcCulloh, line 1550: I believe this is our usual format for this type of error:

```suggestion
					self.spanned_error(callee, format!("Expected a function or method, found \"{}\"", func_type));
```

libs/wingc/src/lib.rs
```diff
@@ -313,7 +313,7 @@ pub fn compile(
 	type_check(&mut scope, &mut types, &source_path, &mut jsii_types);
 
 	// Validate that every Expr in the final tree has been type checked
-	let mut tc_assert = TypeCheckAssert::new(&types);
+	let mut tc_assert = TypeCheckAssert::new(&types, found_errors());
```
MarkMcCulloh, line 316: found_errors() can be true even if there are no type check errors (if there are parser errors). Does that hurt the usefulness of this extra check?

libs/wingc/src/docs.rs
```diff
@@ -74,6 +74,7 @@ impl Documented for TypeRef {
 			| Type::Json
 			| Type::MutJson
 			| Type::Nil
+			| Type::Error
```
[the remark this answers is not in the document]
yoav-steinberg: Yes I feel both aren't perfect, but I think I now prefer your suggestion, I'll make the change.

libs/wingc/src/lib.rs
yoav-steinberg: No, I'll fix the comment above to better reflect what's being done here now. The check verifies the type-checker didn't miss anything, which in practice is:
1. All expressions were resolved to something.
2. If any were resolved to `Unresolved` then we should have generated some error diagnostics (otherwise we'll probably mess something up downstream). I need the new `found_errors()` arg for this second check.

libs/wingc/src/type_check.rs
```diff
@@ -3337,18 +3367,9 @@ impl<'a> TypeChecker<'a> {
...
-					if let Some(ref_udt) = self.reference_to_udt(reference) {
-						let lookup = self.resolve_user_defined_type(&ref_udt, env, self.statement_idx);
-						if let Err(t) = lookup {
-							if t.message.ends_with("to be a type but it's a namespace") {
-								return self.make_error_variable_info(false);
-							}
-						}
-					}
+				// If resolving the object's type failed, we can't resolve the property either
+				if instance_type.is_error() {
```
yoav-steinberg, line 3371: For the vscpde autocomplete to work we need to make sure `resolve_user_defined_type` is called here to eagerly import the types inside a namespace in case this reference only points to a name space (and not to an actual type). Fix!

MarkMcCulloh (approved):

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -47,27 +47,13 @@ error: Unknown symbol \\"this\\"
   |     ^^^^ Unknown symbol \\"this\\"
 
 
-error: Variable this.instanceField is not reassignable
```
eladb, line 50: what happened here?

eladb (dismissed):
missing test

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: `this` in a static method is an unknown symbol (in a best case scenario we can be more specific and say something like "this" isn't available in a static method). So it was a bug in the previous code where after we already know `this` is unknown we still check if it's reassignable. In the past we couldn't distinguish between `any` and an unknown variable, so we needed to account for the case that an `any` var isn't reassignable (even if it was actually an unknown symbol), today we can ignore these cases an reduce this redundant error message..

yoav-steinberg:
> missing test

This is a bit tricky. I added something to the tests (see unknown_symbol.w) which shows this is working, although not ideal. Read top comment for details.

PR #2925, opened 2023-06-14
feat(vscode): docs in completions and signature help for constructors

MarkMcCulloh:
https://github.com/winglang/wing/assets/1237390/253e892c-0680-4bba-ab7c-3a0b065ba686

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/docs.rs
+++ b/libs/wingc/src/docs.rs
@@ -53,7 +53,7 @@ impl Documented for SymbolKind {
 
 impl Documented for Namespace {
 	fn render_docs(&self) -> String {
-		format!("Module '{}'", self.name).to_string()
+		format!("Module `{}`", self.name).to_string()
 	}
 }
 
--- a/libs/wingc/src/lsp/signature.rs
+++ b/libs/wingc/src/lsp/signature.rs
@@ -1,9 +1,10 @@
 use itertools::Itertools;
 use lsp_types::{ParameterInformation, ParameterLabel, Position, SignatureHelp, SignatureInformation};
 
-use crate::ast::{Expr, ExprKind};
+use crate::ast::{Expr, ExprKind, Symbol};
 use crate::lsp::sync::FILES;
 
+use crate::type_check::CLASS_INIT_NAME;
 use crate::visit::{visit_expr, Visit};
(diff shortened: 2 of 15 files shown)


libs/wingc/src/docs.rs
```diff
@@ -53,7 +53,7 @@ impl Documented for SymbolKind {
 
 impl Documented for Namespace {
 	fn render_docs(&self) -> String {
-		format!("Module '{}'", self.name).to_string()
+		format!("Module `{}`", self.name).to_string()
```
eladb, line 56: 👍

libs/wingc/src/lsp/signature.rs
```diff
@@ -31,88 +32,98 @@ pub fn on_signature_help(params: lsp_types::SignatureHelpParams) -> Option<Signa
...
+						documentation: if let Some(structy) = p.1.typeref.maybe_unwrap_option().as_struct() {
+							// print all fields
+							let fields = structy
+								.env
+								.iter(true)
+								.map(|f| format!("  {}: {};", f.0, f.1.as_variable().unwrap().type_))
+								.join("\n");
+							Some(lsp_types::Documentation::MarkupContent(lsp_types::MarkupContent {
+								kind: lsp_types::MarkupKind::Markdown,
+								value: format!("```wing\nstruct {} {{ \n{}\n}}\n```", structy.name.name, fields),
```
eladb, line 110: I think it's confusing to see the struct's name. Lots of people simply don't know about the keyword argument feature because of tht.

I think we should just flatten the keyword arguments. There's no need to mention the "props" structs at all.

For example, the doc of `new cloud.Function` should look something like this:

---


```js
new cloud.Function(handler, ...);
```

### Arguments

- `handler: (event: str): str` - the inflight closure to call...

### Keyword arguments
- `timeout: duration` - duration to wait before ...
- `memory: num` - how much memory to allocate for this function...
- ...

---

What do you think?

eladb (changes requested):

MarkMcCulloh:
@eladb I'm in agreement with the changes you suggested to the struct expansion representation. I made a partial change to spread the struct fields, but the docs still need some work. 
I'm gonna deal with this as a follow-up pr and play around with the representation to be more consumer-oriented rather than producer oriented.

staycoolcall911 (approved):

Issue #2820, opened 2023-06-06
Run unit tests in parallel with e2e tests


MarkMcCulloh:
### Feature Spec

CI builds generally have to follow the process of build and then tests. E2E tests were forced to run after unit tests, even though it is technically not required. 
It would be significantly faster to run `npx nx test` while other E2E tests are running because the unit tests actually take quite a while now.

### Use Cases

vroom vroom

### Implementation Notes

To optimize speed, this would likely require storing the entire workspace as a github artifact. If not reasonable (_cough cough_ node_modules) we need to make sure to at least store the nx cache alongside the source/build stuff.

### Component

Contributor Experience

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

PR #2944, opened 2023-06-16
chore: run unit tests in parallel with build/e2e and fix e2e sharding

MarkMcCulloh:
Time Save: ~17 minutes
Time Save (pr/e2e-full): ~30 minutes

This is a low-effort change, but saves us quite a bit of time.
This adds quite a bit of duplication due to the fact that testing in a new job effectively requires us to redo a lot of work, but at the cost of wasted workflow time we gain overall speed. This could be possible to avoid by saving the entire build steps as an asset, but that would be multiple GBs of data so I'm not sure it's worth it. 
It would also be solved by fully figuring out Nx caching, but tbh I think that'll be a nightmare.

Additionally, it turns out that the test sharding for e2e tests weren't working. So we were just running all tests twice 😭 . Fixing that shaved another ~7 minutes from non-full e2e load.

Misc Change: Decided also to remove the auto-approve workflow, since we're not using it and it adds a useless "skipped" status to all PRs.

Closes https://github.com/winglang/wing/issues/2820

## Checklist

- [x] Tests added (always)


Chriscbr (approved):
Huge perf improvements! 🚀 🚀 🚀

staycoolcall911:
🙏

PR #2950, opened 2023-06-17
fix: esbuild-wasm is slow via child_process

MarkMcCulloh:
We were running esbuild through a spawned node process, which is very slow. See https://github.com/evanw/esbuild/issues/2927 for why this was needed. In that thread, it was mentioned that some changes to esbuild avoided the issue. 

### Change

Added a patch with the previously mentioned workaround to "fix" this (with a small additional change to handle the newly piped stderr). This greatly improves bundling time (saves several hundred ms per inflight).

As far as I can tell, there are no bad side effects for this change (leaks, etc.)

### Root Cause (?)

After doing research on this, I still haven't quite figured out the real issue. esbuild-wasm creates both a worker thread and a subprocess that stays alive "forever". The thread and subprocess are `.unref()`ed, which means that node will not wait for them to finish before the parent process ends. The child process has a handle on stderr, so it seems like when the parent process finishes at a bad time then something in that fd gets messed up. This change removes that handle, which I suppose is good enough?

### Misc

- Started running into an issue that looked like vitest wasn't properly mocking an azure dep in the SDK tests. Not sure why it suddenly started happening. Updating vitest fixed it.
- Updated esbuild-wasm to latest. This wasn't necessary for the fix to work, I just figured it made sense to do.
- I accidentally ran `npm audit fix` and decided that instead of undoing it I would just keep it in, so this PR also updates aws-sdk packages.

### Upstreaming this change

I don't understand esbuild's architecture to know if this change is best for everyone, so maybe we hang on to this for a while before thinking about offering it as a PR upstream. Not sure though, I'd be fine with also doing it right away to see what the author says.

## Checklist

- [ ] Tests added (always)


libs/wingsdk/patches/esbuild-wasm+0.18.4.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/node_modules/esbuild-wasm/lib/main.js b/node_modules/esbuild-wasm/lib/main.js
+index 2ac392f..7b4483f 100644
+--- a/node_modules/esbuild-wasm/lib/main.js
++++ b/node_modules/esbuild-wasm/lib/main.js
+@@ -1926,7 +1926,7 @@ var ensureServiceIsRunning = () => {
+   let [command, args] = esbuildCommandAndArgs();
+   let child = child_process.spawn(command, args.concat(`--service=${"0.18.4"}`, "--ping"), {
+     windowsHide: true,
+-    stdio: ["pipe", "pipe", "inherit"],
```
eladb, line 9: Thanks for taking the initiative!

eladb (approved):
yes yes yes!

PR #2951, opened 2023-06-18
fix(sdk): std.utils.sleep() test

ainvoner:
verify that sleep work without guarantee precision


examples/tests/sdk_tests/util/sleep.w
```diff
@@ -10,5 +10,6 @@ test "sleep 100 mili seconds" {
   util.sleep(oneHundredMiliseconds);
   let end = JSHelper.getTime();
   let delta = end - start;
+  log("slept for ${delta} ms");
   assert(delta > 80 && delta < 220);
```
MarkMcCulloh, line 14: imo, the test should only verify that the sleep happened. It's too hard to guarantee the precision.

```suggestion
   assert(delta >= 100);
```

MarkMcCulloh (approved):

Issue #2953, opened 2023-06-18
Array std tests


revitalbarletz:
### Feature Spec

The test will be written in Wing. 
it will cover the following methods of the array:

- [x] concat
- [x] contains
- [x] copy
- [x] copyMute
- [x] indexOf
- [x] join
- [x] lastIndexOf
- [x] pop

Please be minded to test the methods with all possible arguments compositions, and edge cases ✨ 

The tests should be placed under `examples/tests/sdk_tests/std/array.w` 

### Use Cases

testing

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

PR #2967, opened 2023-06-19
feat(sdk): array tests

revitalbarletz:
Fixes issue #2953.
## Checklist

- [ ] Tests added (always)


examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
-// push()
+// push() & pop()
 
 let a = MutArray<str>["hello"];
 assert(a.length == 1);
 a.push("world");
 assert(a.length == 2);
 assert(a.at(0) == "hello");
 assert(a.at(1) == "world");
+a.pop();
```
staycoolcall911, line 33: `array.pop()` returns the value removed from the array - need to test that too.

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+  assert(a.length == 1);
+  assert(a.at(0) == "hello");
 }
 
 //-----------------------------------------------------------------------------
-// TODO: https://github.com/winglang/wing/issues/2785
-
 // concat()
+
+let b = MutArray<str>["hello"];
```
staycoolcall911, line 52: Let's add another test where 2 `Array`s (instead of `MutArray`s) are concatenated. Maybe also concat `Array` and `MutArray`.

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+assert(b.length == 1);
+let c = MutArray<str>["wing"];
+assert(c.length == 1);
+let d = b.concat(c);
+assert(d.length == 2);
+assert(d.at(0) == "hello");
+assert(d.at(1) == "wing");
+
+test "concat()" {
```
staycoolcall911, line 61: It's absolutely great that you're testing both preflight and inflight!

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+  assert(d.at(0) == "hello");
+  assert(d.at(1) == "wing");
+}
+
+//-----------------------------------------------------------------------------
 // contains()
-// copy()
-// copyMut()
+
+let e = MutArray<str>["hello", "wing"];
```
staycoolcall911, line 75: Let's also test what happens when the searched element is not contained in the array. Maybe search more types than just string

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+
+//-----------------------------------------------------------------------------
 // indexOf()
+
+let g = MutArray<str>["hello", "wing"];
+let h = "wing";
+let index = g.indexOf(h);
+assert(index == 1);
+
+test "indexOf()" {
```
staycoolcall911, line 95: Add another test for when the item is not found in the array (returns -1).

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+  let g = MutArray<str>["hello", "wing"];
+  let h = "wing";
+  let index = g.indexOf(h);
+  assert(index == 1);
+}
+
+//-----------------------------------------------------------------------------
 // join()
+
+let i = MutArray<str>["hello", "wing"];
```
staycoolcall911, line 105: Delimiter is optional (comma by default) - should add a test for that case as well

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+  let joinedString = i.join(delimeter);
+  let expectedString = i.at(0) + delimeter + i.at(1);
+  assert(joinedString == expectedString);
+}
+
+//-----------------------------------------------------------------------------
+// TODO: https://github.com/winglang/wing/issues/2953
+// 
+// The bellow is not implemented:
+// copy()
```
staycoolcall911, line 125: Let's add tests also for these 3 additional methods.

examples/tests/sdk_tests/std/array.w
```diff
@@ -22,32 +22,106 @@ test "at()" {
...
+// copyMut()
 // lastIndexOf()
-// pop()
```
staycoolcall911, line 53: Let's also add some tests for arrays containing more complex types, so we'll guarantee they can hold struct, cloud.Bucket, user defined classes, etc.

staycoolcall911 (commented):
Nice @revitalbarletz - looks great, though missing a few more cases for some methods, e.g. array.indexOf("aaa") returns -1 if "aaa" is not found.
Please see the docstrings for `Array` and `MutArray` to cover all these cases:
https://github.com/winglang/wing/blob/main/libs/wingsdk/src/std/array.ts

examples/tests/sdk_tests/std/array.w
staycoolcall911: Strange why you couldn't find them - where are you looking? They appear in the VSCode ext. completion and also in our [reference documentation](https://www.winglang.io/docs/standard-library/sdk#@winglang/sdk.std.ImmutableArray.lastIndexOf).
```ts
let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9];
arr.copyMut();

let mutArr = MutArray<num> [1, 2, 3, 4, 5, 6, 7, 8, 9];
mutArr.copy();

log("${arr.lastIndexOf(1)}");
```

examples/tests/sdk_tests/std/array.w
revitalbarletz: The latter is not possible now - it does not compile.

examples/tests/sdk_tests/std/array.w
revitalbarletz: This is impossible - I am getting the following error: Capturing collection of preflight classes is not supported yet (type is 'Array<Bucket>')

for this: 
let bucket = new cloud.Bucket() as "myBucket";
let buckets = Array<cloud.Bucket>[bucket];

test "pushAndPopBuckets()" {
  assert(buckets.length == 1);

}

staycoolcall911 (approved):
Awesome job @revitalbarletz - thanks for closing all these loose ends!

PR #2996, opened 2023-06-20
fix: capture diagnostics missing from language server

MarkMcCulloh:
Certain errors were not available in the lsp because we didn't run the jsify phase. This PR changes that and also avoids many panics. 

So far it seems the perf overhead of running jsify in the lsp is minimal. On my PC, it adds between 1ms and 8ms. This time scales with classes/inflight/externs.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -1083,7 +1083,7 @@ impl<'a> JSifier<'a> {
 				_ => {
 					for sym in n {
 						report_diagnostic(Diagnostic {
-							message: format!("Unexpected type: \"{t}\" referenced inflight"),
+							message: format!("Unexpected type \"{t}\" referenced inflight"),
 							span: Some(sym.span.clone()),
 						});
 					}
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -291,7 +291,9 @@ impl<'a> JSifier<'a> {
 				let expression_type = self.get_expr_type(&expression);
 				let is_preflight_class = expression_type.is_preflight_class();
 
-				let class_type = expression_type.as_class().expect("type to be a class");
+				let class_type = if let Some(class_type) = expression_type.as_class() { class_type } else {
+					return "".to_string();
+				};
 				let is_abstract = class_type.is_abstract;
 
 				// if we have an FQN, we emit a call to the "new" (or "newAbstract") factory method to allow
@@ -368,10 +370,6 @@ impl<'a> JSifier<'a> {
 			ExprKind::Call { callee, arg_list } => {
 				let function_type = self.get_expr_type(callee);
 				let function_sig = function_type.as_function_sig();
-				assert!(
-					function_sig.is_some() || function_type.is_anything() || function_type.is_handler_preflight_class(),
-					"Expected expression to be callable"
-				);
 
 				let expr_string = match &callee.kind {
 					ExprKind::Reference(reference) => self.jsify_reference(reference, ctx),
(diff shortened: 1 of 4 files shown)


libs/wingc/src/lsp/sync.rs
```diff
@@ -165,6 +178,14 @@ pub mod test_utils {
 			},
 		});
 
+		let diags = get_diagnostics();
+		// none of these should be panics
+		let panic_diags = diags
+			.iter()
+			.filter(|d| d.message.starts_with("Compiler bug"))
+			.collect::<Vec<_>>();
+		assert_eq!(panic_diags.len(), 0, "Compiler bug detected: {:#?}", panic_diags);
```
Chriscbr, line 187: I see this code reused in two places. One way to keep things DRY could be to put the logic in a trait I think (`impl Vec<Diagnostic> { ... }`)
```suggestion
		diags.assert_no_panics();
```

Chriscbr (approved):
Rad

libs/wingc/src/lsp/sync.rs
Chriscbr: ok I guess in Rust you can't just do `impl Vec<Diagnostic>` but you could do something like this:

```rust
pub trait DiagnosticVecExt {
	fn assert_no_bugs(&self);
}

impl DiagnosticVecExt for Vec<Diagnostic> {
	fn assert_no_panics(&self) {
		if !self.is_empty() {
			panic!("BUG: Diagnostics should be empty at this point");
		}
	}
}
```

naming the trait with "Ext" on the end is a Rust convention IIRC

MarkMcCulloh: Deduping makes sense, but I just went the route of adding a fn to the diagnostics module. That way it was easy to throw `#[cfg(test)]` to reduce the binary size (this shouldn't be used outside of tests imo).

libs/wingc/src/jsify.rs
```diff
@@ -291,7 +291,9 @@ impl<'a> JSifier<'a> {
 				let expression_type = self.get_expr_type(&expression);
 				let is_preflight_class = expression_type.is_preflight_class();
 
-				let class_type = expression_type.as_class().expect("type to be a class");
+				let class_type = if let Some(class_type) = expression_type.as_class() { class_type } else {
+					return "".to_string();
```
eladb, line 295: Wouldn't this result in invalid JavaScript code? Feels like it should be an error, no?

libs/wingc/src/jsify.rs
```diff
@@ -1185,8 +1190,9 @@ impl<'a> JSifier<'a> {
 		// Handle parent class: Need to call super and pass its captured fields (we assume the parent client is already written)
 		let mut lifted_by_parent = vec![];
 		if let Some(parent) = &class.parent {
-			let parent_type = resolve_user_defined_type(parent, env, 0).unwrap();
-			lifted_by_parent.extend(self.get_lifted_fields(parent_type));
+			if let Ok(parent_type) = resolve_user_defined_type(parent, env, 0) {
+				lifted_by_parent.extend(self.get_lifted_fields(parent_type));
+			}
```
eladb, line 1195: Should this also be an error?

libs/wingc/src/jsify.rs
MarkMcCulloh: It will be an error: This is for a `new` expressions and our type checker requires that it be a class. We jsify despite that, so this effectively exits early. 
There's no point in adding a new diagnostic since the existing type error is more useful than anything we could say here.

libs/wingc/src/jsify.rs
MarkMcCulloh: Similar to the above, if we are referencing a udt that doesn't exist then the type checker would've already caught that

Issue #1737, opened 2023-03-07
Json Advanced Functionality


hasanaburayyan:
- [ ] Json.equals()
- [ ]  Json.diff()
- [ ]  Json.patch()
- [ ]  Json.entries()
- [ ]  Structs Schema
- [ ]  To native types:
	- [ ]  unsafe mode
	- [ ]  try_from_json()

staycoolcall911:
@hasanaburayyan - please note I added "Structs Schema" and "To native types" so this issue basically contains all our p2 work related to `json`.

staycoolcall911:
Broken down the tasks listed in this issue to several issues: #3139, #3140, #3142.
Closing

Issue #130, opened 2022-09-20
Wing libraries


staycoolcall911:
### 5.1.2 Internal Libraries
Wing libraries themselves are JSII modules. They can be used in all other JSII
supported languages.

Chriscbr:
Initial thoughts...

At a high level, supporting Wing libraries can be broken into two parts: **packaging** and **publishing**.

Packaging means turning a Wing module (collection of classes, interfaces, structs, etc.) into a shareable artifact. We have several options:
1. turn a Wing module (a `.w` file that isn't an [entrypoint file](https://github.com/winglang/wing/issues/3744)) into a JSII library[^1], as described in this issue's title
2. design some kind of new format?
3. skip packaging altogether (e.g., shipping un-compiled .w files)

The downsides of approaches (2) and (3) are that they also don't move us towards the goal of making Wing libraries usable in other JSII languages. (3) would also likely result in worse compilation and language server performance for users. (1) is my current recommendation. 

Publishing means actually sharing the Wing library artifact in a way that others can consume it. Some options:

1. let users publish Wing/JSII libraries on npm themselves
2. host all Wing libraries (including `.w` and `.js` source code) inside a central repository like `winglang/libraries`, and handle publishing for users
3. host a list of all Wing libraries (pointers to other repositories) inside a central repository like `winglang/libraries`, and handle publishing for users

Some previous discussion about hosting libraries in a central place can be found [here](https://github.com/winglang/wing/discussions/652). In options (2) and (3), we could publish libraries to npm (under a common namespace, like `@winglibs` OR we could publish all libraries under a [GitHub packages scope](https://github.com/features/packages) The appeal of (3) is it's similar to how [homebrew](https://brew.sh/) operates, and has been shown to work up to a pretty reasonable scale - each formula is represented by a single file ([example](https://github.com/Homebrew/homebrew-core/blob/master/Formula/c/cdk8s.rb)).

It might be worth to support support a combination of options, like (1 and 2) or (1 and 3). (1) is the easiest support since it doesn't require any extra work to implement assuming our packaging strategy involves producing an npm tarball. (2) could add some friction to library authors (for example, Wing maintainers may need to approve pull requests and library authors may be unable to run E2E tests in our library monorepo) - but it might be good enough to start with and then switch to (3) once there are more growing pains.

[^1]: A jsii library is just a tarball that can be published to npm, that in addition to `package.json` and JavaScript code, also has a `.jsii` file that describes all of the exported types in a language-agnostic way.

[...]

Issue #129, opened 2022-09-20
Module System - Exports


staycoolcall911:
### 4.2 Exports
In preflight, anything with `public` at block scope level is importable.
This includes functions, classes, structs and interfaces.
In inflight, all of the above excluding preflight classes are importable.
Variables are not exportable.
Preflight classes cannot be instantiated in inflight functions. There is no synthesizer inside
and no deployment system in the inflight body to provision.
"Bringing" other Wing files currently ignores exports, but bringing JSII modules
respect the visibility of JSII module exports.

staycoolcall911:
Duplicates #130

PR #2999, opened 2023-06-21
chore(docs): language reference for wing launch

### LINKED ISSUE #2420: Wing Spec for launch
### Feature Spec

We have a spec that fully describes what we release at launch, and only that

### Use Cases

So developers will not be confused when trying code from the spec and it doesn't work

### Implementation Notes

Should create a spec for the future from what we remove, or make it an RFC

### Component

Documentation

### Community Notes

### LINKED ISSUE #2696: Behavior of for loops is not properly defined
### I tried this:

for i in 1..100 {
    log("${i}");
 }

### This happened:

I got printouts of 1..99

### I expected this:

printouts of 1..100, or at least some documentation that the expected behavior is 1..99

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

I think that the behavior should be to print 1 through 100, including 100.
Or as ChatGPT puts it:
> If you're referring to a mathematical notation or range, the interval 1..100 does include both 1 and 100. This is equivalent to [1, 100] in interval notation. 

> If you're referring to a programming language, it depends on the language. In Python, the range() function doesn't include the end value, so range(1, 100) includes 1 but not 100. However, in languages like Ruby, the range 1..100 does include both 1 and 100. Always consult the documentation or specifications for the particular language you're using.

### Community Notes

staycoolcall911:
This PR introduces an up-to-date language reference - all wishlist/roadmap feature mentions were deleted and instead put into new/existing issues, with a link in a newly added "Roadmap" paragraph at the end of relevant sections..
All code samples (Wing and TypeScript) were tested and should successfully compile.

[Rendered version of the language reference](https://github.com/winglang/wing/blob/urib/lang-ref/docs/docs/03-language-guide/90-reference.md#1149-roadmap).

Fixes #2420 

### TODO
- [ ] I'll remove all typescript code samples in a follow-up PR

### Misc
- [x] Changed the rust debugger to debug the current open `.w` file by default (relevant contributor guide doc updated).
- [x] Small fix of a couple of tree-sitter test headlines.
- [x] Fixes #2696 - by defining the behavior of `for` loops.

### I simply removed the following spec features/requirements:
1. In 1.1.4.8 Json logging
“It is also legal to just log a json object”
2. In 1.2 Utility Functions:
“The above functions can accept variadic arguments of any type except `throw` which only accepts one argument and that is the message to be contained in the error.”
3. In 1.4 Storage modifiers:
“The name of any static data member and static member function must be different from the name of the containing class regardless of the casing.”
4. In 2.6 For:
“Type annotation after an iteratee (left hand side of **in**) is optional.”
5. In 3.2 Classes:
“Optionals are initialized to `nil` if omitted, unless the type is `nil?`, which in that case, absent initialization is a compile error.<br/>
Member function and field access in constructor with the "this" keyword before
all fields are initialized is invalid and would throw a compile error.<br/>
In other words, the `this` keyword is immutable to its field access operator `.`
before all the member fields are properly initialized. The behavior is similar to JavaScript and TypeScript in their "strict" mode.”
6. In 3.3 Preflight classes:
Scope and ID are “both overrideable by user-defined ones in constructor.”
7. In 3.8 Enumeration:
“Last comma is optional in single line definitions but required in multi line definitions.”
“`nameof` operator”
8. Entire section: 6.1.2 Shell strings
9. Entire section: 6.4 Kitchen Sink
10. Entire section: Inspiration (last section)

### Issues opened/changed to track spec completeness:
- [x] Opened issues:
	- [x] #3121 
	- [x] #3123 
	- [x] #3129
	- [x] #3139 
	- [x] #3140 
	- [x] #3142 
- [x] Changed description:
	- [x] #108 
	- [x] #2103 
	- [x] #116 
	- [x] #125 
	- [x] #128
	- [x] #129 
	- [x] #130
	- [x] #1737 

## Checklist

- [ ] Tests added (always)


docs/05-reference/winglang-spec.md
```diff
@@ -2363,64 +1789,59 @@ You may import JSII modules in Wing and they are considered preflight classes if
...
 };
 ```
 
 Given a method of name X, the compiler will map the method to the JavaScript export with the 
 matching name (without any case conversion).
 
-Initially we only support specifying `extern` for static methods (either inflight or preflight),
-but we will consider adding support for instance methods in the future. In those cases the first
-argument to the method will implicitly be `this`.
+We currently only support specifying `extern` for static methods (either inflight or preflight).
```
Chriscbr, line 1837: I think we can remove this, I believe instance methods are supported
```suggestion
```

docs/05-reference/winglang-spec.md
```diff
@@ -69,23 +62,15 @@ import TOCInline from '@theme/TOCInline';
 | ------ | ---------------------------------- |
 | `void` | represents the absence of a type   |
 | `nil`  | represents the absence of a value  |
```
Chriscbr, line 64: Let's remove this as well - `nil` is not a dedicated type that the user can create or specify (nor should they be able to)
```suggestion
```

Chriscbr (commented):
What a cleanup! Though there's a lot of red, it's just as impressive how much of our initial spec we do have working. I have a feeling we will have a lot of new GitHub issues after this 😄

docs/05-reference/winglang-spec.md
```diff
@@ -69,23 +62,15 @@ import TOCInline from '@theme/TOCInline';
...
-"any" must be explicitly declared and annotated.
-
 > ```TS
 > let x = 1;                  // x is a num
 > let v = 23.6;               // v is a num
 > let y = "Hello";            // y is a str
 > let z = true;               // z is a bool
-> let w: any = 1;             // w is an any
 > let q: num? = nil;          // q is an optional num
 > ```
```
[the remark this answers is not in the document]
staycoolcall911: I fully agree. Removed the "Equivalent TypeScript Code" section from 1.1.1 Primitive Types

docs/05-reference/winglang-spec.md
```diff
@@ -229,71 +194,50 @@ let jsonMutObj = MutJson {
...
-// RUNTIME ERROR: Uncaught TypeError: Cannot read properties of undefined (reading 'world')
-```
-
-Like in JavaScript, it is also possible to access object fields using `[]`:
-
-```js
-let foo = j["my-field"].yourField["their-field"];
+```TS
+log(jsonObj.get("boom").get("dude").get("world"));
+// RUNTIME ERROR: Cannot read properties of undefined (reading 'world')
```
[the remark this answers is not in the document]
staycoolcall911: Nice catch

staycoolcall911:
@Chriscbr - do you know what's the planned location of the winglang reference file in the docs? It would be great if you could put the current version of winglang-spec in there so I'll point my file changes to this file and we could see the diff.

docs/docs/03-language-guide/90-reference.md
```diff
@@ -229,71 +213,50 @@ let jsonMutObj = MutJson {
...
-
-```js
-log(jsonObj.boom.dude.world);
-// RUNTIME ERROR: Uncaught TypeError: Cannot read properties of undefined (reading 'world')
+```TS
+let boom: Json = jsonObj.get("boom");
 ```
 
-Like in JavaScript, it is also possible to access object fields using `[]`:
+Trying to access a non-existent field will fail at compile time. For example:
```
Chriscbr, line 230: I think it would have to fail at runtime, since we can't always know ahead of time what value or fields are contained in a `Json` variable.
```suggestion
Trying to access a non-existent field will fail at runtime. For example:
```

docs/docs/03-language-guide/90-reference.md
```diff
@@ -2093,33 +1719,20 @@ is safe to omit optional struct fields, or have order of arguments mixed.
...
+The following features are not yet implemented, but we are planning to add them in the future:
 
-```TS
-let f = (x: num, ...args: Array<num>) => {
-  log(x + y + args.len);
-}
-// last arguments are expanded into their array
-f(1, 2, 3, 4, 5, 6, 34..100);
-```
+* Variadic arguments ()`...args`) - see https://github.com/winglang/wing/issues/125 to track.
```
Chriscbr, line 1735:
```suggestion
* Variadic arguments (`...args`) - see https://github.com/winglang/wing/issues/125 to track.
```

Chriscbr (approved):

Issue #2291, opened 2023-04-27
unhelpful error message for heterogeneous json 


ainvoner:
### What needs to be done to fix this issue

To address this issue we have decided to make `Json` the default inferred object type for `{}` over Map. See #2751 
So the necessary changes to close this issue are:
- [ ] Make the `Json` modifier optional for `{}` objects and inferred if it is not explicitly stated (e.g. both `Json {a: "a"}` and `{a: "a"}` will be inferred by our parser as `Json`s).
- [ ] Make `Map` modifier required for map objects.
- [ ] Change map syntax from `Map {a: "a"}` to `Map {a => "a"}`
### I tried this:

```
class TestMe {
    a: Json;
    init() {
        this.a = {text: "text", number: 0};
    }
}

let t = new TestMe();
```

### This happened:

since it is actually a map and not a Json object, I got this unhelpful error message:
```js
    error: Expected type to be "str", but got "num" instead
  --> test.w:4:41
  |
  |         this.a = {text: "text", number: 0};
  |                                         ^ Expected type to be "str", but got "num" instead
...
```

### I expected this:
The error will notify me that i'm trying to assign a Map to a Json variable.
the correct code is:

```
class TestMe {
    a: Json;
    init() {
        this.a = Json {text: "text", number: 0};
    }
}

let t = new TestMe();
```

### Component

Compiler

### Wing Version

0.13.35

### Node.js Version

19.3.0

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

ekeren:
but actually, @staycoolcall911 has a point saying that we should use the `{}` for json and an explicit `Map<str> {`
```
let jsonIsTheDefault =  {text: "text", number: 0}; // should work
let mapIsTheException =  Map {text: "text", number: 0};
//                                                                                       ^ Expected type to be "str", but got "num" instead
```

monadabot:
Congrats! :rocket: This was released in [Wing 0.23.0](https://github.com/winglang/wing/releases/tag/v0.23.0).

PR #3018, opened 2023-06-21
feat(compiler)!: change map syntax to use `=>`

hasanaburayyan:
Makes `{}` default to Json as well as changes maps to use `=>` syntax

Closes: https://github.com/winglang/wing/issues/2291

BREAKING CHANGE: `std.Map` syntax now uses `=>` instead of `:`

## Checklist

- [x] Tests added (always)


examples/tests/valid/json.w
yoav-steinberg, line 1: Is there a test for the empty literal case (`let x = {}`). To verify it's a json now and not a map/set?

yoav-steinberg (approved):
Looks good

examples/tests/valid/json.w
hasanaburayyan: There is a grammar test, but might be worth making sure I have a wing test as well Ill add it.

examples/tests/valid/json.w
```diff
@@ -52,7 +52,7 @@ someJson.set("x", 111);
 assert(someJson.get("x") == 111);
 
 // assign Map to Json
-let x: Json = {cool: "beans"};
+let x: Json = {cool => "beans"};
```
Chriscbr, line 55: This is Json, not Map, right? This change shouldn't be needed (`Json { a: b }` is still the syntax used in the spec)

Chriscbr (changes requested):

examples/tests/valid/json.w
hasanaburayyan: @Chriscbr so both would be valid

```
let x: Json = {cool: "beans"}; // Json map
OR
let x: Json = {cool => "beans"}; // non-json map
```

I only made the change to show that we can still assign map elements into Json. Like we do with arrays `Json [1, 2, 3]`

hasanaburayyan: Actually this is a current issue where everything is a subtype of Json. Im going to add a subsequent PR with plans to remove that. 

For now we can see the desired spec behavior here: https://github.com/winglang/wing/pull/3018/commits/c295f9ac0a4ff33768cf4e1dc338f98db2bdf89e#diff-25a9b7a88f48644f959355fc63e91f28fbe6751861da3ec0b689da4d37c141efR133-R136

Chriscbr: Got it. I'll be honest, allowing both `Json { a: "b" }` and `Json { a => "b" }` as two ways to construct the same object feels a bit odd (like we've made some mistake in our syntax design somewhere) -- but maybe it will be uncommon enough to happen in practice that it won't be an issue. In any case, I'm down to try this new syntax

Chriscbr (approved):

examples/tests/valid/json.w
hasanaburayyan: Yea I actually really dislike this when working within Structs that contain maps. It feels wonky

```js
struct A {
	foo: str;
	bar: Map<str>;
}

let a = A {
	foo: "foofoo",
	bar: {
		x => "a"
	}
}
```

yoav-steinberg: > Got it. I'll be honest, allowing both `Json { a: "b" }` and `Json { a => "b" }`

I think the 2nd option above isn't legal in our syntax:
```wing
let a = Map<str> { a => "b" }; // a map
let b =  { a => "b" }; // a map initialized with a map without the explicit type
let c: Json =  { a => "b" }; // a Json initialized with a map
let d = Json { a: "b" }; // a Json initialized with a Json object
let e = { a: "b" }; // a Json initialized with a Json object without the explicit type
let f: Map<str> = { a: "b" }; // complication error: can't initialize a map with a Json literal
```

eladb: It makes sense to me that `Map` and `Set` won't be implicitly casted to Json (because they don't have a reversible representation). But primitive types and arrays should be implicitly casted.

examples/tests/valid/json_static.w
```diff
@@ -1,7 +1,7 @@
 bring cloud;
 
 // Get Keys
-let x = Json {a: 123, b: {c: 456, d: 789}};
+let x = Json {a: 123, b: {c => 456, d => 789}};
```
eladb, line 4: This shouldn't work

examples/tests/valid/website_with_api.w
```diff
@@ -9,9 +9,9 @@ let usersTable = new cloud.Table(
   name: "users-table",
   primaryKey: "id",
   columns: {
-    id: cloud.ColumnType.STRING,
-    name: cloud.ColumnType.STRING,
-    age: cloud.ColumnType.NUMBER,
+    id => cloud.ColumnType.STRING,
```
eladb, line 12: Are we also allow keys of maps to be symbolic (`id` instead of `"id"`). I think we should only allow this for json.

```suggestion
    "id" => cloud.ColumnType.STRING,
```

eladb (dismissed):

examples/tests/valid/json_static.w
hasanaburayyan: I have removed map and set from being valid json fields.

examples/tests/valid/website_with_api.w
hasanaburayyan: hmm, it would be an easy change to make but why would we not allow symbolic keys in maps?

examples/tests/valid/json.w
hasanaburayyan: I made some updates here to not allow maps and sets to be valid json. I also added distinct ExprKind for JsonMapLiteral so we arent overloading the use of MapLiteral in typechecking and jsification to deal with both normal maps and Json. Keeps the responsibilities separate I think.

eladb: A winglang map becomes a JavaScript `Map` object. When JSON.stringified it becomes a "regular" JavaScript object, which means that when parsed back to JavaScript it won't be a Map anymore:

```js
const y = new Map();
const x = JSON.parse(JSON.stringify(y));
assert(y instanceof Map); // ok!
assert(x instanceof Map); // boom!
```

Chriscbr: @eladb I'm afriad you're mistaken - Winglang maps (as of this comment) are code-generated as plain JavaScript objects (not [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) objects) - you can try compiling a program in the playground to validate this. This behavior is necessary to interoperate with the CDKTF / AWS CDK libraries, where any type declared in TypeScript as `Record<string, T>` or `{ [key: string]: T }` is treated as a map in the JSII model.

staycoolcall911: Discussed with @eladb -he's ok with the current state. Resolving.

examples/tests/valid/website_with_api.w
staycoolcall911: @hasanaburayyan - our maps are Map<str, T> (map from string to T), so let's enforce that map keys are strings if that's easy. If not, we can save that for a future PR

staycoolcall911 (approved):

examples/tests/valid/website_with_api.w
hasanaburayyan: I updated the grammar to only accept keys that are strings

PR #3021, opened 2023-06-22
feat: integrate the console in the cli

### LINKED ISSUE #723: Wing Console Linux support
### Summary

Ability to run Wing console on Linux

:penguin: 

### Component

Wing Console

### LINKED ISSUE #1831: Console crashes due to misaligned SDK versions
### Problem

We are experiencing occasional console crashes due to a misalignment between the bundled SDK version in the console and the global SDK version used to develop and compile the wing application. We have brainstormed several solutions to eliminate this problem and would like to request feedback on the best approach.

### Possible solutions

1. Split the SDK into two packages, where one package contains APIs for listing resources, controlling the simulator state, etc., and the other package contains the logic for each resource. While this approach would eliminate problems in adding new resources, it would not prevent errors due to inflight method and other internal changes that might crash the console.

2. Implement the "custom resource" solution, where each resource has a "Display"/"Render" object that describes how it is rendered in the console (Map, interaction view and all inflight methods). In this solution the console in un aware of the internal implementation of the SDK and only needs to be familiar with a small subset of API's.

3. Integrate the wing console into the wing repository, where each wing release will create a new console release with the new SDK version.

4. Include wing console tests as part of every wing PR to indicate if the PR breaks the console or not. However, this approach only provides an indication that the console is about to break without being able to prevent it.

5. follow standard procedure for deprecating code and interface changes in the SDK. 

### Suggested solution for the beta

Since some of these solutions may not be realistic for the beta, we need to decide which steps to take now and which to take post-beta. Here are some simple and immediate steps we can take:

1. Auto-update the console dependencies and release a new version of the console after each wing release. This won't prevent breaking changes, however we will get an immediate indication the there was a breaking change in the SDK that could crash the console (the console release flow will fail on failed tests).

2. Add a job for each wing PR to test the latest wing console (main branch) with the current wing PR branch by linking the SDK and wing. The process will provide us with immediate feedback if the console breaks and allow us to fix it before the wing release. This process will let the console team time to refactor the console to work with the future wing release.

3. UI/UX modification in the Console. Whenever loading a new .w file, check the bundled and global versions of the SDK.
- If the console SDK version is more advanced, prompt the user to update their global wing version.
-  If the console SDK version is older and the console version is not the latest, prompt the user to update the console.
- If the console SDK version is older and the console version is the latest, prompt the user that the wing application version is not yet supported by the console.

### Component

SDK, Wing Console

### LINKED ISSUE #2357: Ensure the Console builds correctly during the Wing CI

### LINKED ISSUE #2617: Publish the Console as an npm package
Currently, the Console consists on these packages:

- `@wingconsole/server`: a package that creates an express server that handles the simulation
- `@wingconsole/ui`: a package that exports react components that provide the whole UI
- `@wingconsole/app`: an Electron app that uses the server and UI packages

None of these packages are published to NPM since we bundle the Electron app and publish it to a bucket. Nonetheless, the Playground would really benefit from having a package in NPM that creates a server using both `@wingconsole/server` and `@wingconsole/ui`.

So, my suggestion is to do the following changes:

1. Rename `@wingconsole/app` to `@wingconsole/desktop`
2. Create a new `@wingconsole/app` package that uses `@wingconsole/server` and `@wingconsole/ui`. This one doesn't include anything related to electron
3. Publish `@wingconsole/app` to NPM
4. Change the Playground and Tour so they run `npm install winglang @wingconsole/app`, basically (additionally, the webcontainer should have a package-lock.json so it's faster to install)

These changes should make the Playground and Tour much faster to initialize, while keeping the maintenance and costs to a minimum.

### LINKED ISSUE #3005: feat: use the console in the wing cli
- Add `@wingconsole/app` as a dependency to the `winglang` package
- Change the `wing it` command to start the console app

Depends on https://github.com/winglang/wing/issues/3002.

skyrpex:
- Adds `@wingconsole/app` as a depedency to `winglang`
- Changes `wing it` so it creates a Console server
- Opens the Console URL
- Stops referring to simfiles in favor of wingfiles

Fixes #3005, fixes #1831, fixes #2617, fixes #723 and fixes #2357.


apps/wing/package.json
```diff
@@ -27,6 +27,10 @@
     "package": "bump-pack -b"
   },
   "dependencies": {
+    "@wingconsole/app": "file:../wing-console/console/app",
+    "@winglang/compiler": "file:../../libs/wingcompiler",
+    "@winglang/sdk": "file:../../libs/wingsdk",
+    "add": "^2.0.6",
```
MarkMcCulloh, line 33: What is this dep used for?

skyrpex: It was a typo. Removed.

apps/wing/src/commands/run.test.ts
```diff
@@ -85,3 +111,43 @@ test("wing it with an invalid file throws exception", async () => {
...
+      requestedPort: 5000,
+      hostUtils: expect.anything(),
+    });
+    expect(open).toBeCalledWith("http://localhost:5000/");
+  } finally {
+    process.chdir(prevdir);
+  }
+});
+
+test("wing it with an invalid port ignores it", async () => {
```
staycoolcall911, line 135: Thanks for adding the test - makes it easier to spot that this is what happens when an invalid `port` is passed.
It is better to show an error to the user that they passed wrong input than silently ignore.

apps/wing/src/cli.ts
```diff
@@ -49,8 +49,9 @@ async function main() {
   program
     .command("run")
     .alias("it")
-    .description("Runs a Wing simulator file in the Wing Console")
-    .argument("[simfile]", ".wsim simulator file")
+    .description("Runs a Wing program in the Wing Console")
+    .argument("[entrypoint]", "program .w entrypoint")
```
MarkMcCulloh, line 53: Would it be possible for this to work with both a .w file and .wsim dir? If not, I think it's fine to just work with a .w IMO, but I can see the usefulness of being able to open a .wsim without compilation.

apps/wing/src/commands/run.ts
```diff
@@ -2,21 +2,53 @@ import { readdirSync, existsSync } from "fs";
 import { debug } from "debug";
 import { resolve } from "path";
 import open from "open";
+import { createConsoleApp } from "@wingconsole/app";
+import { parseNumericString } from "../util";
 
-export async function run(simfile?: string) {
-  if (!simfile) {
+/**
+ * Run options for the `run` command.
```
MarkMcCulloh, line 9:
```suggestion
 * Options for the `run` command.
```

apps/wing/src/commands/run.ts
```diff
@@ -2,21 +2,53 @@ import { readdirSync, existsSync } from "fs";
...
+ * This is passed from Commander to the `run` function.
```
MarkMcCulloh, line 10: Maybe something like

```suggestion
 * Customizes the behavior of the created Wing Console
```

apps/wing/src/commands/run.ts
```diff
@@ -2,21 +2,53 @@ import { readdirSync, existsSync } from "fs";
...
+    throw new Error(entrypoint + " doesn't exist");
   }
 
-  simfile = resolve(simfile);
-  debug("calling wing console protocol with:" + simfile);
-  await open("wing-console://" + simfile);
+  entrypoint = resolve(entrypoint);
+  debug("opening the wing console with:" + entrypoint);
+
+  const { port } = await createConsoleApp({
```
MarkMcCulloh, line 42: This stays open after the Wing CLI is done processing the command, right?

MarkMcCulloh (approved):

apps/wing/src/cli.ts
skyrpex: It's technically possible but we currently assume it's a wingfile that needs compilation. We could make the distinction based on the extension and skip the compilation if necessary, but I'd suggest doing it in another issue.

apps/wing/src/commands/run.ts
skyrpex: It should stay open forever (until you Ctrl+C).

PR #3032, opened 2023-06-22
chore(cli): implicitly set progress option in cli for cloud targets

### LINKED ISSUE #2903: Change the test option to always show the progress when running tests on the cloud (but not in the simulator)
### Decided

The bottom line: we have decided to a keep the `--progress` option, but give it different defaults (`false` on sim target, `true` on cloud targets).
You can read the original issue reported below. 

### Feature Spec - original issue

Currently, when running a test on AWS (or any other cloud provider), we are left waiting for the execution without knowing if it's running or frozen.

This happens because tests require the `--progress` option to be used so that we can see the progress of the execution.

It would be more interesting if the default behavior was to always display the progress, and if we wanted to suppress excessive messages, we could use a new option `--quiet`.

image

### Use Cases

Displaying the progress when running a test pointing to a cloud provider.

### Implementation Notes

We would need to create an option in the file `apps/wing/src/cli.ts` with the new option `--quiet`.

Remove the `--progress` option.

And modify the file that runs the tests to include this new option.

### Component

_No response_

### Community Notes

garysassano:
Closes #2903 


MarkMcCulloh (changes requested):

apps/wing/src/cli.ts
```diff
@@ -75,9 +81,15 @@ async function main() {
...
-        .choices(["tf-aws", "sim", "awscdk"])
+        .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
         .default("sim")
     )
+    .hook("preAction", async (cmd) => {
+      const target = cmd.opts().target;
+      if (target != "sim") {
+        process.env.PROGRESS = "1";
+      }
+    })
```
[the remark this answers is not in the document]
garysassano: You can't add it to the root program object, because th `cmd.opts()` is able to see the parameters passed only to that specific command.

apps/wing/src/cli.ts
```diff
@@ -64,6 +64,12 @@ async function main() {
         .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
         .default("sim")
     )
+    .hook("preAction", async (cmd) => {
+      const target = cmd.opts().target;
+      if (target != "sim") {
+        process.env.PROGRESS = "1";
+      }
+    })
```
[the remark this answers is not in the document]
garysassano: The fact is that there's no `--no-progress` since the default is `false` right now. So you are suggesting to add that option?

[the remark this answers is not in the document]
MarkMcCulloh: I was thinking of the following case:

- `wing test -t tf-aws --no-progress`

Which I would expect to not show progress

apps/wing/src/cli.ts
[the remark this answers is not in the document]
MarkMcCulloh: Ah, gotcha. WDYT about refactoring the logic to a new function so the hooks can call it?

apps/wing/src/cli.ts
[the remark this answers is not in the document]
garysassano: I proceeded to add the `--no-progress` option and now it works the way you described.

apps/wing/src/cli.ts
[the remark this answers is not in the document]
garysassano: I don't think that's actually possible for the same reason why we duplicated the target option inside both the `compile` and `test` commands. If we were to define a `target` option at a higher level, it would have been always usable also for commands where it made no sense to be able to set it.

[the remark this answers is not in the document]
garysassano: I don't think that's actually possible for the same reason why we duplicated the `target` option inside both the `compile` and `test` commands. If we were to define a `target` option at a higher level, it would have been always usable also for commands where it made no sense to be able to set it.

apps/wing/src/cli.ts
```diff
@@ -65,6 +67,13 @@ async function main() {
         .default("sim")
     )
     .option("-p, --plugins [plugin...]", "Compiler plugins")
+    .hook("preAction", async (cmd) => {
+      const target = cmd.opts().target;
+      const progress = program.opts().progress;
+      if (progress !== false && target !== "sim") {
+        process.env.PROGRESS = "1";
+      }
+    })
```
MarkMcCulloh, line 76: Okay so it is possible to add this to the top level program, I had to do a little digging though because commander's API is kinda confusing:

```ts
program.hook("preSubcommand", (_: Command, cmd: Command) => {
	const options = cmd.optsWithGlobals();
  const target = options.target;
  const progress = options.progress;
  if (progress !== false && target !== "sim") {
    process.env.PROGRESS = "1";
  }
});
```

(note the `preSubcommand`)

With this you'll be able to have this only once, with access to both target and progress.

apps/wing/src/cli.ts
```diff
@@ -36,6 +36,8 @@ async function main() {
     process.env.PROGRESS = "1";
   });
 
+  program.option("--no-progress", "Don't show compilation progress", () => {});
```
MarkMcCulloh, line 39: Not sure this is needed

```suggestion
  program.option("--no-progress", "Don't show compilation progress");
```

apps/wing/src/cli.ts
```diff
@@ -36,6 +36,8 @@ async function main() {
     process.env.PROGRESS = "1";
```
MarkMcCulloh, line 36: Since we should be able to get the logic of setting this env into one place, it may make sense to get rid of this and move it to the other function so that all the logic stays together.

That way, we can handle all three cases together: if progress is true, if it's false, and if it's undefined.

MarkMcCulloh (commented):
Nice, I think it's functionally correct now. Just had some more comments on refactoring.

apps/wing/src/cli.ts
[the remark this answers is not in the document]
MarkMcCulloh: > If we were to define a target option at a higher level, it would have been always usable also for commands where it made no sense to be able to set it.

I see your point, I would be okay with duplicating the option's definition (they can even have different descriptions) as long as the way we handle setting the env is consolidated by taking the function passed to the hook and extracting it.

apps/wing/src/cli.ts
garysassano: There is a problem with your solution I think. The "preSubcommand" `HookEvent` gets executed before the `test` and `compile` commands. For some reasons the value of `target` option which you specify in the terminal doesn't get passed to the "Subcommand", resulting in an `undefined` target value, which ends up setting the default value of `sim`.

```ts
program.hook("preSubcommand", (_: Command, cmd: Command) => {
  const options = cmd.optsWithGlobals();
  const target = options.target;
  const progress = options.progress;
  log(options);
  log(target);
  if (!progress && target !== "sim") {
    process.env.PROGRESS = "1";
  }
});
```

```console
$ mywing test -t tf-aws examples/tests/sdk_tests/function/invoke.w
```

```console
{ target: 'sim', updateCheck: true }
sim
```

garysassano: There is a problem with your solution I think. The "preSubcommand" `HookEvent` gets executed before the `test` and `compile` commands. For some reasons, the value of `target` option which you specify in the terminal doesn't get passed to the "Subcommand", resulting in an `undefined` target value, which ends up setting the default value of `sim`.

```ts
program.hook("preSubcommand", (_: Command, cmd: Command) => {
  const options = cmd.optsWithGlobals();
  const target = options.target;
  const progress = options.progress;
  log(options);
  log(target);
  if (!progress && target !== "sim") {
    process.env.PROGRESS = "1";
  }
});
```

```console
$ mywing test -t tf-aws examples/tests/sdk_tests/function/invoke.w
```

```console
{ target: 'sim', updateCheck: true }
sim
```

garysassano: There is a problem with your solution I think. The "preSubcommand" `HookEvent` gets executed before the `test` and `compile` commands. For some reason the value of `target` option which you specify in the terminal doesn't get passed to the "Subcommand", resulting in an `undefined` target value, which ends up setting the default value of `sim`.

```ts
program.hook("preSubcommand", (_: Command, cmd: Command) => {
  const options = cmd.optsWithGlobals();
  const target = options.target;
  const progress = options.progress;
  log(options);
  log(target);
  if (!progress && target !== "sim") {
    process.env.PROGRESS = "1";
  }
});
```

```console
$ mywing test -t tf-aws examples/tests/sdk_tests/function/invoke.w
```

```console
{ target: 'sim', updateCheck: true }
sim
```

apps/wing/src/cli.ts
```diff
@@ -36,6 +36,8 @@ async function main() {
...
   });
 
+  program.option("--no-progress", "Don't show compilation progress");
+
   program
     .option("--no-update-check", "Skip checking for toolchain updates")
```
MarkMcCulloh, line 42:
```suggestion
  program
    .option("--no-progress", "Hide show compilation progress");
    .option("--no-update-check", "Skip checking for toolchain updates")
```

apps/wing/src/cli.ts
MarkMcCulloh: Ah okay, fair enough. Sorry for hitting on this so hard and thinks for figuring out the root issue, just really wanted to avoid duplication if possible.

I still think my fallback option is possible, but I'll approve the change regardless. Basically, it is still possible to extract a function:

```ts
const program = new Command();

// ...

const progressHook = (cmd: Command) => {
  const target = cmd.opts().target;
  const progress = program.opts().progress;
  if (progress !== false && target !== "sim") {
    process.env.PROGRESS = "1";
  }
}
```

Then you can use the same function in both hooks (`program` is captured in the scope):

```ts
program.command("compile")

// ...

.hook("preAction", progressHook)
```

MarkMcCulloh (approved):

apps/wing/src/cli.ts
garysassano: Thanks for the suggestion, I honestly didn't think about. I removed the duplicated code.

garysassano: Thanks for the suggestion, I honestly didn't think about that. I removed the duplicated code.

Issue #2165, opened 2023-04-20
SDK reference documentation enhancements


staycoolcall911:
### Feature Spec

Our [SDK reference documentation](https://docs.winglang.io/reference/sdk) needs some love.
Some of the things that should be addressed:
- [x] The initializers section (see [example](https://docs.winglang.io/reference/sdk#initializers--24)) only makes sense for our `cloud` package objects (e.g. Bucket, Queue, etc.). It should be fixed for `std` package objects such as `Map`, 'Json', etc. (note that it currently calls `bring std;` which is done implicitly in wing for the `std` package). - https://github.com/winglang/wing/issues/3162
- [x] Change [ImmutableArray](https://docs.winglang.io/reference/sdk#immutablearray-) => Array, MutableArray => MutArray, etc. (for map, set as well).
- [x] https://github.com/winglang/wing/issues/3161
- [x] Add a link from the [Wing language spec](https://docs.winglang.io/reference/spec) to the relevant types in the [reference docs](https://docs.winglang.io/reference/sdk).
- [x] While you're there, please also add the `redis` package to `VISIBLE_SUBMODULES` and fremove `fs` (which was deleted) in https://github.com/winglang/wing/blob/main/apps/jsii-docgen/src/docgen/view/wing-filters.ts.
- [ ] Should also make sure the @return docstring of methods is visible (see for example [Array.ts.indexOf](https://github.com/winglang/wing/blob/91fae0528452d538e52aac196b95232d24e60e62/libs/wingsdk/src/std/array.ts#L80)).

### Use Cases

A list reference documentation for wing users

### Implementation Notes

We extended [jsii-docgen](https://github.com/cdklabs/jsii-docgen), so  most fixes are expected to be done in https://github.com/winglang/wing/blob/main/apps/jsii-docgen/src/docgen/transpile/wing.ts

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

ekeren:
@Chriscbr according to @staycoolcall911 you are already on it

Chriscbr:
Initializers are hidden for types like MutArray and Array but the docs for static methods still shows `std.MutArray` etc. which is a bug

ekeren:
opened other bugs, closing this for now

PR #3035, opened 2023-06-22
chore(docs): stub cloud docs, collapse categories, update simulator docs

Chriscbr:
Highlights:
- Rename category from "CLI, tools, and plugins" to just "Tools"
- Stub out cloud resource docs. I added some descriptions but they're just meant as a starting point
- Collapse categories by default (by community feedback)
- Refresh the docs on the Simulator. The information here isn't too important since accessing the simulator APIs directly isn't a main use case we are supporting, but I decided to refresh it since the page will be useful to link to or add information to when anyone asks about the simulator. 

## Checklist

- [x] Title matches [Winglang's style guide](https://docs.winglang.io/contributors/pull_requests#how-are-pull-request-titles-formatted)
- [x] Description explains motivation and solution
- [x] Tests added (always)
- [x] Docs updated (only required for features)
- [x] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Monada Contribution License](https://docs.winglang.io/terms-and-policies/contribution-license.html)*.

PR #3036, opened 2023-06-23
feat(docs): add Redis to api reference, and other fixes

Chriscbr:
Highlights:
- generate docs for redis.Redis
- rename built-in types from ImmutableArray to Array and so on so they're generated nicely in our docs
- hide docs for T1
- hide initializer docs for built-in types like String and Number

Addresses most parts of #2165

## Checklist

- [x] Tests added (always)


libs/wingc/src/lsp/completions.rs
```diff
@@ -379,14 +379,8 @@ fn add_std_namespace(type_: &str) -> std::string::String {
...
-		"Map" => "ImmutableMap",
-		"Array" => "ImmutableArray",
-		"MutSet" => "MutableSet",
-		"MutMap" => "MutableMap",
-		"MutArray" => "MutableArray",
-		"Json" | "MutJson" | "MutableArray" | "MutableMap" | "MutableSet" | "ImmutableArray" | "ImmutableMap"
-		| "ImmutableSet" | "String" | "Duration" | "Boolean" | "Number" => type_name,
+		"Json" | "MutJson" | "MutArray" | "MutMap" | "MutSet" | "Array" | "Map" | "Set" | "String" | "Duration"
+		| "Boolean" | "Number" => type_name,
 		_ => return type_name.to_string(),
```
MarkMcCulloh, line 384: looks like this whole match isn't needed anymore

libs/wingsdk/src/std/array.ts
```diff
@@ -10,14 +10,16 @@ import { Code, InflightClient } from "../core";
...
-export class ImmutableArray {
+export class Array {
   /**
    * @internal
    */
   public static _toInflightType(): Code {
     return InflightClient.forType(__filename, this.name);
   }
 
+  private constructor() {}
```
MarkMcCulloh, line 21: Is this needed?

libs/wingsdk/test/testing/__snapshots__/simulator.test.ts.snap
MarkMcCulloh, line 1: another non-deterministic snapshot?

MarkMcCulloh (approved):
Nice JSII patch, would be cool to send upstream as well

libs/wingsdk/src/std/array.ts
Chriscbr: It's not needed for anything in the compiler, but when the constructor is private, the "Initializer" section is removed the generate docs - which is nice since we don't want to advertise to users that these classes are meant to be initialized with `new std.String` etc. Of course we can customize how we generate docs for `std` types differently / revert this if needed in the future

libs/wingsdk/test/testing/__snapshots__/simulator.test.ts.snap
Chriscbr: Fixed in my other PR (https://github.com/winglang/wing/pull/3035) just need to merge in the changes

libs/wingc/src/lsp/completions.rs
Chriscbr: Hmm, if I remove the whole match, I get several test failures in the LSP. I'll leave this as is since refactoring this seems better to leave for another PR

PR #3037, opened 2023-06-23
chore: incident github template

staycoolcall911:
Adds an issue template for production incidents. Intended to make life easier for on-call incident respondents.

Screenshot when creating the issue:
image
Screenshot after the issue is created:
image

## Checklist

- [ ] Tests added (always)


.github/ISSUE_TEMPLATE/incident.yml
```diff
@@ -0,0 +1,30 @@
...
+    id: checklist
+    attributes:
+      label: Checklist
+      description:
+      options:
+        - label: ACKNOWLEDGE the incident on PagerDuty and BetterStack.
+        - label: PRIORITIZE
+            P0 - Extreme/Total user interruption, no reasonable workaround.
+            P1 - Major degradation of user experience, with possible workarounds.
+            P2 - Minor degradation of user experience with workaround, or major contributor experience disruption.
```
Chriscbr, line 24: A nice enhancement could be to have PagerDuty incidents labeled automatically with a default priority, based on what triggered it. For example, when the canary fails, by default it's labeled as P0 since it usually means Wing can't be installed or a critical path isn't working. If the main build pipeline, it could be automatically labeled P1. Of course, in either case the oncall can downgrade the priority if they determine there's a reasonable workaround or it's a false alarm. (disclaimer: this was how oncall worked at Amazon)

Chriscbr (approved):

.github/ISSUE_TEMPLATE/incident.yml
```diff
@@ -0,0 +1,30 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        A production incident started. Please stay cool, no need to call 911. Follow the checklist below - you've got this :muscle: <br/>
+        You may refer to the [Wing maintainers operational guide](https://www.winglang.io/docs/contributors/maintainers/operations) for more details.
+  - type: markdown
+    attributes:
+      value: |
+        ## Incident Response Checklist
+  - type: checkboxes
```
eladb, line 14: Can you share a screenshot of how this will look:
1. when creating the issue?
2. after the issue is created.

I suspect the checklist form is not what we want but rather just some text that contains markdown checklists (e.g. `- [ ] foo`).

eladb (changes requested):

staycoolcall911:
Some comments after discussing with @eladb:
- [ ] Pin the issue
- [ ] Announce template
- [ ] Format priority list
- [ ] DOCUMENT -> UPDATE
- [ ] Add a section - is there a workaround
- [ ] Break acknowledge to 2 items
- [ ] Add to documentation
- [ ] Unpin issue
- [ ] Close the incident issue
- [ ] Update on sla

ck

.github/ISSUE_TEMPLATE/incident.yml
staycoolcall911: I checked and it should actually remain as I initially did.
Screenshot when creating the issue:
image
Screenshot after the issue is created:
image

.github/ISSUE_TEMPLATE/incident.yml
staycoolcall911: It's a nice enhancement, thank you!
For now I prefer to keep prioritization manual, though I believe we should automate as much as we can as our incident response process matures.

staycoolcall911:
Going to merge that in since all verbal comments made by @eladb were met. It's a good starting point if I may say so and we'll iterate on it.

PR #3038, opened 2023-06-24
fix(compiler): assert doesn't work with string interpolation

### LINKED ISSUE #1833: assert doesn't work with string interpolation

Chriscbr:
Fixes #1833

The high level idea behind this fix is to avoid using backtick-strings in our generated JavaScript. These lead to a lot of issues that can be avoided by using other concatenation strategies that do not introduce grammar complexities. In this case, I decided to use `String.raw()`, but translating interpolated strings into a sequence of `+` operations probably could have worked too.

The other major change I did was improving the errors emitted by `assert()` so that it shows the original Wing source code that you asserted, not the emitted JavaScript. For example, now an error looks like:

```
assertion failed: b.list().length == 1
```
     instead of:  
```
assertion failed: '((await b.list()).length === 1)'
```

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify.rs
```diff
@@ -1961,3 +1984,51 @@ fn jsify_type_name(t: &Vec<Symbol>, phase: Phase) -> String {
 		p.join(".")
 	}
 }
+
+fn lookup_span(span: &WingSpan, source_path: &Path) -> String {
+	let source = std::fs::read_to_string(source_path)
```
eladb, line 1989: Does it make sense to read this from a file every time we are doing a lookup? Maybe we can just store the source somewhere and propagate down?

How about creating something like `CompilerContext` object that maintains all the sources as well as e.g. `Types`?

(I am less concerned about performance and more about the fact that very soon we will have multiple sources and then it makes sense to manage them in a centralized facility).

libs/wingc/src/jsify.rs
```diff
@@ -1961,3 +1984,51 @@ fn jsify_type_name(t: &Vec<Symbol>, phase: Phase) -> String {
...
+			result.push('\n');
+		}
+
+		result.push_str(&lines[end_line][..end_col]);
+	}
+
+	result
+}
+
+fn escape_javascript_string(s: &str) -> String {
```
eladb, line 2018: maybe a couple of a small unit tests for these helper functions?

libs/wingc/src/jsify.rs
```diff
@@ -1961,3 +1984,51 @@ fn jsify_type_name(t: &Vec<Symbol>, phase: Phase) -> String {
...
+	let mut result = String::new();
+
+	for c in s.chars() {
+		match c {
```
eladb, line 2022: can you reference the source specification that tells us this is an exhaustive list?

eladb (approved):

libs/wingc/src/jsify.rs
Chriscbr: Updated to use the `Files` utility. I like the compiler context object idea though we have another object also already named `CompilationContext` and I'm not sure if I should be touching it because it's part of a thread_local

libs/wingc/src/jsify.rs
Chriscbr: Done

libs/wingc/src/jsify.rs
Chriscbr: Added a comment

PR #3041, opened 2023-06-25
fix(compiler): prevents a Json string variable from being stringified in a string interpolation

### LINKED ISSUE #3027: json string keys are extracted with extra `""`
### I tried this:

test on sim:
```
let obj = Json {"key": "val"};

let s = "path/with/${obj.get("key")}";
log(s);
```

### This happened:

printed: 
`path/with/"val"`

### I expected this:

to print: 
`path/with/val`

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

marciocadev:
Closes #3027 

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify.rs
```diff
@@ -348,7 +348,7 @@ impl<'a> JSifier<'a> {
 							InterpolatedStringPart::Expr(e) => {
 								match *self.get_expr_type(e) {
 									Type::Json | Type::MutJson => {
-										format!("${{JSON.stringify({}, null, 2)}}", self.jsify_expression(e, ctx))
+										format!("${{typeof {} === 'string' ? {} : JSON.stringify({}, null, 2)}}", self.jsify_expression(e, ctx), self.jsify_expression(e, ctx), self.jsify_expression(e, ctx))
```
eladb, line 351: Expressions run code. If the expression has side-effects, emitting it multiple times can be an issue. For example, if the expression is `doSomeCrazyExpensiveCalculation()` or `send1gOfDataToTheNetwork()`, we can't afford to call it multiple times.

An alternative approach would be to evaluate it only ones and then use it in the ternary.

Something like this:

```
((e) => typeof e === 'string' ? e : JSON.stringify(e, null, 2))({expr})
```

eladb (changes requested):

eladb (approved):

Issue #1485, opened 2023-02-08
Import and use `App` constructs from CDK libraries


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

An `App` is typically a construct at the root of a construct tree. Unlike other constructs, most `App` classes do not expect a `scope` and `id` as their first arguments so they would need to be code-generated differently in the compiler.

### Use Cases

As a user, I would like to use Wing for authoring CDK libraries and apps with other CDK frameworks like cdk8s

### Implementation Notes

Today these classes cannot be imported because our compiler hardcodes an assumption that every construct has an implicit scope and id when emitting JavaScript code.

### Component

Language Design, Compiler

monadabot:
Congrats! :rocket: This was released in [Wing 0.22.23](https://github.com/winglang/wing/releases/tag/v0.22.23).

Issue #2934, opened 2023-06-15
Compiler panics when importing a jsii lib which has non construct types


staycoolcall911:
### I tried this:

In wing versions after this PR (Wing 0.21.6 and above): 
vscode: Missing completions in lazy-loaded jsii namespaces (https://github.com/winglang/wing/pull/2921)
`wing compile` the following:
```ts
bring "cdktf" as cdktf;
```

### This happened:

Compiler error:
```sh
Compiler bug (panicked at 'assertion failed: param.name == \"id\"', libs/wingc/src/type_check/jsii_importer.rs:676:29) during type-checking, please report at https://docs.winglang.io/contributors/bugs
```

### I expected this:

To compile with no errors.

### Component

Compiler

### Wing Version

0.21.6

### Node.js Version

19.3.0

### Platform(s)

MacOS

### Anything else?

#2921 is adding eager import of jsii modules we `bring`. This exposed a bug in our [jsii_importer.rs](https://github.com/winglang/wing/blob/2b0c77f36140f255b2bee1845dbadf7092f60093/libs/wingc/src/type_check/jsii_importer.rs#L676) - we assume any jsii type should have both `id` and `scope`, but we should limit that to constructs only.
What happens here is that CDKTF has some type that is probably not a construct, and therefore don't have an `id` and that causes the referenced code to fail.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

yoav-steinberg:
Currently we have a limitation (by design) where preflight classes must be a construct. This isn't a JSII limitation, so when we encounter such classes I see no other option than to skip them and not import them.

monadabot:
Congrats! :rocket: This was released in [Wing 0.22.23](https://github.com/winglang/wing/releases/tag/v0.22.23).

PR #3057, opened 2023-06-25
fix(compiler): panic when trying to import JSII constructs with no scope/id args

yoav-steinberg:
See #2934 
Importing a construct from JSII that doesn't have a `scope`/`id` arg pair as the first constructor arguments panicked the compiler. We now detect such (unconventional) constructs and mark them as so. This makes them non compatible with our `as`/`in` keywords for explicit scope/id control, but they can still be used by passing the correct (non standard) arguments to them.
This resolves importing the cdktf library because its Backend constructs are an example of such non-standard constructs. 

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -216,6 +216,12 @@ pub struct Class {
 	pub type_parameters: Option<Vec<TypeRef>>,
 	pub phase: Phase,
 	pub docs: Docs,
+
+	// Preflight classes are CDK Constructs which means they have a scode and id as their first arguments
```
staycoolcall911, line 220:
```suggestion
	// Preflight classes are CDK Constructs which means they have a scope and id as their first arguments
```

libs/wingc/src/type_check.rs
```diff
@@ -216,6 +216,12 @@ pub struct Class {
...
+	// this is natively supported by wing used the `as` `in` keywords. However theoretically it is possible
```
staycoolcall911, line 221:
```suggestion
	// this is natively supported by wing using the `as` `in` keywords. However theoretically it is possible
```

libs/wingc/src/type_check.rs
```diff
@@ -216,6 +216,12 @@ pub struct Class {
...
+	// to have a construct which does not have these arguments, in which case we can't use the `as` `in` keywords
+	// and instead the use will need to pass the relevant args to the class's init method.
```
staycoolcall911, line 223:
```suggestion
	// and instead the user will need to pass the relevant args to the class's init method.
```

examples/tests/valid/bring_cdktf.w
```diff
@@ -7,3 +8,11 @@ new aws.s3Bucket.S3Bucket(
...
   },
 ) as "Bucket";
+
+class Foo {
+  init() {
+    // Test importing a non-standard Construct that doesn't have a `scope`/`id` property pair as its first two arguments
+    // Some of cdktf's backend constructs are like this.
+    new cdktf.S3Backend(this, cdktf.S3BackendConfig {bucket: "foo", key: "bar"});
+  }
+}
```
staycoolcall911, line 18: Is it possible to override the `scope` and `id` of a preflight class in its `init()` function?
If the answer is yes, can you please add a test for that?

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -664,23 +665,39 @@ impl<'a> JsiiImporter<'a> {
...
 		// Add constructor to the class environment
 		let jsii_initializer = jsii_class.initializer.as_ref();
-
 		if let Some(initializer) = jsii_initializer {
 			let mut fn_params = vec![];
 			if let Some(params) = &initializer.parameters {
-				for (i, param) in params.iter().enumerate() {
-					// If this is a resource then skip scope and id arguments
-					// TODO hack - skip this check if the resource's name is "App"
-					// https://github.com/winglang/wing/issues/1485
```
staycoolcall911, line 674: Does this PR add support for #1485? If not, can you add a note on #1485 saying what's still missing?

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -664,23 +665,39 @@ impl<'a> JsiiImporter<'a> {
...
-					if class_phase == Phase::Preflight && type_name != "App" {
-						if i == 0 {
-							assert!(param.name == "scope");
-							continue;
-						} else if i == 1 {
-							assert!(param.name == "id");
-							continue;
+				let mut params_iter = params.iter();
+
+				// If this is a preflight class then we need to verify its first two scope and id args:
```
staycoolcall911, line 673:
```suggestion
				// If this is a preflight class then we need to verify its first two args (scope and id):
```

staycoolcall911 (approved):
Nice!

Chriscbr:
@yoav-steinberg I think this might also resolve https://github.com/winglang/wing/issues/1485, not totally sure

Chriscbr (approved):

examples/tests/valid/bring_cdktf.w
yoav-steinberg: It's not possible in Wing. Unconventional classes like `cdktf.S3Backend` do this, but I don't think we want to support it in wing.

libs/wingc/src/type_check/jsii_importer.rs
yoav-steinberg: Tested and it does seem to resolve #1485, good catch.

examples/tests/valid/bring_cdktf.w
staycoolcall911: Ok, thanks. Asked because the spec mentioned it is supported, but I figured out it wasn't. I removed this section from the spec

PR #3058, opened 2023-06-26
feat(cli): adding pre-release + analytics disclaimer on first run of cli

### LINKED ISSUE #2974: Print pre-release +analytics disclaimer on first run of Wing CLI
### Feature Spec

In order to set expectations with users and collaborate with them on making Wing amazing, the Wing CLI will emit a friendly and colorful pre-release disclaimer on first run:

```
🧪 This is an early pre-release of the Wing Programming Language.

We are working hard to make this a great tool, but there's still a pretty good 
chance you'll encounter missing pieces, rough edges, performance issues and even,
god forbid, bugs 🐞. 

Please don't hesitate to ping us at https://t.winglang.io/slack or file an issue at 
https://github.com/winglang/wing. We promise to do our best to respond quickly and help out.

To help us identify issues early, we are collecting anonymous analytics. 
To turn this off, set WING_CLI_DISABLE_ANALYTICS=1.
For more information see https://winglang.io/docs/analytics

(This message will self-destruct after the first run)
```

(feel free to use festive colors)

### Use Cases

* Set expectations with users about quality and readiness
* Encourage users to report issues and collaborate with us
* Inform users about analytics collection

### Implementation Notes

_No response_

### Component

CLI

### Community Notes

garysassano:
Closes #2974 


apps/wing/src/cli.ts
```diff
@@ -24,6 +29,41 @@ function actionErrorHandler(fn: (...args: any[]) => Promise<any>) {
...
+          "https://github.com/winglang/wing"
+        )}. We promise to do our best to respond quickly and help out.\n` +
+        `\n` +
+        `To help us identify issues early, we are collecting anonymous analytics.\n` +
+        `To turn this off, set ${chalk.yellowBright.bold("WING_CLI_DISABLE_ANALYTICS=1")}.\n` +
+        `For more information see ${chalk.blueBright.bold.underline(
+          "https://winglang.io/docs/analytics"
+        )}\n` +
+        `\n` +
+        `${chalk.redBright("(This message will self-destruct after the first run)")}`
```
tsuf239, line 59: Can you please set those as an array of strings, and join them with "\n"? 
(it will make it easier to read)
another option is to write the whole paragraph inside of \` \` as it will allow you to write the entire section without extra characters

tsuf239:
Also, the build isn't passing because of the snapshots probably- as it is always the first run for github actions machines.
We might need to find a way to bypass it 🤔 can we maybe configure them to have this file on setup?

apps/wing/src/cli.ts
garysassano: @tsuf239 Actually I had already considered everything you said. I looked at all the possible ways to achieve multiline string and several people deemed the string array to be "overkill", even tho I actually liked that solution the most. Then I used the template literal but the result ended up being pretty hard to read so I rewrote everything using strings chained with plus operator.

Thanks for the feedback, I'll change the code to use a string array since I liked that solution more anyway.

garysassano: @tsuf239 Actually I had already considered everything you said. I looked at all the possible ways to achieve multiline string and several people deemed the string array to be "overkill", even tho I actually liked that solution the most. Then I used the template literal but the result ended up being pretty messy (had to create a regex to remove trailing spaces since all lines were indented and account also for the blank lines) so I rewrote everything using strings chained with plus operator.

Thanks for the feedback, I'll change the code to use a string array since I liked that solution more anyway.

MarkMcCulloh: > Then I used the template literal but the result ended up being pretty messy (had to create a regex to remove trailing spaces since all lines were indented and account also for the blank lines)

My favorite way is to use a template literal but put the text all the way to the left. It doesn't have an extra runtime cost like the other options it is the easiest to read IMO:

```ts
// The initial backslash escapes the first \n
const text = `\ 
We are working hard to make this a great tool, but there's still a pretty good
chance you'll encounter missing pieces, rough edges, performance issues and even,
god forbid, bugs 🐞.`;
```

Does it look ugly because it's all the way to the left? I guess so, but less ugly than other options to me. If it's really too ugly we could indent it and run code after to dedent it, but that seems like a waste of the user's cpu.

garysassano: > My favorite way is to use a template literal but put the text all the way to the left.

It's funny because I did also that at some point but then thought "better change it, I doubt it meets their coding style". 😂

I like your pragmatic performance-oriented reasoning btw, I also try to think about saving CPU clocks in anything I do.

hasanaburayyan:
Closing this as https://github.com/winglang/wing/pull/3185 is absorbing the work

PR #3064, opened 2023-06-26
feat: capture reassignable variable and mutable instances

ekeren:
## Description

Prior to this change you where not able to capture any reassignable and mutable instances, this limitation was too restricting

Consider the following example where I wanted to have an `onUpdate` `preflight` api that sets a member  `onUpdateCallback`. The `onUpdateCallback` should be called in `inflight`. 

The only way to achieve this is by setting the `onUpdateCallback` on `init`.
After this change I should be able to make `onUpdateCallback` reassignable  

```ts (wing)

bring cloud;
class KeyValueStore {
  bucket: cloud.Bucket;
  onUpdateCallback: inflight (str): void;
  init(store: cloud.Bucket) {
    this.bucket = store;
  }

  onUpdate(fn: inflight (str):void){
    this.onUpdateCallback = fn;
  }

  inflight get(key: str): Json {
    return this.bucket.getJson(key);
  }
  inflight set(key: str, value: Json): void {
    this.onUpdateCallback(key);
    this.bucket.putJson(key, value);
  }
}
``` 

## Checklist

- [ ] Tests added (always)


eladb (approved):

examples/tests/valid/capture_reassigable_class_field.w
Chriscbr, line 1: @ekeren Have you tested this application on the cloud? I am slightly concerned the permissions may not work because the Wing simulator does not simulate permissions right now -- if I understand our language's permission system correctly, this is (one of) the main reasons the restriction is in place.

ekeren: No, I haven't 

Checking...

ekeren: @Chriscbr I had to modify the test a bit so it will run on the cloud, but it works 

(I'll push once my snapshots are updated)

ekeren: @Chriscbr , why did you think it would not work? should we still be worried?

Chriscbr: @ekeren interesting - ok, that's great news. (Our permission system is more powerful than I realized 😁)

> should we still be worried?

Maybe, there's a few other cases I'm worried about. For example, even though we are letting users *capture* a reassignable variable or a mutable object, I _think_ reassigning or mutating them while inflight should not work:

```
let var x = 5;
let y = MutArray<num>[1, 2, 3];

test "inflight" {
  x = 10; // error: preflight variable cannot be reassigned in inflight
  y.push(4); // error: mutable collection defined in preflight cannot be mutated in inflight
}
```

If we don't give this error, my concern is many users will think that we are providing them a distributed array or distributed counter etc. (that can be updated from anywhere), which is not the case.

Chriscbr: If the rule I'm suggesting is valid, then we may also need to add restrictions for these special types when users try to share them across function boundaries:

```js
let y = MutArray<num>[1, 2, 3];

let doStuff = inflight (arr: MutArray<num>) => {
  arr.push(4);
};

test "inflight" {
  doStuff(y); // error: mutable collection defined in preflight cannot be moved, since it could be mutated
}
```

ekeren: #3069 was created as a leftover. It was too hard for me to try implementing

I think that MutArray should be lifted into an Array so any reference in inflight should only "see" the immutable API, similar to Bucket and BucketClient

Chriscbr: ah I didn't see, cool! Thanks for explaining

PR #3087, opened 2023-06-26
chore: ensure bundled deps are actually fully bundled on mac

MarkMcCulloh:
So you may be looking at the diff and saying out loud "get this disgusting change out of my face" but hold on, I have a reason.

`@aws-sdk/util-utf8-node` and `@aws-sdk/util-utf8-browser` are bundled dependencies of the SDK. Both of them have a typo in the package.json `files` field. `"dist-*"` is supposed to designate all of the `dist-*` folders **and** their content. On linux, this works fine. On mac, it does not include these files. This matters when you try to run the equivalent of the ci on mac (basically `npm pack`) because these packages are missing those files. `"dist-*/**"` works for both OSes, and it's what other `@aws-sdk` do as well.

The reason there are 4 patches is because `@aws-sdk/util-utf8-node` also bundles 2 deps of its own with the same problem lol.

This change does not affect the published result since we use linux for that.

## Checklist

- [ ] Tests added (always)


Chriscbr (approved):
Cool. Maybe worth leaving an issue on the AWS SDK github project in case anyone else runs into this / to see if they can upstream this change

PR #3089, opened 2023-06-26
fix: plugin module resolution works unexpectedly

MarkMcCulloh:
In theory, plugin module resolution should work based on where the plugin file is located. Previously it was pulling dependencies from (I think?) the CLI's node_modules. I'm not actually totally sure, but I know it wasn't resolving them based on the actual plugin file location.

This change creates a `require` to invoke the plugin, very similar to the way we do it for preflight execution. This change on it's own broke the plugin e2e tests (because that folder did not have a node_modules with `@cdktf/provider-aws` in it), so I also moved them to `examples/tests/valid/plugins`, that way they can use the node_modules from the valid tests.

## Checklist

- [ ] Tests added (always)


hasanaburayyan (approved):
Hell yea!

PR #3101, opened 2023-06-27
fix: static completions for primitive types not shown in some cases

### LINKED ISSUE #2985: str and Json (possibly more types) don't have intellisense in some cases
### I tried this:

Wrote this code:

```ts
bring cloud;

let helloJson = MutJson { "hello": "world" };
str.
```

and this code:
```ts
bring cloud;

let helloJson = MutJson { "hello": "world" };
Json.
```

### This happened:

No intellisense in both cases

### I expected this:

get intellisense for the types

### Is there a workaround?

_No response_

### Component

IDE Extension

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

When I removed the instantiation of the MutJson from the third line I did get intellisense for the Json but not the str.

So this worked:

So this code works:
```ts
bring cloud;

Json.
```

This still didn't:
```ts
bring cloud;

str.
```

### Community Notes

MarkMcCulloh:
Fixes #2985

This PR handles 2 things:
- Completions for static types that followed an expression got the completions for the expression instead
- Completions were missing for the std types we still rename in the SDK (num, str, etc.)

## Checklist

- [x] Tests added (always)


libs/wingc/src/lsp/completions.rs
```diff
@@ -91,18 +91,25 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse
...
+					let parent_start = parent.start_position();
+					let parent_start_pos = lsp_types::Position {
+						line: parent_start.row as u32,
+						character: parent_start.column as u32,
+					};
+					if nearest_expr.span.contains(&parent_start_pos) {
+						let nearest_expr_type = types.get_expr_type(nearest_expr).unwrap();
+
+						// If we are inside an incomplete reference, there is possibly a type error or an anything which has no completions
+						if !nearest_expr_type.is_anything() && !nearest_expr_type.is_unresolved() {
```
yoav-steinberg, line 103: I have a feeling that in case this is an `any` we probably don't want to continue here (calling `get_completions_from_type` with its empty result is good enough). In case of `unresolved` then we probably do want to continue in this function.

yoav-steinberg (approved):

libs/wingc/src/lsp/completions.rs
MarkMcCulloh: Yeah, makes sense. Changed it to only check for `unresolved`, since get_completions_from_type will return an empty list for `any` anyways.

PR #3114, opened 2023-06-27
feat(sdk): adding math module

marciocadev:
I had a little trouble with the `min` and `max` because JSii didn't accept the methods to receive `Set<number>`.
When I tried to use `number[]` and use it in the Wing code, it only accepted arrays.
I ended up using `any`, but I wasn't very satisfied.

@eladb, I took the liberty of adding a few more functions to the module, such as `round` and I noticed that you placed `random` and `randomInt` in the `Util` module, but I wonder if they should be part of the `Math` module instead.

```ts
preflight inflight min(arr: Set<num>): num;
preflight inflight max(arr: Set<num>): num;
preflight inflight abs(value: num): num;
preflight inflight floor(value: num): num;
preflight inflight ceil(value: num): num;

// adding to the specs
preflight inflight math.PI; // PI constant
preflight inflight math.E; // Euler's number
preflight inflight isPrimen: num): bool;
preflight inflight round(value: num, decimalPlaces?: num): num;
preflight inflight median(arr: Array<num>): num;
preflight inflight mode(arr: Array<num>): Array<num>;
preflight inflight arithmeticMean(arr: Array<num>): num;
preflight inflight geometricMean(arr: Array<num>): num;
preflight inflight harmonicMean(arr: Array<num>): num;
preflight inflight fibonacci(n: num): num;
preflight inflight factorial(n: num): num;
preflight inflight combinations(n: num, r: num): num;
```

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/math/math.ts
```diff
@@ -0,0 +1,145 @@
+import { Code, InflightClient } from "../core";
+
+/**
+ * Utility class for mathematical operations.
+ */
+export class Util {
+  /**
+   * The mathematical constant representing the ratio of a circle's circumference to its diameter.
+   */
+  public static readonly PI = 3.141592653589793;
```
eladb, line 10: Life advice: whenever possible, don't calculate PI on your own:

```suggestion
  public static readonly PI = Math.PI;
```

libs/wingsdk/src/math/math.ts
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Euler's number, a mathematical constant approximately equal to 2.71828.
+   */
+  public static readonly E = 2.718281828459045;
```
eladb, line 14: While we are at it:

libs/wingsdk/src/math/math.ts
```diff
@@ -0,0 +1,145 @@
...
+   */
+  public static ceil(value: number): number {
+    return Math.ceil(value);
+  }
+
+  /**
+   * Rounds the given number to the nearest integer.
+   * @param value - The number to be rounded.
+   */
+  public static round(value: number, decimalPlaces?: number): number {
```
eladb, line 112: Use options in these cases (here and everywhere):

```suggestion
  interface RoundingOptions {
    readonly decimalPlaces?: number;
  }

  public static round(value: number, options?: RoundingOptions): number {
```

This results in a much nicer Wing usage:

```js
math.round(value, decimalPlaces: 1);
```

libs/wingsdk/src/math/math.ts
```diff
@@ -0,0 +1,145 @@
...
+      return Math.round(value * multiplier) / multiplier;
+    }
+    return Math.round(value);
+  }
+
+  /**
+   * Returns the minimum value from an set of numbers.
+   * @param arr The set of numbers.
+   */
+  public static min(arr: any): number {
```
eladb, line 124: Let's always use an array of `number` and make sure we have an easy way to pass in a set (something like `mySet.toArray()`)

```suggestion
  public static min(arr: number[]): number {
```

In the future we would want something like `IIterable<num>`.

libs/wingsdk/src/math/math.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (arr instanceof Set) {
+      return Math.min(...arr);
+    }
+
+    throw new Error("The argument is not a Set");
```
eladb, line 129: This should also work for arrays

eladb (changes requested):

eladb (approved):
Great job!

PR #3125, opened 2023-06-28
feat(compiler): preflight expression lifting

### LINKED ISSUE #1878: Unable to capture non-Wing CDK constructs
### Community Note

### I tried this:

```js
bring cloud;
bring "@cdktf/provider-aws" as tfaws;

class Table {
  table: tfaws.dynamodbTable.DynamodbTable;

  init(table_name: str, hash_key: str) {
    this.table = new tfaws.dynamodbTable.DynamodbTable(
      name: table_name,
      billing_mode: "PAY_PER_REQUEST",
      hash_key: hash_key,
      attribute: [{ name: hash_key, type: "S" }]
    );
  }

  inflight print_name() { 
    log(this.table.name);
  }
}

let table = new Table("my-first-table", "id");

  table.print_name();
});
```

### I expected this:

To print the table name in inflight

### Instead, this happened:

```
preflight error: unable to serialize immutable data object of type DynamodbTable

  note: intermediate javascript code (/Users/eladb/playground/dynamodb/target/main.tfaws/.wing/preflight.js):

     	_toInflight() {
>>   	const table_client = this._lift(this.table);
     	const self_client_path = "./clients/Table.inflight.js".replace(/\\/g, "/");
     	return $stdlib.core.NodeJsCode.fromInline(`(new (require("${self_client_path}")).Table({table: ${table_client}}))`);
--------------------------------- STACK TRACE ---------------------------------
/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/core/internal.js:89
    throw new Error(`unable to serialize immutable data object of type ${obj.constructor?.name}`);
    ^

Error: unable to serialize immutable data object of type DynamodbTable
    at serializeImmutableData (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/core/internal.js:89:11)
    at Table._lift (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/core/resource.js:250:54)
    at Table._toInflight (evalmachine.<anonymous>:36:30)
    at serializeImmutableData (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/core/internal.js:76:28)
    at makeHandler (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/core/internal.js:10:22)
    at new Inflight (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/core/inflight.js:74:43)
    at new MyApp (evalmachine.<anonymous>:44:73)
    at evalmachine.<anonymous>:55:1
    at Script.runInContext (node:vm:141:12)
    at Object.runInContext (node:vm:297:6)
    at compile (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/dist/commands/compile.js:172:12)
```

### Component:

SDK

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

We currently only support Wing resources in the capture mechanism. We need to support arbitrary constructs (`Construct.isConstruct(x)`). In those cases, there is no `_toInflight()` method of course.

Under `internal.js`:

```
// if the object is a resource (i.e. has a "_toInflight" method"), we use it to serialize
// itself.
if (typeof obj._toInflight === "function") {
    return obj._toInflight().text;
}
```

### LINKED ISSUE #2916: Wing test returns 0 exit code
### Feature Spec

Wing test command should have appropriate exit codes based on the outcome of the command. Test failures have a non zero exit code, as well as a compilation failure. Right now everything exits with a 0 exit code.

### Use Cases

CI, developer experience

### Implementation Notes

_No response_

### Component

CLI

### Community Notes

eladb:
## TODO

- [ ] Handle base classes lifts
- [ ] Verify CDK lift
- [ ] Handle unknown operations

## Description

The lifting system (previously known as the `FindReferenceVisitor`) was not robust enough because it tried to identify lifting by looking for references (e.g. `bucket.put()`). This resulted in a million edge cases (e.g. what about `foo.bar.baz.foo().put()`?). I realized that we basically worked at the wrong abstraction level.

So this change refactors the lifting system to work on *expressions* rather than on *references*. During type checking, we used to calculate the type for each expression. Now we are also calculating the *phase* of each expression. This is the key to lifting.

Now, during jsification, when we visit an expression we ask whether its phase it aligned with the scope's phase. If it's a preflight expression inside an inflight phase, it means we need to lift it.

This means, for example, that now we can lift complete expressions. For example, the expression `foo.bar.myArray.get(0)` will be evaluated in *preflight* and only the resulting value will be lifted into inflight (this is because `get()` is a phase-independent method and inherits its phase from the object). In the expression `foo.bar.myBucket.list()` only `foo.bar.myBucket` is a preflight value and will be lifted into inflight as a single object (`$foo_bar_myBucket`).

## Implementation Notes

To support this, I added the lift results into the `JSifyContext` struct as well as the current method name. These are required in `jsify_expression` to record the lift. Additionally, in order to identify the lift operations, I am propagating the `property` of object references into the `jsify_expression(obj)`.

The `mangling` module is responsible to take a multi-component expression (e.g. `a.b.c`, turn it into `a_b_c` and then deconstruct it back to `a.b.c` by emitting a JavaScript object with the same structure. This is primarily used to capture nested type named (e.g. `std.Json`).

## Testing

Added a bunch of Rust unit tests for mangling and for capturing/lifting. These are implemented as snapshot tests which capture some internal capturing details as well as the emitted javascript output. Ideally we should be able to run all of these tests as full tests, but since there are dozens of new tests, we decided to first implement more parallelism into our build so that this won't add too much time to our CI builds. Not ideal. This required extracting a `partial_compile`

## Resolves

* Fixes #1878
* Fixes #2916

## Misc

* Change the `Debug Wing Compiler` launch configuration to simply debug the current file. This allows basically just pressing F5 inside any `.w` file to kick off a debugging session. Sweet!
* Moved the capturing logic from `jsify.rs` into `captures.rs`.
* Change the `variable` portion of `StmtKind::Assignment` to an `Expr` instead of a `Reference` to make sure it's type and phase is captured during type checking.
* Add `()` to jsified closures and improve output of functions.
* Do not emit `$inflight_init` if it doesn't have any statements.
* Do not allow reassigning to preflight variables from inflight
* In the jsii importer, mark structs (and struct fields) as phase independent (because they are).

## Checklist

- [x] Tests added (always)


Chriscbr:
> Moved the capturing logic from `jsify.rs` into `captures.rs`.

👏 👏

libs/wingc/src/jsify/mangling/tests.rs
```diff
@@ -0,0 +1,47 @@
...
+	]));
+	assert_snapshot!(result.to_string());
+}
+
+impl Display for MangleDemangle {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		writeln!(f, "keys: {}", self.keys.join(","))?;
+		writeln!(f, "\n{}", self.javascript)
+	}
+}
```
Chriscbr, line 47: If this only needs to be compiled for unit testing, you could consider slapping this label in front
```suggestion
#[cfg(test)]
impl Display for MangleDemangle {
	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
		writeln!(f, "keys: {}", self.keys.join(","))?;
		writeln!(f, "\n{}", self.javascript)
	}
}
```

libs/wingc/src/jsify/mangling/snapshots/wingc__jsify__mangling__tests__mixed.snap
```diff
@@ -0,0 +1,22 @@
...
+    e: a_b_e,
+  },
+};
+const b = {
+  myBucket: b_myBucket,
+};
+const std = {
+  Json: std_Json,
+  Number: std_Number,
+};
```
Chriscbr, line 20: Pretty neat algorithm 👍

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -139,17 +139,11 @@ Duration <DURATION>"
 `;
 
 exports[`capture_redefinition.w 1`] = `
-"error: Cannot capture symbol \\"y\\" because it is shadowed by another symbol with the same name
-  --> ../../../examples/tests/invalid/capture_redefinition.w:4:7
-  |
-4 |   log(y);
-  |       ^ Cannot capture symbol \\"y\\" because it is shadowed by another symbol with the same name
```
Chriscbr, line 146: Regression perhaps?

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -752,75 +746,97 @@ Duration <DURATION>"
...
-   |
-47 |     let b = this.array.at(1);
-   |                  ^^^^^ Capturing collection of preflight classes is not supported yet (type is 'Array<Bucket>')
-
-
- 
- 
-Tests 1 failed (1)
+"fail ┌ inflight_ref_explicit_ops.wsim » root/env0/test:test1
+     │ Error: Missing environment variable: BUCKET_HANDLE_5982bab0
```
Chriscbr, line 750: Uh oh

examples/tests/invalid/lifting/inflight-from-preflight.w
Chriscbr, line 1: Can we add an E2E test for the fixed issue? (#1878)

Chriscbr (commented):
I only got through a few files so far before I needed to get some fresh air 😄 but I hope some of the initial comments are useful. Really exciting to see this PR converge to a working state

eladb:
Superseded by https://github.com/winglang/wing/pull/3213

PR #3127, opened 2023-06-28
chore(console): prepare and publish the design system

skyrpex:
- Publishes the `@wingconsole/design-system` package
- Exposes the tailwind plugins of both `@wingconsole/design-system` and `@wingconsole/ui` packages


tools/hangar/src/paths.ts
```diff
@@ -26,6 +26,9 @@ export const targetWingConsoleAppSpec =
 export const targetWingConsoleServerSpec =
   process.env.HANGAR_WINGCONSOLE_SERVER_SPEC ??
   `file:${path.join(repoRoot, `apps/wing-console/console/server`)}`;
+export const targetWingConsoleDesignSystemSpec =
+  process.env.HANGAR_WINGCONSOLE_DESIGN_SYSTEM_SPEC ??
+  `file:${path.join(repoRoot, `apps/wing-console/console/design-system`)}`;
```
MarkMcCulloh, line 31: It seems like "@wingconsole/design-system` is only consumed as a devDependency of @wingconsole/ui. Since it's not a runtime dependency, I don't think hangar needs to be concerned about installing it since it won't attempt to actually resolve it. WDYT?

skyrpex: Good catch! It should actually depend on it as a dependency (not devDependency).

ainvoner (approved):

PR #3135, opened 2023-06-28
chore: self mutation does not trigger build

MarkMcCulloh:
Other than security, I now remember the other reason we use a MUTATION_TOKEN rather than GITHUB_TOKEN: Github will not trigger push/pull_request events if a commit is pushed by github actions. We should try to find a way around this, but in the mean time let's go back to using a token

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -14,7 +14,8 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
-  contents: write
+  contents: read
+  packages: read
   statuses: write
 
 jobs:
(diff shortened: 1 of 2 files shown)


Chriscbr:
> Github will not trigger push/pull_request events if a commit is pushed by github actions. We should try to find a way around this, but in the mean time let's go back to using a token

Yeah I think this is a core restriction to prevent automated workflows (that aren't associated with a GitHub user or GitHub app) calling each other recursively. I think using a token or authorizing via an app are the main workarounds. Another option theoretically could be to switch from a "choreography" approach (workflows calling each other) to an "orchestration" approach (centralized service deciding when to start workflows) but I'm not sure if anyone does that in practice 🤔

Chriscbr (approved):

Issue #3133, opened 2023-06-28
ERROR: Must call super constructor in derived class before accessing 'this' or returning from derived constructor


ekeren:
### I tried this:

```ts 
class Bar {

}


class Foo extends Bar  {
  init() {
    super();
  }

  inflight doStuff(h: num) {
    // all code is async and runs on the cloud
  }
}

new Foo();
```

### This happened:

```sh
ERROR: Must call super constructor in derived class before accessing 'this' or returning from derived constructor

target/test/a.wsim.743287.tmp/.wing/preflight.js:38
       class Foo extends Bar {
         constructor(scope, id, ) {
>>         this._addInflightOps("doStuff");
           super(scope,id,);
         }
```

### I expected this:

to work

### Is there a workaround?

remove the inflight api and it will work

### Component

Compiler

### Wing Version

0.22.27

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.22.44](https://github.com/winglang/wing/releases/tag/v0.22.44).

Issue #3104, opened 2023-06-27
`Must call super constructor in derived class before accessing 'this'` with default constructor


eladb:
### I tried this:

```js
class Animal { }
class Dog extends Animal { }
new Dog();
```

### This happened:

```
ERROR: Must call super constructor in derived class before accessing 'this' or returning from derived constructor

target/test.wsim.387169.tmp/.wing/preflight.js:37
       }
       class Dog extends Base {
>>       constructor(scope, id, ) {
         }
         static _toInflightType(context) {
```

### I expected this:

To work

### Is there a workaround?

Define constructors

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.22.44](https://github.com/winglang/wing/releases/tag/v0.22.44).

PR #3136, opened 2023-06-28
fix(compiler): derived classes implicitly call super()

hasanaburayyan:
Okay so ALOT of snapshots got updated (because of ordering of jsify changes) but this change really only contains two main differences:

1. It ensures that the `init` statements are the first thing we jsify for a class constructor. This will make sure that calls to `super()` occur before any of the extra JavaScript code we add is jsified. Otherwise we can run into an issue where we call `this` in a derived class before we call `super()`
2. Since init blocks are optional, we implicitly jsify calls to `super()` in derived classes if they do not implement the call themselves within an `init()`

Closes: https://github.com/winglang/wing/issues/3104
Closes: https://github.com/winglang/wing/issues/3133

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify.rs
```diff
@@ -1060,22 +1060,36 @@ impl<'a> JSifier<'a> {
...
+					match s.kind {
+						StmtKind::SuperConstructor { arg_list: _ } => {} // Already has super do nothing
+						_ => code.line("super(scope, id);"),
+					}
+				}
+				None => {
+					// no statements so we implicitly call super since we know there is a parent
+					code.line("super(scope, id);")
+				}
+			}
```
hasanaburayyan, line 1087: I feel like there should be a way to do this without so many branch arms. So feel free to suggest, I couldnt think of something else

examples/tests/valid/class.w
```diff
@@ -119,4 +119,20 @@ inflight class B extends A {
...
+  }
+
+  inflight doStuff(h: num) {}
+}
+
+new Foo();
+
+// derived classes without defined constructors
+class Baz extends Bar {}
+new Baz();
```
eladb, line 138: Let's add this case as well:

```js
class Boom() {
  init() { }
}
class Bam extends Boom {}
```

libs/wingc/src/jsify.rs
eladb, line 1087: Here's attempt:

```rust
let super_call = if let Some(s) = init_statements.statements.first() {
	matches!(s.kind, StmtKind::SuperConstructor { .. })
} else {
	false
};

// no statements so we implicitly call super since we know there is a parent
if !super_call {
	code.line("super(scope, id);")
}
```

eladb (approved):

libs/wingc/src/jsify.rs
hasanaburayyan: Gorgeous!

Issue #3149, opened 2023-06-29
console analytics uuid might be is lost with electron removal


ekeren:

monadabot:
Congrats! :rocket: This was released in [Wing 0.22.39](https://github.com/winglang/wing/releases/tag/v0.22.39).

PR #3154, opened 2023-06-29
chore(console): enable segment analytics for `@wingconsole/app`

skyrpex:
Uses anonymous data to track the usage of the app.
resolves: https://github.com/winglang/wing/issues/3149


skyrpex:
Before merging, the `SEGMENT_WRITE_KEY` secret needs to be added to the repository.

ainvoner (approved):

PR #3158, opened 2023-06-29
fix(console): open external links in the browser directly

### LINKED ISSUE #3066: The Redis resource's `help` command doesn't open the link for you
When typing the `help` command in a Redis resource, it should open the URL for you (but it doesn't).

So, either display a clickable link or open it automatically.

skyrpex:
Stops relying on the server to open external links.

Fixes #3066.


skyrpex:
This solution doesn't work when opening the console page using the Simple Browser from VSCode, though.

PR #3177, opened 2023-06-30
chore(compiler): type references as expressions and phase tracking

eladb:
No functional changes, preparatory pull request as part of the expression lifting escapade.

Type references are expressions that reference type names. It's valuable to model them as expressions because they are needed at runtime. Once they are modeled as expressions, we can treat all expressions equally when it comes to lifting (which is the motivation for this refactor).

We've changed the following places to use type references instead of a `UserDefinedType`. In all of these cases the types have a runtime presence. The rest of the places are just type annotations that are erased at runtime:

- The class being instantiated in `new` expressions
- Base classes
- The type component in a `Reference::TypeMember` (static access).

Furthermore, to support expression lifting, the type checker now records the phase of each expression during type checking and makes it available through `get_expr_phase`.

Additionally, changed the left-hand-side of an assignment statement to be an `Expr` and not a `Reference`. Also in order to be able to lift it.

Turn the variable component of `StmtKind::Assignment` from a simple `Symbol` to an `Expr` in order for it to also hold type and phase information.

The JSII importer accidentally imported structs as `preflight` while they should be phase-independent.

Misc: strip excess newlines when adding a line to `CodeMaker`.

## Checklist

- [x] Tests added (always)


tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -44,6 +44,20 @@ error: Unknown symbol \\"this\\"
   |     ^^^^ Unknown symbol \\"this\\"
 
 
+error: Unknown symbol \\"this\\"
```
staycoolcall911, line 47: What introduced this change in this snapshot?

staycoolcall911 (approved):
Thanks to the detailed PR description reviewing was a breeze...

tools/hangar/__snapshots__/invalid.ts.snap
eladb: Checking

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -280,11 +279,11 @@ error: Preflight field \\"y\\" is not initialized
    |   ^ Preflight field \\"y\\" is not initialized
 
 
-error: Expected \\"x\\" to be a type but it's a variable
+error: Base class \\"unresolved\\" is not a class
    --> ../../../examples/tests/invalid/class.w:68:18
    |
 68 | class C7 extends x {
-   |                  ^ Expected \\"x\\" to be a type but it's a variable
+   |                  ^ Base class \\"unresolved\\" is not a class
```
MarkMcCulloh, line 286: Regression

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -294,11 +293,11 @@ error: Base class \\"S1\\" is not a class
    |                  ^^ Base class \\"S1\\" is not a class
 
 
-error: Class cannot extend itself
+error: Base class \\"unresolved\\" is not a class
    --> ../../../examples/tests/invalid/class.w:78:19
    |
 78 | class C11 extends C11 {
-   |                   ^^^ Class cannot extend itself
+   |                   ^^^ Base class \\"unresolved\\" is not a class
```
MarkMcCulloh, line 300: Regression

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -336,6 +335,13 @@ error: Expected type to be \\"num\\", but got \\"str\\" instead
    |              ^^^^ Expected type to be \\"num\\", but got \\"str\\" instead
 
 
+error: Variable cannot be reassigned from inflight
+   --> ../../../examples/tests/invalid/class.w:61:5
+   |
+61 |     this.y = 1;
+   |     ^^^^^^^^^^^ Variable cannot be reassigned from inflight
```
MarkMcCulloh, line 342: This line already has a more specific error above:

```
"y" cannot be initialized in the inflight initializer
```

Technically y is not even reassignable at all, so `cannot be reassigned from inflight` is true but misleading.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1965,11 +1964,11 @@ Duration <DURATION>"
...
 
 exports[`void_in_expression_position.w 1`] = `
-"error: Property access unsupported on type \\"void\\"
-  --> ../../../examples/tests/invalid/void_in_expression_position.w:1:1
+"error: Property not found
+  --> ../../../examples/tests/invalid/void_in_expression_position.w:1:12
   |
 1 | log(\\"hey\\").get(\\"x\\");
-  | ^^^^^^^^^^ Property access unsupported on type \\"void\\"
+  |            ^^^ Property not found
```
MarkMcCulloh, line 1971: This diagnostic is accurate but misleading. The original diagnostic does need some work though so it's probably fine to try this new one out.

tools/hangar/__snapshots__/invalid.ts.snap
eladb: yes, I am on it!

tools/hangar/__snapshots__/invalid.ts.snap
eladb: Yes, that was my thought :-)

tools/hangar/__snapshots__/invalid.ts.snap
eladb: Good point. Looking into this.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1528,11 +1527,11 @@ exports[`sorted_errors_no_span.w 1`] = `
   |              ^^^ Expected type to be \\"num\\", but got \\"str\\" instead
 
 
-error: Expected \\"b\\" to be a type but it's a variable
+error: Base class \\"unresolved\\" is not a class
   --> ../../../examples/tests/invalid/sorted_errors_no_span.w:3:26
   |
 3 | inflight class c extends b {}
-  |                          ^ Expected \\"b\\" to be a type but it's a variable
+  |                          ^ Base class \\"unresolved\\" is not a class
```
eladb, line 1534: yak

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1864,32 +1863,32 @@ Duration <DURATION>"
 `;
 
 exports[`unknown_symbol.w 1`] = `
-"error: Unknown symbol \\"clod\\"
-  --> ../../../examples/tests/invalid/unknown_symbol.w:3:18
```
eladb, line 1868: regression

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1864,32 +1863,32 @@ Duration <DURATION>"
...
-error: Unknown symbol \\"y\\"
+"error: Unknown symbol \\"y\\"
   --> ../../../examples/tests/invalid/unknown_symbol.w:9:13
   |
 9 | let x = 2 + y;
   |             ^ Unknown symbol \\"y\\"
 
 
-error: Unknown symbol \\"B\\"
+error: Base class \\"unresolved\\" is not a class
```
eladb, line 1873: regression

tools/hangar/__snapshots__/invalid.ts.snap
eladb: This is actually the correct errors.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -280,11 +294,18 @@ error: Preflight field \\"y\\" is not initialized
    |   ^ Preflight field \\"y\\" is not initialized
 
 
-error: Expected \\"x\\" to be a type but it's a variable
+error: Unable to resolve type \\"x\\"
    --> ../../../examples/tests/invalid/class.w:68:18
    |
 68 | class C7 extends x {
-   |                  ^ Expected \\"x\\" to be a type but it's a variable
```
eladb, line 287: regression

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -280,11 +294,18 @@ error: Preflight field \\"y\\" is not initialized
...
+   |                  ^ Unable to resolve type \\"x\\"
+
+
+error: Base class \\"unresolved\\" is not a class
+   --> ../../../examples/tests/invalid/class.w:68:18
+   |
+68 | class C7 extends x {
+   |                  ^ Base class \\"unresolved\\" is not a class
```
eladb, line 308: regression

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -514,18 +549,11 @@ Duration <DURATION>"
 `;
 
 exports[`enums.w 1`] = `
-"error: Enum \\"SomeEnum\\" does not contain value \\"FOUR\\"
-  --> ../../../examples/tests/invalid/enums.w:5:21
-  |
-5 | let four = SomeEnum.FOUR;
-  |                     ^^^^ Enum \\"SomeEnum\\" does not contain value \\"FOUR\\"
```
eladb, line 521: Regression

libs/wingc/src/type_check.rs
```diff
@@ -2355,12 +2388,23 @@ impl<'a> TypeChecker<'a> {
...
-					self.spanned_error(stmt, format!("Variable {} is not reassignable ", variable));
+				let (exp_type, _) = self.type_check_exp(value, env);
+				let (var_type, var_phase) = self.type_check_exp(variable, env);
+
+				// check if the variable can be reassigned
+
+				if let ExprKind::Reference(r) = &variable.kind {
+					let (var, _) = self.resolve_reference(&r, env);
+					if var_phase == Phase::Preflight && env.phase == Phase::Inflight {
+						self.spanned_error(stmt, "Variable cannot be reassigned from inflight".to_string());
```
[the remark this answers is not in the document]
eladb: added

tools/hangar/__snapshots__/invalid.ts.snap
eladb: Fixed.

eladb:
@MarkMcCulloh I'll make sure all the diagnostics around base classes are fixed. If that's all, I'd appreciate an approval so I can be unblocked with my next step.

MarkMcCulloh (approved):
> I'll make sure all the diagnostics around base classes are fixed. If that's all, I'd appreciate an approval so I can be unblocked with my next step.

Yeah LGTM as long as the diags are fixed. Luckily the grammar still prevents real expressions into places where types are expected so we might be safe from some new big regressions.

I also think @yoav-steinberg should take a look at this, but I'll leave that to you if you want to wait.

libs/wingc/src/type_check.rs
```diff
@@ -74,21 +74,40 @@ pub enum SymbolKind {
 	Namespace(NamespaceRef),
 }
 
+#[derive(Debug, Clone)]
+pub enum VariableKind {
+	/// a free variable not associated with a specific type
+	Free,
```
Chriscbr, line 80: Is this meant to refer to any local variable? I think that might be better compiler nomenclature:
```suggestion
	/// a local variable not associated with a specific type
	Local,
```

libs/wingc/src/type_check.rs
```diff
@@ -1144,18 +1169,28 @@ impl Types {
 	}
 
 	/// Stores the type of a given expression node.
```
Chriscbr, line 1171:
```suggestion
	/// Stores the type and phase of a given expression node.
```

libs/wingc/src/type_check.rs
```diff
@@ -1244,61 +1288,61 @@ impl<'a> TypeChecker<'a> {
...
+		self.types.assign_type_to_expr(exp, t, phase);
+		(t, phase)
 	}
 
 	/// Helper function for type_check_exp. This is needed because we want to be able to `return`
 	/// and break early, while still setting the evaluated type on the expression.
 	///
 	/// Do not use this function directly, use `type_check_exp` instead.
-	fn type_check_exp_helper(&mut self, exp: &Expr, env: &SymbolEnv) -> TypeRef {
+	fn type_check_exp_helper(&mut self, exp: &Expr, env: &SymbolEnv) -> (TypeRef, Phase) {
```
Chriscbr, line 1313: This is pretty cool!

libs/wingc/src/type_check.rs
```diff
@@ -1244,61 +1288,61 @@ impl<'a> TypeChecker<'a> {
...
 					s.parts.iter().for_each(|part| {
 						if let InterpolatedStringPart::Expr(interpolated_expr) = part {
-							let exp_type = self.type_check_exp(interpolated_expr, env);
+							let (exp_type, _) = self.type_check_exp(interpolated_expr, env);
 							self.validate_type_in(exp_type, &self.types.stringables(), interpolated_expr);
 						}
 					});
-					self.types.string()
+					// TODO: do we need to determine the phase based on the inner expressions?
+					(self.types.string(), Phase::Independent)
```
Chriscbr, line 1326: I think as long as strings are phase-independent, it seems okay to me to say that the phase of the template string is independent

libs/wingc/src/type_check.rs
```diff
@@ -1399,26 +1458,26 @@ impl<'a> TypeChecker<'a> {
 									class.phase, class.name, env.phase
 								),
 							);
-							return self.types.error();
+							return (self.types.error(), Phase::Independent);
 						}
 					}
 					// If type is anything we have to assume it's ok to initialize it
-					Type::Anything => return self.types.anything(),
+					Type::Anything => return (self.types.anything(), Phase::Independent),
```
Chriscbr, line 1465: I wonder if at some point we'll need a `Phase::Unknown` (for now I hope not)

libs/wingc/src/type_check.rs
```diff
@@ -3249,8 +3332,8 @@ impl<'a> TypeChecker<'a> {
 					type_: v,
 					reassignable,
 					phase: flight,
-					is_static,
-					is_member: _,
+					kind,
+					..
```
Chriscbr, line 3336: Can be removed
```suggestion
					kind,
```

examples/tests/invalid/inflight_reassign.w
```diff
@@ -0,0 +1,10 @@
+let var xvar = "hello";
+let ylet = 123;
+
+inflight () => {
+  xvar = "hi";
+//^^^^ Variable cannot be reassigned from inflight
```
Chriscbr, line 6: 👏

Chriscbr (approved):

libs/wingc/src/type_check.rs
eladb: It's meant to represent a value that's not bound to an object.

Not sure local is the right term.

We've used "free variables" previously

libs/wingc/src/type_check.rs
eladb: OK.

libs/wingc/src/type_check.rs
Chriscbr: Sorry, I probably didn't explain myself well. I think we're thinking of the term "bound" in different ways (I'm using it in the math/Lisp sense where it refers to a property about substituting variables into functions, and are using it in a way that seems more related to association or membership to types). Anyway I'd be down to chat about it sometime, feels like a shared understanding of compiler terminology is useful long term for taming the complexity monster we have here 😁

Issue #1964, opened 2023-04-02
Compiler reserved words


staycoolcall911:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

The Wing compiler now detects reserved words, such as `this`, `static`, `private`, etc. It is not possible to use these words as variable names, etc.

- [ ] Throw error when a reserved word is used out of place.
- [ ] The full list of reserved words should also appear on the winglang spec.

### Use Cases

Right now, this compiles:
```ts
let this = 1;
```
I'd expect an error message:
```ts
let this = 1;
//   ^^^Cannot use reserved word
```

### Component

Compiler

MarkMcCulloh:
Most of the work for this has been done in https://github.com/winglang/wing/pull/3179

To finish this issue, the reserved list needs to be finalized and documented in the spec.

staycoolcall911:
Just need to compile the list of reserved words and add it to the [language reference](https://www.winglang.io/docs/language-reference).

PR #3179, opened 2023-06-30
fix(compiler): reserved words are not blocked

### LINKED ISSUE #3110: Javascript reserved words (such as null) should be reserved words in Wing as well
### I tried this:

```
// main.w
bring cloud;
bring "cdktf" as cdktf;
bring "@cdktf/provider-null" as null;
```

```
wing compile main.w
```

similar for 

```
// main.w
bring cloud;
bring "cdktf" as cdktf;
bring "@cdktf/provider-null" as nullish;

let null = new nullish.provider.NullProvider();
```

### This happened:

```
 wing compile main.w
Error: ERROR: Unexpected token 'null'

target/main.wsim.016318.tmp/.wing/preflight.js:7
   const cloud = require('@winglang/sdk').cloud;
   const cdktf = require("cdktf");
>> const null = require("@cdktf/provider-null");
     constructor(scope, id) {

    at compile (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/commands/compile.js:117:19)
```

### I expected this:

- No error or
- Immediate feedback from the compiler / language server that this is not working

### Is there a workaround?

Don't use `null` as a variable

### Component

Compiler

### Wing Version

_No response_

### Wing Console Version

0.22.24

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
Fixes #3110

This is list is pretty naive, I think we should eventually do some mangling/cleanup of the JS we emit to avoid so many reserved values. Since I believe there's more work to do, I didn't touch the spec yet (See https://github.com/winglang/wing/issues/1964).

This change will at least prevent strange runtime errors due to using reserved words. Added a special case for "constructor" as a method name, since it's not technically reserved elsewhere.

Misc:
- Changed the add_error in the parser to have a version that doesn't return anything. Sorry that this caused a bit of a noisy diff but I was so tired of that dumb `let _ = ...` thing

## Checklist

- [x] Tests added (always)


libs/wingc/src/parser.rs
```diff
@@ -41,6 +41,99 @@ static UNIMPLEMENTED_GRAMMARS: phf::Map<&'static str, &'static str> = phf_map! {
...
+	"Object",
+	"Set",
+
+	// consts declared by the compiler in preflight
+	"$stdlib",
+	"$outdir",
+	"$wing_is_test",
+	"$App",
+	"$AppBase",
+	"$Root",
```
Chriscbr, line 134: I might just skip these since the list might get out of date more quickly (and if a user gets errors by declaring a variable like "$stdlib", it would probably be better to just track that as a compiler bug instead IMO, since there's no technical reason they have to be reserved). Up to you though

libs/wingc/src/parser.rs
```diff
@@ -41,6 +41,99 @@ static UNIMPLEMENTED_GRAMMARS: phf::Map<&'static str, &'static str> = phf_map! {
...
+	"try",
+	"typeof",
+	"var",
+	"void",
+	"volatile",
+	"while",
+	"with",
+	"yield",
+
+	// wing-specific reserved words
```
Chriscbr, line 111: Do we also need to reserve words like `str`, `num`, etc.? I don't know but figured I'd ask

Chriscbr (approved):

libs/wingc/src/parser.rs
MarkMcCulloh: Ah good catch. I initially left them out because I didn't think so but thinking again yeah they should be reserved.

libs/wingc/src/parser.rs
MarkMcCulloh: Makes sense, I guess I'll take them out since there's no chance for them to show up during inflight (which is where it's a much worse bug)

PR #3187, opened 2023-07-01
chore: switch to pnpm

MarkMcCulloh:
### Pros

- Much faster installation (On CI, ~2m30s => ~1m or ~30s if cached)
- Less disk space usage
- No custom installation logic (AKA the repo makes much more sense)
- No patch-package needed (pnpm has that support built-in)
- `pnpm install` works as expected no matter where you are in the repo

### Cons

- New tool that isn't included with node
- Custom packaging logic to handle bundled deps (See below)

### What's the hack

See [./tools/bump-pack/src/bundle-link.ts](https://github.com/winglang/wing/blob/mark/pnpm-testing/tools/bump-pack/src/bundle-link.ts)
In short: We need to add symlinks for all the transitive dependencies of bundled dependencies. This ensures the final package has all the deps it needs to run because pnpm is intentionally not friendly to bundled deps.

---


apps/wing/src/cli.ts
```diff
@@ -39,6 +39,7 @@ async function main() {
   });
 
   program.option("--progress", "Show compilation progress", () => {
+    console.log(program.processedArgs);
```
skyrpex, line 42: Debug remnants?

libs/tree-sitter-wing/package.json
```diff
@@ -21,16 +21,16 @@
...
-    "test:watch": "npx -y nodemon --watch grammar.js --watch src/scanner.c --exec \"npm run test\"",
+    "test:watch": "npx -y nodemon --watch grammar.js --watch src/scanner.c --exec \"pnpm run test\"",
     "build": "npx -y tree-sitter-cli generate",
-    "build:watch": "npx -y nodemon --watch grammar.js --watch src/scanner.c --exec \"npm run build\"",
-    "build:wasm": "npm run build && npx -y tree-sitter-cli@0.20.6 build-wasm --docker",
-    "playground": "npm run build:wasm && npx -y tree-sitter-cli@0.20.6 playground",
-    "playground:quiet": "npm run build:wasm && npx -y tree-sitter-cli@0.20.6 playground -q",
-    "playground:watch": "npx -y nodemon --watch grammar.js --watch src/scanner.c --exec \"npm run playground:quiet\""
+    "build:watch": "npx -y nodemon --watch grammar.js --watch src/scanner.c --exec \"pnpm run build\"",
+    "build:wasm": "pnpm run build && npx -y tree-sitter-cli@0.20.6 build-wasm --docker",
```
skyrpex, line 27: It should be possible to install tree-sitter-cli as a dev dependency now that we use pnpm? No need to avoid running npm scripts.

MarkMcCulloh: Yep, looks like it works now. Good catch!

.github/workflows/build.yml
```diff
@@ -244,26 +250,23 @@ jobs:
...
-
       - name: Install Dependencies
-        uses: bahmutov/npm-install@v1
-        with:
-          working-directory: |
-            tools/hangar
-            examples/tests/valid
-            examples/tests/invalid
-          install-command: npm ci
+        run: pnpm install --ignore-scripts --frozen-lockfile --filter hangar --filter examples-valid --filter examples-invalid
```
skyrpex, line 265: Is there still a reason to use `--ignore-scripts`?

apps/wing-console/console/server/package.json
```diff
@@ -31,10 +31,10 @@
     "@types/lodash.uniqby": "^4.7.7",
     "@types/ws": "^8.5.5",
     "@vitest/coverage-c8": "^0.31.4",
-    "@wingconsole/error-message": "file:../../packages/error-message",
-    "@wingconsole/eslint-plugin": "file:../../tools/eslint-plugin",
-    "@wingconsole/tsconfig": "file:../../tools/tsconfig",
-    "bump-pack": "file:../../../../tools/bump-pack",
+    "@wingconsole/error-message": "workspace:../../packages/error-message",
```
skyrpex, line 34: All of the workspace dependencies can be installed as if they were third party dependencies, ex: `pnpm i @wingconsole/error-message`. If they belong to the workspace, they'll be symlinked and appear as `"@wingconsole/error-message": "workspace:^"` in the package.json.

So, there's no need to specify the relative path to the package: we can just use `workspace:^`. Thing is, do we prefer to be explicit? I prefer not worrying about the relative paths... less clutter.

skyrpex:
Very nice job! Looks very approvable to me but I'm afraid of doing so because of mergify. Let's talk when everybody is online.

.github/workflows/build.yml
MarkMcCulloh: Yeah, to skip the scripts at the root of the repo (mostly the setup_wasi.sh thing)

apps/wing-console/console/server/package.json
MarkMcCulloh: I'm down to try it out!

skyrpex (approved):
nice job!

PR #3178, opened 2023-06-30
chore(console): add e2e tests

skyrpex:
- The console uses https://playwright.dev/ to run e2e tests
  - Uses `data-testid` parameters to locate elements
- Uses the [IBM Plex font](https://github.com/IBM/plex)
- The `wing it` command uses port 1214 by default, if available
- Adds an optional `--wingfile` parameter to the console dev and preview scripts

Resolves [#2929](https://github.com/winglang/wing/issues/2929).

PR #3218, opened 2023-07-03
chore: use `git diff --binary` so snapshot diffing works

skyrpex:


skyrpex:
Superseded by https://github.com/winglang/wing/pull/3178

PR #3245, opened 2023-07-04
chore: switch to pnpm

MarkMcCulloh:
Second attempt at https://github.com/winglang/wing/pull/3187

### What's new?

The big workaround before was to make sure that all bundled dependencies and their dependencies were symlinked to node_modules. This was great and made it work locally and via file:tgz installs (when all packages are installed directly). However, the final tarballs were not actually correct. 
For each bundled dep, if it had any bundled deps of its own, those were missing. npm normally includes these, but apparently if the bundled dep is a symlink then it doesn't do it 😢 .

So, the [main change](https://github.com/winglang/wing/blob/mark/pnpm-take-2/tools/bump-pack/src/pack.ts) here is that the packaging process creates a completely separate copy of the package. In that copy, we convert the necessary symlinks into copies. Then we package. These packages look almost identical to the ones we publish now.

I think the only way I could possibly test this with enough accuracy is to set up verdaccio, but I'm more confident that this iteration should work. Not sure I want to block this further by taking the time to go back to verdaccio.

Misc:
- Updated nx to latest with a patch to make sure our workflow of using `--` still works. I created a PR upstream in nx to maybe fix the issue. I did this update for two reasons:
  - We are very out of date on nx
  - Theres an old nx bug on linux causing it to watch/search node_modules. This hurts perf and caused nx to try to discover nx projects in node_modules. The update fixed this.


skyrpex (approved):

skyrpex:
Looks good! As you mentioned in our conversation, it'd be responsible to try the whole thing using verdaccio.

MarkMcCulloh:
Tested locally by running `npm install -g winglang` against a local registry and all seemed good. Was able to test, compile to tf-aws, and `wing it`.

PR #3247, opened 2023-07-04
feat(docs): removing `bring std` and hide irrelevant std classes

### LINKED ISSUE #3161: remove T1 section and turn other T1 into <T>

### LINKED ISSUE #3162: static method still shows std.
Initializers are hidden for types like MutArray and Array but the docs for static methods still shows std.MutArray etc. which is a bug

tsuf239:
## Description
fixes #3162 
fixes #3161

* removing `bring std;` and `std.XXX` from std documentation code snippets
* Hiding `mutJson` and `Json` initialization snippet
* Adding `@skipDocs` annotation for hiding resources from the documentation
* Skipping many `std` docs: `Resource`, `Test`, and more
* Removing `HIDDEN_CLASSES` constant for having only one way to hide documentation pieces
* UPDATE: Adding `@wingType` annotation for changing the name of built-in wing types i.e `Number` => `num`
* UPDATE: adding semicolon at the end of snippets lines
* UPDATE: fixes #3161 

### <samp>🤖 Generated by Copilot at ca942b0</samp>

This pull request improves the documentation generation for the Wing language and SDK by introducing a custom `@skipDocs` tag to mark items that should be skipped, and by simplifying the formatting and filtering of usage examples. It also adds private constructors to some utility classes and removes some unused constants. The changes affect the `apps/jsii-docgen` and `libs/wingsdk` directories.

## Checklist

- [ ] Tests added (always)


tsuf239:
image
std docs now

apps/jsii-docgen/src/docgen/schema.ts
```diff
@@ -463,6 +463,11 @@ export interface DocsSchema {
    * Inflight client
    */
   readonly inflight?: string;
+
+  /**
+   * skipping documentation
+   */
+  readonly skipDocs?: boolean;
```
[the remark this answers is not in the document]
tsuf239: @staycoolcall911 and I actually thought about it, the problem with the positive annotation is that we need to find a way to enforce it, otherwise contributors (and code reviewers) might forget, and we might have some things un-documented... 
I do like the `@wingType`/`@wingBuiltIn` annotation :) Can you please elaborate more on the second example? I feel like there is something that I'm missing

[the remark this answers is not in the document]
MarkMcCulloh: > the problem with the positive annotation is that we need to find a way to enforce it

I think this problem is mostly avoidable because there aren't going to be that many `@wingType`s. I think it'd at least as likely to forgot to add a negative annotation.

> Can you please elaborate more on the second example?

In wing, the types: `duration`, `num`, `str`, `bool`
do not match the actual class name in `std` when used in wing. To use them for static method access, you use the lowercase/short version e.g. `num.fromJson()` but currently you're generating `Number.fromJson()`

tsuf239:
oh, I see now, yes, `Number.fromJson` definitely looked weird to me 😂 
I wonder if we need to have a different way to change the name of the type since we do not always need to show the type itself (like T1 in this issue [#3161](https://github.com/winglang/wing/issues/3161))
Also, is there a chance in the future we'll want to hide other types, that aren't in the std package?

tsuf239:
So, I think I'll add the cool `@wingType` annotation, but I'm not completely convinced yet if it should also function as the hiding toggle

tsuf239:
Also, I went through the documentation with @staycoolcall911 and it seems like we need to hide those too, which aren't wing built-ins but a regular "transparent" construct we use for converting handlers:  https://www.winglang.io/docs/standard-library/cloud/api-reference#ibucketeventhandler-
I think I'll keep it as is then, `@wingType` for naming and `@skipDocs` for hiding :)

apps/jsii-docgen/src/docgen/schema.ts
```diff
@@ -463,6 +463,16 @@ export interface DocsSchema {
    * Inflight client
    */
   readonly inflight?: string;
+
+  /**
+   * skipping documentation
```
staycoolcall911, line 468:
```suggestion
   * Skipping documentation
```

apps/jsii-docgen/src/docgen/schema.ts
```diff
@@ -463,6 +463,16 @@ export interface DocsSchema {
...
+   */
+  readonly skipDocs?: boolean;
+
+  /**
+   * showing the wingType instead of the written type
```
staycoolcall911, line 473:
```suggestion
   * Showing the wingType instead of the written type
```

libs/wingsdk/src/std/generics.ts
```diff
@@ -2,5 +2,7 @@
  * Generic type argument. This type is replaced at compile time.
  *
  * @hidden
+ * @skipDocs
+ * @wingType  &lt;T&gt;
```
MarkMcCulloh, line 6: Clever!

MarkMcCulloh (approved):

PR #3257, opened 2023-07-05
fix(docs): removing reference to console installation

ekeren:
There is no need to mention wing console in the installation, the user is going to learn about it when he runs things locally

## Checklist

- [ ] Tests added (always)


docs/docs/01-start-here/02-installation.md
```diff
@@ -81,12 +81,6 @@ wing --version
 This extension adds syntax highlighting and other conveniences for the Wing language in [VSCode]. You don't *have*
 to use it, but it's great. It's available through the VSCode Marketplace [here](https://marketplace.visualstudio.com/items?itemName=Monada.vscode-wing).
 
-## Wing Console
```
ainvoner, line 84: I don't think this should be removed

ekeren: Why not? 

This is the installation guide, and there is nothing to install. I think we should take the opportunity to simplfy stuff and make the docs straight to the point. 

Take a look at 
[installing-cypress](https://docs.cypress.io/guides/getting-started/installing-cypress#npm-install)

They don't tell you that there is an electron application under the hood until you do get to [open the app](https://docs.cypress.io/guides/getting-started/opening-the-app)

ainvoner: ok

ainvoner (approved):

PR #3298, opened 2023-07-06
fix(sdk): updating some tests that should be using fetch

marciocadev:
## Checklist

- [ ] Tests added (always)


examples/tests/sdk_tests/api/patch.w
```diff
@@ -28,7 +28,7 @@ api.patch("/path/{id}", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
 test "http.patch and http.fetch can preform a call to an api" {
   let url = "${api.url}/path/${_id}";
   let response: http.Response = http.patch(url, http.RequestOptions { headers: { "content-type" => "application/json" }, body: Json.stringify(body)});
-  let fetchResponse: http.Response = http.patch(url, http.RequestOptions {method: http_PATCH, headers: { "content-type" => "application/json" }, body: Json.stringify(body)});
+  let fetchResponse: http.Response = http.fetch(url, http.RequestOptions {method: http_PATCH, headers: { "content-type" => "application/json" }, body: Json.stringify(body)});
```
eladb, line 31: Why not like this?

```suggestion
  let fetchResponse: http.Response = http.patch(url, headers: { "content-type" => "application/json" }, body: Json.stringify(body));
```

eladb (dismissed):

examples/tests/sdk_tests/api/patch.w
marciocadev: Because of the title of the test @eladb,

"http.patch and http.fetch can perform a call to an API."

Other tests show the two ways to execute the same operation, I only corrected it to make it consistent.

eladb: I see. Makes sense.

Still, I think we can use keyword arguments instead of explicit structs here:

```js
http.fetch(url, method: http_PATCH, headers: { ... }, ...);
```

marciocadev: You're right, it's much nicer this way

MarkMcCulloh (approved):

PR #3303, opened 2023-07-06
feat: console in vscode and readme touchups

### LINKED ISSUE #2380: Touch ups for the the Wing VSCode extension readme
Give it a little TLC

MarkMcCulloh:
Fixes #2380

https://github.com/winglang/wing/assets/1237390/950e9029-813d-497c-aa7b-809f0b225e9a

Note: We probably need to work on the styles of the console with a way to work within the context of vscode sizing.

Misc:
- Refactored the vscode stuff because it was getting way too messy.
- Added a `--no-open` option to the CLI

## Checklist

- [ ] Tests added (always)


apps/vscode-wing/.projenrc.ts
```diff
@@ -105,6 +105,38 @@ const contributes: VSCodeExtensionContributions = {
       },
     },
   ],
+  commands: [
```
eladb, line 108: Can we make this open as a pane instead of a tab? (like the markdown preview)

apps/vscode-wing/src/console.ts
```diff
@@ -0,0 +1,127 @@
...
+      <!DOCTYPE html>
+        <html lang="en"">
+        <head>
+            <meta charset="UTF-8">
+            <style>
+              body { margin: 0; padding: 0; }
+              iframe { display: block; height: 100vh; width: 100vw; border: none; }
+            </style>
+        </head>
+        <body><iframe src="${url[0]}/"></iframe></body>
```
eladb, line 113: Is there a way to propagate dark/light mode to the console?

eladb (approved):
Very cool!

apps/vscode-wing/.projenrc.ts
MarkMcCulloh: We can, but I don't think we should yet because the console is kinda unusable in a split view on anything other than an ultrawide screen. I think we need to work on the responsive styling first or something to help it be usable.

A user can move it into a split pane if they wish though.

apps/vscode-wing/src/console.ts
MarkMcCulloh: Should be, that's a good idea

apps/vscode-wing/.projenrc.ts
eladb: +1 that we should work on the responsive styling.

Is it possible to check the screen width and decide if we open a tab or a split view based on that?

When I work with the Console, it's extremely useful to be able to see both the code and the console at the same time.

apps/vscode-wing/src/console.ts
MarkMcCulloh: Okay this is harder than it seems, will open up a followup issue

apps/vscode-wing/.projenrc.ts
MarkMcCulloh: > Is it possible to check the screen width and decide if we open a tab or a split view based on that?

I can't find an API for it. Generally, vscode prevents raw access to the DOM so if there's no API that may not be information we can find.

> When I work with the Console, it's extremely useful to be able to see both the code and the console at the same time.

Totally agree, so I'll try it out and split it by default. Hopefully most people have big enough screens until the responsiveness improves.

Chriscbr:
Nice 👌👌👌

Issue #2928, opened 2023-06-15
wing cli analytics


ekeren:
**Goal:**
We would like to be able to learn and measure winglang's project adoption and success, including features adoption.

Having this type of information should help the winglang team to wisely decide which features should get more attention. 

We would also use this information as success indicators, for example:
- 20% of wing cli users are using it 4 days  times a week
- There are 20 organizations using `wing  compile -t awscdk` inside CI
- There is an increase of usage of `tf-gcp` target
- 50% wing test commands inside CI is used with tf-aws vs 10% of tf-aws target outside CI
- wing is used only in repos with less then 5 contributors 

### Requirements: 

```[tasklist]
### Documentation
- [ ] document all data being collected and why it is being collected see [influxdata](https://www.influxdata.com/telemetry/), [mongo](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/telemetry/), [firefox](https://www.mozilla.org/en-US/privacy/firefox/#telemetry),[airbytehq](https://github.com/airbytehq/airbyte/issues/967)
- [ ] document how you can opt-out from data collection
```

```[tasklist]
### Data collection
- [ ] we should use [segment nodejs](https://segment.com/docs/connections/sources/catalog/libraries/server/node/) for reporting from the wing cli
- [ ] it should be able to opt out from reporting anything with an environment variable WING_CLI_DISABLE_ANALYTICS
- [ ] every installation should have a unique id stored in the user's home directory, which will be used for segment's `anonymousId`
- [ ] Every time the the cli is called we should send an event with name `wing cli:<command>` and pass the options information using [track](https://segment.com/docs/connections/sources/catalog/libraries/server/node/#track)
- [ ] Every event should be called with os data collected see [snyk example](https://github.com/snyk/cli/blob/c52782f24be1ed244ae119f8d7368641c6f8b861/src/lib/analytics/getStandardData.ts#L49-L66)
- [ ] Know if we are running inside CI and take report the CI information [see checkov example](https://github.com/bridgecrewio/checkov/blob/d1536477aa2164f80aaffc1f379145b115e8399e/checkov/common/bridgecrew/run_metadata/extractors/github_actions.py#L7-L10), [snyk](https://github.com/snyk/cli/blob/c52782f24be1ed244ae119f8d7368641c6f8b861/src/lib/analytics/sources.ts#L25-L60)
[...]

PR #3310, opened 2023-07-06
feat(cli): collect analytics

hasanaburayyan:
Currently this PR introduces some recorded analytics that can be opted out of using `WING_DISABLE_ANALYTICS` env var. 

Here are some example analytics:
- User running from non git repo
image

- User running from git repo
	- Note `anonymous_repo_id` has been added
image

- Automation running in CI
	- Note `anonymous_repo_id` is added
	- Note `anonymous_id` says 'ci_run' 

image

### Implementation Details

In order to ensure we are not slowing down the user experience when running any CLI command, the process for collecting metrics works in 3 main components. Similar to how other CLI tools I have come across collect metrics. 
- Collectors: Which is responsible for gathering up all the analytics we care about during runtime of the CLI.
- Storage: The analytics gathered by the collector are stored on disk. The storage will house the events until the exporter is able to report the events back to [segment](https://segment.com/)
- Exporter: To avoid any long delays in the CLI commands, the exporter is run as a detached process when the CLI command is complete.

Right now the best way I am seeing to collect cli command specific data is by using Commander hooks around all the commands we want to collect analytics from (which right now I have as all of them)

## TODO:
- [x] Add notification on first run about analytic collection
- [x] Docs
	- [x] 	Document all data being collected
	- [x] Document opt-out policy
- [x] Determine if CLI called within CI
- [x] Collect git metrics (if exists)
- [ ] Legal review
- [ ] Github disclaimer
- [x] Setup privacy email
- [x] Use tmp dir for analytics storage
- [x] Consider best way to manage WriteKey for segment

Closes: https://github.com/winglang/wing/issues/2928
Closes: https://github.com/winglang/wing/issues/2974

## Checklist

- [x] Tests added (always)

### DIFF
--- a/apps/wing/src/analytics/collect.ts
+++ b/apps/wing/src/analytics/collect.ts
@@ -23,7 +23,6 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string> {
   let event: AnalyticEvent = {
     event: `wing cli:${cmd.name()}`,
     properties: {
-      inCI: CICollector.inCI(),
       cli: await cliCollector.collect(),
       os: await osCollector.collect(),
       node: await nodeCollector.collect(),
--- a/apps/wing/src/analytics/collect.ts
+++ b/apps/wing/src/analytics/collect.ts
@@ -5,6 +5,7 @@ import { NodeCollector } from "./collectors/node-collector";
 import { CLICollector } from "./collectors/cli-collector";
 import { CICollector } from "./collectors/ci-collector";
 import { AnalyticsStorage } from "./storage";
+import { GitCollector } from "./collectors/git-collector";
 
 
 /**
@@ -17,6 +18,7 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string | un
   const osCollector = new OSCollector();
   const nodeCollector = new NodeCollector();
   const ciCollector = new CICollector();
+  const gitCollector = new GitCollector();
   const cliCollector = new CLICollector(cmd);
 
   let event: AnalyticEvent = {
@@ -26,6 +28,7 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string | un
       os: await osCollector.collect(),
       node: await nodeCollector.collect(),
       ci: await ciCollector.collect(),
+      anonymous_repo_id: (await gitCollector.collect())?.anonymous_repo_id
     }
   }
 
--- a/apps/wing/src/analytics/collect.ts
+++ b/apps/wing/src/analytics/collect.ts
@@ -21,8 +21,10 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string | un
   const gitCollector = new GitCollector();
   const cliCollector = new CLICollector(cmd);
 
+  const eventName = `cli_${cmd.opts().target ?? ''}_${cmd.name()}`;
+
   let event: AnalyticEvent = {
-    event: `wing cli:${cmd.name()}`,
+    event: eventName.replace(/[^a-zA-Z_]/g, ""),
     properties: {
       cli: await cliCollector.collect(),
       os: await osCollector.collect(),
--- a/apps/wing/src/analytics/collect.ts
+++ b/apps/wing/src/analytics/collect.ts
@@ -14,7 +14,7 @@ import { PACKAGE_VERSION } from "../cli";
  * @param cmd The commander command to collect analytics for
  * @returns string the file path of the stored analytic
  */
-export async function collectCommandAnalytics(cmd: Command): Promise<string> {
+export async function collectCommandAnalytics(cmd: Command): Promise<string | undefined> {
   const osCollector = new OSCollector();
   const nodeCollector = new NodeCollector();
   const ciCollector = new CICollector();
@@ -26,17 +26,16 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string> {
       cli: await cliCollector.collect(),
       os: await osCollector.collect(),
       node: await nodeCollector.collect(),
+      ci: await ciCollector.collect(),
     }
-  }
-
-  if (await ciCollector.canCollect()) {
-    event.properties.ci = await ciCollector.collect();
   }
   
   let analyticFilePath = storeAnalyticEvent(event);
+  
   if (getWingAnalyticsCollectionConfig().debug) {
     console.log(`Analytics event stored at ${analyticFilePath}`);
   }
+  
   return analyticFilePath;
 }
 
--- a/apps/wing/src/analytics/collect.ts
+++ b/apps/wing/src/analytics/collect.ts
@@ -1,0 +1,95 @@
+import { Command } from "commander";
+import { loadAnalyticsConfig, storeAnalyticEvent } from "./storage";
+import { AnalyticEvent } from "./event";
+import { OSCollector } from "./collectors/os-collector";
+import { NodeCollector } from "./collectors/node-collector";
+import { CLICollector } from "./collectors/cli-collector";
+import { CICollector } from "./collectors/ci-collector";
+import { PACKAGE_VERSION } from "../cli";
+
+
+/**
+ * Collects analytics for a given command, stores it for later export
+ * 
+ * @param cmd The commander command to collect analytics for
+ * @returns string the file path of the stored analytic
+ */
+export async function collectCommandAnalytics(cmd: Command): Promise<string> {
+  const osCollector = new OSCollector();
+  const nodeCollector = new NodeCollector();
+  const ciCollector = new CICollector();
+  const cliCollector = new CLICollector(cmd);
+
+  let event: AnalyticEvent = {
+    event: `wing cli:${cmd.name()}`,
+    properties: {
+      inCI: CICollector.inCI(),
+      cli: await cliCollector.collect(),
+      os: await osCollector.collect(),
+      node: await nodeCollector.collect(),
+    }
+  }
+
+  if (await ciCollector.canCollect()) {
+    event.properties.ci = await ciCollector.collect();
+  }
+  
+  let analyticFilePath = storeAnalyticEvent(event);
+  if (getWingAnalyticsCollectionConfig().debug) {
+    console.log(`Analytics event stored at ${analyticFilePath}`);
+  }
+  return analyticFilePath;
+}
+
+/**
+ * Analytics configuration that determines whether to collect or export analytics
+ * Having these options separate makes it easy to debug when developing
+ */
+export interface AnalyticsCollectionConfig {
+  collect: boolean;
+  export: boolean;
+  debug: boolean;
+}
+
+export function getWingAnalyticsCollectionConfig(): AnalyticsCollectionConfig {
+  const optOutCollectionConfig: AnalyticsCollectionConfig = {
+    collect: false,
+    export: false,
+    debug: false,
+  }
+
+  // If user has provided opt out flag, do not collect or export anything
+  if (process.env.WING_DISABLE_ANALYTICS) {
+    return optOutCollectionConfig
+  }
+
+  // If no environment variable is set, check the analytics config file for opt-out
+  const config = loadAnalyticsConfig();
+  // If opt out is set do not 
+  if (config.optOut) {
+    return optOutCollectionConfig;
+  }
+
+  // If no opt out is set, then we should collect and export analytics
+  const optInConfig: AnalyticsCollectionConfig = {
+    collect: true,
+    export: true,
+    debug: false,
+  }
+
+  // check for debug environment variable
+  if (process.env.WING_ANALYTICS_DEBUG) {
+    // If debug is set, do not export analytics but we still want to generate them
+    optInConfig.export = false;
+    optInConfig.debug = true;
+  }
+
+  // If using local development version of wing, do not collect or export anything
+  // unless debug mode is enabled
+  if (PACKAGE_VERSION == "0.0.0" && !optInConfig.debug) {
+    return optOutCollectionConfig;
+  }
+
+  // If we've made it this far, we should collect and export analytics
+  return optInConfig;
+}
--- a/apps/wing/src/analytics/collect.ts
+++ b/apps/wing/src/analytics/collect.ts
@@ -1,11 +1,10 @@
 import { Command } from "commander";
-import { loadAnalyticsConfig, storeAnalyticEvent } from "./storage";
 import { AnalyticEvent } from "./event";
 import { OSCollector } from "./collectors/os-collector";
 import { NodeCollector } from "./collectors/node-collector";
 import { CLICollector } from "./collectors/cli-collector";
 import { CICollector } from "./collectors/ci-collector";
-import { PACKAGE_VERSION } from "../cli";
+import { AnalyticsStorage } from "./storage";
 
 
 /**
@@ -29,65 +28,11 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string | un
       ci: await ciCollector.collect(),
     }
   }
+
+  const storage = new AnalyticsStorage({debug: process.env.DEBUG ? true : false});
   
-  let analyticFilePath = storeAnalyticEvent(event);
-  
-  if (getWingAnalyticsCollectionConfig().debug) {
-    console.log(`Analytics event stored at ${analyticFilePath}`);
-  }
+  let analyticFilePath = storage.storeAnalyticEvent(event);
   
   return analyticFilePath;
 }
 
-/**
- * Analytics configuration that determines whether to collect or export analytics
- * Having these options separate makes it easy to debug when developing
- */
-export interface AnalyticsCollectionConfig {
-  collect: boolean;
-  export: boolean;
-  debug: boolean;
-}
-
-export function getWingAnalyticsCollectionConfig(): AnalyticsCollectionConfig {
-  const optOutCollectionConfig: AnalyticsCollectionConfig = {
-    collect: false,
-    export: false,
-    debug: false,
-  }
-
-  // If user has provided opt out flag, do not collect or export anything
-  if (process.env.WING_DISABLE_ANALYTICS) {
-    return optOutCollectionConfig
-  }
-
-  // If no environment variable is set, check the analytics config file for opt-out
-  const config = loadAnalyticsConfig();
-  // If opt out is set do not 
-  if (config.optOut) {
-    return optOutCollectionConfig;
-  }
-
-  // If no opt out is set, then we should collect and export analytics
-  const optInConfig: AnalyticsCollectionConfig = {
-    collect: true,
-    export: true,
-    debug: false,
-  }
-
-  // check for debug environment variable
-  if (process.env.WING_ANALYTICS_DEBUG) {
-    // If debug is set, do not export analytics but we still want to generate them
-    optInConfig.export = false;
-    optInConfig.debug = true;
-  }
-
-  // If using local development version of wing, do not collect or export anything
-  // unless debug mode is enabled
-  if (PACKAGE_VERSION == "0.0.0" && !optInConfig.debug) {
-    return optOutCollectionConfig;
-  }
-
-  // If we've made it this far, we should collect and export analytics
-  return optInConfig;
-}
--- a/apps/wing/src/analytics/collectors/ci-collector.ts
+++ b/apps/wing/src/analytics/collectors/ci-collector.ts
@@ -4,73 +4,33 @@ export interface CIData {
   name: string;
 }
 
-export enum CIType {
-  GITHUB_ACTIONS = "github_action",
-  GITLAB_CI = "gitlab_ci",
-  JENKINS = "jenkins",
-  CIRCLECI = "circleci",
-  BITBUCKET = "bitbucket",
-  AZURE_DEVOPS = "azure_devops",
-  TEAMCITY = "teamcity",
-  AWS_CODEBUILD = "aws_codebuild",
-  NONE = "none",
+// Map of CI environment variable identifiers to CI names
+const CI_ENV: {[key: string]: string} = {
+  // https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+  GITHUB_ACTIONS: "GITHUB_ACTIONS",
+  // https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
+  GITLAB_CI: "GITLAB_CI",
+  // https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
+  JENKINS_URL: "JENKINS",
+  // https://circleci.com/docs/variables/#built-in-environment-variables
+  CIRCLECI: "CIRCLECI",
+  // https://support.atlassian.com/bitbucket-cloud/docs/variables-and-secrets/
+  BITBUCKET_BUILD_NUMBER: "BITBUCKET",
+  // https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
+  BUILD_BUILDID: "AZURE_DEVOPS",
+  // https://www.jetbrains.com/help/teamcity/predefined-build-parameters.html
+  TEAMCITY_VERSION: "TEAMCITY",
+  // https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-variables.html
+  CODEBUILD_BUILD_ID: "CODEBUILD",
 }
 
 export class CICollector extends Collector {
-  
-  static inCI(): boolean {
-    return this.getCIType() !== CIType.NONE;
-  }
-
-  static getCIType(): CIType {
-    // Github builtin environment variables
-    // https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
-    if (process.env.GITHUB_ACTIONS) {
-      return CIType.GITHUB_ACTIONS;
+  async collect(): Promise<CIData | undefined> {
+    for (const e in CI_ENV) {
+      if (e in process.env) {
+        return {name: CI_ENV[`${e}`]}
+      }
     }
-    // Gitlab builtin environment variables
-    // https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
-    if (process.env.GITLAB_CI) {
-      return CIType.GITLAB_CI;
-    }
-    // Jenkins builtin environemnt variables
-    // https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
-    if (process.env.JENKINS_URL) {
-      return CIType.JENKINS;
-    }
-    // CircleCI builtin environment variables
-    // https://circleci.com/docs/variables/#built-in-environment-variables
-    if (process.env.CIRCLECI) {
-      return CIType.CIRCLECI;
-    }
-    // Bitbucket builtin environment variables
-    // https://support.atlassian.com/bitbucket-cloud/docs/variables-and-secrets/
-    if (process.env.BITBUCKET_BUILD_NUMBER && process.env.CI) {
-      return CIType.BITBUCKET;
-    }
-    // Azure DevOps builtin environment variables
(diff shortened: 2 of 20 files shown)


apps/wing/src/analytics/storage.ts
```diff
@@ -0,0 +1,94 @@
...
+  return config.anonymousId;
+}
+
+export function loadAnalyticsConfig(): AnalyticsConfig {
+  try {
+    const fileContents = readFileSync(WING_ANALYTICS_CONFIG_FILE, 'utf-8');
+    return JSON.parse(fileContents) as AnalyticsConfig;
+  } catch (error) {
+    // Possible issue reading analytics config file, first determine if file exists yet
+    if (existsSync(WING_ANALYTICS_CONFIG_FILE)) {
```
MarkMcCulloh, line 55: nit: Instead of checking for existence, check out the error codes from the error to see if it's ENOENT or EACCES.

apps/wing/src/analytics/disclaimer.ts
```diff
@@ -0,0 +1,41 @@
+import chalk from "chalk";
+import { loadAnalyticsConfig, saveAnalyticsConfig } from "./storage";
+
+const disclaimer = `
+🧪 This is an early pre-release of the Wing Programming Language (aka "alpha").
```
eladb, line 5:
```suggestion
🧪 This is an early pre-release of the Wing Programming Language.
```

apps/wing/src/analytics/event.ts
```diff
@@ -0,0 +1,17 @@
...
+import { CLIData } from "./collectors/cli-collector";
+import { NodeData } from "./collectors/node-collector";
+import { OSData } from "./collectors/os-collector";
+
+export interface AnalyticEvent {
+  event: string;
+  timestamp?: string;
+  anonymousId?: string;
+  properties: {
+    inCI: boolean;
```
eladb, line 11: do we really need both `inCI` and `ci`. We can assume that if `ci` has contents we are in a CI context, no?

apps/wing/src/analytics/scripts/detached-export.ts
```diff
@@ -0,0 +1,30 @@
+import { getAnonymousId, loadAnalyticsEvent } from '../storage';
+import * as fs from 'fs';
+import Analytics from '@segment/analytics-node';
+import { getWingAnalyticsCollectionConfig } from '../collect';
+
+// When this file is run as a child process, it will be passed the path 
+// to the analytics report file
+const filePath = process.argv[2];
+
```
eladb, line 9: error is argv[2] is not supplied

apps/wing/src/analytics/scripts/detached-export.ts
```diff
@@ -0,0 +1,30 @@
...
+      properties: event.properties,
+    });
+  }
+
+  // Keeps the file around for debugging purposes
+  if (!analyticsConfig.debug) {
+    fs.unlinkSync(filePath);
+  }
+} catch(err) {
+  // TODO: add mechanism to retry
```
eladb, line 29: i am assuming if we fail, we will simply send the file again next time, no? that seems like a reasonable retry

apps/wing/src/analytics/storage.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @param event the event to store
+ * @returns string the file path of the stored analytic
+ */
+export function storeAnalyticEvent(event: AnalyticEvent): string {
+  const eventId = uuidv4();
+  const tmpdir = path.join(os.tmpdir(), "wing-analytics");
+  
+  if (!existsSync(tmpdir)) {
+    mkdirSync(tmpdir);
+  }
```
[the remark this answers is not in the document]
eladb: i think we *never* want to fail if something doesn't work with our analytics. probably need to surround everything with try/catch like you did in the `detached-export` script

hasanaburayyan:
@eladb and @MarkMcCulloh thanks for starting to take a look at this one. Im going to add the `do-not-merge` I forgot I made this not a draft and still have TODOs 😁 but love the feedback keep it coming

apps/wing/src/cli.ts
```diff
@@ -7,7 +7,16 @@ import { satisfies } from "compare-versions";
...
-const PACKAGE_VERSION = require("../package.json").version as string;
+import { collectCommandAnalytics, getWingAnalyticsCollectionConfig } from "./analytics/collect";
+import { exportAnalytics } from "./analytics/export";
+import { optionallyDisplayDisclaimer } from "./analytics/disclaimer";
+
+
+export const PACKAGE_VERSION = require("../package.json").version as string;
+
+const ANALYTICS_COLLECTION_CONFIG = getWingAnalyticsCollectionConfig(); 
+let ANALYTICS_EXPORT_FILE: string | undefined = undefined;
```
eladb, line 18: ALL_CAPS usually represent constants

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Wing Cli Analytics
+id: wing-analytics
+keywords: [Analytics, Telemetry]
+---
+
+## Wing Anonymous Analytics
+
+The Wing cli collects anonymous usage data in order to help us improve the cli experience. The intent is to collect data that will help us understand how the cli is being used throughout the community, with hopes that this will help us to build the right features and improve the right areas.
```
eladb, line 9:
```suggestion
The Wing CLI collects anonymous usage data in order to help us improve. The intent is to collect data that will help us understand how the CLI is being used throughout the community, with hopes that this will help us to build the right features and improve the right areas.
```

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+
+You may not feel comfortable sharing this data with us, and that is ok. The gathering of these analytics is completely optional to use the Wing cli, and can be opted out of at any time.
```
eladb, line 11:
```suggestion
You may not feel comfortable sharing this data with us, and that is ok. The gathering of these analytics is completely optional to use the Wing CLI, and can be opted out of at any time.
```

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+```sh
+wing compile -t sim app.w --no-analytics
+```
+
+### What data is collected
+
+Each run of the Wing cli collects information about various aspects of the environment it is running in. Below are the analytics we gather. As well
+you can always head over to our Github and view the code that collects these metrics, as the code is open source.
+
+- The Wing cli version
```
eladb, line 26:
```suggestion
- The Wing CLI version
```

apps/wing/src/analytics/scripts/detached-export.ts
hasanaburayyan: currently in order to avoid a race condition (if multiple wing commands run in parallel)  the detached script is called with a specific analytic file uuid, so there would need to be some sort of watch or failure log or something for the subsequent runs to know to try exporting the failed metric.

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+- The command information, including the command name, arguments, and flags
+- Operating system information, including the architecture, version, and platform
+- Node js version
```
eladb, line 29:
```suggestion
- Node.js version
```

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+- CI environment name (if applicable)
```
eladb, line 30: What about Wing SDK version? Console version?

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
+---
```
eladb, line 1: Why is this under 07-examples?

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+title: Wing Cli Analytics
```
eladb, line 2:
```suggestion
title: Analytics
```

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+id: wing-analytics
+keywords: [Analytics, Telemetry]
+---
+
+## Wing Anonymous Analytics
```
eladb, line 7:
```suggestion
## Analytics Collection
```

docs/docs/07-examples/14-wing-analytics.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Wing Cli Analytics
+id: wing-analytics
```
eladb, line 3:
```suggestion
id: analytics
```

apps/wing/src/analytics/event.ts
hasanaburayyan: I went back and forth on this. In my mind it would depend on what we are using downstream to query/visualize these analytics. I assume you are right, but I wonder if having this easy true/false field would make some query easier downstream. 🤷 Happy to take it out, I wonder if @ekeren has any thoughts here?

docs/docs/07-examples/14-wing-analytics.md
hasanaburayyan: Just a placeholder for now, wanted to confirm where this was actually going to go.

apps/wing/src/cli.ts
```diff
@@ -83,6 +126,7 @@ async function main() {
     )
     .option("-p, --plugins [plugin...]", "Compiler plugins")
     .hook("preAction", progressHook)
+    .hook("preAction", collectAnalyticsHook)
```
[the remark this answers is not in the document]
hasanaburayyan: let me try that again, I could not get "preSubcommand" to work last week (iirc I was not getting some command information I expected) but perhaps I was goofing something up. I would much rather have it once and done.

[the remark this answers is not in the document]
hasanaburayyan: Ah I see I was trying to parse the root command rather than the sub command when using that hook, which is why I thought data was missing. This works like a charm good call

apps/wing/src/cli.ts
```diff
@@ -53,6 +66,36 @@ async function main() {
...
     });
 
+
+    async function collectAnalyticsHook(cmd: Command) {
+      if (!ANALYTICS_COLLECTION_CONFIG.collect) { return; }
+  
+      // Fail silently if collection fails
+      try {
+        optionallyDisplayDisclaimer();
+        ANALYTICS_EXPORT_FILE = await collectCommandAnalytics(cmd);
```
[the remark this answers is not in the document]
hasanaburayyan: ahh I see, yea that voodoo works I had not thought about it. 😅

apps/wing/src/analytics/scripts/detached-export.ts
MarkMcCulloh: imo I don't think we should burden the user's system/io with retries. Metrics are great but let's not be greedy, especially since our metrics are opt-out. 
At most we should retry once in this same process (which luckily doesn't block anything) and if it doesn't work then just consider the data lost.

apps/wing/src/analytics/storage.ts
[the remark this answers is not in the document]
MarkMcCulloh: For this particular chunk, the exception is actually used for logic rather than just handling failures. But yeah I would agree this whole thing also needs a try/catch anyways.

apps/wing/src/analytics/collectors/ci-collector.ts
```diff
@@ -0,0 +1,76 @@
...
+  CIRCLECI = "circleci",
+  BITBUCKET = "bitbucket",
+  AZURE_DEVOPS = "azure_devops",
+  TEAMCITY = "teamcity",
+  AWS_CODEBUILD = "aws_codebuild",
+  NONE = "none",
+}
+
+export class CICollector extends Collector {
+
```
eladb, line 20: Nit: This class feels a bit over designed...

I'd just write it like so:

```js
// list of supported CI environment variables
const ENV = [
  "GITHUB_ACTIONS",
  "GITLAB_CI",
  // ...
};

export class CICollector extends Collector {
  async collect(): Promise<CIData | undefined> {
    for (const e of ENV) {
      if (e in process.env) {
        return { name: e };
      }
    }

    return undefined;
  }
}
```

I don't see a real need to mangle the env name in the `name` and also, we can update the contract of the `Collector` to allow `undefined` to be returned, indicating this collection is empty.

(this requires a change to `collect.ts` to not call `canCollect()`.

ainvoner:
@hasanaburayyan wing and wing console should use the same analytics anonymous id.
In order for the console to have access to the analytics module, it should be developed as a library and not as part of the wing app (perhaps only the anonId creation and getter?).
can you please refactor your solution to support it?
basically the console needs only the `getAnonymousId()`

apps/wing/src/analytics/collectors/ci-collector.ts
hasanaburayyan: thats fair, yea this originally was a lot more complicated when I had it designed to get much more information than just the CI environment name. (repo id, commit hash, etc..) Ill simplify this

hasanaburayyan:
> @hasanaburayyan wing and wing console should use the same analytics anonymous id. In order for the console to have access to the analytics module, it should be developed as a library and not as part of the wing app (perhaps only the anonId creation and getter?). can you please refactor your solution to support it? basically the console needs only the `getAnonymousId()`

@ainvoner sure let me take a look at extracting this today

apps/wing/src/analytics/collectors/ci-collector.ts
hasanaburayyan: Although with mangling the name, its not always the case that an environment id is a nice indicator of what CI it is in. Like with Jenkins we only know based on JENKINS_URL which is not really a good name, but we would know its Jenkins (which could be okay) but a step worse might be azure which is identified by the environment variable `BUILD_BUILDID` I would have no idea looking at downstream analytics that is azure

eladb: So you can make `ENV` a map from variable to identifier.

apps/wing/src/analytics/collect.ts
```diff
@@ -0,0 +1,93 @@
...
+  const ciCollector = new CICollector();
+  const cliCollector = new CLICollector(cmd);
+
+  let event: AnalyticEvent = {
+    event: `wing cli:${cmd.name()}`,
+    properties: {
+      cli: await cliCollector.collect(),
+      os: await osCollector.collect(),
+      node: await nodeCollector.collect(),
+      ci: await ciCollector.collect(),
```
ekeren, line 29: We need to flatten this all things under `properties,` nested maps are not good for our analytics consumers 

Instead of:
```
properties: { 
  cli: { 
    k1: "v1", 
    k2: { 
      k3: "v2"
    }
  }
}
```
Is should be:
```
properties: { 
  cli_k1: "v1",
  cli_k2_k3: "v2
}
```

apps/wing/src/analytics/storage.ts
```diff
@@ -0,0 +1,96 @@
...
+    const analyticsConfig: AnalyticsConfig = {
+      anonymousId: randomBytes(16).toString('hex'),
+      optOut: false,
+    }
+
+    saveAnalyticsConfig(analyticsConfig);
+
+    return analyticsConfig;
+  }
+}
```
ekeren, line 75: @hasanaburayyan , 

I was thinking that in CI we will not have this value, the result is that in every CI run the system will generate a new `anonymousId` 

One option I have is to take the git repo and run a hash function on it (e.g. md5) and using the hashed versions  as an anonymousId in the case of CI 

I this case we can't determine the real github repo, (unless we guess it and run the same hash function) 

It might be also nice to add this anonymouRepoId in all the events, we can use it to understand correlation  between anonymouId and anonymouRepoId  without any privacy issues

apps/wing/src/analytics/collect.ts
ekeren: also notice that properties should be snake case (just a naming convention)

apps/wing/src/analytics/storage.ts
ekeren: actually anonymouRepoId should be inside `properties` and should be called : `anonymou_repo_id`

ekeren: @ainvoner - notice that the console should also have the `anonymou_repo_id` property

apps/wing/src/analytics/collect.ts
hasanaburayyan: Love the call out. I did not think about consumers querying nested maps aspect. Will fix this conventions

apps/wing/src/analytics/storage.ts
ekeren: yes, I am full of typos

apps/wing/src/analytics/collectors/cli-collector.ts
```diff
@@ -0,0 +1,30 @@
...
+    super();
+    this.cmd = cmd;
+  }
+
+  async collect(): Promise<CLIData> {
+    return {
+      target: this.cmd.opts().target,
+      options: `${JSON.stringify(this.cmd.opts())}`,
+      version: PACKAGE_VERSION,
+      wingSDKVersion: require(`@winglang/sdk/package.json`).version as string,
```
ekeren, line 26:
```suggestion
      wing_sdk_version: require(`@winglang/sdk/package.json`).version as string,
```

apps/wing/src/analytics/collectors/cli-collector.ts
```diff
@@ -0,0 +1,30 @@
...
+      wingConsoleVersion: require(`@wingconsole/app/package.json`).version as string
```
ekeren, line 27:
```suggestion
      wing_console_version: require(`@wingconsole/app/package.json`).version as string
```

apps/wing/src/analytics/collect.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @returns string the file path of the stored analytic
+ */
+export async function collectCommandAnalytics(cmd: Command): Promise<string | undefined> {
+  const osCollector = new OSCollector();
+  const nodeCollector = new NodeCollector();
+  const ciCollector = new CICollector();
+  const cliCollector = new CLICollector(cmd);
+
+  let event: AnalyticEvent = {
+    event: `wing cli:${cmd.name()}`,
```
ekeren, line 24: Some naming convention change, it is <source>_<object>_<action>
in this case:
-  source is `cli`
- object is `--target` 
- action is `test/compile/etc`
 
if object (target) is empty (like in `wing docs` it should be "nil"), unless there is a default (then it should be the default) 
Also, lets remove anything in the object/action that is not normal chars [a-zA-Z] `str.replace(/[^a-zA-Z]/g, '');`

meaning it should be:
- `name: cli_sim_test`
- `name: cli_tfaws_compile`
- `name: cli_sim_compile`
- `name: cli_tfaws_compile`

apps/wing/src/analytics/collectors/cli-collector.ts
```diff
@@ -0,0 +1,30 @@
...
+  private cmd: Command;
+
+  constructor(cmd: Command) {
+    super();
+    this.cmd = cmd;
+  }
+
+  async collect(): Promise<CLIData> {
+    return {
+      target: this.cmd.opts().target,
```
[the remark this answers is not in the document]
hasanaburayyan: Yes since the commander CLI uses sim as default for commands that do not require `--target` like `test`

apps/wing/src/analytics/collect.ts
hasanaburayyan: sorry not sure if I get this completely. In the above are `cli_sim_test`, `cli_tfaws_compile`, etc.. values in `event`? Also does that mean that running docs would produce `cli_nil_docs` or `cli_docs` or `cli__docs`

apps/wing/src/analytics/storage.ts
hasanaburayyan: Also @ekeren I believe we have to keep anonymousID even if it was in CI this is because its a required value in segment. Though if its in CI I can just make the value say "see anonymous_repo_id"

apps/wing/src/analytics/collect.ts
ekeren: Let's go with `cli__docs`, I think it is the best approach

hasanaburayyan: alright heres what it looks like
image

apps/wing/src/cli.ts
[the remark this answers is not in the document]
hasanaburayyan: @MarkMcCulloh I reverted back to adding this as a "preAction" hook to each command. It appears that the lifecycle hook for "preSubcommand" does not get the options on the subcommand. The only options I can parse are the defaults.

apps/wing/src/analytics/collect.ts
```diff
@@ -0,0 +1,43 @@
...
+  const eventName = `cli_${cmd.opts().target ?? ''}_${cmd.name()}`;
+
+  let event: AnalyticEvent = {
+    event: eventName.replace(/[^a-zA-Z_]/g, ""),
+    properties: {
+      cli: await cliCollector.collect(),
+      os: await osCollector.collect(),
+      node: await nodeCollector.collect(),
+      ci: await ciCollector.collect(),
+      anonymous_repo_id: (await gitCollector.collect())?.anonymous_repo_id
```
eladb, line 33: Why is this not just?

```suggestion
      git: await gitCollector.collect()
```

apps/wing/src/analytics/collectors/cli-collector.ts
```diff
@@ -0,0 +1,30 @@
...
+      options: `${JSON.stringify(this.cmd.opts())}`,
+      version: PACKAGE_VERSION,
+      wing_sdk_version: require(`@winglang/sdk/package.json`).version as string,
+      wing_console_version: require(`@wingconsole/app/package.json`).version as string
```
eladb, line 27: Maybe add some try/catch around this so that not everything fails if we can't require the module:

```js
function tryRequire(m) {
  try {
    return require(m);
  } catch {
    return undefined;
  }
}
```

apps/wing/src/analytics/collectors/collector.ts
```diff
@@ -0,0 +1,28 @@
+import { execFile } from "child_process";
+
+export interface Data {
+  [key: string]: any;
+}
+
+export abstract class Collector {
+  abstract collect(): Promise<Data | undefined>;
+
+  async runCommand(cmd: string, args: string[]): Promise<any> {
```
eladb, line 10:
```suggestion
  protected async runCommand(cmd: string, args: string[]): Promise<any> {
```

apps/wing/src/analytics/storage.ts
```diff
@@ -0,0 +1,202 @@
...
+   */
+  public loadConfig(): AnalyticsConfig {
+    try {
+      const fileContents = readFileSync(this.analyticsConfigFile, 'utf-8');
+      return JSON.parse(fileContents) as AnalyticsConfig;
+    } catch (error: any) {
+      // Possible issue reading analytics config file, first determine if file exists yet
+  
+      if (error.code === 'EACCES') {
+        throw new Error("Not able to access analytics config file due to permissions");
```
eladb, line 148: Is this caught somewhere? I don't think we need this to ever fail even if we can't store the file

apps/wing/src/analytics/storage.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Helper method to flatten objects
+   * 
+   * @param properties The properties to flatten
+   * @param parentKey The key of the parent (defaults to "")
+   * @returns 
+   */
+  private flattenProperties(properties: {[key: string]: any}, parentKey: string = ""): {[key: string]: any} {
```
eladb, line 181: Why is this needed?

docs/docs/09-analytics.md
```diff
@@ -0,0 +1,30 @@
...
+### What data is collected
+
+Each run of the Wing cli collects information about various aspects of the environment it is running in. Below are the analytics we gather. As well
+you can always head over to our Github and view the code that collects these metrics, as the code is open source.
+
+- The Wing CLI, SDK, and Console versions
+- The command information, including the command name, arguments, and flags
+- Operating system information, including the architecture, version, and platform
+- Node.js version
+- CI environment name (if applicable)
```
eladb, line 30: Missing anonymous git repo

eladb (approved):
LGTM. Some notes.

apps/wing/src/analytics/scripts/detached-export.ts
```diff
@@ -0,0 +1,58 @@
...
+  const analytics = new Analytics({ writeKey: 'sCqPF5xSscOjJdi5Tbkqu73vfF8zkZdw'});
+  const storage = new AnalyticsStorage();
+  const event = storage.loadEvent(filePath);
+
+  if (!event) {
+    throw new Error(`No analytic event found at: ${filePath}`);
+  }
+
+  const params = {
+    anonymousId: (event.properties as any).ci_name ? 'ci_run' : storage.getAnonymousId(),
```
ekeren, line 29: `ci_run` is not a good value. But because I have `anonymous_repo_id` I can work the details in the dashboards. Lets revert to using `getAnonymousId`

```suggestion
    anonymousId: storage.getAnonymousId(),
```

apps/wing/src/analytics/storage.ts
```diff
@@ -0,0 +1,202 @@
...
+   * Saves the analytics config to disk
+   * 
+   * @param config the analytics config to save to disk
+   */
+  public saveConfig(config: AnalyticsConfig) {
+    if (!existsSync(WING_HOME_DIR)) {
+      mkdirSync(WING_HOME_DIR);
+    }
+    writeFileSync(this.analyticsConfigFile, JSON.stringify(config))
+  }
```
ekeren, line 172: Maybe you should guard this with try catch ? (we might find this is a readonly fs)

apps/wing/src/analytics/storage.ts
ekeren: https://github.com/winglang/wing/pull/3310#discussion_r1258312403

ekeren: some analytics consumers don't like nested property values

apps/wing/src/analytics/collect.ts
ekeren: Current solution is ok (talked to Elad)

docs/docs/09-analytics.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Analytics
+id: analytics
+keywords: [Analytics, Telemetry]
+---
+
+## Analytics Collection
+
+The Wing CLI collects anonymous usage data in order to help us improve. The intent is to collect data that will help us understand how it is being used throughout the community, with hopes that this will help us to build the right features and improve the right areas.
```
staycoolcall911, line 9:
```suggestion
The Wing CLI collects anonymous usage data in order to help us improve. The intent is to collect data that will help us understand how it is being used throughout the community, with hopes that this will help us build the right features and improve the right areas.
```

apps/wing/src/cli.ts
```diff
@@ -27,6 +39,33 @@ function actionErrorHandler(fn: (...args: any[]) => Promise<any>) {
   };
 }
 
+async function collectAnalyticsHook(cmd: Command) {
+  if (process.env.WING_DISABLE_ANALYTICS) { return; }
```
staycoolcall911, line 43: If the user opted out of anlytics collection, it would be nice to show some feedback that no analytics are sent (maybe print to stdout)

hasanaburayyan: Let me think about how best to handle output info on analytics collection. I think this might become noise, maybe its 1. if disable AND if debug then we print the message.

Ill make suggestion in subsequent pr.

PR #3312, opened 2023-07-07
chore(sdk): add out of bounds exception to string.at

### LINKED ISSUE #3240: String.at should return an exception in case of a failure
### I tried this:

`"book".at(100)`

### This happened:

Undefined is returned 

### I expected this:

An exception should be thrown

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

fynnfluegge:
Adds exception to `string.at` if index is out of bounds.

closes #3240 

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/std/string.ts
```diff
@@ -41,6 +41,8 @@ export class String {
   /**
    * Returns the character at the specified index.
    *
+   * @macro ((args) => { if ($args$ >= $self$.length) {throw new Error("index out of bounds")}; return $self$.at($args$) })($args$)
```
tsuf239, line 44: what about `"boom".at(-5)`?  (in js is undefined) 
Should it throw the same error?

ekeren: yes

fynnfluegge: done 👌✅

examples/tests/sdk_tests/std/string.w
```diff
@@ -39,19 +39,28 @@ test "length" {
...
   assert("boom".at(0) == "b");
   // Negative integers count back from the last string character.
   assert("boom".at(-4) == "b");
   assert("boom".at(-1) == "m");
-  
-  // Should throw an exception instead of returns undefined see issue #3240
-  // assert("boom".at(-10) == "m");
+  // Should throw an exception
+  try { "boom".at(4); } catch s { assert(s == INDEX_OUT_OF_BOUNDS_ERROR ); }
+  try { "boom".at(-5); } catch s { assert(s == INDEX_OUT_OF_BOUNDS_ERROR ); }
```
tsuf239, line 63: Is there a reason it's not `assertThrows`? 
since we are visiting the `catch` scope only if there is an error, it will pass either way...
It seems like it can be removed :)

tsuf239 (approved):

revitalbarletz:
Nice!

examples/tests/sdk_tests/std/string.w
fynnfluegge: True, but trying to call `assertThrows` in `test {}` throws `error: Cannot call into preflight phase while inflight`. I think that's the reason why many tests doing this that way

fynnfluegge: So sure it can be removed, but then we should remove it in many tests, I prefer to be consistent 😅

tsuf239: > True, but trying to call `assertThrows` in `test {}` throws `error: Cannot call into preflight phase while inflight`. I think that's the reason why many tests doing this that way

That's a good point!! I didn't know that, can you please open an issue and add a comment next to those lines? (no need to remove them), it would be super helpful 🙏

fynnfluegge: okidoki 🙏✅

PR #3333, opened 2023-07-09
feat(sdk): adding uuidv4, nanoid and sha256 to Utils

marciocadev:
Unfortunately, I had to limit `nanoid` in version 3 because starting from version 4, it no longer works with CommonJS.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/util/util.ts
```diff
@@ -1,3 +1,6 @@
+import { SHA256 } from "crypto-js";
```
eladb, line 1: why not use node.js's `crypto` library? (always prefer not to use a dependency if possible)

eladb (changes requested):

MarkMcCulloh (approved):

Issue #1935, opened 2023-03-29
`winglang` npm package description needs some work


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

https://www.npmjs.com/package/winglang

### I expected this:

To see something similar to what we have in the root README

### Instead, this happened:

Something that seems unmaintained

### Component:

CLI

PR #3335, opened 2023-07-09
chore(cli): copy package README from root README

hasanaburayyan:
This just copies the README and referenced gif asset into `app/wing` on build of the wing cli.

Closes: https://github.com/winglang/wing/issues/1935

TODO: still looking into if symlinks could work. (though this is pretty simple use case atm)

## Checklist

- [x] Tests added (always)


apps/wing/README.md
```diff
@@ -1,42 +1,144 @@
...
+
+We are incredibly grateful to our entire community for contributing bug fixes and improvements:
+
+<a href="https://github.com/winglang/wing/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=winglang/wing" />
+</a>
+
+## License 📜
+
+Wing is licensed under the  [MIT License](./LICENSE.md).
```
staycoolcall911, line 139: You're referencing the license in the cli dir here, which is different from our license on the root dir. Let's ask the expert if we should also copy the license.md here. @ShaiBer

staycoolcall911 (approved):

apps/wing/README.md
staycoolcall911: Discussed this with @ShaiBer and the conclusion is that it is actually the right approach to link to the license found on the same dir. Resolving.

PR #3359, opened 2023-07-11
fix: prs can be accidentally merged with faulty self-mutation content

### LINKED ISSUE #3318: PRs can be accidentally merged with faulty self-mutation content
https://github.com/winglang/wing/pull/3302 was merged automatically with a faulty snapshots  which caused a regression. The flow was:

1. PR was approved
2. Build failed
3. I tried to fix it and pushed a commit with a fix, but then realized it had other regressions
4. In the meantime, self-mutation happened and new (faulty) snapshot committed.
5. Now, all checks passed and the PR was merged... Ouch

Had to revert it.

---

A simple solution is for our self-mutation flow to somehow block auto-merge. This can be done either by adding a conversation or a label as self mutation occurs (I suspect the lowest hanging fruit is to simply apply a `do-no-merge` label.

WeepingClown13:
Adds a "🚧 pr/do-not-merge" label during mutation to prevent auto merge of pull requests with mutation.
closes #3318

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -34,10 +34,12 @@ jobs:
           name_is_regexp: true
           if_no_artifact_found: ignore
           path: patches
+
       - uses: marocchino/action-workflow_run-status@54b6e87d6cb552fc5f36dbe9a722a6048725917a
         if: steps.download-artifacts.outputs.found_artifact == 'true'
         with:
           github_token: ${{secrets.GITHUB_TOKEN}}
+
       - name: Token check
         if: steps.download-artifacts.outputs.found_artifact == 'true'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/mutation.yml
```diff
@@ -49,10 +51,22 @@ jobs:
...
+
+      - name: Add label to block auto merge
+        if: github.event.workflow_run.event == 'pull_request' && steps.download-artifacts.outputs.found_artifact == 'true'
+        run: |
+          curl \
+            -X POST \
+            -H "Authorization: Bearer ${{ secrets.MUTATION_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            "https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.workflow_run.pull_requests[0].number }}/labels" \
+            -d '["🚧 pr/do-not-merge"]'
```
MarkMcCulloh, line 63: I think we should use a new label to make sure the intent is clear. I think people would be hesitant to remove a  `do-not-merge` label. Maybe something like `pr/review-mutation`. With a new label we can have a description that explains what to do.

The new label would need to be in the mergify config as well.

WeepingClown13: Yes, this is something I thought about as well. Do we have an emoji for this? xD

MarkMcCulloh: Not sure, here are the first ones that came to mind:
- ❗ or ❕
- ⚠️ or ⚠︎

Feel free to pick 😄

.mergify.yml
```diff
@@ -20,7 +20,7 @@ pull_request_rules:
           {{ body }}
     conditions:
       - -title~=(WIP|wip)
-      - -label~=(do-not-merge)
+      - -label~=(🚧 pr/do-not-merge|⚠️ pr/review-mutation)
```
MarkMcCulloh, line 23: Make sure to check in the config editor to see if it picks these up properly. Mostly concerned about the `/`

MarkMcCulloh (approved):

.mergify.yml
WeepingClown13: No worries, I did thoroughly check first before commiting.

PR #3364, opened 2023-07-11
ci(Mergify): configuration update

WeepingClown13:
This change has been made by @WeepingClown13 from Mergify config editor.


WeepingClown13:
Testing in mergify playground went wrong... *sweats*

PR #3367, opened 2023-07-12
fix: lazy importing perf improvements

### LINKED ISSUE #2980: Let's do another performance optimization round
### I tried this:

Lately it seems the toolchain has become a bit slower. Not sure exactly why, but let's do another performance profiling round and identify some low hanging fruit.

### This happened:

Slowness

### I expected this:

Fastness

### Is there a workaround?

_No response_

### Component

Compiler, SDK, CLI

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
Fixes #2980

Use dynamic `import`s and `requires` it certain places to limit the loaded deps. Primarily in two places:

- In the cli, each command is only loaded when run
  - Avoids any aws-sdk imports for sim target compilations/tests
  - Avoids any `@wingconsole` imports for non-`wing it` usage
  - Avoids `vscode/languageserver` import for non-lsp usage
- Each individual simulator resource is required dynamically in the `DefaultSimulatorFactory`.
  - Avoids unused dockerode, redis, express, etc.

Misc Changes:
- bundled wasi-js. Very slightly improves load time and gets rid of the patched bundled dep
- Always do release builds for rust, even locally. It doesn't take much longer to build and it runs significantly faster (the wasm file is like 10x smaller and serde is properly optimized)
- Switch to https://www.npmjs.com/package/tiny-updater from https://www.npmjs.com/package/update-notifier
  - update-notifier took ~150ms just to import, which happens every invocation regardless of actually checking the internet
- Disabled test isolation in hangar. We don't need it and it should run slightly faster
- Removed esbuild devdep from wing cli, we aren't using it right now

Currently working on benchmark comparisons, here's a preview for this PR (Uses the e2e benchmark data):

```
main -> PR#3367
Avg Change: -41.66%
hello_world.w -t sim
         -58.24% (1091ms -> 455ms)
hello_world.w -t tf-aws
         -11.26% (5567ms -> 4940ms)
jsii_small.w -t sim
         -61.66% (1043ms -> 400ms)
jsii_small.w -t tf-aws
         -53.26% (1158ms -> 541ms)
functions_10.w -t sim
         -55.67% (1128ms -> 500ms)
functions_10.w -t tf-aws
         -15.01% (4740ms -> 4028ms)
jsii_big.w -t sim
         -26.15% (8449ms -> 6240ms)
jsii_big.w -t tf-aws
         -19.75% (8103ms -> 6502ms)
functions_1.w -t sim
         -59.10% (1128ms -> 461ms)
functions_1.w -t tf-aws
         -24.79% (2761ms -> 2076ms)
empty.w -t sim
         -59.14% (1039ms -> 424ms)
empty.w -t tf-aws
         -55.82% (1191ms -> 526ms)
```

## Checklist

- [ ] Tests added (always)


apps/wing/src/commands/upgrade.ts
```diff
@@ -1,26 +1,22 @@
...
 
 const log = debug("wing:update");
-const PJSON = require("../../package.json");
 const DEFAULT_UPDATE_RATE = 1000 * 60 * 60 * 24; // 1 day
 
 /** Handles checking for toolchain updates */
 export async function checkForUpdates() {
-  if (!process.stdout.isTTY) {
-    log("not a TTY, skipping update check");
+  if (!process.stdout.isTTY || process.env.CI === "true" || currentPackage.version === "0.0.0") {
```
skyrpex, line 9: Is it possible that `process.env.CI` is equal to `"1"` in some places? Maybe a more fuzzy check would be preferred.

apps/wing/src/commands/upgrade.ts
```diff
@@ -1,26 +1,22 @@
...
   }
 
   log("checking for updates...");
-  // eval is because "update-notifier" is an ESM package and we're CommonJS.
-  const updateNotifier = await eval('import("update-notifier")');
-  const notifier = updateNotifier.default({
-    updateCheckInterval: DEFAULT_UPDATE_RATE,
-    pkg: PJSON,
+  // eval is because "tiny-updater" is an ESM package and we're CommonJS.
+  const updateNotifier = await eval('import("tiny-updater")');
```
skyrpex, line 16: Not related to this PR but can't we use `await import()` even in CommonJS? 🤔

MarkMcCulloh: Yessir, but tsc converts dynamic imports into a `require`-based promise when you are compiling to cjs. the eval bypasses that and forced us to use `import`

I can expand the comment to be more clear.

MarkMcCulloh: If we run the cli through an actual bundler we probably don't need this, but I had some issues with that so I kept things mostly as-is

skyrpex: Thanks for the info! Much better with that comment of top of it :)

skyrpex (approved):
LGTM

PR #3368, opened 2023-07-12
feat(sdk): add utilities for custom bind methods

Chriscbr:
This PR introduces the capability to write Wing classes with `_bind` methods that specify custom permissions for resources. This addresses a key limitation that has prevented classes like those found in the standard library (cloud.Bucket, cloud.Counter, etc.) from being written in Wing.

Writing `_bind` methods that add permissions hasn't been possible before because the `_bind` method only provides a `std.IInflightHost` as an argument, while in practice it will be provided a concrete class, like the `sim` target's `cloud.Function` implementation or the `tf-aws` target's `cloud.Function` implementation. To narrow the type, the SDK source code currently uses the JS `instanceof` operator (see [here](https://github.com/winglang/wing/blob/d7769ada65a6f3a917fd54843d4690280d421e11/libs/wingsdk/src/target-tf-aws/bucket.ts#L113-L116)) but this isn't supported in Wing today.

To support this use case, I added a module called `aws` that exposes a bare minimum class `AwsFunction` that can safely cast a `std.IInflightHost` into a `aws.IAwsFunction`, which has methods like `addEnvironment` and `addIamPolicy` for customizing permissions. I've added an example to our test suite that demonstrates how it's used.

```js
class DynamoTable {
  // ...
  _bind(host: std.IInflightHost, ops: Array<str>) {
    if let host = aws.AwsFunction.from(host) {
      if ops.contains("putItem") {
        host.addIamPolicy(aws.PolicyStatement {
          actions: ["dynamodb:PutItem"],
          resources: [this.table.arn],
          effect: aws.Effect.ALLOW,
        });
      }
    }
  }
}
```

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/shared-aws/aws-function.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export interface IAwsFunction {
+  /**
+   * Add an environment variable to the function.
+   */
+  addEnvironment(key: string, value: string): void;
+  /**
+   * Add a policy statement to the function's IAM role.
+   */
+  addIamPolicy(policy: PolicyStatement): void;
```
Chriscbr, line 17:
```suggestion
  addPolicyStatement(policy: PolicyStatement): void;
```

libs/wingsdk/src/index.ts
```diff
@@ -4,6 +4,7 @@ export * as core from "./core";
 export * as ex from "./ex";
 export * as http from "./http";
 export * as math from "./math";
+export * as aws from "./shared-aws";
```
Chriscbr, line 7: Need to generate docs for this module

examples/tests/valid/dynamo.w
```diff
@@ -0,0 +1,94 @@
+bring "@cdktf/provider-aws" as tfaws;
```
Chriscbr, line 1: This test only works on AWS, so this test might need a `skip-sim` annotation (how do I do this?)

examples/tests/valid/dynamo.w
```diff
@@ -0,0 +1,94 @@
...
+          name: "Flavor",
+          type: "S",
+        },
+      ],
+    );
+    this.tableName = this.table.name;
+  }
+
+  _bind(host: std.IInflightHost, ops: Array<str>) {
+    if let host = aws.AwsFunction.from(host) {
```
Chriscbr, line 41: Alternative API naming (thoughts?)
```suggestion
    if let host = aws.Hosts.asAwsFunction(host) {
```

examples/tests/valid/dynamo.w
tsuf239:
```
/*\
skip: true
\*/
```

This works in Hangar, but not via the cli, I also thought about it today when I wanted to skip a test, I think we should think of a way to unify all of those skipping scenarios 🤔

examples/tests/valid/dynamo.w
```diff
@@ -0,0 +1,94 @@
...
+  table: tfaws.dynamodbTable.DynamodbTable;
+  tableName: str;
+  init() {
+    let target = util.env("WING_TARGET");
+    if target != "tf-aws" {
+      throw("Unsupported target: ${target} (expected 'tf-aws')");
+    }
+
+    this.table = new tfaws.dynamodbTable.DynamodbTable(
+      name: "Default",
```
eladb, line 27:
```suggestion
      name: this.node.addr,
```

examples/tests/valid/dynamo.w
eladb:
```
aws.Function.from(host)
```

Not sure we need the Aws prefix here. Ha?

examples/tests/valid/dynamo.w
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  extern "./dynamo.js" inflight _putItem(tableName: str, item: Json): void;
+
+  inflight putItem(item: Map<Attribute>) {
+    let json = this.itemToJson(item);
+    this._putItem(this.tableName, json);
+  }
+
+  inflight itemToJson(item: Map<Attribute>): Json {
```
eladb, line 59:
```suggestion
  inflight _itemToJson(item: Map<Attribute>): Json {
```

Signal that it's private

libs/wingsdk/src/shared-aws/aws-function.ts
```diff
@@ -0,0 +1,54 @@
...
+        },
+      };
+    }
+
+    if (host instanceof AwsCdkFunction) {
+      return {
+        addEnvironment(key: string, value: string): void {
+          host.addEnvironment(key, value);
+        },
+        addIamPolicy(policy: PolicyStatement): void {
```
eladb, line 46:
```suggestion
        addPolicyStatement(policy: PolicyStatement): void {
```

libs/wingsdk/src/shared-aws/aws-function.ts
```diff
@@ -0,0 +1,54 @@
...
+ * A helper class for working with AWS functions.
+ */
+export class AwsFunction {
+  /**
+   * If the inflight host is an AWS function, return a helper interface for
+   * working with it.
+   * @param host The inflight host.
+   */
+  public static from(host: IInflightHost): IAwsFunction | undefined {
+    if (host instanceof TfAwsFunction) {
```
eladb, line 30: Why not just return `host`?

Make TfAwsFunction and AwsCdkFunction implement IAwsFunction and just return them directly. No?

eladb (approved):

libs/wingsdk/src/shared-aws/aws-function.ts
Chriscbr: Yeah that would be a lot nicer - let me give it a try. I initially jumped over this solution because I thought I might run into some kind of circular import issue between namespaces, but I suspect this is just a case of me overthinking things :)

Chriscbr: Done

PR #3370, opened 2023-07-12
feat: `tryParse()` now accepts `str?`

eladb:
The use case:

```js
let res = http.get("/foo");
let body = Json.tryParse(res.body);
```

Since `res.body` is `str?` (which makes sense), it also makes sense that `tryParse` just return `nil` when takes `nil`.

Previously, I had to do something like this:

```js
Json.tryParse(res.body ?? "");
```

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/std/json.ts
```diff
@@ -109,12 +109,12 @@ export class Json {
   /**
    * Try to parse a string into a Json
    *
-   * @macro ((args) => { try { return JSON.parse(args); } catch (err) { return undefined; } })($args$)
+   * @macro ((args) => { try { return args && JSON.parse(args); } catch (err) { return undefined; } })($args$)
```
tsuf239, line 112: if you'll put here `false`, it'll return `false` instead of `nil`, the same with `0`, I think

eladb: Since the signature is `str?` it won't be possible to pass these values from wing.

tsuf239: and what about passing `""` ? it'll return `""`

eladb: Nice catch. It should return `undefined` in that case. Checking.

eladb: Fixed.

MarkMcCulloh (approved):

Issue #3373, opened 2023-07-12
cli analytics issues


ekeren:
Issues seen (order by importance) 
- [ ] duplicating events: when running `wing test` one time , you get two `wing_sim_test` events (BTW, when running `wing doc` is doesn't reproduce it and `wing compile` does)
- [ ] the repo is not taken from the file location, it is taken from the CWD
consider using something like ` { cwd: entrypoint.dirname } ` when calling execFile
- [ ] anonymous_repo_id is missing on our CI test maybe related to ☝ 
- [ ] can we also have a `cli_entrypoint`

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.15](https://github.com/winglang/wing/releases/tag/v0.24.15).

PR #3381, opened 2023-07-12
fix(cli): duplicate reporting, add entrypoint

hasanaburayyan:
This PR is to cleanup and enhance the cli analytics. The main points are: 
- I had duplicate reporting because of adding the export hook to the error hook in the wrong spot, that is now fixed so the error handler only calls export hook if exit code is 1.
- The Git collector was running queries from CWD when it needed to run them from the directory of the app entry point.
- Added `cli_entrypoint` data

Closes: https://github.com/winglang/wing/issues/3373

*Note:* Im opening this for review and adding do-not-merge label so I can have some time to think about adding more robust test coverage to the analytics. Since we want to get these changes in before code-freeze Id like to ask for reviews on the changes before I add tests.

TODO:
- add some more test coverage
## Checklist

- [x] Tests added (always)


hasanaburayyan:
cc @ekeren @staycoolcall911 I have manually tested these changes and it all looks good You can see in the image below that even though Im running my command from my home dir (the entrypoint shows the path im using) I still get the git metrics from the entrypoint dir.
image


Im opening the PR for review now but I still want to add some automated tests in the meantime before its merged.

ekeren (approved):

apps/wing/src/analytics/collect.ts
```diff
@@ -18,11 +18,16 @@ export async function collectCommandAnalytics(cmd: Command): Promise<string | un
   const osCollector = new OSCollector();
   const nodeCollector = new NodeCollector();
   const ciCollector = new CICollector();
-  const gitCollector = new GitCollector();
   const cliCollector = new CLICollector(cmd);
 
+  // If entrypoint to app is provided, we will give that to git collector to use for
+  // running queries against the git repo, otherwise we will use the current working directory
+  const gitCollector = new GitCollector({
+    appEntrypoint: cmd.args.length > 0 ? cmd.args[0] : '.'
```
MarkMcCulloh, line 26: Is there any concern that this data can possibly contain PII? e.g. an absolute path could have a name.

apps/wing/src/analytics/scripts/detached-export.ts
```diff
@@ -17,7 +26,7 @@ async function reportAnalytic() {
     throw new Error('No file analytic path provided');
   }
 
-  const analytics = new Analytics({ writeKey: 'sCqPF5xSscOjJdi5Tbkqu73vfF8zkZdw'});
+  const analytics = new Analytics({ writeKey: process.env.DEBUG === undefined ? segmentWriteKey : segmentDebugWriteKey});
```
MarkMcCulloh, line 29: nit: I think it makes sense to invert this, so it matches the other DEBUG checks

```suggestion
  const analytics = new Analytics({ writeKey: process.env.DEBUG ? segmentDebugWriteKey : segmentWriteKey});
```

apps/wing/src/analytics/collect.ts
ekeren: Name is considered PII, but not when it is the only thing that we collect 

> That means that something like a first name is not PII, because it’s likely that many people have that first name. However, a full name — first, middle, and last — is considered PII by many organizations. It’s much less likely that many people have that same full name.

We can easily use `path.basename` 

@hasanaburayyan has asked me about this before, and I didn't consider the PII aspect

hasanaburayyan: So Im curious @ekeren if someone for some strange reason had their absolute path as
<first_name>/<last_name>/<social_security_number>/<address>/repos/my-wing-app

we would technically be collecting PII right?

I mean I guess the same could be said if the entrypoint file was `hasan_aburayyan.w`

MarkMcCulloh: I think we should just avoid automatically collecting user-entered data whenever possible, hashing it for tracking whenever needed

hasanaburayyan: From the point of avoiding collecting user-entered data then that would mean we also should skip collecting what flags and parameters were passed. IE `wing compile -t tf-aws app.w --plugin some_path/plugin.js` could also result in pii. 

I mean we could collect what --options were used but ignore any values passed to them. It would still give a sense for usage I think

MarkMcCulloh: For options with known values (like -t) we can collect those known values, skipping unexpected ones that may have been accidentally entered

ekeren: We are going to deep here, we aren't collecting name deliberately. it might be part of a file, part of a path, part of plugin, etc...

If I go to any website and add a `?utm_campaign="eyal keren, <id>"` it doesn't mean that they are collecting PII 

No need to change the code, it is fine to collect the full path.

MarkMcCulloh: Is accidental PII not still PII? If we can avoid it then I think we should.

staycoolcall911: This should be a team decision, and I want to make sure we all get a chance to voice our opinions.
Let's remove that potentially-PII field from our analytics for now and discuss it tomorrow on team time. If the team decides, we can add it back in tomorrow.

@hasanaburayyan - please let's get that merged in after this minor change.

PR #3392, opened 2023-07-12
fix: labeling during self mutation not working for pull requests from forks

### LINKED ISSUE #3391: labeling during self mutation not working for pull requests from forks
### I tried this:

Since #3359 the self mutation workflow adds a `⚠️ pr/review-mutation` label to the pr.

### This happened:

This labeling did not happen during self mutation in #3366.

### I expected this:

For the self mutation workflow to label the pr like normal.

### Is there a workaround?

Manually add the label.

### Component

Other

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

Upon looking into it, while the branches from wing repo works normally, it seems all pull_request contexts from github context returns empty arrays in case of forks and hence the curl request is getting an improper url.

### Community Notes

WeepingClown13:
Labeling during self mutation was added as a cautionary method but seems like it is not working for PRs from forks. This is because github context does not return pull request data (at least in this case) and so there is no way to form the proper url for rest api call using just github context. Instead retrieve the PR number with a rest api call and use it. Also change the labeling to after mutation commit.
Modified the step to use script module instead of run as per @MarkMcCulloh's suggestion.
fixes #3391.
## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -52,16 +52,6 @@ jobs:
             exit 1
           fi
 
-      - name: Add label to block auto merge
-        if: github.event.workflow_run.event == 'pull_request' && steps.download-artifacts.outputs.found_artifact == 'true'
-        run: |
-          curl \
-            -X POST \
-            -H "Authorization: Bearer ${{ secrets.MUTATION_TOKEN }}" \
-            -H "Accept: application/vnd.github.v3+json" \
-            "https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.workflow_run.pull_requests[0].number }}/labels" \
-            -d '["⚠️ pr/review-mutation"]'
-
       - name: Disable Git Hooks
         if: steps.download-artifacts.outputs.found_artifact == 'true'
         run: |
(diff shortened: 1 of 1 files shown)


WeepingClown13:
While we are at it, just another thought. Currently the label is immediately added if diff is detected so that even if the workflow somehow fails at a later point the label still exists and the author can take caution. Should this be changed to be labeled after the mutation commit instead?

.github/workflows/mutation.yml
```diff
@@ -55,11 +55,14 @@ jobs:
       - name: Add label to block auto merge
         if: github.event.workflow_run.event == 'pull_request' && steps.download-artifacts.outputs.found_artifact == 'true'
         run: |
+          PR_NUMBER=$(curl -H "Authorization: Bearer ${{ secrets.MUTATION_TOKEN }}" \
+            "https://api.github.com/repos/${{ github.repository }}/pulls?head=${{ github.event.workflow_run.head_repository.full_name }}:${{ github.event.workflow_run.head_branch }}" \
+            --silent | jq -r '.[0].number')
```
MarkMcCulloh, line 60: This would allow shell injection by having a branch name that exploits this (https://securitylab.github.com/research/github-actions-untrusted-input/)

Maybe setup the URLs in an env first

.github/workflows/mutation.yml
```diff
@@ -55,11 +55,14 @@ jobs:
...
           curl \
             -X POST \
             -H "Authorization: Bearer ${{ secrets.MUTATION_TOKEN }}" \
             -H "Accept: application/vnd.github.v3+json" \
-            "https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.workflow_run.pull_requests[0].number }}/labels" \
+            "https://api.github.com/repos/${{ github.repository }}/issues/$PR_NUMBER/labels" \
```
MarkMcCulloh, line 65: WDYT about using https://github.com/actions/github-script instead of curls?

MarkMcCulloh (changes requested):
> Should this be changed to be labeled after the mutation commit instead?

👍 Makes sense to me

.github/workflows/mutation.yml
WeepingClown13: Truth be told, I just try avoid js whenever I can because of my super unfamiliarity 🤣

MarkMcCulloh (approved):

PR #3407, opened 2023-07-13
fix: mergify comment is incorrect if the pull request contains mergify config change

### LINKED ISSUE #3389: mergify comment is incorrect if the pull request contains mergify config change
### I tried this:

Added a pr that contained a change to mergify configuration in `.mergify.yml`.

### This happened:

Got this comment from mergify:
image

### I expected this:

Mergify is not currently configured to auto merge PRs if the PR contains changes to the very config file of mergify itself. So the message should not say the PR will be merged automatically but indicate the need for manual merge instead.

### Is there a workaround?

No.

### Component

Contributor Experience

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

There are two ways to fix this, the simplest one would be to allow auto merge on mergify config change. The other one is rather simple too except it'll require a conditional comment config which will result in a few more lines of config compared to the single line change in the former method.

### Community Notes

WeepingClown13:
Mergify is not configured to auto merge if the PR contains mergify config changes, however the message that is commented is still that it'll be auto merged. This PR changes the response comment to appear conditionally and adds some explicit conditions to make sure that mergify config change is given the necessary attention.
fixes #3389.
## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,6 +2,7 @@
 queue_rules:
   - name: default
     conditions:
+      - files=.mergify.yml
       - status-success=Validate PR title
       - or:
         - status-success=Quality Gate
@@ -19,7 +20,8 @@ pull_request_rules:
           {{ title }} (#{{ number }})
           {{ body }}
     conditions:
-      - -title~=(WIP|wip)
+      - -files=.mergify.yml
+      - -title~=(?i)wip
       - -label~=(🚧 pr/do-not-merge|⚠️ pr/review-mutation)
       - -merged
       - -closed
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -19,7 +20,8 @@ pull_request_rules:
           {{ title }} (#{{ number }})
           {{ body }}
     conditions:
-      - -title~=(WIP|wip)
+      - -files=.mergify.yml
+      - -title~=(?i)wip
```
MarkMcCulloh, line 24: 💯

.mergify.yml
```diff
@@ -39,6 +41,35 @@ pull_request_rules:
...
+        - base=dev
+    actions:
+      comment:
+        message: Thank you for contributing! Your pull request contains megify configuration changes and needs manual merge from a maintainer (be sure to [allow changes to be pushed to your fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).
+      label:
+        add:
+          - ⚠️ mergify/review-config
+      request_reviews:
+        teams:
+          - "@winglang/maintainers"
```
MarkMcCulloh, line 71: Due to our current CODEOWNERS, they will always be reviewers on a PR. Maybe add it to the comment body so it's an `@mention` instead.

MarkMcCulloh (approved):

.mergify.yml
WeepingClown13: Yes but it is right after the pr is opened, right? This one is for when all conditions are satisfied and it is ready to merge. Add a label and then notify the reviewers. Doesn't that make sense?

MarkMcCulloh: Sure, I was just thinking an `@mention` would be more noticeable (e.g. I get push notifications on my phone for mentions, but not requested reviews).

I'm totally fine with just a requested review though 👍

WeepingClown13: Nah actually I thought about this and my original plan was to mention you specifically (since you work with the mergify configs most) and then I found there's an action to request reviews in mergify and went with that. This is because I didn't know we could mention teams itself with @mention. Just checked and found this is possible so definitely the more straightforward and noticeable method. I'll change it to that.

WeepingClown13: Seems like there are some quirks regarding team mentions so I will stick to the current config for now.

PR #3421, opened 2023-07-13
chore(vscode): monada to wing cloud, and remove alpha from ext name

ekeren:
Monada -> Wing Cloud
VSCode name ~Alpha~
## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/vscode-wing/.projenrc.ts
+++ b/apps/vscode-wing/.projenrc.ts
@@ -8,10 +8,10 @@ const VSCODE_BASE_VERSION = "1.70.0";
 const project = new TypeScriptAppProject({
   defaultReleaseBranch: "main",
   name: "vscode-wing",
-  authorName: "Monada",
-  authorEmail: "ping@monada.co",
+  authorName: "Wing Cloud",
+  authorEmail: "ping@wing.cloud",
   authorOrganization: true,
-  authorUrl: "https://monada.co",
+  authorUrl: "https://winglang.io",
   repository: "https://github.com/winglang/wing.git",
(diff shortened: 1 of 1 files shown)


apps/vscode-wing/.projenrc.ts
```diff
@@ -153,10 +153,10 @@ const contributes: VSCodeExtensionContributions = {
 };
 
 project.addFields({
-  publisher: "Monada",
+  publisher: "Wing Cloud",
```
MarkMcCulloh, line 156: This might be a problem, changing the publisher name is a process we'd have to look into

MarkMcCulloh: also I don't think it can have a space in it.

This publisher name becomes part of the id of the extension

MarkMcCulloh: So people would have to re-download the extension

ekeren: Good catch https://stackoverflow.com/questions/42745643/how-to-change-publisher-name-of-already-published-extension

I'll this change

ekeren, line 156:
```suggestion
  publisher: "Monada",
```

ekeren: https://github.com/winglang/wing/issues/3422

ekeren:
@MarkMcCulloh  added label `pr/backport` FYI

MarkMcCulloh (approved):

PR #3435, opened 2023-07-13
feat: adding adding uuidv4, nanoid and sha256 to utils

marciocadev:
Unfortunately, I had to limit nanoid in version 3 because starting from version 4, it no longer works with CommonJS.

## Checklist

- [x] Tests added (always)


libs/wingsdk/.projenrc.ts
```diff
@@ -98,6 +98,7 @@ const project = new cdk.JsiiProject({
     "tar",
     "express",
     "uuid",
+    "nanoid@^3.3.6",
```
MarkMcCulloh, line 101: You mentioned there is an issue with later versions of nanoid, maybe put a comment here to explain for future reference

libs/wingsdk/src/util/util.ts
```diff
@@ -97,6 +115,33 @@ export class Util {
...
+   * Generates a version 4 UUID.
+   */
+  public static uuidv4(): string {
+    return v4();
+  }
+
+  /**
+   * Generates a unique ID using the nanoid library.
+   * @param options - Optional options object for generating the ID.
+   */
```
MarkMcCulloh, line 136: Adding a link of some kind would be useful here for quick reference

```suggestion
  /**
   * Generates a unique ID using the nanoid library.
   # @link https://github.com/ai/nanoid
   * @param options - Optional options object for generating the ID.
   */
```

MarkMcCulloh (approved):

Issue #3405, opened 2023-07-13
Pre-release and analytics disclaimer is not printed to the user's standard output


staycoolcall911:
### I tried this:

(Followup bug related to issue #2974)

Installed wing CLI and ran `wing compile`.

### This happened:

No disclaimer (as described in #2974) is printed to stdout.

### I expected this:

The disclaimer to be printed on first run.

### Component

CLI

### Wing Version

0.24.17

### Node.js Version

18.15.0

### Platform(s)

MacOS

### Anything else?

I suspect this is not showing because the VSCode extension is running `wing lsp` and other `wing` commands behind the scenes. We should make sure that this disclaimer appears the first time when *a user* is running a `wing` command (and not the vscode extension, CI/CD, etc.). Perhaps a good direction for solution would be to verify that the command has `stdin`? 

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

hasanaburayyan:
@staycoolcall911 oh I suspect you are spot on about vscode running the lap. Can you check in your ~/.wing config to see what the displayed analytics disclaimer variable is set to?

staycoolcall911:
> @staycoolcall911 oh I suspect you are spot on about vscode running the lsp. Can you check in your ~/.wing config to see what the displayed analytics disclaimer variable is set to?

"disclaimerDisplayed":true

ekeren:
I verified that if you change this to false, it will present the message on the next command. So it should be an easy fix. Great work setting up this mechanism

hasanaburayyan:
Yea I think im going to add some sort of version or date identifier to the `disclaimerDisplayed` variable so that if we update the message we have a way to force it to be redisplayed.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.41](https://github.com/winglang/wing/releases/tag/v0.24.41).

PR #3439, opened 2023-07-13
fix(cli): display analytics disclaimer only in tty

hasanaburayyan:
This PR introduces 1 fix and 1 baby feature

1. The fix is to ensure we only display disclaimer when there is a stdin terminal. This stops things like vs-code extension from altering the `displayedDisclaimer` value.
2. The feature is to add a version to our config file for displayed disclaimer. This way if we need to update the disclaimer we can just tweak the `WING_DISCLAIMER_VERSION` variable and it will force display on the next run for the user.

Closes: https://github.com/winglang/wing/issues/3405

## Checklist

- [x] Tests added (always)


apps/wing/src/analytics/disclaimer.ts
```diff
@@ -1,5 +1,9 @@
 import chalk from "chalk";
-import { AnalyticsStorage } from "./storage";
+import { AnalyticsConfig, AnalyticsStorage } from "./storage";
+
+// simplest way I could think of to allow us to update the disclaimer
+// and force it to be displayed again
```
MarkMcCulloh, line 5: You're making it sound like this is a crazy thing to do haha

Versioning the disclaimer is a great idea.

MarkMcCulloh (approved):

apps/wing/src/analytics/disclaimer.ts
hasanaburayyan: Oh yea I agree, 😅 I was thinking more its just a hardcoded string we need to manually update.

PR #3444, opened 2023-07-13
fix(sdk): generate cloudwatch role for cloud.Api on tfaws target

Chriscbr:
I was trying to debug some issues with an app using a `cloud.Api` deployed to AWS when I discovered I couldn't enable logging because there wasn't a CloudWatch role associated with the API Gateway. I did some digging and found that in AWS CDK, this role is created for you automatically ([link](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.RestApi.html#cloudwatchrole)). For parity, I added code for specifying the role in the tf-aws target.

## Checklist

- [ ] Tests added (always)


Chriscbr:
Looking a bit closer I found this blurb from the AWS CDK docs:

> Note: there can only be a single apigateway.CfnAccount per AWS environment so if you create multiple RestApis with cloudWatchRole=true each new RestApi will overwrite the CfnAccount. It is recommended to set cloudWatchRole=false (the default behavior if @aws-cdk/aws-apigateway:disableCloudWatchRole is enabled) and only create a single CloudWatch role and account per environment.

It seems like creating an API Gateway account per API gateway resource is not supported, so I might just skip this for now

PR #3447, opened 2023-07-14
Refactor-to-turbo

skyrpex:
## Checklist

- [ ] Tests added (always)


skyrpex:
@MarkMcCulloh not really sure if the dev script for the vscode extension works the same... could you please check it out?

.clippy.toml
skyrpex, line 1: Is this empty file necessary? @MarkMcCulloh Haven't noticed a change.

.github/workflows/build.yml
```diff
@@ -358,9 +291,9 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - build
-      - e2e
-      - benchmarks
-      - unit-tests
+      # - e2e-test
```
skyrpex, line 294:
```suggestion
      - e2e-test
```

skyrpex:
Docs have been updated.

Ready for review.

skyrpex:
Will keep the [🚧 pr/do-not-merge](https://github.com/winglang/wing/labels/%F0%9F%9A%A7%20pr%2Fdo-not-merge) label until we decide if we want to do this change or not :)

.github/workflows/build.yml
```diff
@@ -74,73 +76,24 @@ jobs:
...
+          name: dist
+          path: dist/*.tgz
 
       - name: Check git diff of VSCode Extension
         id: git-diff-vscode-wing
         run: |
-          if git diff --quiet ${{ steps.setSHAs.outputs.base }} ${{ github.sha }} -- apps/vscode-wing; then
-            echo "::set-output name=diff::false"
+          if git diff --quiet origin/${{ github.base_ref }} ${{ github.sha }} -- apps/vscode-wing; then
+            echo "diff=false" >> "$GITHUB_OUTPUT"
```
MarkMcCulloh, line 100: This step intentionally used https://github.com/nrwl/nx-set-shas, regardless of the nx connection. That step got the latest successful SHA.
Either add https://github.com/nrwl/nx-set-shas back in or do it in a more turbo-y way.

apps/jsii-docgen/test/__fixtures__/libraries/construct-library/turbo.json
```diff
@@ -0,0 +1,13 @@
...
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "compile": {
+      "inputs": ["lib/**"],
+      "outputs": [".jsii"]
+    },
+    "package": {
+      "outputs": ["dist/**"]
+    }
```
MarkMcCulloh, line 11: I'm not sure we need to package this library at all

apps/jsii-docgen/test/__fixtures__/libraries/construct-library/.npmignore
MarkMcCulloh, line 1: If we don't need to package this (see below) then we can remove this file

apps/vscode-wing/scripts/dev.mjs
MarkMcCulloh, line 1: Could you change the vscode launch configuration to use this script at well?

MarkMcCulloh (commented):
I'm on board with this change, really excited about getting rid of nx.

For this PR to be ready, it needs a few things:

1. Many documentation updates to make sure the new turbo-based workflows are clear. Obviously removing any nx mentions. 
2. Make sure the devcontainer setup still works. It should, just worth a check though.
3. Uncomment anything in the build workflow that was removed (publishing/mutation/etc.)

Also, when those are ready, would be a good idea to get some feedback from the #dev slack channel to make sure contributors and aware and can review too.

Thanks man ❤️

.clippy.toml
Chriscbr: Not sure. Feel free to remove

turbo.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "lint": {
+      "inputs": ["src/**"]
+    },
+    "eslint": {
+      "inputs": ["src/**"]
+    },
+    "package": {
+      "dependsOn": ["compile", "post-compile"],
+      "cache": false
```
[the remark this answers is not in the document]
skyrpex: My reasoning is:

- The packaging task is fast (I think)
- The tgz file is generated in the root `dist/`, outside of the package dir, which would make the config harder to maintain

apps/jsii-docgen/test/__fixtures__/libraries/construct-library/turbo.json
skyrpex: If we don't need to package it, I'll just remove the package script from package.json and its turbo config.

apps/vscode-wing/scripts/dev.mjs
skyrpex: I'm not sure I can. The `"type": "extensionHost",` vscode launch script can't run custom commands AFAIK 🤔

.github/workflows/build.yml
```diff
@@ -198,70 +164,45 @@ jobs:
...
+
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
           cache: "pnpm"
           node-version: ${{ env.NODE_VERSION }}
 
       - name: Install Dependencies
-        run: pnpm install --ignore-scripts --frozen-lockfile --filter hangar --filter examples-valid --filter examples-invalid
+        run: pnpm install --frozen-lockfile --filter hangar --filter examples-valid --filter examples-invalid
```
MarkMcCulloh, line 201: scripts can be ignored here to avoid the extra wasi SDK setup

```suggestion
        run: pnpm install --ignore-scripts --frozen-lockfile --filter hangar --filter examples-valid --filter examples-invalid
```

docs/contributing/01-start-here/05-development.md
```diff
@@ -35,22 +35,22 @@ cd wing
...
-[Nx] commands in this document are structured as
+:::note Turbo Commands
+[Turbo] commands in this document are structured as
 
 ```sh
-pnpm nx <target> <project>
+pnpm turbo <task>
 # or
-pnpm nx <target> <project> -- <args>
+pnpm turbo <task> --filter=<project>
```
MarkMcCulloh, line 44: How do args work?

```suggestion
pnpm turbo <task> --filter=<project> -- <args>
```

or just

```suggestion
pnpm turbo <task> --filter=<project> <args>
```

package.json
```diff
@@ -1,23 +1,22 @@
...
+    "turbo": "^1.10.7"
   },
   "scripts": {
-    "build": "nx run-many --all --targets build,test,package",
-    "test": "nx run-many --all --target=test",
-    "package": "nx run-many --all --target=package",
-    "build:ci": "nx run-many --targets lint,build --all --configuration=release --output-style=stream --verbose",
-    "test:ci": "nx run-many --exclude=hangar --target=test --all --configuration=release --output-style=stream --verbose",
-    "package:ci": "nx run-many --target=package --all --configuration=release --output-style=stream --verbose",
+    "compile": "turbo compile --filter=!hangar",
```
MarkMcCulloh, line 11: Not sure we want a "compile" command at the root

package.json
```diff
@@ -1,23 +1,22 @@
...
+    "build": "turbo compile post-compile lint eslint test package --filter=!hangar",
+    "test": "turbo lint eslint test --filter=!hangar",
+    "package": "turbo package --filter=!hangar",
```
MarkMcCulloh, line 14: I believe @eladb wanted to have root commands available that handled running everything (including hangar).

package.json
```diff
@@ -1,23 +1,22 @@
...
+  "name": "@winglang/monorepo",
+  "private": true,
   "devDependencies": {
-    "@nrwl/devkit": "16.5.0",
     "bump-pack": "workspace:^",
     "changelogen": "^0.5.3",
-    "nx": "16.5.0",
-    "semver": "^7.3.8"
+    "semver": "^7.3.8",
+    "turbo": "^1.10.7"
```
MarkMcCulloh, line 8: Would be nice to update this to at least 1.10.10 before merging, looks like there is a useful fix for `workspace:` references: https://github.com/vercel/turbo/releases/tag/v1.10.10

package.json
skyrpex: What exact commands should we include? Currently:

- `build` executes everything
- `compile`, `test` and `package` execute the related subsets

package.json
skyrpex: Removed.

package.json
eladb: I think something like:

* `build` - full build: exactly what happens in CI


* `compile` - only compile
* `test` - only test
* `package` - only package

The rule of thumb is that `git clone && pnpm i && npm run build` should be almost the same as CI.

skyrpex: Ok, that's how it works now, with the difference that:

- Both test and package tasks will run, beforehand, any dependencies they have. For example, the `package` task requires the projects to be compiled first

It shouldn't really affect performance much since the cache is quite aggressive.

turbo.json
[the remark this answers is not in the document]
skyrpex: I suggest revisiting this in a different issue. I'd also like to improve the `post-compile` task (used to generate the API docs from one of the packages).

package.json
eladb: Sounds great!

MarkMcCulloh: To be clear, build runs everything and **shouldn't** filter out hangar.

In CI, we can't use this command because we split up build, unit testing, and e2e testing, but the package.json command should still do everything

MarkMcCulloh: I'm actually okay with filtering out hangar, it's just not what was previously requested

eladb: Totally fine if we break this down in CI, but the leading principle is that "build" should mimick CI, and also include hangar.

MarkMcCulloh (approved):

.github/workflows/build.yml
```diff
@@ -198,58 +162,33 @@ jobs:
...
-            echo "diff=true" >> $GITHUB_OUTPUT
-          fi
-      - name: Upload Diff
-        if: steps.diff.outputs.diff == 'true'
-        uses: actions/upload-artifact@v3
-        with:
-          name: build.diff
-          path: build.diff
+      - name: Test
+        run: pnpm test
```
MarkMcCulloh, line 170: hangar needs to be excluded here

libs/tree-sitter-wing/turbo.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "build:generate": {
+      "inputs": ["grammar.js"],
```
MarkMcCulloh, line 6:
```suggestion
      "inputs": ["grammar.js", "src/scanner.c"],
```

MarkMcCulloh (approved):

Issue #2986, opened 2023-06-20
Support calling super methods


hasanaburayyan:
### Feature Spec

Wing now supports calling methods from inherited classes within derived classes with `super.` 
```js
class Student {
  name: str;
  major: str;
  init(name: str, major: str) {
    this.name = name;
    this.major = major;
  }
  
  toString(): str {
    return "Name: ${this.name}, Major: ${this.major}";
  }
}

class PaidStudent {
  hrlyWage: num;
  
  init(name: str, major: str, hrlyWage: num) {
    super(name, major);
    this.hrlyWage = hrlyWage;
  }

  // Overridden method that calls method on super
  toString(): str {
    return "${super.toString()}, hrlyWage: ${this.hrlyWage}";  
  }
}
```

### Use Cases

Allows calling of inherited methods

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

eladb:
@staycoolcall911 this needs to be p1. We can't really extend anything without this.

hasanaburayyan:
@eladb I think it might fall into p1 territory as well. Right now you can still access methods on the `super` class with `this.someMethod()` but this is troublesome for overriding methods. It would require you to do something like

```js

class A {
	init() {}
	toString(): str {
		return "parent";
     }
}

class B extends A {
   init() { super() }

  toStringNotOverriden() {
    return this.toString() + " child";
   }
}

staycoolcall911:
Changed to p1
@eladb , @hasanaburayyan

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.41](https://github.com/winglang/wing/releases/tag/v0.24.41).

PR #3468, opened 2023-07-16
feat(compiler): support calling `super` methods 

yoav-steinberg:
see #2986
This adds the ability to call the superclass's overridden methods from an extended class
```wing
class A { do() {}  }
class B extends A { 
  do() { 
    super.do(); // call A's "do" 
  }
}
```  

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -11,15 +11,16 @@ use crate::{
 	ast::{
 		ArgList, BinaryOperator, Class as AstClass, Expr, ExprKind, FunctionBody, FunctionDefinition,
 		InterpolatedStringPart, Literal, Phase, Reference, Scope, Stmt, StmtKind, Symbol, TypeAnnotationKind,
-		UnaryOperator, UserDefinedType,
+		UnaryOperator, UserDefinedType, CalleeKind,
 	},
 	comp_ctx::{CompilationContext, CompilationPhase},
 	dbg_panic, debug,
 	diagnostic::{report_diagnostic, Diagnostic, WingSpan},
 	files::Files,
 	type_check::{
-		lifts::Lifts, symbol_env::SymbolEnv, ClassLike, Type, TypeRef, Types, VariableKind, CLASS_INFLIGHT_INIT_NAME,
+		lifts::Lifts, symbol_env::SymbolEnv, ClassLike, Type, TypeRef, Types, VariableKind, CLASS_INFLIGHT_INIT_NAME, resolve_super_method,
 	},
+	visit_context::VisitContext,
 	MACRO_REPLACE_ARGS, MACRO_REPLACE_ARGS_TEXT, MACRO_REPLACE_SELF, WINGSDK_ASSEMBLY_NAME, WINGSDK_RESOURCE,
 	WINGSDK_STD_MODULE,
 };
@@ -44,13 +45,13 @@ const ROOT_CLASS: &str = "$Root";
 const JS_CONSTRUCTOR: &str = "constructor";
 
 pub struct JSifyContext<'a> {
-	pub in_json: bool,
 	/// The current execution phase of the AST traversal.
 	/// The root of any Wing app starts with the preflight phase, and
 	/// the `inflight` keyword specifies scopes that are inflight.
-	pub phase: Phase,
 	pub files: &'a mut Files,
 	pub lifts: Option<&'a Lifts>,
+
+	pub visit_ctx: &'a mut VisitContext,
 }
 
 pub struct JSifier<'a> {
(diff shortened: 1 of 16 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -4249,6 +4262,56 @@ pub fn resolve_user_defined_type(
...
+		} else {
+			Err(TypeError {
+				message: format!("Cannot call super method because class {} has no parent", type_),
+				span: method.span.clone(),
+			})
+		}
+	} else {
+		Err(TypeError {
+			message:
+				"Cannot call super method because we're not in an instance method (\"super\" is unknown in this context)"
```
MarkMcCulloh, line 4308: overly verbose imo

```suggestion
				"Cannot call super method because we're not in an instance method"
```

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1862,6 +1862,56 @@ error: Cannot instantiate type \\"BucketProps\\" because it is a struct and not
...
+Test Files 1 failed (1)
+Duration <DURATION>"
+`;
+
+exports[`super_call.w 1`] = `
+"error: Cannot call super method because we're not in an instance method (\\"super\\" is unknown in this context)
+   --> ../../../examples/tests/invalid/super_call.w:32:7
+   |
+32 | super.do();
+   |       ^^ Cannot call super method because we're not in an instance method (\\"super\\" is unknown in this context)
```
Chriscbr, line 1875: Is it possible to simplify this error message to something like:
```suggestion
   |       ^^ Cannot call super method inside of a static method
```

Chriscbr, line 1875: Can we add `super` as a reserved keyword?

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1862,6 +1862,56 @@ error: Cannot instantiate type \\"BucketProps\\" because it is a struct and not
...
+   |
+20 |     super.child_method();
+   |           ^^^^^^^^^^^^ super class \\"A\\" does not have a method named \\"child_method\\"
+
+
+error: Cannot call super method because we're not in an instance method (\\"super\\" is unknown in this context)
+   --> ../../../examples/tests/invalid/super_call.w:26:11
+   |
+26 |     super.method();
+   |           ^^^^^^ Cannot call super method because we're not in an instance method (\\"super\\" is unknown in this context)
```
Chriscbr, line 1903: suggestion
```suggestion
   |           ^^^^^^ "super" can only be used inside of classes
```

libs/wingc/src/jsify.rs
```diff
@@ -390,17 +404,24 @@ impl<'a> JSifier<'a> {
...
+								_ => expr_string,
 							}
-
-							_ => expr_string,
+							CalleeKind::SuperCall{..} =>
+								// Note: in case of a $self$ macro override of a super call there's no clear definition of what $self$ should be,
+								// "this" is a decent option because it'll refer to the object where "super" was used, but depending on how 
+								// $self$ is used in the macro it might lead to unexpected results if $self$.some_method() is called and is
+								// defined differently in the parent class of "this".
+								"this".to_string(),
```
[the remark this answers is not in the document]
yoav-steinberg: We wanna figure out what `$self$` needs to be in our macro system. Imagine the case of `Array`'s `indexOf`:
the macro is `$self$.indexOf($args$)` so it'll translate to `super.indexOf(args...)`  which is good, but in case of `copy`
the macro is `Object.freeze([...($self$)])` so it'll translate to `Object.freeze([...(super)])` which is a JS error.

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: it's already there (https://github.com/winglang/wing/blob/ce24a6a4df52f1e4bff80c3363ab00105abbefb1/libs/wingc/src/parser.rs#L95). Do you mean something else?

yoav-steinberg: There's also the case of calling a `super` method from the global (top level) context. Anyway I'll shorten it :+1:

Chriscbr: Ok cool 👍

libs/wingc/src/type_check.rs
```diff
@@ -1746,22 +1742,39 @@ impl<'a> TypeChecker<'a> {
...
+				if let CalleeKind::Expr(call_expr) = callee {
+					if let ExprKind::Reference(Reference::Identifier(ident)) = &call_expr.kind {
+						if ident.name == "wingc_env" {
+							println!("[symbol environment at {}]", exp.span().to_string());
+							println!("{}", env.to_string());
+						}
+					}
+				}
+
+				if let CalleeKind::Expr(call_expr) = callee {
```
Chriscbr, line 1754: ?
```suggestion
				if let CalleeKind::SuperCall(call_expr) = callee {
```

yoav-steinberg: this was a piece of redundant code

PR #3503, opened 2023-07-17
fix(compiler): (performance) duplicate JSII deps are loaded more than once   

yoav-steinberg:
This affected performance of JSII loading.
Profiled using:
`cargo flamegraph --profile flamegraph --freq 20000 --example compile -- ../../examples/tests/valid/hello.w`
Results show that wingii's  `load_module` previously took 40% of the compilation time and now is down to 30%.

This PR also adds a profile to our `Cargo.toml` that's more compatible with `cargo flamegraph` (see the command above).

## Checklist

- [ ] Tests added (always)

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Monada Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.

### DIFF
--- a/libs/wingii/src/util.rs
+++ b/libs/wingii/src/util.rs
@@ -3,15 +3,16 @@ extern crate serde_json;
 
 pub mod package_json {
 	use std::{
+		collections::HashSet,
 		fs,
 		path::{Path, PathBuf},
 	};
 
 	use crate::node_resolve::{is_path_dependency, resolve_from};
 
-	pub fn dependencies_of(package_json: &serde_json::Value) -> Vec<String> {
+	pub fn dependencies_of(package_json: &serde_json::Value) -> HashSet<String> {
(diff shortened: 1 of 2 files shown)


Chriscbr (approved):

libs/wingii/src/util.rs
```diff
@@ -3,15 +3,16 @@ extern crate serde_json;
...
 	use std::{
+		collections::HashSet,
 		fs,
 		path::{Path, PathBuf},
 	};
 
 	use crate::node_resolve::{is_path_dependency, resolve_from};
 
-	pub fn dependencies_of(package_json: &serde_json::Value) -> Vec<String> {
+	pub fn dependencies_of(package_json: &serde_json::Value) -> HashSet<String> {
```
eladb, line 13: BTW, what happens if two separate JSII modules have the same transitive dependency? Are we loading the transitive once or multiple times?

Issue #3511, opened 2023-07-17
Cannot reuse dependency that is used elsewhere (diamond dependency problem)


Chriscbr:
### I tried this:

A project where module A depends on module B and module C, and B and C both depend on module D, so that D is used twice. Example:

```js
// file1.w
bring "./file2.w" as file2;
bring "./file3.w" as file3;

class A {
  b: file2.B;
  c: file3.C;
  init() {
    this.b = new file2.B();
    this.c = new file3.C();
  }
}


// file2.w
bring "./file4.w" as file4;

class B {
  d: file4.D;
  init() {
    new file4.D();
  }
}

// file3.w
bring "./file4.w" as file4;

class C {
  d: file4.D;
  init() {
    new file4.D();
  }
}

// file4.w
class D {}
```

### This happened:

An error occurred during compiling
```
error: Compiler bug (panicked at 'Failed adding source file to parser: Cannot add two files with the same name: file4.w', libs/wingc/src/parser.rs:169:17) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
  --> file3.w:1:1
  |
1 | bring "./file4.w" as file4;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Compiler bug (panicked at 'Failed adding source file to parser: Cannot add two files with the same name: file4.w', libs/wingc/src/parser.rs:169:17) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
```

### I expected this:

The example compiles safely.

### Component

Compiler

### Anything else?

Inside preflight.js, we need to only parse `file4.w` once and generate the code for `file4.w` only once. One approach that could simplify things is to generate separate `preflight.js` files for each module (file1, file2, file3, and file4).

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.53](https://github.com/winglang/wing/releases/tag/v0.24.53).

Chriscbr:
Reopening - the fix was reverted in Wing 0.24.55 due to other breakages it caused in the language server.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.81](https://github.com/winglang/wing/releases/tag/v0.24.81).

Issue #3509, opened 2023-07-17
Bad error message on bring/import cycles


Chriscbr:
### I tried this:

```
// file1.w
bring "./file2.w" as foo;
```

```
// file2.w
bring "./file1.w" as foo;
```

### This happened:

A confusing error message:
```
error: Compiler bug (panicked at 'Failed adding source file to parser: Cannot add two files with the same name: file1.w', libs/wingc/src/parser.rs:169:17) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
  --> file2.w:1:1
  |
1 | bring "./file1.w" as foo;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Compiler bug (panicked at 'Failed adding source file to parser: Cannot add two files with the same name: test_bucket.w', libs/wingc/src/parser.rs:169:17) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
```

### I expected this:

An error message saying that an import cycle was detected.

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.53](https://github.com/winglang/wing/releases/tag/v0.24.53).

Chriscbr:
Reopening - the fix was reverted in Wing 0.24.55 due to other breakages it caused in the language server.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.81](https://github.com/winglang/wing/releases/tag/v0.24.81).

PR #3512, opened 2023-07-17
feat(compiler): multi-file Wing applications

### LINKED ISSUE #476: Module System - Imports (bring): local .w files

This was done, then reverted - related to #401.

Chriscbr:
Introduce a basic form of multi-file Wing projects through the `bring` keyword. To use classes, interfaces, enums, or structs from another Wing file, bring a file with a path that ends in `.w`, and assign it to an alias:

```js
bring "./other-file.w" as shared;

new shared.Store();
```

This initial implementation follows a (naive) strategy of recursively instantiating the parser to create a large AST. This has some limitations -- for example it's not possible for two Wing files to reuse the same third Wing file -- but this still gives users a lot more refactoring power for building Wing apps. #3511 is the main follow-up issue for addressing this.

Closes #476

Follow up:
- https://github.com/winglang/wing/issues/3509
- https://github.com/winglang/wing/issues/3511

## Checklist

- [x] Tests added (always)

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Monada Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.


Chriscbr:
@MarkMcCulloh raised an important issue which is that before shipping this, we should aim to have a solution which prevents imports from having side effects. For example, with this PR the following code would be valid:

```js
// file1.w
bring "./file2.w" as file2;

// file2.w
bring cloud;

let global_bucket = new cloud.Bucket();
class B {
  // uses global_bucket
}
```

However, notice that bringing file2.w into the first file has the side effect of creating a bucket in the user's application, even though no classes from that module actually gets constructed (such as B). In some past discussions it's been brought up that only the entrypoint Wing file should have an implicit scope, so I'd like to see if there's a way to incorporate that here.

eladb:
> prevents imports from having side effects. For example, with this PR the following code would be valid.

Totally. We discussed this a few times. Non-entrypoint files should not allow any root-level statements, only type declarations (`class`, `enum`, `interface` and `struct`).

In the future (let's open a tracking issue) we want to support constants as root level statements in non-entrypoint files - `let` statements with an immutable data type (primitives or immutable collections).

libs/wingc/src/fold.rs
```diff
@@ -87,6 +87,10 @@ where
 			module_name: f.fold_symbol(module_name),
 			identifier: identifier.map(|id| f.fold_symbol(id)),
 		},
+		StmtKind::Module { name, statements } => StmtKind::Module {
```
eladb, line 90: So I am assuming "library" will be used to represent libraries (we need a way to distinguish between module/namespace and a library as a reusable artifact).


@yoav-steinberg what do you think?

libs/wingc/src/jsify.rs
```diff
@@ -578,6 +571,27 @@ impl<'a> JSifier<'a> {
 					}
 				))
 			}
+			StmtKind::Module { name, statements } => {
+				let mut code = CodeMaker::default();
```
eladb, line 575: nice and elegant!

examples/tests/valid/bring_local.w
```diff
@@ -0,0 +1,26 @@
+bring "./store.w" as file1;
+
+// classes from other files can be used
```
eladb, line 3: Feels like we need more tests here:

1. bring modules from subdirectories (`bring "./foo/bar/baz.w" as baz`).
2. prevent cyclic references
3. empty modules

what else?

libs/wingc/src/jsify.rs
```diff
@@ -578,6 +571,27 @@ impl<'a> JSifier<'a> {
...
+				code.add_code(self.jsify_scope_body(statements, ctx));
+
+				let exports = get_public_symbols(statements);
+				code.line(format!(
+					"return {{ {} }};",
+					exports.iter().map(ToString::to_string).join(", ")
+				));
+
+				code.close("})();");
+				code
```
MarkMcCulloh, line 586: I mentioned this irl, but just want to make sure it's commented: I don't think inlining the module is sufficient and probably shouldn't wait for a followup change imo. 
- The modules's code will have access to the parent scope. Because we have type checking I'm not 100% sure this will cause issues, but I think there's a chance for sneaky bugs in the emitted js.
- (more important) This will result in duplicate js when the same module is imported in different places. Not only does this contribute to https://github.com/winglang/wing/issues/3511, but it's also very wasteful. It'll also cause issues if there are any `instanceof` checks of classes from modules.

breaking the module out to a new file via a require() instead makes a lot of sense

eladb: I also share this intuition but I think we can totally iterate on this going forward.

MarkMcCulloh: > I think we can totally iterate on this going forward.

I would agree if changing this were difficult, but I think switching to a require is a very simple diff. Just emit a `module.exports =` `jsify_scope_body()` to a new file then emit the `require` to the current file.

examples/tests/valid/bring_local.w
Chriscbr: Added tests for bringing modules from subdirectories and empty modules. I've added an issue for handling cyclic references nicely: https://github.com/winglang/wing/issues/3509

libs/wingc/src/fold.rs
Chriscbr: I'm not sure I see any functional difference between libraries and modules, as far as the type system / compiler is concerned. Both represent a collection of API elements (classes, structs, constants, etc.). My understanding is a library is just a module that a user has exported as an artifact for sharing and reuse.

libs/wingc/src/jsify.rs
Chriscbr: Right now I'm just hijacking the `module` AST node for the use case of multi-file Wing applications -- I'm planning to address these problems as part of the follow up work (handling import cycles etc.) where we won't have to combine all of the ASTs together. If people run into bugs before that, let's create issues to track them.

libs/wingc/src/fold.rs
eladb: A library, for example, has dependencies.

eladb (approved):

libs/wingc/src/fold.rs
Chriscbr: I see. Yeah, thinking ahead, I'm expecting we won't want to inject libraries in the AST - it makes more sense to store library/dependency information inside the type system. But we'll see.

eladb: Totally agree. My comment was about terminology. If we are cool with calling libraries "libraries", then we can use the term "module" to refer to a namespace.

libs/wingc/src/jsify.rs
MarkMcCulloh: To be clear, this comment was talking about the emitted JS rather the sharing of the AST. I also think that we need to create multiple ASTs for several reasons, but that's a non-trivial change. Emitting the JS separately should be easy and can be done before any AST-weirdness.

PR #3524, opened 2023-07-18
feat(cli): implementing `test --watch` for wing cli

### LINKED ISSUE #3198: wing test --watch
### Feature Spec

`wing test watch <file.w>` reruns tests when the wing source changes.

### Use Cases

DX

### Implementation Notes

Ideally this would watch other dependenat files as well (e.g. extern) but watching the wing file itself would be a good start

### Component

_No response_

### Community Notes

garysassano:
Closes #3198

### DIFF
--- a/apps/wing/package.json
+++ b/apps/wing/package.json
@@ -34,6 +34,7 @@
     "@winglang/compiler": "workspace:^",
     "@winglang/sdk": "workspace:^",
     "chalk": "^4.1.2",
+    "chokidar": "^3.5.3",
     "codespan-wasm": "0.4.0",
     "commander": "^10.0.0",
     "compare-versions": "^5.0.3",
--- a/apps/wing/src/cli.ts
+++ b/apps/wing/src/cli.ts
@@ -114,7 +114,7 @@ async function main() {
 
       watcher.on("change", async () => {
         log(`Wing source file changed, rerunning tests…`);
-        await import(`./commands/test`).then((m) => m["test"](cmd.args, cmd.opts()));
+        await import(`./commands/test`).then((m) => m["test"](cmd.args, cmd.optsWithGlobals()));
       });
 
       watcher.on("error", (error) => {
--- a/apps/wing/src/cli.ts
+++ b/apps/wing/src/cli.ts
@@ -9,7 +9,6 @@ import { collectCommandAnalytics } from "./analytics/collect";
 import { exportAnalytics } from "./analytics/export";
 import { optionallyDisplayDisclaimer } from "./analytics/disclaimer";
 import { currentPackage } from "./util";
-import { log } from "console";
 import * as chokidar from "chokidar";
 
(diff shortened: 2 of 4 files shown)


apps/wing/src/cli.ts
```diff
@@ -100,6 +102,27 @@ async function main() {
...
+    if (watch) {
+      const watcher = chokidar.watch(entrypoint, {
+        ignoreInitial: true,
+      });
+
+      log(`Starting to watch ${entrypoint}`);
+
+      watcher.on("change", async () => {
+        log(`Wing source file changed, rerunning tests…`);
+        await import(`./commands/test`).then((m) => m["test"](cmd.args, cmd.opts()));
```
MarkMcCulloh, line 117: I think we need to make sure the top level options are carried through, just in case (might prevent subtle bugs later)

```suggestion
        await import(`./commands/test`).then((m) => m["test"](cmd.args, cmd.optsWithGlobals()));
```

apps/wing/src/cli.ts
```diff
@@ -100,6 +102,27 @@ async function main() {
     }
   }
 
+  async function watchHook(cmd: Command) {
```
MarkMcCulloh, line 105: I think handling the watch should be inside the test command itself. It may require some refactoring, but it has several benefits. One of the important ones is that it prevents the cli from loading `chokidar` unless needed.

apps/wing/src/cli.ts
```diff
@@ -9,6 +9,8 @@ import { collectCommandAnalytics } from "./analytics/collect";
 import { exportAnalytics } from "./analytics/export";
 import { optionallyDisplayDisclaimer } from "./analytics/disclaimer";
 import { currentPackage } from "./util";
+import { log } from "console";
```
MarkMcCulloh, line 12: I think console.log is ok, no need to have this import

apps/wing/package.json
```diff
@@ -34,6 +34,7 @@
     "@winglang/compiler": "workspace:^",
     "@winglang/sdk": "workspace:^",
     "chalk": "^4.1.2",
+    "chokidar": "^3.5.3",
```
eladb, line 37: Is this the same watch library wing console used?

@skyrpex ?

We should be very careful to add a dependency to our cli. Let's make sure this is absolutely needed.

MarkMcCulloh: It's the same one, so it doesn't broaden our dependency closure

Chriscbr:
Hey @garysassano - sorry for the delayed review. 😅  I think it will be a bit longer while before we can ship this PR since the current implementation won't work for multi-file Wing projects. Currently we don't have an easy way to get this kind of tracking information (I've created an issue about it here: https://github.com/winglang/wing/issues/3730) so I think that will be needed before a watch mode can be added, otherwise users will be confused when they change files or add new files to their project and the tests do not get reloaded. I'll mark the PR as a draft for now to make it clear there is still some remaining work needed.

PR #3528, opened 2023-07-18
feat(sdk): adding base64 helper to utils

### LINKED ISSUE #3497: Base64 helper
### Feature Spec

Now it's possible to en/decode base64 data with Wing.

Example

```
bring util;

let foo = util.base64encode("foo")
log(foo) // -> "Zm9v"
log(util.base64decode(foo))  // -> "foo"
```

As an additional feature, [url safe encoding](https://en.wikipedia.org/wiki/Base64#URL_applications) is supported with `util.base64encode("foo", url: true)`

### Use Cases

- encoding / decoding payloads in urls, request / response bodies and headers

### Implementation Notes

not sure if `util` is the right place

### Component

SDK

### Community Notes

garysassano:
Closes #3497


apps/wing/src/util.ts
```diff
@@ -95,4 +96,22 @@ export function parseNumericString(text?: string) {
...
+export function base64Encode(stringToEncode: string, url?: boolean): string {
+  return url ? base64url(stringToEncode) : btoa(stringToEncode);
+}
+
+/**
+ * Converts a base64/base64url string to a binary string
+ */
+export function base64Decode(stringToDecode: string, url?: boolean): string {
+  return url ? base64url.decode(stringToDecode) : atob(stringToDecode);
+}
```
MarkMcCulloh, line 111: I don't think a library is needed. I didn't test this, but this is built in since node18 afaik

```suggestion
export function base64Encode(stringToEncode: string, url?: boolean): string {
	return Buffer.from(stringToEncode).toString(url ? "base64url" : "base64");
}

/**
 * Converts a base64/base64url string to a binary string
 */
export function base64Decode(stringToDecode: string, url?: boolean): string {
	return Buffer.from(stringToDecode, url ? "base64url" : "base64").toString("utf8")
}
```

MarkMcCulloh (changes requested):

apps/wing/src/util.ts
```diff
@@ -95,4 +95,22 @@ export function parseNumericString(text?: string) {
   return number;
 }
 
-export const currentPackage: { name: string, version: string, engines: { node: string } } = require("../package.json");
+/**
+ * Converts a binary string to a base64/base64url string
```
MarkMcCulloh, line 99: The docstring should have the `@params` and `@returns` parts written as well. 
imo the summary doesn't need to mention the base64/base64url distinction, that can be in the `@returns` part instead

libs/wingsdk/src/util/util.ts
```diff
@@ -78,6 +78,26 @@ export class Util {
     return process.env[name];
   }
 
+  /**
+   * Converts a string from UTF-8 to base64.
+   * @param name The name of the UTF-8 string to encode.
```
MarkMcCulloh, line 83: Missing the `url` param as well

libs/wingsdk/src/util/util.ts
```diff
@@ -78,6 +78,28 @@ export class Util {
...
+   * @param url Whether the string should be treated as an url.
```
MarkMcCulloh, line 84:
```suggestion
   * @param url If `true`, a URL-safe base64 is returned.
```

libs/wingsdk/src/util/util.ts
```diff
@@ -78,6 +78,28 @@ export class Util {
...
+   * @returns The base64 string.
+   */
+  public static base64Encode(stringToEncode: string, url?: boolean): string {
+    return Buffer.from(stringToEncode).toString(url ? "base64url" : "base64");
+  }
+
+  /**
+   * Converts a string from base64 to UTF-8.
+   * @param name The name of the base64 string to decode.
```
MarkMcCulloh, line 93:
```suggestion
   * @param stringToDecode base64 string to decode.
```

libs/wingsdk/src/util/util.ts
```diff
@@ -78,6 +78,28 @@ export class Util {
...
+   * @param url Whether the string should be treated as an url.
```
MarkMcCulloh, line 94:
```suggestion
   * @param url If `true`, the source is expected to be a URL-safe base64 string
```

MarkMcCulloh (approved):

PR #3540, opened 2023-07-20
feat(console): zoom to fit when map view is out of bounds

### LINKED ISSUE #3537: application map view in console should always be visible
### I tried this:

open the console and agreed to the terms of use

### This happened:

the map of my application was hidden somewhere out side of my view area

### I expected this:

the map to be centered and properly zoomed in.
also I expect that a use won't be able to drag the map out of view

### Is there a workaround?

click on the auto center button... but users won't have an idea that it is possible.

### Component

Wing Console

### Wing Version

latest

### Node.js Version

_No response_

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

skyrpex:
If the map view is out of bounds, it will zoom to fit automatically.

Uses the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

Fixes #3537.


skyrpex:
As @ainvoner mentioned, the 35% threshold triggers when zooming in, too.

ainvoner (approved):

PR #3567, opened 2023-07-22
fix(compiler): improve JSII loading speed via binary format cache

yoav-steinberg:
This PR caches the `.jsii` manifests in `/tmp/.wing/jsii_manifest_cache/` in a binary format which is quicker to deserialize. It compares the original files to the cached ones using canonical path, file size and modification time. If no match is found it loads the original file and attempts to cache the file, if a match is found it'll load the JSII manifest from the cache.

In addition I identified some extra memory copying (cloning) of the imported JSII assemblies which was redundant. Eliminating this also added some performance.

A few notes:
1. I ended up using file metadata instead of the "fingerprint" field in the JSII manifest because doing so required unzipping gzipped .jsii files just to read the fingerprint field.
2. I didn't use bson or flexbuffer serde implementations because their performance is worse than plain serde_json(!!!). See: https://github.com/djkoloski/rust_serialization_benchmark.
3. ~~I ended up using rmp_serde (msgpack serde). There are faster ones which we might move to in the future but the problem is that their implementation is incomplete. The main issue being lack of support for "self-describing" formats where the deserializer needs to figure out the output based on the schema information in the file itself (which is mainly required for deserializing rust algebraic enums and also handle missing fields in structs).~~
4. ~~Even with `rmp_serde` I needed to remove the missing fields support from our serde definitions to get this working. rmp_serde actually does support missing fields but that hinders its performance (see `rmp_serde::encode::write` vs `rmp_serde::encode::write_named`).~~
5. I ended up using `bincode`, and specifically its non-serde serializer. I `bincode` as most fast serializer's doens't support the [`tag`](https://serde.rs/enum-representations.html#internally-tagged) field for enum serialization which is required to deserialize .jsii files using `serde_json`. By using `bincode`'s non-serde serializer we don't need to deal with the tag field. `bincode` seems like one of the faster serializers, we can consider some more options from [here](https://github.com/djkoloski/rust_serialization_benchmark) in the future.

**update**: I endded up using [**speedy**](https://github.com/koute/speedy/) which provided the best performance (see benchmarks)

## benchmarks
normal test:
`hyperfine -w 1 "cargo run --example compile -- testfile.w"`
empty cache test (replace FORMAT with speedy/bincode/bson based on the name of the cache file generated):
`hyperfine -w 1 -p "find ../.. | grep ".jsii.FORMAT" | xargs rm || true" "cargo run --example compile -- testfile.w"`

test | ~~rmp_serde no cache~~ | ~~rmp_serde cached~~ | main branch | bincode no cache | bincode cached | speedy no cache | speedy cached
---|---|---|---|---|---|---|---
hello.w | 136ms | 106ms | 105ms | 134ms | 100ms | 111ms | 99ms
bring_cdktf.w | 10.4s | 2.9s | 4.1s | 9.1 | 1.39s | 4.1s | 740ms
bring_awscdk.w | 3.9s | 1.2s | 1.8s | 3.42s | 734ms | 1.83s | 416ms
bring_projen.w | 263ms | 141ms | 161ms | 260ms | 118ms | 189ms | 111ms

## profiling
Looking a the flamegraphs it's clear that deserializing the JSII manifests is still taking the bulk of the compilation time when importing large ones (cdktf...). I think there's value in continuing this work by either using a faster deserializer than ~~`rmp_serde`~~ `bincode` or dropping JSII deserialization and implementing our own type-system serde per namespace and caching that.

## Checklist

- [x] Tests added (always)


yoav-steinberg:
@MarkMcCulloh, @eladb Do you guys think I should add some env var setting to skip generating the cache for faster CI/testing?

libs/wingii/src/lib.rs
```diff
@@ -48,20 +53,45 @@ pub mod spec {
 		}
 	}
 
+	fn try_load_from_cache(hash: &str) -> Option<Assembly> {
+		let file_path = format!("{CACHE_FILE_DIR}/{hash}.{CACHE_FILE_EXT}");
```
eladb, line 57: We should also support bundling the cached version of the jsii manifest into the library.

For example, if there's a file called `wing.types` (or something like that) inside the npm module (next to `.jsii`), then use it instead of trying to load `.jsii`. Then we can create this file during build and bundle it into `@winglang/sdk/wing.types`.

(we will also need some tool that can generate this `.types` file so we can use it in our build).

(not sure about the name)

yoav-steinberg: In practice the only .jsii files we have are wingsdk's and constructs, both of which load pretty quickly and also serialize to the cache pretty quickly so I don't think it'll make much of a difference and not worth the effort.
Pre-caching cdktf/awscdk makes sense, but they aren't part of our bundle, so I think it's out of this scope, and more related to adding support for binary manifests in .JSII (should I do a PR there?).

.gitignore
```diff
@@ -13,6 +13,9 @@ node_modules/
 # Generated wing output
 *.w.out/
 
+# Generated wingc .jsii binary cache
+*.jsii.bincode
```
MarkMcCulloh, line 17:
```suggestion
**/*.jsii.bincode
```

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -1022,18 +1024,16 @@ impl From<&Option<jsii::Docs>> for Docs {
 			return Docs::default()
 		};
 
-		let docs = docs.clone();
```
MarkMcCulloh, line 1025: Why remove the clone? I saw below that you removed the Clone derive on Docs, but why not keep it instead?

libs/wingii/src/lib.rs
```diff
@@ -179,7 +248,15 @@ pub mod type_system {
...
 			}
 
-			let asm = self.load_assembly(&assembly_file)?;
+			// Get the module version, this is used for fingerprinting the JSII manifest
+			// needed for comparing it to the JSII manifest cache
+			let module_version = if let Some(Value::String(ver_str)) = package.get("version") {
+				Some(ver_str.clone())
+			} else {
+				None
+			};
```
MarkMcCulloh, line 257: I think we're going to have an issue with a massive pile of `jsii.bincode` caches in the wingsdk folder in development.

I can't really think of a good way to do this other than disabling caching if the version of the jsii is `0.0.0`

libs/wingc/src/type_check/jsii_importer.rs
yoav-steinberg: The performance optimization was removing an unneeded clone of the entire assembly in `add_assembly`. To ensure we don't do any unneeded memory copies I figured just making sure the JSII structs aren't cloneable to begin with. This way we need to be explicit about what we copy out of these structs. Since we copy lots of data out of `Doc` then this might seem a bit tedious but still seemed worth it.

libs/wingii/src/lib.rs
yoav-steinberg: It also compares the modification date and file size of the .jsii in addition to the module version. So it should take care of cleaning the old ones when it encounters an update.

yoav-steinberg: also added a test for cleanups

libs/wingii/src/lib.rs
```diff
@@ -32,6 +34,7 @@ pub mod spec {
 
 	pub const SPEC_FILE_NAME: &str = ".jsii";
 	pub const REDIRECT_FIELD: &str = "jsii/file-redirect";
+	pub(crate) const CACHE_FILE_EXT: &str = ".jsii.speedy";
```
MarkMcCulloh, line 37: Cute 👀

libs/wingii/src/test.rs
```diff
@@ -14,54 +14,157 @@ mod tests {
...
 	}
 
 	#[test]
 	fn can_load_assembly_from_single_file_compressed() {
-		let assembly_path_pre = create_temp_assembly();
-		let assembly_path = assembly_path_pre.with_extension("jsii.gz");
+		let (name, assembly_path) = create_temp_gz_assembly();
+
+		let assembly = spec::load_assembly_from_file(&name, assembly_path.to_str().unwrap(), Some("gzip"), &None).unwrap();
+		assert_eq!(assembly.name, "jsii-test-dep"); // TODO: write a better test
```
MarkMcCulloh, line 48: I think the TODOs can be removed, you added some nice tests

tools/hangar/__snapshots__/test_corpus/valid/debug_env.w_test_sim.md
MarkMcCulloh, line 1: Any idea why this change happened? Generally we don't have ansi sequences in the snapshots.

MarkMcCulloh (approved):
Super nice perf numbers, looking forward to this change going in ❤️

tools/hangar/__snapshots__/test_corpus/valid/debug_env.w_test_sim.md
yoav-steinberg: no idea, I'll investigate..

yoav-steinberg: It's a result of an upgraded colored crate in the cargo lockfile. It also exposed that we don't correctly pass the color argument to the compiler package since the breakup of the compiler and wing packages. Fixing.

Issue #3569, opened 2023-07-23
Making sure that when JSII libraries are imported we map Date to this new type


tsuf239:
### Feature Spec

Making sure that when JSII libraries are imported we map Date to this new type.
See https://github.com/winglang/wing/pull/2095/files#r1167668099 

https://github.com/winglang/wing/blob/23712eff9768576bdd852cb9b989a9cd44af014a/libs/wingc/src/type_check/jsii_importer.rs#L93

### Use Cases

Mapping `Date` properly to JSII libraries 

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

PR #3570, opened 2023-07-23
feat(sdk): adding `datetime` type

### LINKED ISSUE #2102: Missing `datetime` std type
See https://github.com/winglang/wing/issues/1392

Also, make sure that when JSII libraries are imported we map `Date` to this new type.
See https://github.com/winglang/wing/pull/2095/files#r1167668099

tsuf239:
## Description
closes #2102 
(besides adding the type mapping, opened a separate issue: https://github.com/winglang/wing/issues/3569)

a wing test added, checked both preflight and inflight both in tf-aws and sim.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/std/datetime.ts
+++ b/libs/wingsdk/src/std/datetime.ts
@@ -106,7 +106,9 @@ export class Datetime {
   }
 
   /**
-   *  returns a timestamp of non-leap seconds since epoch
+   * Retrun a timestamp of non-leap year seconds since epoch
+   * 
+   * @returns a number representing the current timestamp in seconds
    */
   get timestamp(): number {
     return this.timestampMs / 1000;
--- a/libs/wingsdk/src/std/datetime.ts
+++ b/libs/wingsdk/src/std/datetime.ts
@@ -65,7 +65,9 @@ export class Datetime {
   }
 
   /**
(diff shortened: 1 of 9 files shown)


hasanaburayyan (commented):
Looking good so far, I added some nits to the comment docstrings I didnt do all of them but if you could take a look.

libs/wingsdk/src/std/datetime.ts
```diff
@@ -0,0 +1,194 @@
...
+   * returns a Datetime represents the same date in utc
+   */
+  toUtc(): Datetime {
+    return new Datetime(new Date(this.timestampMs));
+  }
+
+  /**
+   * returns ISO-8601 string
+   */
+  toIso(): string {
```
Chriscbr, line 191: Let's aim for consistency in our public APIs and set a model for others to follow. Please use the public keyword and start docstrings with a capital letter for all methods. (When the codebase follows conventions, the contributor experience is easier for newcomers)

tsuf239: No problem, good to know, can I add those two points to the docs somewhere? what will be the best place?

Chriscbr: Somewhere here would be good: https://www.winglang.io/contributing/start-here/pull_requests

Chriscbr (approved):

Issue #1894, opened 2023-03-23
Regex in std lib


staycoolcall911:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Wing introduces a new built-in `regex` type as part of the standard library (similar to `duration`, `string`, `array`, etc.).

### Use Cases

Make regular expressions easy to use in wing and reduce boilerplate code currently needed to use regex. Currently this is supported via the `extern` syntax - see an example as part of [story 15](https://docs.winglang.io/contributors/rfcs/stories/story-15).

### Implementation Notes

As other builtin types, this should be implemented in [wingsdk's std library](https://github.com/winglang/wing/tree/main/libs/wingsdk/src/std) as a thin wrapper over javascripts RegExp.

Please also make sure to update the reference docs (should be done automatically after you add the regex class to the std lib and run `npm run build` from the [wingsdk folder](https://github.com/winglang/wing/tree/main/libs/wingsdk/)).

### Component

SDK

Bharadwajshivam28:
@staycoolcall911 Hello sir i want to work on this issue

Bharadwajshivam28:
sir please elaborate this issue little more in simple words @staycoolcall911

revitalbarletz:
Hi @Bharadwajshivam28! I want to chat with you, how about joining our [slack](https://t.winglang.io/slack) channel?

Bharadwajshivam28:
> Hi @Bharadwajshivam28! I want to chat with you, how about joining our [slack](https://t.winglang.io/slack) channel? 

Ya sure

Bharadwajshivam28:
> Hi @Bharadwajshivam28! I want to chat with you, how about joining our [slack](https://t.winglang.io/slack) channel? 

I joined the channel

revitalbarletz:
Since there is no spec for Regex, I am removing the 'good first issue' tag.

ekeren:
I'll work on the spec first, and then we can turn this into "good first issue"

Bharadwajshivam28:
> Since there is no spec for Regex, I am removing the 'good first issue' tag.

ok

Bharadwajshivam28:
> I'll work on the spec first, and then we can turn this into "good first issue"

ok

olivernybroe:
@Bharadwajshivam28 are you currently working on this?

staycoolcall911:
[...]

PR #3592, opened 2023-07-24
feat(sdk): adding `string.replace()`

### LINKED ISSUE #2988: SDK String replace
### Feature Spec

Replacing within strings is now possible.

```
let repoName = "foo/bar";
let repoNameNormalized = repoName.replace("/", "-") // or even /[^a-zA-Z0-9]/g, "-"
```

### Use Cases

String manipulation 

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

garysassano:
Closes #2988


libs/wingsdk/src/std/string.ts
```diff
@@ -150,6 +150,24 @@ export class String {
...
+  /**
+   * Replaces occurrences of a substring within a string.
+   *
+   * @macro $self$.replace($args$)
+   *
+   * @param searchString The substring to search for.
+   * @param replaceString The replacement substring.
+   * @returns The modified string after replacement.
+   */
+  public stringReplace(
```
tsuf239, line 162: shouldn't it be called just `replace`?

tsuf239 (commented):
Hi Gary, thanks a lot! the code looks great!
can you please:
- add tests to this file: `examples/tests/sdk_tests/std/string.w`, testing both replacement of a string and a regex
- running `pnpm build` in your own machine, so it'll generate the snapshots and docs :)

libs/wingsdk/src/std/string.ts
garysassano: Yeah, clearly an oversight.

tsuf239: thanks! 

now just those two: 

> * add tests to this file: examples/tests/sdk_tests/std/string.w, testing both replacement of a string and a regex 
> * running pnpm build in your own machine, so it'll generate the snapshots and docs :)

garysassano: Do you know why it fails the build?

tsuf239: I'm not sure Wing supports regex yet (Does it @Chriscbr?) 
if not, let's open an issue and add a comment to this method
maybe the easier thing to do is to accept only a string and put it into `new RegExp(...);` (maybe check if the string is  a regex first, otherwise, expressions like `string.replace("^", "any")` won't work quite well...)

garysassano: In fact, I noticed that there's an open issue to add `Regex` type to std library: #1894

I therefore proceeded to create another issue to add the feature to `string.replace()` in the future: #3644

garysassano: In fact, I noticed that there's an open issue to add `Regex` type to std library: #1894

I therefore proceeded to create another issue to add the feature to `String.replace()` in the future: #3644

tsuf239 (approved):

PR #3594, opened 2023-07-24
chore: simplify url-shortener example

### LINKED ISSUE #2823: possible race condition in url shortener example
https://github.com/winglang/wing/blob/e4a8477a317c7fc34ad5051d0d035638438d66ab/examples/proposed/url-shortener.w#L37-L56

In the above code, if 2 callers are entering the function with the same URL and cross all the way to line 46. 
The result data structure will be:
```
url -> id2 
id1 -> url
id2 -> url
```

and although the functions returned `id1`, it is not a valid id (because there is no URL that points to it)

Also, I am not sure I understand why we have ` this.idLookup.put(id, url);` in line 42

Chriscbr:
The URL shortener example as its written has a race condition because it's trying to ensure that every URL gets a unique shortened ID. If we relax this constraint (which is reasonable in practice), then the race condition is avoided and we can simplify the implementation.

Closes #2823

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/proposed/url-shortener.w
+++ b/examples/proposed/url-shortener.w
@@ -2,69 +2,33 @@ bring cloud;
 bring math;
 bring http;
 
-let ALPHANUMERIC_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
-
 class UrlShortener {
-  urlLookup: cloud.Bucket;
-  idLookup: cloud.Bucket;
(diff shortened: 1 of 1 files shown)


ekeren:
Love it

ekeren (approved):

eladb (commented):
@Chriscbr @skorfmann I've noticed the examples under `proposed` are not being compiled and oftentimes get stale.

Perhaps we should move them all into the examples repository?

PR #3595, opened 2023-07-24
feat(cli): add analytics opt out message

hasanaburayyan:
Adds a confirmation message when users opt out of analytics.

## Checklist

- [ ] Tests added (always)


staycoolcall911 (approved):
Thanks for following up on that point!

MarkMcCulloh:
imo I don't think we should have this message. Not only because it breaks in CI (checking for 0.0.0 isn't sufficient) but because I think it would annoy users who opt out since it pops up every time

hasanaburayyan:
> imo I don't think we should have this message. Not only because it breaks in CI (checking for 0.0.0 isn't sufficient) but because I think it would annoy users who opt out since it pops up every time

As far as the annoyance personally I think if I want to explicitly opt-out I would prefer some sort of confirmation. This way there’s some sort of indication if I mistype the environment variable or something along those lines.

as for why checking for 0.0.0 isn’t sufficient I’m not sure I understand what happened in the CI job yet

MarkMcCulloh:
In CI the version is set to the soon-to-be-published version. This does not happen in PRs, only in main.

> As far as the annoyance personally I think if I want to explicitly opt-out I would prefer some sort of confirmation. This way there’s some sort of indication if I mistype the environment variable or something along those lines.

I get the idea, but are there other CLI tools that do this? I'm just not sure it's expected. If the person is conscious enough to disable telemetry then they probably understand that it's their responsibility to do it correctly. And once they do it correctly once, every other time it shows up it's basically noise.

hasanaburayyan:
> I get the idea, but are there other CLI tools that do this? I'm just not sure it's expected.

Im not sure how many CLIs do this but I know brew does ask you to re-enable analytics.

image


As well from my perspective I think we have been trying to opt for as much transparency with the analytics gathering as possible. I would feel different about displaying a message like "btw we are gonna collect stuff this run" because that is the normal behavior, but when it comes to opt out the user is asking for a deviation from standard process. Which to me makes me feel less bad about pestering them with additional output.

But maybe im totally off

ekeren:
Joining late to the party. Here are my thoughts. 

I like the transparency, and dislike the verbosity.

When looking at brew example, it is important to notice:
- Brew message only appear once, there are even 2 issue about this message re-appearing 
  - https://github.com/Homebrew/brew/issues/15718
  - https://github.com/Homebrew/brew/issues/15678

In particular:
> After seeing this message once and having influxanalyticsmessage = true being set in /opt/homebrew/.git/config along with analyticsmessage = true. It doesn't show up again. I realize I have analyticsdisabled = true on but I shouldn't be spammed every time I run brew up and have updated formulas.

There is a different here that brew is asking the user to re-enable it and we are not.

I **feel** that presenting this each time is actually nagging the customer that he should open it, which is not the author intention I understand, but maybe considered. . 

My suggestion:
1. Show it once (only when tty is avaiable)
2. Show it always on --debug option

PR #3597, opened 2023-07-25
chore(cli): use tsup instead of tsc

skyrpex:
[tsup](https://github.com/egoist/tsup) is used by the Console packages to compile TS to JS. Uses esbuild, but also can generate type definitions using tsc.

Additionally, it will bundle all imported dependencies except for packages listed as dependencies or peerDependencies. The fact that tsup does that, improves the Console DX by allowing to work with the `@wingconsole/server` package without starting an additional dev script.


skyrpex:
Depends on https://github.com/winglang/wing/issues/3601.

skyrpex:
Dropping this since I found a way to improve the Console DX without having to change the Wing CLI.

PR #3605, opened 2023-07-25
fix(compiler): cannot bring modules more than once

### LINKED ISSUE #3509: Bad error message on bring/import cycles

### LINKED ISSUE #3511: Cannot reuse dependency that is used elsewhere (diamond dependency problem)

### LINKED ISSUE #3578: Cannot bring the same module from multiple files
### I tried this:

`main.w`:

```js
bring math;
bring "./other.w" as other;
```

`other.w`:

```js
bring math;`
```

### This happened:

```
`math` is already defined
```

### I expected this:

To work

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

Chriscbr:
Refactors how the Wing compiler handles `bring` for local Wing files in order to support more complex `bring` use cases, like allowing multiple files to `bring` the same module.

The high level implementation approach is instead of generating a single, monolithic AST, we augment the parser so that after a Wing source file is parsed, an isolated AST and a list of dependent Wing files is produced. Using this, we can write a function that parses an entire Wing program given an entrypoint, and use the `petgraph` crate to produce topological ordering of the files for type checking, and each AST is independently type checked.

While this unblocks several use cases, there are still some limitations. For example, consider how it's possible for classes to recursively reference each other in TypeScript:

```ts
// foo.ts
import * as bar from "./bar";

export class Foo {
  constructor(public bar: bar.Bar | undefined) {}
}

new Foo(new bar.Bar(new Foo(undefined)));


// bar.ts
import * as foo from "./foo";

export class Bar {
  constructor(public foo: foo.Foo | undefined) {}
}
```

This doesn't work in Wing currently because each file is still type checked independently. To relax this constraint, we would need to perform symbol resolution on all files first (to identify all available types and members) before performing type checking. This kind of refactoring would be pretty large and is out of scope for this PR.

Fixes #3511
Fixes #3578
Fixes #3509

## Checklist

- [x] Tests added (always)


libs/wingc/src/ast.rs
```diff
@@ -418,16 +418,19 @@ pub struct Interface {
...
 pub enum StmtKind {
 	Bring {
-		module_name: Symbol, // Reference?
+		source: BringSource,
 		identifier: Option<Symbol>,
 	},
-	Module {
-		name: Symbol,
-		statements: Scope,
-	},
```
Chriscbr, line 430: I think there's merit to adding a syntax for creating modules within files, e.g.

```js
pub module Foo {
  pub class Bar {}
  pub class Baz {}
}
```

But that's not really the goal of this PR / I wanted to keep this PR on the smaller side, so I decided to remove it for now since the code would otherwise be unused.

eladb:
> it's possible for classes to recursively reference each other in TypeScript

I'd argue that this type of cyclic dependency is not really something we want to support. In typescript/JavaScript it's oftentimes a source of headaches.

examples/tests/valid/store.w
```diff
@@ -1,3 +1,6 @@
+bring "./subdir/empty.w" as file3;
```
eladb, line 1: I got to say that I really dont like that we are requiring the `.w` here. @MarkMcCulloh i remember you were very opinionated about it. Can you remind me why?

Chriscbr:
Yeah, I agree it seems like the kind of thing that's healthy to try avoiding to start

libs/wingc/src/jsify.rs
```diff
@@ -129,11 +137,14 @@ impl<'a> JSifier<'a> {
 				"const {} = process.env.WING_IS_TEST === \"true\";",
 				ENV_WING_IS_TEST
 			));
+		} else {
+			output.open(format!("module.exports = function({{ {} }}) {{", STDLIB));
+			output.line(format!("const std = {STDLIB}.{WINGSDK_STD_MODULE};"));
```
eladb, line 142: I think this can be moved outside of the `if` blocks along with line 35

libs/wingc/src/jsify/snapshots/base_class_with_lifted_field_object.snap
```diff
@@ -58,7 +58,7 @@ const $stdlib = require('@winglang/sdk');
 const $outdir = process.env.WING_SYNTH_DIR ?? ".";
 const std = $stdlib.std;
 const $wing_is_test = process.env.WING_IS_TEST === "true";
-const cloud = require('@winglang/sdk').cloud;
+const cloud = $stdlib.cloud;
```
eladb, line 61: Should we merge those to:

```js
const { cloud, std } = $stdlib;
```

eladb (approved):
After playing around with this feature, I tend to agree with @MarkMcCulloh that we should distinguish between entrypoint files and module files more explicitly. This is especially important for the IDE experience and Wing Console.

I was wondering if maybe entrypoint files should have the `.main.w` suffix (and also `main.w`).

Thoughts?

examples/tests/valid/store.w
skyrpex: Omitting the file extension has caused many issues in the Node ecosystem:

- Runtime code has to do expensive look ups. For example, when doing `require("./subdir/empty")`, Node needs to check:
  - Does `./subdir/empty.js` exist?
  - Does `./subdir/empty/index.js` exist?
  - Not sure if it also checks for `.json` files...
- Nowadays, there's not only `.js` files. There are:
  - TypeScript files `.ts`
  - TSX and JSX `.tsx, .jsx`
  - CommonJS files `.cjs`
  - ESM files `.mjs`
  - You can configure a bundler to load any file extension with the logic you need (`.vue`, `.svelte`

Mix all of those and we get the mess we have now. This has been corrected in ESM now, since providing the full filename is necessary now.

I think we should follow the modern trend and force the full filename since the beginning. I believe we will end up being able to bring/import more file formats.

MarkMcCulloh: ^+1, basically we should learn from past mistakes. 
Considering the file extension is one letter, the price to pay isn't even too bad

yoav-steinberg: I think currently omitting the extension would mix this up with the JSII bringing. Of course we can have some heuristics about this, but still. I suggest for now to require it. Removing it later as a feature can also work.

libs/wingc/src/files.rs
```diff
@@ -58,6 +58,10 @@ impl Files {
 		self.data.get(path.as_ref())
 	}
 
+	pub fn contains_file<S: AsRef<Path>>(&self, path: S) -> bool {
+		self.data.contains_key(path.as_ref())
```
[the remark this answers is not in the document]
MarkMcCulloh: We unfortunately can't use canonicalize, but storing the inode makes sense.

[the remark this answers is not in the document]
Chriscbr: I don't think we can use inodes since they're not available on Windows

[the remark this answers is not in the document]
Chriscbr: As written, `Files` assumes each file is only indexed with a single, normalized path (to avoid issues of multiple ways to represent the same path) - so path normalization is the responsibility of the caller. Definitely something we can improve on, though I think I'll leave it for follow-up PRs.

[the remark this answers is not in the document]
MarkMcCulloh: > I don't think we can use inodes since they're not available on Windows

While that's true, we should check to see what we actually get from WASI. It's very possible it's providing an equivalently useful number for us. I'm also cool with this being a followup issue

libs/wingc/src/parser.rs
```diff
@@ -131,6 +132,138 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
+		let curr_path = needs_parsing.remove(0);
+		if files.contains_file(&curr_path) {
+			continue;
+		}
+
+		let source = match fs::read(&curr_path) {
+			Ok(source) => source,
+			Err(err) => {
+				report_diagnostic(Diagnostic {
+					message: format!("Error reading source file: {}: {:?}", curr_path.display(), err),
```
yoav-steinberg, line 186: I suspect rust io errors are `Dispaly`able.
```suggestion
					message: format!("Error reading source file: {}: {}", curr_path.display(), err),
```

libs/wingc/src/type_check.rs
```diff
@@ -2622,40 +2649,78 @@ impl<'a> TypeChecker<'a> {
...
-						library_name = WINGSDK_ASSEMBLY_NAME.to_string();
-						namespace_filter = vec![module_name.name.clone()];
-						alias = identifier.as_ref().unwrap_or(&module_name);
-					} else if module_name.name.as_str() == WINGSDK_STD_MODULE {
-						self.spanned_error(stmt, format!("Redundant bring of \"{}\"", WINGSDK_STD_MODULE));
-						return;
-					} else {
-						self.spanned_error(stmt, format!("\"{}\" is not a built-in module", module_name.name));
+					BringSource::JsiiModule(name) => {
+						if identifier.is_none() {
```
yoav-steinberg, line 2667: didn't we check this in parser.rs? can't we just `expect()` here?

libs/wingc/src/type_check.rs
```diff
@@ -2622,40 +2649,78 @@ impl<'a> TypeChecker<'a> {
...
+							);
+							return;
+						}
+						library_name = name.name.to_string();
+						// no namespace filter (we only support importing entire libraries at the moment)
+						namespace_filter = vec![];
+						alias = identifier.as_ref().unwrap();
+					}
+					BringSource::WingFile(name) => {
+						if identifier.is_none() {
```
yoav-steinberg, line 2683: didn't we check this in parser.rs? can't we just expect() here?

libs/wingc/src/type_check.rs
```diff
@@ -2622,40 +2649,78 @@ impl<'a> TypeChecker<'a> {
...
+						}
+
+						let lookup_name = sanitize_dots(&name.name);
+						let lookup_result = self.types.libraries.lookup_nested_str(&lookup_name, None);
+						match lookup_result {
+							LookupResult::Found(kind, _) => match kind {
+								SymbolKind::Type(_) => panic!("expected a namespace, found a type"),
+								SymbolKind::Variable(_) => panic!("expected a namespace, found a variable"),
+								SymbolKind::Namespace(nsref) => {
+									if !nsref.is_bringable {
```
yoav-steinberg, line 2701: Could be cool is instead of a `bool` here we'd have some span info of the invalid statement, and generate an appropriate diag.

yoav-steinberg (commented):
Amazing!

libs/wingc/src/parser.rs
```diff
@@ -131,6 +132,138 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
+					span: None,
+				});
+				continue;
+			}
+		};
+		tree_sitter_trees.insert(curr_path.clone(), tree_sitter_tree.clone());
+
+		let tree_sitter_root = tree_sitter_tree.root_node();
+
+		// Parse the single file, and collect a list of all wing files that it references
```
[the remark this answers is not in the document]
Chriscbr: Fixed

libs/wingc/src/type_check.rs
Chriscbr: Oops, yeah you're right, this is duplicated. Thanks

libs/wingc/src/parser.rs
```diff
@@ -131,6 +132,138 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
+				.collect::<String>();
+			let component1 = component1.trim_end();
+
+			report_diagnostic(Diagnostic {
+				message: format!(
+					"Could not compile \"{}\" due to cyclic bring statements:\n{}",
+					init_path.display(),
+					component1
+				),
+				span: None,
```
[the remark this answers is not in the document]
Chriscbr: https://github.com/winglang/wing/issues/3626

libs/wingc/src/jsify/snapshots/base_class_with_lifted_field_object.snap
Chriscbr: I think it's doable but it will add some complexity, since right now the JSifier follows a simple model of generating one statement per AST statement. The change would also be tricky since currently the compiler allows `bring` statements anywhere in the AST, e.g.:

```js
if true {
  bring math;
}
```

In any case, sounds like a nice future optimization.

libs/wingc/src/jsify.rs
Chriscbr: Done

libs/wingc/src/files.rs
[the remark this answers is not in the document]
Chriscbr: 👀  interesting. I'll create an issue

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
Chriscbr: Looks like bringing inflight doesn't totally work at the moment. Also the `Utils` convention doesn't seem to be working with multi-file projects (I had to use `my_utils.Utils.get_str` to fix a type checking error). Since `bring` inside inflights isn't as common I'll open a separate issue for it. I reckon it will need special treatment in order to support it.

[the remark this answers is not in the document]
Chriscbr: https://github.com/winglang/wing/issues/3628

libs/wingc/src/parser.rs
```diff
@@ -590,75 +727,58 @@ impl<'s> Parser<'s> {
...
-			} else {
-				source_path
-			};
-			let source_path = if source_path.is_absolute() {
-				source_path.to_path_buf()
-			} else {
-				Path::new(&self.source_name).parent().unwrap().join(source_path)
-			};
+			let source_path = normalize_path(source_path, Some(&Path::new(&self.source_name)));
 			if source_path == Path::new(&self.source_name) {
```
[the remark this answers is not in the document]
Chriscbr: We have a unit test that checks the error if a file imports itself (the case you mention). I'll double check this and see if there's a bug I can tease out

libs/wingc/src/type_check.rs
Chriscbr: Created an issue https://github.com/winglang/wing/issues/3626

libs/wingc/src/parser.rs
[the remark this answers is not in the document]
Chriscbr: self.source_name should always be normalized when it's provided to `Parser`, so I don't think there's an issue here. Though I did discover some limitations with how we're normalizing, so I fixed that and added another test.

libs/wingc/src/lsp/sync.rs
```diff
@@ -76,75 +77,80 @@ pub fn on_document_did_change(params: DidChangeTextDocumentParams) {
...
-			);
+			let result = partial_compile(&uri_path, &mut jsii_types.borrow_mut());
 			files.borrow_mut().insert(uri, result);
 		});
 	})
 }
 
 /// Runs several phases of the wing compile on a file, including: parsing, type checking, and capturing
-fn partial_compile(source_file: &str, text: &[u8], jsii_types: &mut TypeSystem) -> FileData {
+fn partial_compile(source_path: &Path, jsii_types: &mut TypeSystem) -> FileData {
```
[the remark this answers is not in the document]
Chriscbr: Perhaps..

libs/wingc/src/files.rs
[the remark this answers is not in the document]
yoav-steinberg: > We unfortunately can't use canonicalize, but storing the inode makes sense.

Maybe delegating canonicalize to node through some WASM callback can solve this too. It'll be useful in other cases, like the JSII file caching where initially I used `canonicalize` to uniquely identify the .jsii file.

Chriscbr:
> I was wondering if maybe entrypoint files should have the `.main.w` suffix (and also `main.w`).
> 
> Thoughts?

+1 from me. I think a convention based on the file name seems pretty good as far as options go. Since most source files in larger Wing projects will be non-entrypoint files, I think it makes sense that the any extra rules or annotations required by the compiler should be placed on the entrypoint files, not the non-entrypoint files - if that makes any sense.

skyrpex:
Hey, this PR caused an issue in the LSP. We had to revert Learn and Play to a previous version.

See the Slack convo https://winglang.slack.com/archives/C04QQD2SDM2/p1690534520811589.

We'll have to revert this PR or fix whatever problem there is.

ShaiBer:
Let's revert for now so newer releases will not break the playground again

skyrpex:
Here's the revert PR https://github.com/winglang/wing/pull/3641

PR #3608, opened 2023-07-26
fix(compiler): compile to cjs rather than js

### LINKED ISSUE #3601: Invoking functions using a wingfile inside an ESM package fails
### I tried this:

Given the following files:

`package.json`
```json
{
  "type": "module"
}
```

`index.w`:
```wing
bring cloud;
  return "hello world";
});
```

Run `wing it` and execute the function.

### This happened:

The function invocation errors `module not defined`. Using the Simulator API instead of the Console should give the same result.

### I expected this:

The function invocation should succeed and return `"hello world"`

### Is there a workaround?

You can place an empty `package.json` (with empty meaning `{}`)
 inside the `target/` directory so the wing-compiled files are treated as CJS instead of ESM.

### Component

Compiler, SDK

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

The easiest and most portable solution is to change the extension of the wing-compiled JS files from `.js` to `.cjs`. This will make them as CJS instead of being dependent on any `package.json` that exists in upper directories.

### Community Notes

skyrpex:
The Wing suite only works with CJS files. Trying to use the Console or the SImulator inside an ESM package will fail during function invocations because Wing-generated files will be treated as ESM.

Fixes #3601.

### DIFF
--- a/libs/wingc/src/jsify.rs
+++ b/libs/wingc/src/jsify.rs
@@ -28,7 +28,7 @@ use crate::{
 
 use self::codemaker::CodeMaker;
 
-const PREFLIGHT_FILE_NAME: &str = "preflight.js";
+const PREFLIGHT_FILE_NAME: &str = "preflight.cjs";
 
 const STDLIB: &str = "$stdlib";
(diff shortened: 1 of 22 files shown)


bronifty:
https://github.com/winglang/wing/pull/4297

Issue #1973, opened 2023-04-03
Logs are not displayed during `wing test -t tf-aws`


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

Writing a wing test that uses `log`, and then testing it with `wing test -t tf-aws`

### I expected this:

I expected the logs to be displayed in my terminal, just like they do when I run `wing test -t sim`, for example:

<img width="363" alt="Screenshot 2023-04-03 at 5 36 12 PM" src="https://user-images.githubusercontent.com/5008987/229632955-99526d02-87c6-4799-ae3b-3abb92e6858d.png">

### Instead, this happened:

No logs appeared

### Component:

SDK

skorfmann:
This would be pretty useful to have, in particular since all the infra is cleaned up automatically after running the tests. I created another issue to make this configurable https://github.com/winglang/wing/issues/2884. However, the most useful thing right now would be to just see log statements directly in the wing test output.

skorfmann:
Just ran into this again. At least a hint that logs are not visible and should be checked with the cloud provider would go a long way already (and since we should know the test function arns, even creating a direct URL to clouddwatch should be doable)

skorfmann:
Yeah, there's no way finding the corresponding log group is something a user could do manually if there's more than one test function

![Screenshot 2023-06-29 at 23 33 49](https://github.com/winglang/wing/assets/136789/0a2c034f-c2fb-4e11-a8c4-520f1917eddd)

Chriscbr:
Yep. Also I think it might be sensible if these functions are named based on the name of the test

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.67](https://github.com/winglang/wing/releases/tag/v0.24.67).

PR #3622, opened 2023-07-26
feat(sdk, compiler): displaying logs of tf-aws functions

tsuf239:
fixes: #1973
## Description
* Collecting and printing `log(...)` outputs when running a test on target tf-aws

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -11,12 +13,49 @@ export class FunctionClient implements IFunctionClient {
...
+        ) ?? [];
+
+      logsCollector.push(
+        ...logs.map((item: string) => ({
+          data: {
+            message: (item.match(/(?<=winglogstart:)[\s\S]*?(?=:winglogend)/) ??
+              [])[0],
+          },
+          sourceType: FUNCTION_TYPE,
+          sourcePath: functionName, //TODO: is that the right thing to do?
```
Chriscbr, line 43: I think it's worth trying to be consistent with simulator (if possible) and try providing the construct path. It might be possible to do this by inserting a new parameter to the constructor in this file:

```ts
constructor(
    private readonly functionArn: string,
    private readonly constructPath: string,
    private readonly lambdaClient = new LambdaClient({})
  ) {}
```

and injecting the value when creating the inflight client:

https://github.com/winglang/wing/blob/7d9910208966d69005cc87b06186d07f23839162/libs/wingsdk/src/target-tf-aws/function.ts#L253

to
```ts
[`process.env["${this.envName()}"]`, this.node.path]
```

If it doesn't work, using the AWS function ARN sounds ok

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -11,12 +13,49 @@ export class FunctionClient implements IFunctionClient {
...
+        }))
+      );
+    } else {
+      console.log(`Invoking ${functionName}:`);
+      console.log(
+        atob(response.LogResult ?? "")
+          .match(/winglogstart:[\s\S]+?:winglogend/gm)
+          ?.join("\n") ?? ""
+      );
+    }
```
Chriscbr, line 57: Do we need this else branch?

tsuf239: Yes, when we have a function that is called in another function or a test, it doesn't get caught, unless it prints to the test function log

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -11,12 +13,49 @@ export class FunctionClient implements IFunctionClient {
   /**
    * Invoke the function, passing the given payload as an argument.
    */
-  public async invoke(payload: string): Promise<string> {
+  public async invoke(
+    payload: string,
+    logsCollector?: Trace[]
+  ): Promise<string> {
```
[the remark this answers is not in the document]
tsuf239: I'll split the implementation

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -11,12 +13,49 @@ export class FunctionClient implements IFunctionClient {
...
+          // date, then any char then a line of winglog
+          /(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z)(.*)winglogstart:[\s\S]+?:winglogend/gm
+        ) ?? [];
+
+      logsCollector.push(
+        ...logs.map((item: string) => ({
+          data: {
+            message: (item.match(/(?<=winglogstart:)[\s\S]*?(?=:winglogend)/) ??
+              [])[0],
+          },
```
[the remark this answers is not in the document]
tsuf239: I agree, but I think we should let the user choose, (if they got a test calling to 100 cloud function, that only 20 of them log something, instead of getting 20 lines, it will get messy pretty soon), I think I'll add a flag, what about `-log=full | exact | none` with the default to `exact`? (I really have no idea what it should be called)

tsuf239:
works on `awscdk` too

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -25,14 +129,66 @@ export class FunctionClient implements IFunctionClient {
...
+
+    const value = await this.executeFunction(payload);
+
+    if (
+      value?.context?.logGroupName &&
+      value?.context?.logStreamName &&
+      logType !== LogType.NONE
+    ) {
+      // waiting for the logs to be created
+      await new Promise((resolve) => setTimeout(resolve, 10000));
```
Chriscbr, line 181: Does this mean every time a user runs `wing test` on the cloud they will need to wait an extra ten seconds to get feedback of their test results? We should probably avoid making developers wait longer to get feedback in "happy path" situations. (ten seconds is a lot!)

I'd think of the logs shown in `wing test` as a *best effort* mechanism. A function can produce thousands of logs, so what's most important is handling common use cases well (like showing the last several dozens of logs that were printed). When that isn't enough, the user should be provided enough information (like the name of the function and the request ID) to continue their debugging. Maybe for tools like the Wing Console we will add some other kind of moethod like `tailLogs()` for when you want to run a test and stream all of the logs as the test runs. But I feel like we should treat that as a separate use case since it requires extra resources. WDYT?

apps/wing/src/cli.ts
```diff
@@ -151,6 +152,11 @@ async function main() {
         .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
         .default("sim")
     )
+    .addOption(
+      new Option("-l, --log <LogType>", "Log type")
+        .choices([LogType.DEFAULT, LogType.NONE, LogType.EXTENDED])
+        .default(LogType.DEFAULT)
+    )
```
Chriscbr, line 159: How do these logging levels line up with the log levels in the Wing Console? And what's the difference between `--log=extended` and `--debug` CLI options?

I feel like some more discussion is needed about logging. Maybe we can leave this as a follow up enhancement since it wasn't part of the original issue

Screenshot 2023-07-31 at 4 04 01 PM

tsuf239: I think you're right, maybe it will be good for this PR to allow `--debug` to behave like `-log=extended`.
I'll open an issue for thinking of a better and more systematic approach

eladb: +1 that this all needs to align with the console

tsuf239:
I found an interesting way to get the logs in O(1) without any async calls, and without wrapping the string,
by introducing a logger class.
This way we can send them to a 3rd party log collector of the user's choice in the future :) and have better control over log levels and formatting. 
Therefore, I'll move this to draft now, and let you know when the new code is ready! 🎉

libs/wingsdk/src/shared-aws/function.inflight.ts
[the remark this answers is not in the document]
tsuf239: update: removed

tsuf239:
Gave up the logger class idea- chose a much easier approach of changing the `console.log` behavior in the aws cloud functions to save the wing logs, and display them as traces

libs/wingsdk/src/shared-aws/function.inflight.ts
tsuf239: changed the mechanism a bit now to prevent waiting, and logging extra chars,
I like the `tailLogs` idea, we can add it to the function execution in the future, should I open an issue?

apps/wing/src/cli.ts
tsuf239: https://github.com/winglang/wing/issues/3680

Chriscbr (approved):
Looks good to me - approved conditional on addressing a few small nitpicks. This will be a huge improvement for testing apps on AWS. Once we have more resources on other cloud targets, I imagine we might be able to reuse the approach there as well 🚀

libs/wingsdk/src/target-awscdk/function.ts
```diff
@@ -71,13 +72,48 @@ export class Function extends cloud.Function {
...
+   * @param inflightClient inflight client code
+   * @returns cloud function code string
+   * @internal
+   */
+  protected _generateLines(inflightClient: core.Code): string[] {
+    const lines = new Array<string>();
+
+    lines.push("exports.handler = async function(event, context) {");
+    lines.push(`
+  const console_log = console.log.bind({});
```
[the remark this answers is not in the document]
tsuf239: no, it will reassign the function, and make a recursion...
the `.bind()` is used for cloning:
 https://stackoverflow.com/questions/1833588/javascript-clone-a-function

examples/tests/sdk_tests/function/invoke.w
```diff
@@ -2,15 +2,18 @@ bring cloud;
 bring util;
 
 let payload = "hello";
+log("log preflight");
```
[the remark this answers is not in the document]
tsuf239: image

https://github.com/winglang/wing/blob/16d78df99849d79d107fe647b2acc9488f216fb0/examples/tests/sdk_tests/function/logging.w#L1-L35

tsuf239:
Resolving all comments :) @Chriscbr please let me know if something is missing and I'll push it into a further PR

libs/wingsdk/src/target-awscdk/function.ts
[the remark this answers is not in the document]
Chriscbr: TIL!

PR #3636, opened 2023-07-27
chore(docs): changed title to "pulumi vs. wing [main differences]"

RoniWDS:
## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/docs/999-faq/040-alternatives/042-pulumi-vs-wing.md
+++ b/docs/docs/999-faq/040-alternatives/042-pulumi-vs-wing.md
@@ -1,5 +1,5 @@
 ---
-title: What are the main differences between Wing and Pulumi?
+title: Pulumi vs. Wing [Main differences]
 sidebar_label: Pulumi
 id: differences-from-pulumi
 keywords: [faq, winglang, Wing programming language, Wing language, Pulumi, IAC]


hasanaburayyan (commented):
Hey @RoniWDS 👋 Thanks for the suggested title change. I realize the file names say "vs" but Im wondering if that sends the wrong message. For instance I could imagine a world where Wing could have a target that leverages Pulumi (as it does with CDKTF and CDKAWS). Which makes me think that in the FAQ section for comparisons it might make sense to use "main differences" over "vs" 

But Im interested to hear from @ShaiBer on this one.

ShaiBer:
The 'vs' is for SEO - to make it easier to find, but if it has a negative connotation and reduces the accuracy of the message then we will not use it

ekeren:
I like `vs` , this is how I personally find solutions that compete with each other. I write in Google "pulumi vs" and it completes the competition 

Pulumi also likes `vs` 

https://www.pulumi.com/docs/concepts/vs/

https://www.pulumi.com/docs/concepts/vs/terraform/

MarkMcCulloh:
I think going with `vs` makes sense even though I personally dislike the adversarial phrasing. One day when we support a pulumi target, the comparison for using wing as a solution vs pulumi as a solution is still valid. It's like comparing wing to terraform.

hasanaburayyan (approved):
Alright I think we have consensus on this change being worth it. So Im going to approve these PRs

PR #3643, opened 2023-07-29
feat(compiler): variadic arguments

### LINKED ISSUE #125: Function Declaration: Variadic Arguments

### LINKED ISSUE #397: Jsii variadic parameter support

marciocadev:
This PR adds the possibility of using variadic arguments

Unfortunately, I had to add the `Clone` attribute to many `enum` and `struct` to be able to create a list of arguments compatible with the requirements for verifying the typing of the variadic argument.

I needed to implement `PartialEq` for `Type`, `TypeRef`, `Class`, `Interface`, `Struct`, and `Enum`.
I'm not sure if it was done in the best way, but it fulfills this specific case.

- [x] Validates if the last argument (variadic) of the function accepts multiple elements.
- [x] Validates if the only argument (variadic) of the function accepts multiple elements.
- [x] Validates if all types of the variadic argument are the same.
- [ ] Accept range as a variadic argument.

Closes #125 
Closes #397

## Checklist

- [x] Tests added (always)


revitalbarletz:
@marciocadev we are looking into this and will add comments soon!

libs/tree-sitter-wing/grammar.js
```diff
@@ -511,14 +511,34 @@ module.exports = grammar({
...
+      "(",
+      choice(
+        seq(
+          commaSep($.parameter_definition),
+          optional(seq(",", $.variadic_definition)),
+        ),
+        $.variadic_definition,
+      ),
+      ")"
+    ),
```
Chriscbr, line 541: Even though it makes the grammar a bit less precise, I think we can improve the error diagnostics if we specify in the grammar any parameter can be variadic, and then handle the validation that only the last parameter can be variadic inside `parser.rs`. This way we can give a more friendly error than "unexpected variadic definition".

Something like this maybe?

 ```
    parameter_definition: ($) =>
      seq(
        optional(field("reassignable", $.reassignable)),
        optional(field("variadic", $.variadic)),
        field("name", $.identifier),
        $._type_annotation
      ),

    parameter_list: ($) => seq("(", commaSep($.parameter_definition), ")"),
```

(also totally okay keeping the current implementation if there are issues with that one)

File: libs/wingc/src/type_check.rs [hunk 2]
```diff
@@ -2076,10 +2118,43 @@ impl<'a> TypeChecker<'a> {
...
+			for (arg_expr, arg_type, param) in izip!(arg_list.pos_args.iter(), arg_list_types.pos_args.iter(), params) {
+				self.validate_type(*arg_type, param.typeref, arg_expr);
+			}
+		} else {
+			let mut new_arg_list: Vec<Expr> = Vec::new();
+			let mut new_arg_list_types: Vec<TypeRef> = Vec::new();
+			for i in 0..index_last_item {
+				new_arg_list.push(arg_list.pos_args.get(i).unwrap().clone());
+				new_arg_list_types.push(*arg_list_types.pos_args.get(i).unwrap());
+			}
```
Chriscbr: I refactored some of the code in this area to avoid the need for Clone and PartialEq implementations. Let me know if that's ok / if there's any issues with it

Chriscbr (changes requested):

libs/wingc/src/type_check.rs
marciocadev: It worked perfectly, thank you @Chriscbr

marciocadev:
@Chriscbr I will put this PR back into WIP for refactoring.

examples/tests/valid/function_variadic_arguments.w
```diff
@@ -0,0 +1,18 @@
...
+let func1 = (var x: num, y: str?, ...args: Array<num>) => {
+  assert(x == 1);
+  assert(y == "something" || y == nil);
+  assert(args.length == 4);
+  for i in args {
+    assert(i > 0 && i < 5);
+  }
+};
+func1(1, "something", 1, 2, 3, 4);
+func1(1, nil, 1, 2, 3, 4);
```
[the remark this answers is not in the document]
Chriscbr: @marciocadev I don't think the "optional" marker should be needed. This is the same as how it works in TypeScript ([playground](https://www.typescriptlang.org/play?#code/DYUwLgBAhgJjByBXAtgZwgXggCgHT4DsVUAuCAQQCdKoBPAHiOQCMRKA+ASjKdcs3YQA3gCgI4iKEhgA9mCjBMEAAwBuMRIBmM-tgDGMgqkhMIASwIQmqTsI0SJs+YoDUWAIzqHAX3sRK4IiUlk4K6t7qIgZGMqC4wDIA5tiwCMTY7gA0EABM2QDMnJzq0aixIPFJKXBIaBlFJYZlcQnJqbWo2A1AA)) and Golang ([playground](https://go.dev/play/p/M6AomVI8QU8)).

examples/tests/invalid/function_variadic_definition.w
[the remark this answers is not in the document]
marciocadev: @Chriscbr I don't undestand this `when a user tries to define a function with more than one parameter that is parameter`

did you mean `when a user tries to define a function with more than one variiadic parameter that is parameter` ?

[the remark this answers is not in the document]
Chriscbr: @marciocadev Yes that's what I meant, sorry for the typo :)

libs/wingc/src/parser.rs
```diff
@@ -1125,7 +1125,7 @@ impl<'s> Parser<'s> {
 
 				Ok(udt)
 			}
-			"mutable_container_type" | "immutable_container_type" => {
+			"mutable_container_type" | "immutable_container_type" | "variadic_container_type" => {
```
Chriscbr, line 1128: Revert

libs/wingc/src/parser.rs
```diff
@@ -1228,7 +1229,7 @@ impl<'s> Parser<'s> {
 					other => self.with_error(format!("invalid json container type {}", other), &type_node),
 				}
 			}
-			"mutable_container_type" | "immutable_container_type" => {
+			"mutable_container_type" | "immutable_container_type" | "variadic_container_type" => {
```
Chriscbr, line 1232: Revert

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -465,6 +465,11 @@ impl<'a> JsiiImporter<'a> {
 							name: param.name.clone(),
 							typeref: self.parameter_to_wing_type(&param),
 							docs: Docs::from(&param.docs),
+							variadic: if param.variadic.is_some() {
+								param.variadic.unwrap()
+							} else {
+								false
+							},
```
Chriscbr, line 472:
```suggestion
							variadic: param.variadic.unwrap_or(false),
```

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -724,6 +729,11 @@ impl<'a> JsiiImporter<'a> {
 						name: param.name.clone(),
 						typeref: self.parameter_to_wing_type(&param),
 						docs: Docs::from(&param.docs),
+						variadic: if param.variadic.is_some() {
+							param.variadic.unwrap()
+						} else {
+							false
+						},
```
Chriscbr, line 736:
```suggestion
						variadic: param.variadic.unwrap_or(false),
```

Chriscbr (approved):
Nice work @marciocadev ! 👏

Issue #3139, opened 2023-06-29
Json: schema validation and assignment to struct


staycoolcall911:
##### 1.1.4.5 Schema validation
All `fromJson()` methods will validate that the runtime type is compatible with the target type in
order to ensure type safety (at a runtime cost):
```js
```TS
str.fromJson(jsonNumber);      // RUNTIME ERROR: unable to parse number `123` as a string.
num.fromJson(Json "\"hello\""); // RUNTIME ERROR: unable to parse string "hello" as a number
```

For each `fromJson()`, there is a `tryFromJson()` method which returns an optional `T?` which
indicates if parsing was successful or not:
```js
let s = str.tryFromJson(myJson) ?? "invalid string";
```
##### 1.1.4.7 Assignment to user-defined structs
All [structs](#31-structs) also have a `fromJson()` method that can be used to parse `Json` into a
struct:
```js
struct Contact {
  first: str;
  last: str;
  phone: str?;
}

let j = Json { first: "Wing", last: "Lyly" };
let myContact = Contact.fromJson(j);
assert(myContact.first == "Wing");
```
When a `Json` is parsed into a struct, the schema will be validated to ensure the result is
type-safe:
```js
let p = Json { first: "Wing", phone: 1234 };
Contact.fromJson(p);
// RUNTIME ERROR: unable to parse Contact:
// - field "last" is required and missing
// - field "phone" is expected to be a string, got number.
```

hasanaburayyan:
@staycoolcall911 this needs bumped up to P1 before Beta right? Well at least the `struct.fromJson()` bit

staycoolcall911:
Yes, definitely, this came up much during dogfooding. Bumping to p1

hasanaburayyan:
Splitting out some features into different issues that wont be covered by initial work in: https://github.com/winglang/wing/pull/3648

additional issues:
- https://github.com/winglang/wing/issues/3789
- https://github.com/winglang/wing/issues/3788

monadabot:
Congrats! :rocket: This was released in [Wing 0.25.1](https://github.com/winglang/wing/releases/tag/v0.25.1).

PR #3648, opened 2023-07-30
feat(compiler): support json to struct conversion

hasanaburayyan:
The changes in this PR make it possible to call `fromJson` on any compatible struct definition. 

## Implementation notes:
Previously we treated the jsification of structs as a no-op since there is not a JS equivalent. So with this change structs now JSify into a `Struct` file that contains a class with a static `jsonSchema` and a `fromJson` function which will allow for field validation at runtime. 

The schema generated adheres to: https://json-schema.org/understanding-json-schema/

take this simple example Wing code:
```js
struct MyStruct {
	myField: str;
	myOtherField: num;
}
```

this will now generate a JS file named `MyStruct.Struct.js` which looks like this:
```js
module.exports = function(stdStruct, fromInline) {
  class MyStruct {
    static jsonSchema() {
      return {
        id: "/MyStruct",
        type: "object",
        properties: {
          myField: { type: "string" },
          myOtherField: { type: "number" },
        },
        required: [
          "myField",
          "myOtherField",
        ],
        $defs: {
        }
      }
    }
    static fromJson(obj) {
      return stdStruct._validate(obj, this.jsonSchema())
    }
    static _toInflightType(context) {
      return fromInline(`require("./MyStruct.Struct.js")(${ context._lift(stdStruct) })`);
    }
  }
  return MyStruct;
};

```

The piece that brings this all together is the addition of the `Struct` class in our std that only has a `fromJson()` methods at the moment that is a Wing macro. The macro just calls the `fromJson()` method in the generated javascript.

### Misc
We want to stop the user at compile time from calling `fromJson` on a struct that cannot be represented by a Json value ie
```js
struct MyStruct {
	b: cloud.Bucket;
}
let j = {};
MyStruct.fromJson(j);
```

to prevent this I added a check in the typechecker for structs to confirm that if `fromJson` is called that all the fields in the struct are valid for conversion attempt

See image below for error when attempting:
image

Closes: https://github.com/winglang/wing/issues/3653
Closes: https://github.com/winglang/wing/issues/3139

## TODO:
- [x] separate the work done here and the remaining work into different tickets.
- [x] update language reference

## Followup issues that are out of scope for this PR:
- https://github.com/winglang/wing/issues/3792
- https://github.com/winglang/wing/issues/3790
- https://github.com/winglang/wing/issues/3789
- https://github.com/winglang/wing/issues/3788

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify.rs
```diff
@@ -411,6 +455,9 @@ impl<'a> JSifier<'a> {
 								ExprKind::Reference(Reference::InstanceMember { object, .. }) => {
 									self.jsify_expression(&object, ctx)
 								}
+                ExprKind::Reference(Reference::TypeMember { .. }) => {
+                  expr_string.clone().split(".").next().unwrap().to_string()
```
yoav-steinberg, line 459: Have a feeling this might break with namespaces: `Namespace.Namesapce.Type.property`

libs/wingc/src/jsify.rs
```diff
@@ -758,9 +805,88 @@ impl<'a> JSifier<'a> {
...
+					if let Some(js_type) = self.jsify_type(&field.member_type.kind) {
+						let field_fqn = format!("{}.{}", &obj, &field.name.name);
+						let field_path = format!("{}.{}", name.name, &field.name.name);
+
+						code.open(format!(
+							"if ({}) {{",
+							self.jsify_struct_field_validation_check(&field.member_type.kind, &field_fqn)
+						));
+						code.line(format!(
+							"{}.push(`field: \"{}\" expected type of: \"{}\" but got type: ${{typeof {}}}`)",
```
yoav-steinberg, line 841:
```suggestion
							"{}.push(`field: \"{}\" expected type of: \"{}\" but got type: \"${{typeof {}}}\"`)",
```

libs/wingc/src/type_check.rs
```diff
@@ -4069,11 +4116,19 @@ impl<'a> TypeChecker<'a> {
 	}
 
 	/// Get's the type of an instance variable in a class
-	fn get_property_from_class_like(&mut self, class: &impl ClassLike, property: &Symbol) -> VariableInfo {
+	fn get_property_from_class_like(
+		&mut self,
+		class: &impl ClassLike,
+		property: &Symbol,
+		allow_static: bool,
```
yoav-steinberg, line 4123: why did you add this?

libs/wingsdk/src/std/struct.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  private constructor() { }
+
+  /**
+   * Converts a json to struct
+   * 
+   * This macro takes a json object and the struct definition file name then calls the validate method
+   * 
+   * @macro ((j, s) => {const errs = s.validate(j); if(errs.length !== 0) {throw new Error(`Unable to create struct: ${s.name} from json: ${JSON.stringify(j)}`)} return j;})($args$, $self$)
```
yoav-steinberg, line 25: Are we not adding the actual errors into the thrown exception? Is that why you `console.log` them in the validate function? If so you should make sure you `console.log` to stderr somehow. But don't you think it'll be better to add them to the exception and avoid outputing the the schema validate function?

yoav-steinberg (commented):
Very nice work
added some questions.

libs/wingc/src/jsify.rs
eladb, line 841:
```suggestion
							"{}.push(`field: \"{}\" expected type of: \"{}\" but got type: ${{typeof {}}}`);",
```

libs/wingc/src/jsify.rs
```diff
@@ -758,9 +805,88 @@ impl<'a> JSifier<'a> {
...
+						// additional checks based on type of field
+						match &field.member_type.kind {
+							TypeAnnotationKind::Array(t) => {
+								code.open(format!("for(let i = 0; i < {}.length; i++) {{", &field_fqn));
+								code.open(format!(
+									"if({}) {{",
+									self.jsify_struct_field_validation_check(&t.kind, &format!("{}[i]", field_fqn))
+								));
+								code.line(format!(
+									"{}.push(`field: \"{}\" expected type of: \"{}\"`)",
```
eladb, line 855:
```suggestion
									"{}.push(`field: \"{}\" expected type of: \"{}\"`);",
```

libs/wingc/src/jsify.rs
```diff
@@ -758,9 +805,88 @@ impl<'a> JSifier<'a> {
...
+							}
+							_ => {}
+						}
+					}
+				}
+
+				code.open(format!("if({}.length > 0) {{", &invalid_errors));
+				code.line(format!("console.log(`- ${{{}.join(\"\\n-\")}}`);", &invalid_errors));
+				code.close("}");
+				code.line(format!("return {}", &invalid_errors));
```
eladb, line 871: Missing ; here are well. 

Please check that all statements end with ;

libs/wingc/src/jsify.rs
```diff
@@ -758,9 +805,88 @@ impl<'a> JSifier<'a> {
...
+				code.line(format!("return {};", name));
+				code.close("};");
+
+				// Emits struct class file
+				self.emit_struct_file(name, code, ctx);
+
+				// Reset the code maker for code to be inserted in preflight.js
+				code = CodeMaker::default();
+				code.line(format!(
+					"const {} = require(\"{}\")($stdlib.core.NodeJsCode.fromInline)",
```
eladb, line 885: Perhaps just return a string and call fromInline here instead of passing in the function?

eladb (commented):
Great job 👏

eladb:
I was wondering if it might make more sense to generate a json schema and use a standard schema validation utility.

We also want to support something like:

```js
let s = MyStruct.jsonSchema();

s.validate()
s.toJson()
// etc..

```

The use case is for example using wing structs as api shapes with OpenAPI specs.

libs/wingsdk/src/std/struct.ts
eladb: I think the errors should be accumulated and included in the error message, not emitted by the validation code

eladb (changes requested):

libs/wingsdk/src/std/struct.ts
hasanaburayyan: Yes I remember telling myself this should probably not be a console.log at some point but I never went back to readdress 😅totally agree I’ll just return the array of errors and have them in the exception

yoav-steinberg:
> I was wondering if it might make more sense to generate a json schema and use a standard schema validation utility.

After reviewing I thought of this too. I think we should have a getSchema method in `std.Struct`, the validate method should be:
```js
var v = new require('jsonschema').Validator(); // see https://www.npmjs.com/package/jsonschema
v.validate(obj, self.getSchema());
```

hasanaburayyan:
> I was wondering if it might make more sense to generate a json schema and use a standard schema validation utility.

@yoav-steinberg @eladb When I originally started I wanted to generate schemas and went down a few learning rabbit holes, and did not think it was something we could easily do (because of nested arrays that can contain user defined structs), thats when I shifted gears to this approach. 

However, since then I have grown and learned more about the way bundling and some of the other JavaScript vodoo works 😂 😂 and I think generating these schemas for use in a standard validator can be "easishly" done. 

Im going to convert this PR to a WIP and investigate changing to that pattern (assuming there are not any objections)

libs/wingc/src/type_check.rs
hasanaburayyan: I need to allow calling static methods on Struct definitions, and we cant do the same thing as we do for Json, str, etc.. because the name of the Struct is user defined. (I believe we have special checks for language primitive static functions)

I was not sure how to get around adding this, but would love thoughts.

libs/wingsdk/src/std/struct.ts
```diff
@@ -0,0 +1,32 @@
...
+   */
+  public static fromJson(json: Json): T1 {
```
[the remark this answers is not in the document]
hasanaburayyan: Id like to implement the `tryFromJson` in an immediate followup PR just because that is going to need to do some changes to macro system (hydrating optionals)

here is the issue: https://github.com/winglang/wing/issues/3653

libs/wingc/src/jsify.rs
hasanaburayyan: hmm, so my intent was to reduce the boiler plating in the preflight.js file for what I am hoping is a readability improvement 🤞. Is there a concern with passing `fromInline`? I could totally be naive here

libs/wingc/src/type_check.rs
```diff
@@ -3944,6 +3964,29 @@ impl<'a> TypeChecker<'a> {
...
+										property,
+										format!(
+											"Struct \"{}\" contains a non-json compatible field \"{}\" and cannot call fromJson method",
+											type_, name
+										),
+									);
+									return (self.make_error_variable_info(), Phase::Independent);
+								}
+							}
+						}
```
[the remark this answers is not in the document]
hasanaburayyan: Well initially I added it because I swore it was in the spec, but Im not seeing it at the moment 😁 (maybe its in some thread). Though Im torn because I dislike adding the special case checks, but I think there is a pretty big value add in giving the user a heads up in compilation time versus allowing them to march forward to a runtime error. Especially in the case of when I add `tryFromJson` the user may get pretty confused as to why they always get undefined when trying to parse.

examples/tests/valid/struct_json_conversion.w
```diff
@@ -0,0 +1,147 @@
...
+  firstName: "Sally",
+  lastName: "Reynolds",
+  enrolled: false,
+  schoolId: "s2-xyz",
+  dob: { month: 5, day: 31, year: 1987},
+  courses: [
+    { name: "COMP 101", credits: 2 },
+    { name: "COMP 121", credits: 4 },
+  ]
+};
```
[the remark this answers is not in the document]
hasanaburayyan: Updated test, but yes that will work because the validation does structural equivalence.

libs/wingc/src/jsify.rs
hasanaburayyan: I added an `unwrap_or()`

examples/tests/error/struct_from_json_2.w
```diff
@@ -0,0 +1,24 @@
...
+}
+
+let missingAdvisor = {
+  name: "cool",
+  age: "not a number"
+};
+
+Student.fromJson(missingAdvisor);
+//              ^ ERROR: Unable to create struct: Student from json: {"name":"cool","age":"not a number"}
+// - field: "Student.advisor" expected type of: "Advisor" but got type: undefined
```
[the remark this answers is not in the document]
hasanaburayyan: It does now 😁

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
yoav-steinberg: Ok let's keep this approach, the verbose error is worth it I think.

libs/wingc/src/jsify.rs
```diff
@@ -227,10 +227,71 @@ impl<'a> JSifier<'a> {
...
+			_ => None,
+		}
+	}
+
+	// This helper determines what requirement and dependency to add to the struct schema based
+	// on the type annotation of the field.
+	// I.E. if a struct has a field named "foo" with a type "OtherStruct", then we want to add
+	// the field "foo" as a required  and the struct "OtherStruct" as a dependency. so the result is
+	// a tuple (required, dependency)
+	fn extract_struct_schema_dependencies(
```
yoav-steinberg, line 259: Maybe `extrac_struct_field_schema_dependency` (note the added field, because we're dealing with a single field and removal of the plural "dependencies" because we're looking for a single dependency for a field).

libs/wingc/src/jsify.rs
```diff
@@ -227,10 +227,71 @@ impl<'a> JSifier<'a> {
...
+					Some((None, deps.1))
+				} else {
+					None
+				}
+			}
+			_ => None,
+		}
+	}
+
+	fn jsify_struct_field_to_json_schema_type(&self, typ: &TypeAnnotationKind, field_name: &String) -> String {
```
yoav-steinberg, line 279: `field_name` seems to never be used

libs/wingc/src/jsify.rs
```diff
@@ -227,10 +227,71 @@ impl<'a> JSifier<'a> {
...
+			TypeAnnotationKind::UserDefined(udt) => {
+				format!("\"$ref\": \"/{}\"", udt.root.name)
+			}
+			TypeAnnotationKind::Array(t) => {
+				format!(
+					"type: \"array\", items: {{ {} }}",
+					self.jsify_struct_field_to_json_schema_type(&t.kind, field_name)
+				)
+			}
+			TypeAnnotationKind::Optional(t) => self.jsify_struct_field_to_json_schema_type(&t.kind, field_name),
```
yoav-steinberg, line 293: Where in the schema do we specify this is optional?

libs/wingc/src/jsify.rs
```diff
@@ -227,10 +227,71 @@ impl<'a> JSifier<'a> {
...
+			_ => "type: \"null\"".to_string(),
```
yoav-steinberg, line 294: What about all the other types? Map/Set? Another Json?

libs/wingc/src/jsify.rs
```diff
@@ -411,6 +472,9 @@ impl<'a> JSifier<'a> {
 								ExprKind::Reference(Reference::InstanceMember { object, .. }) => {
 									self.jsify_expression(&object, ctx)
 								}
+                ExprKind::Reference(Reference::TypeMember { .. }) => {
```
yoav-steinberg, line 475: spaces instead of tabs. Also I think there's an issue here if the type name is under a namespace:
`MyModule.MyStruct.field`

libs/wingc/src/jsify.rs
```diff
@@ -758,9 +822,127 @@ impl<'a> JSifier<'a> {
 				// This is a no-op in JS
 				CodeMaker::default()
 			}
-			StmtKind::Struct { .. } => {
-				// This is a no-op in JS
-				CodeMaker::default()
+			StmtKind::Struct { name, fields, extends } => {
```
yoav-steinberg, line 825: I think we're at the point this can be moved into a `jsify_struct`

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
yoav-steinberg, line 1: I don't see the actual struct `Bar.Struct.js` files in the snapshot.

libs/wingc/src/jsify.rs
```diff
@@ -758,9 +822,127 @@ impl<'a> JSifier<'a> {
...
+				// To allow for struct validation at runtime this will generate a JS class that has a static
+				// getValidator method that will create a json schema validator.
+
+				let mut code = CodeMaker::default();
+				let json_schema_require = "require(\"jsonschema\")".to_string();
```
yoav-steinberg, line 830: I'm wondering how this work in inflight, who makes sure that `jsonschema` will be bundled? I see there's a test for this, but don't understand how it works.

hasanaburayyan: Yea that was actually one of the challenges here since we don’t have other examples of bringing in external libraries for the inflight bundle, but if you checkout bundling.ts I added the sdk node modules as a resolvable path so esbuild can bundle it. Then esbuild does it’s magic when it sees the require() in the _toInflightType.

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
hasanaburayyan: Oh good point, let me check our snapshot stuff not sure why it’s not being included

libs/wingc/src/jsify.rs
hasanaburayyan: Everything in the schema is optional, unless required. So I add a required entry for everything that is not optional. If the snapshots had the generated schema file it would be clearer 😅 I’ll get that added

libs/wingc/src/jsify.rs
yoav-steinberg: Nice solution! I was struggling with something similar a while back and just gave up... So I guess we can now share external deps required by the language between inflight an preflight :clap:

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
hasanaburayyan: @yoav-steinberg PR has been updated to include struct files in snapshots

libs/wingc/src/jsify.rs
hasanaburayyan: Good call, I tunnel visioned on the legal json fields. I have added the other types and set up some tests

libs/wingc/src/jsify.rs
hasanaburayyan: hmm I think the namespaces issue is going to be a lot more complex.. IE if I have the following two wing files
lib.w
```js
struct Person {
  name: str;
  age: num;
}
```

main.w
```js
bring "./lib.w" as lib;

struct Advisor extends lib.Person {
	empId: str;
}
```

Then when we make it to jsification the We end up with a parent that looks like this
```js
 UserDefinedType {
    root: Symbol {
        name: "lib",  // <=== this is a problem
        span: WingSpan {
```

I have some ideas, but I think I would prefer to introduce them in a follow-up PR since this one is already pretty large. If there arent any objections?

libs/wingc/src/jsify.rs
hasanaburayyan: Yea Mark saved me from my bundling pit of sorrow 😁 Id like to get it to a whitelist sort of sharing, but that seems to be harder than it sounds. For now if its in the bundled dependencies of the SDK it can be used.

libs/wingc/src/jsify.rs
yoav-steinberg: ok, but maybe add a comment there or a failing test that should pass in the future. I think the issue isn't only related to bringing other .w files. We also generate namespaces when importing a JSII lib. And JSII lib's with what's known as `datatype` interfaces generate wing structs (under the brought namespace). So you can actually test this now. Trying something like
```wing
bring cloud;
let x = cloud.BucketProps.fromJson( { public: false } );
```

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/events.w_compile_tf-aws.md
```diff
@@ -1,5 +1,51 @@
...
+      }});
+      addDependencies(this._getDependencies());
+      return v;
+    }
+    static validate(obj) {
+      const result = this._getValidator().validate(obj, this._schema);
+      return result.errors;
+    }
+    static _toInflightType() {
+      return fromInline(`require("./CheckHitCountOptions.Struct.js")(require("jsonschema"))`);
```
yoav-steinberg, line 41: Any chance this can be replaced with `__filename`? Something like:
```suggestion
      return fromInline(`require("./"+__filename)(require("jsonschema"))`);
```
It'll be great if we can move all this boilerplate somewhere instead of duplicating it in every struct file, not critical though.

libs/wingsdk/src/std/struct.ts
```diff
@@ -0,0 +1,31 @@
+import { T1 } from "./generics";
+import { Json } from "./json";
+import { Code, InflightClient } from "../core";
+
+/**
+ * Base class for all structs
+ *
+ * @typeparam T1
+ */
+export class Struct {
```
yoav-steinberg, line 10: Can we expose the struct's schema through public API: `getSchema`? I think all we need to do is a macro that'll create the validator and returns its internal `v.schemas` field.
I'd the fact that struct's a "schema"''d JSons to be a wing "feature".

yoav-steinberg (approved):
Very nice work using schemas

docs/docs/04-standard-library/02-std/api-reference.md
```diff
@@ -2051,6 +2051,37 @@ The length of the string.
 ---
 
 
+### Struct <a name="Struct" id="@winglang/sdk.std.Struct"></a>
+
+Base class for all structs.
```
Chriscbr, line 2056: I think we want to avoid confusing people by saying that structs are classes. Maybe we could tweak the description to something like
```suggestion
### Struct <a name="Struct" id="@winglang/sdk.std.Struct"></a>

Shared behavior for all structs.
```

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
```diff
@@ -0,0 +1,794 @@
...
+
+## Advisor.Struct.js
+```js
+module.exports = function(validator_req, fromInline) {
+  class Advisor {
+    static _schema = {
+      id: "/Advisor",
+      type: "object",
+      properties: {
+        ...require("./Person.Struct.js")()._schema.properties,
```
Chriscbr, line 11: I'm curious what's the reason we can't inline all of the properties / constraints into a single schema? Wouldn't that generated code be easier to debug (and run faster, since there's no function calls?)

tools/hangar/__snapshots__/test_corpus/valid/structs.w_compile_tf-aws.md
```diff
@@ -1,5 +1,212 @@
...
+      const addDependencies = ((deps) => { for (const [key, value] of Object.entries(deps)) {
+        v.addSchema(value, key);
+      }});
+      addDependencies(this._getDependencies());
+      addDependencies(require("./A.Struct.js")()._getDependencies());
+      return v;
+    }
+    static validate(obj) {
+      const result = this._getValidator().validate(obj, this._schema);
+      return result.errors;
```
Chriscbr, line 78:
```suggestion
      return this._getValidator().validate(obj, this._schema).errors;
```

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
```diff
@@ -0,0 +1,794 @@
...
+    }
+    static _getValidator() {
+      const v = new validator_req.Validator();
+      const addDependencies = ((deps) => { for (const [key, value] of Object.entries(deps)) {
+        v.addSchema(value, key);
+      }});
+      addDependencies(this._getDependencies());
+      addDependencies(require("./Person.Struct.js")()._getDependencies());
+      return v;
+    }
```
Chriscbr, line 374: Maybe we can remove the indirection of `_getDependencies()`? I think we would just need a transitive closure of all schemas our struct reaches/uses, e.g.
```suggestion
    static _getValidator() {
      const v = new validator_req.Validator();
      v.addSchema(require("./Advisor.Struct.js")()._schema, "/Advisor");
      v.addSchema(require("./Course.Struct.js")()._schema, "/Course");
      v.addSchema(require("./CourseResults.Struct.js")()._schema, "/CourseResults");
      v.addSchema(require("./Person.Struct.js")()._schema, "/Person");
      v.addSchema(require("./Date.Struct.js")()._schema, "/Date");
      return v;
    }
```

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
hasanaburayyan: Not a super strong reason, I just opted for the function calls to avoid having to do a bunch of env lookups in the jsify methods. Looking up each extends and any fields that were structs felt heavy, and personally I didnt think the requires were too terrible to follow. (but I have been looking at this for a bit so im biased 😅 )

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
hasanaburayyan: @Chriscbr So I dont have a strong opinion on this, but my initial reason for this route was to avoid putting the need to do quite a bit of env lookups in the jsification steps.

tools/hangar/__snapshots__/test_corpus/valid/structs.w_compile_tf-aws.md
hasanaburayyan: Oh yea, originally I was doing some error message formatting here but now I can just oneline it.

tools/hangar/__snapshots__/test_corpus/valid/struct_json_conversion.w_compile_tf-aws.md
Chriscbr: Gotcha - yeah, my preference would be to shift more work onto the compiler since we're otherwise just deferring more work to runtime. But I'm okay having it taken on in a separate issue, especially if you think it requires any heavy refactoring

hasanaburayyan: Yea I see the benefit. I have a follow-up PR im planning to do with `tryFromJson()` Ill play with it there and see how tough it is. 👍

libs/wingsdk/src/std/struct.ts
hasanaburayyan: I was thinking something similar which is why I made _getValidator() its own function rather than creating it in the validate method. Should make supporting other features easier.

examples/tests/valid/struct_json_conversion.w
```diff
@@ -0,0 +1,217 @@
...
+let student3 = Student.fromJson(jStudent3);
+
+if let additionalData = student3.additionalData {
+  let notes = additionalData.get("notes");
+  assert(notes == "wow such notes");
+} else {
+  assert(false); // shouldnt happen
+}
+
+// struct Student extends Person {
```
eladb, line 166: Why is this commented out?

libs/wingcompiler/src/compile.ts
```diff
@@ -168,12 +168,15 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
   const artifact = await fs.readFile(artifactPath, "utf-8");
   log?.("artifact: %s", artifact);
 
+  // TODO: is there a way to whitelist the modules we want to include?
+  const wingSDKNodeModules = join(__dirname, "..", "..", "wingsdk", "node_modules");
```
eladb, line 172: How does this behave in a production install? This depends on the repository structure...

Also, maybe you just want to include a specific dependency here and not all of them?

You can use `require.resolve(dep)` to get its path maybe...

libs/wingsdk/src/shared/bundling.ts
```diff
@@ -21,13 +21,16 @@ export function createBundle(entrypoint: string, outputDir?: string): Bundle {
   mkdirSync(outdir, { recursive: true });
   const outfile = join(outdir, "index.js");
 
+  // TODO: is there a way to whitelist the modules we want to include?
+  const sdkNodeModules = join(__dirname, "..", "..", "node_modules");
```
eladb, line 25: You can rely on node modules location...

libs/wingsdk/src/std/struct.ts
eladb: I think `MyStruct.jsonSchema()` is what the spec says

eladb (changes requested):
Some issue with module resolution.

Also, please update the language reference with the relevant implemented features from the spec (basically copy from [here](https://docs.winglang.io/contributing/rfcs/language-spec) to [here](https://docs.winglang.io/docs/language-reference).

I would also add a list of follow up items in the pull request with tracking issues.

libs/wingcompiler/src/compile.ts
hasanaburayyan: I believe as long as the dependency is a bundled dependency then in the production install the resolution will still work. 
M

eladb: It's a bad practice to rely on the fact that our dependencies are bundled. Let's find a more robust solution.

I think something like passing it the specific dependency would be better. It also means we are not exposing all of the deps internally.

MarkMcCulloh: You should be able to use aliases to map known packages to their path with an [alias](https://esbuild.github.io/api/#alias):

```
alias: {
  "jsonschema": require.resolve("jsonschema"),
},
```

The most robust solution would be a small esbuild plugin, but that could wait for another time if we have issues

hasanaburayyan: I wonder if the more straight forward path is to add the "jsonschema" dependency to both wingsdk and wingcompiler and use `alias` for the preflight code and the bundle code.

examples/tests/valid/struct_json_conversion.w
hasanaburayyan: was leftover debug, the struct definition is uncommented earlier in the file I removed this.

libs/wingcompiler/src/compile.ts
hasanaburayyan: Okay so I managed to figure out how to keep this all in the sdk and not need to do any of this dependency resolution stuff here. 

The std.struct class now has a `_getValidator` method that returns a validator from the `jsonschema` library so no requires to `jsonschema` are needed in the generated code.

hasanaburayyan:
> Also, please update the language reference with the relevant implemented features from the spec (basically copy from [here](https://docs.winglang.io/contributing/rfcs/language-spec) to [here](https://docs.winglang.io/docs/language-reference).

@eladb forgot to update the docs before I re-requested review. Added as a todo but will complete it in the morning.

examples/tests/error/struct_from_json_4.w
```diff
@@ -0,0 +1,29 @@
...
+  age: 22,
+  advisors: [
+    {id: "advisor1", name: "Bob", age: 34},
+    {id: "advisor1", name: "Bob", age: 34},
+  ]
+};
+
+Student.fromJson(invalidAdvisorInArray);
+//              ^ ERROR: unable to parse Student:
+// - instance.advisors contains duplicate item
```
eladb, line 29: It's debatable whether this is an error or just results in a single entry. When you add the same item to a set twice, you don't get an error.

libs/wingsdk/src/std/struct.ts
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @macro ((j, s) => {return s._validate(j)})($args$, $self$)
+   */
+  public static fromJson(json: Json): T1 {
+    json;
+    throw new Error("Macro");
+  }
+
+  /** @internal */
+  public static _getValidator(deps?: { [key: string]: any }): Validator {
```
eladb, line 32: Add doc string. What is deps?

eladb, line 32:
```suggestion
  public static _getValidator(deps: { [key: string]: any } = {}): Validator {
```

This way you can avoid the `if`

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/events.w_compile_tf-aws.md
```diff
@@ -1,5 +1,48 @@
...
+        "count",
+      ]
+    }
+    static _getDependencies() {
+      return {
+        "/Source": require("./Source.Struct.js")()._schema,
+        ...require("./Source.Struct.js")()._getDependencies(),
+      }
+    }
+    static _validate(obj) {
```
eladb, line 29: Do we really need this function here? Seems like all structs will have the exact same validate method, which implies it should be under the `std.Struct` type.

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/events.w_compile_tf-aws.md
```diff
@@ -1,5 +1,48 @@
 # [events.w](../../../../../../examples/tests/sdk_tests/bucket/events.w) | compile | tf-aws
 
+## CheckHitCountOptions.Struct.js
+```js
+module.exports = function(stdStruct, fromInline) {
+  class CheckHitCountOptions {
+    static _schema = {
```
eladb, line 7: Shouldn't this class have a `fromJson` method?

tools/hangar/__snapshots__/test_corpus/valid/struct_from_json.w_compile_tf-aws.md
```diff
@@ -0,0 +1,761 @@
...
+module.exports = function({ $Student }) {
+  class $Closure1 {
+    constructor({  }) {
+      const $obj = (...args) => this.handle(...args);
+      Object.setPrototypeOf($obj, this);
+      return $obj;
+    }
+    async handle() {
+      const jStudent3 = Object.freeze({"firstName":"struct","lastName":"greatest","enrolled":true,"schoolId":"s3-inflight","dob":Object.freeze({"month":4,"day":1,"year":1999}),"coursesTaken":[Object.freeze({"grade":"B","dateTaken":Object.freeze({"month":5,"day":10,"year":2021}),"course":Object.freeze({"name":"COMP 101","credits":2})}), Object.freeze({"grade":"A","dateTaken":Object.freeze({"month":5,"day":10,"year":2021}),"course":Object.freeze({"name":"COMP 121","credits":4})})]});
+      const studentInflight1 = ((j, s) => {return s._validate(j)})(jStudent3, $Student);
```
eladb, line 366: Curious why isn't this just calling a method `MyStruct.fromJson()`?

What's the motivation for all the macros?

eladb (commented):
Hey!

This is amazing and I know you are keen to get this over the finish line, but there are a few tweaks I think we should make.

The generated struct code and the interaction between it and the std.Struct seem a bit convoluted.

I imagined something like;

Foo.Struct.js

```js
class Foo {
  static get schema() {
    // return a full self-contained json schema, including all nested structs...
  }

  static fromJson(x) {
    return std.Struct.validate(x, this.schema);
  }

  // lifting boilerplate 

}
```

Something along these lines. KISS...

hasanaburayyan:
I think Im going to go ahead and leave this PR open and just add the remaining features for the issue rather than split them into a followup pr

eladb:
> I think Im going to go ahead and leave this PR open and just add the remaining features for the issue rather than split them into a followup pr


That's usually something I regret, especially with a big project like this... I would try to get this merged first and follow up with subsequent PRs.

libs/wingsdk/src/std/struct.ts
hasanaburayyan: I was able to refactor the schema to use $defs https://json-schema.org/understanding-json-schema/structuring.html#defs 

rather than adding external schemas as dependencies to the validator. So this now makes the `getSchema()` method return a self contained schema. 🎉 so this method is now gone 

Note:
In a future iteration I would like to get rid of the require() statements within the schema (when referencing parent fields) but that is going to take some heavier refactoring,

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/events.w_compile_tf-aws.md
hasanaburayyan: Its gone in its place is a `fromJson` method that just makes a call to std.Struct and passes in the self contained schema

tools/hangar/__snapshots__/test_corpus/valid/struct_from_json.w_compile_tf-aws.md
hasanaburayyan: It does now 😁

libs/wingc/src/type_check.rs
```diff
@@ -4270,6 +4291,29 @@ impl<'a> TypeChecker<'a> {
 							)
 						}
 					}
+					Type::Struct(ref s) => {
+						if property.name == "fromJson" || property.name == "tryFromJson" {
```
eladb, line 4295: put in `"fromJson"` and `"tryFromJson"` in constants...

libs/wingsdk/src/std/struct.ts
```diff
@@ -0,0 +1,65 @@
...
+   * @internal
+   */
+  public static _toInflightType(): Code {
+    return InflightClient.forType(__filename, this.name);
+  }
+
+  /**
+   * Converts a Json to a Struct
+   *
+   * This macro takes a Json object and the Struct definition file name then calls the validate method
```
eladb, line 22: update the description

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/events.w_compile_tf-aws.md
```diff
@@ -1,5 +1,42 @@
 # [events.w](../../../../../../examples/tests/sdk_tests/bucket/events.w) | compile | tf-aws
 
+## CheckHitCountOptions.Struct.js
+```js
+module.exports = function(stdStruct, fromInline) {
+  class CheckHitCountOptions {
+    static getSchema() {
```
eladb, line 7:
```suggestion
    static jsonSchema() {
```

eladb (approved):
LGTM. @yoav-steinberg / @Chriscbr / @MarkMcCulloh please give the final approval

libs/wingc/src/jsify.rs
hasanaburayyan: I added a commented out test and tracking issue: https://github.com/winglang/wing/issues/3792

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -102,6 +103,8 @@ impl<'a> JsiiImporter<'a> {
 					self.wing_types.json()
 				} else if type_fqn == &format!("{}.{}", WINGSDK_ASSEMBLY_NAME, WINGSDK_MUT_JSON) {
 					self.wing_types.mut_json()
+				} else if type_fqn == &format!("{}.{}", WINGSDK_ASSEMBLY_NAME, WINGSDK_STRUCT) {
+					self.wing_types.anything()
```
MarkMcCulloh, line 107: Does this actually happen? Can leave it out if not

libs/wingc/src/type_check.rs
```diff
@@ -4270,6 +4289,32 @@ impl<'a> TypeChecker<'a> {
 							)
 						}
 					}
+					Type::Struct(ref s) => {
+						let from_json = "fromJson";
+						let try_from_json = "tryFromJson";
```
MarkMcCulloh, line 4294: nit

```suggestion
						const from_json: str = "fromJson";
						const try_from_json: str = "tryFromJson";
```

MarkMcCulloh (approved):
🎉

MarkMcCulloh (approved):
🎉

libs/wingc/src/type_check/jsii_importer.rs
hasanaburayyan: I thought it did but it appears it does not, removed 👍

PR #3654, opened 2023-08-01
chore(sdk): optimizing `bucket.exists()` for aws targets

garysassano:
I was randomly thinking that using `ListObjectsV2Command` may not be the most efficient way to check if an object exists in a S3 bucket, since you are asking AWS to look for all the objects that have a specific prefix even though the API returns only one object when filtering with `MaxKeys: 1`. I thought that probably just checking for the metadata of the single object using `HeadObjectCommand` would be enough to determine whether the object exists.

### DIFF
--- a/libs/wingsdk/src/shared-aws/bucket.inflight.ts
+++ b/libs/wingsdk/src/shared-aws/bucket.inflight.ts
@@ -1,6 +1,8 @@
 import { Readable } from "stream";
 import * as consumers from "stream/consumers";
 import {
+  HeadObjectCommand,
+  HeadObjectCommandOutput,
   DeleteObjectCommand,
   GetObjectCommand,
   ListObjectsV2Command,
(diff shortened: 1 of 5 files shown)


garysassano:
Not sure if this optimization is overkill to be honest, and it would require to update also the tests using mock AWS resources.

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -27,13 +29,13 @@ export class BucketClient implements IBucketClient {
    * @param key Key of the object
    */
   public async exists(key: string): Promise<boolean> {
-    const command = new ListObjectsV2Command({
+    const command = new HeadObjectCommand({
```
tsuf239, line 32: Nice! That's a great initiative 💫 
Can you please replace this in the tests as well? 
(`wing/libs/wingsdk/test/shared-aws/bucket.inflight.test.ts`)

Chriscbr (approved):

libs/wingsdk/src/shared-aws/bucket.inflight.ts
Chriscbr, line 32: Do the AWS permissions need to be updated at all? see here https://github.com/winglang/wing/blob/f8a5abf8a7a6a143f3e089d1fbe2c2ef47904bfc/libs/wingsdk/src/shared-aws/permissions.ts#L84

garysassano: Yes, in order to retrieve the object's metadata, the `HeadObjectCommand` requires also the `s3:GetObject` and `s3:GetObjectAttributes` permissions. Thanks for reminding me, I've fixed it now.

PR #3677, opened 2023-08-02
fix(compiler): cannot bring modules more than once (attempt 2)

### LINKED ISSUE #3509: Bad error message on bring/import cycles

### LINKED ISSUE #3511: Cannot reuse dependency that is used elsewhere (diamond dependency problem)

### LINKED ISSUE #3578: Cannot bring the same module from multiple files

Chriscbr:
Re-roll of the changes in #3605, but with more care taken to support the language server. The previous version of this PR  introduced some bugs where the LSP was using the text from disk instead of the text in the user's editor, resulting in invalid text completions and crashes. To fix this, I had to rethink the original multi-file parsing solution so that we store the latest file text from the LSP in memory.

----

Refactors how the Wing compiler handles `bring` for local Wing files in order to support more complex `bring` use cases, like allowing multiple files to `bring` the same module.

The high level implementation approach is instead of generating a single, monolithic AST, we augment the parser so that after a Wing source file is parsed, an isolated AST and a list of dependent Wing files is produced. Using this, we can write a function that parses an entire Wing program given an entrypoint, and use the `petgraph` crate to produce topological ordering of the files for type checking, and each AST is independently type checked.

Fixes #3511
Fixes #3578
Fixes #3509

## Checklist

- [x] Tests added (always)


libs/wingc/src/lsp/sync.rs
```diff
@@ -1,42 +1,59 @@
...
+	}
+}
+
 thread_local! {
 	/// When consumed as a WASM library, wingc is not in control of the process/memory in which it is running.
 	/// This means that it cannot reliably manage stateful data like this between function calls.
 	/// Here we will assume the process is single threaded, and use thread_local to store this data.
-	pub static FILES: RefCell<HashMap<Url,FileData>> = RefCell::new(HashMap::new());
+	pub static WING_TYPES: RefCell<Types> = RefCell::new(Types::new());
+	pub static PROJECT_DATA: RefCell<ProjectData> = RefCell::new(ProjectData::new());
```
Chriscbr, line 56: I was running into some hairy borrow checking errors (due to trying to mutably borrow `Types` and immutably borrow an AST `Scope` at the same time) at some point - the best solution I could come up with was to put "Types" into its own thread-local variable.

libs/wingc/src/parser.rs
```diff
@@ -131,23 +131,218 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
 	"Object",
 };
 
+#[derive(Default)]
+pub struct FileGraph {
+	graph: petgraph::graph::DiGraph<PathBuf, ()>,
+	path_to_node_index: IndexMap<PathBuf, petgraph::graph::NodeIndex>,
+}
+
+impl FileGraph {
```
Chriscbr, line 140: Write unit tests for this?

libs/wingc/src/parser.rs
```diff
@@ -131,23 +131,218 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
+	files.update_file(&init_path, source_text);
+
+	// Update our collections of trees and ASTs and our file graph
+	tree_sitter_trees.insert(init_path.to_owned(), tree_sitter_tree);
+	asts.insert(init_path.to_owned(), ast);
+	file_graph.update_file(init_path, &dependent_wing_files);
+
+	// Track which files we have parsed and which still need parsing
+	let mut files_to_parse = dependent_wing_files;
+	let mut files_parsed = HashSet::new();
```
Chriscbr, line 243: This hashset should be initialized with init_path

libs/wingc/src/parser.rs
```diff
@@ -131,23 +131,218 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
+
+	// Parse all remaining files in the project (skipping files we have seen before)
```
Chriscbr, line 245:
```suggestion
	// Parse all remaining files in the project (skipping files we have already parsed)
```

libs/wingc/src/parser.rs
```diff
@@ -131,23 +131,218 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
+	while let Some(file_to_parse) = files_to_parse.pop() {
+		if files_parsed.contains(&file_to_parse) {
+			continue;
+		}
+
+		// TODO: skip the re-parsing a file if we have already have an AST saved from a previous compilation
+		// This requires more refactoring since currently we store some type-checking information
+		// (SymbolEnv) on the ASTs, so we can't just re-use the ASTs
```
Chriscbr, line 253: Create issue

libs/wingc/src/parser.rs
```diff
@@ -131,23 +131,218 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
...
 	pub source: &'a [u8],
 	pub source_name: String,
 	pub error_nodes: RefCell<HashSet<usize>>,
-	files: RefCell<&'a mut Files>,
 	// Nesting level within JSON literals, a value larger than 0 means we're currently in a JSON literal
 	in_json: RefCell<u64>,
 	is_in_mut_json: RefCell<bool>,
 	is_in_loop: RefCell<bool>,
+	/// Track all file paths that have been found while parsing the current file, along with the
+	/// span of the `bring` statement that referenced them.
```
Chriscbr, line 336:
```suggestion
	/// Track all file paths that have been found while parsing the current file
```

MarkMcCulloh:
Testing locally in vscode, opening files in certain sequences causes a panic. Couple examples:

- open examples/tests/valid/baz.w then examples/tests/valid/bring_local.w
- open examples/tests/valid/subdir/bar.w then examples/tests/valid/subdir/empty.w

The panic is the same either way:

```
panicked at 'assertion failed: (*env).is_none()', libs/wingc/src/ast.rs:678:9) during type-checking
```

Editing the file with the panic makes it go away

Chriscbr:
ah, I think I might know what's causing this. Yeah it all goes back to how I'm trying to re-use ASTs even after they've been type checked which isn't really supported (https://github.com/winglang/wing/issues/3700). Maybe it makes sense to try fixing that first...

Chriscbr:
OK - fixed @MarkMcCulloh

libs/wingc/src/lsp/sync.rs
```diff
@@ -71,83 +98,99 @@ pub unsafe extern "C" fn wingc_on_did_change_text_document(ptr: u32, len: u32) {
...
-	// Reset diagnostics before new compilation (`partial_compile` can be called multiple)
+/// Runs several phases of the wing compiler on a file, including: parsing, type checking, and lifting
+/// `ProjectData` is passed with results from previous compilations, and is updated with the results of this compilation.
+fn partial_compile(
+	source_path: &Path,
+	source_text: String,
+	mut types: &mut Types,
+	jsii_types: &mut TypeSystem,
+	project_data: &mut ProjectData,
+) {
```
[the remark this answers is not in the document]
Chriscbr: @yoav-steinberg That's mostly correct. Some of the parsing results are cached between executions of `partial_compile`, but everything after that (type-checking onwards) is re-done from scratch each time.

More specifically, we re-type-check "known" files every time. That includes all files the user has opened in the LSP, as well as any unopened Wing source files they depend on.

As far as what's reset and what's kept:
- `TypeSystem` is kept between compilations since it stores all of the information about JSII dependencies, and we don't want to reload / parse that every time.
- `ProjectData` (renamed from `FileData`) stores several pieces of information used by the LSP hooks, like the text, tree sitter tree, and AST for each source file. (Right now all of these pieces of data are just collected in a struct for convenience.) A lot of these that pertain to parsing are incrementally updated on each compilation.

`Types` is currently shared between compilations because our `Scope`'s have pointers to SymbolEnv's that we don't want to invalidate, but that might be addressed by https://github.com/winglang/wing/pull/3712, so we might be able to remove it as a parameter to this function. I'll add a TODO comment

libs/wingc/src/file_graph.rs
```diff
@@ -0,0 +1,159 @@
...
+				PathBuf::from("d"),
+				PathBuf::from("c"),
+				PathBuf::from("b"),
+				PathBuf::from("a")
+			]
+		);
+	}
+
+	#[test]
+	fn toposort_cycle() {
```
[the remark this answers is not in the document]
Chriscbr: I'll add this

libs/wingc/src/lsp/snapshots/completions/struct_literal_value.snap
```diff
@@ -1,64 +1,16 @@
...
+  kind: 5
+  detail: num
   documentation:
     kind: markdown
-    value: "```wing\nassert: (condition: bool): void\n```\n---\nAsserts that a condition is true\n\n### Parameters\n- `condition` — The condition to assert"
-  sortText: cc|assert
-  insertText: assert($0)
+    value: "```wing\ny: num\n```"
+  sortText: "ab|y:"
+  insertText: "y: $1"
```
MarkMcCulloh, line 11: This is a regression because we should be showing normal expression completion. This is probably related to the preceding text calculation (in this case, the last character should be ":")

Chriscbr: Fixed

libs/wingc/src/file_graph.rs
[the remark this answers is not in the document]
Chriscbr: Added

libs/wingc/src/file_graph.rs
```diff
@@ -0,0 +1,159 @@
...
+					.rev()
+					.map(|n| self.graph[n].clone())
+					.collect::<Vec<_>>(),
+			),
+			Err(_) => {
+				// toposort function in the `petgraph` library doesn't return the cycle itself,
+				// so we need to use Tarjan's algorithm to find one instead
+				let strongly_connected_components = petgraph::algo::tarjan_scc(&self.graph);
+				Err(
+					strongly_connected_components[0]
```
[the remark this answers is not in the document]
Chriscbr: Yeah, we're not even returning an exact cycle, just a list of files that we know are part of _some_ cycle (i.e. from one of these files, it's possible to reach all of the others). I think the maximum possible unique cycles in a graph is exponential in the number of nodes, so this should be good enough as a first approximation anyways.

Issue #3689, opened 2023-08-03
RFC: Wing Cloud Preview Environments


ainvoner:
# Introduction

The **Wing Cloud Preview Environments** project is designed to achieve two key objectives. Firstly, it aims to offer a a preview environment solution for those interested in using winglang. Secondly, it seeks to enhance the winglang ecosystem and its toolchain by building a production application using the language.

# Wing Cloud Preview Environments

see [rendered file](https://github.com/winglang/wing/blob/rfc-preview-environment/docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md)

Issue #2227, opened 2023-04-24
Webhook Local Development Experience (ngrok-like capabilities)


eladb:
### Feature Spec

Wing now has built-in support for developing applications that receive webhooks (such as slack bots, github apps and more).

DX to be defined

### Use Cases

Webhook development

### Implementation Notes

Basically, we will need some kind of support for an ngrok-like service where users are able to allocate some stable public dns name on the internet and requests will be routed to their local machine.

### Component

SDK, Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

eladb:
Changing to p1

skorfmann:
btw: a hacky workaround for this could be something like https://gist.github.com/skorfmann/c3bb0f53e5f244cb7b1f13f15c6a499b in combination with [Tailscale Funnel](https://tailscale.com/kb/1223/tailscale-funnel/)

eladb:
super keep

staycoolcall911:
Please look at the thread #2271

eladb:
https://vt.tiktok.com/ZSLKXLfWU/

Chriscbr:
@eladb Is this solved by the ngrok winglib?

eladb:
We are looking into a more holistic approach. @eladcon is on it.

staycoolcall911:
We have that implemented via tunnels. Closing

PR #3696, opened 2023-08-03
rfc: wing cloud preview environments

ainvoner:
View the [rendered file](https://github.com/winglang/wing/blob/rfc-preview-environment/docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md).

The Wing Cloud Preview Environments project is designed to achieve two key objectives. Firstly, it aims to offer a preview environment solution for those interested in using winglang. Secondly, it seeks to enhance the winglang ecosystem and its toolchain by building a production application using the language.

Resolves: https://github.com/winglang/wing/issues/3689


docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+Developer reading our docs can easily start integrate Wing Preview Environments in his repository.
+In our docs we have information about:
+1. What is Wing Preview Environment
+2. Getting started instructions
+4. How can I configure multiple entry points
+5. Populate environment with initial data
+6. Running tests
+
+### Security
+1. The preview environments shouldn’t be able to access files from other preview environments
```
skyrpex, line 84: I guess flyio got us covered here. If every environment uses a container, it should be already safe by default.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Self-Cleaning Environments Mechanism
+To ensure efficient resource utilization:
+1. Preview environments associated with stale PRs (without access or code changes for over 30 days) will automatically deactivate.
+2. The PR's preview environment comment will indicate its inactive status.
+3. Changes in code will trigger redeployment.
+
+#### Additional Features
+Other valuable features include:
+1. Running tests on pull requests, noting that tests will operate within isolated environments to prevent disruption for other users.
```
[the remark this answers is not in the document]
eladb: From a UX perspective, I think posting the test results in the PR as part of the comment would be freaking amazing.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+2. Configuration of multiple entry points files for creating multiple preview environments in a single PR
+3. Populate a preview environment with an initial data
+
```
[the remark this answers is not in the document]
eladb: Maybe for MVP we don't even need the web UI

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+### Analytics and Logs
+To improve the product we should collect data that we can analyze
+1. Collect usage analytics
+2. use OpenTelemetry for collecting metrics, logs, traces
+
```
eladb, line 73: What about monitoring?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+4. How can I configure multiple entry points
+5. Populate environment with initial data
+6. Running tests
+
+### Security
+1. The preview environments shouldn’t be able to access files from other preview environments
+2. Only authenticated users should be able to use Wing Preview Environments
+
+## Technical Design
+TBD
```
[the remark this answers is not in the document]
skyrpex: @eladcon can we create docker containers in flyio? also, can we open multiple ports to the outside?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Developer Experience
+A developer using winglang has the capability to integrate preview environments with his/her repository.
+
+#### Installation
+The installation process is straightforward using GitHub application:
+
+1. Goto https://wing.cloud/preview website and click "Deploy with Wing"
+2. Complete GitHub authentication.
+3. Grant repository access permission.
```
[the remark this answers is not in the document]
ainvoner: In my mind it should be something similar to Vercel signup as well (see image). What you describe it the flow in the Vercel dashboard which I think we should not handle in this product MVP.

Screenshot 2023-08-06 at 10 24 26

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+#### Preview Environment for production branch
+For the production branch, there is a single preview environment that will constantly be up to date with the latest code committed.
+A link to the preview environment is available in the repository main page ("About" section).
+
+#### Preview Environments for Pull Requests
+Upon each creation of a pull request, an automatic comment will be added to the PR and will guide the developer to a dedicated preview environment. The deployed preview environment will have a unique url for easy identification.
+To provide a streamlined process for updating a pull request preview environment, each PR code changes will redeploy the preview environment and also will:
+1. provide a real-time build and deployment updates
+2. ensure a consistent url for the preview environment
+3. provide the ability to download the deployment logs
```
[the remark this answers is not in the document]
eladb: What about test output?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+#### Installation
+The installation process is straightforward using GitHub application:
+
+1. Goto https://wing.cloud/preview website and click "Deploy with Wing"
+2. Complete GitHub authentication.
+3. Grant repository access permission.
+
+#### Preview Environment for production branch
+For the production branch, there is a single preview environment that will constantly be up to date with the latest code committed.
+A link to the preview environment is available in the repository main page ("About" section).
```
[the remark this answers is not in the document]
ainvoner: we can. this is how vercel is doing it:

Screenshot 2023-08-06 at 10 42 42

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
ainvoner: updated. let me know if it is still not clear

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,88 @@
...
+Allow developers to iterate with their teammates using a shared Wing Console instance.
+Seamlessly integrating with GitHub.
+
+### Developer Experience
+A developer using winglang has the capability to integrate preview environments with his/her repository.
+
+#### Installation
+The installation process is straightforward using GitHub application:
+
+1. Goto https://wing.cloud/preview website and click "Deploy with Wing"
```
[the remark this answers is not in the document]
ainvoner: updated this section, let me know if you still thinks we should change it to "Sign up"

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
ainvoner: I don't think it should be part of the MVP. We should store the user / gh account / repository so we will be able to add this dashboard and functionalities later

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+In order to use Wing Cloud Preview Environments, you should sign up to Wing Cloud.
+The signup and installation process is straightforward using GitHub application:
+
+1. Goto https://wing.cloud website and click "Deploy with Wing Cloud"
+2. Complete GitHub authentication.
+3. Grant repository access permission.
+4. You are now signed up and ready to use Wing Cloud Preview Environments.
+
+#### Preview Environment for production branch
+For the production branch, there is a single preview environment that will constantly be up-to-date with the latest code committed.
```
eladb, line 38: how will URLs look like?

Ideally something like:

```
https://wing.cloud/gh-account/gh-repo/gh-branch
```

e.g.


https://wing.cloud/winglang/docsite/main

---

this should be the same for PRs (because they must be branched), no?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
ainvoner: added to the rfc, although I think it should be a P2. adding the test results to the PR will add complexity
when we will have time estimations we will be able to decide if you add this feature or not

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+A link to the preview environment is available in the repository main page ("About" section).
+
+#### Preview Environments for Pull Requests
+Upon each creation of a pull request, an automatic comment will be added to the PR and will guide the developer to a dedicated preview environment. The deployed preview environment will have a unique url for easy identification.
```
eladb, line 42: see above re urls

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+To provide a streamlined process for updating a pull request preview environment, each PR code changes will redeploy the preview environment and also will:
+1. provide a real-time build and deployment updates
+2. each entry point will have a seperated preview environment
+3. ensure a consistent url for each preview environment
+4. provide the ability to download the preview environment deployment logs for debugging purposes in case of failure
+
```
eladb, line 48: what happens if the branch is private? how do we ensure other people can't access it? do we use github auth there too?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+PR comment example:
+
+| Entry Point     | Status | Preview | Updated (UTC)        |
+|-----------------| ------------- | ------------- |----------------------| 
+| main.w          |  ✅ Ready ([logs](https:/main-pr.wing-preview.com/logs)) | [Visit Preview](https:/main-pr.wing-preview.com) | Jul 31, 2023 8:01am  |
```
eladb, line 53: fix urls

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+#### Multiple Entry Points
+A developer can configure multiple entry points in a single PR.
+Every `*.main.w` file in the repository will be considered as an entry point and will have a dedicated preview environment.
+All preview environments will be available in the PR comment and will be updated upon code changes.
+
+#### Populate Environment with Initial Data
+TBD
+
+#### Running Tests
+Running tests on your environment is easy using the Wing Console tests UI.
```
eladb, line 68: i think we should also run tests automatically and report to the PR

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+This document describes the Wing Cloud Preview Environments feature.
+
+## Introduction
+The **Wing Cloud Preview Environments** project is designed to achieve two key objectives. Firstly, it aims to offer a preview environment solution for those interested in using winglang.
+Secondly, it seeks to enhance the winglang ecosystem and its toolchain by building a production application using the language.
+
+## Requirements
+Create the first production feature of Wing Cloud - Wing cloud Preview Environments.
+Build and Deploy wing applications to a Wing Console simulated environment.
+Allow developers to iterate with their teammates using a shared Wing Console instance.
```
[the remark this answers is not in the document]
ainvoner: updated the text, the reason for adding it as a requirement is to make sure the environment and data will be shared.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+Seamlessly integrating with GitHub.
+
+### Developer Experience
+A developer using winglang has the capability to integrate preview environments with his/her repository.
+
+#### Sign Up and Installation
+In order to use Wing Cloud Preview Environments, you should sign up to Wing Cloud.
+The signup and installation process is straightforward using GitHub application:
+
+1. Goto https://wing.cloud website and click "Deploy with Wing Cloud"
```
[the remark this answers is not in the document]
ainvoner: why login / signup? we would like them to deploy their application, during this process they signup to wing cloud, authenticate via GitHub and grant access to a repository

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: I thought of using subdomains for wing.cloud:
1. Production branch environment url structure is: `https://<gh-repository>-<production-branch-name>.wing.cloud.app`
2. PR preview environments urls structure is: `https://<gh-repository>-<gh-branch>-<pr-number>-<entry-point>.wing.cloud.app`

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+
+| Entry Point     | Status | Preview | Updated (UTC)        |
+|-----------------| ------------- | ------------- |----------------------| 
+| main.w          |  ✅ Ready ([logs](https:/main-pr.wing-preview.com/logs)) | [Visit Preview](https:/main-pr.wing-preview.com) | Jul 31, 2023 8:01am  |
+| another.main.w  |  ✅ Ready ([logs](https:/main-pr.wing-preview.com/logs)) | [Visit Preview](https:/main-pr.wing-preview.com) | Jul 30, 2023 10:01am |
+| failed.main.w   | ❌ Failed ([logs](https:/main-pr.wing-preview.com/logs))  |  | Jul 31, 2023 8:01am  |
+| building.main.w |  🔄 Building ([logs](https:/main-pr.wing-preview.com/logs))  | | Jul 31, 2023 8:01am  |
+| stale.main.w    | ⏸️ Stale  ([lean more](https:/winglang.io/docs/preview))  | | Jul 31, 2023 8:01am  |
+
+#### Multiple Entry Points
```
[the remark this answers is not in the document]
ainvoner: the motivation to support a configuration file is to support our own dog fooding and to run wing console preview environment.
we will need to run the locally built wing to test the console new code

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: perhaps for our MVP we can assume that all users allowed to view the PR can also view the preview env. and logs.
what do you think?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
eladb: I think we need to tackle the port proxying issue generally (and not just for preview environments). See #2227

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,102 @@
...
+In our docs we have the following information:
+1. What is Wing Cloud Preview Environments
+2. Getting started instructions
+3. How can I configure multiple entry points
+4. Populate environment with initial data
+5. Running tests
+
+### Security
+1. The preview environments shouldn’t be able to access files from other preview environments
+2. Only authenticated users should be able to use Wing Preview Environments
```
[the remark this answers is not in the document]
eladb: If a preview environment is working against a private GitHub repository, I think we will need to make sure that only GitHub users that have access to this repo can view the environment, no?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
eladb: vercel's CTA is "start deploying" (which leads to https://vercel.com/new but they also have a "login" and "sign up" buttons on their home page.

let's do the same thing.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
eladb: Let's be careful with introducing config files. If we do, these should be read by the Wing CLI and not by the preview environments system.

* Entryfile(s) should be any file called `main.w` or has the `.main.w` extension.
* Install is always through `npm` right now.
* What's `buildScript`?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
eladb: Why not normal API routes? What's the benefit of using subdomains? I think normal URL routes are more suitable for this need.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
eladb: yes, but just bear in mind that this means that access to the environment must be authenticated and there should be some authorization performed against github.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
ainvoner: I came to a conclusion that we can't avoid a sign up process to Wing Cloud which should not be dependent on the Preview Environment feature.
I think we need a different issue to discuss and define it

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: accessing env? or creating new ones? even vercel doesn't require authentication for private repositories preview environments.
We had to pay for this service when our console was a private repo.
I don't think we should limit viewing environments.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
ainvoner: see https://github.com/winglang/wing/pull/3696#discussion_r1285576005
I don't think we should handle it at the moment

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: Using subdomains is ~preferred~ mandatory here, security wise.

The preview environments will run code created by users. If we serve them under wing.cloud, we expose us to [XSRF attacks](https://en.wikipedia.org/wiki/Cross-site_request_forgery).

Even if this preview environment is served under the Console App control, it's very likely that users may be able to abuse it. Attacks that would be possible are:

- Gathering information from the user previewing the environment
- Executing calls to wing.cloud in the name of this user

I think we can run the subdomains under `*.wing.cloud` but I need to do some research first... there's some HTTP headers that can protect the main site from the subdomains. But the safest option is to use a totally different domain, something like `*.wing-preview.cloud`, so there's no open doors for XSRF.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: The way Vercel works is what makes most sense IMO:

- GitHub authentication is only used to authorize the Wing Cloud GitHub App to have access to the repository
- The Preview Environment can be private but in order to see it, a user needs to be logged in to Wing Cloud and have view permission to the environment (the permissions are controlled by Wing Cloud, not GitHub anymore)

This is why is makes sense to me:

- A GitHub user having read permission to the repository code doesn't mean they should be able to see the preview environment
- Our auth code won't need to closely work along with GitHub's API, so it'll be much easier to maintain

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
skyrpex: I'd implement a USER_x_PROJECT permission on our end, as stated here: https://github.com/winglang/wing/pull/3696#discussion_r1285605394

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
skyrpex: What if we go (again) the Vercel route?

- In Wing Cloud, you can create more than one Preview Environment for a GitHub repository
- When you create a new Preview Environment, you select the repository
- Then, you select the wing entry file, and maybe some more configuration
- The configuration stays on Wing Cloud, so we don't need to define a config file

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: @skyrpex isn't wing.cloud.app is a different domain than wing.cloud? isn't it enough?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
eladb: Building websites is something we should discuss outside of the scope of preview environments. It's relevant for local development as well. Let's open an issue and figure this out separately. Preview environments will use the same mechanism.

Regarding wing version, we *will* have some form of config for Wing, but again, this is not a preview environment concern. Let's open an issue and discuss separately.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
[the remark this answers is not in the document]
ainvoner: I think we should start with a simple gh auth for wing cloud without any extra permissions.
not all signed in users to wing cloud would like to install the preview app in their repo.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
eladb: I am okay to use subdomains if that's the only way to defend against attacks but it should be under the wing.cloud parent

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,249 @@
...
+Production branch environment url structure is: `https://<gh-repository>-<gh-branch>-<entry-point>.wing.cloud.app`
+
+#### Preview Environments for Pull Requests
+
+Upon each creation of a pull request an automatic comment will be added to the PR and will guide the developer to dedicated preview environments.
+provide a streamlined process for updating a pull request preview environment, each PR code changes will redeploy the preview environment and also will:
+
+1. provide a real-time build and deployment status updates
+2. each entry point will have a seperated preview environment
+3. ensure a consistent and unique url for each preview environment. The url structure: `https://<gh-repository>-<gh-branch>-<entry-point>-<pr-number>.wing.cloud.app`
```
skyrpex, line 73: @ainvoner Instead of having an additional subdomain with the PR number, I'd use a random id mixed with repo information such as the repo name and/or org, like Vercel does. It's easier for us to generate a unique ID this way, and also generate an alias for it using `https://<gh-repository>-<gh-branch>-<entry-point>.wingcloud.app`, if available.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,282 @@
...
+
+1. Preview environments will require authentication to view.
+2. Assign users to Teams.
+3. enable preview environments viewers to add comments in the environment itself
+
+## Technical Design
+
+### Technology
+
+- https://wing.cloud for the infrastructure (although it may prove a challenge due to time constrains)
```
ainvoner, line 162: @skyrpex did you mean winglang?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,282 @@
...
+- https://react.dev/ for the frontend
+
+### Architecture
+
+- A GitHub app
+- A Cloud.Function that will be triggered by GitHub events, such as `pull_request.synchronize` or `pull_request.closed`
+- A Flyio account to deploy the preview environments
+- A Cloud.Table to store different data:
+  - The registered users
+  - The Wing Cloud projects and their config
```
ainvoner, line 176: is it relevant for the mvp? (next line as well)

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,282 @@
...
+    fn --> rest[...]
+```
+
+The probot npm package will take care of ensuring the events received are legitimate.
+
+#### Process: Creating Preview Environments
+
+When a GitHub event `pull_request.opened` is received, we will look for `*.main.w` files and create a preview environment for each of them. We will use the Flyio API to create a new project(s) and deploy the code to it (the Flyio specifics may change).
+
+During this process, we will also create new entries in the Cloud.Table to store the Flyio project ID, the GitHub PR ID, and the entry point.
```
ainvoner, line 242: can you please explain why do we need to save this information?
(we discussed about reducing the gh api requests for minimum - this was one reason for saving the PR and the VM ID, for knowing which PR to update if the VM was stale for X amount of time)

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: but we need a different domain... accessing our wing.cloud auth cookie for example would be possible from preview environments if we will use sub domains of wing.cloud.

skyrpex: > I am okay to use subdomains if that's the only way to defend against attacks but it should be under the wing.cloud parent

I'm afraid we'll have to use a different domain altogether. There's multiple security risks involved...

Just to give you an example:
- After signing in to `https://wing.cloud`, there'll be an secure HTTP-only cookie on `wing.cloud` used for authentication
- If the preview environments are `*.wing.cloud`, the browser will send this cookie to them and they'll be able to read your session credentials and stuff on your behalf

There are security measures we could take to prevent attacks, but it'd be like if the whole Wing Cloud's security was hanging from a thread. The smallest mistake would become a security breach. We need to go for something like `*.wingcloud.app`.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: Yes, thanks.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: Not anymore since we don't want configuration yet. I'll rephrase it.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: I rephrased everything a bit, but those reasons are correct. Another reason is that we have to resolve a domain such as `<gh-repository>-<gh-branch>-<entry-point>.wingcloud.app` or `<gh-repository>-<nanoid>.wingcloud.app` to a Flyio domain.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Sign Up to Wing Cloud
+
+1. Goto https://wing.cloud and click on sign up.
+2. sign up to Wing Cloud using GitHub authentication.
+
+### Sign In to Wing Cloud
+
+1. Goto https://wing.cloud and click on sign in.
+2. If a developer was already signed in before, he will be automatically signed in to Wing Cloud.
```
skyrpex, line 35: @ainvoner signing in and signing up are the same process. I'd simplify it to just "Sign In".

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+
+Wing Cloud users can build and deploy Wing applications to an ephemeral (simulated) environment and use the Wing Console UI to interact with it.
+Automatically run tests on the preview environment once created and provide a UI for viewing the test results.
+Allow developers to collaborate with their teammates using a shared environment.
+Seamlessly integrating with GitHub.
+
+### Sign In to Wing Cloud
+
+1. Goto https://wing.cloud and click on sign in.
+2. If a developer was already signed in before, he will be automatically signed in to Wing Cloud.
```
yoav-steinberg, line 30:
```suggestion
2. If a developer was already signed in before, they will be automatically signed in to Wing Cloud.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+3. If a developer was not signed in before, he will be redirected to GitHub authentication and then redirected back to Wing Cloud.
```
yoav-steinberg, line 31:
```suggestion
2. If a developer was already signed in before, they will be automatically signed in to Wing Cloud.
3. If a developer was not signed in before, he will they redirected to GitHub authentication and then redirected back to Wing Cloud.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+### Authentication and Authorization
+
+1. Wing Cloud Preview Environments will use GitHub for authentication and authorization.
+2. Developers will be able to install Wing Cloud Preview Environments on a repository only if they have the GitHub permissions.
+3. Developers will be able to create a PR with a preview environment only if they have the necessary GitHub permissions to the repository **and** their GitHub user is signed up to Wing Cloud.
+4. Everyone can view a preview environment if they have the url.
+
+### Wing Cloud Preview Environments Developer Experience
+
+A developer using winglang has the capability to integrate Wing Cloud Preview Environments with their repository.
```
yoav-steinberg, line 42:
```suggestion
A developer using winglang has the capability to integrate Wing Cloud Preview Environments with their (github) repository.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+
+**Disclaimer**: Developers that have write access and can create PRs but didn't signed-up to Wing Cloud will not be able to create preview environments.
+In this case the above comment will contain a direct link for signing up to Wing Cloud.
+After signing up, the developer will be able to create a preview environment for the PR and one will be created automatically after the next code change.
+
+#### Multiple Entry Points
+
+A developer can configure multiple entry points in a single PR.
+Every `*.main.w` file in the repository will be considered as an entry point and will have a dedicated preview environment.
+All preview environments will be available in the PR comment and will be updated upon code changes.
```
yoav-steinberg, line 92: What is the experience in the console when a PE is updated? Do you get some popup saying "please wait while redeploying PE..."? Or does it break because the backend container died under the hood?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+
+#### Populate Environment with Initial Data
+
+Out of scope for this project.
+
+#### Automatically Running Tests
+
+Upon PR creation and code changes, Wing Cloud Preview Environments will automatically run all tests defined for each entry point.
+For each test run, a new simulator will be created and will be destroyed upon test completion.
+The test results will be available in the PR comment for each entry point.
```
yoav-steinberg, line 102: Might be cool in the console, next to each test you'll automatically have a ✅/❌ based on the test results.
Also what happens if there's a flaky test and the automatic running of tests passes but in the PE it fails? Will it update the comment? Is there a way to mark flaky tests? (Assuming there's some connection between the PE and the test running mechanism).

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+#### Running Tests Manually
+
+Running tests on your environment is easy using the Wing Console tests UI.
+Each test will reload the simulator to create a clean environment.
+
+Once the test is over it will **not** clean up the simulator data, so you can debug the environment and share outputs with your teammates.
+
+#### User-Controlled Environments termination
+
+Closing a PR will terminate the corresponding preview environments.
```
yoav-steinberg, line 113: I really think we'll need some mechanism for suspending stale PE's. If there's important state we might need to store it before suspending. But just spinning up a container per PR means that at some point we'll have lots of containers running forever.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+5. Running tests
+
+### Security
+
+1. The preview environments shouldn’t be able to access files from other preview environments
+2. Only Wing Cloud signed up developers with write access to the repository can create preview environments
+
+## NON Requirements
+
+1. Preview environments will require authentication to view.
```
yoav-steinberg, line 150: What about non-public github repos? I think if we support them the PE needs to be github authenticated.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+2. Assign users to Teams.
+3. enable preview environments viewers to add comments in the environment itself
+
+## Technical Design
+
+### Technology
+
+- https://winglang.io for the infrastructure (although it may prove a challenge due to time constrains)
```
yoav-steinberg, line 158: Would be interesting to understand the main concerns here

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Architecture
+
+- A GitHub app
+- A Cloud.Function that will be triggered by GitHub events, such as `pull_request.synchronize` or `pull_request.closed`
+- A Flyio account to deploy the preview environments
+- A Cloud.Table to store different data:
+  - The registered users
+  - The connected GitHub repositories
+  - The related Flyio projects and environments
```
yoav-steinberg, line 173: * some timer/service for cleanups/stale PRs
* some storage for logs

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+When a GitHub event `pull_request.opened` is received, we will look for `*.main.w` files and create a preview environment for each of them. We will use the Flyio API to create a new project(s) and deploy the code to it (the Flyio specifics may change).
+
+New subdomains will be created for each preview environment:
+
+- `<gh-repository>-<nanoid>.wingcloud.app`
+- `<gh-repository>-<gh-branch>-<entry-point>.wingcloud.app`, as an alias of the above, if available. Otherwise, a random ID could be added somewhere, such as in `<gh-repository>-<gh-branch>-<nanoid>-<entry-point>.wingcloud.app`
+
+During this process, we will also create new entries in the Cloud.Table to store the Flyio project ID, the GitHub PR ID, the entry points and whatever unique IDs we generate for every environment. This data is necessary to later resolve the subdomains to the Flyio projects, to update them later, and to delete them.
+
+The technical details on how the subdomain resolution will work at DNS level are still to be determined.
```
yoav-steinberg, line 245: Did we ever discuss a `cloud.Dns` abstraction? Wouldn't this be useful here?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,278 @@
...
+A link to the preview environment is available in the repository main page ("About" section).
+Production branch environment url structure is: `https://<gh-repository>-<gh-branch>-<entry-point>.wingcloud.app`
+
+#### Preview Environments for Pull Requests
+
+Upon each creation of a pull request an automatic comment will be added to the PR and will guide the developer to dedicated preview environments.
+provide a streamlined process for updating a pull request preview environment, each PR code changes will redeploy the preview environment and also will:
+
+1. provide a real-time build and deployment status updates
+2. each entry point will have a seperated preview environment
```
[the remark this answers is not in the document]
Chriscbr: I think multiple entry points sounds OK to me - if there are too many entry points detected, maybe we can just throw an error (or only create the first 5 etc.). But it also feels like a P2 requirement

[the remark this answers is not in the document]
ainvoner: it is a good point, however we agreed to handle it post mvp. see: https://github.com/winglang/wing/pull/3696#discussion_r1285184277

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: for the MVP we decided to do what Vercel is doing, which is not to have authentication for PE by default. In order to have auth in Vercel PE you should specifically enable the feature and pay extra for it.
We will support auth protected envs. post MVP

yoav-steinberg: So in that case a private vercel repo has a public preview env? If so maybe we need some mangling of the DNS? What does vercel do? Or perhaps just warn the user that since they signed up to wing cloud and the repo is private they should be aware that we're deploying a public PE.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: There's no signal of change on the frontend. It'd be a nice feature, though!

The websocket connection will break, so events from backend to frontend will stop working. The rest of the information, which is fetched via HTTP requests, will still work but there will be a drastic change because the environment will be wiped clean (eg, no files in buckets, no events, no logs, etc).

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: There's an API that we can call to update this comment. I think we can send markdown. And, AFAIK, we can update the comment anytime.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: I fear hitting a wall but it's being OK so far :)

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: I don't recall, but sure it'd be helpful.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: Correct, will update the list.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: We can, although I'll need to design that. There's, again, security implications we have to handle.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: added this as a requirement, @yoav-steinberg good point

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: added to the non requirements section in this rfc. we can priorities it later

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
ainvoner: this is how vercel is doing it:
https://vercel.com/docs/concepts/deployments/deployment-protection#vercel-authentication

TL;DR
- Free to all plans:
enable vercel authentication to all deployed environment. users should be authenticated in vercel before viewing an environment.
- paid
password protected. define specific password for each project.

I think we can start with the first option. WDYT?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,285 @@
...
+- https://fly.io for the preview environments
+- https://react.dev/ for the frontend
+
+### Architecture
+
+- A GitHub app
+- A Cloud.Function that will be triggered by GitHub events, such as `pull_request.synchronize` or `pull_request.closed`
+- A Flyio account to deploy the preview environments
+- A Cloud.Table to store different data:
+  - The registered users
```
ainvoner, line 173: @skyrpex we should start thinking about how to group users and preview environments.
I've added those fields so we will have the needed data in our DB, and once we will start implementing the environments dashboard we will already have the needed data per user and env.
When thinking about it, we will also need a table to know which env. belongs to which gh project / repo / user.

WDYT?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### Preview Environments for Pull Requests
+
+Upon each creation of a pull request an automatic comment will be added to the PR and will guide the developer to dedicated preview environments.
+provide a streamlined process for updating a pull request preview environment, each PR code changes will redeploy the preview environment and will also:
+
+1. Provide a real-time build and deployment status updates
+2. While preview environment is being updated, the user that is currently using the environment will be notified and will be able to refresh the page to get the latest changes.
+3. Each entry point will have a seperated preview environment
+4. Ensure a consistent and unique url for each preview environment. The url structure: `https://<gh-repository>-<gh-branch>-<entry-point>.wingcloud.app`
```
polamoros, line 85: Will we have the option to also have a URL per commit like Vercel does?

Will the 'entry-point' be the full path of the file? (will not be unique otherwise)

I think <entry-point>-<gh-branch>-<repo>.wingcloud.app will be more readable. 

For example:
```
entry1-add-new-feat-my-app.wingcloud.app
entry2-add-new-feat-my-app.wingcloud.app
```

vs

```
my-app-add-new-feat-entry1.wingcloud.app
my-app-add-new-feat-entry1.wingcloud.app
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,329 @@
...
+    queue[(Events Queue)]
+    fn2[Enqueue Events Fn]
+    ghapp -- Events --> fn2
+    fn2 --> queue
+    queue --> fn
+    gh -- Integration --> ghapp
+    fn --> rest[...]
+```
+
+The probot npm package will take care of ensuring the events received are legitimate.
```
polamoros, line 279: What part of the `github -> wingcloud` flow will be managed by the probot and which part by the wing project?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,285 @@
...
+- **Stage Date**: 2023-08-03
+
+This document describes the Wing Cloud Preview Environments feature.
+
+## Introduction
+
+The **Wing Cloud Preview Environments** project is designed to achieve two key objectives. Firstly, it aims to offer a preview environment solution for those interested in using winglang.
+Secondly, it seeks to enhance the winglang ecosystem and its toolchain by building a production application using the language.
+
+## Requirements
```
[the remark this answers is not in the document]
skyrpex: This implies implementing some basic authorization roles so different people can access/maintain the projects.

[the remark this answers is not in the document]
eladb: Yes I think there's no way around this

eladb:
@skyrpex @ainvoner it will be very useful to create an entity relationship model for the backend. I think it will help align everyone and we can make decisions around which entities we have in the MVP and how deep we go.

skyrpex:
> 

I'm on it.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,381 @@
...
+        UserId userId
+        string roleType
+    }
+    Project {
+        ProjectId projectId
+        UserIdOrTeamId accountId
+        string name
+        string githubRepoId
+        string wingEntrypoint
+        map environmentVariables
```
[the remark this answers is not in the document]
skyrpex: Yes, I couldn't figure how to describe such data structure in mermaid. Vercel also does this:

image

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+
+Wing Cloud will be integrated with GitHub.
+
+### Wing Cloud User Experience
+
+#### Sign-In to Wing Cloud (#1)
+
+1. On https://wing.cloud website click on sign in.
+2. Authenticate using your GitHub account.
+3. If you were already signed in, you will see the Wing Cloud dashboard (under https://wing.cloud directly).
```
ekeren, line 33: It make more sense to me to have `app.wing.cloud` or something like that.

ekeren: I understand there where already discussion about this, I am fine with what you decide

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+5. For each preview environment, you can see the deployment logs and the test results and logs.
+
+#### Create a new personal Project (#4)
+
+1. To create a new project, click on "New Project" under "Personal Account".
+2. Choose between the GitHub repositories that you have access to.
+3. Define your secret values.
+4. Make sure there's a `main.w` file in the root of the repository.
+5. Click on "Deploy".
+6. You'll be redirected to your new project's page (https://wing.cloud/-user/-project).
```
ekeren, line 55: Personal account is mentioned here and also `-user` in the path. 
Did you guys have the organization vs personal account discussion? 

From my experience this is a complex topic, especially when you start to think about analytics, permission, multi organizations and migrating projects. 

I suggest we spend some time together on this topic to make sure we start as simple as possible.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+3. Define your secret values.
+4. Make sure there's a `main.w` file in the root of the repository.
+5. Click on "Deploy".
+6. You'll be redirected to your new project's page (https://wing.cloud/-team/-project).
+
+#### Preview Environment for production branch (#7)
+
+For the repository production branch, there is a single preview environment that will constantly be up-to-date with the latest code committed.
+A link to the preview environment is available in the repository main page ("About" section).
+Production branch environment url structure is: `https://wing.cloud/-account/-project/-branch`
```
ekeren, line 76: What is `account` here?

skyrpex: Either a team ID or user ID

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+3. Optionally, invite other Wing Cloud users to the team (by GitHub user ID).
+
+#### Create a new team Project (#6)
+
+1. Click on "New Project" under the team of your choice.
+2. Choose between the GitHub repositories that you have access to.
+3. Define your secret values.
+4. Make sure there's a `main.w` file in the root of the repository.
+5. Click on "Deploy".
+6. You'll be redirected to your new project's page (https://wing.cloud/-team/-project).
```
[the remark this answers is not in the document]
skyrpex: Yes, they will be two separate projects that happen to use the same repository as source. Both projects will be governed by different people (one by you, the other one by a team), and will have different secrets.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
skyrpex: I actually prefer that, too. It's technically simpler and will avoid us doing some black magic with CloudFront.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+2. You can see and manage all your projects and teams.
+3. You will see all preview environments associated with your personal account and teams you are member in.
+4. In order to see a preview environment, just click on the preview environment URL from the list of all your preview environments.
+5. For each preview environment, you can see the deployment logs and the test results and logs.
+
+#### Create a new personal Project (#4)
+
+1. To create a new project, click on "New Project" under "Personal Account".
+2. Choose between the GitHub repositories that you have access to.
+3. Define your secret values.
```
eladb, line 52: environment variables

eladb, line 52: environment variables

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+4. Make sure there's a `main.w` file in the root of the repository.
```
eladb, line 53:
```suggestion
4. Make sure there's a `main.w` or `*.main.w` file in the root of the repository.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+5. Click on "Deploy".
+6. You'll be redirected to your new project's page (https://wing.cloud/dashboard/-user/-project).
```
eladb, line 55:
```suggestion
6. You'll be redirected to your new project's page (https://wing.cloud/-user/-project).
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+1. Once signed in, you will be able to navigate directly to https://wing.cloud/dashboard.
+2. You can see and manage all your projects and teams.
+3. You will see all preview environments associated with your personal account and teams you are member in.
+4. In order to see a preview environment, just click on the preview environment URL from the list of all your preview environments.
+5. For each preview environment, you can see the deployment logs and the test results and logs.
+
+#### Create a new personal Project (#4)
+
+1. To create a new project, click on "New Project" under "Personal Account".
+2. Choose between the GitHub repositories that you have access to.
```
eladb, line 51: project:repo == 1:1 ?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+3. Define your secret values.
+4. Make sure there's a `main.w` file in the root of the repository.
+5. Click on "Deploy".
+6. You'll be redirected to your new project's page (https://wing.cloud/dashboard/-user/-project).
+
+#### Create a new Team (#5)
```
eladb, line 57: are teams p1 for the mvp? can we allow users to add other users to their personal projects without a team?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+
+1. Click on "New Team".
+2. Write a team name.
+3. Optionally, invite other Wing Cloud users to the team (by GitHub user ID).
+
+#### Create a new team Project (#6)
```
eladb, line 63: seems duplicated with the "personal team"  no need to duplicate

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+
+1. Click on "New Project" under the team of your choice.
+2. Choose between the GitHub repositories that you have access to.
+3. Define your secret values.
+4. Make sure there's a `main.w` file in the root of the repository.
+5. Click on "Deploy".
+6. You'll be redirected to your new project's page (https://wing.cloud/dashboard/-team/-project).
+
+#### Preview Environment for production branch (#7)
```
eladb, line 72: i would frame this as "preview environment for branches", and then allow users to basically create an environment for any branch in the repo (`main` is the just a case of that)

maybe p2 for mvp

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+
+For the repository production branch, there is a single preview environment that will constantly be up-to-date with the latest code committed.
+A link to the preview environment is available in the repository main page ("About" section).
+Production branch environment url structure is: `https://wing.cloud/-account/dashboard/-project/-branch`
+Once the installation of the Wing Cloud application is completed, an initial preview environment will be created and will be available in the repository main page.
+
+#### Preview Environments for Pull Requests (#8)
```
eladb, line 79: this is p1.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+
+Upon each creation of a pull request an automatic comment will be added to the PR and will guide the developer to a dedicated preview environment.
+Provide a streamlined process for updating a pull request preview environment, each code changes in a PR will redeploy the preview environment and will also:
+
+1. Provide a real-time build and deployment status updates
```
eladb, line 84: what about tests?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+2. While preview environment is being updated, the user that is currently using the environment will be notified and will be able to refresh the page to get the latest changes.
+3. Ensure a consistent and unique url for each preview environment. The url structure: `https://wing.cloud/dashboard/-account/-project/-branch` (account is a <user> or a <team>)
```
eladb, line 86:
```suggestion
3. Ensure a consistent and unique url for each preview environment. The url structure: `https://wing.cloud/-account/-project/-branch` (account is a <user> or a <team>)
```

If we really need another URL component, it should be `/environments` or `/projects`

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+| building.main.w | 🔄 Building ([logs](https://wing.cloud/dashboard/-account/-project/-branch/logs/build/))   |                                                               |                                                                                                                                                        | Jul 31, 2023 8:01am |
+| stale.main.w    | ⏸️ Stale ([lean more](https://wing.cloud/dashboard/-account/-project/-branch/logs/build/)) |                                                               |                                                                                                                                                        | Jul 31, 2023 8:01am |
+
+#### Preview Environment Environment Variables (Secrets) (#9)
+
+The project management dashboard will include a section to manage the secrets of the project. These secrets will be available during the Wing compilation.
```
eladb, line 100: is this at the project level?
how do we handle the case of different environment variables per environment (e.g. prod/staging).
maybe p2

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+
+#### Run Tests Automatically Upon PR Creation and Code Changes (#10)
+
+Upon PR creation and code changes, Wing Cloud Preview Environments will automatically run all tests defined for the entry point.
+For each test run, a new simulator will be created and will be destroyed upon test completion.
```
eladb, line 105:
```suggestion
For each test run, a new simulator instance will be created and will be destroyed upon test completion.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+The PR comment will indicate the termination of each preview environment and the URLs won't be valid anymore.
+
+#### Self-Cleaning Environments Mechanism (#13)
+
+To ensure efficient resource utilization:
+
+1. Preview environments associated with stale PRs (without access or code changes for over 30 days) will automatically deactivate.
+2. The PR's preview environment comment will indicate its inactive status.
+3. Changes in code will trigger redeployment for all related preview environments in this PR.
+
```
eladb, line 127: missing: how do users access application endpoints, such as websites and api endpoints?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+### Security (#16)
+
+1. Preview environments shouldn’t be able to access files from other preview environments.
+2. Only Wing Cloud signed up developers with access to the repository can view preview environments created for this repository.
+
+## Non-Requirements
+
+1. password protected preview environments.
+2. wing configuration file support. (for stating a specific wing version, entry points, etc.)
+3. update PR comment with tests results triggered from the console.
```
eladb, line 156: i don't think this will even be needed

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+   Every `*.main.w` file in the repository will be considered as an entry point and will have a dedicated preview environment.
+   All preview environments will be available in the PR comment and will be updated upon code changes.
+
+## Technical Details
+
+### Entity Relationship Diagram
+
+```mermaid
+erDiagram
+    User ||--0+ RoleAssignment : has
```
eladb, line 172: I don't think users should own projects, just teams.

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,364 @@
...
+#### User-Controlled Environments termination (#12)
+
+Closing a PR will terminate the corresponding preview environments.
+The PR comment will indicate the termination of each preview environment and the URLs won't be valid anymore.
+
+#### Self-Cleaning Environments Mechanism (#13)
+
+To ensure efficient resource utilization:
+
+1. Preview environments associated with stale PRs (without access or code changes for over 30 days) will automatically deactivate.
```
skorfmann, line 124: Perhaps being more aggressive could make sense? Perhaps 3 or 7 days?

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+
+For the repository production branch, there is a single preview environment that will constantly be up-to-date with the latest code committed.
+A link to the preview environment is available in the repository main page ("About" section).
+Production branch environment url structure is: `https://wing.cloud/-account/dashboard/-project/-branch`
+Once the installation of the Wing Cloud application is completed, an initial preview environment will be created and will be available in the repository main page.
+
+#### Preview Environments for Pull Requests (#8)
+
+Upon each creation of a pull request an automatic comment will be added to the PR and will guide the developer to a dedicated preview environment.
+Provide a streamlined process for updating a pull request preview environment, each code changes in a PR will redeploy the preview environment and will also:
```
staycoolcall911, line 82:
```suggestion
Provide a streamlined process for updating a pull request preview environment, each commit in a PR will redeploy the preview environment and will also:
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+
+1. Provide a real-time build, tests and deployment status updates
+2. While preview environment is being updated, the user that is currently using the environment will be notified and will be able to refresh the page to get the latest changes.
+3. Ensure a consistent and unique url for each preview environment. The url structure: `https://wing.cloud/dashboard/-account/-project/-branch` (account is a <user> or a <team>)
```
staycoolcall911, line 86: `<something>` (without backticks) does not render on markdown, so I added backticks:
```suggestion
3. Ensure a consistent and unique url for each preview environment. The url structure: `https://wing.cloud/dashboard/-account/-project/-branch` (account is a `<user>` or a `<team>`)
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+| building.main.w | 🔄 Building ([logs](https://wing.cloud/dashboard/-account/-project/-branch/logs/build/))   |                                                               |                                                                                                                                                        | Jul 31, 2023 8:01am |
+| stale.main.w    | ⏸️ Stale ([lean more](https://wing.cloud/dashboard/-account/-project/-branch/logs/build/)) |                                                               |                                                                                                                                                        | Jul 31, 2023 8:01am |
+
+#### Preview Environment Environment Variables (Secrets) (#9)
+
+The project management dashboard will include a section to manage the secrets of the project. These secrets will be available during the Wing compilation.
+
+#### Run Tests Automatically Upon PR Creation and Code Changes (#10)
```
staycoolcall911, line 102:
```suggestion
#### Run Tests Automatically Upon PR Creation and Code Commits (#10)
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+
+Upon PR creation and code changes, Wing Cloud Preview Environments will automatically run all tests defined for the entry point.
```
staycoolcall911, line 104:
```suggestion
Upon PR creation and code commits, Wing Cloud Preview Environments will automatically run all tests defined for the entry point.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+#### User-Controlled Environments termination (#13)
+
+Closing a PR will terminate the corresponding preview environments.
+The PR comment will indicate the termination of each preview environment and the URLs won't be valid anymore.
+
+#### Self-Cleaning Environments Mechanism (#14)
+
+To ensure efficient resource utilization:
+
+1. Preview environments associated with stale PRs (without access or code changes for over 10 days) will automatically deactivate.
```
staycoolcall911, line 128:
```suggestion
1. Preview environments associated with stale PRs (without access or code commits for over 10 days) will automatically deactivate.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+2. wing configuration file support. (for stating a specific wing version, entry points, etc.)
+3. update PR comment with tests results triggered from the console.
+4. enable preview environments viewers to add comments in the environment.
+5. populate Environment with Initial Data
+6. Allow users to define specific environment machine type / configuration (cpu, memory, etc.)
+7. Support multiple entry points in a single PR:
+
+   There is a support for multiple entry points in a single PR.
+   Every `*.main.w` file in the repository will be considered as an entry point and will have a dedicated preview environment.
+   All preview environments will be available in the PR comment and will be updated upon code changes.
```
staycoolcall911, line 168:
```suggestion
   All preview environments will be available in the PR comment and will be updated upon code commit.
```

docs/docs/09-rfcs/2023-03-08-wing-preview-environment.md
```diff
@@ -0,0 +1,368 @@
...
+#### Access application endpoints from the Wing Console (#12)
+
+Users will be able to access the applications' public endpoints such as APIs and web pages.
+
+#### User-Controlled Environments termination (#13)
+
+Closing a PR will terminate the corresponding preview environments.
+The PR comment will indicate the termination of each preview environment and the URLs won't be valid anymore.
+
+#### Self-Cleaning Environments Mechanism (#14)
```
staycoolcall911, line 124:
```suggestion
#### Environments Self-Cleaning Mechanism (#14)
```

PR #3712, opened 2023-08-04
chore(compiler): do not store symbol environments on AST

Chriscbr:
## Checklist

- [ ] Tests added (always)


libs/wingc/src/ast.rs
```diff
@@ -652,28 +650,27 @@ pub enum InterpolatedStringPart {
 	Expr(Expr),
 }
 
-#[derive(Derivative, Default)]
-#[derivative(Debug)]
+// do not derive Default, as we want to explicitly generate IDs
+#[derive(Debug)]
 pub struct Scope {
+	/// An identifier that is unique among all scopes in the AST.
+	pub id: usize,
```
MarkMcCulloh, line 657: wdyt about using a marker type for this? (Is marker type a thing? Does this have a better name?)

e.g.

```rust
/// An identifier that is unique among all scopes in the AST.
pub type ScopeId = usize
```

```suggestion
	pub id: ScopeId,
```

I feel like it helps make the documentation for it more accessible, plus if we ever want to get away from usize for it (we should) then it'll be an easy change.

MarkMcCulloh (approved):
Makes sense and the change looks good, I'm assuming this will work well with https://github.com/winglang/wing/pull/3677 once it's merged as well

libs/wingc/src/ast.rs
Chriscbr: good idea!

Issue #3717, opened 2023-08-06
Cannot extend non-Wing constructs


eladb:
### I tried this:

```js
bring "cdk8s" as cdk8s;

class MyChart extends cdk8s.Chart {
  init() {
    new cdk8s.ApiObject(kind: "Pod", apiVersion: "v1");
  }
}

let c = new MyChart();
```

### This happened:

```
ERROR: this._addInflightOps is not a function
```

### I expected this:

To work

### Component

Compiler, SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.26.0](https://github.com/winglang/wing/releases/tag/v0.26.0).

staycoolcall911:
I believe this now works.
CC: @eladb, @Chriscbr

Chriscbr:
Thanks for the heads up @staycoolcall911. It looks like the immediate bug above was fixed (I think via https://github.com/winglang/wing/pull/4640), but if I extend the example a little further with some inflight code, we still get errors:

```js
bring "cdk8s" as cdk8s;
bring cloud;

let global_b = new cloud.Bucket();

class MyChart extends cdk8s.Chart {
  local_b: cloud.Bucket;
  init() {
    new cdk8s.ApiObject(kind: "Pod", apiVersion: "v1");
    this.local_b = new cloud.Bucket();
  }

  pub inflight save() {
    global_b.put("data.txt", "Hello World!");
    this.local_b.put("data.txt", "Hello World!");
  }
}

let c = new MyChart();
new cloud.Function(inflight () => {
  c.save();
});
```

error:

```
ERROR: unable to serialize immutable data object of type MyChart

target/main.wsim.988477.tmp/.wing/preflight.js:79
         _registerOnLift(host, ops) {
           if (ops.includes("handle")) {
>>           $Closure1._registerOnLiftObject(c, host, ["save"]);
           }
           super._registerOnLift(host, ops);
```

I think the root cause is our lifting system is still assuming `std.Resource` is in the class's hierarchy. Refactoring the SDK to remove `std.Resource` feels like the cleanest solution IMO

Chriscbr:
From my testing it looks like the previous example now works. I was able to uncover a separate issue but it's more minor: https://github.com/winglang/wing/issues/6629

Closing.

PR #3718, opened 2023-08-06
fix: cannot extend non-wing construct

eladb:
When trying to bring `cdk8s` and extend `Chart`, the call to `this._addInflightOps()` fails because non-wing constructs don't have this method.

The fix is to use `?.()` (e.g. `this._addInflightOps?.()`) in order to chain the undefined.

Added a test which verifies that cdk8s can be brought and used.

Fixes https://github.com/winglang/wing/issues/3717

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/valid/bring_cdk8s.w
+++ b/examples/tests/valid/bring_cdk8s.w
@@ -1,0 +1,13 @@
+bring "cdk8s" as cdk8s;
+
+class MyChart extends cdk8s.Chart {
+  init() {
+    new cdk8s.ApiObject(kind: "Pod", apiVersion: "v1");
+  }
+}
+
+let c = new MyChart();
+let first: Json = c.toJson().at(0);
+assert(first.get("apiVersion") == "v1");
+assert(first.get("kind") == "Pod");
+assert(first.get("metadata").get("name").asStr().length > 0);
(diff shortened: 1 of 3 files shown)


examples/tests/valid/bring_cdk8s.w
yoav-steinberg, line 1: never thought of extending a non `cloud.Resource`. I guess the ideal test would be to import `constructs` and try to extent `contsructs.Construct`.
In any case what happens if there are inflights in `MyChart`? Wouldn't we need `_addInflightOps` then?
Maybe a quick fix is to error if there are inflights in a class that doesn't inherit `cloud.Resource`?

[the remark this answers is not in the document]
eladb: yes, good point.

eladb: Yes, good idea with `Construct`. Definitely has the exact same issue.
Also, you are right about inflights.

eladb:
OK, moving to Draft and will iterate.

eladb:
I think the likely better approach would be to get rid of `Resource` altogether and support any `Construct`.

eladb:
@Chriscbr is going to pick this up in short order. I am closing for now.

PR #3721, opened 2023-08-07
fix(build): playwright cannot find browser executables in a new env

flyingImer:
For a new development environment, [`pnpm build`](https://www.winglang.io/contributing/start-here/development#full-build) fails due to missing browser deps for playwright tests with error messages like:
```
Error: browserType.launch: Executable doesn't exist at /<...>/.cache/ms-playwright/<...>
```
It is required to install manually with `pnpm exec playwright install` within `apps/wing-console/console/app`, which breaks the `pnpm build` flow.

This fix is a try to ensure the browser deps are always installed and being the latest before running tests. It works fine for local development. Although the installation "runs" before every test run, turbo does a pretty good job to cut corners:
```
@wingconsole/app:pre-test: cache hit (outputs already on disk), replaying logs 00d0d93700261d4e
```

Just not sure about if this is okay for the CI environment, please suggest

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing-console/console/app/package.json
+++ b/apps/wing-console/console/app/package.json
@@ -14,6 +14,7 @@
     "dev": "tsx watch scripts/dev.mjs",
     "compile": "tsup",
     "eslint": "eslint --ext .js,.cjs,.ts,.cts,.mts,.tsx --no-error-on-unmatched-pattern . --fix",
+    "pre-test": "playwright install",
     "test": "playwright test --update-snapshots",
     "package": "bump-pack -b"
   },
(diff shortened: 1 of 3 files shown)


skyrpex:
Hey @flyingImer thanks for the PR. This is something I noticed from other people as well, and wanted to fix the issue.

@polamoros and I checked out the solution you worked on and has a problem: in Ubuntu (and maybe other OS) the `playwright install` will ask for sudo permissions. Additionally, `playwright install --with-deps` may be necessary, depending on your system.

For now, I think the best solution is to rename the Console/App test to `test:playwright` and run it only in the root script `test:ci`. This way, it won't bother people running `test`, but still we be run in the CI.

Issue #3527, opened 2023-07-18
Deploying / testing multiple apps in the same AWS account can lead to conflicting resource names


skorfmann:
### I tried this:

Deploying multiple apps into an AWS account, using resources like Counter and API handler.

See theses [test runs](https://github.com/winglang/examples/actions/runs/5592325224/jobs/10224642055?pr=11) in the [examples](https://github.com/winglang/examples/) repo.

### This happened:

Resource name conflicts across apps.

```
╷
│ Error: creating Amazon DynamoDB Table (wing-counter-cloud.Counter-c8d87bc1): ResourceInUseException: Table already exists: wing-counter-cloud.Counter-c8d87bc1
│ 
│   with aws_dynamodb_table.env0_cloudCounter_D1CE665B,
│   on main.tf.json line 152, in resource.aws_dynamodb_table.env0_cloudCounter_D1CE665B:
│  152:       }
│ 
╵
╷
│ Error: creating Lambda Function (cloud-Api-OnRequest-ee03792e-c8663789): operation error Lambda: CreateFunction, https response error StatusCode: 409, RequestID: 5d6fe52c-2964-4b09-aead-79114e99d439, ResourceConflictException: Function already exist: cloud-Api-OnRequest-ee03792e-c8663789
│ 
│   with aws_lambda_function.env0_cloudApi_cloudApi-OnRequest-ee03792e_458F7938,
│   on main.tf.json line 304, in resource.aws_lambda_function.env0_cloudApi_cloudApi-OnRequest-ee03792e_458F7938:
│  304:       },
│ 
```

### I expected this:

- Globally unique resource names per application by default
- or at least the option to let me define a seed to influence the generated resource names

### Is there a workaround?

manually define resource names via `as`

### Component

Language Design, Compiler, SDK

### Wing Version

0.24.31

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

Related to https://github.com/winglang/wing/issues/3224 but not sure if that is really gonna be the solution. Likely better ways to solve this.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skorfmann:
Note: When running tests agains AWS, my local test run shouldn't be interfering with other CI test runs of the same application as well.

skorfmann:
Related to https://github.com/winglang/wing/issues/2689 (point 5)

monadabot:
[...]

Issue #3224, opened 2023-07-03
Configurable App / Stack Name


skorfmann:
### Feature Spec

It's now possible to configure individual application names for Wing apps. Since the generated resource names are directly depending on the application name as a seed for the randomized name, this enables the deployment of the same Wing application to the same target account (e.g. AWS) multiple times, without having to change individual name to avoid naming conflicts. 

### Use Cases

- tests
- sdk resources

### Implementation Notes

https://github.com/winglang/wing/blob/9f24a2165bf4cc4555bd6576fb0f190c916d0647/libs/wingsdk/src/shared-tf/app.ts#L12

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skorfmann:
Deploying multiple Wing apps in the same account becomes difficult, and essentially requires manual, explicit and globally unique naming for otherwise totally unrelated resources across stacks. Here's an [example](https://github.c`winglang/examples/actions/runs/5592325224/jobs/10224642055?pr=11) - while there are other permission related issues in this run, the Counter resource and an API handler are affected.

```
╷
│ Error: creating Amazon DynamoDB Table (wing-counter-cloud.Counter-c8d87bc1): ResourceInUseException: Table already exists: wing-counter-cloud.Counter-c8d87bc1
│ 
│   with aws_dynamodb_table.env0_cloudCounter_D1CE665B,
│   on main.tf.json line 152, in resource.aws_dynamodb_table.env0_cloudCounter_D1CE665B:
│  152:       }
│ 
╵
╷
│ Error: creating Lambda Function (cloud-Api-OnRequest-ee0[37](https://github.com/winglang/examples/actions/runs/5592325224/jobs/10224642055?pr=11#step:8:38)92e-c8663789): operation error Lambda: CreateFunction, https response error StatusCode: 409, RequestID: 5d6fe52c-2964-4b09-aead-79114e99d439, ResourceConflictException: Function already exist: cloud-Api-OnRequest-ee03792e-c8663789
│ 
│   with aws_lambda_function.env0_cloudApi_cloudApi-OnRequest-ee03792e_458F79[38](https://github.com/winglang/examples/actions/runs/5592325224/jobs/10224642055?pr=11#step:8:39),
│   on main.tf.json line 304, in resource.aws_lambda_function.env0_cloudApi_cloudApi-OnRequest-ee03792e_458F7938:
[...]

Issue #2689, opened 2023-05-29
sdk test spec github action improvements


tsuf239:
### Feature Spec

- [x] Report failures and notify the on-call member
- [x] Run right before releasing a version/ instead of every day
- [ ] ~Add a flag to run it when placing a PR (similar to pr/e2e-full label)~
- [x] adding a flag to disable cleaning the resources + supply the temp dir path to the user.  #2884
- [ ] Randomize resources prefix (root id), by default (and a flag to keep it steady).  #3723 
This way the names of the resources will be randomized and we won't be dependent on cleaning them up before running a test. 
- [ ] Run tests in parallel on a few machines  #3723 
- [ ] Update SDK workflow to run on a few machines #3726 
- [ ] ~Condition the run on the latest changes location (do not run when console, or docs changes)~

moved point 3 and 8 to https://github.com/winglang/wing/issues/3727.

### Use Cases

testing

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.24.90](https://github.com/winglang/wing/releases/tag/v0.24.90).

PR #3723, opened 2023-08-07
feat(cli): parallel sdk testing

tsuf239:
Fixes the 5th and 6th points in #2689
Fixes: #3527
Fixes: #3224
## Description:
* Stop copying wing files when testing on cloud targets, instead changed the target dir to a random tmp dir.
* Separating env vars to be present in compilation context only (those are present only on the compiled artifact, not in its imports, so vars like WING_TARGET and WING_IS_TEST are still out). 
* Moving WING_SOURCE_DIR env var to the app constructor- this way it can be passed in the code easily and doesn't depend on global env.
* Upon testing, adding an extra nanoid to the tf and cdk root resource name, to prevent having resources with the same id over the different tests
* Fixing `Bucket.tryGet`, `Bucket.tryGetJson`, `Bucket.tryDelete` since were failing on aws targets
* adjusting the tests
* generating snapshots

**All code changes can be found on the 1st commit (16 FILES), 2nd one is for snapshots (290~ FILES)**

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/shared-tf/app.ts
```diff
@@ -40,7 +41,10 @@ export abstract class CdktfApp extends App {
     const cdktfApp = new cdktf.App({ outdir: cdktfOutdir });
     const cdktfStack = new cdktf.TerraformStack(cdktfApp, TERRAFORM_STACK_NAME);
 
-    super(cdktfStack, "Default");
+    const rootId = props.isTestEnvironment
+      ? "Default_" + nanoid(10)
```
Chriscbr, line 45: Random IDs for constructs should be strongly avoided. We want compilations to be deterministic and reproducible, so when you instantiate a class or call a method during preflight, the behavior should be the same each time as long as your environment is the same (e.g. same environment variables are set and so on).

How about we instead allow the root ID to be customized through AppProps or an environment variable? This way choosing the ID will be the responsibility of the CLI or user instead.

libs/wingsdk/src/core/app.ts
```diff
@@ -47,6 +47,11 @@ export interface AppProps {
    * @default false
    */
   readonly isTestEnvironment?: boolean;
+
+  /**
+   * App's external context
```
Chriscbr, line 52: What does this mean? (add more description)

Chriscbr (changes requested):
Great start - reducing resource naming conflicts for tests will be very helpful. Changing the name/ID of the root construct seems like the correct approach to me, I've just requested one change to where we keep the logic for customizing app IDs. Otherwise looks good to me.

Chriscbr:
(Thinking out loud) In the future we might also want to write up some guidelines for implementing resources in the SDK. For example, we by using the `ResourceNames` class/utility in the SDK since it will guarantee that whenever the construct path  (like `root/Store/cloud.Bucket`) changes, the generated physical name of the resource should change (preventing name conflicts).

libs/wingcompiler/src/compile.ts
```diff
@@ -42,6 +42,9 @@ export interface CompileOptions {
 
   /// Enable/disable color output for the compiler (subject to terminal detection)
   readonly color?: boolean;
+
+  // target directory for the output files
```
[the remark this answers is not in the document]
eladb: This is TypeScript...

libs/wingsdk/src/core/app.ts
```diff
@@ -47,6 +47,11 @@ export interface AppProps {
...
+   */
+  readonly sourceDir: string;
```
eladb, line 54: Seems like should actually be `entrypointDir`, right? It seems like this is used by the `Website` resource to find the website assets relative to the entrypoint.

Which begs the question, now that we have support for multiple files, how would that work?

@Chriscbr any thoughts?

It feels like this should somehow be implicitly passed when `new cloud.Website` is called. Maybe we can add some magic keyword argument called `__dirname`. If the compiler sees it, it will assign the file's directory to it implicitly.

Something like:

```js
new cloud.Website();

// implies
new cloud.Website(__dirname: __dirname);
```

libs/wingsdk/src/shared-tf/app.ts
eladb: If I understand correctly, this is only for test environments which are expected to be teared down anyway and be isolated from the environment. Kind of feels like a reasonable case for a random id, no?

eladb, line 45:
```suggestion
      ? `Test.${nanoid(10)}`
```

libs/wingsdk/src/target-awscdk/app.ts
```diff
@@ -74,7 +75,10 @@ export class App extends CoreApp {
     const cdkApp = new cdk.App({ outdir: cdkOutdir });
     const cdkStack = new cdk.Stack(cdkApp, stackName);
 
-    super(cdkStack, "Default");
+    const rootId = props.isTestEnvironment
+      ? "Default_" + nanoid(10)
```
eladb, line 79: DRY (find a way to reuse this across apps)

libs/wingsdk/src/shared-tf/app.ts
tsuf239: True @eladb :) 
I can also pass it from the user, the default will be just "Default", and in testing it can be automated to
 "<test folder>_<test file>". it may be better but still not collision-proof.... what do you think? @eladb @Chriscbr

libs/wingcompiler/src/compile.ts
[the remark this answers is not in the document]
Chriscbr: Oops 😅

libs/wingsdk/src/shared-tf/app.ts
Chriscbr: @eladb What about snapshot tests?

eladb: > @eladb What about snapshot tests?

Good point. Here's a suggestion: let's add `--rootid` switch which will *default to nanoid* but snapshot tests can pass some fixed value there. WDYT?

Eventually we need snapshot testing to be built into `wing test`, so this should be handled internally.

libs/wingsdk/src/core/app.ts
tsuf239: It just means that I experimented with a different property before and forgot to update the docstring, thanks for noticing!

libs/wingsdk/src/core/app.ts
tsuf239: Opened an issue: https://github.com/winglang/wing/issues/3736

libs/wingcompiler/src/compile.ts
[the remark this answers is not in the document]
tsuf239: still good to know! I'll remember that 😊

libs/wingsdk/src/shared-tf/app.ts
tsuf239: Added the switch, the snapshot tests are currently on the sim only, there I haven't changed the root name, since it's hardcoded in a few places in our code... (and we're not worried about tf not working well, the main reason to randomize the names- since we are not using tf for the sim)

apps/wing/src/commands/test.ts
```diff
@@ -131,14 +133,15 @@ function printResults(
 async function testOne(entrypoint: string, options: TestOptions) {
   // since the test cleans up after each run, it's essential to create a temporary directory-
   // at least one that is different then the usual compilation dir,  otherwise we might end up cleaning up the user's actual resources.
-  const tempFile: string =
-    options.target === Target.SIM ? entrypoint : await generateTmpDir(entrypoint);
+  const tempFile: string = entrypoint;
```
Chriscbr, line 136: Looks like this extra variable can be removed

libs/wingsdk/src/target-sim/app.ts
```diff
@@ -59,7 +59,8 @@ export class App extends core.App {
   private synthed = false;
 
   constructor(props: core.AppProps) {
-    super(undefined as any, "root");
+    // doesn't allow customize the root id- as used hardcoded in the code
+    super(undefined as any, "root", props);
```
Chriscbr, line 63: In theory I'm not sure why we shouldn't also be able to customize the root id for the `sim` target

Chriscbr (approved):

libs/wingsdk/src/shared-tf/app.ts
Chriscbr: > since it's hardcoded in a few places in our code

Maybe open an issue to address it later?

tsuf239: opened: https://github.com/winglang/wing/issues/3749 :)

libs/wingsdk/src/target-sim/app.ts
tsuf239: I agree, we use this string mostly `"root/cloud.TestRunner"`, and in the console too, so we really need to be careful with that...

PR #3726, opened 2023-08-07
feat(sdk): parallel sdk spec testing via gh action

### LINKED ISSUE #2689: sdk test spec github action improvements

tsuf239:
## Description

Thanks to @skorfmann's amazing [code](https://github.com/winglang/examples/blob/main/.github/workflows/wing-sdk.yml), I managed to split the SDK spec tests to run in parallel on a few machines based on the tests directory!

Cannot be deployed before #3723 

finally fixes #2689! 🥳 

It looks like that:
image

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/workflows/tf-aws-test.yml
+++ b/.github/workflows/tf-aws-test.yml
@@ -64,7 +64,7 @@ jobs:
           node-version: 18
       - name: Install winglang globally
         run: npm install -g winglang
-      - name: Check for package.json file
+      - name: Installing external js modules
         run: | 
           cd examples/tests/sdk_tests
           npm install
@@ -75,13 +75,13 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ env.AWS_REGION }}
       - name: Execute wing test in matrix directory
-        # env:
-        #   TF_LOG: info
-        #   TF_LOG_PATH: ${{ runner.workspace }}/terraform.log
-        run: wing test -t sim ${{ matrix.test.directory }}/*.w
-      # - name: Output Terraform log
-        # if: failure()
-        # run: cat ${{ runner.workspace }}/terraform.log
+        env:
+          TF_LOG: info
+          TF_LOG_PATH: ${{ runner.workspace }}/terraform.log
+        run: wing test -t tf-aws ${{ matrix.test.directory }}/*.w
+      - name: Output Terraform log
+        if: failure()
+        run: cat ${{ runner.workspace }}/terraform.log
     
     
     # steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tf-aws-test.yml
```diff
@@ -5,39 +5,63 @@ on:
...
+    - name: Checkout Repository
+      uses: actions/checkout@v3
+    - name: Get list of directories
+      id: setdirs
+      shell: bash
+      run: |
+        dirs=$(ls -d examples/tests/sdk_tests/*/ | sed 's/\/$//' | grep -v "external" | jq -R -s -c 'split("\n")[:-1]')
+        processed_dirs=$(echo "{ \"directory\": $dirs }" | jq -c '[ .directory[] | {directory: ., name: (split("/") | last)}]')
+        wrapped_dirs=$(echo "{ \"test\": $processed_dirs }" | jq -c .)
+        echo "DIRS=$wrapped_dirs" >> $GITHUB_ENV
```
Chriscbr, line 24: Perhaps add a small script in the `scripts/` folder to handle this logic?

tsuf239: I'm not sure this script, as is, is reusable- as it reads the directories from `examples/tests/sdk_tests/*/` and saves the output in some sort of object as an env nar, I don't think we use it again or need to save an env var when running locally...
If we want to enable this workflow to run locally, we'll have to rewrite the whole thing (considering we're not running on 14 different machines) or just use [act](https://github.com/nektos/act),

tsuf239:
checked again now, it takes around 10~ minutes to run the tests:
image
there is some sort of a problem with the bucket props: 
image
and function rate concurrency (requested increase from AWS it might take a bit of time)

Chriscbr (approved):

.github/workflows/tf-aws-test.yml
tsuf239: I agree! On the other hand, chatGPT did a pretty good job at explaining that, the next person to maintain it will be just fine

PR #3729, opened 2023-08-07
feat(compiler): structural typing for known json objects

### LINKED ISSUE #2366: Infer types of struct literals
### Feature Spec

Starting in Wing x.y.z, it's possible to create complex, nested structs without needing to specify the exact struct type you're creating.

Previously, when you were instantiating complex types, you might have to write a lot of types by hand:

```js
bring cloud;
bring "@cdktf/provider-kubernetes" as kube;

new kube.provider.KubernetesProvider(
  config_path: "~/.kube/config",
);

let ns = new kube.namespace.Namespace(metadata: kube.namespace.NamespaceMetadata {
  name: "nginx",
});

let deployment = new kube.deployment.Deployment(metadata: kube.deployment.DeploymentMetadata {
  name: "nginx",
  namespace: ns.metadata.name,
}, spec: kube.deployment.DeploymentSpec {
  template: kube.deployment.DeploymentSpecTemplate {
    metadata: kube.deployment.DeploymentSpecTemplateMetadata {
      labels: {
        app: "nginx",
      },
    },
    spec: kube.deployment.DeploymentSpecTemplateSpec {
      container: [
        kube.deployment.DeploymentSpecTemplateSpecContainer {
          name: "nginx-container",
          image: "nginx",
          port: kube.deployment.DeploymentSpecTemplateSpecContainerPort {
            container_port: 80,
          },
        }
      ]
    }
  }
});
```

Now, it's possible to infer the types in locations where only a single struct type is expected, so you can instead write:

```js
bring cloud;
bring "@cdktf/provider-kubernetes" as kube;

new kube.provider.KubernetesProvider(
  config_path: "~/.kube/config",
);

let ns = new kube.namespace.Namespace(metadata: {
  name: "nginx",
});

let deployment = new kube.deployment.Deployment(metadata: {
  name: "nginx",
  namespace: ns.metadata.name,
}, spec: {
  template: {
    metadata: {
      labels: {
        app: "nginx",
      },
    },
    spec: {
      container: [
        {
          name: "nginx-container",
          image: "nginx",
          port: {
            container_port: 80,
          },
        }
      ]
    }
  }
});

```

### Use Cases

Reduce boilerplate in code

### Implementation Notes

_No response_

### Component

Language Design

### Community Notes

### LINKED ISSUE #2597: Expected type to be "str", but got "Array<str>" instead
### I tried this:

```
bring "cdktf" as cdktf;

let provider = new cdktf.TerraformHclModule(
    source:  "github.com/philips-labs/terraform-aws-github-oidc?ref=0.7.0//modules/provider"
);

let github_oidc = new cdktf.TerraformHclModule(
    source:  "github.com/philips-labs/terraform-aws-github-oidc?ref=0.7.0",
    variables: {
        openid_connect_provider_arn: provider.get("arn"),
        repo: "winglang/examples",
        role_name: "repo-examples",
        conditions: [{
            test: "StringLike",
            variable: "token.actions.githubusercontent.com:sub",
            values: ["repo:my-org/my-repo:pull_request"]
        }]
    }
) as "repo-examples";
```

### This happened:

Screenshot 2023-05-20 at 17 42 12

`Expected type to be "str", but got "Array<str>" instead`

### I expected this:

Since the [variables](https://github.com/hashicorp/terraform-cdk/blob/f13e58902d1804358a13d4bd75909c508a39f43b/packages/cdktf/lib/terraform-hcl-module.ts#L8) are typed as `any`, that I'm able to provide a list of strings as specified in the [terraform module docs](https://registry.terraform.io/modules/philips-labs/github-oidc/aws/latest#usages)

### Is there a workaround?

Resort to a string token

```
values: cdktf.Token.asString(["repo:my-org/my-repo:pull_request"])
```

### Component

Compiler

### Wing Version

0.17.16

### Wing Console Version

0.17.16

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

### LINKED ISSUE #3697: Inference of struct type name
### Feature Spec

When specifying a struct literal, it's possible to omit the struct type name when the literal is assigned to a parameter or field with a known struct type name.

For example:

```js
struct MyStruct {
  foo: str;
  bar: num;
}

// before:
let x: MyStruct = MyStruct { foo: "hello", bar: 123 };

// after:
let x: MyStruct = { foo: "hello", bar: 123 };
```

### Use Cases

Less typing.

### Implementation Notes

Intuitively I hope that we can special-case this in the compiler *not* to need structural typing between `json` and structs.

### Component

Compiler

### Community Notes

### LINKED ISSUE #3848: Type inference doesn't work with `if let` expressions
### I tried this:

Wing uses `log` to infer that variables may be strings:

```
let f = (x) => {
  log(x);
};
```

But this information doesn't seem to carry through `if let` expressions:

```
let f = (x) => {
  if let y = x {
    log(y);
  }
};
```

### This happened:

```
error: Expected type to be optional, but got "unknown" instead
  --> inference.w:2:14
  |
2 |   if let y = x {
  |              ^ Expected type to be optional, but got "unknown" instead
```

### I expected this:

It infers `x` is `str?`

### Is there a workaround?

Add type annotations

### Component

Compiler

### Wing Version

0.25.14

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
Fixes #3697 (it seemed easier to do it this way than to special case adding the struct name)
Fixes #2366
Fixes #2597
Fixes #3848

Mostly putting this up in case anyone wants to see the spec change. The code needs cleanup and a few more test cases need to be added. (it works though!)

The main complications here are due to my recent inference changes, otherwise the implementation is pretty straightforward.

Completions inside Json literals also work like structs if they are being used as a struct.

Misc:
- Added InferenceId and ExprId to wrap over usize
- Made sure hover worked for `if let`
- Json and Map keys are now Symbols so that they have span info. This also allows hover to work on them
- Fix bug where `if let` did not do inference properly


eladb (commented):
Awesome stuff!

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -981,24 +981,17 @@ error: Expected type to be \\"inflight (key: str, type: BucketEventType): void\\
    |                               ^^^^ Expected type to be \\"inflight (key: str, type: BucketEventType): void\\", but got \\"inflight (request: ApiRequest): ApiResponse\\" instead
 
 
-error: Expected \\"Json\\" elements to be Json values (https://www.json.org/json-en.html), but got \\"Array<unknown>\\" which is not a Json value
-   --> ../../../examples/tests/invalid/inference.w:71:14
-   |
-71 | Json { cool: anotherEmptyArray };
-   |              ^^^^^^^^^^^^^^^^^ Expected \\"Json\\" elements to be Json values (https://www.json.org/json-en.html), but got \\"Array<unknown>\\" which is not a Json value
-
```
hasanaburayyan, line 989: hmm, why does this no longer fail?

```ts
let anotherEmptyArray = [];
Json { cool: anotherEmptyArray };
```

MarkMcCulloh: Array<unknown> either means:

1. It's an empty array. That's always valid Json.
2. It's an array with an inference that we figure out during type checking. With this PR we now validate Json legal values after type checking is complete, so we should be safe to not preemptively show this error
3. It's an array with an inference that we never figure out. No reason to show an error here because it would be a cascading error.

libs/wingc/src/type_check/inference_visitor.rs
```diff
@@ -101,6 +122,31 @@ impl<'a> crate::visit_types::VisitType<'_> for InferenceVisitor<'a> {
 		crate::visit_types::visit_typeref(self, node);
 	}
 
+	fn visit_json(&mut self, node: &'_ JsonData) {
+		match &node.kind {
+			JsonDataKind::Type(_) => {
+				visit_json(self, node);
+			}
+			JsonDataKind::Fields(fields) => {
```
yoav-steinberg, line 130: Wouldn't it be ideal to just verify the json schema of `expected_struct` against `node`?
I'm not sure @hasanaburayyan's json schema is in yet, but ideally I think we should use it as a basis for all struct/json matching.

yoav-steinberg: Can we generate the schema at compile time? Then Hasan can just use that schema when generating the runtime code.

MarkMcCulloh: The purpose of the type checking here is to basically cast Json data to structs. If there's already a struct then that schema will be generated by Hasan's change.

If you're interested in generating schemas for Json data directly with statically known types, that's a cool idea but out of scope for this PR IMO.

libs/wingc/src/type_check.rs
```diff
@@ -794,7 +861,13 @@ impl Display for Type {
...
 			Type::Boolean => write!(f, "bool"),
 			Type::Void => write!(f, "void"),
-			Type::Json => write!(f, "Json"),
+			Type::Json(t) => {
+				if let Some(t) = t {
+					write!(f, "Json {}", t)
+				} else {
+					write!(f, "Json")
+				}
+			}
```
Chriscbr, line 870: I'm not sure we should be showing the inferred Json literal information publicly as part of the types.

Screenshot 2023-08-11 at 12 59 40 PM

It looks like this info appears in the LSP. IMO (at least initially) this info should just be information we're storing under the hood, and not something leaks into the user's mental model of the Wing type system.

Here's another example where this leads to some confusing behavior:

Screenshot 2023-08-11 at 1 07 45 PM

libs/wingc/src/type_check.rs
```diff
@@ -203,13 +203,13 @@ pub enum Type {
 	Duration,
 	Boolean,
 	Void,
-	Json,
+	Json(Option<JsonData>),
 	MutJson,
```
Chriscbr, line 207: Why does this extra field exist on Json and not MutJson? (add comment)

libs/wingc/src/type_check.rs
```diff
@@ -1299,7 +1419,19 @@ impl Types {
 
 		let existing_type = self.inferences.get_mut(id).expect("Inference id out of bounds");
 
-		assert!(existing_type.is_none(), "Inference id already has a type");
+		if let Some(existing_type) = existing_type {
+			// if the types are the same, ok, otherwise error
+			if std::ptr::eq(&**existing_type, &*new_type) {
```
Chriscbr, line 1424: str::ptr::eq only checks if the values to the same internal instance of `Type`, not necessarily if the they're logically the same in Wing's type system. For example its possible to have two `Type` instances of `Type::Array(Type::Number)` with different pointers. Is that the intention here? (otherwise it might make sense to use `existing_type.is_same_type_as(new_type)`)

libs/wingc/src/type_check.rs
```diff
@@ -1299,7 +1419,19 @@ impl Types {
...
+				// this can happen when we have a type that can have multiple references to the same inference inside it
+				// e.g. function, json
+				return;
+			} else {
+				report_diagnostic(Diagnostic {
+					message: format!("Inferred type {new_type} conflicts with already inferred type {existing_type}"),
+					span: None,
+				});
```
Chriscbr, line 1432: Do we have a test case for this?

libs/wingc/src/type_check.rs
```diff
@@ -2113,18 +2278,24 @@ impl<'a> TypeChecker<'a> {
...
-          if !t.is_json_legal_value() {
+          if self.is_in_mut_json && !known_type.is_json_legal_value() {
             self.spanned_error(
               v,
               format!(
                 "Expected \"Json\" elements to be Json values (https://www.json.org/json-en.html), but got \"{}\" which is not a Json value",
-                t
+                known_type
               ),
             );
```
Chriscbr, line 2288: style: use tabs, not spaces

(can we set up an automated linter for this?)

libs/wingc/src/type_check.rs
```diff
@@ -2475,6 +2657,73 @@ impl<'a> TypeChecker<'a> {
...
+		}
+
+		// if the actual type is a Json with known data, we can attempt to structurally type check
+		if expected_types.len() == 1 {
+			let expected_type = self.types.maybe_unwrap_inference(expected_types[0]);
+			let expected_type_unwrapped = expected_type.maybe_unwrap_option();
+			if let Type::Json(Some(data)) = &**json_type.maybe_unwrap_option() {
+				if expected_type_unwrapped.is_json_legal_value()
+					|| expected_type_unwrapped.is_struct()
+					|| matches!(**expected_type_unwrapped, Type::Map(_))
```
Chriscbr, line 2694:
```suggestion
					|| expected_type_unwrapped.is_map()
```

libs/wingc/src/type_check.rs
```diff
@@ -2700,11 +2955,23 @@ impl<'a> TypeChecker<'a> {
 						"Cannot assign nil value to variables without explicit optional type",
 					);
 				}
+				if *reassignable && inferred_type.is_json() {
+					if let Type::Json(ref mut data) = *inferred_type {
+						// We do not have the required analysis to know the type of the Json data after reassignment
+						data.take();
+					}
+				}
```
Chriscbr, line 2963: Do we have a test case for this?

libs/wingc/src/type_check_assert.rs
```diff
@@ -1,12 +1,14 @@
...
-	type_check::Types,
+	diagnostic::{report_diagnostic, Diagnostic},
+	type_check::{JsonData, JsonDataKind, Type, Types},
 	visit::{self, Visit},
 };
 
 /// This visitor validates:
 /// 1. All expressions in the AST were resolved to some type.
 /// 2. If any were resolved to `Unresolved` then we should have generated some error diagnostics.
+/// 3. Any Json literal expression only has legal json values unless it is being used as a struct.
```
Chriscbr, line 11: Why is this check added here instead of during the main type checking phase? This code smells off to me because it seems like `TypeCheckAssert` is no longer having just a single responsibility.

libs/wingc/src/type_check.rs
MarkMcCulloh: > this info should just be information we're storing under the hood, and not something leaks into the user's mental model of the Wing type system.

Not sure I agree, I think the user's mental model could handle the extra information. I will go ahead and remove it for now though.


> Here's another example where this leads to some confusing behavior:

This is actually a bug. I think I was too optimistic by letting the return statement set that. Will fix.

libs/wingc/src/type_check.rs
MarkMcCulloh: It was the intention, in theory the only time this should legally happen is when they are the same instance. I can explore this with an additional test case though and see if it should change.

libs/wingc/src/type_check.rs
MarkMcCulloh: will do

libs/wingc/src/type_check.rs
MarkMcCulloh: Interesting, don't we already run rustfmt? My editor does as well, surprised this made it in, will look into it

libs/wingc/src/type_check.rs
MarkMcCulloh: added

libs/wingc/src/type_check.rs
Chriscbr: > Okay it turns out fmt stops working when the line is too long (max_width is set to 120). Unfortunately, it also doesn't provide any feedback about that

interesting! Yeah I was wondering why it seemed like `fmt` just stopped working sometimes... that explains a lot

libs/wingc/src/type_check_assert.rs
Chriscbr: Is this unique to Json? How is this different from inferring the type of an Array:

```js
let foo = () => {
  return "hi";
};
let bar = foo();

let x = [bar];
```

(BTW this example throws the error `Expected type to be "unknown", but got "unknown" instead`... smells like a bug)

libs/wingc/src/type_check.rs
Chriscbr: As a rule of thumb, I think most [1] type annotations a user gets on hover should be annotations that they can type into their code - it makes for a more consistent experience. Showing the inferred Json type would be an issue since this isn't supported in Wing:

```js
let x: Json { base: str } = { base: "d" };
```

[1] Exceptions in partially-compiled programs like "unknown" are probably OK

libs/wingc/src/type_check_assert.rs
MarkMcCulloh: Json-legalness is not really part of our types, so inference can't known whether or not a given inferred type will be json legal or not. 
If I make an array `let x = [bar]` but I don't know what `bar` is (yet), nothing unsound will be introduced so you don't need to go back and double check it. but if it's `let x = Json [bar]` then there are limitations on what `bar` is allowed to be.

Additionally, the added feature of casting Json to structs is unique and requires this check to happen later. Consider this example:

```
bring cloud;

struct HasBucket {
  b: cloud.Bucket;
}

let regularJson = {
  b: new cloud.Bucket()
//   ^^^^^^^^^^^^^^^^^^ This is an error
};

let hasBucket: HasBucket = {
  b: new cloud.Bucket()
//.  ^^^^^^^^^^^^^^^^^^ This is not an error because the literal is only used in the context of being casted to a struct
};

let hasBucket2: HasBucket = regularJson;
//                          ^^^^^^^^^^^ This is not an error because there's no need to double-report the issue
```

(worth noting I just found a bug writing out this example, but I'll fix it now and add this example)

---

> BTW this example throws the error Expected type to be "unknown", but got "unknown" instead... smells like a bug

You are right, there is an issue with inference. I'll look into a fix and add that as a test case.

examples/tests/invalid/json.w
```diff
@@ -18,23 +17,81 @@ let a: Array<str> = j;
...
+
+let var mutableJson = {
+  foo: "bar",
+  stuff: [1, 2, 3],
+  maybe: {
+    good: true,
+  }
+};
+let mutableJsonStruct: StructyJson = mutableJson;
+//                                   ^^^^^^^^^^^ Expected type to be "StructyJson", but got "Json" instead
```
Chriscbr, line 97: How come this error disappears when the `var` keyword is removed?

I sort of get how this is coming about (if Json-typed variable is reassignable, then type inference becomes much more complicated, so the compiler stops trying to track the exact structure) -- but it feels weird because we're essentially saying that casting rules depend not only on types, but also whether we are casting from an lvalue or rvalue.

It's also weird because code like this seems like it should compile:
```js
let var mutableJson: StructyJson = {
  foo: "bar",
  stuff: [1, 2, 3],
  maybe: {
    good: true,
  }
};
```

If `mutableJson` is reassignable then `StructyJson` has to be on the right hand of the equal sign for the code to compile -- but if it's _not_ reassignable, then `StructyJson` can be on either side of the equal sign. This feels confusing.

I don't have a solution off the top of my head / not sure if this is a blocker, just sharing some raw thoughts here

@eladb @yoav-steinberg This seems like an interesting edge case

libs/wingc/src/type_check_assert.rs
Chriscbr: Thanks for the explanation 👍  - very helpful.

examples/tests/invalid/json.w
MarkMcCulloh: > It's also weird because code like this seems like it should compile:

I agree, and it looks like this edge case is actually pretty easy to handle. Will add a fix.

As for the general case of `let mutableJsonStruct: StructyJson = mutableJson;`, it actually was pretty helpful when the displayed type included the known json info (the text that was removed in a recent thread). It told the user what our type system knew about that object, where just having `Json` told the user we had no info.

MarkMcCulloh: An additional hint has been added to the error message to let the user know that they should use `.fromJson` because the data of the Json is not known.

Chriscbr (approved):
Nicely done 👍

libs/wingc/src/type_check.rs
```diff
@@ -1283,11 +1347,13 @@ impl Types {
 	pub fn maybe_unwrap_inference(&self, t: TypeRef) -> TypeRef {
 		if let Type::Inferred(id) = &*t {
 			if let Some(t) = self.get_inference_by_id(*id) {
-				return self.maybe_unwrap_inference(t);
+				self.maybe_unwrap_inference(t)
+			} else {
+				t
 			}
+		} else {
+			t
```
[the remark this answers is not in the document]
MarkMcCulloh: Yeah you're right, I'll check if I should either update the functionality or the docstring

libs/wingc/src/type_check.rs
```diff
@@ -1283,11 +1347,13 @@ impl Types {
 	pub fn maybe_unwrap_inference(&self, t: TypeRef) -> TypeRef {
 		if let Type::Inferred(id) = &*t {
 			if let Some(t) = self.get_inference_by_id(*id) {
-				return self.maybe_unwrap_inference(t);
+				self.maybe_unwrap_inference(t)
+			} else {
```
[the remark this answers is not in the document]
MarkMcCulloh: My answer is maybe :)

As currently implemented, it is not possible to avoid inference chains if we include cases of invalid programs:

```js
(arg /*: inference(0)*/) /*: inference(1)*/ => {
  return arg;
  // at this point, we only know that inference(1) -> inference(0)
}
```

I can't think of a valid example at the moment, although I vaguely remember running into a one as well. I'll see what I can do in this PR though, because I would prefer that invariant.

libs/wingc/src/ast.rs
```diff
@@ -547,12 +548,12 @@ pub enum ExprKind {
 		fields: IndexMap<Symbol, Expr>,
 	},
 	JsonMapLiteral {
-		fields: IndexMap<String, Expr>,
+		fields: IndexMap<Symbol, Expr>,
 	},
```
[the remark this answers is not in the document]
MarkMcCulloh: I'm into it, I'll add it to this PR if it doesn't seem to cause any complications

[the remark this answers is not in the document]
MarkMcCulloh: This change is a little more involved than I first expected, added a followup https://github.com/winglang/wing/issues/3873

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
MarkMcCulloh: Ok so for valid cases, inferences are used but I suppose it's more of an implementation detail of the compiler rather than something that a user situationally triggers. Given that, it should be possible to avoid.

However, also given that invalid cases like the above exist and still need to be modeled, I'm not sure there's much value in trying to figure out a different implementation.

PR #3733, opened 2023-08-07
chore(compiler): remove object freezing

Chriscbr:
Today, object freezing in Wing doesn't provide much value to users. The compiler emits `Object.freeze()` in lots of the JS code for the promise of runtime safety, but to date it isn't a capability we test.

To that end, I'm proposing to remove object freezing until we have stronger use cases for this. By removing any details on freezing, we're also able to simplify Wing's type model and reduce the code we're emitting in both preflight and inflight.

## Checklist

- [ ] Tests added (always)


yoav-steinberg:
Here's my concern - currently there's a chance that a compiler but will will allow you to write code that mutates a non-mutable which might result in a runtime error when trying to mutate a frozen js object. With this change you're removing that safety check. I think in reality we haven't encountered something like that so that's a +1 for your change which does simplify things. But I'm still not totally convinced removing this safety net is a "net gain".

eladb (approved):
👍

Chriscbr:
> Here's my concern - currently there's a chance that a compiler but will will allow you to write code that mutates a non-mutable which might result in a runtime error when trying to mutate a frozen js object. With this change you're removing that safety check. I think in reality we haven't encountered something like that so that's a +1 for your change which does simplify things. But I'm still not totally convinced removing this safety net is a "net gain".

I guess the question to me is whether the `Object.freeze()` strategy is meant to be a public API guarantee. That is, should we promise users that `Array`-type values always refer to frozen JavaScript arrays? Today there are several holes with this model, for example:
- if you `bring` a jsii module and you call a method that returns an `Array`, you may get an unfrozen array
- if you call an extern JavaScript function and annotate it as returning an `Array`, it may return an unfrozen array

Closing these gaps (by introducing object freezing at function call boundaries etc.) seems more effort than it's worth IMO

PR #3739, opened 2023-08-08
feat(console): distinguish between user-named resources and compiler-named 

polamoros:
- Set non-user-defined Function resources as children of the related resource.
- Use inflight names for Function resources generated by the compiler.
- Enhance the ELK configuration.
- Fix the issue where the icon for test resources was not being displayed.
- Enable test selection between test explorer and map view.

**Current**
image

**New**
image

### DIFF
--- a/tools/hangar/__snapshots__/tree_json.ts.snap
+++ b/tools/hangar/__snapshots__/tree_json.ts.snap
@@ -31,7 +31,6 @@ exports[`tree.json for an app with many resources 1`] = `
                     },
                     "display": {
                       "hidden": true,
-                      "sourceModule": "@winglang/sdk",
                     },
                     "id": "$Closure1",
                     "path": "root/Default/Default/$Closure1",
@@ -46,7 +45,6 @@ exports[`tree.json for an app with many resources 1`] = `
                     },
                     "display": {
                       "hidden": true,
-                      "sourceModule": "@winglang/sdk",
                     },
                     "id": "$Closure5",
                     "path": "root/Default/Default/$Closure5",
@@ -150,7 +148,6 @@ exports[`tree.json for an app with many resources 1`] = `
                             },
                             "display": {
                               "description": "A distributed atomic counter",
-                              "sourceModule": "@winglang/sdk",
                               "title": "Counter",
                             },
                             "id": "cloud.Counter",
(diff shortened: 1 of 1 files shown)


libs/wingsdk/src/std/resource.ts
```diff
@@ -506,6 +518,16 @@ export class Display {
...
 
+  /**
+   * Whether the resource is user-named or compiler-named.
+   */
+  public type?: "user-named" | "compiler-named";
+
+  /**
+   * The name of the resource.
+   */
+  public name?: string;
```
Chriscbr, line 529: How is this different from Display.title?

libs/wingsdk/src/std/resource.ts
```diff
@@ -506,6 +518,16 @@ export class Display {
    */
   public description?: string;
 
+  /**
+   * Whether the resource is user-named or compiler-named.
+   */
+  public type?: "user-named" | "compiler-named";
```
Chriscbr, line 524: I feel like the generalized question of what we're trying get at is, who authored the class? Was it the user creating an app? Was it their teammate? Was it the author of the library? Was it us (Wing Cloud Library maintainers)?

The most robust way to get this answer is based on where the class was defined. Right now Wing SDK is the only library in Wing's ecosystem, but in the future there will be others. I'm thinking perhaps we could simply store the name of the library / file where the class was defined, like so:

```suggestion
  /**
   * The source file or library where the resource was defined.
   */
  public sourceModule?: string;
```

This could take on the name of a library, or the name of a source file, like "@winglang/sdk", "wing-redis", "main.w", etc.

libs/wingsdk/src/std/resource.ts
polamoros: I can use the title instead, for sure 👍

libs/wingsdk/src/std/resource.ts
polamoros: I replaced the `type` attribute for `sourceModule` and set it as `@winglang/sdk` for the Functions generated by the SDK. 

Let me know if you have a better idea!

Chriscbr (approved):

skyrpex (approved):

Issue #3653, opened 2023-07-31
Implement `Struct.tryFromJson`


hasanaburayyan:
### Feature Spec

Wing supports `tryFromJson()` methods on user defined Structs which will return an optional struct.

### Use Cases

struct Foo {
  f: str;
}

let j = { f: "bar" };

let f = Foo.fromJson(j);

if let f = f {
 // success!
}

### Implementation Notes

I split this into a follow up issue to not be included in https://github.com/winglang/wing/pull/3648 because this will require some changes to our baby generics and Id like to not add more scope to the existing PR. 

(assigning to myself as well to implement as a followup)

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.25.1](https://github.com/winglang/wing/releases/tag/v0.25.1).

Issue #3575, opened 2023-07-23
`Map.get()` should throw if there is no value, and we should also offer `tryGet()`


eladb:
### I tried this:

```js
let m = { "foo" => "bar" };
let s: str = m.get("zoo");
assert(s.length > 0);
```

### This happened:

```
ERROR: Cannot read properties of undefined (reading 'length')
```

### I expected this:

I expected this to throw.

I also expected to find `tryGet()` to return `T?`.

### Is there a workaround?

If we assign the result to a `T?`, it works.

```js
let m = { "foo" => "bar" };
let s: str? = m.get("zoo");
assert(s == nil);
```

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

giovanni-orciuolo:
Hello! Can I work on this issue? I want to get my hands dirty with the codebase and this seems like a good starting point.

ekeren:
Yes, it does make a lot of sense to start with this

ekeren:
I've assigned it to you @giovanni-orciuolo , let me know if you need any help with this

revitalbarletz:
Hi @giovanni-orciuolo ! How about joining our slack channel? https://t.winglang.io/slack

staycoolcall911:
Keep

monadabot:
Congrats! :rocket: This was released in [Wing 0.38.0](https://github.com/winglang/wing/releases/tag/v0.38.0).

Issue #1868, opened 2023-03-21
Add `tryAt` (and other `tryXXX` methods) to standard library APIs


eladb:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

When using `array.at(i)`, if the element does not exist in the array, an error is thrown. The new `tryAt()` method can be used to return a optional:

```js
if let v1 = my_array.tryAt(i) {
  print("array contains a value: ${v1}");
}

// or
let v2 = my_array.tryAt(i) ?? "default_value";
print(v2);

// etc...
```

This also applies to other retrieval standard collection APIs.

### Use Cases

It is a common use.

### Implementation Notes

I believe that currently `at()` does _not_ throw an error, which is a bug.

### Component

Compiler, SDK

Chriscbr:
If `array.at(i)` is going to be equivalent to `array[i]`, what is `array.try_at(i)` going to be equivalent to?

eladb:
Why do we need a shorthand syntax for `try_at()`?

revitalbarletz:
@eladb what other methods were you thinking? tryConcat and other friends? or anything specific?

giovanni-orciuolo:
This is related to issue #3575. I'm in the process of implementing `tryGet` (even though I'm in the process of getting familiarity with the codebase).

staycoolcall911:
The `Array.tryAt()` method has been added - I'll close this for housekeeping until we get more specific requirements to add any additional `tryXXX` methods. 
Feel free to reopen if you disagree.

PR #3741, opened 2023-08-08
fix(compiler): could not hydrate optionals

hasanaburayyan:
There are several open issues in progress that need to be able to define generic optionals. 
Such as: 
- https://github.com/winglang/wing/issues/1868
- https://github.com/winglang/wing/issues/3575
- https://github.com/winglang/wing/issues/3653

I only added `tryAt` method for Array since the actual work is defined in another issue, but I needed an example to write some test cases with. Since there are several efforts that need this change I opted to not wait and include it in one of those PRs but just add the support here so other efforts can rebase.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -3931,38 +3931,34 @@ impl<'a> TypeChecker<'a> {
 		} else {
 			// Handle generic return types
 			// TODO: If a generic class has a method that returns another generic, it must be a builtin
+			if let Type::Optional(t) = *type_to_maybe_replace {
+				let concrete_t = self.get_concrete_type_for_generic(t, types_map);
+				return self.types.add_type(Type::Optional(concrete_t));
+			}
 
-			match *type_to_maybe_replace {
-				Type::Optional(t) => {
-					let concrete_t = self.get_concrete_type_for_generic(t, types_map);
-					return self.types.add_type(Type::Optional(concrete_t));
-				}
-				_ => {
-					if let Some(c) = type_to_maybe_replace.as_class() {
-						if let Some(type_parameters) = &c.type_parameters {
-							// For now all our generics only have a single type parameter so use the first type parameter as our "T1"
-							let t1 = type_parameters[0];
-							let t1_replacement = *types_map
-								.get(&format!("{t1}"))
-								.filter(|(o, _)| t1.is_same_type_as(o))
-								.map(|(_, n)| n)
-								.expect("generic must have a type parameter");
-							let fqn = format!("{}.{}", WINGSDK_STD_MODULE, c.name.name);
+			if let Some(c) = type_to_maybe_replace.as_class() {
+				if let Some(type_parameters) = &c.type_parameters {
+					// For now all our generics only have a single type parameter so use the first type parameter as our "T1"
+					let t1 = type_parameters[0];
+					let t1_replacement = *types_map
+						.get(&format!("{t1}"))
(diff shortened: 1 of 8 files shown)


hasanaburayyan:
CC: @revitalbarletz

libs/wingc/src/type_check.rs
```diff
@@ -3932,28 +3931,38 @@ impl<'a> TypeChecker<'a> {
...
-						WINGSDK_MUT_SET => self.types.add_type(Type::MutSet(t1_replacement)),
-						_ => {
-							self.unspanned_error(format!("\"{}\" is not a supported generic return type", fqn));
-							self.types.error()
+
+			match *type_to_maybe_replace {
+				Type::Optional(t) => {
+					let concrete_t = self.get_concrete_type_for_generic(t, types_map);
+					return self.types.add_type(Type::Optional(concrete_t));
+				}
```
MarkMcCulloh, line 3939: nit: to remove the extra nesting I would just forgo the extra match add this before the original code:

```rust
if let Type::Optional(t) = *type_to_maybe_replace {
  let concrete_t = self.get_concrete_type_for_generic(t, types_map);
  return self.types.add_type(Type::Optional(concrete_t));
}
```

libs/wingsdk/src/std/array.ts
```diff
@@ -38,6 +38,19 @@ export class Array {
...
+   *
+   * @macro ($self$[$args$])
+   *
+   * @param index index of the value to get
+   * @returns the value at the given index, or undefined if the index is out of bounds
+   */
+  public tryAt(index: number): T1 | undefined {
+    index;
+    throw new Error("Macro");
+  }
```
MarkMcCulloh, line 52: Technically the current at() already does this, we just don't have it typed correctly. I guess we should add a macro to the current `at` to throw if undefined instead.

libs/wingsdk/src/std/array.ts
```diff
@@ -38,6 +38,19 @@ export class Array {
     throw new Error("Abstract");
   }
 
+  /**
+   * Try to get the value at the given index
+   *
+   * @macro ($self$[$args$])
```
MarkMcCulloh, line 44: to handle negative indexes? idk I guess this stuff is out of scope for the focus of this PR.

```suggestion
   * @macro ($self$.at($args$))
```

libs/wingsdk/src/std/array.ts
hasanaburayyan: This is included in the description for https://github.com/winglang/wing/issues/1868 which is already being worked on (not by me)

libs/wingc/src/type_check.rs
hasanaburayyan: I constantly forget that you can check for a single case this way.

libs/wingsdk/src/std/array.ts
```diff
@@ -38,6 +38,19 @@ export class Array {
     throw new Error("Abstract");
   }
 
+  /**
+   * Try to get the value at the given index
```
Chriscbr, line 42:
```suggestion
   * Get the value at the given index, returning nil if the index is out of bounds.
```

libs/wingsdk/src/std/array.ts
Chriscbr, line 1: Shouldn't the regular `array.at` function throw a runtime error if the index is out of bounds? The method says it returns a `T` so failing silently doesn't seem right.

MarkMcCulloh (approved):

PR #3743, opened 2023-08-08
fix(compiler): cannot reuse class names across multiple files

### LINKED ISSUE #3640: Compile error when classes in different modules have the same name
### I tried this:

```js
// main.w
bring "./foo.w" as foo;
bring "./bar.w" as bar;

assert(foo.Util.foo() == "foo");
assert(bar.Util.bar() == "bar");

// foo.w
class Util {
  static foo(): str {
    return "foo";
  }
}

// bar.w
class Util {
  static bar(): str {
    return "bar";
  }
}
```

### This happened:

error: Cannot add two files with the same name: ./inflight.Util.js

### I expected this:

It compiles

### Is there a workaround?

Give the classes different names

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

MacOS

### Anything else?

The JSifier needs to emit inflight clients in a way so that classes with the same name still get different generated names. For example, the file stem could be used to distinguish the names like `inflight.foo-Util.js` and `inflight.bar-Util.js`, or an incrementing number could be used.

### Community Notes

Chriscbr:
Fixes #3640

This change affects a lot of files, so I've implemented the changes and updated examples in the first commit, and all snapshot updates in the second commit.

## Checklist

- [x] Tests added (always)


MarkMcCulloh:
Looks like the hash is not deterministic (self mutation ran multiple times)

Chriscbr:
😬  Looking into it

Chriscbr:
Figured out the source of the mutations: the unit tests in wingc are compiling in temporary directories each time so the paths keep changing.

libs/wingc/src/closure_transform.rs
```diff
@@ -131,24 +131,29 @@ impl Fold for ClosureTransformer {
 			ExprKind::FunctionClosure(func_def) => {
 				self.closure_counter += 1;
 
+				// we are creating AST nodes out of thin air so most won't have real spans, but we try to
+				// preserve the original file ID for jsification & error reporting
+				let mut fake_span = WingSpan::default();
```
eladb, line 136: Maybe `Wingspan::for_file(x)`

libs/wingc/src/jsify.rs
```diff
@@ -1325,7 +1326,10 @@ fn get_public_symbols(scope: &Scope) -> Vec<Symbol> {
 }
 
 fn inflight_filename(class: &AstClass) -> String {
-	format!("./inflight.{}.js", class.name.name)
+	let mut hasher = Sha1::new();
+	hasher.update(&class.name.span.file_id);
```
eladb, line 1330: Is it promised that file_id is not a full path? If it is well have a determinism issue, no?

Maybe a simple counter can be sufficient

eladb (approved):

eladb:
This issue bit me as well

libs/wingc/src/jsify.rs
Chriscbr: It is promised that it's not a full path, with the exception of cases where you provided a full path to `wing compile <file>`

libs/wingc/src/jsify.rs
```diff
@@ -1325,7 +1326,10 @@ fn get_public_symbols(scope: &Scope) -> Vec<Symbol> {
 }
 
 fn inflight_filename(class: &AstClass) -> String {
-	format!("./inflight.{}.js", class.name.name)
+	let mut hasher = std::collections::hash_map::DefaultHasher::new();
```
yoav-steinberg: An alternative can be to use the file_id plus a counter to create the file name, something like:
`format!("./inflight.{}${}-{}.js", sanitize(class.name.file_id), inflight_name_counter++, class.name.name)`

Chriscbr: Yeah, I guess I wanted to try something non-counter based to make `.wing` outputs a little more stable, but it's not critical for sure. Let's try the current implementation / see if it causes any issues

eladb:
Yass!

PR #3746, opened 2023-08-08
chore(sdk): update `queue.push()` to use variadic parameter for messages

### LINKED ISSUE #2278: cloud.Queue - push batch of items
### Feature Spec

Push a batch of items into a queue

(suggested by @garysassano on Slack)

### Use Cases

Make it easier to test scenarios where multiple items are added to a queue at once, or are processed in a batch

### Implementation Notes

If we add support for variadics (https://github.com/winglang/wing/issues/125) then we can extend the current API:

```js
let code = inflight () => {
  queue.push("message1", "message2", "message3"); // any number of args
};
```

Or we can have a separate method like:

```js
let code = inflight () => {
  queue.push_batch(["message1", "message2", "message3"]); // any number of args
};
```

### Component

SDK

### Community Notes

garysassano:
Closes #2278


docs/docs/04-standard-library/01-cloud/queue.md
```diff
@@ -172,12 +172,12 @@ Purge all of the messages in the queue.
 ##### `push` <a name="push" id="@winglang/sdk.cloud.IQueueClient.push"></a>
 
 ```wing
-inflight push(message: str): void
+inflight push(messages: str): void
```
Chriscbr, line 175: Looks like a bug in how we're generating documentation in the `jsii-docgen` tool. I think we need some extra logic to make sure variadic parameters like this one are displayed as `...message: Array<str>`. Can we open an issue about this?

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -58,13 +58,15 @@ export class Queue
...
+  // TODO: enforce maximum queue message size?
+  public async push(...messages: string[]): Promise<void> {
+    messages.forEach(async (message) => {
+      await this.context.withTrace({
+        message: `Push (message=${message}).`,
+        activity: async () => {
+          this.messages.push(new QueueMessage(this.retentionPeriod, message));
+        },
+      });
     });
```
Chriscbr, line 70: By using `forEach` like this, I'm not if all of the promises will be awaited. For example in this simplified version, if you try running it, you'll see it prints "done" before "task1" or "task2".

```js
async function main() {
    let sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
    let task1 = sleep(1000).then(() => console.log("task1"));
    let task2 = sleep(1000).then(() => console.log("task2"));
    let tasks = [task1, task2];
    tasks.forEach(async (task) => {
        await task;
    });
    console.log("done");
}
main();
```

Instead I think we need to use something like [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) or a regular for loop...

libs/wingsdk/src/cloud/queue.ts
Chriscbr, line 1: We need new unit tests [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/test/target-sim/queue.test.ts) and [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/test/shared-aws/queue.inflight.test.ts)

Chriscbr (changes requested):

libs/wingsdk/src/target-sim/queue.inflight.ts
garysassano: Ok, I didn't know about the differences of how `forEach` handles promises internally. I tested both `Promise.all` and `for...loop` but I couldn't notice a significant performance difference. My bias was that the former one would be more efficient since it can await multiple promises in parallel and stop execution as soon as one rejects.

garysassano: Ok, I didn't know that `forEach()` cannot handle promises properly. I tested both `Promise.all` and `for...loop` but I couldn't notice a significant performance difference. I thought that the former one would be more efficient since it can await multiple promises in parallel and stop execution as soon as one rejects.

docs/docs/04-standard-library/01-cloud/queue.md
garysassano: Done: #3781

garysassano:
@Chriscbr I was wondering if it wouldn't be better to preserve the `push` method while introducing an additional `pushBatch` method that makes use of the variadic parameter. This new method could leverage the [SendMessageBatch](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html) API for AWS targets.

Chriscbr:
@garysassano What would the pros and cons be?

garysassano:
@Chriscbr Not many things come to my mind, but probably:
- **Pro:** you can make a single API call to AWS instead of multiple ones.
- **Con:** if you fail to push a message, you still get a HTTP 200 returned because the batch operation is considered successful even if some operations fail, so you need to check for batch errors.

libs/wingsdk/src/cloud/queue.ts
garysassano: Unit tests added.

garysassano:
I think it's fine to just keep a single method for now.

Chriscbr (approved):

PR #3757, opened 2023-08-09
fix(sdk): website has no url protocol on tf-aws

### LINKED ISSUE #3529: Website url failed to parse inflight test function
### I tried this:

```
bring cloud;
bring util;
bring http;

let website = new cloud.Website(
  path: "./static",
);

// workaround for https://github.com/winglang/wing/issues/3289
// this shouldn't be necessary, since api.url should
// be directly accessible  in the test
// let apiUrl = api.url;

// However, while this worked for API, it doesn't work for
// the website.url property :/
let websiteUrl = website.url;

let invokeAndAssert = inflight(response: http.Response, expected: str) => {
  log("response: ${response.status} ");
  assert(response.status == 200);
  assert(response.body?.contains(expected) == true);
};

// Doesn't work right now
test "renders the index page" {
  invokeAndAssert(http.get(websiteUrl), "Hello, Wing");
}

```

### This happened:

```
fail ┌ main.tfaws » root/Default/env0/test:renders the index page
     │ Error: Invoke failed with message: "Unhandled". Full error: "{"errorType":"TypeError","errorMessage":"Failed to parse URL from d28y7bxgj0r2sp.cloudfront.net","trace":["TypeError: Failed to parse URL from d28y7bxgj0r2sp.cloudfront.net","    at Object.fetch (node:internal/deps/undici/undici:11457:11)","    at async Util.fetch (/var/task/index.js:1619:21)","    at async $Closure4.handle (/var/task/index.js:16:34)","    at async exports.handler (/var/task/index.js:1751:10)"]}"
     │     at FunctionClient.invoke (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/shared-aws/function.inflight.js:21:19)
     │     at async TestRunnerClient.runTest (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/shared-aws/test-runner.inflight.js:27:13)
     │     at async /Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/commands/test.js:295:30
     │     at async withSpinner (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/util.js:55:24)
     │     at async testTfAws (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/commands/test.js:292:25)
     │     at async testOne (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/commands/test.js:124:20)
     │     at async Object.test (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/commands/test.js:54:39)
     └     at async Command.<anonymous> (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/dist/cli.js:48:30)
```

### I expected this:

no error

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

0.24.31

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

Seems related to https://github.com/winglang/wing/issues/3289

### Community Notes

tsuf239:
fixes #3529 

## Checklist

- [x] Tests added (always)


eladb:
@tsuf239 please check the [pull request guidelines] for writing titles. The title for `fix` should reflect what's not working. In this case, it should be something like `fix(sdk): fetch fails without url protocol`. Or something like this.

[pull request guidelines]: https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted

libs/wingsdk/src/http/http.ts
```diff
@@ -181,7 +181,11 @@ export class Util {
     url: string,
     options?: RequestOptions
   ): Promise<Response> {
-    const res = await fetch(url, { ...defaultOptions, ...options });
+    const hasProtocol = url.match(/^(.*):\/\/(.*)/);
```
eladb, line 184: This feels like a bad practice. I will be more comfortable with just failing with a nice error if the URL doesn't have a protocol and documenting that the URL requires a protocol. Assuming `http://` could be unsafe and surprising.

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -126,7 +126,9 @@ export class Website extends cloud.Website {
       policy: allowDistributionReadOnly.json,
     });
 
-    this._url = distribution.domainName;
+    const url = distribution.domainName;
+
+    this._url = url.match(/^(.*):\/\/(.*)/) ? url : `https://${url}`;
```
eladb, line 131: Do we really need this test? I think domain name by definition doesn't have a protocol, no?

eladb (changes requested):

libs/wingsdk/src/target-tf-aws/website.ts
tsuf239: I think you're right, I just wonder what will happen when we'll let the user apply their own domain, instead of having the default cloudFront one... I think it's ok to be on the safe side here

eladb: Then the check should be when we process user input. It's okay to assume that `domainName` is just a domain name and not a URL.

As a rule of thumb, we want to maintain the integrity of an object on the surface area.

For example:

```js
class MyClass {
  positiveNumber: num;

  setNumber(x: num) {
    // verify that the "invariant" is kept and `positiveNumber` is always positive.
    if (x < 0) { throw("error"); }

    this.positiveNumber = x;
  }

}
```

tsuf239: ok, removed the test

eladb (approved):

PR #3398, opened 2023-07-12
fix(cli): breaks without globally installed aws-cdk-lib

Chriscbr:
Fixes #3396

This "instanceof" check is causing problems since it results in several other modules to get imported, including those that require `aws-cdk-lib` to be installed -- even if you're not compiling to the `awscdk` target. Removing this code path as a hotfix.

## Checklist

- [ ] Title matches [Winglang's style guide](https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted)
- [ ] Description explains motivation and solution
- [ ] Tests added (always)
- [ ] Docs updated (only required for features)
- [ ] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Monada Contribution License](https://www.winglang.io/terms-and-policies/contribution-license.html)*.

PR #3769, opened 2023-08-10
fix(sdk): custom bind method on awscdk targets

marciocadev:
Following @eladb's recommendation in PR #3398, I removed the import of `target-awscdk` and `target-tf-aws`, and checked if the `host` implements the methods of the `IAwsFunction` interface. This way, the error that was occurring due to not having `aws-cdk-lib` globally installed no longer happens.

## Checklist

- [x] Tests added (always)


examples/tests/valid/dynamo_awscdk.w
```diff
@@ -0,0 +1,96 @@
+/*\
+skip: true
```
eladb, line 2: @tsuf98 is this still going to run in our integration tests on aws?

examples/tests/valid/dynamo_awscdk.w
```diff
@@ -0,0 +1,96 @@
...
+      billingMode: awscdk.aws_dynamodb.BillingMode.PAY_PER_REQUEST,
+      partitionKey: awscdk.aws_dynamodb.Attribute {
+        name: "Flavor",
+        type: awscdk.aws_dynamodb.AttributeType.STRING,
+      },
+    );
+    this.tableName = this.table.tableName;
+  }
+
+  _bind(host: std.IInflightHost, ops: Array<str>) {
```
eladb, line 42: @Chriscbr it feels weird that people need to override an internal method... maybe we should formalize it? (At least change it to `bind()`.

(Would be nice to have something like a "private protected" method which can only be overridden by derived classes but not called from outside)

examples/tests/valid/dynamo_awscdk.w
```diff
@@ -0,0 +1,96 @@
...
+
+let table = new DynamoTable();
+
+test "put an item in the table" {
+  table.putItem({
+    "Flavor" => Attribute {
+      type: AttributeType.String,
+      value: "Chocolate",
+    },
+  });
```
eladb, line 95: No assertion in the test

libs/wingsdk/src/shared-aws/function.ts
```diff
@@ -30,10 +29,18 @@ export class Function {
...
   public static from(host: IInflightHost): IAwsFunction | undefined {
-    if (host instanceof TfAwsFunction) {
+    if (this.checkIAwsFunctionImplementation(host)) {
       return host;
     }
-
     return undefined;
   }
+
+  private static checkIAwsFunctionImplementation(
```
eladb, line 38:
```suggestion
  private static isAwsFunction(
```

examples/tests/valid/dynamo_awscdk.w
tsuf239: Only when running tests through hangar, we don't have it built-in in the CLI

examples/tests/valid/dynamo_awscdk.w
Chriscbr: I agree. Let's make it public

marciocadev: Do you want me to make the changes in this PR, or should I open another one?

Chriscbr: In another PR is fine 👍

marciocadev: Open a issue #3775 and a PR #3776

Chriscbr (approved):

PR #3770, opened 2023-08-10
chore(sdk): adding auto retry to the sdk gh workflow

tsuf239:
From the few runs I looked into, it seems like the reason for failure is either a network error installing wing (is it blocked by NPM in any way?) or occasional fails in the tests themselves...
I added a retry option to make them more stable:
https://github.com/winglang/wing/actions/runs/5818534372 

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/workflows/tf-aws-test.yml
+++ b/.github/workflows/tf-aws-test.yml
@@ -51,6 +51,7 @@ jobs:
         with:
           max_attempts: 3
           retry_on: error
+          timeout_minutes: 5
           command: npm install -g winglang
       - name: Installing external js modules
         run: |
@@ -70,6 +71,7 @@ jobs:
         with:
           max_attempts: 3
           retry_on: error
+          timeout_minutes: 20
           command: wing test -t tf-aws ${{ matrix.test.directory }}/*.w
       - name: Output Terraform log
         if: failure()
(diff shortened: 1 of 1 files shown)


.github/workflows/tf-aws-test.yml
```diff
@@ -56,10 +61,15 @@ jobs:
...
+        uses: nick-fields/retry@v2
         env:
           TF_LOG: info
           TF_LOG_PATH: ${{ runner.workspace }}/terraform.log
-        run: wing test -t tf-aws ${{ matrix.test.directory }}/*.w
+        with:
+          max_attempts: 3
+          retry_on: error
+          timeout_minutes: 20
+          command: wing test -t tf-aws ${{ matrix.test.directory }}/*.w
```
eladb, line 72: I kind of feel like maybe we want to add this retry in `wing test`, no?

tsuf239: Good point, I'll open an issue!
From what I experienced so far while developing and heavily using this command, it hardly fails for "no reason", and when it fails I want to respond the fastest to save my time (if I ran a whole directory and one file failed, next time I'll run only the failed once, and only after adjusting the code, I hardly run the tests without changing something at my end first), this retry option will mostly be valued for automation when we want to make sure something isn't working just because of "environmental factors" rather a problem in the code, it still be good and make the overall testing time shorter- running over only the failed ones.

tsuf239: https://github.com/winglang/wing/issues/3796

Issue #3024, opened 2023-06-22
`cloud.api` GET request body is an empty json in sim


tsuf239:
### I tried this:

created an API instance with a GET endpoint:
```
bring cloud;
let api = new cloud.Api();

api.get("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
    assert(req.body?.length == 0);
    
    return cloud.ApiResponse {
        status: 200,
        body: body
    };
});
```
preformed a get request to the end point with an empty body

### This happened:

assertion fails in sim. the body equals to an empty json (was verified later with other assert)

### I expected this:

the body to be a string and an empty one.

### Component

SDK

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

tsuf239:
seems to be resolved somehow

tsuf239:
Wasn't resolved in sim, reopening

monadabot:
Congrats! :rocket: This was released in [Wing 0.25.2](https://github.com/winglang/wing/releases/tag/v0.25.2).

Issue #3539, opened 2023-07-20
api.url is not working on aws, wrong WING_TOKEN_TFTOKEN number


ekeren:
### I tried this:

```ts wing
bring cloud;
bring ex;
bring http;
bring util;

let api = new cloud.Api();
// counter to generate uuid
let counter = new cloud.Counter() as "uuid generator";
// our employee database
let db = new ex.Table(
  name: "employees",
  primaryKey: "id",
  columns: {
    "name" => ex.ColumnType.STRING
  }
);

api.get("/employees", inflight (request: cloud.ApiRequest): cloud.ApiResponse => {
  let result = MutJson [];
  let var i = 0;
  for employee in db.list() {
    result.setAt(i, employee);
    i = i + 1;
  }
  return cloud.ApiResponse {
    status: 200,
    body: Json.stringify(result)
  };
});


api.get("/employees/{id}", inflight (request: cloud.ApiRequest): cloud.ApiResponse => {
  let employee = db.get(request.vars.get("id"));
  if let id = employee.tryGet("id") {
    return cloud.ApiResponse {
      status: 200,
      body: Json.stringify(employee)
    };
  }
  return cloud.ApiResponse {
    status: 404
  };
});

api.post("/employees", inflight (request: cloud.ApiRequest): cloud.ApiResponse => {
   if let body = request.body {
    let employeeData = Json.parse(body);
    let id = "${counter.inc()}";
    db.insert(id, employeeData);
    return cloud.ApiResponse {
      status: 201,
      body: id
    };
   }
});

api.put("/employees/{id}", inflight (request: cloud.ApiRequest): cloud.ApiResponse => {
  if let body = request.body {
    let employeeData = Json.parse(body);
    let id = request.vars.get("id");
    try {
      db.update(id, employeeData);
    } catch {
      return cloud.ApiResponse {
        status: 404
      };
    }
    return cloud.ApiResponse {
      status: 200,
      body: Json.stringify(employeeData)
    };
  }
});

api.delete("/employees/{id}", inflight (request: cloud.ApiRequest): cloud.ApiResponse => {
  let id = request.vars.get("id");
  try {
    db.delete(id);
  } catch {
    return cloud.ApiResponse {
      status: 404
    };  
  }
  return cloud.ApiResponse {
    status: 204
  };
});

// TEST Section

inflight class AssertUtils { 
  static assertNumEqual(expected: num? , actual: num?) {
    if expected != actual {
[...]

Issue #3342, opened 2023-07-10
`api.url` isn't working on aws 


tsuf239:
### I tried this:

```
bring cloud;
let api = new cloud.Api();

api.get("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {

  return cloud.ApiResponse {
    status: 200,
    body: "ok"
  };
});

log(api.url); // logs the token

let apiUrl = api.url;

test "main" {
  log(apiUrl); // a string with " at the beginning and " at the end :(
  log(api.url); // this will be undefined (therefore throws "Unexpected token u in JSON at position 0") see https://github.com/winglang/wing/blob/7bbc81ff26a97b799dafa2a976dab4d32135cd2e/libs/wingsdk/src/shared-tf/tokens.ts#L22-L26
}
```

### This happened:

either getting an error or a URL wrapped in quotation marks 

### I expected this:

getting no errors and valid string 

### Component

Compiler, SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skorfmann:
Seems like this is related to https://github.com/winglang/wing/issues/3289

tsuf239:
fixed by #3772

PR #3772, opened 2023-08-10
fix(sdk): fixing `api.url` token inconsistency

tsuf239:
Fixes: #3024 #3539 #3342 
The problem with the token was the way cdktf is tokenizing strings, I assume it tried to create a new token every time we call `api.url`, or maybe tokenize a token. Now they're all the same- no matter if you call it on preflight or inflight it or assign it to another variable- it'll generate exactly the same token number.

other than that:
- brought back the aws api SDK tests
- changed sim response to be an empty string (in express it's an empty object by default)

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -414,6 +415,9 @@ class WingRestApi extends Construct {
       stageName: STAGE_NAME,
       deploymentId: this.deployment.id,
     });
+
+    //should be exported from here, otherwise won't be mapped to the right token
+    this.url = this.stage.invokeUrl;
```
eladb, line 420: @skorfmann I am okay with this as a point solution to the `api.url` issue, but that's a more general problem with CDKTF.

Do we have a CDKTF issue to associate with this? Any chance you can pick this up upstream?

libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -179,7 +179,7 @@ function isApiResponse(response: unknown): response is ApiResponse {
 function transformRequest(req: express.Request): ApiRequest {
   return {
     headers: sanitizeParamLikeObject(req.headers),
-    body: req.body,
+    body: Object.keys(req.body).length > 0 ? req.body : "",
```
eladb, line 182: Are we sure `req.body` is an object and not e.g. a string?

eladb (approved):

libs/wingsdk/src/target-sim/api.inflight.ts
tsuf239: if the body is empty it'll be an empty object, by default, if the body is a non-empty string or a non-empty object `Object.keys(req.body).length` results in a number rather than 0 
(if the body is a string it counts the number of chars, so if it's empty by purpose it'll still return an empty string)

PR #3776, opened 2023-08-10
feat(sdk): changing the bind method from internal to public

### LINKED ISSUE #3775: Change the internal bind method to public
### Feature Spec

Changing the `bind`method from internal to public

### Use Cases

When creating a new custom resource with Wing, it's necessary to implement the binding. Currently, it's a method tagged as internal. It makes sense to change it to public and allow its implementation if necessary.

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

marciocadev:
This PR changes the _bind methods, making them public and removing the internal tag. It makes more sense to make this method public, as a Wing class can rewrite the method when necessary.

Closes #3775 

## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/docs/04-standard-library/01-cloud/api.md
+++ b/docs/docs/04-standard-library/01-cloud/api.md
@@ -149,7 +149,6 @@ new cloud.Api(props?: ApiProps);
 
 | **Name** | **Description** |
 | --- | --- |
-| <code><a href="#@winglang/sdk.cloud.Api.bind">bind</a></code> | Binds the resource to the host so that it can be used by inflight code. |
 | <code><a href="#@winglang/sdk.cloud.Api.connect">connect</a></code> | Add a inflight handler to the api for CONNECT requests on the given path. |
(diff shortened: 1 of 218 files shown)


docs/docs/04-standard-library/01-cloud/api.md
```diff
@@ -160,6 +161,31 @@ new cloud.Api(props?: ApiProps);
 
 ---
 
+##### `bind` <a name="bind" id="@winglang/sdk.cloud.Api.bind"></a>
```
Chriscbr, line 164: Can we hide this from the docs? It should be possible to add it to the list of hidden methods here: https://github.com/winglang/wing/blob/8df5d0e87549e61105710c1a0a18191d1abb2179/apps/jsii-docgen/src/docgen/view/wing-filters.ts#L10

The method isn't meant to be directly called in Wing applications, so I think it might be more confusing to have it shown here.

PR #3777, opened 2023-08-10
fix: hangar snapshots have machine-dependent output

Chriscbr:
In #3743 I introduced some code to avoid name collisions in our code generation by hashing file paths. It seems this was adequate for many cases, but it caused problems for contribution since our `hangar` E2E tests currently compile most Wing programs using absolute paths. That means in these scenarios the hashes will be machine dependent. According to @MarkMcCulloh these tests are needed since there were bugs with absolute paths in the past. To that end, this PR changes the name deduplication strategy to be based around incrementing numbers.

## Checklist

- [ ] Tests added (always)


MarkMcCulloh (approved):
🙏

PR #3780, opened 2023-08-10
feat(sdk)!: remove cloud.Queue initialMessages

### LINKED ISSUE #281: Support cloud.Queue `initialMessages` prop for AWS target

Chriscbr:
Removes the `initialMessages` property from `cloud.Queue` to reduce API surface area in the standard library. This capability has only worked on the `sim` target to date, and it's possible to easily achieve the capability (across clouds) using `cloud.OnDeploy`:

```js
bring cloud;

new cloud.OnDeploy(inflight () => {
  queue.push("hello");
  queue.push("world");
});
```

This style of API would have more merit if it's something we could provided an optimized implementation for, but it doesn't appear there are dedicated Terraform resources available that represent AWS SQS messages or Google PubSub items. `OnDeploy` also offers more flexibility since it allows the user to dynamically calculate the queue's initial messages based on the results of other inflight operations.

Closes #281.

BREAKING CHANGE: The field `initialMessages` of `cloud.Queue` has been removed. Instead, use the `cloud.OnDeploy` resource to perform any deploy-time initialization logic.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/docs/04-standard-library/01-cloud/queue.md
+++ b/docs/docs/04-standard-library/01-cloud/queue.md
@@ -239,22 +239,8 @@ let QueueProps = cloud.QueueProps{ ... };
 
 | **Name** | **Type** | **Description** |
 | --- | --- | --- |
-| <code><a href="#@winglang/sdk.cloud.QueueProps.property.initialMessages">initialMessages</a></code> | <code>MutArray&lt;str&gt;</code> | Initialize the queue with a set of messages. |
 | <code><a href="#@winglang/sdk.cloud.QueueProps.property.retentionPeriod">retentionPeriod</a></code> | <code><a href="#@winglang/sdk.std.Duration">duration</a></code> | How long a queue retains a message. |
 | <code><a href="#@winglang/sdk.cloud.QueueProps.property.timeout">timeout</a></code> | <code><a href="#@winglang/sdk.std.Duration">duration</a></code> | How long a queue's consumers have to process a message. |
-
----
-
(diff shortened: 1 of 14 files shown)


eladb (approved):
Love it

ekeren:
@Chriscbr, Why not use the same mechanism to remove the need for `bucket.addObject`
```
new cloud.OnDeploy(inflight () => {
  bucket.put("file.txt", "some text");
});
```

eladb:
`addObject()` is a common use case for buckets in production (as oppose to `initialMessages` which would normally be used for development purposes), so I think we should leave `addObject()`.

ekeren:
> `addObject()` is a common use case for buckets in production (as oppose to `initialMessages` which would normally be used for development purposes), so I think we should leave `addObject()`.

Out of curiosity, does it makes sense to re-implement `addObject` in the abstraction layer using cloud.onDeploy?

Chriscbr:
@ekeren I addressed this in the PR description:

> This style of API would have more merit if it's something we could provided an optimized implementation for, but it doesn't appear there are dedicated Terraform resources available that represent AWS SQS messages or Google PubSub items.

In other words, for `addObject()`, we're able to implement it today using Terraform's [aws_s3_object](https://registry.terraform.io/providers/figma/aws-4-49-0/latest/docs/resources/s3_object) resource, so there's no need to create any compute resources. 

We could implement `addObject` using `cloud.OnDeploy`, but `cloud.OnDeploy` requires several more Terraform resources (we have to bundle your code, creating an IAM role, etc.), so it would be a much heavier solution to the use case (without any benefits as far as I can tell).

PR #3791, opened 2023-08-11
chore(repo): skip hangar when not needed

MarkMcCulloh:
### Goal
Skip e2e tests when the changes in the PR don't effect it. Notably, this usually means unit tests, README, or anything in the console.

"changes" here means one of 2 things:
- **If non-pr:** Last successful build workflow on the same branch
- **If pr:** git diff of the PR compared to the base, ignoring any commits from upstream

This high-level logic is the same that we've been using to determine whether or not we deploy the vscode extension, so this extends it to the e2e and benchmarking tests as well.
Unfortunately turbo itself was not sufficient for this goal (see note at the top of `tools/bump-pack/src/turbo-diff.ts`), so I had to write some junk to properly determine "what changed". I swear I've had to write something like this like 9 times already at various points in my life.

#### Limitations

The "Last successful build workflow on the same branch" does not apply to PRs, even though I wish it did. Implementing this safely is more difficult than it seems.

### The future of bump-pack

Part of the this PR is also effectively turning "bump-pack" into a general-purpose toolkit. In a followup PR I plan to simply turn the entire package into a full CLI to serve misc repo development purposes. My hope is that this will make it easier to actually add testing to it with a fixture based on https://github.com/kiegroup/mock-github and https://github.com/kiegroup/act-js

### Misc

- Added a script for the git patching stuff, didn't like so much repetitious bash in the workflow
- ~Effectively forked `nrwl/nx-set-shas` since it was not sufficient for the workflow I wanted here (It always assumed you wanted to compare against `main` in a PR)~ Nevermind, it turns out that the original action is actually good enough


libs/wingc/turbo.json
```diff
@@ -3,16 +3,17 @@
   "extends": ["//"],
   "pipeline": {
     "compile": {
+      "dependsOn": ["@winglang/wingii#compile", "@winglang/tree-sitter-wing#compile"],
```
MarkMcCulloh, line 6: wingc compile does not depend on `^compile` because it does not need `@winglang/sdk#compile`

scripts/create_patch.sh
Chriscbr, line 1: Good call moving this to the scripts folder 👍

turbo.json
```diff
@@ -7,7 +7,9 @@
     "insta.yaml",
     ".github/workflows/build.yml",
     "scripts/*",
-    "patches/*"
+    "patches/*",
+    "tools/bump-pack/**",
+    "!tools/bump-pack/node_modules/**"
```
Chriscbr, line 12: Should we ignore all `node_modules`?

tools/bump-pack/src/turbo-diff.ts
Chriscbr, line 1: Could you add a description to the top of the file that describes what the script does? What's the expected state of the world before running it, and what's the expected state of the world after?

tools/bump-pack/src/derive-shas.ts
Chriscbr, line 1: Add description (see other comment)

tools/bump-pack/src/turbo-diff.ts
```diff
@@ -0,0 +1,116 @@
...
+  const dataEntry = taskData[task.taskId];
+  if (dataEntry.changes) {
+    continue;
+  }
+  let dependencies = task.dependencies;
+  if (task.package === "hangar") {
+    // ignore the wing console
+    dependencies = dependencies.filter(
+      (dependency: string) => !dependency.startsWith("@wingconsole")
+    );
```
Chriscbr, line 105: How come the Wing console is ignored?

Chriscbr (approved):
LGTM. It looks like this change will improve the contributing experience a lot - let's ship it!

I wonder if we should only skip build steps on PR branches so that each commit on `main` still has a full build? (I'm not too sure)

MarkMcCulloh:
> I wonder if we should only skip build steps on PR branches so that each commit on main still has a full build? (I'm not too sure)

I'm not sure either.
In theory, if we trust this for PRs then we should trust it for main too. The process _should_ be deterministic. I do acknowledge the risk though. This would be mitigated if we started running the canary test inline with the build, and always do it regardless of the change.

Honestly I kind of want to try this unsafe route, but if literally anyone thinks we shouldn't then I don't mind immediately changing that.

Chriscbr:
Will benchmarks still be run on every commit to main?

tools/bump-pack/src/turbo-diff.ts
MarkMcCulloh: `@wingconsole` stuff is a transitive dependency of hangar (`hangar->winglang->@console/*`). By definition then, hangar also depends on it and turbo knows that. However, the hangar does interact with the console at all, so it's functionally a dependency. 

There is no way to tell turbo to ignore a transitive dependency for a specific project. So that's what this is here. I can add this note to the file

MarkMcCulloh:
> Will benchmarks still be run on every commit to main?

No, for the same reason

turbo.json
MarkMcCulloh: Shouldn't be needed since other globs are more specific, while `tools/bump-pack/**` isn't. Will double check others though.

PR #3793, opened 2023-08-12
feat(sdk): sim.Object

Chriscbr:
Today, it can be difficult or awkward to write `sim` implementations for user-defined Wing classes because there isn't a way to persist data across multiple inflight consumers. To illustrate this point, suppose you create a class in Wing that is implemented as follows:

```js
class Table {
  inflight items: MutArray<num>;
  inflight putItem(x: num) {
    this.items.push(x);
  }
}
```

Since `items` is an inflight field, when `putItem` is called by a resource like a `cloud.Function`, the array mutation will only be visible to that running `cloud.Function` instance. Other instances of `cloud.Function` or other compute resources will not see the change. To make an analogy to the operating systems world, inflight class members are a lot like thread-local variables -- they may persist as long as a piece of compute is running, but gets cleared as soon as that thread (or compute instance) is killed.

It's possible to work around this by using one of the existing SDK resources like `cloud.Bucket` or `cloud.Queue` etc. that have data persistence built in. However this doesn't always feel idiomatic, since `cloud.Bucket` doesn't provide strong typing for the data stored inside, thus requiring you to write data wrapping / unwrapping logic yourself. `cloud.Bucket` can also only store serializable data.

To that end, I'm proposing a dedicated resource / class in the standard library called `sim.SharedState` for sharing mutable data between inflight consumers. The class is only available on the `sim` target. Example usage:

```js
class QueueSim {
  messages: sim.SharedState;
  init() {
    this.messages = new sim.SharedState(MutArray<str>[]);
  }
  inflight push(message: str) {
    let messages = this.messages.get();
    messages.push(message);
  }
  inflight pop(): str {
    let messages = this.messages.get();
    return messages.pop();
  }
}
```

So far this seems most useful for sharing mutable types like `MutArray`, `MutSet`, `MutMap`, and `MutJson`.

Any and all ideas / feedback welcome. 🙂 

Some of the questions that crossed my mind:

- Is it possible to make this class parameterized like some other built-in types (`SharedState<T>`)? What changes to the compiler would this need?
- How could we support initializing `SharedState` with a lifted value like a `cloud.Bucket` client or an inflight class instance? Perhaps instead of accepting `initial: any` in the constructor it should accept `initializer: inflight () -> any`?
- Alternative names?
- Would it make sense to make this a dedicated language feature or keyword at some point? Calling `.get()` and `.set()` seems like extra work for the user that could be desugared away.

TODO:
- [ ] unit tests
- [ ] rename impl to sim.Object or sim.SharedState

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/examples/tests/sdk_tests/std/shared-state.w
+++ b/examples/tests/sdk_tests/std/shared-state.w
@@ -1,0 +1,33 @@
+bring cloud;
+
+class QueueSim {
+  rawMessages: std.SharedState;
+
+  init() {
+    this.rawMessages = new std.SharedState(MutArray<str>[]);
+  }
+
+  inflight push(message: str) {
+    let rawMessages = this.rawMessages.get();
+    rawMessages.push(message);
+  }
+
+  inflight pop(): str {
+    let rawMessages = this.rawMessages.get();
+    return rawMessages.pop();
+  }
+}
+
+let queue = new QueueSim();
+
+let pusher = new cloud.Function(inflight () => {
+  queue.push("hello");
+}) as "pusher";
+let popper = new cloud.Function(inflight (): str => {
+  return queue.pop();
+}) as "popper";
+
+test "push and pop from different functions" {
+  pusher.invoke("");
+  assert(popper.invoke("") == "hello");
+}
--- a/libs/wingsdk/src/std/index.ts
+++ b/libs/wingsdk/src/std/index.ts
@@ -9,6 +9,7 @@ export * from "./number";
 export * from "./range";
 export * from "./resource";
 export * from "./set";
+export * from "./shared-state";
 export * from "./string";
 export * from "./test";
 export * from "./test-runner";
--- a/libs/wingsdk/src/std/shared-state.ts
+++ b/libs/wingsdk/src/std/shared-state.ts
@@ -1,0 +1,69 @@
+import { Construct } from "constructs";
+import { IResource, Resource } from "./resource";
+import { fqnForType } from "../constants";
+import { App } from "../core";
+
+/**
+ * Global identifier for `SharedState`.
+ */
+export const SHARED_STATE_FQN = fqnForType("std.SharedState");
+
+/**
+ * A container for a value that can be shared and mutated across inflight consumers.
+ * Only supported in the `sim` compilation target.
+ *
+ * @inflight `@winglang/sdk.std.ISharedStateClient`
+ */
+export abstract class SharedState extends Resource implements IResource {
+  /**
+   * Create a new SharedState instance.
+   * @internal
+   */
+  public static _newSharedState(
+    scope: Construct,
+    id: string,
+    initial: any
+  ): SharedState {
+    return App.of(scope).newAbstract(SHARED_STATE_FQN, scope, id, initial);
+  }
+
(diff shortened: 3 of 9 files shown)


MarkMcCulloh:
> Should this class be part of the sim module, i.e. sim.SharedState?

Probably, being in std implies it works in any target.

---

> cloud.Bucket can also only store serializable data.

This is a pretty useful limitation though, I'm not clear how SharedState would work without it. Even in the context of sim we would need a way of transferring the shared data across non-trivial boundaries.

staycoolcall911:
Interesting idea.
Just some sporadic feedback on a specific point:
> However this doesn't always feel idiomatic, since `cloud.Bucket` doesn't provide strong typing for the data stored inside

Once we have generics we could use `cloud.Bucket<Person>` for example.

examples/tests/sdk_tests/std/shared-state.w
```diff
@@ -0,0 +1,33 @@
+bring cloud;
+
+class QueueSim {
+  rawMessages: std.SharedState;
```
eladb, line 4: Something feels wrong about this.

At the minimum, this should be `sim.SharedState` and not `std.SharedState`.

And then maybe it should just be `sim.Object`?

eladb: Maybe this needs a sync discussion...

Chriscbr: > Something feels wrong about this.

Just curious, do you mean just with the name, or with other stuff? (FWIW I'm not 100% sold it's the most elegant thing, but also intuitively feels to me like an abstraction we will need in one shape or another). I'm cool with the rename suggestions 👍 

I'll comment that the fact that you have to create `sim.Object` in preflight init feels a bit restrictive to me -- `sim.Object` might be a good candidate for a future phase-independent class. That is, `sim.Object<T>` can hold any `T` that is also phase-independent (like collections, structs, and so on), and it can be declared in preflight or just in inflight if needed. But until we iron out phase-independence, I think this might be adequate.

eladb: In inflight, an object is just an object, no?

libs/wingsdk/src/std/index.ts
```diff
@@ -9,6 +9,7 @@ export * from "./number";
 export * from "./range";
 export * from "./resource";
 export * from "./set";
+export * from "./shared-state";
```
eladb, line 12: Move to `sim`

libs/wingsdk/src/std/shared-state.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export const SHARED_STATE_FQN = fqnForType("std.SharedState");
+
+/**
+ * A container for a value that can be shared and mutated across inflight consumers.
+ * Only supported in the `sim` compilation target.
+ *
+ * @inflight `@winglang/sdk.std.ISharedStateClient`
+ */
+export abstract class SharedState extends Resource implements IResource {
```
eladb, line 17: I am not sure the abstract version is required. Users should explicitly instantiate the concrete `sim.Object`.

PR #3799, opened 2023-08-13
fix(compiler): remove some dead capturing code

yoav-steinberg:
This is a tiny fix removing dead capturing code. For some reason there's code that tries to capture the parent class of a inflight class in the irrelevant case that we're inside an inflight context. In this case the code is dead because an inflight class within an inflight class generates inlined code.

The PR also adds some coloring improvements to the env printing I used to debug this and fixed a comment related to parent class capturing.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingc/src/lifting.rs
+++ b/libs/wingc/src/lifting.rs
@@ -285,16 +285,6 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
 			None
 		};
 
-		if self.ctx.current_phase() == Phase::Inflight {
-			if let Some(parent) = &node.parent {
-				if self.should_capture_expr(parent) {
-					let mut lifts = self.lifts_stack.pop().unwrap();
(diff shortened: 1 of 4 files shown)


staycoolcall911 (approved):
LGTM

libs/wingc/src/lifting.rs
```diff
@@ -285,16 +285,6 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
 			None
 		};
 
-		if self.ctx.current_phase() == Phase::Inflight {
```
yoav-steinberg, line 288: Note this is dead because it'll reach this code only if the current phase is inflight and the visited class is a preflight class. But it's not possible to define a preflight class within an inflight context. I'm not sure what we were trying to achieve here.
@eladb?

eladb: Safe to delete. It's a leftover I hadn't had a chance to clean up!

eladb (approved):

PR #3801, opened 2023-08-13
feat(sdk): `addfile()` is a preflight method which adds file from absolute path

0018akhil:
## Checklist

- [x] Tests added (always)


libs/wingsdk/src/std/file.ts
```diff
@@ -0,0 +1,34 @@
...
+    return InflightClient.forType(__filename, this.name);
+  }
+
+  public static getExtension(fileName: string): string {
+    return pathext.extname(fileName).toLowerCase();
+  }
+
+  public static getFileEncoding(fileName: string): "base64" | "utf8" {
+    const type = this.getExtension(fileName);
+    switch (type) {
```
tsuf239, line 20: I think it's better to let the user specify the encoding of the file, and set the default to a common one (utf-8 for example) as there are probably dozens of "utf-8" compatible files that aren't in this list, as well as cases which the user prefers to go with a completely other encoding
(it's ok to add an extra argument to the `addFile` signature 🙂 ) 
What do you think?

libs/wingsdk/src/target-awscdk/bucket.ts
```diff
@@ -55,6 +56,15 @@ export class Bucket extends cloud.Bucket {
     });
   }
 
+  public addFile(fileName: string, path: string): void {
+    path = isAbsolute(path)
+      ? path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", path);
```
tsuf239, line 62: Nice! 
(but it probably fail if a few files are running on the same process, as it will be overridden, unfortunatly) 
can you please change it to `App.of(scope).entrypointDir`
(like in here: https://github.com/winglang/wing/blob/86779dbddea789f84496437fce03d1fabd739144/libs/wingsdk/src/cloud/website.ts#L62-L65)
Since this argument is more consistent

libs/wingsdk/src/target-sim/bucket.ts
```diff
@@ -41,6 +42,15 @@ export class Bucket extends cloud.Bucket implements ISimulatorResource {
     this.initialObjects[key] = body;
   }
 
+  public addFile(fileName: string, path: string): void {
+    path = isAbsolute(path)
+      ? path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", path);
+    const data = fs.readFileSync(path, File.getFileEncoding(path));
```
tsuf239, line 49: since this code block is the same in all of the target, we can add it to the parent class :) 
here: libs/wingsdk/src/cloud/bucket.ts

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -73,6 +74,15 @@ export class Bucket extends cloud.Bucket {
...
   }
 
+  public addFile(fileName: string, path: string): void {
+    path = isAbsolute(path)
+      ? path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", path);
+    const data = fs.readFileSync(path, File.getFileEncoding(path));
+
+    this.addObject(fileName, data);
+  }
```
tsuf239, line 84: same here^

libs/wingsdk/src/target-tf-azure/bucket.ts
```diff
@@ -97,6 +99,15 @@ export class Bucket extends cloud.Bucket {
...
   }
 
+  public addFile(fileName: string, path: string): void {
+    path = isAbsolute(path)
+      ? path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", path);
+    const data = fs.readFileSync(path, File.getFileEncoding(path));
+
+    this.addObject(fileName, data);
+  }
```
tsuf239, line 109: same here^

libs/wingsdk/src/target-tf-gcp/bucket.ts
```diff
@@ -82,6 +84,15 @@ export class Bucket extends cloud.Bucket {
...
   }
 
+  public addFile(fileName: string, path: string): void {
+    path = isAbsolute(path)
+      ? path
+      : resolve(process.env.WING_SOURCE_DIR ?? "", path);
+    const data = fs.readFileSync(path, File.getFileEncoding(path));
+
+    this.addObject(fileName, data);
+  }
```
tsuf239, line 94: same here^

libs/wingsdk/test/target-awscdk/bucket.test.ts
```diff
@@ -57,6 +57,20 @@ test("bucket with two preflight objects", () => {
...
+  const bucket = Bucket._newBucket(app, "my_bucket", { public: true });
+  bucket.addFile("file1.txt", "test/testFiles/test1.txt");
+  bucket.addFile("file2.txt", "test/testFiles/test2.txt");
+  const output = app.synth();
+
+  // THEN
+  const template = Template.fromJSON(JSON.parse(output));
+  template.resourceCountIs("Custom::CDKBucketDeployment", 2);
+  expect(template.toJSON()).toMatchSnapshot();
+});
```
tsuf239, line 72: Nice!

tsuf239 (changes requested):
Great work! Really awesome 1st PR 🎉 
Can you please add a test or two written in Wing to a new file in this folder:  examples/tests/sdk_tests/bucket
It help us to keep stability between versions :)

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -73,6 +73,12 @@ export abstract class Bucket extends Resource {
    */
   public abstract addObject(key: string, body: string): void;
 
+  /**
+ * Add a file to the bucket from system folder
+ */
+
+  public abstract addFile(fileName: string, path: string): void;
```
Chriscbr, line 80: First parameter should be named `key` to align with the rest of our terminology/docs for `cloud.Bucket`
```suggestion
  public abstract addFile(key: string, path: string): void;
```

libs/wingsdk/src/std/file.ts
0018akhil: Yeah, that would be great. Let me add an extra argument file encoding.

libs/wingsdk/src/target-awscdk/bucket.ts
0018akhil: Sure.

libs/wingsdk/src/cloud/bucket.ts
0018akhil: Okay.

libs/wingsdk/src/target-sim/bucket.ts
```diff
@@ -17,8 +17,10 @@ import { BaseResourceSchema } from "../testing/simulator";
 export class Bucket extends cloud.Bucket implements ISimulatorResource {
   private readonly public: boolean;
   private readonly initialObjects: Record<string, string> = {};
+  private scope: Construct
   constructor(scope: Construct, id: string, props: cloud.BucketProps = {}) {
     super(scope, id, props);
+    this.scope = scope
```
tsuf239, line 23: please remove this unused var :)

tsuf239 (approved):
Awesome ✨ 
Just one comment is left before merging

libs/wingsdk/src/target-sim/bucket.ts
0018akhil: Hi @tsuf239 

I need **scope** variable from the constructor. So I can use it in the **addFile** method.

image

I'm actually **reading the scope value in the addFile method** but I'm still getting this error in codespaces.

tsuf239: You are right! you need the scope in libs/wingsdk/src/**cloud**/bucket.ts I asked you to remove the lines from libs/wingsdk/src/**target-sim**/bucket.ts it's not the same file :)

0018akhil: OH!
Thank you. Got it.
I have removed it.

tsuf239:
pushed a fix for the test paths and missing `entrypointDir`, the tests should pass now :)

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -73,6 +77,28 @@ export abstract class Bucket extends Resource {
    */
   public abstract addObject(key: string, body: string): void;
 
+  /**
+   * Add a file to the bucket from system folder
+   */
+
```
eladb, line 83: extra newline. Add "@param" to describe parameters.

libs/wingsdk/src/target-sim/bucket.ts
```diff
@@ -17,9 +17,9 @@ import { BaseResourceSchema } from "../testing/simulator";
 export class Bucket extends cloud.Bucket implements ISimulatorResource {
   private readonly public: boolean;
   private readonly initialObjects: Record<string, string> = {};
+
```
eladb, line 20: revert

tsuf239:
@MarkMcCulloh do you have any idea why does "@monadabot
[chore: self mutation](https://github.com/winglang/wing/pull/3801/commits/2c16f1d5364fcfd2458f7e96a2b1f43be52909be)" keep adding those license files and changing the package.json?

MarkMcCulloh:
Looks like a change to projen slipped without updating the actual file. I say go ahead and keep whatever the self mutation does, my PR here https://github.com/winglang/wing/pull/3913 fixes things as well.

PR #3807, opened 2023-08-14
fix(sdk): fix website source_hash

tsuf239:
Fixing the source_hash in the sdk snapshots to be a string and not a path

## Checklist

- [ ] Tests added (always)


Chriscbr (approved):
🙏

Chriscbr:
Having to do these sanitizations twice (in hangar and in the SDK) is annoying... I wonder if we can find some way to reuse the code in both place

tsuf239:
> Having to do these sanitizations twice (in hangar and in the SDK) is annoying... I wonder if we can find some way to reuse the code in both places

We can, but it won't be so straightforward- in hangar we manipulate string (to overcome nested fields) and in the SDK we manipulate json, maybe when we'll start snapshotting with our test command- we can export the sanitizing function and use it in the SDK as well

Issue #3493, opened 2023-07-17
No resource found with handle "sim-4"


skorfmann:
### I tried this:

```
bring cloud;
bring ex;
bring util;

let queue = new cloud.Queue();
let redis = new ex.Redis();

queue.setConsumer(inflight (message: str) => {
  redis.set("hello", message);
}, timeout: 3s);

test "Hello, world!" {
  queue.push("world!");

  util.waitUntil((): bool => {
    log("Checking if redis key exists");
    return redis.get("hello") != nil; // works without the `return`
  });

  assert("world!" == "${redis.get("hello")}");
}
```

adding an explicit `return` to the `waitUntil` function which worked in other contexts, like the in the following example.

```
test "Hello, world!" {
  util.waitUntil((): bool => {
    log("Checking if wing.txt exists");
    return bucket.exists("wing.txt");
  });
}
```

### This happened:

```
wing test main.w
/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:359
            throw new Error(`No resource found with handle "${handle}".`);
            ^

Error: No resource found with handle "sim-4".
    at HandleManager.find (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:359:19)
    at Object.findInstance (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/testing/simulator.js:229:38)
    at Queue.processMessages (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/queue.inflight.js:77:47)
    at Timeout._onTimeout (/Users/sebastian/.nvm/versions/node/v18.16.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/target-sim/queue.inflight.js:15:50)
    at listOnTimeout (node:internal/timers:569:17)
    at process.processTimers (node:internal/timers:512:7)
```

### I expected this:

either no error at all, or an error which states what the problem is

### Is there a workaround?

Don't `return` 

### Component

Compiler

### Wing Version

0.24.31

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
[...]

PR #3810, opened 2023-08-14
fix(sdk): changing redis empty value to be nil

tsuf239:
fixes [#3493](https://github.com/winglang/wing/issues/3493)
Changing Redis empty value to be nil- before it was js' null, which doesn't have an equivalent in wing, therefore could not be used 

## Checklist

- [x] Tests added (always)


staycoolcall911 (approved):
Nice! Makes perfect sense

PR #3814, opened 2023-08-14
chore(sdk): refactor connections API

Chriscbr:
Background: in order for Wing classes to support extending non-Wing classes (#3717), I'm trying to remove the `Resource` class in the SDK so that all classes generated by our compiler just need to extend the ol' [`construct.Construct`](https://github.com/aws/constructs/tree/10.x) class -- the "base class" of the JSII ecosystem. This is a tricky effort since `Resource` is used in many places, and it collected a lot of responsibilities over the past year.

This PR takes a first step towards that by removing the connection-related information from the `Resource` class:

https://github.com/winglang/wing/blob/1c9954e50ebdd4f2b290e91665c920fb504c9cb7/libs/wingsdk/src/std/resource.ts#L222-L223

Instead, we now store the information inside the root of the construct tree. It can be accessed through `Connections.of(x)` where `x` is any class instance. See `connections.ts` for the full implementation (its short).

With this refactoring, I took the opportunity to clean up how connections are modeled. Previously, we were storing each connection twice inside `tree.json` (once next to the connection's source, and a second time next to the connection's target). Now, we just store each connection once, in a flat array inside `connections.json`. (I wasn't sure if it makes more sense to put it in a separate file - I'm interested in feedback on this). I also removed the `implicit` field which wasn't used by the SDK or console anywhere.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/apps/wing-console/console/server/src/index.ts
+++ b/apps/wing-console/console/server/src/index.ts
@@ -100,12 +100,18 @@ export const createConsoleServer = async ({
   });
 
   const compiler = createCompiler(wingfile);
+  let isStarting = false;
+  let isStopping = false;
+
   const simulator = createSimulator();
   if (onTrace) {
     simulator.on("trace", onTrace);
   }
   compiler.on("compiled", ({ simfile }) => {
-    simulator.start(simfile);
+    if (!isStarting) {
+      simulator.start(simfile);
+      isStarting = true;
+    }
   });
 
   let lastErrorMessage = "";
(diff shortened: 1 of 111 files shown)


polamoros (approved):

eladb:
I love this direction. See my comment on the `Display` PR. I think we should merge `std.Display` and `std.Connections` into `std.Node`. This is all node-level general-purpose metadata.

PR #3838, opened 2023-08-16
feat(sdk): adding set() method to mutarray

### LINKED ISSUE #3832: array.set(index, value)
### Feature Spec

```js
let x = MutArray<num>[5, 10, 15];
assert(x.at(1) == 10);
x.set(1, 12);
assert(x.at(1) == 12);
```

### Use Cases

Replacing elements of arrays

### Implementation Notes

Setting an index that is out of bounds should throw an error

### Component

SDK

### Community Notes

WeepingClown13:
Adding a mutarray set method to change array values. Now setting values can be done with;
```ts
let x = MutArray<num>[2, 3];
x.set(1, 4); // changes x to [2, 4]
```

closes #3832

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/std/array.ts
```diff
@@ -249,4 +249,19 @@ export class MutArray {
   public pop(): T1 {
     throw new Error("Abstract");
   }
+
+  /**
+   * Sets a new value at the given index of an array
+   *
+   * @macro ((obj, args) => { if (args[0] < 0 || args[0] >= $self$.length) throw new Error("Index out of bounds"); obj[args[0]] = args[1]; })($self$, [$args$])
```
tsuf239, line 256:
```suggestion
   * @macro ((obj, args) => { if (args[0] < 0 || args[0] >= ($self$.length -1)) throw new Error("Index out of bounds"); obj[args[0]] = args[1]; })($self$, [$args$])
```

examples/tests/sdk_tests/std/array.w
```diff
@@ -236,3 +236,13 @@ test "lastIndexOf()" {
...
+
+//-----------------------------------------------------------------------------
+// set()
+
+test "set()" {
+  let mutArr = MutArray<num>[1, 3, 5, 7, 9];
+  assert(mutArr.at(0) == 1);
+  mutArr.set(0, 2);
+  assert(mutArr.at(0) == 2);
+}
```
tsuf239, line 248: can you please add the following cases:
- index smaller than 0
- index larger than array's length -1

libs/wingsdk/src/std/array.ts
WeepingClown13: @tsuf239  Since it is evaluating `greater than or equal`, isn't `$self$.length` what is needed? Since the max index can only be `$self$.length -1`. Or does `$self$.length -1` do something different than what I think it does?

tsuf239: No, you're actually right, resolving this comment

tsuf239 (approved):

PR #3850, opened 2023-08-17
chore(sdk): running the sdk spec tests action before the build

tsuf239:
Running the sdk spec tests action after pushing to main

## Checklist

- [ ] Tests added (always)


.github/workflows/tf-aws-test.yml
```diff
@@ -3,10 +3,16 @@ on:
   release:
     types:
       - published
+  push:
+    branches:
+      - main
```
MarkMcCulloh, line 8: In order for this workflow to block the build, you would need to do one of these things:

1. Move this workflow to be a job of the build.yml workflow, making it run in run similar to hangar
2. Make the build workflow wait for this to succeed, probably by checking the github API or finding an action that handles that

(1) is much easier. The problem is that to run it before the build means you cannot install wing normally, so you'll have to use the same mechanism that hangar uses.

MarkMcCulloh (changes requested):

.github/workflows/tf-aws-test.yml
```diff
@@ -1,12 +1,19 @@
 name: SDK Spec AWS Tests
 on:
+  workflow_call: {}
   release:
     types:
       - published
+  push:
+    branches:
+      - main
   workflow_dispatch: {}
```
MarkMcCulloh, line 10: should only need workflow_call

```suggestion
```

MarkMcCulloh (changes requested):
It's possible that these tests need to instead run inside of hangar so that it can install the dev version of wing, but triggered separately from the existing e2e tests. This would be similar to the benchmark tests.

.github/workflows/tf-aws-test.yml
tsuf239: I'll keep the `release` for now

.github/workflows/tf-aws-test.yml
```diff
@@ -42,15 +49,43 @@ jobs:
...
+
+      ########################################################################
       - name: Install winglang globally
+        if: ${{github.event_name == 'release'}}
         uses: nick-fields/retry@v2
         with:
           max_attempts: 3
           retry_on: error
           timeout_minutes: 5
           command: npm install -g winglang
```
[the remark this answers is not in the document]
MarkMcCulloh: When you call this workflow during the build workflow, `npm install -g winglang` will install the latest published version of wing and run it against the not-yet-published version of the SDK tests. This mismatch will work sometimes, but it won't always work if there is a breaking change.

tsuf239:
Ready :)

.github/workflows/build.yml
```diff
@@ -265,6 +265,20 @@ jobs:
...
+    if: github.event_name == 'push'
+    uses: ./.github/workflows/tf-aws-test.yml
+    secrets: inherit
+
+  file-diff-sdk-spec-test:
+    needs:
+      - build
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/wing-test.yml
+    secrets: inherit
```
MarkMcCulloh, line 280: Both PR and main branch pushes can make use of a diff-based approach. Ideally turbo should be used to get a good diff. This use case is not well documented and I would be happy to meet on it.

.github/workflows/tf-aws-test.yml
```diff
@@ -39,21 +42,43 @@ jobs:
...
+        with:
+          name: dist
+          path: dist
+
+      - name: Install Wing from Dist
+        if: ${{ env.LOCAL_BUILD ==  'true' }}
+        # npm isn't installing everything on the first try, so we need to call it twice...
+        run: |
+          npm install -g ./dist/*-[0-9]*.[0-9]*.[0-9]*.tgz
+          npm install -g ./dist/*-[0-9]*.[0-9]*.[0-9]*.tgz
```
MarkMcCulloh, line 64: We should probably try to figure this out, it's very strange

.github/workflows/wing-test.yml
```diff
@@ -0,0 +1,99 @@
...
+          max_attempts: 3
+          retry_on: error
+          timeout_minutes: 30
+          command: |
+            wing --version
+            results=$(wing test -t ${{ matrix.target }} ${{ needs.setup.outputs.files }})
+            EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+            echo "text<<$EOF" >> $GITHUB_OUTPUT
+            echo "$results" >> $GITHUB_OUTPUT
+            echo "$EOF" >> $GITHUB_OUTPUT
```
MarkMcCulloh, line 87: Would it be easier to add these tests to the hangar project?

.github/workflows/build.yml
```diff
@@ -265,6 +265,20 @@ jobs:
...
+    needs:
+      - build
+    if: github.event_name == 'push'
+    uses: ./.github/workflows/tf-aws-test.yml
+    secrets: inherit
+
+  file-diff-sdk-spec-test:
+    needs:
+      - build
+    if: github.event_name == 'pull_request'
```
MarkMcCulloh, line 278: Is running this in PRs just for testing? I feel like we'll hit rate limits if we run this in every PR

PR #3870, opened 2023-08-17
fix(compiler): comparison between mutable and immutable objects

### LINKED ISSUE #2867: Comparison of Mutable and Immutable types is not possible
### I tried this:

```
test "foo" {
  let actual = MutArray<str>["a"];
  let expected = Array<str>["a"];
  assert(actual == expected);
}
```

See in [playground](https://play.winglang.io/?code=dGVzdCAiZm9vIiB7CiAgbGV0IGFjdHVhbCA9IE11dEFycmF5PHN0cj5bImEiXTsKICBsZXQgZXhwZWN0ZWQgPSBBcnJheTxzdHI+WyJhIl07CiAgYXNzZXJ0KGFjdHVhbCA9PSBleHBlY3RlZCk7Cn0=)

### This happened:

```
Error: Command failed: wing compile ./test.w -t sim
Error: error: Expected type to be "MutArray", but got "Array" instead
  --> ./test.w:4:10
  |
4 |   assert(actual == expected);
  |          ^^^^^^^^^^^^^^^^^^ Expected type to be "MutArray", but got "Array" instead


    at compile (file:///home/httpsplaywinglangio-eqhr-zoxh9o2f/node_modules/winglang/dist/index.js:61995:19)
```

Same applies for `Set`, `MutSet` so assuming it applies for similar types in general.

### I expected this:

It's possible to compile this

### Is there a workaround?
Stringify before testing equality:
```ts
test "foo" {
  let actual = MutArray<str>["a"];
  let expected = Array<str>["a"];
  assert(Json.stringify(actual) == Json.stringify(expected));
} 
```

### Component

Compiler

### Wing Version

0.19.6

### Wing Console Version

_No response_

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

Turns out that while the version with identical types compiles, the assertion is failing

```
test "foo" {
  let actual = Array<str>["a"];
  let expected = Array<str>["a"];
  assert(actual == expected);
}
```

yields 

```
Unknown error: assertion failed: '(actual === expected)'
```

### Community Notes

marciocadev:
Closes #2867 

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -2584,6 +2596,49 @@ impl<'a> TypeChecker<'a> {
...
+		actual_type: TypeRef,
+		expected_type: TypeRef,
+		span: &impl Spanned,
+	) -> TypeRef {
+		if actual_type.is_array_type()
+			&& expected_type.is_array_type()
+			&& actual_type
+				.collection_item_type()
+				.unwrap()
+				.is_same_type_as(&expected_type.collection_item_type().unwrap())
```
MarkMcCulloh, line 2614: The inner types don't need to be the same, you should be able to compare subtypes as well. So you should still use `self.validate_type_in` for the inner types of the actual and expected

libs/wingc/src/type_check.rs
```diff
@@ -2584,6 +2596,49 @@ impl<'a> TypeChecker<'a> {
...
+	///
+	/// Returns the given type on success, otherwise returns the expected type.
+	fn validate_type_binary_operator(
+		&mut self,
+		actual_type: TypeRef,
+		expected_type: TypeRef,
+		span: &impl Spanned,
+	) -> TypeRef {
+		if actual_type.is_array_type()
+			&& expected_type.is_array_type()
```
MarkMcCulloh, line 2610: 1. My hunch is that using pattern matching would be cleaner here. 
2. We don't need to rely on the inner type being the exact same, we should keep using validate_type because we can use `==` with subtypes.
3. If we're validating the inner type, this function should be recursive since you can do `MutArray<num>[Array<num>[1]] == Array<num>[Array<num>[1]]`

So maybe something like this?

```rust
		if let Type::Array(inner_actual) | Type::MutArray(inner_actual) = *actual_type {
			if let Type::Array(inner_expected) | Type::MutArray(inner_expected) = *expected_type {
				return self.validate_type_binary_operator(inner_actual, inner_expected, span);
			}
		}
```

If using this, the new `is_*_type()` functions can be removed since they'll be unused.

libs/wingc/src/type_check.rs
```diff
@@ -2584,6 +2596,49 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	/// Validate that the given type is a subtype (or same) as the expected type, disregarding whether
+	/// the object is mutable or immutable. If not, add an error to the diagnostics.
+	///
+	/// Returns the given type on success, otherwise returns the expected type.
+	fn validate_type_binary_operator(
```
MarkMcCulloh, line 2603: For now I think this behavior is specific to equality, maybe the function name can reflect that just to make sure it's not used elsewhere.

MarkMcCulloh (changes requested):

libs/wingc/src/type_check.rs
MarkMcCulloh: Another option, with less nesting (although requiring an extra dereferencing):

```rust
		if let (
			Type::Array(inner_actual) | Type::MutArray(inner_actual),
			Type::Array(inner_expected) | Type::MutArray(inner_expected),
		) = (&*actual_type, &*expected_type)
		{
			return self.validate_type_binary_operator(*inner_actual, *inner_expected, span);
		}
```

marciocadev:
Thank you, @MarkMcCulloh . I basically just opened the PR, you're the one who did the heavy lifting. 😁

libs/wingc/src/type_check.rs
```diff
@@ -2584,6 +2584,36 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	/// Validate that the given type is a subtype (or same) as the expected type, disregarding whether
+	/// the object is mutable or immutable. If not, add an error to the diagnostics.
+	///
+	/// Returns the given type on success, otherwise returns the expected type.
+	fn validate_equal_types(&mut self, actual_type: TypeRef, expected_type: TypeRef, span: &impl Spanned) -> TypeRef {
```
MarkMcCulloh, line 2591: This function name still feels off, it sounds like it's checking if types are equal. Maybe something like this, and we can make a pattern out of `validate_type_*` like we do with `type_check_*`:

```suggestion
	fn validate_type_binary_equality(&mut self, actual_type: TypeRef, expected_type: TypeRef, span: &impl Spanned) -> TypeRef {
```

idk, I'm also struggling to come up with a name...

MarkMcCulloh (approved):
❤️

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
MarkMcCulloh: If you add handling for this case, make sure to add a test as well

PR #3884, opened 2023-08-18
feat(sdk)!: std.Node

### LINKED ISSUE #3717: Cannot extend non-Wing constructs

Chriscbr:
As part of the effort to fix #3717, I'm trying to refactor some parts of the SDK so that we do not need to rely on the `std.Resource` class.  The issue this PR addresses is that the `Resource` class has an extra property named `display` which is not available if you extend `cdk8s.Chart`:

```js
bring cdk8s;

class Bar {
  init() {
    this.display.title = "hi"; // OK
  }
}

class MyChart extends cdk8s.Chart {
  init() {
    this.display.title = "hello"; // error: Unknown symbol "display"
  }
}
```

The solution implemented by this PR is to standardize display property access through `std.Node`:

```js
bring cdk8s;

class Bar {
  init() {
    std.Node.of(this).title = "hi"; // OK
  }
}

// once #3717 is fixed
class MyChart extends cdk8s.Chart {
  init() {
    std.Node.of(this).title = "hello"; // OK
  }
}
```

`std.Node.of(this)` also allows you toaccess all other fields and methods normally available through the `constructs.Node` class.

Other changes:
- I've replaced the `_addInflightOps` method from `std.Resource` with an API contract where a class can (optionally) implement a method named `_getInflightOps()`. This removes one more of the places where the compiler assumes behavior that only exists on `std.Resource` and not on `constructs.Construct`. 
- I've removed the `Code` class from the SDK. I looked through our codebase and 100% of the usages have just been to access its `text` field, so the class isn't really doing anything important. Just using plain `string`'s seems like a nice simplification.

BREAKING CHANGE: `.display` property is no longer available in Wing classes. To change how classes are displayed in the Wing Console, use `std.Node.of(this)` and modify fields like `.title` and `.description`.

## Checklist

- [ ] Tests added (always)


libs/wingc/src/jsify.rs
```diff
@@ -307,30 +307,6 @@ impl<'a> JSifier<'a> {
...
-	// This helper determines what requirement and dependency to add to the struct schema based
-	// on the type annotation of the field.
-	// I.E. if a struct has a field named "foo" with a type "OtherStruct", then we want to add
-	// the field "foo" as a required  and the struct "OtherStruct" as a dependency. so the result is
-	// a tuple (required, dependency)
-	fn extract_struct_field_schema_dependency(
-		&self,
-		typ: &TypeAnnotationKind,
-		field_name: &String,
-	) -> (Option<String>, Option<String>) {
```
Chriscbr, line 319: I moved this to be a free floating function since it doesn't need/use the `&self` parameter

libs/wingsdk/src/std/display.ts
```diff
@@ -0,0 +1,50 @@
+import { IConstruct } from "constructs";
+
+const DISPLAY_SYMBOL = Symbol.for("@winglang/sdk.std.Display");
```
eladb, line 3: Let's merge the concepts of "display" and "connections". I think they can both be modeled through something like `std.Node`, which we should submit upstream to `constructs.Node`. They are both relevant as general purpose metadata in the construct tree for other CDKs as well.

In the meantime, we can just support this like this:

And then:

```js
std.Node.of(x).title = "hello";
std.Node.of(x).addConnection(target: x, edge: ["put()"]);
std.Node.of(x).hidden = true;
std.Node.of(x).id
std.Node.of(x).path
std.Node.of(x).addr
std.Node.of(x).addDependency(...)
```

for c in std.Node.of(x).connections {
  log(std.Node.of(c).path);
}
```

And ideally, add a keyword: `nodeof(x)` as sugar.

WDYT?

**Review by eladb — CHANGES_REQUESTED** [review]

File: libs/wingsdk/src/std/display.ts (hunk 2 above)
Comment (Chriscbr) [reply libs/wingsdk/src/std/display.ts:3]: Love the idea. I had an inkling it might be possible to combine Connections and Display under some common API but I didn't have a concrete idea at the time. But using the `nodeof` concept (that we've discussed some times in the past) sounds like a great fit. I'll try this out 👍

**Comment by Chriscbr** [comment]
@eladb Updated the API. Leaving `nodeof` for a separate PR since it requires some refactoring (https://github.com/winglang/wing/issues/2027)

File: libs/wingsdk/src/std/node.ts [hunk 3]
```diff
@@ -0,0 +1,349 @@
...
+  // https://github.com/aws/constructs/blob/10.x/src/construct.ts
+
+  /**
+   * Returns the scope in which this construct is defined.
+   *
+   * The value is `undefined` at the root of the construct scope tree.
+   */
+  public get scope(): IConstruct | undefined {
+    return this._constructsNode.scope;
+  }
```
Chriscbr, line 80: Pre-answering a question you might be wondering: why did I copy all of the public APIs from `constructs.Node` instead of just extending the `Node` class? The reason I did this was JSII didn't allow extending `constructs.Node` because the new signature of `std.Node.of` (returns a `std.Node`) didn't match the signature of `constructs.Node.of` (returns a `constructs.Node`). I wasn't sure if there are better alternatives 🤷

eladb: I knew you will need to do that 😛

libs/wingc/src/jsify/snapshots/capture_identifier_from_preflight_scope.snap
```diff
@@ -43,26 +43,28 @@ class $Root extends $stdlib.std.Resource {
     class $Closure1 extends $stdlib.std.Resource {
       constructor(scope, id, ) {
         super(scope, id);
-        this._addInflightOps("handle", "$inflight_init");
```
eladb, line 46: Is this a regression?

eladb (approved):
I am wondering if we need to add a doc topic about nodes? 

Maybe even just an initial thing with a reference to the construct documentation?

libs/wingc/src/jsify/snapshots/capture_identifier_from_preflight_scope.snap
Chriscbr: Fortunately no - there's a new _getInflightOps() method generated at the end of the snapshot instead

PR #3888, opened 2023-08-19
feat(compiler): elif let statement

marciocadev:
## Checklist

- [x] Tests added (always)

### DIFF
--- a/tools/hangar/__snapshots__/test_corpus/valid/optionals.w_compile_tf-aws.md
+++ b/tools/hangar/__snapshots__/test_corpus/valid/optionals.w_compile_tf-aws.md
@@ -515,9 +515,9 @@ class $Root extends $stdlib.std.Resource {
     }
     const b = 1;
     {
-      const $IF_LET_VALUE = b;
-      if ($IF_LET_VALUE != undefined) {
-        const z = $IF_LET_VALUE;
+      const $if_let_value = b;
+      if ($if_let_value != undefined) {
+        const z = $if_let_value;
         {((cond) => {if (!cond) throw new Error("assertion failed: z == 1")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(z,1)))};
       }
     }


libs/wingc/src/jsify.rs
```diff
@@ -896,6 +897,15 @@ impl<'a> JSifier<'a> {
 					if_let_value,
 					self.jsify_expression(value, ctx)
 				));
+				let elif_let_value = "$ELIF_LET_VALUE";
+				for i in 0..elif_statements.len() {
+					let value = format!("{}{}", elif_let_value, i);
+					code.line(format!(
```
yoav-steinberg, line 903: I think it's wrong to evaluate all the values in advance. Consider:
```wing
if let x = tryA() {
} elif let x = tryB() {
} elif let x = TryC() {
}
```
1. There's a performance penalty evaluating `TryA`, `TryB` and `TryC` in advance if `TryA` succeeds.
2. If there are side effects to the evaluation then the use wouldn't expect them to happen if `TryA` succeeds.

libs/wingc/src/type_check.rs
```diff
@@ -3728,6 +3669,81 @@ impl<'a> TypeChecker<'a> {
 		}
 	}
 
+	fn type_check_if_let_statement(
```
yoav-steinberg, line 3672: nice

yoav-steinberg (commented):
nice! added some minor comments, mainly the issue of per-evaluating the elif statements.

examples/tests/valid/optionals.w
```diff
@@ -58,6 +58,17 @@ let tryParseName = (fullName: str): Name? => {
...
+let json_obj = Json { ghost: "spooky" };
+if let y = json_obj.tryAsBool() {
+  log("it's a boolean!");
+} elif let y = json_obj.tryAsNum() {
+  log("it's a number!");
+} elif let y = json_obj.tryAsStr() {
+  log("it's a str!");
+} else {
+  log("it's something else!");
+}
```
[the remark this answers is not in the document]
marciocadev: nice

libs/wingc/src/jsify.rs
```diff
@@ -896,6 +897,15 @@ impl<'a> JSifier<'a> {
 					if_let_value,
 					self.jsify_expression(value, ctx)
 				));
+				let elif_let_value = "$ELIF_LET_VALUE";
```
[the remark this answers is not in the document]
marciocadev: @yoav-steinberg I follow this line

https://github.com/winglang/wing/blob/19ef86cd590e7d531b0c843c08456ec81ac1bc9e/libs/wingc/src/jsify.rs#L893

libs/wingc/src/jsify.rs
```diff
@@ -905,6 +915,19 @@ impl<'a> JSifier<'a> {
...
 
+				for i in 0..elif_statements.len() {
+					let value = format!("{}{}", elif_let_value, i);
+					code.open(format!("else if ({value} != undefined) {{"));
+					let elif_block = elif_statements.get(i).unwrap();
+					if elif_block.reassignable {
+						code.line(format!("let {} = {};", elif_block.var_name, value));
+					} else {
+						code.line(format!("const {} = {};", elif_block.var_name, value));
+					}
```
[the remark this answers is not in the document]
Chriscbr: My reading of the situation is that `let` and `const` sometimes have performance costs relative to `var` in JavaScript (I found some details in this stackoverflow post: https://stackoverflow.com/questions/40070631/v8-javascript-performance-implications-of-const-let-and-var). So I could see the benefits of switching to just code-generating `var` to eek out some more performance. The difference between `const` and `let` is less clear but I'm not too opinionated about it. (Though I don't think using `const` negatively impacts the readability of emitted JS code)

The other distinction with `object.freeze` is that `object.freeze` had a semantic difference, not just a performance difference (in extern JavaScript methods, you would receive frozen objects which couldn't be modified with extra methods etc.)

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
yoav-steinberg: I see, maybe change that as well. Not critical but I think all vars including generated stuff should be lowercase.

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
Chriscbr: Sort of related - apparently esbuild sometimes rewrites JS statements from let / const into var: https://esbuild.github.io/faq/#top-level-var

libs/wingc/src/jsify.rs
marciocadev: From what I understand, this code in Wing:
```ts (wing)
if let x = tryA() {
   ...
} elif let x = tryB() {
   ...
} elif let x = TryC() {
   ...
}
```
will generate this code in TypeScript:
```ts
let $IF_LET_VALUE = tryA();
let $ELIF_LET_VALUE0 = tryB();
let $ELIF_LET_VALUE1 = tryC();
if ($IF_LET_VALUE !== undefined) {
   ...
} else if ($ELIF_LET_VALUE0 !== undefined) {
   ...
} else if ($ELIF_LET_VALUE1 !== undefined) {
   ...
}
```
In order for the evaluation of the `elif` conditions to only occur in the case of the else, you can use the following code:
```ts
let $IF_LET_VALUE = tryA();
if ($IF_LET_VALUE !== undefined) {
   ...
} else {
   let $ELIF_LET_VALUE0 = tryB();
   if ($ELIF_LET_VALUE0 !== undefined) {
      ...
   } else {
      let $ELIF_LET_VALUE1 = tryC();
      if ($ELIF_LET_VALUE1 !== undefined) {
         ...
      }
   }
}
```
This code might be considered less elegant, but it's effective in this case.

libs/tree-sitter-wing/grammar.js
```diff
@@ -280,9 +280,20 @@ module.exports = grammar({
         "=",
         field("value", $.expression),
         field("block", $.block),
+        repeat(field("elif_let_block", $.elif_let_block)),
         optional(seq("else", field("else_block", $.block)))
       ),
 
+    elif_let_block: ($) =>
+      seq(
+        "elif let",
```
MarkMcCulloh, line 289: nit: I'm not 100% sure, but I think tree-sitter handles things like this better when each has its own token

```suggestion
        "elif",
        "let",
```

libs/wingc/src/lsp/hover.rs
```diff
@@ -169,13 +169,19 @@ impl<'a> Visit<'a> for HoverVisitor<'a> {
...
 				self.with_scope(statements, |v| {
 					v.visit_symbol(var_name);
 				});
 				self.visit_expr(value);
 				self.visit_scope(statements);
+				for elif in elif_statements {
+					self.visit_symbol(&elif.var_name);
+					self.visit_expr(&elif.value);
+					self.visit_scope(&elif.statements);
+				}
```
MarkMcCulloh, line 184: Weird thing to handle with `if let`-stuff: Technically the symbol is defined within the inner scope, so we have to let the visitor know

```suggestion
				for elif in elif_statements {
				  self.with_scope(&elif.statements, |v| {
				  	v.visit_symbol(&elif.var_name);
				  });
					self.visit_expr(&elif.value);
					self.visit_scope(&elif.statements);
				}
```

MarkMcCulloh (approved):

libs/tree-sitter-wing/grammar.js
marciocadev: Thank you for pointing that out, @MarkMcCulloh. You're right, if I put 2 spaces between the `elif` and the `let`, an error occurs if I use the string ("elif let"). However, separating the tree-sitter takes care of this type of detail.

It's becoming a habit for you to correct me, keep it up! 😁

PR #3891, opened 2023-08-20
chore(console): console preview environment

eladcon:
Builds the console (`@wingconsole/app`) and deploys it to fly.io on every pr (if its not from a fork). 

## Checklist

- [ ] Tests added (always)


.github/workflows/console-preview.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.action != 'closed' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install flyctl
+        uses: superfly/flyctl-actions/setup-flyctl@master
```
skyrpex, line 36: Maybe we should pin to a specific version, and update manually from time to time?

apps/wing-console/console/app/preview/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+COPY . .
+
+ENV PATH "$PATH:/emsdk:/emsdk/upstream/emscripten:/app/.cargo/wasi-sdk-19.0/bin"
+
+RUN pnpm install --no-frozen-lockfile
+RUN pnpm turbo compile --filter @wingconsole/app --concurrency=1
+RUN pnpm --filter @wingconsole/app deploy --prod pruned
+
+FROM node:20-alpine
```
skyrpex, line 45: I think it would be better if we use node 18, since that's what we require in the monorepo. See the `.nvmrc` file. The specific version is `18.16.0`.

eladb:
I think this should be `chore` and not `feat` because there is no user facing impact here.

.github/workflows/console-preview.yml
```diff
@@ -0,0 +1,68 @@
...
+            popd
+          fi
+          flyctl deploy . --config ./apps/wing-console/console/app/preview/fly.toml --app "$APP_NAME" --image-label latest --vm-memory 512 --strategy immediate
+          flyctl scale count 1 --yes --app "$APP_NAME"
+      - name: Post preview comment
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            Console preview environment is available at https://${{env.APP_NAME}}.fly.dev :rocket:
+          comment_tag: Console preview environment
```
MarkMcCulloh, line 51: You can use the monadabot token here so the comment is a little more friendly. Actually you should probably make a new token that only has permission to post comments.

```suggestion
      - name: Post preview comment
        uses: thollander/actions-comment-pull-request@v2
        with:
          GITHUB_TOKEN: ${{ secrets.PROJEN_GITHUB_TOKEN }}
          message: |
            Console preview environment is available at https://${{env.APP_NAME}}.fly.dev :rocket:
          comment_tag: Console preview environment
```

.github/workflows/console-preview.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install flyctl
+        uses: superfly/flyctl-actions/setup-flyctl@master
+        with:
+          version: ${{env.FLY_VERSION}}
+      - name: Deploy to Fly.io
```
MarkMcCulloh, line 37: Is there a way to better integrate this with our build system and workflow? Seems like the entire build happens in a new container, but really we could be using the build artifact that we are already creating as part of PRs.

Worst case scenario, the build that's happening in the container should at least make use of the turbo remote caching

.github/workflows/build.yml
```diff
@@ -306,6 +306,55 @@ jobs:
...
+    env:
+      FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+      FLY_VERSION: "0.1.80"
+      APP_NAME: wing-console-pr-${{github.event.number}}
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
MarkMcCulloh, line 323: Is full git history needed?

MarkMcCulloh (approved):

.github/workflows/build.yml
```diff
@@ -306,6 +306,55 @@ jobs:
           name: ${{ steps.diff.outputs.diff_name }}
           path: ${{ steps.diff.outputs.diff_name }}
 
+  create-console-preview:
+    name: "Create console preview"
```
MarkMcCulloh, line 310:
```suggestion
  console-preview:
    name: "Console Preview"
```

.github/workflows/build.yml
```diff
@@ -306,6 +306,55 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    env:
+      FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+      FLY_VERSION: "0.1.80"
+      APP_NAME: wing-console-pr-${{github.event.number}}
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
```
MarkMcCulloh, line 318: This should also ignore merge queue PRs like https://github.com/winglang/wing/pull/4003

(should be able to ignore any branches that start with `mergify/merge-queue/`)

polamoros (approved):

.github/workflows/build.yml
```diff
@@ -306,6 +306,55 @@ jobs:
...
+          fetch-depth: 1
+
+      - name: Download Dist Artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: dist
+          path: dist
+
+      - name: Install flyctl
+        uses: superfly/flyctl-actions/setup-flyctl@master
```
MarkMcCulloh, line 332:
```suggestion
        uses: superfly/flyctl-actions/setup-flyctl@1.4
```

ainvoner (approved):

PR #3899, opened 2023-08-20
fix(compiler): capture and lifting cleanups and fixes

yoav-steinberg:
* Types aren't expressions anymore. This reverts a change in #3213. Back then we turned all user defined types into expressions so we can lift/capture them the same way we handle expressions. The downside was that the type system didn't make sense anymore (we added a type variable kind and the symbol env wasn't normalized). A side effect is better error spans on unknown type errors.
* To handle capturing/lifting of Types I created a `Liftable` enum that represents AST elements that can be lifted. A lift token now points to a `Liftable` and not to an expression. Currently a `Liftable` can be either an expression or a `UserDefinedType`.
* Simplified and normalized the `Lifts` data structure stored per class. It now contains:
  * `lifts_qualifications` - Stores information for each class inflight method which preflight code it lifts and for that code which ops are performed on it. It's a map of map: "method" -> "preflight js code" -> "set of ops". This data structure is used for generating the lift bindings.
  * `captures` - Stores capture tokens and for each token the preflight js code used for capture and also whether this is a free capture or an object field. This is used to initialize the tokens in preflight.
  * `token_for_liftable` - Maps capturable AST elements (`Liftable`s) to a capture token. This is used for generating inflight code when we encounter a captured type or expression during jsification.
* Fixed a bug where a type referencing itself but calling an inflight static method wasn't properly lifted. No bindings were generated for it leading to potential permission issues. Added a snapshot test.
* Fixed bug parsing static method calls on json `Json.something()` generated a wrong AST expression where we wrapped a `TypeMember` reference inside a `InstanceMemeber` reference (instead of just being a `TypeMemeber` reference). So the end jsification result was `Json.something().something()`??
* Fixed visitor type annotation's default implementation to call `visit_user_defined_type` for user defined types.

## Checklist
- [x] Tests added (always)


tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -212,12 +212,14 @@ Duration <DURATION>"
 `;
 
 exports[`capture_redefinition.w 1`] = `
-"error: Cannot capture symbol \\"y\\" because it is shadowed by another symbol with the same name
-   --> ../../../examples/tests/invalid/capture_redefinition.w:14:9
-   |
-14 |     log(y);
-   |         ^ Cannot capture symbol \\"y\\" because it is shadowed by another symbol with the same name
+"ERROR: Cannot access 'x' before initialization
```
yoav-steinberg, line 215: Figure out why we don't get these errors anymore and if this is a good thing?

libs/wingc/src/parser.rs
```diff
@@ -1394,37 +1391,34 @@ impl<'s> Parser<'s> {
 		let object_expr = self.get_child_field(nested_node, "object")?;
 
 		if let Some(property) = nested_node.child_by_field_name("property") {
-			let object_expr = if object_expr.kind() == "json_container_type" {
-				Expr::new(
+			if object_expr.kind() == "json_container_type" {
+				Ok(Expr::new(
 					ExprKind::Reference(Reference::TypeMember {
```
yoav-steinberg, line 1396: todo: figure our why the double reference here didn't jsify into "Json.something().something()". Or did it?

libs/wingc/src/jsify.rs
```diff
@@ -1172,13 +1190,14 @@ impl<'a> JSifier<'a> {
 		self.emit_inflight_file(&class, inflight_class_code, ctx);
 
 		// lets write the code for the preflight side of the class
+		// TODO: why would we want to do this for inflight classes?? maybe return here in that case?
 		let mut code = CodeMaker::default();
 
 		// default base class for preflight classes is `core.Resource`
 		let extends = if let Some(parent) = &class.parent {
-			let base = parent.as_type_reference().expect("resolve parent type");
+			//let base = parent.as_type_reference().expect("resolve parent type");
```
eladb, line 1198: Remove comment

libs/wingc/src/jsify.rs
```diff
@@ -1154,6 +1170,8 @@ impl<'a> JSifier<'a> {
 			ctx.lifts
 		};
 
+		//println!("LIFTS: {:#?}", lifts);
```
eladb, line 1173: Remove comment

libs/wingc/src/lifting.rs
```diff
@@ -138,40 +129,56 @@ impl<'a> LiftVisitor<'a> {
 		self.ctx.pop_phase();
 		res
 	}
+
+	fn jsify_udt(&mut self, node: &UserDefinedType) -> String {
```
eladb, line 133: Maybe move this to jsify.rs as the public api

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -5,7 +5,7 @@ exports[`access_hidden_namespace.w 1`] = `
   --> ../../../examples/tests/invalid/access_hidden_namespace.w:7:5
   |
 7 | new core.NodeJsCode(\\"/tmp/test.txt\\"); // This should fail even though \`fs.TextFile\` extends \`core.FileBase\` because we didn't bring in \`core\` explicitly.
-  |     ^^^^^^^^^^^^^^^ Unknown symbol \\"core\\"
+  |     ^^^^ Unknown symbol \\"core\\"
```
eladb, line 8: 👍

tools/hangar/__snapshots__/invalid.ts.snap
eladb: This seems like an unwanted regression

eladb (approved):
Thanks for cleaning up my mess 🙏💓

Same behavior with less code is always better!

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: Fixed by adding a dedicated test for var shadowing in the lifting phase by cleaning up the previous code and moving it to the right place. The new test is more in-line with how JS works so this actually produces the desired result in all three test cases in _capture_redefinition.w_. Added a big comment explaining the logic.

yoav-steinberg: Actually this still isn't perfect, need to better distinguish between _capture_redefinition.w_ and _use_before_defined.w_ tests..

libs/wingc/src/lifting.rs
```diff
@@ -25,69 +30,42 @@ impl<'a> LiftVisitor<'a> {
...
-				}
-				LookupResult::NotFound(_) | LookupResult::ExpectedNamespace(_) => {}
+			if matches!(
+				env.lookup_ext(symbol, Some(self.ctx.current_stmt_idx())),
+				LookupResult::DefinedLater
+			) {
+				report_diagnostic(Diagnostic {
+					span: Some(symbol.span.clone()),
+					message: format!(
+						"Cannot acceess symbol \"{symbol}\" because it is shadowed by another symbol with the same name"
```
MarkMcCulloh, line 57: Typo and imo the first "symbol" is redundant:

```suggestion
						"Cannot access \"{symbol}\" because it is shadowed by another symbol with the same name"
```

libs/wingc/src/jsify.rs
```diff
@@ -1280,10 +1295,9 @@ impl<'a> JSifier<'a> {
 		code.open(format!("require(\"{client_path}\")({{"));
 
 		if let Some(lifts) = &ctx.lifts {
-			for capture in lifts.captures() {
-				let preflight = capture.code.clone();
-				let lift_type = format!("context._lift({})", preflight);
-				code.line(format!("{}: ${{{}}},", capture.token, lift_type));
+			for (token, capture) in lifts.captures.iter().filter(|(_, cap)| !cap.is_field) {
+				let lift_type = format!("context._lift({})", capture.code);
+				code.line(format!("{}: ${{{}}},", token, lift_type));
```
Chriscbr, line 1300: How come fields don't have to be lifted?

libs/wingc/src/lsp/completions.rs
```diff
@@ -309,7 +309,10 @@ pub fn on_completion(params: lsp_types::CompletionParams) -> CompletionResponse
 						},
 						arg_list,
 					)),
-					ExprKind::New(new_expr) => Some((types.get_expr_type(&new_expr.class), &new_expr.arg_list)),
+					ExprKind::New(new_expr) => {
+						// TODO: I think `e` is enough here because a new expression's type is the same as the class type. Is this good enough?
+						Some((types.get_expr_type(&e), &new_expr.arg_list))
```
Chriscbr, line 314: Sounds right to me

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -212,11 +212,25 @@ Duration <DURATION>"
 `;
 
 exports[`capture_redefinition.w 1`] = `
-"error: Cannot capture symbol \\"y\\" because it is shadowed by another symbol with the same name
+"error: Cannot acceess symbol \\"y\\" because it is shadowed by another symbol with the same name
+  --> ../../../examples/tests/invalid/capture_redefinition.w:5:7
+  |
+5 |   log(y);
+  |       ^ Cannot acceess symbol \\"y\\" because it is shadowed by another symbol with the same name
```
Chriscbr, line 219:
```suggestion
  |       ^ Cannot access symbol \\"y\\" because it is shadowed by another symbol with the same name
```

libs/wingc/src/type_check/lifts.rs
```diff
@@ -1,140 +1,93 @@
...
-	/// Map from token to lift
-	lift_by_token: BTreeMap<String, Lift>,
+	/// All the captures. The key is token the value is the preflight code.
+	/// Used for preflight setup of inflight captures.
+	pub captures: BTreeMap<String, Capture>,
 
-	/// Map between expression id and a lift token.
-	token_by_expr_id: BTreeMap<ExprId, String>,
+	/// Map between liftable AST element and a lift token (used for inflight jsification of captures)
+	pub token_for_liftable: HashMap<Liftable, String>,
```
Chriscbr, line 19: Was the choice to use different map data structures intentional? (If so maybe add a comment, I'm not sure if the difference is obvious)

libs/wingc/src/type_check/lifts.rs
```diff
@@ -1,140 +1,93 @@
...
 #[derive(Debug)]
 pub struct Lifts {
-	/// All the lifts. The key is "<method>/<token>"
-	lifts: BTreeMap<String, MethodLift>,
-
-	/// All the captures. The key is token.
-	captures: BTreeMap<String, Capture>,
+	// TODO: make all these private and add accessors+helper logic
+	/// All the lifts. Map from method to a map from inflight code to lift qualifications.
+	pub lifts_qualifications: BTreeMap<String, BTreeMap<String, LiftQualification>>,
```
Chriscbr, line 12: super nitpick / optional
```suggestion
	pub lift_qualifications: BTreeMap<String, BTreeMap<String, LiftQualification>>,
```

Chriscbr (approved):

libs/wingc/src/jsify.rs
yoav-steinberg: This creates the `_toInflightType` which lifts free floating vars (and types) used by the inflight (client) type. 
Fields, on the other hand, are lifted by `_toInfilght` when it calls the constructor of the type returned by `_toInflightType`.
I'm not totally sure this complexity of separating the fields from the free floating vars, and doing so in this way, is really required. But I didn't change the previous logic here (previously the `lifts.captures` map didn't include all captures just non-field ones, I unified it).

libs/wingc/src/type_check/lifts.rs
yoav-steinberg: I'm passing on this because we're inside a `Lifts` (plural) struct so this seems more correct.

libs/wingc/src/type_check/lifts.rs
yoav-steinberg: In general I don't like our use of `BTreeMap` at all because it adds a ordering property that isn't needed and since we usually map from strings and ints we don't encounter an issue since ordering is implemented for them by default. In this case I didn't have or want to implement `Ord` on `Liftable`.
In cases where this affects output order I'm for using an `IndexedMap` which is ordered without the need for `Ord` either. But in this case the order has no implications on the output so I went for the least requirements solution (`HashMap`). I think there's no point in adding a comment here since moving back to a `BTreeMap` will give a complication error and ordering isn't important here.

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: fixed

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: fixed by note checking for capturing a shadowed var for unresolved expressions (unknown symbols). This avoids cascading errors.

libs/wingc/src/parser.rs
yoav-steinberg: the reason is because of another bug in how `reference_to_udt` used to handle a `TypeReference` (it just ignored any property access on a reference containing this reference. Anyway not relevant anymore since we don't have `TypeReference`s anymore.

PR #3929, opened 2023-08-23
chore(repo): add VS Code workspace file

skorfmann:
Make it easier to navigate around the various sub projects in the mono repo by using a workspace configuration file https://code.visualstudio.com/docs/editor/multi-root-workspaces

Not exactly sure if that structure makes sense, but it's a starting point. 

Screenshot 2023-08-23 at 13 51 04

Thoughts?

## Checklist

- [ ] Tests added (always)


wing.code-workspace
```diff
@@ -0,0 +1,56 @@
+{
+  "folders": [
+    {
+      "name": "ROOT",
+      "path": "."
+    },
+    {
+      "name": "Wing",
```
eladb, line 8:
```suggestion
      "name": "CLI",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+      "path": "apps/wing"
+    },
+    {
+      "name": "Wing SDK",
+      "path": "libs/wingsdk"
+    },
+    {
+      "name": "Wing SDK Integration Tests",
```
eladb, line 16:
```suggestion
      "name": "SDK Tests",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+      "path": "examples/tests/sdk_tests"
+    },
+    {
+      "name": "Wing Console",
```
eladb, line 20:
```suggestion
      "name": "Console",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+      "path": "apps/wing-console"
+    },
+    {
+      "name": "Wing VSCode Extension",
```
eladb, line 24:
```suggestion
      "name": "VSCode",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+      "path": "apps/vscode-wing"
+    },
+    {
+      "name": "Wing Docs",
```
eladb, line 28:
```suggestion
      "name": "Docs",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+      "path": "docs/docs"
+    },
+    {
+      "name": "jsii-docgen",
```
eladb, line 32:
```suggestion
      "name": "API Docs Generator",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+      "path": "apps/jsii-docgen"
+    },
+    {
+      "name": "wing-api-checker",
```
eladb, line 36:
```suggestion
      "name": "API Checker",
```

wing.code-workspace
```diff
@@ -0,0 +1,56 @@
...
+    {
+      "name": "Tests: Invalid",
+      "path": "examples/tests/invalid"
+    },
+    {
+      "name": "Tests: Error",
+      "path": "examples/tests/error"
+    },
+    {
+      "name": "Proposals",
```
eladb, line 52:
```suggestion
      "name": "Examples",
```

eladb (approved):
A few tweaks

wing.code-workspace
MarkMcCulloh, line 1: As much as I like multi-root workspaces, the downside is that they tend to become unmaintained. It'd be nice if we could automate a process to ensure all projects in the pnpm workspace are present here.

eladb: Can we generate this from the pnpm definition?

MarkMcCulloh: Not directly from the pnpm-workspace.yaml since it uses globs, but it's really easy to programmatically query pnpm ([example we are already using](https://github.com/winglang/wing/blob/main/tools/bump-pack/src/bundle-link.ts#L115-L123))

skorfmann: that's a great idea - will implement that

skorfmann: see latest version

package.json
```diff
@@ -17,7 +18,7 @@
     "docs": "./scripts/docsite.sh",
     "changelog": "bump-pack",
     "install": "bash scripts/setup_wasi.sh",
-    "postinstall": "link-bundles",
+    "postinstall": "link-bundles && generate-workspace",
```
skorfmann, line 21: not sure if that's the best hook, other ideas?

docs/docs/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "docs",
```
skorfmann, line 2: alternative to adding this would be hardcoding a few entries. thoughts?

tools/generate-workspace/src/cli.ts
```diff
@@ -0,0 +1,118 @@
...
+const nameMapping: Record<string, string> = {
+  "@winglang/monorepo": "ROOT",
+  "winglang": "CLI",
+  "@winglang/compiler": "Compiler",
+  "@winglang/sdk": "SDK",
+  "examples-sdk": "SDK Tests",
+  "docs": "Docs",
+  "@wingconsole/app": "Console App",
+  "@wingconsole/server": "Console Server",
+  "@wingconsole/ui": "Console UI",
```
skorfmann, line 31: not sure if the `console` entries should be really a single one

skorfmann:
maybe the mapping and ignore list should be configurable, so everyone can define their own git ignored version of it or fallback to a default. And then perhaps publish to NPM, since this could be useful beyond wing 🤷

Issue #2171, opened 2023-04-20
publish redis as a third party library


ekeren:
I keep trying to write the following code:
```ts (wing)
bring redis;

let redis = new redis.Redis(); // PROBLEM
```

But the above code doesn't compile because the `redis` keyword is already taken 

One option was to replace it with experimental, but this word feels too long.

Chriscbr:
Maybe you could write
```js
bring redis as experimental;
```

If I try doing this today, I get a bug though:
```js
error: Unknown symbol "Redis" in namespace "redis"
  --> hello.w:4:26
  |
4 | let r = new experimental.Redis();
  |                          ^^^^^ Unknown symbol "Redis" in namespace "redis"
```

subh-cs:
Hi @Chriscbr @staycoolcall911, I would like to work on this. Can you assign this to me?

staycoolcall911:
Hey @subh-cs thank you! However this task is dependent on Chris's current task #130, so let's wait for after it is done.

subh-cs:
Hi @staycoolcall911, Can I start working on this? as it seems that #130 has been completed successfully.

staycoolcall911:
Yes, I think you can start. Assigned to you.
Note that @eladb is working on publishing a library for #4350 (Wing Containers Library) - you can see his work on [this repo](https://github.com/eladb/wing-containers) and also a captain's log on [Slack](https://winglang.slack.com/archives/C047QFSUL5R/p1696868156845019).

subh-cs:
Thanks for the reference.

Chriscbr:
`@winglibs/redis` is now published on npm - great job @subh-cs! 🎉

subh-cs:
Hi @Chriscbr , please don't close this issue because `@winglibs/redis` doesn't have tf-aws support but `ex.Redis()` has that.

staycoolcall911:
Thanks for keeping it organized @subh-cs and @Chriscbr 🙏

PR #3938, opened 2023-08-23
feat(cli): `wing pack`

Chriscbr:
Introduces the `wing pack` command - a standard way to package libraries in Wing.

`wing pack` can be run in any Wing project directory that has a `package.json` in it. `wing pack` will validate your `package.json` produce a tarball that contains the libraries source files. For more documentation, see https://winglang.io/docs/libraries.

## Implementation

When you `bring "foo"`, the parser now performs some Node module resolution to look up the `foo` package and see if it's a JSII library or a Wing library based on its `package.json`. If `bring "foo"` refers to a Wing library, then it's equivalent to bringing that library's root directory, per https://github.com/winglang/wing/pull/4210.

## Future related work
- https://github.com/winglang/wing/issues/4415
- https://github.com/winglang/wing/issues/4294
- https://github.com/winglang/wing/issues/2171

## Checklist

- [x] Tests added (always)


apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,83 @@
...
+      throw new Error(`Missing required field "${field}" in package.json`);
+    }
+  }
+
+  // check if "main" points to a valid file, and if not, create a dummy file
+  let main = pkgJson.main ?? "index.js";
+  if (!fs.existsSync(main)) {
+    const lines = [];
+    lines.push("// This file was generated by wing pack");
+    lines.push(`console.log("${pkgJson.name} isn't ready for JS consumption yet!")`);
```
Chriscbr, line 67: Is this sort of what you were suggesting @eladb on the Wingly?

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,83 @@
...
+import * as cp from "child_process";
+
+export interface PackageOptions {
+  /**
+   * Directory to to save the generated package to.
+   */
+  readonly outdir: string;
+}
+
+export async function pack(options: PackageOptions): Promise<string> {
```
Chriscbr, line 12: Write unit tests

libs/wing-fixture/dist/wing-fixture-0.0.0.tgz
Chriscbr, line 1: gitignore

libs/wingc/src/ast.rs
```diff
@@ -423,6 +424,8 @@ pub struct Interface {
 #[derive(Debug)]
 pub enum BringSource {
 	BuiltinModule(Symbol),
+	// we can remove this once Wing modules are compiled to Jsii modules :)
+	WingModule { name: Symbol, root_file: PathBuf },
```
Chriscbr, line 428: Is this path relative or absolute? Specify

examples/tests/valid/bring_wing_library.w
Chriscbr, line 1: Battle test this a bit more... I should give the fixture library a submodule and try importing that as well

libs/wing-fixture/package.json
Chriscbr, line 1: Specify this package as private / un-publishable

libs/wingc/src/parser.rs
```diff
@@ -731,12 +731,57 @@ impl<'s> Parser<'s> {
...
 		}
 
-		if module_name.name.starts_with("\"") && module_name.name.ends_with("\"") {
+		if module_symbol.name.starts_with("\"") && module_symbol.name.ends_with("\"") {
+			// we need to inspect the npm dependency to figure out if it's a JSII library or a Wing library
+			// first, find where the package.json is located
+			let module_name = module_symbol.name[1..module_symbol.name.len() - 1].to_string();
+			let source_dir = Path::new(&self.source_name).parent().unwrap().to_str().unwrap();
+			let module_dir =
+				wingii::util::package_json::find_dependency_directory(&module_name, &source_dir).ok_or_else(|| {
```
Chriscbr, line 740: Change `find_dependency_directory` to accept a Path instead of a String perhaps?

libs/wingc/src/lsp/hover.rs
```diff
@@ -648,13 +648,13 @@ bring cloud;
...
-                        //^
-"#
-	);
+	// 	test_hover_list!(
+	// 		test_bring_library,
+	// 		r#"
+	// bring "@winglang/sdk" as bar;
+	//                         //^
+	// "#
+	// 	);
```
Chriscbr, line 657: I noticed this test was breaking because the LSP hover unit tests compile files in temporary directories, but the parser now does some surface-level package lookups in order to determine if something's a JSII library or a Wing library. Trying to bring `@winglang/sdk` doesn't represent a real use case / I think would work in an E2E test, so I'm thinking of dropping this

libs/wing-fixture/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/winglang/wing.git"
+  },
+  "scripts": {
+    "package": "npm pack"
+  },
+  "wing": {
+    "entrypoint": "lib.w"
+  },
```
Chriscbr, line 30: @eladb besides simplifying the implementation, my initial feeling is that having a single file represent the "entrypoint" of a module could be a good practice since adding or removing exports is in a single place. (versus, if we inferred that all files at the root of your project are public modules, then you would have to move a Wing file into a folder in order to hide it / make it private). This sort of aligns with the some parts of the language's philosophy around "private by default".

WDYT / how are you thinking about the DX tradeoffs? (I'm not totally sold on either direction personally)

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -88,18 +88,18 @@ Duration <DURATION>"
 `;
 
 exports[`bring_jsii.w 1`] = `
-"error: bring \\"jsii-code-samples\\" must be assigned to an identifier (e.g. bring \\"foo\\" as foo)
+"error: bring jsii-code-samples must be assigned to an identifier (e.g. bring \\"foo\\" as foo)
```
Chriscbr, line 91: regression

libs/wing-fixture/package.json
MarkMcCulloh, line 1: It feels off to me to have this project in `libs`, seems like it belongs somewhere in `examples`

libs/wing-fixture/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "type": "git",
+    "url": "git+https://github.com/winglang/wing.git"
+  },
+  "scripts": {
+    "package": "npm pack"
+  },
+  "private": true,
+  "wing": {
+    "entrypoint": "lib.w"
+  },
```
MarkMcCulloh, line 31: imo it might be nice to initially avoid any configuration for this, I think this default is reasonable.
The convention of `lib.w is the entrypoint of a library` should buy us time to better spec out the kind of options we want to be configurable.

eladb: I also don't think we really need an entrypoint for libraries (chatted with @Chriscbr about this today), but I think we at least need a way to know that this is a wing library from reading `package.json`.

But it could be `wing: true`.

apps/wing/src/cli.ts
```diff
@@ -156,6 +156,13 @@ async function main() {
     .hook("preAction", collectAnalyticsHook)
     .action(runSubCommand("test"));
 
+  program
+    .command("pack")
+    .description("Package a Wing module as a shareable library")
```
eladb, line 161:
```suggestion
    .description("Package a Wing module as a shareable npm library")
```

apps/wing/src/cli.ts
```diff
@@ -156,6 +156,13 @@ async function main() {
...
+    .addOption(new Option("-o --outdir <outdir>", "Output directory").default("dist/"))
```
eladb, line 162:
```suggestion
    .addOption(new Option("-o --outdir <outdir>", "Output directory").default("."))
```

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // check package.json exists
+  const currentDir = process.cwd();
+  const pkgJsonPath = path.join(currentDir, "package.json");
+  if (!fs.existsSync(pkgJsonPath)) {
+    throw new Error(`No package.json found in the current directory. Run \`npm init\` first.`);
+  }
+
+  const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, "utf8"));
+  pkgJson.wing = pkgJson.wing ?? {};
```
eladb, line 28:
```suggestion
  pkgJson.wing = pkgJson.wing ?? true;
```

apps/wing/src/commands/pack.ts
eladb, line 67:
```suggestion
    lines.push(`console.log("${pkgJson.name} is a winglang library and cannot be used from javascript (yet)")`);
```

libs/wing-fixture/package.json
eladb: +1

libs/wing-fixture/package.json
eladb: Once we have access modifiers, only types that are marked `pub` will be part of the library's public API. The fact that something is under a subdirectory does not imply that it's private.

libs/wingc/src/ast.rs
```diff
@@ -423,6 +424,12 @@ pub struct Interface {
 #[derive(Debug)]
 pub enum BringSource {
 	BuiltinModule(Symbol),
+	/// A Wing library (containing Wing source files).
+	/// root_file is a relative path to the root file of the library (relative to the compilation entrypoint)
+	WingModule {
```
eladb, line 429: Shouldn't this be `WingLibrary`?

libs/wingc/src/ast.rs
```diff
@@ -423,6 +424,12 @@ pub struct Interface {
...
+		name: Symbol,
+		root_file: PathBuf,
+	},
 	JsiiModule(Symbol),
 	WingFile(Symbol),
```
eladb, line 434: I believe this should have been `WingModule`, no?

libs/wingc/src/parser.rs
```diff
@@ -731,28 +731,73 @@ impl<'s> Parser<'s> {
...
+								span: module_symbol.span,
+							},
+							root_file: entrypoint_path,
+						},
+						identifier: Some(alias),
+					})
+				} else {
+					self.with_error::<StmtKind>(
+						format!(
+							"bring {} must be assigned to an identifier (e.g. bring \"foo\" as foo)",
```
eladb, line 771:
```suggestion
							"bring \"{}\" must be assigned to an identifier (e.g. bring \"foo\" as foo)",
```

libs/wingc/src/ast.rs
Chriscbr: Good call, I'll rename 👍

eladb:
A quick summary of our conversation about modules, libraries, directories and everything in between:

Basically, a *module* is any directory that has a bunch of `.w` files in it (excluding the `.main.w` files). All the declarations marked `pub` within this directory are exported from this module. If two (sibling) files in the same module declare the same symbol, it's a compilation error.

When we `bring "./my_module" as my_module`, then the `my_module` namespace will include the collection of all the public declarations from all the files in the module, as well as submodules (as sub-namespaces).

If a module doesn't have any `pub` declarations, it is considered private (same as directories in git).

A *library* is a module packaged into an npm package.

All the `pub` declarations within the `.w` files at the root directory of the library (next to `package.json`) will be at the root of the library, as well as any submodules.

apps/wing/src/commands/pack.ts
MarkMcCulloh: @eladb Any reason not to just say "wing" here? Seems like an unambiguous situation

```suggestion
    lines.push(`console.log("${pkgJson.name} is a wing library and cannot be used from javascript (yet)")`);
```

eladb: If I see this error I likely don't know what wing is... saying winglang will prompt me to look it up and find a whole new world of goodness joy love and creativity.

libs/wingc/src/parser.rs
```diff
@@ -731,28 +736,87 @@ impl<'s> Parser<'s> {
...
+					}
+
+					// generate a file name for their concatenation
+					let fake_path = PathBuf::from(module_dir).join("$root.w");
+
+					// track that the current file depends on the concatenated file, so it will get parsed downstream
+					self
+						.referenced_wing_files
+						.borrow_mut()
+						.push(WingSourceFile(fake_path.clone(), source));
```
Chriscbr, line 775: So, this file concatenation strategy does "work" (air quotes), but it's not a great solution since we're fudging up the line numbers from original source files, which means the LSP and error messages could get filled with incorrect information. 😕 

Need to find a better solution than this...

Chriscbr:
Keep

eladb:
Really looking forward for this!

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  // check if "main" points to a valid file, and if not, create a dummy file
+  let main = pkgJson.main ?? "index.js";
+  if (!fs.existsSync(main)) {
+    const lines = [];
+    lines.push("// This file was generated by wing pack");
+    lines.push(
+      `console.log("${pkgJson.name} is a winglang library and cannot be used from javascript (yet)")`
```
eladb, line 60:
```suggestion
      `console.log("${pkgJson.name} is a winglang library and cannot be used from javascript (yet). See winglang.io/docs/libraries")`
```

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+    lines.push();
+    fs.writeFileSync(main, lines.join("\n"));
```
eladb, line 63: Not sure I like the idea of writing index.js and package.json into the project directory.

Perhaps we should create a temporary staging directory? This will give us more control over exactly what goes into the package.

(With the price of having to be more explicit about the anatomy of a wing npm package, which is good in my opinion)

examples/wing-fixture/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "organization": true
+  },
+  "files": [
+    "**/*.js",
+    "**/*.w",
+    "!/target/**"
+  ],
+  "license": "MIT",
+  "engines": {
+    "node": ">=v18.0.0"
```
eladb, line 18: We should embed the wing version here

examples/wing-fixture/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/winglang/wing.git"
+  },
+  "scripts": {
+    "package": "npm pack"
```
eladb, line 26:
```suggestion
    "package": "wing pack"
```

examples/wing-fixture/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "private": true,
+  "wing": {
+    "entrypoint": "lib.w"
```
eladb, line 30: Not needed

examples/wing-fixture/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "volta": {
+    "extends": "../../package.json"
```
eladb, line 33: Is this needed?

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export interface PackageOptions {
+  /**
+   * Directory to to save the generated package to.
+   */
+  readonly outdir: string;
+}
+
+export async function pack(options: PackageOptions): Promise<string> {
+  // check npm is installed
```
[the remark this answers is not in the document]
Chriscbr: @eladb see https://github.com/winglang/wing/pull/4342

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,77 @@
...
+  try {
+    cp.execSync("npm --version", { stdio: "ignore" });
+  } catch {
+    throw new Error(`npm is not installed. Install it from https://www.npmjs.com/get-npm`);
+  }
+
+  // check there is at least one source file
+  const sourceFiles = fs.readdirSync(process.cwd()).filter((f) => f.endsWith(".w"));
+  if (sourceFiles.length === 0) {
```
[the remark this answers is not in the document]
Chriscbr: As a library author I imagine it would be pretty common to want to have some entrypoint files (maybe for testing) alongside my other files. Types from entrypoint files would not be exported by the library.

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!fs.existsSync(pkgJsonPath)) {
+    throw new Error(`No package.json found in the current directory. Run \`npm init\` first.`);
+  }
+
+  const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, "utf8"));
+  pkgJson.wing = pkgJson.wing ?? true;
+
+  // check source files will be included in tarball
+  const pkgJsonFiles = pkgJson.files ?? [];
+  const expectedGlobs = ["**/*.js", "**/*.w", "!/target/**"];
```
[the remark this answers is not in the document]
Chriscbr: We don't support `extern "./util.ts"`, do we?

apps/wing/src/commands/pack.ts
Chriscbr: I feel like there are two high level approaches we can take for the MVP:

1. Have a dedicated Wing configuration file, like `wing.toml`. This is where you can specify the library name, author, dependencies, license, description, etc. There's no `package.json` or `index.js` file in the project directory. Clean. 
2. Use the existing JS ecosystem - every Wing project has a `package.json` file. We're using the npm ecosystem to manage dependencies, so folks can control this file and reuse `package.json` snippets from their previous JS experiences, like adding custom scripts etc (this could be a good DX for folks coming from JS/TS, worse DX for folks not coming from JS/TS). Running `wing pack` performs a few validations on `package.json`, telling you if anything is configured wrong -- and it might "own" one or two fields, like `version`, but besides that, what the user wrote in `package.json` is what goes in the tarball. No temporary staging directory.

Do you have an opinion on what direction we should follow? (If not I'm happy to continue down option (2) since I wasn't sure if you had warmed up to (1))

Chriscbr: I'm not sure I see the value in a hybrid approach where there's a package.json in the project directory, but some subset of the fields are copied into a new `package.json` inside of a staging directory. It seems a little surprising (users have to look into docs if they're trying to figure out why a field wasn't included in their tarball). I'm also reminded of how users couldn't customize `tsconfig.json` in JSII, which was a large pain point personally :-/

examples/wing-fixture/package.json
Chriscbr: I think it's just a thing for our monorepo

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,77 @@
...
+  // check there is at least one source file
+  const sourceFiles = fs.readdirSync(process.cwd()).filter((f) => f.endsWith(".w"));
+  if (sourceFiles.length === 0) {
+    throw new Error(`No Wing source files (.w) found in the current directory.`);
+  }
+
+  // check package.json exists
+  const currentDir = process.cwd();
+  const pkgJsonPath = path.join(currentDir, "package.json");
+  if (!fs.existsSync(pkgJsonPath)) {
```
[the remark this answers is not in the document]
Chriscbr: Updated to async

apps/wing/src/commands/pack.ts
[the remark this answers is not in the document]
eladb: Sounds good

apps/wing/src/commands/pack.ts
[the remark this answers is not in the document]
eladb: I heard rumors it works, but not sure. If we don't, then we shouldn't include them of course.

apps/wing/src/commands/pack.ts
eladb: I am not suggesting to create a new package.json and only copy certain fields, just to clone the project directory into a staging area and make modifications there instead of the user code.

It will give us more control over the package structure without polluting the user project, but the DX is will going to be pretty standard package.json.

I am less comfortable with modifying the user directory because it takes us into Projen-land and eventually will require things like self mutations, read only files, and maybe we should try to avoid this (for now).

Once we would want to add jsii manifests, api docs and JavaScript files, there will be more stuff we will want to include in the tarball that we shouldn't just spit into the user directory.

Stuff like this...

If we do `wing.toml` (which we will likely add at some point), we will also have to pack from a staging directory anyway, so staging is basically a step into the right direction regardless of the manifest format I think.

Chriscbr: > just to clone the project directory into a staging area and make modifications there instead of the user code.

> If we do wing.toml (which we will likely add at some point), we will also have to pack from a staging directory anyway, so staging is basically a step into the right direction regardless of the manifest format I think.

I guess I just wonder if there are any ways we can avoid cloning the user's entire directory. I worry it may be pretty heavy handed as a solution. Sure, we can use excluding heuristics like "ignore node modules" and "ignore .git", but in general there can always be extra stuff in the user's current directory (files for CI, editor configs, helper scripts, symlinks, etc.).

The safer approach might be to just copy files that our compiler is actually seeing as part of the compilation (all of the .w and referenced .js files, plus package.json for now), but I guess we can refine upon this later.

examples/wing-fixture/package.json
Chriscbr: I've updated `wing pack` so that a `wing` field is added in the the staging directory / in the final tarball

libs/wingc/src/parser.rs
```diff
@@ -885,10 +885,55 @@ impl<'s> Parser<'s> {
...
+							format!(
+								"Unable to load \"{}\": Module not found in \"{}\"",
+								module_name, self.source_name
+							),
+							&statement_node,
+						)
+						.err();
+				})?;
+
+			// If the package.json has a `wing` field, then we treat it as a Wing library
```
Chriscbr, line 905: Note. When we were modifying the user's `package.json` in-place (in a previous implementation), users wouldn't really be required to run `wing pack` to test a new version of the library locally -- they could just have package.json point to their local Wing library directory.

But since we're no longer modifying `package.json`, the package.json may be missing the "wing" flag or its "main" field may be invalid, and that strategy doesn't work anymore. I've cheated in our monorepo by adding `"wing": true` to the package.json and adding a stub `index.js` file.

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    // add "wing" top-level field
+    pkgJson.wing = true;
+
+    // write package.json
+    await fs.writeFile(pkgJsonPath, JSON.stringify(pkgJson, null, 2) + "\n");
+
+    // make tarball in the specified outdir
+    await fs.mkdir(outdir, { recursive: true });
+    const command = `npm pack --json --pack-destination "${outdir}"`;
```
MarkMcCulloh, line 102: Any consideration towards doing this programmatically? https://github.com/npm/npm-packlist is what npm and pnpm (and probably yarn but idk) use for packing.

It keeps us compliant with expected npm tarball behavior while not relying on the user's system configuration.

apps/wing/src/cli.ts
```diff
@@ -163,6 +163,13 @@ async function main() {
     .hook("preAction", collectAnalyticsHook)
     .action(runSubCommand("test"));
 
+  program
+    .command("pack")
+    .description("Package a Wing module as a shareable npm library")
+    .addOption(new Option("-o --outdir <outdir>", "Output directory").default("."))
```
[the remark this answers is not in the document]
Chriscbr: Gotcha, yeah that makes sense to me. I just copied this from `npm pack` and didn't think much about it. Specifying the path sounds ok to me

apps/wing/src/cli.ts
```diff
@@ -163,6 +163,13 @@ async function main() {
     .hook("preAction", collectAnalyticsHook)
     .action(runSubCommand("test"));
 
+  program
+    .command("pack")
+    .description("Package a Wing module as a shareable npm library")
```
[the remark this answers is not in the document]
Chriscbr: Agree with both

apps/wing/src/commands/pack.ts
Chriscbr: I'll try it out

eladb (approved):
Shippppppit

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,165 @@
...
+   */
+  readonly outfile?: string;
+}
+
+export async function pack(options: PackageOptions = {}): Promise<string> {
+  // check that the library compiles to the "sim" target
+  // TODO: let the user specify all supported targets in package.json
+  console.log('Compiling to the "sim" target...');
+
+  // ugly hack
```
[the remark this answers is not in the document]
Chriscbr: Updated the comment - tl;dr there's a bug so I couldn't run `wing compile .`

apps/wing/src/commands/pack.ts
Chriscbr: Added!

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,165 @@
...
+  // check npm is installed
+  await checkNpmVersion();
+
+  const userDir = process.cwd();
+  const outfile = options.outfile ? resolve(options.outfile) : undefined;
+  const outdir = outfile ? path.dirname(outfile) : userDir;
+
+  // perform our work in a staging directory to avoid making a mess in the user's current directory
+  return withTempDir(async (workdir) => {
+    const excludePaths = [
```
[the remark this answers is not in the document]
Chriscbr: Technically your example would also include .w and .js files from  `node_modules`. But I guess we can have both include and exclude lists. Updated

libs/wingc/src/parser.rs
eladb: Wow great point. I think we definitely want to support monorepos. Let's open an issue and discuss there.

apps/wing/src/commands/pack.ts
eladb: Nice!

apps/wing/src/commands/pack.ts
```diff
@@ -0,0 +1,196 @@
...
+    // copy the user's directory to the staging directory
+    await copyDir(userDir, workdir, excludeGlobs, includeGlobs);
+
+    // check package.json exists
+    const pkgJsonPath = path.join(workdir, "package.json");
+    if (!(await exists(pkgJsonPath))) {
+      throw new Error(`No package.json found in the current directory. Run \`npm init\` first.`);
+    }
+
+    const pkgJson = JSON.parse(await fs.readFile(pkgJsonPath, "utf8"));
```
eladb, line 66: Why read the file again? Isn't this the same as the originalJsonFile above?

Chriscbr: Oops, you're right

Issue #3792, opened 2023-08-11
Json: schema validation support fromJson for structs in namespaces


hasanaburayyan:
with the initial support for `fromJson` methods on structs there is an issue when trying to call `fromJson` on structs that are defined within names spaces and in separate wing files

Need to support something like:
```js
bring cloud;

let j = { public: false };
let x = cloud.BucketProps.fromJson(j);
```

Also multiple files:

example:

file_1.w
```js
struct Person {
	name: str;
}
```

file_2.w
```js
bring "./file1.w" as f1;

let j = { name: "cool" };
f1.Person.fromJson(j);
```

staycoolcall911:
Upgraded to p1 based on feedback from @hasanaburayyan

monadabot:
Congrats! :rocket: This was released in [Wing 0.25.45](https://github.com/winglang/wing/releases/tag/v0.25.45).

Issue #3943, opened 2023-08-23
Json: schema validation support fromJson for jsii imported structs


hasanaburayyan:
Currently we can only call fromJson() on user defined structs, this is because we jsify user defined structs into Json schemas. However, we also treat interfaces imported through jsii as struct types, and since these are just imported from the stdlib we do not create jsified schemas.

Not super sure what the best approach to support this would be just yet.

support something like:
```js
bring cloud;

let j = { public: false };
let x = cloud.BucketProps.fromJson(j);
```

(broke this work out of https://github.com/winglang/wing/issues/3792 since its a separate issue)

hasanaburayyan:
@staycoolcall911 I extracted this from 3792 since it orignally covered this and imported structs from other wing files. Not sure if this is as high priority as the work in [3792](https://github.com/winglang/wing/issues/3792)

staycoolcall911:
Thanks @hasanaburayyan - I think it is high priority. I'll set it as p1, unless someone else has any interesting input on why it could/should be added later. @yoav-steinberg and @Chriscbr might be able to assist you with ideas how to impl

Chriscbr:
I can try to suggest an implementation -- though it's just one approach, so feel free to take with a grain of salt. I think the solution can be split into two parts: type checking and jsification.

For type checking, we'd need to make sure `fromJson` and `schema` is modeled in the type system as available methods on every json-safe struct (regardless of where it came from). This might already work - I haven't checked.

For jsification, we could try changing how we're emitting struct files into a lazy process, so instead of eagerly emitting a `.struct.js` file for each `Struct` statement, we only emit a `.struct.js` file the first time we encounter a `Reference` expression that refers to a given struct (or a `Call` expression to `fromJson` / `schema`).

hasanaburayyan:
@Chriscbr thanks for the feedback,

> For type checking, we'd need to make sure fromJson and schema is modeled in the type system as available methods on every json-safe struct (regardless of where it came from). This might already work - I haven't checked.

Yes this already works.

As for emitting a struct definition when we encounter a reference expression, let me look into that. It sounds logical that I should just be able to do a lookup on the struct and get everything I need for emitting the schema, though I guess that also means I need to jsify the reference to reference the schema file definition rather than the actual imported reference (maybe).

PR #3944, opened 2023-08-23
fix(compiler): cant call fromJson on imported structs

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/3792

Now supports calling fromJson on structs that were brought from other wing files

example: 
file_1.w
```js
struct Person {
	name: str;
}
```

file_2.w
```js
bring "./file1.w" as f1;

let j = { name: "cool" };
f1.Person.fromJson(j);
```

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/valid/struct_from_json.w
+++ b/examples/tests/valid/struct_from_json.w
@@ -237,3 +237,15 @@ test "lifting a student" {
   assert(studentInflight1.dob.day == 10);
   assert(studentInflight1.dob.year == 2005);
 }
+
+// bring structs from other files
+bring "./subdir/structs.w" as externalStructs;
+
+let jj1 = {
+  data: {
+    val: 10
+  }
+};
+
+let externalBar = externalStructs.MyOtherStruct.fromJson(jj1);
+assert(externalBar.data.val == 10);
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -4593,8 +4593,6 @@ impl<'a> TypeChecker<'a> {
 								}
 							}
 						}
-						let lookup = env.lookup(&s.name, None);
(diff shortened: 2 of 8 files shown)


examples/tests/valid/struct_from_json.w
```diff
@@ -236,4 +236,16 @@ test "lifting a student" {
   assert(studentInflight1.dob.month == 10);
   assert(studentInflight1.dob.day == 10);
   assert(studentInflight1.dob.year == 2005);
-}
\ No newline at end of file
+}
+
+// bring structs from other files
+bring "./subdir/structs.w" as externalStructs;
```
eladb, line 242: Does this also work for structs imported via JSII? Maybe add a test..,

libs/wingc/src/type_check.rs
```diff
@@ -4593,8 +4593,6 @@ impl<'a> TypeChecker<'a> {
 								}
 							}
 						}
-						let lookup = env.lookup(&s.name, None);
-						let type_ = lookup.unwrap().as_type().unwrap();
```
MarkMcCulloh, line 4597: nice catch

examples/tests/valid/struct_from_json.w
MarkMcCulloh: Based on this https://github.com/winglang/wing/issues/3943 I take it does not and is being left for later

MarkMcCulloh (approved):

examples/tests/valid/struct_from_json.w
hasanaburayyan: Yupp thats exactly right @MarkMcCulloh, Im brainstorming the JSII imported structs now. The issue here is that we arent jsifying those so we never create a jsonschema for them to use for validation. I have a few ideas that are probably not gonna work but may 😁

Issue #2405, opened 2023-05-04
good dev experience to a local website


tsuf239:
### Feature Spec

following: https://github.com/winglang/wing/pull/2164#discussion_r1175668067

We need to think of ways to make the development of wing's local website better- 
it starts with exposing the website URL and making it permanent (for easier development ) and continues with running the website while developing it

UPDATE: 
Ultimately, I decided to create a new resource- `ReactWebsite`.
The resource points to a react folder, and either builds it or starts it, allowing hot reloading and reacting to wing file changes.
variables can be passed from the wing file to the react app via the `addEnvironment` method.
in order to use it, the user needs to add this line: `<script src="%PUBLIC_URL%/wing.js"></script>` to the react's index file. 

### Use Cases

running a website locally

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skyrpex:
Looking good! I’m wondering if it would make sense to use more specific JS APIs to build the websites (instead of relying on cli commands).

For example, Vite has the [createServer API](https://vitejs.dev/guide/api-javascript.html#javascript-api). Vite it supports any JS framework out there.

skyrpex:
Would be wonderful if we could generate type definitions for the environment variables provided from Wing code.

In Vite, you can define what variables you expect to come from the environment as explained here https://vitejs.dev/guide/env-and-mode.html#intellisense-for-typescript.

The [Nuxt](https://nuxt.com/) framework generates a `.nuxt` directory that contains a tsconfig.json and some useful types.

<img width="638" alt="image" src="https://github.com/winglang/wing/assets/1077520/aa082536-e39a-454c-bb8d-968e1c8ab9a6">

skyrpex:
One of the types that Nuxt generates are the API endpoints of your application. Let's say, you have a `server/api/hello.ts` endpoint that returns a `{ hello: "world" }` response.

In any web component of your app, you can call `const { data } = await useFetch("/api/hello");`, and `data` is correctly typed to `{ hello: "world" }`. This is achieve thanks to one of the generated types `.nuxt/types/nitro-routes.d.ts`, which references our `server/api/hello.ts` file as seen below:

```ts
// Generated by nitro
import type { Serialize, Simplify } from 'nitropack'
declare module 'nitropack' {
  type Awaited<T> = T extends PromiseLike<infer U> ? Awaited<U> : T
  interface InternalApi {
[...]

Issue #2271, opened 2023-04-26
cloud.Api - static port for api simulator


RaphaelManke:
### Feature Spec

I would like to have a static port when the cloud.Api resource is started in simulator.

### Use Cases

This helps in cases where I develop against the api e.g. with a Frontend or want to make api calls with an external api client like insomnia oder postman. 
On top of that it enables the use of http reverse proxy like ngrok.

I discovered this lack of capability while developing a slack bot. To test it I used ngrok to forward the slack commands to the local machine. 

### Component

SDK, Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

RaphaelManke:
I would add an idea like 
http://localhost:<staticPort>/<apiIdentifier>

This allows a port be defined by the user e.g. if a default port is not suitable and also support multiple APIs in an wing app.

eladb:
This is a duplicate with #2227 -- I am not sure a static port is necessarily the correct way to address this use case.

skorfmann:
> This is a duplicate with #2227 -- I am not sure a static port is necessarily the correct way to address this use case.

@eladb I'm not sure that #2227 is really covering this. I think exposing a service to the public internet is quite a different use-case from wanting to have a locally available, static port - even if not necessarily technically, certainly from an organizational policy point of view e.g.: I might have another application I develop running locally (no matter if frontend or backend), where all I want is a known address to a Wing API. In particular in enterprise contexts, I could see this become a dealbreaker if the only option was exposing this via something like ngrok.

staycoolcall911:
Duplicates #2227

staycoolcall911:
Oops, missed some important comments on this issue...
Reopening - we'll discuss it on a next team time

Chriscbr:
We've made two changes that should help here:
- The ports used by `cloud.Api` on the simulator are now reused (https://github.com/winglang/wing/pull/5736)
- An ngrok winglib is now available: https://github.com/winglang/winglibs/tree/main/ngrok

Let us know if these improvements solve your use case(s). I'll downgrade this issue to P2 for now - but if there are use cases or situations that aren't addressed by these changes, I think it could still be feasible to support static ports via an optional custom platform parameter: https://www.winglang.io/docs/concepts/platforms#defining-custom-platform-parameters

PR #3958, opened 2023-08-24
feat(sdk): react x wing integration

tsuf239:
fixes: [#2405](https://github.com/winglang/wing/issues/2405)
demo: https://www.loom.com/share/0bd2d7e8982e4a568f0db3eeb9e427cd?sid=503d1dca-c28e-4bf2-82ff-f6fbcc5433cc

This resource allows to enjoy developing react applications, using wing as a backend, and then deploy them to the cloud (or building them to the console) with no additional steps or configurations. 🎉 🎉 🎉 

I created a new Resource called `ReactWebsite`, it takes a file path to the react project, and a flag to know if the project should be started (`npm run start` by default, allowing hot reloading) or built (`np run build` by default). 
In a cloud target, the website will always be built.
Variables can be passed from the wing files to the react app via the `addEnvironment` method.
The resource's URL is accessible to wing via the `url` getter.
The wing variables are accessible via `window.wingEnv` in react.

//TODO:
- [x] Add wing tests
- [x] Add ts tests
- [x] Add documentation details and doc strings
- [ ] Let the world know :) 

## Checklist

- [ ] Tests added (always)


marciocadev:
This is really cool @tsuf239

I can't wait to try this.

olivernybroe:
As this is just tying itself to having a watch and build command, maybe it should be named more generally? 🤔

tsuf239:
> As this is just tying itself to having a watch and build command, maybe it should be named more generally? 🤔

That's a good point :) what name would you choose to describe this resource?

libs/wingsdk/src/target-sim/react-website.inflight.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.startCommand = props.startCommand;
+    this.environments = props.environments;
+    this.isDevRun = props.isDevRun;
+  }
+
+  public async init(): Promise<ReactWebsiteAttributes> {
+    this.addTrace(`Executing start command: ${this.startCommand}`);
+
+    writeFileSync(
+      join(`${this.path}`, "/public/wing.js"),
```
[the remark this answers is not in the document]
olivernybroe: oh that's interesting, did not thing about how to build that for aws 🤔 

 Maybe it can be solved with dotenv vault feature?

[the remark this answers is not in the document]
tsuf239: I'm not sure... Are the env vars resolved before calling the build command? 
I know that react build is switching the env variables with strings upon the build, and then there isn't a very good way to change them (other than bypassing it using the `window` global variable or another variable, or making sure they'll be available before the build)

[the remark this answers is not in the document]
olivernybroe: Hmm... 

Alright so I misunderstood the issue. So we have 3 phases: 
- Build
- preflight
- inflight

We are building the website in `build` step, but we need data from `preflight` for building. 
That is obviously not possible, so we therefore needs to move the website building step into `preflight` instead. 

This means for `tf-aws` target, we need to build the website in Terraform. Maybe we can do it by "just" running the npm command in terraform via `external`? 
https://registry.terraform.io/providers/hashicorp/external/latest/docs/data-sources/external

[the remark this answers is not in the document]
tsuf239: As far as I understood, it will bind the command to the terraform process, running it before `apply`.  At this point in time we still don't have the actual URL, (until the build is over)... so I am afraid it will have the same result (unresolved tf token bound as the value of the environment variable to the build output)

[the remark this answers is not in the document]
olivernybroe: Hmm I see. 
My terraform is not good enough for knowing the solution here (if one exist)

libs/wingsdk/src/cloud/react-website.ts
```diff
@@ -0,0 +1,125 @@
...
+  public get url(): string {
+    return this._websiteHost?.url ?? `http://localhost:${this._localPort}`;
+  }
+
+  private _parsePath(scope: Construct, path: string) {
+    if (isAbsolute(path)) {
+      return path;
+    } else if (!App.of(scope).entrypointDir) {
+      throw new Error("Missing environment variable: WING_SOURCE_DIR");
+    }
```
Chriscbr, line 113: It feels to me like the logic for checking the environment variable should be handled in a single place in the SDK. Maybe it's possible to do it in `App`?

```ts
// app.ts
get entrypointDir() {
  if (!process.env.WING_SOURCE_DIR) {
    throw new Error("missing environment variable: WING_SOURCE_DIR");
  }
  return process.env.WING_SOURCE_DIR;
}
```

libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -240,6 +241,17 @@ export interface WebsiteSchema extends BaseResourceSchema {
   };
 }
 
+export interface ReactWebsiteSchema extends BaseResourceSchema {
+  readonly type: typeof REACT_WEBSITE_TYPE;
+  readonly props: {
+    path: string;
+    startCommand: string;
+    environments: Record<string, string>;
```
Chriscbr, line 249:
```suggestion
    environmentVariables: Record<string, string>;
```

libs/wingcompiler/src/compile.ts
```diff
@@ -213,7 +213,7 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
   });
 
   try {
-    vm.runInContext(artifact, context);
+    await vm.runInContext(artifact, context);
```
Chriscbr, line 216: Is this to change handle async code from somewhere?

libs/wingsdk/src/cloud/react-website.ts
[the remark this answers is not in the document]
tsuf239: I'm not sure it also renders those docs from the `ex` directory, I'll check and open an issue if needed

libs/wingcompiler/src/compile.ts
tsuf239: isn't needed at the end- removed, thanks for noticing!

libs/wingsdk/src/ex/react-website.ts
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * In sim, if `true` - will use the start command, and if `false` - the build command
+   * @default false
+   */
+  readonly isDevRun?: boolean;
+  /**
+   * Additional properties to run the website host with
+   * @default {}
+   */
+  readonly hostProps?: BaseWebsiteProps;
```
Chriscbr, line 51: Instead of requiring the user pass `Website` props through `hosProps`, can we flatten all of the props? i.e. `ReactWebsiteProps extends BaseWebsiteProps`

libs/wingsdk/src/target-sim/react-website.ts
```diff
@@ -0,0 +1,70 @@
+import { Construct } from "constructs";
+import { ISimulatorResource } from "./resource";
+import { ReactWebsiteSchema, REACT_WEBSITE_TYPE } from "./schema-resources";
+import { bindSimulatorResource, makeSimulatorJsClient } from "./util";
+import * as cloud from "../cloud";
+import * as core from "../core";
+import * as ex from "../ex";
+import { IInflightHost } from "../std";
+import { BaseResourceSchema } from "../testing/simulator";
+// import { simulatorAttrToken } from "./tokens";
```
Chriscbr, line 10:
```suggestion
```

libs/wingsdk/src/ex/react-website.ts
```diff
@@ -0,0 +1,163 @@
...
+      throw new Error(`non existent directory '${this._projectPath}'`);
+    }
+  }
+
+  /**
+   * Website's url
+   */
+  public get url(): string {
+    return this._websiteHost?.url ?? `http://localhost:${this._localPort}`;
+  }
```
Chriscbr, line 141: localhost is a detail about how ReactWebsite works on the `sim` target, not for all targets, so I think this logic for `url()` and `_localPort` should be inside the simulator class implementation (not in this base class).

libs/wingsdk/src/target-sim/react-website.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /** @internal */
+  public _toInflight(): string {
+    return makeSimulatorJsClient(__filename, this);
+  }
+
+  /** @internal */
+  public _getInflightOps(): string[] {
+    return [];
+  }
```
Chriscbr, line 69: I think we might be able to move this to the base class since `ReactWebsite` has the same inflight methods on all targets

libs/wingsdk/src/target-tf-aws/react-website.ts
```diff
@@ -0,0 +1,78 @@
...
+  public _preSynthesize(): void {
+    const env = Object.fromEntries(this._environmentVariables.entries());
+    const content = `window.wingEnv = ${JSON.stringify(env, null, 2)};`;
+
+    new S3Object(this._websiteHost as Website, `File--wing.js`, {
+      dependsOn: [(this._websiteHost as Website).bucket],
+      content,
+      bucket: (this._websiteHost as Website).bucket.bucket,
+      contentType: "text/javascript",
+      key: "wing.js",
```
Chriscbr, line 60: Extract this to a constant

examples/tests/sdk_tests/website/react-website/package.json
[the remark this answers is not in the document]
tsuf239: Good question, let me check, perhaps I'll be able to remove it

libs/wingsdk/src/ex/react-website.ts
tsuf239: good idea!

libs/wingsdk/src/ex/react-website.md
```diff
@@ -0,0 +1,88 @@
...
+
+or customizing them:
+
+```ts
+bring cloud;
+bring util;
+
+let website = new cloud.Website(
+  projectPath: "./client",
+  isDevRun: util.tryEnv("ENV") == "dev"
```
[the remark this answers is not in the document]
tsuf239: it's a flag indicating that we are running a local development server for react (by running npm start) other than building it (and serving the static files via express/s3), how would you call that?

`NODE_ENV ` is a common env var used in node and react, indicating what environment are we on (production, testing, development, etc...) since it's not a node app, I omitted the `NODE` and left with `ENV` should I call it differently?

examples/tests/sdk_tests/website/react-website/package.json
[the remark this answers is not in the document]
tsuf239: https://github.com/winglang/wing/blob/e264095fc9a67b78218e7ecd08e8b96996758227/pnpm-workspace.yaml#L9
it should be ok, I think

libs/wingsdk/src/ex/react-website.ts
tsuf239: unfortunately, it's not easy to take out only the BaseWebsiteProps from all the props and put them into the website (since I can't iterate over interface keys, and I don't want to have to update it manually- it can cause bugs pretty easily)

libs/wingsdk/src/ex/react-website.md
```diff
@@ -0,0 +1,86 @@
...
+The `ex.ReactWebsite` resource represents a website, built using React, that can be both hosted in the cloud or to run on a development hot-reload server in the sim target.
+
+## Usage
+
+### Initialization
+
+Using the default arguments:
+
+```ts
+bring cloud;
```
staycoolcall911, line 18: Shouldn't we bring ex; instead of cloud here? The whole example should be for an `ex.ReactWebsite` instead of a cloud.Website or did I get it wrong?

libs/wingsdk/src/ex/react-website.md
```diff
@@ -0,0 +1,86 @@
...
+  startCommand: "pnpm start" // default is "npm start"
+  buildCommand: "pnpm build" // default is "npm build"
+  hostProps: {} // website resource props (expect of the "path" prop)
+  localPort: "4000" // default is 3001
+ );
+```
+
+- When calling `wing it` when `isDevRun` is `true`- wing starts both the simulator and React's development server
+- When calling `wing it`/`wing compile` on the sim target, when `isDevRun` is `false`- wing builds React and serves it over the [Website resource](./website.md).
+- When calling `wing compile` to any other target - wing builds react locally then deploys it to the cloud! (using the [Website resource](./website.md))
```
staycoolcall911, line 43: This behavior is a little bit strange - usually `wing compile` is not deploying anything. I understand you're probably doing that because you want to support hot-reloading, but it is surprising that `wing compile` behaves differently here. Perhaps we should introduce `wing watch`?

libs/wingsdk/src/ex/react-website.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### using wing variables within react code
+
+`ex.ReactWebsite` allows you to pass preflight arguments from wing to the React app using `addEnvironment` method:
```
staycoolcall911, line 47: I second @Chriscbr that `addEnvironment` is not clear and should be replaced by `addEnvironmentVariable`
```suggestion
`ex.ReactWebsite` allows you to pass preflight arguments from wing to the React app using `addEnvironmentVariable` method:
```

libs/wingsdk/src/cloud/react-website.ts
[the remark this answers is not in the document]
tsuf239: Added to the docs site in the current PR

libs/wingsdk/src/ex/react-website.md
tsuf239: No, it doesn't actually deploys, I'll change the working, thanks for noticing!

libs/wingsdk/src/ex/react-website.md
tsuf239: ok

libs/wingsdk/src/target-tf-aws/react-website.ts
```diff
@@ -0,0 +1,73 @@
...
+      target: this._websiteHost as Website,
+      name: `host`,
+    });
+  }
+
+  public _preSynthesize(): void {
+    const env = Object.fromEntries(this._environmentVariables.entries());
+    const content = `window.wingEnv = ${JSON.stringify(env, null, 2)};`;
+
+    new S3Object(this._websiteHost as Website, `File--${ex.WING_JS}`, {
```
eladb, line 55: Why not use bucket.addFile?

libs/wingsdk/src/target-tf-aws/react-website.ts
```diff
@@ -0,0 +1,73 @@
...
+import * as cloud from "../cloud";
+import { Connections } from "../core";
+import * as ex from "../ex";
+
+/**
+ * AWS implementation of `ex.ReactWebsite`.
+ *
+ * @inflight `@winglang/sdk.cloud.IReactWebsiteClient`
+ */
+export class ReactWebsite extends ex.ReactWebsite {
```
eladb, line 18: Feels like this should be shared across all non-sim implementations.

eladb (dismissed):

libs/wingsdk/src/ex/react-website.ts
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * A command for building the React app
+   * @default "npm run build"
+   */
+  readonly buildCommand?: string;
+  /**
+   * In sim, if `true` - will use the start command, and if `false` - the build command
+   * @default false
+   */
+  readonly isDevRun?: boolean;
```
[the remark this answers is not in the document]
tsuf239: I wanted to enable two possible behaviors on the sim target (similar to when developing a react app)-
1- The developer is working on his react app- interacting with his files and developing more features (98% of the time I guess)
2- The developer is done working on his app, so he builds it (and its servers over a nodeJs or a similar static server ) - I usually did it only when testing specific parts in that area, it'll used by the preview env (otherwise won't be accessible)
As a developer, I'd still want to be able to choose whether to run build or start- but I agree that the default of the sim env should be start

libs/wingsdk/src/ex/react-website.ts
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * Additional properties to run the website host with
+   * @default {}
+   */
+  readonly hostProps?: BaseWebsiteProps;
```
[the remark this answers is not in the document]
tsuf239: because then we'll have to pass them into the `websiteHost`-and since we cannot iterate over `BaseWebsiteProps` there is a good chance that the next time someone adds something to it- we'll forget to update this file as well :/

libs/wingsdk/src/ex/react-website.ts
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * A port to start a local build of the React app on.
+   * @default 3001
+   */
+  readonly localPort?: string | number;
```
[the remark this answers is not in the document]
tsuf239: This allows us to run more than one React website at the same time, I believe it's needed

libs/wingsdk/src/target-tf-aws/react-website.ts
tsuf239: since it doesn't support content type, I can add an issue (also for the cloud.website files)

libs/wingsdk/src/target-tf-aws/react-website.ts
tsuf239: since it doesn't support content-type, I can add an issue to add it, as well as convert cloud.website files to be added this way

libs/wingsdk/src/target-tf-aws/react-website.ts
tsuf239: I tend to agree, but we still don't know if all of them are implementing website host using a bucket... it should be considered when developing to different target, I'll add this point to the other targets tickets

libs/wingsdk/src/ex/react-website.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * A cloud deployable React website.
+ *
+ * @inflight `@winglang/sdk.ex.IReactWebsiteClient`
+ */
+export abstract class ReactWebsite extends Resource {
```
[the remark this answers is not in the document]
tsuf239: Any class that extends a resource is an abstract class, and the target-specific ones are concrete classes...
am I missing something?

examples/tests/sdk_tests/website/react-website/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "react-website",
```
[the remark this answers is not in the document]
tsuf239: we need a folder with package.json in order to mock react for the tests (instead of handling with a heavy real react project in our repo- that isn't really needed)

examples/tests/sdk_tests/website/react-website/index.js
```diff
@@ -0,0 +1,8 @@
+const express = require("express");
```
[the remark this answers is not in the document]
tsuf239: it serves the mocked "react" files- as a real react development server would do (I'm testing `wing it` manually with that, and also ensures that the `reactApp.url` is working when running it on a local server)

examples/tests/sdk_tests/website/react-website.main.w
```diff
@@ -0,0 +1,29 @@
...
+
+api.get("/", inflight () => {
+  return {body: "hi!", status: 200};
+});
+
+// the project isn't actually react (since it can make wing's size much larger) 
+// but a mock with start and build commands
+let website = new ex.ReactWebsite(
+  projectPath: "./react-website", 
+  buildFolder: "/build/public",
```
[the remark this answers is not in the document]
tsuf239: it works either way- it's joined to the project path

examples/tests/sdk_tests/website/react-website/public/index.js
```diff
@@ -0,0 +1,2 @@
+document.getElementById("par1").innerText = `apiUrl is: ${window.wingEnv?.apiUrl}`;
```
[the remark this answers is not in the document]
tsuf239: it is, 
it's just for a case we are serving this file in a n independent way, before running it via `ex.ReactApp`

examples/tests/sdk_tests/website/react-website.main.w
```diff
@@ -0,0 +1,29 @@
...
+);
+
+let preflightVariable = "preflight variable";
+website.addEnvironment("apiUrl", api.url);
+website.addEnvironment("anotherEnvVar", preflightVariable);
+
+test "website is working" {
+  let res = http.get(website.url);
+  assert(res.ok);
+  // unfortunately we can't test the env variables, since it's js and get added in the browser...
```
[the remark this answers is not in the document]
tsuf239: found a simpler way :)

examples/tests/sdk_tests/website/react-website.main.w
```diff
@@ -0,0 +1,29 @@
...
+// but a mock with start and build commands
+let website = new ex.ReactWebsite(
+  projectPath: "./react-website", 
+  buildFolder: "/build/public",
+  isDevRun: util.tryEnv("ENV") == "dev"
+);
+
+let preflightVariable = "preflight variable";
+website.addEnvironment("apiUrl", api.url);
+website.addEnvironment("anotherEnvVar", preflightVariable);
```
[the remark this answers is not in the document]
tsuf239: yes, does it require any change?

libs/wingsdk/src/cloud/react-website.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Options for `ReactWebsite`.
+ */
+export interface ReactWebsiteProps {
+  readonly projectPath: string;
+  readonly buildFolder?: string;
+  readonly startCommand?: string;
+  readonly buildCommand?: string;
+  readonly isDevRun?: boolean;
```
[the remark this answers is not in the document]
tsuf239: https://github.com/winglang/wing/pull/3958#discussion_r1331366601

libs/wingsdk/src/ex/react-website.md
```diff
@@ -0,0 +1,86 @@
...
+bring cloud;
+bring util;
+
+let website = new cloud.Website(
+  projectPath: "./client",
+  isDevRun: util.tryEnv("ENV") == "dev"
+  buildFolder: "/dist" // default is "/build"
+  startCommand: "pnpm start" // default is "npm start"
+  buildCommand: "pnpm build" // default is "npm build"
+  hostProps: {} // website resource props (expect of the "path" prop)
```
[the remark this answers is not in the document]
tsuf239: I rewrote it, is it ok now?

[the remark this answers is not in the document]
staycoolcall911: Yes, much better!
Please just delete the extra react-website.md file, now that you renamed to react-app

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,88 @@
...
+  hostProps: {} // properties that apply to the react app host, which is a `cloud.Website` resource
+  localPort: "4000" // default is 3001
+ );
+```
+
+- When calling `wing it` with `isDevRun` set to `true`, Wing will start both the simulator and React's development server.
+- When calling `wing it` or `wing compile` for the `sim` target with `isDevRun` set to `false`, Wing will build React and serve it via the [Website resource](../01-cloud/website.md).
+- When calling `wing compile` for any other target, Wing will build React locally and generate output files, preparing it for deployment to the cloud using the [Website resource](../01-cloud/website.md).
+
+#### using wing variables within react code
```
Chriscbr, line 45:
```suggestion
### Using wing variables within react code
```

libs/wingsdk/src/ex/react-website.md
Chriscbr: @staycoolcall911 I don't have a strong preference between these names (maybe @eladb has an opinion?), but the name should match the corresponding method in `cloud.Function` since it has the same functionality.

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,88 @@
...
+
+`ex.ReactApp` allows you to pass preflight arguments from wing to the React app using `addEnvironmentVariable` method:
```
staycoolcall911, line 47: > @staycoolcall911 I don't have a strong preference between these names (maybe @eladb has an opinion?), but the name should match the corresponding method in cloud.Function since it has the same functionality.

@Chriscbr - thanks for the feedback, I wasn't aware of the `cloud.Function` method. In that case, let's keep it consistent. Sorry for the back and forth @tsuf239... let's revert to use `addEnvironment`

libs/wingsdk/src/target-tf-aws/react-website.ts
eladb: I think `ReactWebsite` should be implemented using a `Website`, no? And as far as I recall, there's an API for [adding a JSON](https://www.winglang.io/docs/standard-library/cloud/website#@winglang/sdk.cloud.Website.addJson) config to the website.

I don't see a need for `ReactWebsite` to access the underlying bucket. It should just use the `cloud.*` abstractions.

libs/wingsdk/src/target-tf-aws/react-website.ts
tsuf239: added

tsuf239:
Now the ReactApp is using only the website resource (no additional cloud-specific methods) for the non-sim targets. In the following PR I'll move the class to a shared place

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,88 @@
...
+bring util;
+
+let website = new ex.ReactApp(
+  projectPath: "./client",
+  isDevRun: util.tryEnv("ENV") == "dev" // `true`` by default. will run the start command if true, and the build command if not
+  buildDir: "/dist" // default is "/build"
+  startCommand: "pnpm start" // default is "npm start"
+  buildCommand: "pnpm build" // default is "npm build"
+  hostProps: {} // properties that apply to the react app host, which is a `cloud.Website` resource
+  localPort: "4000" // default is 3001
```
Chriscbr, line 37: A few thoughts/questions:
1. Should the default be a random local port? If you have more than one react app, they will both try to serve to 3001 and it could error.
2. It smells a little weird that this is customizable for ReactApp but not for Website.
3. If this option is relevant for the `sim` target I'm not sure we should have it on the public, target-agnostic API - usually the public API only has options that can be configured for all targets. I guess this is related to https://github.com/winglang/wing/issues/2271. All that said I'm ok keeping `localPort` for now and revisiting this when we decide on a solution for `Api`

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,88 @@
...
+ );
+```
+
+- When calling `wing it` with `isDevRun` set to `true`, Wing will start both the simulator and React's development server.
+- When calling `wing it` or `wing compile` for the `sim` target with `isDevRun` set to `false`, Wing will build React and serve it via the [Website resource](../01-cloud/website.md).
+- When calling `wing compile` for any other target, Wing will build React locally and generate output files, preparing it for deployment to the cloud using the [Website resource](../01-cloud/website.md).
+
+### Using wing variables within react code
```
Chriscbr, line 45:
```suggestion
### Using Wing variables within react code
```

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,88 @@
...
+
+`ex.ReactApp` allows you to pass preflight arguments from wing to the React app using `addEnvironment` method:
```
Chriscbr, line 47:
```suggestion
`ex.ReactApp` allows you to pass preflight string values from Wing to the React app using `addEnvironment` method:
```

libs/wingsdk/src/cloud/website.ts
```diff
@@ -91,7 +91,27 @@ export abstract class Website extends Resource {
...
+   * If the path conflicts with file path from the website's static assets, an error will be thrown.
+   * @param path the file path it will be uploaded as
+   * @param data the data to write to the file
+   * @param contentType the file content-type
+   */
+  public abstract addFile(
+    path: string,
+    data: string,
+    contentType: string
+  ): string;
```
Chriscbr, line 114: I think this method name could confuse users since the methods `addFile` exists on `Bucket` and `Website` but they have different meanings:

- In `Bucket`, `addFile()` expects a file path (typed as `str`)
- In `Website`, `addFile()` expects raw data (typed as `str`)

Maybe we can rename the second one to `addFileData` for now? (I don't love this name - open to suggestions)

libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -231,16 +232,32 @@ export interface RedisSchema extends BaseResourceSchema {
...
+
 /** Schema for cloud.Website */
 export interface WebsiteSchema extends BaseResourceSchema {
   readonly type: typeof WEBSITE_TYPE;
   readonly props: {
     /** Path to the directory where all static files are hosted from */
     staticFilesPath: string;
     /** Map of `.json` file paths to dynamic content inserted from preflight */
-    jsonRoutes: Record<string, Json>;
+    fileRoutes: FileRoutes;
```
Chriscbr, line 244: Fix docstring

libs/wingsdk/src/ex/react-website.md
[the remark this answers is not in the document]
tsuf239: but we don't really use nodeJS 🤔

libs/wingsdk/src/ex/react-website.ts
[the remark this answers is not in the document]
tsuf239: that's a good question, `ENV` or `NODE_ENV` will be set to "production" when the app is in the cloud (I guess we don't want to set it that way locally since we can accidentally use production apis and services... )
it's set by the tool that builds the react to its dist files 
I think I just remove that part to eliminate confusion, good catch Chris!

docs/docs/04-standard-library/07-regex/api-reference.md
```diff
@@ -0,0 +1,59 @@
+---
+title: API reference
```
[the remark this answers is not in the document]
tsuf239: image
done :)

libs/wingsdk/src/cloud/website.ts
```diff
@@ -91,7 +91,27 @@ export abstract class Website extends Resource {
...
+   * Add a file with custom value during the website's deployment.
+   * If the path conflicts with file path from the website's static assets, an error will be thrown.
+   * @param path the file path it will be uploaded as
+   * @param data the data to write to the file
+   * @param contentType the file content-type
+   */
+  public abstract addFile(
+    path: string,
+    data: string,
+    contentType: string
```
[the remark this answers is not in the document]
tsuf239: not if you pass a plain string (as far as I know)

libs/wingsdk/src/ex/react-app.md
tsuf239: it's only customizable for running react on dev mode (when building it it just get the port of the hosting website), but I agree with the rest.
it shouldn't be a random port since it will make it impossible to develop in dev mode- it will close the react dev server every time the wing file updates/ running `wing it` which breaks the hot reloading (and a bad experience for the developer too)

libs/wingsdk/src/cloud/website.ts
tsuf239: maybe `createFileFromStr` (on the website)?

libs/wingsdk/src/ex/react-website.md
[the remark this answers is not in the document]
tsuf239: because the sim got 2 modes- development and build... (at the end I just passed a boolean in my examples I think that's ok )

docs/docs/04-standard-library/06-ex/dynamodb-table.md
```diff
@@ -1,14 +1,10 @@
 ---
-title: API reference
-id: api-reference
-description: Wing standard library API reference for the ex module
-keywords: [Wing sdk, sdk, Wing API Reference]
-hide_title: true
-sidebar_position: 100
+title: Dynamodb Table
```
staycoolcall911, line 2: This seems like an unintended change, right?

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,87 @@
+---
+title: React Website
```
staycoolcall911, line 2: Name consistency - shouldn't this match the .md file name? I'd expect both to be React App.
```suggestion
title: React App
```

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,87 @@
...
+id: react-website
```
staycoolcall911, line 3:
```suggestion
id: react-app
```

libs/wingsdk/src/target-awscdk/website.ts
```diff
@@ -62,10 +62,14 @@ export class Website extends cloud.Website {
...
-  public addFile(path: string, data: string, contentType: string): string {
+  public addFile(
+    path: string,
+    data: string,
+    options?: cloud.AddFileOptions
+  ): string {
     new BucketDeployment(this, `S3Object-${path}`, {
       destinationBucket: this.bucket,
-      contentType: contentType,
+      contentType: options?.contentType ?? "tesxt/plain",
```
Chriscbr, line 72:
```suggestion
      contentType: options?.contentType ?? "text/plain",
```

libs/wingsdk/src/target-tf-aws/react-app.ts
```diff
@@ -56,11 +63,9 @@ export class ReactApp extends ex.ReactApp {
...
       2
     )};`;
-    (this._websiteHost as Website).addFile(
-      ex.WING_JS,
-      content,
-      "text/javascript"
-    );
+    (this._websiteHost as Website).addFile(ex.WING_JS, content, {
+      contentType: "text/javascript",
+    });
```
Chriscbr, line 68: Can we make `_host` typed as `cloud.Website` so that we don't have to perform this type cast?

libs/wingsdk/.projenrc.ts
```diff
@@ -1,28 +1,6 @@
...
-  cloudFiles.map((filename) => filename.split(".")[0])
-);
-
-UNDOCUMENTED_CLOUD_FILES.forEach((file) => cloudResources.delete(file));
-
-const undocumentedResources = Array.from(cloudResources).filter(
-  (file) => !cloudFiles.includes(`${file}.md`)
-);
-
-if (undocumentedResources.length) {
```
staycoolcall911, line 18: I guess you moved this code because it didn't really work? Does it work now? IMO the build should fail when a new resource doesn't have docs

libs/wingsdk/.projenrc.ts
```diff
@@ -33,9 +11,10 @@ const CDKTF_PROVIDERS = [
   "google@~>4.63.1",
 ];
 
-const PUBLIC_MODULES = ["std", "http", "util", "aws", "ex", "math", "regex"];
+const PUBLIC_MODULES = ["std", "http", "util", "aws", "math", "regex"];
```
staycoolcall911, line 14: Why does `ex` gets special treatment?

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,87 @@
...
+
+or customizing them:
+
+```ts
+bring ex;
+bring util;
+
+let website = new ex.ReactApp(
+  projectPath: "./client",
+  isDevRun: false // `true`` by default. will run the start command if true, and the build command if not
```
staycoolcall911, line 32:
```suggestion
  isDevRun: false // `true` by default. Will run the start command if true, and the build command if not
```

libs/wingsdk/src/ex/react-app.md
```diff
@@ -0,0 +1,87 @@
...
+  startCommand: "pnpm start" // default is "npm start"
+  buildCommand: "pnpm build" // default is "npm build"
+  localPort: 4000 // default is 3001
+ );
+```
+
+When `ReactApp` is compiled to the `sim` target, by default it runs the start command (default: `npm start`) inside `projectPath` to serve your app in development mode on a local port.
+
+If the `CI` environment variable is set OR if `ReactApp` is compiled to any other target, it will run the build command (default: `npm build`) inside of the `projectPath` to build the React app for production to `buildDir` and serve the app.
+
```
Chriscbr, line 43: It doesn't look like this was implemented-

If we're keeping `isDevRun` then let's update the docs to make sure it's clear that the option only has an effect on the `sim` target, and that the `buildCommand` is always used for building the react app on other targets.

Chriscbr (approved):

libs/wingsdk/src/ex/react-app.ts
```diff
@@ -0,0 +1,159 @@
...
+import { WebsiteOptions, Website } from "../cloud/website";
+import { fqnForType } from "../constants";
+import { App } from "../core";
+import { Resource, Node } from "../std";
+
+const DEFAULT_BUILD_FOLDER = "/build";
+export const DEFAULT_REACT_APP_BUILD_COMMAND = "npm run build";
+const DEFAULT_PORT = 3001;
+
+export const WING_JS = "wing.js";
```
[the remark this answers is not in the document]
staycoolcall911: To me that's still unclear

libs/wingsdk/src/ex/react-app.ts
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * A command for building the React app
+   * @default "npm run build"
+   */
+  readonly buildCommand?: string;
+  /**
+   * In sim, if `true` - will use the start command, and if `false` - the build command
+   * @default true
+   */
+  readonly isDevRun?: boolean;
```
staycoolcall911, line 49: I know it came up before, but this property still feels a bit strange. Let's try to discuss and find a better name for it or a better docstring

staycoolcall911 (approved):
Great work @tsuf239! I would split the missing `ex` resources docs to a different PR.
This is almost ready to go live 💪

docs/docs/04-standard-library/06-math/api-reference.md
```diff
@@ -0,0 +1,772 @@
+---
+title: API reference
```
staycoolcall911, line 2: I noticed something went wrong here. We have both 06-ex and 06-math, while we already have 07-math on the doc site: https://www.winglang.io/docs/standard-library/math/api-reference (which is defined in https://github.com/winglang/wing/blob/main/docs/docs/04-standard-library/07-math)

staycoolcall911 (changes requested):

libs/wingsdk/src/ex/react-app.ts
tsuf239: switched to `useBuildCommand`

docs/docs/04-standard-library/06-ex/dynamodb-table.md
tsuf239: image
github is just being confused - it thinks I took the api-reference file and renamed it into "dynamodb-table" while I deleted the api-reference and added a new dynamodb md file

docs/docs/04-standard-library/06-math/api-reference.md
tsuf239: fixed the numbers

libs/wingsdk/src/cloud/website.ts
tsuf239: I'll postpone it to a following PR

staycoolcall911 (approved):
Let's get it merged!

skyrpex (approved):

PR #3963, opened 2023-08-25
feat(sdk): use arm64 for `awscdk` and `tf-aws` functions

Chriscbr:
This change should give better price/performance for most workloads, and not have any effect on runtime behavior since all cloud functions are using the AWS-provided NodeJS runtimes.

## Checklist

- [x] Tests added (always)


marciocadev:
I went to check if all regions are already with the arm64 (about 1 year ago there wasn't here in Brazil yet), but apparently all regions are already available.

I think this PR is great.

PR #3972, opened 2023-08-25
chore(compiler): make preflight.js more self-contained

Chriscbr:
Currently the `preflight.js` files generated by Wing contain a reference to a value named `$plugins` which has to be injected by a NodeJS VM. Because of this, it's not possible to run `node path/to/.wing/preflight.js` for debugging purposes without manually editing the file to stub in a fake value.

I've removed the need to inject this value by passing the list of plugin paths through an environment variable instead.

## Checklist

- [ ] Tests added (always)


Chriscbr:
Added the e2e-full label to make sure this env var usage doesn't break anything on Windows

MarkMcCulloh (approved):

PR #3973, opened 2023-08-26
fix(sdk): align 404 status across cloud.Api targets

skorfmann:
**Issue:** 

When an inexistent resource path is provided, AWS Rest API Gateway currently returns a 403 - Forbidden response with the message `Missing Authentication Token`. This is the default behavior of AWS Rest API Gateway. However, it's misleading and differs semantically from the local simulator's response.

**Previous Behavior:** 

- Local simulator: `curl ./not-existing -> 404`
- tf-aws: `curl ./not-existing -> 403` 

**Proposed Change:** 

- Local simulator: `curl ./not-existing -> 404`
- tf-aws: `curl ./not-existing -> 404`

My intention was also to synchronize the response bodies. However, I struggled to find the appropriate way to integrate an expressjs middleware as the final handler to modify the 404 response. An attempt was made [here](https://github.com/winglang/wing/blob/2b2aa817bef5f56c56c8a0d02961858dd47c5860/libs/wingsdk/src/target-sim/api.inflight.ts#L132-L135) without success. Suggestions on how to achieve this are welcome.

**Solution Considerations:**

Two general solutions were explored:

1. Modify the [Api Gateway Response](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gatewayResponse-definition.html)

This would likely override actual 403 responses, see details [here](https://dev.to/aws-builders/better-error-messages-for-non-existing-api-resources-with-amazon-api-gateway-50pb). Hence, it doesn't seem like a good approach.

2. Implement a Greedy Handler

I opted for the greedy mock handler method as it appeared to be the more straightforward solution. As long as there's no such option on the `cloud.API` it's seems safe to do so. Once something like this would be implemented, this will have to be adapted accordingly.

## Checklist

- [ ] Tests added (always)


Chriscbr (approved):
Looks great - the implementation makes sense to me. A default 404 handler that doesn't require invoking any other cloud functions sounds like a sane default - I've created an issue [here](https://github.com/winglang/wing/issues/3983) for covering use cases where the user wants to customize it further. 👍

PR #3976, opened 2023-08-26
feat(sdk): `expect` module with assertions for testing

skorfmann:
After some back and forth we settled on using `expect` as the module name. Mostly for the reason that `assert` is a builtin function already and would cause naming conflicts. The pull request discussions for more details. 

## Checklist

- [ ] Tests added (always)


apps/wing/src/commands/test.ts
```diff
@@ -247,7 +247,7 @@ function noCleanUp(synthDir: string) {
 }
 
 async function testSimulator(synthDir: string, options: TestOptions) {
-  const s = new testing.Simulator({ simfile: synthDir });
+  const s = new simulatorTesting.Simulator({ simfile: synthDir });
```
skorfmann, line 250: just `simulator` or `simulatorTesting` - what do you think? /cc @eladb

libs/wingsdk/src/testing/assertions.ts
```diff
@@ -0,0 +1,31 @@
+import { InflightClient } from "../core";
+
+/**
+ * The Assert class provides methods for making assertions in tests,
+ * such as comparing two strings for equality.
+ * @inflight
+ */
+export class Assert {
```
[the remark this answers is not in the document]
skorfmann: would you name it `assertions` or `testing`?

examples/tests/sdk_tests/testing/string-expecations.w
```diff
@@ -0,0 +1,7 @@
+bring util;
+bring testing;
+
+test "inflight base64" {
+    testing.Expectations.expect("hello").toEqual("hello");
+    testing.Expectations.expect(1).totoBeGreaterThan(0);
```
skorfmann, line 6: the string expectation works, but the numeric one fails

```
error: Expected type to be "str", but got "num" instead
  --> /Users/sebastian/projects/wing-cloud/wing/examples/tests/sdk_tests/testing/string-expecations.w:6:33
  |
6 |     testing.Expectations.expect(1).totoBeGreaterThan(0);
  |                                 ^ Expected type to be "str", but got "num" instead


error: Unknown symbol "totoBeGreaterThan"
  --> /Users/sebastian/projects/wing-cloud/wing/examples/tests/sdk_tests/testing/string-expecations.w:6:36
  |
6 |     testing.Expectations.expect(1).totoBeGreaterThan(0);
  |                                    ^^^^^^^^^^^^^^^^^ Unknown symbol "totoBeGreaterThan"
```

skorfmann:
Since generic function overloading is not working at the moment, I'm curious in which direction assertions should go. Thoughts?

## 1. Pure Functions

All possible matchers at the same level, no scoping.

```
assertStrEqual("a", "a")
assertStrIncludes("asflkasf", "asflkasfasflkasf")
...
assertNumEqual(1,1)
...
assertFnThrows(fn, new Error())
```

## 2. Chained Objects

Type specific high level "entry point" functions (e.g. `expectStr`) which return an object with all possible matchers for that specific type.

```
expectStr("a").toEqual("a")
expectStr("asflkasfasflkasf").toInclude("asflkasf")
...
expectNum(1).toEqual(1)
...
expectFn(fn).toThrow(new Error())
```

examples/tests/sdk_tests/testing/assert.w
```diff
@@ -0,0 +1,62 @@
...
+testing.Assert.notEqual(b, "hello");
+testing.Assert.notEqual(b, true);
+
+test "equal num" {
+    testing.Assert.equal(1, 1);
+    testing.Assert.equal(b, 1);
+    testing.Assert.isNil(a);
+    testing.Assert.notEqual(b, 2);
+    testing.Assert.notEqual(b, "hello");
+    testing.Assert.notEqual(b, true);
```
skorfmann, line 22: no type safety, but I think much nicer than `strEqual`, `numEqual`, ...

examples/tests/sdk_tests/testing/assert.w
```diff
@@ -0,0 +1,62 @@
+bring util;
+bring testing;
+
```
skorfmann, line 3: would prefer to alias as `bring testing as t;` but this doesn't work right now https://github.com/winglang/wing/issues/4032

skorfmann (commented):
Played around with various type definitions for the assertions and was surprised that `unknown` is accepted by the Wing compiler. 

Allows  `testing.Assert.notEqual(1, "hello");` which is kind of what I'd expect from an assertions library. Flagged the relevant parts with comments in the code.

Don't know how this would show up in intellisense.

What do you think?

apps/wing-console/console/server/src/utils/createRouter.ts
```diff
@@ -1,5 +1,5 @@
 import { initTRPC } from "@trpc/server";
-import { testing } from "@winglang/sdk";
+import { simulatorTesting } from "@winglang/sdk";
```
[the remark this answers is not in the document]
skorfmann: done

libs/wingsdk/src/testing/assertions.ts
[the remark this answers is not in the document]
skorfmann: Is that in lieu of named imports, e.g. `bring { Assert } from testing` or generally a preferences for more granular packages over monolithic packages?

e.g.

```
bring assertions;
bring mocks;
bring testing;
```

vs 

```
bring testing

testing.Assert.equal
...
```

mostly trying to understand the reasoning here and also wondering how this would read best in the code without an alias:

```
bring assertions

assertions.equal
assertions.expectEqual
assertions.isEqual
... ?
```

examples/tests/sdk_tests/testing/assert.main.w
```diff
@@ -0,0 +1,62 @@
+bring util;
+bring testing;
```
eladb, line 2: it kind of feels like this module needs to be called `assert` and then `assert.equal()`, but then we might need to make sure the compiler still treats `assert()` as a builtin function.

@Chriscbr wdyt?

Chriscbr: you mean something like "if `blah` refers to a namespace, then `blah(xyz)` desugars into `blah.Util.default(xyz)`"?

skorfmann:
How about we split this PR in 2 parts:

1. renaming from testing -> simulator  (assuming that's something we'd wanna do anyway)
2. adding assertions 

Keeping this PR open for longer will likely cause frequent merge conflicts which'll need to be resolved.

examples/tests/sdk_tests/testing/assert.main.w
eladb: Yes, one off situation for `assert`.

My ideal name for this module is `assert` but I also want `assert()` to be a built-in function.

libs/wingsdk/src/testing/assertions.ts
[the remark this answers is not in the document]
eladb: I think `assert` will read best:

```js
bring assert;

assert(x == y);
assert.equals(x, y);
assert.throws(() => {...}, "my error");
```

Can we make this happen?

examples/tests/sdk_tests/testing/assert.main.w
Chriscbr: (personal opinions disclaimer)

> Yes, one off situation for `assert`.

I'm not sure I feel the same about the value of one off exceptions like this.

I think it can be a really healthy design constraint to try and build as much of a language's standard library only using the capabilities that are available in userland. If being unable to have an identifier represent both a namespace and a function is a language capability that you feel is _necessary_ for building Wing-idiomatic libraries and APIs, then IMO it's also a capability we should give to users -- and we should figure out how to design that. (Otherwise, we're making Wing libraries second-class citizens). If the exception doesn't represent an essential capability for building Wing-idiomatic APIs, then it would be preferable to change the name of `assert` (either the built-in function or the SDK module).

> My ideal name for this module is assert but I also want assert() to be a built-in function.

It's not clear to me this is a better DX. Learning programming languages is hard for a lot of people, and saying that `assert` can be both a function and a namespace feels like something that would be surprising to many users, and make Wing harder to learn. (What is the type of `assert` in Wing's type system -- when you mouse over `assert`)? Even though we're spending a lot of time designing Winglang, I think it's important to design the language with empathy for people that can't / don't want to maintain all of the exceptions and edge cases of a language in their head, since this makes it harder to pick up the language and get productive in a new or existing Wing codebase.

I'd compare this a little bit to operator overloading -- this is a capability that many languages add because of the ergonomic benefits (why write `bigInt1.add(bigInt2)` when you can let the user write `bigInt1 + bigInt2`?) -- but it's also _avoided_ in many languages because of the ambiguity it creates for people reading code (it can be hard to determine what function the "+" will call -- see https://journal.stuffwithstuff.com/2010/12/31/rethinking-user-defined-operators/). It also reminds me a bit of the weird oddities of JavaScript, like how classes are also functions `new Foo()` does something different from `Foo()`.

skorfmann:
> How about we split this PR in 2 parts:
> 
> 1. renaming from testing -> simulator  (assuming that's something we'd wanna do anyway)
> 2. adding assertions

Done

examples/tests/sdk_tests/testing/assert.main.w
skorfmann: Why is `assert` a global function in the first place? both https://www.winglang.io/docs/language-reference#12-utility-functions and https://github.com/winglang/wing/issues/105 aren't providing much context.

eladb: Okay let's leave this testing.Assert for now.

I hear you folks

eladb (approved):

eladb:
Reopening. @skorfmann lets get this over the finish line?

skorfmann:
> Reopening. @skorfmann lets get this over the finish line?

Yes!

> Okay let's leave this testing.Assert for now.

So, we'll go with the current `testing.Assert.xyz` style, correct?

libs/wingsdk/src/testing/assert.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Compares two values for equality.
+   * @param expected The expected value.
+   * @param actual The value to test.
+   * @throws Will throw an error if the actual value does not match the expected value.
+   * @returns void
+   */
+  public static equal(actual: unknown, expected: unknown): void {
+    nodeAssert.strictEqual(actual, expected);
+  }
```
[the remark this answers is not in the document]
skorfmann: @eladb thoughts?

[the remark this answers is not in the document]
eladb: How does that look like when used from Wing?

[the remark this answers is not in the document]
skorfmann: This looks like this https://github.com/winglang/wing/pull/3976/files#diff-00da8f3e4f26325a9d8576f796e8c8a471b2239c8ed0b1f8fe7567c5032374d2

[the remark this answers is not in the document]
eladb: This is much better than the typed versions.

examples/tests/valid/api_cors_custom.test.w
```diff
@@ -25,16 +25,16 @@ test "GET /users has cors headers" {
   let response = http.get(api.url + "/users");
 
   let headers = response.headers;
-  t.Assert.equalNum(response.status, 200);
+  testing.Assert.equal(response.status, 200);
```
eladb, line 28: Oooof, this is so cumbersome `testing.Assert.xxx()`...

I would love to find a good name for this module so we can avoid the class. I understand @Chriscbr that you are not comfortable with special casing so we can use `assert.equals()`, and I understand...

Some ideas:

```js
assertions.equal(x, y);
check.equal(x, y);
testing.equals(x, y);
asserts.equal(x, y);
assrt.equals(x, y);
ass.equals(x, y); // couldn't resist
tst.equals(x, y);
```

Any other ideas?

garysassano: > ass.equals(x, y); // couldn't resist

:laughing:

I would use `equal` over`equals`, and avoid the third person form in general.

Additionally, will there be a universal `not` for all assertions?

```js
assert.not.equal(x, y);
assert.notEqual(x, y);
!assert.equal(x, y);
```

Is it feasible to adopt Vitest's syntax?

```js
expect(x).toEqual(y);
expect(x).toStrictEqual(y);
expect(x).not.toEqual(y);
```

skorfmann: > Is it feasible to adopt Vitest's syntax?

To some extent I think yes, but it wouldn't be possible to have this typed due to no support of generics in wing. 

Haven't tried the "not" part yet.

Having said that, I think the general sentiment was that "assert" style is preferred to "expect". 

It would solve the naming conflicts though 😀

eladb: How about `expect` as the name of the module?

```js
bring expect;

expect.equals(x, y);
```

skorfmann: > bring expect;
>
> expect.equals(x, y);

I'd sort of expect `expect` to behave similar to vitest / jest expectations, e.g. `expect(1).toEqual(1)`. 

I'm still wondering if this entire discussion would be obsolete if there was either of the following

1. A way to alias this https://github.com/winglang/wing/issues/4032 e.g. `bring "testing" as t ; t.assert.equals` or `bring "assertions" as foo`
1. Move the `assert` keyword into the `assert` module and make it something like `assert.ok` (pretty much exactly how [node is doing it](https://nodejs.org/api/assert.html), except for the [assert](https://nodejs.org/api/assert.html#assertvalue-message) alias)
1. Rename the bultin `assert` to e.g. `assert!`

eladb: Not sure how #1 will solve this because users will still need to decide on an alternative name.

#2 hurts the ergonomics of the common case which is against our design principles.

#3 is interesting. I am wondering if maybe all of our builtin functions should use `!`. @Chriscbr @yoav-steinberg @MarkMcCulloh what do you think?

yoav-steinberg: I'd avoid adding special characters to our builtins since they aren't really special in any way and I don't think there's much value in flagging them as builtins. For all the user cares they're like any other function.

MarkMcCulloh: > hurts the ergonomics of the common case which is against our design principles.

I disagree that a plain binary assert is a common case. I mean it's common in the sense that it's often available, but IMO it's not what people want most of the time. My guess is that asserting equality is the most common. I think having a simplistic global assert encourages crappily written assertions and may even hide rich assertion capabilities. Looking at the current assertions we have in the examples folder, *most of them are checking equality

image


> I am wondering if maybe all of our builtin functions should use !

I don't think we have (or should have) enough global functions to make this a useful convention. I don't have a particularly strong opinion on this I guess.
It's worth noting that `!` is not a valid identifier in js (or wing currently), so that could rear its head as an issue in the implementation.

skorfmann: > #2 hurts the ergonomics of the common case which is against our design principles.

Is there a common case beyond testing? Haven't found reasoning about it, only [this](https://github.com/winglang/wing/issues/105) and [this](https://github.com/winglang/wing/pull/1030). Also, wouldn't this apply to something like [fetch](https://www.winglang.io/docs/standard-library/http/api-reference#@winglang/sdk.http.Util.fetch) as well, since it could be consider a pretty common case as well?

Chriscbr: re: an earlier suggestion -  naming the module `asserts` or `check` sounds like a decent option to me (we haven't really established a convention that modules must be singular nouns):

```js
bring asserts;

asserts.check(fluxCapacitor.enabled(), "forgot to turn on flux capacitor"); // or `asserts.isTrue()` maybe
asserts.notEqual(fluxCapacitor.power, 0, "more power is needed");
asserts.throws(() => { ... });

// or
bring check;

check.true(fluxCapacitor.enabled(), "forgot to turn on flux capacitor");
check.notEqual(fluxCapacitor.power, 0, "more power is needed");
check.throws(() => { ... });
```

eladb: > Is there a common case beyond testing? 

Yes, defensive programming - expressing preconditions in your code.

> naming the module `asserts` or `check` sounds like a decent option

If we go in that direction, I think `expect` is actually a better name as it is more commonly used for this use case.

@skorfmann I hear what you are saying about expecting that `expect` will have a different style because of jest/vitest, but I am not sure it's a huge blocker from trying this name.

---

Let's move forward with `expect` please, and we will see how it feels. We can still change it later.

examples/tests/sdk_tests/testing/assert.main.w
```diff
@@ -0,0 +1,62 @@
...
+bring expect;
+
+// Num tests
+
+let a: num? = nil;
+let b: num? = 1;
+
+expect.equal(1, 1);
+expect.equal(b, 1);
+expect.isNil(a);
```
skorfmann, line 11: what are your thoughts about `isX` vs `x`, e.g. `expect.nil` or `expect.isNil`?

examples/tests/valid/api_cors_custom.test.w
skorfmann: Done

examples/tests/sdk_tests/testing/assert.main.w
eladb:
```js
expect.nil(bang);
```

I would go with `nil()`.

eladb: What about the reverse one? `expect.notNil(opt)` or `expect.defined(opt)`?

examples/tests/sdk_tests/testing/assert.main.w
```diff
@@ -0,0 +1,62 @@
...
+    expect.notEqual(b, "hello");
+    expect.notEqual(b, true);
+}
+
+// String tests
+
+let c: str? = nil;
+let d: str? = "hello";
+
+expect.equal("hello", "hello");
```
eladb, line 30: I like `expect`. It reads well in my mind.

eladb (approved):
This is looking great.

examples/tests/sdk_tests/expect/assert.main.w
```diff
@@ -0,0 +1,210 @@
...
+// Doesn't work, while expect.equal([1,2,3,4], [1,2,3,4]); works and the JS equivalent works too
+// 
+//ERROR: Values have same structure but are not reference-equal:
+
+// [
+//     1,
+//     2,
+//     3,
+//     4
+//   ]
```
skorfmann, line 154: Don't really know why these tests are failing. Ideas?

WeepingClown13: There is a [bug](https://github.com/winglang/wing/issues/4118) currently that fails assertions for any array returned by something. I'm not sure this happens all the time but it has happened whenever I tried something, and now it looks like yet another example.

skorfmann: Thanks for the pointer!

I'm generally ok to merge this as is, mainly wanted to flag the cases which I've found to be not comparable right now:

1. DateTime
2. Range
3. Custom classes (in particular inflight classes)

examples/tests/sdk_tests/testing/assert.main.w
skorfmann: added `notNil` as well

eladb (approved):
SHIP IT

Issue #3928, opened 2023-08-23
Preserve state inside resource interaction views


staycoolcall911:
### I tried this:

1. Created a `cloud.Api`:
```ts
bring cloud;

let api = new cloud.Api();
```
2. Clicked the Api resource in the Wing Console and edited the request body (added "hello").
3. Clicked the Root panel.
4. Clicked the Api resource again.

### This happened:

The request body text field is cleared (does not contain "hello").

### I expected this:

The request body text field to still contain "hello".
A "clear all fields" button could be added to quickly get a clean slate, but in many cases I'd prefer to keep the panel's state longer.

### Component

Wing Console

### Wing Version

0.25.33

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

polamoros:
I've updated the issue title as it's affecting all resources

PR #3981, opened 2023-08-28
feat(console): preserve state inside resource interaction views

polamoros:
Preserve inputs state for the interaction views when:
* Changing from one resource to another
* Saving the file (if the resource still exists)

https://github.com/winglang/wing/assets/5547636/b5161eae-f180-4d96-bd61-63f5f051765a

resolves https://github.com/winglang/wing/issues/3928


apps/wing-console/console/ui/src/layout/default-layout.tsx
```diff
@@ -239,6 +245,40 @@ export const DefaultLayout = ({
...
+    }
+
+    const newInstances = { ...metadataInstances };
+    for (const nodeId of Object.keys(metadataInstances)) {
+      if (!hasChild(node, nodeId)) {
+        delete newInstances[nodeId];
+      }
+    }
+    setMetadataInstances(newInstances);
+  }, [items]);
```
skyrpex, line 280: You should pass a callback inside `setMetadataInstances` or otherwise it won't correctly work since the effect depends on `metadataInstances`

polamoros: you're right!

apps/wing-console/console/server/src/router/app.ts
```diff
@@ -114,6 +114,34 @@ export const createAppRouter = () => {
           input?.showTests,
         );
       }),
+    "app.itemIds": createProcedure
```
skyrpex, line 117: Can we name it `app.nodeIds`? They are only called items in the context of a tree view, but here they are nodes.

apps/wing-console/console/server/src/router/app.ts
```diff
@@ -114,6 +114,34 @@ export const createAppRouter = () => {
...
+        const getItemIds = (item: ExplorerItem, list: string[] = []) => {
+          list.push(item.id);
+          if (item.childItems) {
+            for (const child of item.childItems) {
+              getItemIds(child, list);
+            }
+          }
+          return list;
+        };
+        return getItemIds(node);
```
skyrpex, line 143:
```suggestion
        const list = new Array<string>();
        const getItemIds = (item: ExplorerItem) => {
          list.push(item.id);
          for (const child of item.childItems ?? []) {
            getItemIds(child);
          }
        };
        getItemIds(node);
        return list;
```

I think this is more readable.

apps/wing-console/console/ui/src/layout/use-layout.tsx
```diff
@@ -105,6 +106,7 @@ export const useLayout = ({
     },
     {
       enabled: selectedItems.length > 0,
+      keepPreviousData: false,
```
skyrpex, line 109: Why this change?

polamoros: I added this to prevent the UI from displaying the metadata of the previously selected node when switching from one resource to another. `useLayout` is providing the trpc object directly, and we are using the `metadata.data` in the UI, so we need to clean it up.

We could do other things like check if it's loading I guess..

with:
https://github.com/winglang/wing/assets/5547636/f2d5334e-8e66-4574-8029-352e3ec356db

without:
https://github.com/winglang/wing/assets/5547636/1b69808a-116b-4e75-87f8-08cf3205dc6a

apps/wing-console/console/server/src/router/app.ts
polamoros: yep

apps/wing-console/console/ui/src/layout/use-layout.tsx
skyrpex: Makes sense what you're doing, although I think there's something strange. Even with preserve enabled, there's a flash where the panel is empty 🤔 

Also, when did it get so slow?

polamoros: It's 'slow' as usual, but now the metadata pane was hidden while fetching new data. 
I fixed this by keeping the previos metadata panel until the new metadata is fetched.

apps/wing-console/console/ui/src/layout/default-layout.tsx
```diff
@@ -239,6 +245,43 @@ export const DefaultLayout = ({
     ],
   );
 
+  const [currentNodeMetadata, setCurrentNodeMetadata] = useState<
```
ainvoner, line 248: isn't it just currentNodeId? ```const currentNode = selectedItems[0]``` 
 it is not the current node metadata

apps/wing-console/console/ui/src/layout/default-layout.tsx
```diff
@@ -239,6 +245,43 @@ export const DefaultLayout = ({
...
+      return instances;
+    });
+    setCurrentNodeMetadata(currentNode);
+  }, [metadata.data, selectedItems]);
+
+  useEffect(() => {
+    if (!nodeIds.data) {
+      return;
+    }
+    setMetadataInstances((instances) => {
```
ainvoner, line 275: I would add a comment describing this logic.
If I understand correctly, this cleans up all removed resources from the `metadataInstances` array, right?

apps/wing-console/console/ui/src/layout/default-layout.tsx
```diff
@@ -367,27 +410,42 @@ export const DefaultLayout = ({
...
-                          node={metadata.data.node}
-                          inbound={metadata.data.inbound}
-                          outbound={metadata.data.outbound}
-                          onConnectionNodeClick={(path) => {
-                            expand(path);
-                            setSelectedItems([path]);
-                          }}
-                        />
-                      )}
+                      {Object.values(metadataInstances).map((instance) => (
```
ainvoner, line 423: why do we need to load into the DOM everything and not only the one that is currently selected?

polamoros: That's the trick to preserve the input's state for each opened node. Loading only the selected node is what causes the inputs to be 'reset' every time you switch from one node to another.

ainvoner: but this is why we now have the instanceMetadata.data, no?
when we select a node, set also set the metadata data of this node (populate the relevant metadata fields)

polamoros: One thing is the metadata (node info) and the other the inputs the user fills in (for the api the request type, the endpoint...) 🤔.

ainvoner: wouldn't it be a better solution to have a ResourceMetadataContext in our app that will have the user input for each resource? this way, once a resource is selected, its' user input can be fetched and displayed.
I think it will be better than to have all resources metadata panels and their content on the DOM

skyrpex: If the component is removed from the DOM, we'd lose their state (which is what we really want to keep).

An alternative to this is to build a more complex set of hooks that preserve the state for us.

apps/wing-console/console/design-system/src/key-value-list.tsx
```diff
@@ -11,8 +11,14 @@ export interface KeyValueItem {
   value: string;
 }
 
-export const useKeyValueList = () => {
-  const [items, setItems] = useState<KeyValueItem[]>([]);
+export interface UseKeyValueListOptions {
+  useExternalState?: typeof useState;
```
skyrpex, line 15: I think the name `useState` makes more sense. `useExternalState` sounds like it's asking for a boolean, while `useState` is a well known method. Also, since `useExternalState` is a prop that you pass from the outside, the "external" part of the name is redundant.

apps/wing-console/console/ui/src/services/use-explorer.tsx
```diff
@@ -56,15 +57,17 @@ export const useExplorer = () => {
...
-    setSelectedNode.mutate({
-      resourcePath: "root",
-    });
-  }, [tree.data, setItems]);
+    if (!selectedNode.data || !nodeIds.data?.includes(selectedNode.data)) {
+      setSelectedNode.mutate({
+        resourcePath: "root",
+      });
+    }
+  }, [tree.data, setItems, nodeIds.data]);
```
skyrpex, line 70: I feel like this should become two different effects. Otherwise, `setItems` and `createTreeMenuItemFromExplorerTreeItem` will be called more than once if I'm not mistaken.

apps/wing-console/packages/use-persistent-state/src/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useContext,
+  useEffect,
+  useRef,
+  MutableRefObject,
+  useState,
+} from "react";
+
+const PersistentStateContext = createContext<{
+  state: MutableRefObject<Map<string, any[]>>;
+}>(undefined!);
```
skyrpex, line 15: If we do `undefined!`, at some point we may end up trying to use `createPersistentState` without using the `PersistentStateProvider`, and it will fail silently.

It's better to:

```suggestion
const PersistentStateContext = createContext<{
  state: MutableRefObject<Map<string, any[]>>;
} | undefined>(undefined);
```

And create an auxiliary `usePersistentState` method that throws if the context is undefined.

skyrpex (changes requested):

skyrpex (commented):
Can you solve the new ESLint warnings?

apps/wing-console/packages/use-persistent-state/src/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+export const PersistentStateProvider = (props: PropsWithChildren) => {
+  const state = useRef(new Map<string, any[]>());
+  return (
+    <PersistentStateContext.Provider value={{ state }}>
+      {props.children}
+    </PersistentStateContext.Provider>
+  );
+};
+
+export const createPersistentState = (prefix: string) => {
```
ainvoner, line 39: the prefix is actually the unique id for the state, right? why call it a prefix?

polamoros:
> ounter state after hot reload?

My idea was to only save user inputs and not data that is loaded from the server.

Do you think we should save the other info as well?

polamoros:
> @polamoros is it intentional not to save a counter state after hot reload? same question for bucket files that were already uploaded

My idea was to only save user inputs and not data that is loaded from the server.

Do you think we should save the other info as well?

apps/wing-console/packages/use-persistent-state/src/index.tsx
polamoros: true! it make no sense now.

fixed

skyrpex: Initially it make sense but yeah, we should name it key or id.

ainvoner:
regarding the counter - I agree there is no need to save the state.
regarding the bucket - i'm not sure. Users may want to have their already populated date in the bucket throughout hot reloading... on the other hand, if I had an issue with my date population to the bucket and now I fixed it, I will probably want to have a clean state.

skyrpex:
> > @polamoros is it intentional not to save a counter state after hot reload? same question for bucket files that were already uploaded
> 
> My idea was to only save user inputs and not data that is loaded from the server.
> 
> Do you think we should save the other info as well?

I feel like we could use the `props.useState` everywhere in the components that have that prop. They shouldn't decide whether to use it in some instances or not.

polamoros:
> regarding the counter - I agree there is no need to save the state. regarding the bucket - i'm not sure. Users may want to have their already populated date in the bucket throughout hot reloading... on the other hand, if I had an issue with my data population to the bucket and now I fixed it, I will probably want to have a clean state.

I wouldn't manage the resources state here.

We should distinguish between preserving the resource state and the view's state (user inputs).
If we want to preserve the resource state, it shouldn't be managed on the frontend. Preserving the state of resources when they are destroyed and recreated is more complex than simply storing some user inputs, I bet.

ainvoner (approved):

skyrpex (approved):

PR #3986, opened 2023-08-28
fix(sdk): connections are kept alive in api/website simulator

### LINKED ISSUE #3709: Test hangs for a few seconds for after completion
### I tried this:

Running this test with `wing test -t sim main.w`:

```js
bring cloud;
bring http;

let api = new cloud.Api();
api.get("/hello", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
  return cloud.ApiResponse {
    status: 200,
    body: "Hello, world!"
  };
});

let url = api.url;

test "call endpoint" {
  let resp = http.fetch("${url}/hello");
  assert(resp.ok == true);
}
```

### This happened:

The following text is printed within a few hundred milliseconds:

```
pass ─ add.wsim » root/env0/test:call endpoint
 
 
Tests 1 passed (1)
Test Files 1 passed (1)
Duration 0m0.38s
```

Then, 2-3 seconds passes, and the command exits.

### I expected this:

The command should exit immediately

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

Removing the `http.fetch` call causes the test to no longer hang. 

### Community Notes

MarkMcCulloh:
Fixes #3709

`.close()` keeps existing connections alive, and those connections prevent the node process from closing.

Misc:
 - https://github.com/winglang/wing/pull/3964 Caused a failure in the playwright tests. I'm not sure why it would only fail sometimes, but I don't think there's a need to wait for the loading animation, regardless.

### DIFF
--- a/apps/wing-console/console/app/test/describe.ts
+++ b/apps/wing-console/console/app/test/describe.ts
@@ -36,8 +36,6 @@ export const describe = (wingfile: string, callback: () => void) => {
 
     await page.goto(`http://localhost:${server.port}/`);
 
-    await page.getByTestId("loading-overlay").waitFor({ state: "hidden" });
-
     await page.waitForLoadState("domcontentloaded");
   });
 
--- a/libs/wingsdk/src/target-sim/api.inflight.ts
+++ b/libs/wingsdk/src/target-sim/api.inflight.ts
@@ -155,6 +155,7 @@ export class Api
   public async cleanup(): Promise<void> {
     this.addTrace(`Closing server on ${this.url}`);
     this.server?.close();
+    this.server?.closeAllConnections();
   }
 
   private addTrace(message: string): void {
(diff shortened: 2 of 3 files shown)


apps/wing-console/console/app/test/describe.ts
```diff
@@ -36,8 +36,6 @@ export const describe = (wingfile: string, callback: () => void) => {
 
     await page.goto(`http://localhost:${server.port}/`);
 
-    await page.getByTestId("loading-overlay").waitFor({ state: "hidden" });
-
```
MarkMcCulloh, line 40: @polamoros This part does not seem to work properly after https://github.com/winglang/wing/pull/3964. Removing it seems to work fine, WDYT?

polamoros: It should work without that. it was just a trick to double check that the Console ui was loaded.

polamoros (approved):

PR #4008, opened 2023-08-29
feat(compiler)!: `throw` becomes a keyword

### LINKED ISSUE #2150: Utility function `throw()` (function-like): change to `throw` (similar to `return`)
Currently the grammar for `throw()` is a function-like. We'd like to change it to be aligned with `return`, since it changes the control flow of the user's program.

So instead of:
```ts (wing)
throw("no such key: ${id}");
```
It should be:
```ts (wing)
throw "no such key: ${id}";
```

Chriscbr:
Changes the built-in function "throw" into a dedicated keyword. The rationale for this change (described in #2150) is that throwing is a primary operation that changes the control flow of your program - just like `if`, `while`, `return`, or `try`/`catch` - so it's more natural to have it as a dedicated statement. This means it will receive the same syntax highlighting treatment as `return` and `while` etc.

Closes #2150.

BREAKING CHANGE: `throw` is now a built-in keyword, instead of a global function. Use `throw "blah"` instead of `throw("blah")`

## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/docs/03-language-reference.md
+++ b/docs/docs/03-language-reference.md
@@ -516,12 +516,10 @@ log("UTC: ${t1.utc.toIso())}");            // output: 2023-02-09T06:21:03.000Z
 | Name     | Extra information                                        |
 | -------- | -------------------------------------------------------- |
 | `log`    | logs str                                                 |
-| `throw`  | creates and throws an instance of an exception           |
 | `assert` | checks a condition and _throws_ if evaluated to false    |
 
 > ```TS
 > log("Hello ${name}");
-> throw("a recoverable error occurred");
 > assert(x > 0);
 > ```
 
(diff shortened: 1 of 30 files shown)


docs/docs/03-language-reference.md
```diff
@@ -1186,6 +1184,19 @@ The loop invariant in for loops is implicitly re-assignable (`var`).
 
 ---
 
+### 2.8 throw
+
+The **throw** statement throws a user-defined exception.
```
MarkMcCulloh, line 1189: We should probably be explicit here that the thing you throw must be a `str`

docs/docs/03-language-reference.md
```diff
@@ -1186,6 +1184,19 @@ The loop invariant in for loops is implicitly re-assignable (`var`).
...
+Execution of the current function will stop (the statements after throw won't be executed), and control will be passed to the first catch block in the call stack.
+If no catch block exists among caller functions, the program will terminate.
```
MarkMcCulloh, line 1191: Should we specify what "terminate" means depending on phase? e.g. throwing in preflight causes a compilation failure.

MarkMcCulloh (approved):

docs/docs/03-language-reference.md
Chriscbr: Added

PR #4009, opened 2023-08-30
chore(compiler): store paths with Utf8PathBuf

Chriscbr:
A `str` or `String` in Rust represents a sequence of Unicode characters, while a `Path` or `PathBuf` in Rust represents a sequence of any bytes that represents a path. In practice though, non-Unicode paths are vanishingly uncommon, and there isn't a strong need to support them in Wing. (https://crates.io/crates/camino has some more details)

In this PR I've updated our Rust crates to use the `Utf8PathBuf` and `Utf8Path` types throughout. By doing so we're able to check that our paths are UTF-8 once, and then manipulate them as valid UTF-8 from there on, avoiding repeated lossy and confusing conversions like `path.to_str().unwrap()` and `path.display()`.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "89b2fd2a0dcf38d7971e2194b6b6eebab45ae01067456a7fd93d5547a61b70be"
 
 [[package]]
+name = "camino"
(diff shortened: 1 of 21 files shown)


MarkMcCulloh (approved):
Nice, I'm into it

Issue #3337, opened 2023-07-10
Jump to error doesn't go to the correct tab in VSCode


eladb:
### I tried this:

Click on a compiler error link in the blue screen

### This happened:

https://github.com/winglang/wing/assets/598796/1936e978-5fac-4b84-8303-f3dca73d1cd7

### I expected this:

To jump to the right line in my code window instead of opening a new tab

### Component

IDE Extension

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.29.1](https://github.com/winglang/wing/releases/tag/v0.29.1).

PR #4012, opened 2023-08-30
feat(console):  improve jump to error when runing in vscode

polamoros:
When opening a hyperlink in vs-code, the default behavior is to open the hyperlink in the currently focused 'Editor Group.' 

However, when we have the Console open side by side with the wing file, this default behavior causes vs-code to open the file as a new tab in the Console panel group (if the group where you have the Console is focused).

This PR adds two new endpoints: `[POST] app.openFileInEditor` and `app.openFileInEditorSubscription`
When clicking on an error link the Console emits an event named `openFileInEditor`.
Code editors like vscode (and potentially others) can subscribe to the new sub endpoint to receive these events.

Previous behaviour (using hyperlinks)

https://github.com/winglang/wing/assets/5547636/305a58ef-1f90-499e-a72d-60a00cefe2ca

New behaviour (using sub)

https://github.com/winglang/wing/assets/5547636/ae9c5d84-c70a-4160-b019-7b9144509797

Weird case when you have 425924593 groups

https://github.com/winglang/wing/assets/5547636/ec52da55-887d-4b46-84a2-8d0b566669b0

resolves https://github.com/winglang/wing/issues/3337


apps/wing-console/console/ui/src/shared/use-vscode-links.tsx
```diff
@@ -0,0 +1,51 @@
...
+    error: string,
+    className: string,
+    expanded: boolean = false,
+  ) => {
+    return error
+      .replaceAll(
+        /\B((?:[a-z]:)?[/\\]\S+):(\d+):(\d+)/gi,
+        (match, path, line, column) => {
+          const link = `vscode://file/${path}:${line}:${column}`;
+          return `<a class="${className}" href="${link}">${match}</a>`;
```
skyrpex, line 16: We can use data attributes on the element to store the path, line and columns separately. Then, in the `OpenFileInEditorButton` you only need to retrieve them instead of parsing them from the link.

apps/wing-console/console/ui/src/ui/blue-screen-of-death.tsx
```diff
@@ -16,13 +19,15 @@ export const BlueScreenOfDeath = ({
 }) => {
   const [formattedPathsError, setFormattedPathsError] = useState("");
 
+  const { createHtmlLink, onClick } = useVSCodeLinks();
```
skyrpex, line 22: Is `onClick` used?

apps/wing-console/console/ui/src/shared/use-vscode-links.tsx
```diff
@@ -0,0 +1,51 @@
...
+      )
+      .replaceAll(/(\r\n|\n|\r)/gm, expanded ? "<br />" : "\n");
+  };
+
+  return {
+    createHtmlLink,
+  };
+};
+
+export const OpenFileInEditorButton = ({ children }: PropsWithChildren<{}>) => {
```
skyrpex, line 27:
```suggestion
export const OpenFileInEditorButton = ({ children }: PropsWithChildren) => {
```

skyrpex (changes requested):

skyrpex:
By default, clicking on a FileLink should use https://github.com/yyx990803/launch-editor/tree/master, which is what Vite uses.

For the VSCode extension specifically, it should be disabled since it does more than just opening the file.

polamoros:
> By default, clicking on a FileLink should use https://github.com/yyx990803/launch-editor/tree/master, which is what Vite uses.
> 
> For the VSCode extension specifically, it should be disabled since it does more than just opening the file.

I created a new issue to add it:
https://github.com/winglang/wing/issues/4067

skyrpex:
The impl looks good to me 👍🏻

skyrpex (approved):

PR #4018, opened 2023-08-30
feat(compiler): addition and subtraction assignment operators implementation

### LINKED ISSUE #4005: Operators: addition and subtraction assignment operators (`+=` and `-=` respectively)
We'd like to support the addition assignment (`+=`) and subtraction assignment (`-=`) operators and add them to the [language reference](https://www.winglang.io/docs/language-reference#643-mathematics-operators).
```ts
test "addition/subtraction" {
  let var i = 5;
  let var j = 5;
 
  i += 3;
  // 8 is logged
  log("${i}");

  j -= 3;
  // 2 is logged
  log("${j}");
}
```
For reference: see a former PR that added the power operator to the language #1654
Small note: the operator was named "exponent" in this PR, but later renamed to "power".

wzslr321:
Fixes #4005

Implemented a `+=` and `-=` operators, as a new type of `StmtKind::Assignment`.
- [Type Check](https://github.com/winglang/wing/compare/main...wzslr321:feature/compiler/add-and-sub-assignment-ops?expand=1#diff-fe246e49d57ca72d41851e8dacc3d2ccee194881769083dd8ca508058cff13b4R1829) allows only for numbers to be used on both sides of the operator.

## Checklist

- [x] Tests added (always)


MarkMcCulloh (changes requested):

libs/tree-sitter-wing/grammar.js
```diff
@@ -607,6 +609,8 @@ module.exports = grammar({
         //['>>', PREC.SHIFT],
         //['>>>', PREC.SHIFT],
         ["??", PREC.UNWRAP_OR],
+        ["+=", PREC.ADD_ASSIGNMENT],
+        ["-=", PREC.SUBTRACT_ASSIGNMENT],
```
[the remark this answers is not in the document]
wzslr321: Great point, thank you. I had some trouble testing, therefore I didn't catch that. I will try to change it according to your request, so no weird behavior can be valid.

libs/wingc/src/ast.rs
```diff
@@ -702,6 +702,8 @@ pub enum BinaryOperator {
 	LogicalAnd,
 	LogicalOr,
 	UnwrapOr,
+	SubtractAssignment,
+	AddAssignment,
```
yoav-steinberg, line 706: I think this shouldn't be a `BinaryOperator`. It should be implemented as a new type of assignment statement.
I'd change `StmtKind::Assignment` so it includes a new enum field `AssignmentKind` of a type 'Assign/AssignIncr/AssignDecr`. Then in _type_check.rs_ for the new assignment types we need to validate not only that the source and target expressions are of the same type, but that they are also numbers.

yoav-steinberg (changes requested):

libs/wingc/src/ast.rs
wzslr321: Thank you for the review and advice, I will try to change it according to it.

libs/tree-sitter-wing/grammar.js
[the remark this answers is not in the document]
wzslr321: I changed it, it is now implemented as a new type of `StmtKind::Assignment`

libs/wingc/src/ast.rs
wzslr321: I changed it according to your advice, `StmtKind::Assignment` now includes a new enum field to specify the assignment type, the type check is also updated.

yoav-steinberg (changes requested):
Nicely done, added a few comments

libs/wingc/src/type_check.rs
```diff
@@ -3145,7 +3147,18 @@ impl<'a> TypeChecker<'a> {
...
+						self.validate_type(exp_type, var.type_, value);
+					}
+					AssignmentKind::AssignIncr => {
+						self.validate_type(exp_type, self.types.number(), value);
+						self.validate_type(var.type_, self.types.number(), variable);
+					}
+					AssignmentKind::AssignDecr => {
+						self.validate_type(exp_type, var.type_, value);
+					}
+				}
```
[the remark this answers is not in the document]
wzslr321: Great change, it looks much better now. I did not know about `matches!` macro in rust, thank you.

libs/wingc/src/type_check/class_fields_init.rs
```diff
@@ -20,7 +20,11 @@ impl VisitClassInit {
 impl Visit<'_> for VisitClassInit {
 	fn visit_stmt(&mut self, node: &Stmt) {
 		match &node.kind {
-			StmtKind::Assignment { variable, value: _ } => match variable {
+			StmtKind::Assignment {
+				kind: _,
```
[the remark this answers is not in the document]
wzslr321: Thank you for noticing that, I was not sure about it too, I should've asked or researched it more deeply first.

yoav-steinberg (approved):
:rocket:

staycoolcall911 (approved):
Nice! Great work @wzslr321!

libs/tree-sitter-wing/grammar.js
```diff
@@ -160,6 +162,22 @@ module.exports = grammar({
         $._semicolon
       ),
 
+    variable_assignment_incr_statement: ($) =>
+      seq(
+        field("name", alias($.reference, $.lvalue)),
+        "+=",
+        field("value", $.expression),
```
Chriscbr, line 169: There's some duplication here - could we combine these? e.g.

```
    variable_assignment_statement: ($) =>
      seq(
        field("name", alias($.reference, $.lvalue)),
        field("operator", $.assignment_operator),
        field("value", $.expression),
        $._semicolon
      ),
   
   assignment_operator: _ => choice("=", "+=", "-="),
```

wzslr321: Good catch, this `grammar.js` syntax is new to me, so I settled with a way that worked, which turned out not to be the best - yours looks much better. I will look into it later in the day, thank you :)

wzslr321: I have tried to combine it, but it seems like it can not be translated properly this way. The compiler does not differentiate between assignments - it treats all - `=`, `+=`, `-=` -  as `=`, therefore tests fail.

I am trying to deep deeper and maybe find some work around.

wzslr321: I modified `grammar.js` the same way as you suggested. In order for it to work I also had to modify [parser](https://github.com/winglang/wing/pull/4018/files#diff-f8d8b2a32f3fb730f7c3e76c059919266d79127c00c57c1eb98599f6ae284da1R441) and adjust [statements.txt](https://github.com/winglang/wing/pull/4018/files#diff-20471e3433dc0ea2e3af17c3568071e229b05d5eb15516a3deab142e64b351e6R91).

Chriscbr (approved):

libs/tree-sitter-wing/grammar.js
Chriscbr: Sweet

PR #4034, opened 2023-08-31
fix(sdk)!: change Json.get() and Json.getAt() to throw

garysassano:
Currently `Json.get()` and `Json.getAt()` don't throw any error if the JSON property you are accessing doesn't exist.

Besides, the descriptions of the methods say that they can return `undefined`, which is not true: it's not specified in the return values and that's actually the scope of `Json.tryGet()` and `Json.tryGetAt()`.

- [x] Fix `Json.get()`
- [x] Fix `Json.getAt()`


libs/wingsdk/src/std/json.ts
```diff
@@ -146,12 +146,13 @@ export class Json {
   private constructor() {}
 
   /**
-   * Returns a specified element from the Json.
```
revitalbarletz, line 149: Do we want to keep the existing implementation and add TryGet()?

garysassano: `Json.tryGet()` and `Json.tryGetAt()` already exist.

libs/wingsdk/src/std/json.ts
```diff
@@ -146,12 +146,13 @@ export class Json {
...
+   * Returns the value associated with the specified Json key
    *
-   * @macro ($self$)[$args$]
+   * @macro ((obj, args) => { if (obj[args] === undefined) throw new Error("Json property does not exist"); return obj[args] })($self$, $args$)
```
hasanaburayyan, line 151:
```suggestion
   * @macro ((obj, args) => { if (obj[args] === undefined) throw new Error(`Json property "${args}" does not exist`); return obj[args] })($self$, $args$)
```

hasanaburayyan:
@garysassano also I believe we are going to want to indicate this as a breaking change

hasanaburayyan (approved):

PR #4037, opened 2023-08-31
rfc: iterators

### LINKED ISSUE #4036: Pagination: Iterator type / protocol / interface
### Feature Spec

We would like to have a way to to model APIs that let you iterate over some number of items, but without having to necessarily load all of the items in memory.

### Use Cases

For example, `bucket.list()` should return `Iterator<string>` instead of `Array<string>`. 

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

Chriscbr:
Proposal for a spec update to add support for a dedicated Iterator type in the language.

Closes #4036

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/docs/03-language-reference.md
+++ b/docs/docs/03-language-reference.md
@@ -1041,29 +1041,14 @@ for x in arr.iter() {
 > Note: `Array<T>` does not implement the `Iterator<T>` interface directly
 > since doing so would mean an array could not be iterated over more than once.
 
-Custom iterators can be defined by implementing the `Iterator<T>` interface:
-
-```TS
-class MyIterator implements Iterator<num> {
-  private i: num = 0;
-
-  next(): num {
-    this.i = this.i + 1;
-    return this.i;
-  }
-
-  hasNext(): bool {
-    return this.i < 10;
-  }
-}
-
-let it = new MyIterator();
-for x in it {
-  log(x); // prints 1 through 10
-}
-```
-
 [`▲ top`][top]
+
+#### 1.9.1 Roadmap
+
+The following features are not yet implemented, but we are planning to add them in the future:
+
+* generator functions, which return `Iterator<T>` - see issue #XYZ (TODO)
+* custom iterator implementations (e.g. `class MyIterator implements Iterator<num>`) - see issue #XYZ (TODO)
 
 ---
 
--- a/docs/docs/03-language-reference.md
+++ b/docs/docs/03-language-reference.md
@@ -978,7 +978,96 @@ In the presence of `catch` the variable holding the exception (`e` in the exampl
 
 ---
 
-### 1.9 Recommended Formatting
+### 1.9 Iterators
+
+The language supports iteration through the built-in `Iterator<T>` type.
+An iterator represents a sequence of elements of type `T` that may be lazily evaluated.
+Iterators are handy for working with large sequences of data without having to store the entire sequence in memory.
+
+Every iterator must satisfy the following interface:
+
+```TS
+interface Iterator<T> {
+  next(): T;
+  hasNext(): bool;
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/docs/03-language-reference.md
```diff
@@ -978,7 +978,81 @@ In the presence of `catch` the variable holding the exception (`e` in the exampl
...
+assert(it.hasNext() == true);
+assert(it.next() == 1);
+assert(it.hasNext() == true);
+assert(it.next() == 2)
+assert(it.hasNext() == false);
+```
+
+If `next()` is called when `hasNext()` returns `false`, then a runtime error will be thrown.
+
+Many types in Wing implement the `Iterator<T>` interface.
```
eladb, line 1010:
```suggestion
Many types in Wing implement the `IIterator<T>` interface.
```

docs/docs/03-language-reference.md
```diff
@@ -978,7 +978,81 @@ In the presence of `catch` the variable holding the exception (`e` in the exampl
...
+### 1.9 Iterators
+
+The language supports iteration through the built-in `Iterator<T>` type.
+An iterator represents a sequence of elements of type `T` that may be lazily evaluated.
+Iterators are handy for working with large sequences of data without having to store the entire sequence in memory.
+
+Every iterator must satisfy the following interface:
+
+```TS
+interface Iterator<T> {
```
eladb, line 990:
```suggestion
interface IIterator<T> {
```

docs/docs/03-language-reference.md
```diff
@@ -978,7 +978,81 @@ In the presence of `catch` the variable holding the exception (`e` in the exampl
...
+
+The `for` statement can be used to iterate over an iterator:
+
+```TS
+for x in it {
+  log(x);
+}
+```
+
+If an object has a method named `iter()` that returns an `Iterator<T>`, then it will automatically be used by the `for` statement.
```
eladb, line 1026: Nice!

Should we have an `IIterable<T>` as well?

docs/docs/03-language-reference.md
eladb, line 990: Is this compatible with JavaScript? Let's make sure they are seamlessly interoperable

eladb (changes requested):

docs/docs/03-language-reference.md
Chriscbr: Sort of. The way it works in JavaScript is a little bit different -- here's a repl session I whipped up that shows how iteration works:

```
$ node
Welcome to Node.js v18.16.0.
Type ".help" for more information.
> let x = [1, 2, 3]
undefined
> let it = x[Symbol.iterator]();
undefined
> it.next()
{ value: 1, done: false }
> it.next()
{ value: 2, done: false }
> it.next()
{ value: 3, done: false }
> it.next()
{ value: undefined, done: true }
```

My thinking was that in Wing we can provide a more ergonomic API for iteration, where you can obtain individual values or check if the iteration is finished directly.

For interop with JavaScript, I was imagining that given a Wing object that implements the `IIterator<T>` API, we could convert it into the corresponding JavaScript iterator so it will still work in JavaScript contexts that expect the JavaScript iteration scheme -- and given any JavaScript iterator, when it's lifted into Wing, we could converted into a form with extra `step()` and `hasNext()` methods so it can be iterated in Wing-idiomatic contexts.

(as I write this down, I realize a POC would be good to do first)

Alternatively, I could see it making sense just to model the JavaScript API directly in Wing and call it a day.

eladb: If the JS API is good enough, I think there's value in aligning to save glue and special casing. We would also want to support generator functions at some points, so there's an interplay with these APIs there as well.

garysassano:
Keep.

Issue #3790, opened 2023-08-11
Json: schema validation expose schema() object


hasanaburayyan:
##### 1.1.4.7 Schemas
Structs have a `schema` static method which returns a `JsonSchema` object (P2):
```js
let schema = Contact.schema();
schema.validate(j);
```

PR #4040, opened 2023-09-01
fix(compiler): cannot use fromJson on jsii imported struct

hasanaburayyan:
## Summary
So this is actually quite a substantial refactor to how we generate jsonschemas for structs. Previously anytime we encountered a user defined wing struct we generated a json schema file with a schema and fromJson methods. However this meant that we were not created schema files for structs defined outside of the wing code (I.E. JSII imported). With the new changes I am taking a lazy creation approach where we only generate schemas for structs if there are references that would require the schema to exist.

This change also no longer generates a separate file for the schema, instead it creates an instance of a stdlib.std.StructSchema using the jsonschema inline.

The other big change is the removal of the require statements and defs from schemas that were dependent on other schemas. As in take for example this Wing struct:
```js
struct A {
  val: num;
}

struct B {
  data: str;
}

struct C extends A {
  b: B;
}
```

The struct schema for `C` used to look like this:
```js
{
        id: "/C",
        type: "object",
        properties: {
          ...require("./A.Struct.js")().jsonSchema().properties,
          b: { "$ref": "#/$defs/B" },
        },
        required: [
          "b",
          ...require("./A.Struct.js")().jsonSchema().required,
        ],
        $defs: {
          "B": { type: "object", "properties": require("./B.Struct.js")().jsonSchema().properties },
          ...require("./A.Struct.js")().jsonSchema().$defs,
        }
      }
```

Now it looks like this:
```js
{
        id: "/C",
        type: "object",
        properties: {
          b: {
            type: "object",
            properties: {
              data: { type: "string" },
            },
            required: [
              "data",
            ]
          },
          val: { type: "number" },
        },
        required: [
          "b",
          "val",
        ]
      }
```

### Implementation Notes
1. Lazy creation of schemas, only emit struct schema files if there exists a reference to the struct type in the code.
2. Schemas no longer consist of require statements for dependent schemas. As in the schemas are now fully self contained.
3. Pre-Jsification parse of the ast to emit struct files for referenced schemas.

Closes: https://github.com/winglang/wing/issues/3943
Closes: https://github.com/winglang/wing/issues/3790

## Checklist

- [x] Tests added (always)


hasanaburayyan:
@MarkMcCulloh @Chriscbr @yoav-steinberg Still have a bug to squash with 2 namespaces that have the same struct def (i.e. A.SomeStuct and B.SomeStruct) will cause collisions. But wanted to get some thoughts on these changes. The description has some notes about the big ideas with these changes. If you get a chance to review, please rip it apart 😁 

Im planning to address the namespace issue tomorrow.

libs/wingc/src/files.rs
```diff
@@ -38,7 +38,28 @@ impl Error for FilesError {}
 
 #[derive(Default)]
 pub struct Files {
-	data: HashMap<Utf8PathBuf, String>,
+	pub data: HashMap<Utf8PathBuf, FileData>,
+}
+
+pub struct FileData {
```
hasanaburayyan, line 44: Still iffy about this

libs/wingc/src/struct_schema.rs
```diff
@@ -0,0 +1,60 @@
+use crate::{
+	ast::{Reference, Scope},
+	jsify::{JSifier, JSifyContext},
+	type_check::{is_udt_struct_type, resolve_user_defined_type, Struct},
+	visit::{self, Visit},
+	visit_context::VisitContext,
+};
+
+pub struct StructSchemaVisitor<'a> {
```
hasanaburayyan, line 9: Maybe this can be a more general phase like "pre-emit-visitor" or something along those lines. Naming is also hard

Chriscbr: Sounds like a pre-optimization to me. 🤓 Keeping this focusing around a single purpose (struct schemas) feels like the right call - if we end up with multiple pre-jsify passes, then we can abstract things when we get there

libs/wingc/src/jsify.rs
```diff
@@ -655,107 +656,105 @@ impl<'a> JSifier<'a> {
...
+		let mut code = CodeMaker::default();
+		code.open("required: [");
+		for (field_name, (_stmt_idx, kind)) in env.symbol_map.iter() {
+			if !matches!(*kind.as_variable().unwrap().type_, Type::Optional(_)) {
+				code.line(format!("\"{}\",", field_name));
+			}
+		}
+		code.close("]");
 		code
 	}
```
Chriscbr, line 681: I see there's a lot of `jsify_struct_*` methods, I wonder if it would help to corral them all in a single file like `struct_schema.rs`? It could make it easier to write unit tests for individual functions and stuff

libs/wingc/src/jsify.rs
```diff
@@ -655,107 +656,105 @@ impl<'a> JSifier<'a> {
...
-				self.jsify_struct_field_to_json_schema_type(&field.member_type.kind, ctx)
+				"{}: {},",
+				field_name,
+				self.jsify_struct_schema_field(&kind.as_variable().unwrap().type_, ctx)
 			));
 		}
+		code
+	}
 
+	pub fn jsify_struct_schema_required_fields(&self, env: &SymbolEnv) -> CodeMaker {
```
Chriscbr, line 671: `self` doesn't need to be included as a parameter if it's not being used

libs/wingc/src/files.rs
Chriscbr: Yeah this does feel a bit weird. I'm not a fan of exposing `data` as a public field, since that means `Files` is no longer well encapsulated (anyone can mutate the hashmap as they please, regardless of what contracts its methods are trying to enforce)

tools/hangar/__snapshots__/test_corpus/valid/struct_from_json.w_compile_tf-aws.md
```diff
@@ -629,10 +541,17 @@ const $plugins = ((s) => !s ? [] : s.split(';'))(process.env.WING_PLUGIN_PATHS);
...
 const externalStructs = require("./preflight.structs-1.js")({ $stdlib });
 class $Root extends $stdlib.std.Resource {
   constructor(scope, id) {
     super(scope, id);
+    const Foo = require("./Foo.Struct.js")($stdlib.std.Struct);
+    const Bar = require("./Bar.Struct.js")($stdlib.std.Struct);
+    const BucketProps = require("./BucketProps.Struct.js")($stdlib.std.Struct);
+    const Student = require("./Student.Struct.js")($stdlib.std.Struct);
+    const MyOtherStruct = require("./MyOtherStruct.Struct.js")($stdlib.std.Struct);
+    const Foosible = require("./Foosible.Struct.js")($stdlib.std.Struct);
```
MarkMcCulloh, line 554: The order that these are emitted does not seem deterministic (see all the mutations on this PR)

MarkMcCulloh: I didn't check the code yet, but make sure any structures you're iterating over to jsify use a data structure with deterministic iteration (like BTreeMap over HashMap)

libs/wingc/src/jsify.rs
hasanaburayyan: Oh my gewdness.. I thought it was required, but I guess thats just leftover Python-isms TIL

hasanaburayyan: Actually it seems like it does have to be included in oder to be callable from other methods within itself

libs/wingc/src/files.rs
hasanaburayyan: Yea I changed this back to private and put all the work in the impl.

libs/wingc/src/struct_schema.rs
hasanaburayyan: I think you are right. Thanks for the sanity check!

hasanaburayyan:
ironed out the namespaces bugs, and did some cleanups should be ready for review now

libs/wingc/src/files.rs
```diff
@@ -38,33 +38,73 @@ impl Error for FilesError {}
 
 #[derive(Default)]
 pub struct Files {
-	data: HashMap<Utf8PathBuf, String>,
+	data: BTreeMap<Utf8PathBuf, FileData>,
+}
+
+pub struct FileData {
+	content: String,
+	is_struct_schema: bool,
```
yoav-steinberg, line 46: Once we output a 3rd file type this might become messy. Maybe a `FileKind` enum?

libs/wingc/src/jsify.rs
```diff
@@ -208,13 +210,29 @@ impl<'a> JSifier<'a> {
...
+	fn jsify_struct_schema_constants(&self) -> CodeMaker {
+		let mut code = CodeMaker::default();
+		if let Some(struct_files) = self.output_files.borrow().get_struct_files() {
+			for (path, _data) in struct_files {
+				let struct_name = struct_name_from_filename(&path.to_string());
+
+				code.line(format!(
+					"const {} = require(\"{}\")($stdlib.std.Struct);",
+					struct_name,
+					path.to_string()
```
yoav-steinberg, line 229: nit
```suggestion
					"const {struct_name} = require(\"{path}\")($stdlib.std.Struct);"
```

libs/wingc/src/jsify.rs
```diff
@@ -1667,6 +1507,10 @@ fn struct_filename(s: &String) -> String {
 	format!("./{}.Struct.js", s)
 }
 
+fn struct_name_from_filename(s: &String) -> String {
```
yoav-steinberg, line 1510: Do Windows file systems today still have the silly property of being case insensitive? Do we really want to have this dependency between file name and struct name? One alternative would be adding some [id](https://json-schema.org/understanding-json-schema/structuring.html#id) into the schema file which translates into a struct name.

libs/wingc/src/lib.rs
```diff
@@ -359,6 +361,16 @@ pub fn compile(
 		return Err(());
 	}
 
+	// -- PRE-EMIT PHASE --
```
yoav-steinberg, line 364: Maybe a comment or more descriptive name of what this phase **does**.

libs/wingc/src/struct_schema.rs
```diff
@@ -0,0 +1,185 @@
...
+	fn visit_reference(&mut self, node: &'a crate::ast::Reference) {
+		match node {
+			Reference::InstanceMember { .. } => {
+				visit::visit_reference(self, node);
+			}
+			Reference::TypeMember { type_name, .. } => {
+				if is_udt_struct_type(type_name, self.ctx.current_env().unwrap()) {
+					let type_ = resolve_user_defined_type(type_name, self.ctx.current_env().unwrap(), 0);
+					let struct_code = self.jsify_struct(type_.unwrap().as_struct().unwrap());
+					self.jsify.emit_struct_file(
```
yoav-steinberg, line 161: Is `emit_struct_file` only called from here? If so maybe move it out of `jsify` and drop the jsifier dependency for this visitor. (Only have a `&types` dependency?)

libs/wingc/src/struct_schema.rs
yoav-steinberg, line 1: It seems like the methods here do 2 things:
1. Jsify the struct.
2. Write an inline schema into the struct.
One minor thing that might make this clearer is to rename the functions that output a schema from `jsify...` to `get_schema_for...`. While keeping the functions that actually write JS code with the `jsify...` prefix.
Another more difficult refactor would be to **only** generate the **schemas** in this visitor and move the jsifying code back _jsify.rs_, at that phase we can dump the boilerplate code in `jsify_struct_schema` and insert the relevant pre-generated schema into it. This can happen in the jsifying code each time we reach a reference to the struct type (plus some mechanism to avoid outputting the same struct js file multiple times, which I think is missing here too). 

The reason I'm bugging you about this is because it'll be cool if we keep the JS specific code localized while the schema generation code is relevant no matter what our target lang is.

hasanaburayyan: I love the feedback, I think once concern I have is how to tell jsify.rs that it needs to emit these struct_schemas. Right now the inline requires are determined based on what files were emitted from the visitor, but maybe there is a nice way to make that happen.

Ill have some time tomorrow morning to think this through clearly, but Im really intrigued by the thoughts

libs/wingc/src/jsify.rs
hasanaburayyan: Was able to remove this with recent refactor changes.

libs/wingc/src/files.rs
hasanaburayyan: I moved away from this model, now having the jsifier hold a BTree of schemas that are used to generate files.

libs/wingc/src/struct_schema.rs
hasanaburayyan: Okay so I moved all the jsifying code back to jsify and the visitor now just generates struct schema object and stores it in a referenced_struct_schemas map which jsifier will jsify each one. Let me know what you think of the changes.

hasanaburayyan:
@yoav-steinberg @Chriscbr Okay so I got in a grove today and might have gone crazy but I just did a refactor on this code. Please let me know your thoughts.

Summary:
 I no longer need to emit schema files, rather now Its creating an instance of `stdlib.std.StructSchema` from the `jsonschema` objects.

I.E. The following wing code:
```js
struct Foo {
  val: num;
  val2: str;
}

Foo.fromJson({val: 10, val2: "10"});
```

Just emits this into the prefilight.js
```js
const Foo = $stdlib.std.Struct._createStructSchema({id:"/Foo",type:"object",properties:{val:{type:"number"},val2:{type:"string"},},required:["val","val2",]});
```

What I like about this is I no longer have to worry about file names and require calls. Also gets rid of a lot of boiler plate. 

Also I added the implementation for being able to call `.schema()` on a struct as such:
```js
struct Foo {
  val: num;
  val2: str;
}

let schema = Foo.schema();
schema.validate({val: 10, val2: "10"});
log(schema.asStr());
```

libs/wingc/src/struct_schema.rs
```diff
@@ -0,0 +1,148 @@
...
+			code.line(format!(
+				"{}: {},",
+				field_name,
+				self.get_struct_schema_field(&kind.as_variable().unwrap().type_, env)
+			));
+		}
+		code
+	}
+
+	fn get_struct_schema(&self, struct_: &Struct) -> CodeMaker {
```
Chriscbr, line 96: Is it possible to unit test this function in Rust somehow?

libs/wingc/src/jsify.rs
```diff
@@ -1431,13 +1293,23 @@ impl<'a> JSifier<'a> {
...
 
-	fn emit_struct_file(&self, name: &Symbol, struct_code: CodeMaker, _ctx: &mut JSifyContext) {
-		let mut code = CodeMaker::default();
-		code.add_code(struct_code);
+	pub fn add_referenced_struct_schema(&self, struct_name: String, schema: CodeMaker) {
+		let mut struct_schemas = self.referenced_struct_schemas.borrow_mut();
+		struct_schemas.insert(struct_name, schema);
+	}
+
+	pub fn emit_struct_file(&self, name: &String, struct_code: &CodeMaker) {
```
Chriscbr, line 1301: is this function still needed?

libs/wingc/src/jsify.rs
```diff
@@ -1039,20 +912,9 @@ impl<'a> JSifier<'a> {
...
-				code.line(format!(
-					"const {} = require(\"{}\")({}.std.Struct);",
-					name,
-					struct_filename(&name.name),
-					STDLIB
-				));
-				code
+			StmtKind::Struct { .. } => {
+				// Struct schemas are emitted before jsification phase
+				CodeMaker::default()
```
Chriscbr, line 917: 👍

examples/tests/valid/struct_from_json.w
```diff
@@ -1,7 +1,9 @@
-// TODO: https://github.com/winglang/wing/issues/3792
-// bring cloud;
-// let j = { public: false };
-// let x = cloud.BucketProps.fromJson(j);
+bring cloud;
+
+// JSII structs
+let j = { public: false };
+let x = cloud.BucketProps.fromJson(j);
+assert(x.public == false);
```
Chriscbr, line 6: Can we add a test that this also works in inflight scopes?

libs/wingc/src/jsify.rs
```diff
@@ -53,6 +54,8 @@ pub struct JSifier<'a> {
 	pub types: &'a mut Types,
 	/// Store the output files here.
 	pub output_files: RefCell<Files>,
+	/// Stored struct schemas that are referenced in the code.
+	pub referenced_struct_schemas: RefCell<BTreeMap<String, CodeMaker>>,
```
Chriscbr, line 58: I'm curious how this works for multi-file programs. I have a sense either the file names (or some associated counter) needs to be part of the keys, or we might need to store one of these maps per AST.

(I'm ok with leaving this for a follow up issue)

Chriscbr:
@hasanaburayyan Looks much better! The inlining solution sounds good to me as well

docs/docs/04-standard-library/02-std/api-reference.md
```diff
@@ -2679,6 +2688,64 @@ Converts a Json to a Struct, returning nil if the Json is not valid.
...
+
+---
+
+##### `asStr` <a name="asStr" id="@winglang/sdk.std.StructSchema.asStr"></a>
+
+```wing
+asStr(): str
+```
+
+Retrieve the json schema as a string.
```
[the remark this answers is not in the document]
hasanaburayyan: maybe additionally instead?

My reason for adding `asStr` was so that if I have a schema
```js
struct Foo {
	val: num;
}
let schema = Foo.schema();
```

Then I can easily just log the schema 

```js
log(schema.asStr())
```

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
hasanaburayyan: The keys are struct names, with namespaces. As in `mylib.SomeStruct`, `cloud.BucketProps`, `Foo` but yea Ill add that to the comment

docs/docs/04-standard-library/02-std/api-reference.md
```diff
@@ -2679,6 +2688,64 @@ Converts a Json to a Struct, returning nil if the Json is not valid.
...
 
+### StructSchema <a name="StructSchema" id="@winglang/sdk.std.StructSchema"></a>
+
+Struct Schema.
+
+#### Initializers <a name="Initializers" id="@winglang/sdk.std.StructSchema.Initializer"></a>
+
+```wing
+new StructSchema(schema: any);
+```
```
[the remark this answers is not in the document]
hasanaburayyan: Good call, as for the constructor though it needs to be public so Struct class can call it.

libs/wingc/src/jsify.rs
hasanaburayyan: I knew I was going to miss something when I deleted all the struct code 😁

libs/wingc/src/jsify.rs
Chriscbr: Ah cool - LGTM

Chriscbr (approved):

docs/docs/04-standard-library/02-std/api-reference.md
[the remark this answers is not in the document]
Chriscbr: Gotcha - yeah, maybe someday we'll support passing `Json` directly to `log`

libs/wingc/src/struct_schema.rs
```diff
@@ -0,0 +1,148 @@
...
+			}
+			Reference::TypeMember { type_name, .. } => {
+				if is_udt_struct_type(type_name, self.ctx.current_env().unwrap()) {
+					let type_ = resolve_user_defined_type(type_name, self.ctx.current_env().unwrap(), 0);
+					let struct_code = self.get_struct_schema(type_.unwrap().as_struct().unwrap());
+
+					// add the schema to the jsifier's referenced struct schemas
+					self
+						.jsify
+						.add_referenced_struct_schema(type_name.clone().to_string(), struct_code);
```
yoav-steinberg, line 133: I'm down with this. But would like to say that what we're doing here is assigning a schema to a type. This, again, isn't strictly related to the `JSifier`. We could have just as well, I think, added this map to `Types`. Imagine our type-system having a schema per struct regardless of the `JSifier`.

yoav-steinberg (approved):
I like this. Added a minor comment.
Didn't do an in depth review thought, this time. Hopefully good enough.

PR #4041, opened 2023-09-01
chore(build): no need to generate tree-sitter wasm as part of build

MarkMcCulloh:
Generating the .wasm for the grammar is only needed when running the tree-sitter playground, otherwise the rust build handles consuming the C files when targeting wasm32-wasi. Skipping this unnecessary work is much faster and removes a pretty big dependency from our core build path (docker, although it's still used in tests)


hasanaburayyan (approved):
🤌

PR #4042, opened 2023-09-01
feat: create astro resource

skyrpex:
Closes [#4075](https://github.com/winglang/wing/issues/4075).


skyrpex:
Got something:

image

PR #4062, opened 2023-09-03
feat(sdk): dynamodb table resource

### LINKED ISSUE #4030: Implement ex.DynamodbTable resource

eladcon:
Adds a Dynamodb table resource to the `tf-aws` and `sim` (using a docker image) targets. This allows write transactions and unstructured rows which are not supported in `ex.Table`.

Closes #4030 

## Checklist

- [X] Tests added (always)
- [X] Docs updated (only required for features)


ekeren:
I would love to use this for my CDK days demo, any chance this can be reviewed?

libs/wingsdk/src/ex/dynamodb-table.ts
```diff
@@ -0,0 +1,396 @@
...
+
+  public async updateItem(
+    key: string,
+    props?: DynamodbTableUpdateItemProps
+  ): Promise<Json> {
+    const client = await this._rawClient();
+    const result = await client.send(
+      new UpdateItemCommand({
+        TableName: this.tableName,
+        Key: { [this.primaryKey]: { S: key } },
```
skyrpex, line 299: If we're going to marshall automatically, maybe we should do it here as well:

```suggestion
        Key: marshall({ [this.primaryKey]: key }),
```

libs/wingsdk/src/target-sim/dynamodb-table.inflight.ts
```diff
@@ -0,0 +1,117 @@
...
+        {
+          AttributeName: this.primaryKey,
+          KeyType: "HASH",
+        },
+      ],
+      BillingMode: "PAY_PER_REQUEST",
+    });
+
+    // dynamodb server process might take some time to start
+    let attempNum = 0;
```
skyrpex, line 100: There's a typo: `attemptNum`. What about not shortening words? `attemptNumber`

skyrpex (changes requested):

skyrpex:
I couldn't make use of the interaction widget. I think we should use a similar design as in the `ex.Table` resource 🤔

libs/wingsdk/src/ex/dynamodb-table.ts
```diff
@@ -0,0 +1,396 @@
...
+}
+
+/**
+ * Properties for transact write item's update operation.
+ */
+export interface DynamodbTransactWriteItemUpdateProps {
+  /**
+   * The item to update.
+   */
+  readonly key: string;
```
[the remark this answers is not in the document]
skyrpex: Yes, that may work... In the end, Dynamodb wants the key as `{ S: "my_key" }` or `{ N: "123" }`.

libs/wingsdk/src/ex/dynamodb-table.ts
```diff
@@ -0,0 +1,396 @@
...
+  /**
+   * An expression that defines one or more attributes to be updated.
+   * @default undefined
+   */
+  readonly updateExpression?: string;
+  /**
+   * One or more values that can be substituted in an expression.
+   * @default undefined
+   */
+  readonly expressionAttributeValues?: Json;
```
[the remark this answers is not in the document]
skyrpex: It can be `{ S: "value" }`, `{ N: "123" }`, `{ B: true }` and not sure if there's more. But if Wing needs to support the exported type, I guess we just can't use something else than Json.

skyrpex:
> it seems that table is using typed columns, we need something that supports any json value

Sorry, I wasn't clear: makes sense for the input to be JSON, but the design looks quite different. I didn't see any of existing the rows either, but maybe I didn't create any 🤔

libs/wingsdk/src/target-sim/dynamodb-table.inflight.ts
```diff
@@ -0,0 +1,117 @@
...
+      - Make sure you have docker installed and running`);
+    }
+  }
+
+  public async cleanup(): Promise<void> {
+    // disconnect from the dynamodb server
+    this.client?.destroy();
+    // stop the dynamodb container
+    await runCommand("docker", ["rm", "-f", `${this.containerName}`]);
+  }
```
skyrpex, line 71: I wonder if the cleanup gets called during kill signals. @Chriscbr ?

If it doesn't, we may want to use something like the `death` package:

```ts
      const { default: DEATH } = await import("death");
      DEATH(() => {
        // Do all cleanups
      });
```

apps/wing-console/console/ui/src/features/dynamodb-table-interaction-view.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const onAddRow = useCallback(
+    async (row: RowData) => {
+      try {
+        await addRow(row);
+      } catch (error: any) {
+        showError(error.message);
+      }
+    },
+    [addRow, showError, table.data?.keySchema],
+  );
```
skyrpex, line 44: I wonder, why is this callback dependant on `table.data?.keySchema`?

apps/wing-console/console/ui/src/services/use-dynamodb-table.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const removeRow = useCallback(
+    async (index: number) => {
+      if (!table.data?.rows[index]) {
+        return;
+      }
+      await tableDelete.mutateAsync({
+        resourcePath,
+        data: table.data.rows[index] || {},
+      });
```
skyrpex, line 32: I'd recommend using a variable to store the row so you don't have to add `|| {}`.

 ```suggestion
      const row = table.data?.rows[index];
      if (!row) {
        return;
      }
      await tableDelete.mutateAsync({
        resourcePath,
        data: row,
      });
```

File: libs/wingsdk/src/ex/dynamodb-table.ts [hunk 8]
```diff
@@ -0,0 +1,405 @@
...
+
+  public async putItem(
+    item: Json,
+    props?: DynamodbTablePutItemProps
+  ): Promise<void> {
+    const client = await this._rawClient();
+    await client.send(
+      new PutItemCommand({
+        TableName: this.tableName,
+        Item: marshall(item),
```
Chriscbr, line 294: Does `marshall` perform validation (like telling you if there is a missing key etc.)?

libs/wingsdk/src/target-tf-aws/dynamodb-table.ts
```diff
@@ -0,0 +1,161 @@
...
+    const hashKey = Object.entries(props.keySchema).find(
+      ([_, v]) => v === "HASH"
+    );
+    if (!hashKey) {
+      throw Error("missing HASH key in `keySchema`.");
+    }
+
+    const rangeKey = Object.entries(props.keySchema).find(
+      ([_, v]) => v === "RANGE"
+    );
```
Chriscbr, line 38: Should this error handling logic be shared by both targets? Maybe put this in a separate `validateKeySchema` function (checking that all values are valid, there's no more than one HASH or RANGE, etc.)

libs/wingsdk/src/ex/dynamodb-table.ts
```diff
@@ -0,0 +1,405 @@
...
+export abstract class DynamodbTableClientBase implements IDynamodbTableClient {
+  /**
+   * @param tableName the table name.
+   * @param attributeDefinitions the table attribute definitions.
+   * @param keySchema the table key schema.
+   */
+  constructor(
+    protected tableName: string,
+    protected attributeDefinitions: Json,
+    protected keySchema: Json
```
Chriscbr, line 277: Dynamodb noob question - why does this need to be a `Json` type? is it possible to pass more than two keys? If there's only a hash key and range key, we could simplify it two parameters:
```suggestion
    protected hashKey: string,
    protected rangeKey: string | undefined,
```
This will also reduce the amount of validation we have to do etc.

libs/wingsdk/src/target-tf-aws/dynamodb-table.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  private envName(): string {
+    return `DYNAMODB_TABLE_NAME_${this.node.addr.slice(-8)}`;
+  }
+
+  private attributeDefinitionsEnvName(): string {
+    return `${this.envName()}_ATTRIBUTE_DEFINITIONS`;
+  }
+
+  private KeySchemaEnvName(): string {
```
Chriscbr, line 158:
```suggestion
  private keySchemaEnvName(): string {
```

Chriscbr (approved):
I only reviewed the stuff in the SDK but everything looks great from what I can see @eladcon 👍

skyrpex (changes requested):
- The UI should be almost identical to the Table resource, with the exception of the input. The input should be a single input that accepts JSON

Instead of this row representation:

image

It should be:

image

The list of columns can be derived from the keys of the rows that we are displaying.

- There's no feedback for invalid inputs

In the Table resource, we show this message when there's an error:

image

I think we should do the same

skyrpex (approved):

PR #4072, opened 2023-09-04
feat(sdk): cloud.Domain

### LINKED ISSUE #3951: RFC: Define domain for `cloud.Website`
### Feature Spec

Today, when setting up a website with a domain on AWS, the domain is not configured because AWS requires specific information to configure the domain on CDN (CloudFront).

My initial idea to address this is to extend the `cloud.WebsiteProps` within the `share-aws` folder and include the missing specific information.

However, this approach might be confusing for users.

@Chriscbr suggested creating a preflight class that is only available to AWS through the `bring aws` command. This preflight class represents a `Domain` and provides the following code as an example.

```ts (wing)
bring cloud;
bring aws;

let w = new cloud.Website(path: "./website", domain: "boldbytes.co");
let d = new aws.Domain(w,
  hostedZoneId: "Z111111111AAAA1A111AA",
  acmCertificateArn: "arn:aws:acm:us-east-1:111111111111:certificate/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
);
```

@eladb pointed out that domains are not specific to AWS and suggested the need for a `cloud.Domain` class. They recommended that the `cloud.Website` class should accept a `cloud.Domain` object as input. Here's the provided code example:

```ts (wing)
bring cloud;

let exampleDomain = new cloud.Domain("www.example.com") as "hello-domain";
 
new cloud.Website(
  path: "./website",
  domain: exampleDomain,
  // ...
);
```
However, when compiling to AWS (ts-aws or awscdk), you encounter the following error:
```ts (wing)
wing compile -t tf-aws main.w
ERROR: "hostedZoneId" is missing from /root/hello-domain
ERROR: "acmCertificateArn" is missing from /root/hello-domain

These are required properties of platform-specific types. You can set these values
either through `-v|--value` switches or `--values` file.
```
Now, you should be able to assign values to this:
```ts (wing)
wing compile \
  -t tf-aws main.w \
  -v /root/hello-domain.hostedZoneId=Z111111111AAAA1A111AA \
  -v /root/hello-domain.acmCertificateArn= arn:aws:ac...
```
Alternatively, a values.yaml file can be used:
```ts (wing)
/root/hello-domain:
  hostedZoneId: "Z111111111AAAA1A111AA",
  acmCertificateArn: "arn:aws:acm:us-east-1:111111111111:certificate/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"  
```
```ts (wing)
wing compile -t tf-aws main.w --values values.yaml
```

### Use Cases

The basic usage is to configure the domain when creating a website.

### Implementation Notes

I like the approach that Elad suggested.

I think we can initially create a class called `cloud.Domain` with the fields `domain` and `subDomain`. This way, we could create targets like `www.example.com` and `app.example.com`.

In the AWS-specific `Website` class, the `Domain` class would add the necessary fields (`hostedZoneId`, `acmCertificateArn`, `iamCertificate`).

I don't know how we would create a specific error linked to the target. Correct me if I'm wrong, but perhaps an annotation in the `Domain` class could indicate that it depends on the target.

### Component

SDK

### Community Notes

marciocadev:
Now, to configure a domain for your website on AWS, you need to provide the `hostedZoneId` and either the `acmCertificateArn` or `iamCertificate`.

> If the target is different from AWS, these parameters will not be required.

These values can be passed to the compiler in two ways for the code below:

```ts
// file domain.w
bring cloud;

let myDomain = new cloud.Domain(domain: "www.lazinessdevs.com") as "my-domain";

new cloud.Website(path: "./site", domain: myDomain);
```

1) By passing the values on the command line as shown in the example below:

```bash
wing compile \
-t tf-aws \
-v root/Default/Default/my-domain.hostedZoneId=Z0XXXXXXXXXXXXXXXXXXF \
-v root/Default/Default/my-domain.acmCertificateArn=arn:aws:acm:us-east-1:111111111111:certificate/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee  domain.w
```

2) By providing the path to a YAML file with the configuration:

```yaml
root/Default/Default/my-domain:
  hostedZoneId: Z0XXXXXXXXXXXXXXXXXXF
  acmCertificateArn: arn:aws:acm:us-east-1:111111111111:certificate/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee
```

Command:

```bash
wing compile -t tf-aws --values config.yaml domain.w
```

If the target is AWS and these parameters are not passed, the following error message will be displayed.

```bash
wing compile -t tf-aws domain.w 
ERROR: 
  - 'iamCertificate' or 'acmCertificateArn' is missing from root/Default/Default/my-domain
  - 'hostedZoneId' is missing from root/Default/Default/my-domain

These are required properties of platform-specific types. You can set these values
either through '- v | --value' switches or '--values' file.

target/domain.tfaws.892227.tmp/.wing/preflight.js:9
     constructor(scope, id) {
       super(scope, id);
>>     const d = this.node.root.newAbstract("@winglang/sdk.cloud.Domain",this,"my-domain",{ domain: "www.lazinessdevs.com" });
       this.node.root.newAbstract("@winglang/sdk.cloud.Website",this,"cloud.Website",{ path: "./site", domain: d });
     }
```

Closes #3951

## Checklist

- [x] Tests added (always)


skorfmann:
Another thought: Why not delegate the config management to Terraform, given that's only relevant over there? Make 

- hostedZoneId
- acmCertificateArn

Terraform [variables](https://developer.hashicorp.com/terraform/language/values/variables) and Terraform takes care of validating presence of these. Since deployments are done outside of Winglang anyway, this could simplify this quite a bit. The only part where some custom plumbing would be required is for `wing test -t tf-aws` workflows

marciocadev:
> > root/Default/Default/my-domain.hostedZoneId
> 
> There's a lot of magic involved here. How does a user know the correct path for the resource - only by reading the error message? Also, relying on the path seems to be very brittle when it comes to refactoring. Are there ideas how this could be improved (if not now, at least in following iterations)?
> 
> One thought which crossed my mind exceeding the scope of this PR:
> 
> How about allowing target specific code, something like `domain.tf-aws.w` which would be compiled for the specified target only? Sort of the reverse of what the SDK is doing to allow platform specific code. This would allow e.g. to do platform specific cert validation as well. Have no idea if that would practically work though :)

I understand what you're saying, @skorfmann , and I find the definition of the platform as an extension of the file interesting.

One of the things that would need to be implemented for this solution is for the auto-complete to also capture the platform. This way, a file with the AWS extension would have fields for `hostedZoneId` and `Certificate`, while without this extension, they wouldn't be necessary.

Could you please add this suggestion to issue #3951?

apps/wing/src/commands/compile.ts
```diff
@@ -20,6 +20,8 @@ export interface CompileOptions {
   readonly target: wingCompiler.Target;
   readonly plugins?: string[];
   readonly rootId?: string;
+  readonly value?: string;
```
eladb, line 23: can you please add some docstrings here? (I know there weren't docstrings here but maybe now is a good opportunity...)

marciocadev: I added docstrings to all the variables here, @eladb, maybe I overdid it with the `values`

apps/wing/src/cli.ts
```diff
@@ -142,6 +148,8 @@ async function main() {
     )
     .option("-p, --plugins [plugin...]", "Compiler plugins")
     .option("-r, --rootId <rootId>", "App root id")
+    .option("-v, --value <value>", "Platform-specific value", addValue, [])
```
eladb, line 151:
```suggestion
    .option("-v, --value <value>", "Platform-specific value in the form KEY=VALUE", addValue, [])
```

apps/wing/src/commands/compile.ts
```diff
@@ -17,9 +17,33 @@ const log = debug("wing:compile");
...
+  /**
+   * Path to the YAML file with specific platform values
+   * 
+   * example of the file's content:
+   * root/Default/Domain:
+   *   hostedZoneId: Z0111111111111111111F
+   *   acmCertificateArn: arn:aws:acm:us-east-1:111111111111:certificate/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee
+   */
+  readonly values?: string;
```
eladb, line 46: @hasanaburayyan this is related to the PDK work. can you please review this PR?

docs/docs/04-standard-library/01-cloud/domain.md
```diff
@@ -0,0 +1,179 @@
...
+To provide these values, there are two options. You can either pass the values in the command line as shown in the example below:
+
+```bash
+wing compile \
+-t tf-aws \
+-v root/Default/Default/cloud.Domain.hostedZoneId=Z0XXXXXXXXXXXXXXXXXXF \
+-v root/Default/Default/cloud.Domain.acmCertificateArn=arn:aws:acm:us-east-1:111111111111:certificate/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee  domain.w
+```
+
+Or you can provide a YAML file with the configurations as follows:
```
eladb, line 54: Let's move the documentation about how to provide platform values to the cli documentation and just reference the yaml option here

docs/docs/04-standard-library/01-cloud/domain.md
```diff
@@ -0,0 +1,179 @@
...
+);
+
+new cloud.Website(path: "./site", domain: domain);
+```
+
+## Target-specific details
+
+### Simulator (`sim`)
+
+Under the hood, the simulator stores the domain value in memory.
```
eladb, line 37: @eladcon feels like sim.Domain might be where we want to inject the port forwarding behavior into the sdk.

docs/docs/04-standard-library/01-cloud/domain.md
```diff
@@ -0,0 +1,179 @@
...
+    Wing language,
+    language,
+    Wing standard library,
+    Wing programming language,
+    Domain,
+  ]
+sidebar_position: 1
+---
+
+The `cloud.Domain` resource represents the domain that will be configured when creating a website in the cloud.
```
eladb, line 17: This should also be useful for other types of endpoints such as `cloud.Api`, not just websites.

libs/wingsdk/src/cloud/domain.ts
```diff
@@ -0,0 +1,61 @@
...
+export const DOMAIN_FQN = fqnForType("cloud.Domain");
+
+/**
+ * Options for `Domain`.
+ */
+export interface DomainProps {
+  /**
+   * The website's custom domain name.
+   * @example "example.com"
+   * @default - a domain is generated by the cloud provider
```
eladb, line 18: What's the use case of defining a Domain without a domain name? Feels like it should be a required field.

libs/wingsdk/src/shared-aws/domain.ts
```diff
@@ -0,0 +1,89 @@
...
+    super(scope, id, props);
+
+    const values = getPlatformSpecificValues(
+      this.node.path,
+      "iamCertificate",
+      "acmCertificateArn",
+      "hostedZoneId"
+    );
+
+    let errors: Array<String> = new Array<String>();
```
eladb, line 27: Can we move this error rendering code to the common utility function? Seems like it's going to be needed anywhere we have platform values. Maybe `getPlatformValues()` can receives some more info about each value like defaults and description and whether it's required or not.

libs/wingsdk/src/cloud/domain.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  /** @internal */
+  public _getInflightOps(): string[] {
+    return [];
+  }
+
+  /**
+   * The domain value.
+   */
+  public get domainName(): string | undefined {
```
eladb, line 58:
```suggestion
  public get domainName(): string {
```

eladb (changes requested):

libs/wingsdk/src/cloud/domain.ts
marciocadev: You're right

change to not optional

libs/wingsdk/src/shared-aws/domain.ts
marciocadev: I did what you requested, @eladb.

Now, the following call:

```ts
const values = getPlatformSpecificValues(
      this.node.path,
      "iamCertificate||acmCertificateArn",
      "hostedZoneId"
    );
```

I did what you requested, Elad.

Now, the following call:

ts

const values = getPlatformSpecificValues(
      this.node.path,
      "iamCertificate||acmCertificateArn",
      "hostedZoneId"
    );

checks if the value of `hostedZoneId` was provided and if one of the certificate values (`iamCertificate` or `acmCertificateArn`) was provided.

If these conditions are not valid, an exception is generated within the `getPlatformSpecificValues` function itself, allowing it to be reused by other resources in the future.

eladb (changes requested):
Missing .w sdk tests for the new Domain resource

eladb:
> Why not delegate the config management to Terraform, given that's only relevant over there?

That's an interesting idea, relevant more generally of course to any terraform based platform provider.

Let's keep the implementation here as is and @hasanaburayyan should consider as part of his PDK RFC.

marciocadev:
Hi @eladb 

I am trying to run a test with `cloud.Domain`, but I encountered a small issue.

When I run the test below for the simulator, it does not require the `hostedZoneId` or a value for the certificate (`iamCertificate` or `acmCertificateArn`)

```ts (wing)
bring cloud;

let d = new cloud.Domain(domainName: "example.com") as "my-domain";

test "domain name" {
  assert(d.domainName == "example.com");
}
```
I get this
```bash
$ ~/mywing test domain.main.w
pass ─ domain.main.wsim » root/env0/test:domain name


Tests 1 passed (1)
Test Files 1 passed (1)
Duration 0m0.29s
```
But when I run the same test by specifying 'tf-aws' as the target, the path varies because the tests are created uniquely. Therefore, it's challenging to create a file indicating the correct path to provide the necessary data

First execution
```bash
$ ~/mywing test -t tf-aws domain.main.w

✖ Compiling domain.main.w to tf-aws...

ERROR:
  - 'iamCertificate' or 'acmCertificateArn' is missing from root/Test.gfvqV9U482/env0/my-domain
  - 'hostedZoneId' is missing from root/Test.gfvqV9U482/env0/my-domain

These are required properties of platform-specific types. You can set these values
either through '-v | --value' switches or '--values' file.

../../../../../tmp/-wing-compile-testPNi238/target/test/domain.main.tfaws.344717.tmp/.wing/preflight.js:42
         }
       }
>>     const d = this.node.root.newAbstract("@winglang/sdk.cloud.Domain",this,"my-domain",{ domainName: "example.com" });
       this.node.root.new("@winglang/sdk.std.Test",std.Test,this,"test:domain name",new $Closure1(this,"$Closure1"));
     }



Tests 1 failed (1)
Test Files 1 failed (1)
Duration 0m0.15s
```

Second
```bash
$ ~/mywing test -t tf-aws domain.main.w                                                                                                      ─╯


✖ Compiling domain.main.w to tf-aws...

ERROR:
  - 'iamCertificate' or 'acmCertificateArn' is missing from root/Test.OcZ_aYTvRy/env0/my-domain
  - 'hostedZoneId' is missing from root/Test.OcZ_aYTvRy/env0/my-domain

These are required properties of platform-specific types. You can set these values
either through '-v | --value' switches or '--values' file.

../../../../../tmp/-wing-compile-testJAjNPL/target/test/domain.main.tfaws.346476.tmp/.wing/preflight.js:42
         }
       }
>>     const d = this.node.root.newAbstract("@winglang/sdk.cloud.Domain",this,"my-domain",{ domainName: "example.com" });
       this.node.root.new("@winglang/sdk.std.Test",std.Test,this,"test:domain name",new $Closure1(this,"$Closure1"));
     }



Tests 1 failed (1)
Test Files 1 failed (1)
Duration 0m0.15s
```

The path changes with each execution
1) `root/Test.gfvqV9U482/env0/my-domain`
2) `root/Test.OcZ_aYTvRy/env0/my-domain`

eladb (approved):
Ship it!

Issue #4067, opened 2023-09-04
open error links in the default code editor


polamoros:
### Feature Spec

Users of different code editors can open error links in their default editor (other than vs-code).

The console uses [lunch-editor](https://github.com/yyx990803/launch-editor/tree/master) to open error links in the default code editor.

<img width="960" alt="image" src="https://github.com/winglang/wing/assets/5547636/b5d5a0f1-0d26-4d7c-b387-7bdd6723bfa7">

### Use Cases

Users of code editors other than vs-code want to use error links.

### Component

Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.29.6](https://github.com/winglang/wing/releases/tag/v0.29.6).

PR #4084, opened 2023-09-05
feat(console): open error links using the active code editor

polamoros:
Use [launch-editor](https://www.npmjs.com/package/launch-editor) to open the wing file in the currently active code editor.

https://github.com/winglang/wing/assets/5547636/cda55017-ded4-4b85-80f0-875465d7227a

resolves https://github.com/winglang/wing/issues/4067


apps/wing-console/console/server/src/router/app.ts
```diff
@@ -601,12 +601,21 @@ export const createAppRouter = () => {
...
+          });
+          return;
+        }
+        const launch = require("launch-editor");
+        launch(`${input.path}:${input.line}:${input.column}`);
       }),
+
+    // Subscribing to this procedure will override the default behavior of opening files in the editor
+    // provided by the "app.openFileInEditor" procedure.
+    // Needed to manage file opening within the vs-code extension.
```
skyrpex, line 618: This should be a `/** ... */` docstring, otherwise it won't show up in the IDE when consuming the subscription.

```suggestion
    /**
     * Warning! Subscribing to this procedure will override the default behavior of opening files in the editor
     * provided by the "app.openFileInEditor" procedure.
     * Needed to manage file opening within the vs-code extension.
     */
```

apps/wing-console/console/server/src/router/app.ts
```diff
@@ -601,12 +601,21 @@ export const createAppRouter = () => {
...
-        });
+        if (ctx.emitter.listenerCount("openFileInEditor") > 0) {
+          ctx.emitter.emit("openFileInEditor", {
+            path: input.path,
+            line: input.line,
+            column: input.column,
+          });
+          return;
+        }
+        const launch = require("launch-editor");
```
skyrpex, line 612: Can we use `const launch = await import("launch-editor");` instead? Requires aren't ESM compatible, and would like to stay future proof.

skyrpex (changes requested):

apps/wing-console/console/server/src/router/app.ts
polamoros: fixed!

skyrpex (approved):

Issue #3967, opened 2023-08-25
If a self mutation happens in a merge queue, the failure is confusing and the next step is unclear


MarkMcCulloh:
Example https://github.com/winglang/wing/pull/3966

This PR failure was because main was merged into the branch, which caused a snapshot change. The self mutation was applied to the merge queue, but this is a waste of time because mergify already throws away this PR.

The next action depends on the exact circumstance, but generally will require the same update and mutation to be applied to the original PR. At the very least there should be some messaging to clarify that the mergify queue failed for this reason and some possible next steps.

staycoolcall911:
This happened a couple of times already - bumping it to p1

PR #4088, opened 2023-09-05
fix(build): missing files @winglang/sdk - cannot find module

### LINKED ISSUE #4021: pnpm compile command fails does not generate needed files if empty providers directory exists
### I tried this:

After working on a PR regarding compiler I wanted to test the entire project. 

### This happened:

It failed on stage `@winglang/sdk#compile`.

```
// one of many
@winglang/sdk:compile: 12 import { S3Bucket } from "../.gen/providers/aws/s3-bucket";
@winglang/sdk:compile:                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@winglang/sdk:compile: src/target-tf-aws/website.ts:13:32 - error TS2307: Cannot find module '../.gen/providers/aws/s3-bucket-policy' or its corresponding type declarations.
```

### I expected this:

I expected to compile properly and eventually fail on tests stage.

### Is there a workaround?

The workaround is to remove `libs/wingsdk/src/.gen/providers` directory by hand and run `pnpm pre-compile`. It then generates missing files, which were not generated, because near-empty `providers` directory existed before - it didn't check if all needed files are there.

### Component

SDK

### Wing Version

0.26.8

### Node.js Version

v18.17.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

wzslr321:
fixes  #4021

When `libs/wingsdk/src/.gen/providers` directory is not empty, it fails to generate the required files during `pre-compile` stage. The easy - but not obvious - fix is to remove *providers* directory entirely.

`pnpm pre-compile` command executed `cdktf get` command, which does not generate files if it spots the aforementioned directory.  This command has a `--force` flag, which forces to generate all the files. I changed `pre-compile` script to use it by default.

## Checklist

- [ ] Tests added (always)


libs/wingsdk/.projen/tasks.json
```diff
@@ -345,6 +345,15 @@
         }
       ]
     },
+    "pre-compile-force": {
```
MarkMcCulloh, line 348: imo I think the existing `pre-compile` task should just always run with --force.

Since we use `turbo`, you generally will not run this more than once anyways

wzslr321: Well, I did not really take into account that it is only used once, and I apparently over-thinked this. I will change it right away, thank you :)

MarkMcCulloh:
You'll have to make the change in the projenrc.ts file rather than the task.json

wzslr321:
> You'll have to make the change in the projenrc.ts file rather than the task.json

Thank you, I updated it. I definitely need yet to study how exactly is this project built and configured.

MarkMcCulloh (approved):

MarkMcCulloh:
@wzslr321 This PR needs to be updated to be merged (sorry for any confusion, it's due to https://github.com/winglang/wing/issues/3967)

PR #4092, opened 2023-09-05
chore(sdk): add `Json.entries()`

### LINKED ISSUE #3142: Json.entries()
Json static method `.entries()`

garysassano:
Closes #3142


hasanaburayyan:
Thanks @garysassano! Before we get this merged we should probably go ahead and update the language reference as well. There is a section that says: `Json.entries() - see https://github.com/winglang/wing/issues/3142 to track.` and we should instead add a section showing how to use entries (nothing big just like the other examples above it)

garysassano:
@hasanaburayyan I see documentation for `Json.entries()` is already present [HERE](https://www.winglang.io/contributing/rfcs/language-spec#11-types).

Maybe I should just remove this line?
>NOTE: values() and entries() return an array inside a Json object because at the moment we cannot represent heterogenous arrays in Wing.

hasanaburayyan:
@garysassano sorry just saw this, the link you sent is from an RFC of the language spec, which the current language spec was derived from and @staycoolcall911 removed the bits that thad not been implemented yet. So basically we should put what is in that RFC into the actual language reference

garysassano:
@hasanaburayyan docs updated, should be ready to 🚢

hasanaburayyan (approved):
🚢

libs/wingsdk/src/std/json.ts
```diff
@@ -32,6 +32,32 @@ export class Json {
...
+   * @param json to get values from
+   * @returns the values from the Json as array of Json
+   */
+  public static values(json: Json): Json[] {
+    json;
+    throw new Error("Macro");
+  }
+
+  /**
+   * Returns the entries from the Json.
```
MarkMcCulloh, line 49: Might be worthwhile to add an example to make it clear that it's an array of arrays.

libs/wingsdk/src/std/json.ts
```diff
@@ -32,6 +32,32 @@ export class Json {
...
+   *
+   * @macro (Object.entries($args$))
+   *
+   * @param json to get values from
```
MarkMcCulloh, line 53:
```suggestion
   * @param json to get entries from
```

Also, I guess we should mention here that it should be a map json and not just any arbitrary one.

libs/wingsdk/src/std/json.ts
```diff
@@ -32,6 +32,32 @@ export class Json {
...
+   * @returns the entries from the Json as array of Json
```
MarkMcCulloh, line 54: Maybe be specific about the [[key, value]] format here

libs/wingsdk/src/std/json.ts
garysassano: I actually added an example in the Library Reference: https://github.com/winglang/wing/pull/4092/files#diff-930e1858fb1950a337183b7f181fd0706fac2a9daf0e5b135b884e09cbaf6fe7R242-R251

libs/wingsdk/src/std/json.ts
garysassano: Then it should be the same also for `keys()` and `values()`. Should I change all three?

garysassano: I updated the same also for `keys()` and `values()`, can you take a look and let me know what you think?

libs/wingsdk/src/std/json.ts
garysassano: Updated the JSDoc.

libs/wingsdk/src/std/json.ts
garysassano: I updated the same also for `keys()` and `values()`. Can you take a look and let me know what you think?

PR #4098, opened 2023-09-06
feat(sdk): only generate test resources when running `wing test`

Chriscbr:
It was brought up in a recent discussion about testing that it's unusual how `test` statements create resources in cloud provider accounts even when you have only compiled your application to run in production. This PR changes the behavior so that test resources are only generated when running `wing test`, or when using the simulator target (since there's no cost of simulated resources). 

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/core/app.ts
```diff
@@ -100,6 +100,12 @@ export abstract class App extends Construct {
     }
   }
 
+  /**
+   * The name of the compilation target.
+   * @internal
+   */
+  public abstract readonly _target: string;
```
MarkMcCulloh, line 107: IMO it'd be useful to have a literal union type for this instead of just string

libs/wingsdk/src/std/test.ts
```diff
@@ -38,8 +38,12 @@ export class Test extends Resource implements IInflightHost {
     return App.of(scope).newAbstract(TEST_FQN, scope, id, inflight, props);
   }
 
-  /** @internal */
-  public readonly _fn: Function;
+  /**
+   * The function that will be called when the test is run. This will not be created
+   * if the app is not compiled as a test.
```
MarkMcCulloh, line 43: Comment is not entirely correct (will also be created for sim). Also the double negative could probably be avoided.

MarkMcCulloh (approved):
LGTM, some nits

libs/wingsdk/src/std/test.ts
Chriscbr: I also don't not dislike double negatives 😄

libs/wingsdk/src/core/app.ts
Chriscbr: I thought JSII might complain about this but I guess not 🤷 - added

PR #4105, opened 2023-09-06
feat(sdk)!: expose simulator via HTTP server

Chriscbr:
This PR evolves the Wing simulator so that every simulation spins up an HTTP server. This HTTP server accepts requests to call inflight methods on resources, making it possible for simulated resources to interact across multiple processes on the same machine (such as from containers running on your system).

Since it's not possible to serialize all forms of data as HTTP, there's a small regression where the `Redis` class's `rawClient` API no longer worked. For the time being, I've removed the method to work around the issue. We might be able to recover the functionality by writing a dedicated implementation of `toInflight` for the `sim` target's implementation of `Redis`, but it's a bit beyond the scope of this PR.

BREAKING CHANGE: The `rawClient()` method on `ex.Redis` has been removed. This API was infrequently used and was provided without any type information. If you have a use case for this API, let us know!

## Checklist

- [ ] Tests added (always)


examples/tests/valid/redis.w
```diff
@@ -18,11 +18,11 @@ queue.setConsumer(inflight (message: str) => {
...
-  // Using raw client
-  let connection = r.rawClient();
-  connection.set("wing", "does redis");
-  let value = connection.get("wing");
-  assert(value == "does redis");
+  // // Using raw client
+  // let connection = r.rawClient();
+  // connection.set("wing", "does redis");
+  // let value = connection.get("wing");
+  // assert(value == "does redis");
```
Chriscbr, line 25: Should I just remove this?

libs/wingsdk/test/target-sim/queue.test.ts
```diff
@@ -110,12 +110,7 @@ test("queue with one subscriber, batch size of 5", async () => {
     "OnDeployHandler",
     `\
 async handle() {
-  await this.queue.push("A");
-  await this.queue.push("B");
-  await this.queue.push("C");
-  await this.queue.push("D");
-  await this.queue.push("E");
-  await this.queue.push("F");
+  await this.queue.push("A", "B", "C", "D", "E", "F");
```
Chriscbr, line 113: This was to fix a flakey test

examples/tests/valid/redis.w
MarkMcCulloh: I'm good with removing the raw client. If people really want unchecked access to redis within wing, using extern would basically give the same experience (if not better)

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -86,7 +89,6 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
           timeout: this.timeout,
           context: {
             process: sandboxProcess,
-            $simulator: this.context,
```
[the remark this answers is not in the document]
Chriscbr: What kind of opt-in mechanism do you have in mind?

It's probably doable (maybe e.g. we add a parameter to `toInflight()` opt into a "network safe" version of the client), I just wasn't sure if the optimization and increased API surface area was worth it. (For what it's worth, we already spin up an HTTP server per website, API, and public bucket in your Wing app 😅). I'll double check perf numbers tomorrow.

re: big picture -- some other benefits are that this change eliminates some of the differences between the simulator clients and actual-cloud clients (they're all doing HTTP requests now, and are susceptible to latency and network errors), and it paves the way for making simulated resources available in other languages.

libs/wingsdk/src/target-sim/util.ts
```diff
@@ -50,13 +50,35 @@ export function bindSimulatorResource(
...
+    throw new Error("Missing environment variable: WING_SIMULATOR_URL");
+  }
+
+  return new Proxy({}, {
+    get: function(target, method, receiver) {
+      return async function() {
+        const resp = await fetch(simulatorUrl + "/v1/call", {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({ handle, method, args: Array.from(arguments) }),
```
skyrpex, line 71: We're subject to the JSON stringify limitations here. Some data types like BigInt, Date, etc won't work. I'd suggest using https://github.com/blitz-js/superjson instead.

libs/wingsdk/src/testing/simulator.ts
```diff
@@ -231,6 +243,67 @@ export class Simulator {
...
+    app.use(express.urlencoded({ extended: true }));
+
+    app.post("/v1/call", async (req, res, next) => {
+      try {
+        const request: SimulatorServerRequest = req.body;
+        let { handle, method, args } = request;
+
+        // when a JSON object is stringified to be sent over the wire, any occurrences of `undefined` are replaced with `null`.
+        // deeply replace all nulls with undefineds to avoid issues
+        args = JSON.parse(JSON.stringify(args), (_k, v) =>
```
skyrpex, line 261: `superjson` might help with the undefined values, too.

libs/wingsdk/test/target-sim/__snapshots__/file-counter.test.ts.snap
```diff
@@ -17,20 +17,60 @@ return class Handler {
...
+        });
+        const parsed = await resp.json();
+        if (parsed.error) {
+          throw new Error(parsed.error);
+        }
+        return parsed.result;
+      };
+    }
+  });
+})(),
```
skyrpex, line 46: Can this generated code do something like the example below?

```ts
import { createSimulatorProxy } from "@winglang/sdk";

// ...
counter: createSimulatorProxy(process.env.WING_SIMULATOR_URL),
// ...
```

Reasoning is smaller bundles, and also due to possibly importing superjson (as a dependency from @winglang/sdk). It's reasonable to require having `@winglang/sdk` installed and available inside the astro project.

skyrpex (commented):
Love it! The 15% slowdown tradeoff is a win if it enables the use cases we're looking for :)

libs/wingsdk/src/target-sim/util.ts
Chriscbr: Sounds like a good fit, I'll try it out

MarkMcCulloh: If the goal is to more closely resemble real cloud clients in the sim, should we not keep the same limitations that plain old JSON APIs force on us?

libs/wingsdk/src/target-sim/function.inflight.ts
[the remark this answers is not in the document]
MarkMcCulloh: I was thinking more of an implicit opt-in. The simulator client can check for the existence of a $simulator global, and only go with HTTP if that global isn't found.

It would be cool to keep the same API and always use JSON as the way to transport data, regardless of using HTTP or calling directly. This keeps things that matter the same imo.

> eliminates some of the differences between the simulator clients and actual-cloud clients

Arguably I think we're willing to have differences for non-functional things (e.g. latency and networking hiccups). I do think requiring serialization is a reasonable functional change keep the same between them.

libs/wingsdk/src/target-sim/util.ts
skyrpex: > If the goal is to more closely resemble real cloud clients in the sim, should we not keep the same limitations that a plain old JSON APIs force on us?

I think we should talk about this, it's important. The initial solution to this problem was to change the way sim clients are implemented: they should be aware that they're being executed in a server and the communication protocol is HTTP.

The problem is that this is not the case currently. The quick solution was to create an HTTP server as a means of an invisible proxy for other processes to interact with the sim clients that currently exist...

skyrpex: But this HTTP proxy comes with the problem that it can't serialize everything: Date, BigInt, Blob, and more importantly, Functions passed as callbacks, and other stuff that may be implemented in inflight methods.

examples/tests/valid/redis.w
hasanaburayyan: Lets do it

libs/wingsdk/src/testing/simulator.ts
```diff
@@ -653,3 +734,12 @@ export interface ConnectionData {
   /** A name for the connection. */
   readonly name: string;
 }
+
+interface SimulatorServerRequest {
```
hasanaburayyan, line 738: Should we export this interface?

hasanaburayyan (commented):
Freaking sick 🔥 

Thinking out-loud, does it make sense for some documentation around the simulator's HTTP API? Right now its just the single POST call but perhaps itd be helpful for contributor experience.

libs/wingsdk/src/target-sim/util.ts
Chriscbr: I believe the protocol used by our HTTP clients to send data to/from the process performing the simulation is just an implementation detail -- it's no different from how AWS SDK clients could be sending data to AWS through protobuf or via JSON strings.

MarkMcCulloh: It's not an implementation detail, it's part of the contract of an API. If we plan to expose it at any capacity then we should probably just use standard json.

If we don't plan to expose it and always intend to control both sides, then this is fine and there's no need to read this comment any further.

> it's no different from how AWS SDK clients could be sending data to AWS through protobuf or via JSON strings.

I'm not sure I understand the comparison, the raw JSON is what I'm concerned about. Regardless of the existence of a client that can take a `Date()`, the format that the JSON uses should ideally use a standard representation of the data. I think this would prevent surprises from hitting the endpoint without our client and/or when implementing a client in a new language.

Chriscbr: @skyrpex Definitely -- I think eventually it would be ideal support other kinds of values as arguments / return types in inflight APIs (like callbacks), but that won't be possible with this current HTTP server approach. This limitation is OK for now -- we can evolve our protocol later to support this.

@MarkMcCulloh you're right, it depends on whether we're defining the request/responses of HTTP requests as part of our public contract. For now let's not make it part of our public API contract (don't advertise it in our docs) and tell anyone that asks that it may break at any point. After some time / once there's a need,  we can choose to make it a public contract.

Chriscbr: @skyrpex I did try adding superjson support in this PR -- unfortunately it's not working because of an issue in our simulator. https://github.com/winglang/wing/issues/4131

I'm not quite sure what the root cause is so I'll probably go back to the previous hack for now

libs/wingsdk/src/testing/simulator.ts
Chriscbr: Per the discussion in the other thread -- I think for now it's best to avoid adding this to the public API surface area if there isn't a need for it, so that folks don't assume our simulator's HTTP API is stable.

libs/wingsdk/test/target-sim/__snapshots__/file-counter.test.ts.snap
Chriscbr: Added something like this (see docs). Totally experimental of course

libs/wingsdk/src/target-sim/function.inflight.ts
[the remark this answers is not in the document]
Chriscbr: For now I'd like to at least try out using the HTTP everywhere to see if it works / is reliable. The main upside in my mind is to avoid bugs that only reproduce when the client is in a different process. All that said I'm totally fine changing things if it ends up creating issues for contributing or has performance drawbacks (or if this implementation falls flat on its face 🙃 - which is totally possible)

docs/docs/02-concepts/05-simulator.md
```diff
@@ -94,6 +94,34 @@ const response = await fn.invoke("hello!");
 console.log(response);
 ```
 
+### Interacting with the simulation across processes
```
Chriscbr, line 97: (disclaimer) Note that earlier in this doc it says these APIs are subject to breaking changes. So take with a grain of salt...

skyrpex:
Awesome job! Here's what I think about the whole idea:

_First of all, we recently decided to drop Astro in favor of a SPA and an API. We still would appreciate a way to access simulated resources from a dev server, though._

We **do** need an HTTP server to enable communication from different places to the simulated resources. For example, it is needed for projects like Astro, which create a separate server for development. The backend endpoints from Astro will only be able to reach the simulated resources via an HTTP API.

On the other hand, I think the solution provided in this PR should be treated as temporary. We must embrace the server and client separation (even in the simulator target) because that's how things are going to work: a Bucket will have the server implementation living inside the simulator process, while the Bucket clients will be instantiated in different processes, such as a dev server. Also, there's many data types that clients may want to use in their interfaces (dates, big ints, bytes, blobs, even callbacks) which can't be automatically serialized and be transported through HTTP.

It may feel cumbersome to force contributors to think in server/client for all custom resources, but I'm sure we can make the experience easier somehow.

libs/wingsdk/src/testing/simulator.ts
```diff
@@ -231,6 +246,66 @@ export class Simulator {
...
+        request = JSON.parse(JSON.stringify(request), (_key, value) => {
+          return value === null ? undefined : value;
+        });
+
+        const { handle, method, args } = request;
+        const resource = this._handles.find(handle);
+
+        try {
+          const result = await (resource as any)[method](...args);
+          // assumption: Wing APIs are only returning JSON-serializable values
```
eladb, line 269: This needs to be enforced by the compiler for user defined inflight APIs of preflight classes and also in our sdk using the api checker. 

I am ok if this is not part of this PR but we should fast follow because it will easily break

libs/wingsdk/test/target-sim/__snapshots__/immutable-capture.test.ts.snap
```diff
@@ -124,19 +124,81 @@ if (!(await this.my_buckets[1].get(\\"foo\\") === \\"bar\\")) { throw new Error(
...
-})(\\"BUCKET_HANDLE_4fecd6d0\\"),(function(env) {
-  let handle = process.env[env];
+  const simulatorUrl = process.env.WING_SIMULATOR_URL;
+  if (!simulatorUrl) {
+    throw new Error(\\"Missing environment variable: WING_SIMULATOR_URL\\");
+  }
+
+  return (function makeSimulatorClient(url, handle) {
+    return new Proxy({}, {
+        get: function (_target, method, _receiver) {
```
eladb, line 139: Any way we can avoid emitting this into all inflight code? Perhaps implement this in the sdk and call from here?

eladb (approved):

libs/wingsdk/test/target-sim/__snapshots__/immutable-capture.test.ts.snap
Chriscbr: I don't remember why but I think when I tried generating code that called `require("@winglang/sdk")` it didn't seem to like it. I can look into it again and create an issue

libs/wingsdk/src/testing/simulator.ts
Chriscbr: Opened https://github.com/winglang/wing/issues/4510 👍

libs/wingsdk/test/target-sim/__snapshots__/immutable-capture.test.ts.snap
Chriscbr: So `require("@winglang/sdk")` still doesn't work, as the error I get from esbuild is like this:

```
test/target-sim/topic-producer.test.ts > publishing messages to topic
Error: Error: Failed to bundle function: Error: Build failed with 1 error:
../../../../../../private/var/folders/q3/0yxxhjr95vx4yfkk3lvy6dj40000gn/T/wingsdk.y2Spfd/.wing/function_c88517b0.js:24:17: ERROR: Could not resolve "@winglang/sdk"
    at createBundle (/Users/chrisr/dev/wing1/libs/wingsdk/src/shared/bundling.ts:40:11)
    at Sandbox.loadBundleOnce (/Users/chrisr/dev/wing1/libs/wingsdk/src/shared/sandbox.ts:98:32)
    at Sandbox.call (/Users/chrisr/dev/wing1/libs/wingsdk/src/shared/sandbox.ts:110:5)
    at Object.withTrace (/Users/chrisr/dev/wing1/libs/wingsdk/src/simulator/simulator.ts:520:24)
    at /Users/chrisr/dev/wing1/libs/wingsdk/src/simulator/simulator.ts:268:26
```

I'm not totally sure, but this could just be an issue with the SDK unit tests, because `require("@winglang/sdk")` _does_ seem to work in `.wing/preflight.js`. But it's also possible we are using some tricks in the CLI to make that work.

Anyways, I think I've found a workaround for the time being

Chriscbr: Hmm maybe I could also use symlink shenanigans?...

MarkMcCulloh:
@Chriscbr The benchmarks look like they're consistently worse every time they've run for this PR. I was expecting running `wing test` to be slower due to the additional HTTP server, but these benchmarks are just `wing compile` so it's surprising. Any idea?

libs/wingsdk/test/target-sim/__snapshots__/immutable-capture.test.ts.snap
MarkMcCulloh: I think adding an `alias` to the esbuild invocation could work. Something like:

`alias: require.resolve("@winglang/sdk")`

Chriscbr:
@MarkMcCulloh If I had to guess, it could be due to added imports. I'll try a commit where the import is inlined

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -231,6 +250,70 @@ export class Simulator {
     this._running = true;
   }
 
+  /**
+   * Start a server that allows any resource to be accessed via HTTP.
+   */
+  public async startServer(): Promise<void> {
+    // import lazily to reduce SDK import time (20ms or so)
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const express_: typeof express = require("express");
```
MarkMcCulloh, line 259: Since this is an async function you can do a dynamic import. Also I don't think the top level type import or annotation here is necessary (maybe it is, not sure)

```suggestion
    const express = await import("express");
```

Chriscbr: Ah good catch, I forgot I was in async :)

Chriscbr:
There might be at least one more issue that may need addressing. Some integ tests have been occassionally failing on my local testing. I suspect the issue is that some HTTP requests are timing out when they're calling a `cloud.Function` that take a long time. If it's not possible to simply increase the request timeouts... then we'll have to modify the client/server scheme a bit.

libs/wingsdk/test/target-sim/__snapshots__/immutable-capture.test.ts.snap
Chriscbr: `require.resolve()` still didn't work for me but I was able to use `alias` to clean things up a bit

Chriscbr:
It looks like the fetch errors are caused by an issue with headers:
```
cause: HeadersTimeoutError: Headers Timeout Error
```
Looking into possible solutions

eladb:
Wow!

PR #4127, opened 2023-09-09
feat(sdk): ex.Table - gcp support for table creation

### LINKED ISSUE #1316: cloud.Table - GCP implementation

wzslr321:
resolves #1316

Support for creating ex.Table with tf-gcp target

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-tf-gcp/table.ts
```diff
@@ -0,0 +1,76 @@
...
+      name: tableName + `-${randomId.hex}`,
+      instanceName: instanceName,
+      columnFamily: columnsFamily,
+    };
+
+    new BigtableTable(this, "Default", config);
+  }
+
+  public addRow(key: string, row: Json): void {
+    throw new Error(`Can not insert ${row} at given ${key}`);
```
wzslr321, line 70: As far as I can understand, it is impossible to implement this function at this point, since it requires a client.

reference: https://cloud.google.com/bigtable/docs/samples/bigtable-insert-update-rows

wzslr321:
I added some support for creating the table and a corresponding test - it seems to pass, but I feel like this MR is not ready yet. Among other,  I am not sure what to do with `primaryKey`, since I can not find any information in `BigtableTableConfig` about where it should be supplied. 

I can also see on [Compatbility Matrix ](https://www.winglang.io/docs/standard-library/compatibility-matrix) that I should implement inflight methods to, therefore I will 'skip' to doing that. 

In the meantime, can someone review and tell me if I am going in the right direction or if some part of configuration is missing that I am unaware of? The only reference about GCP implementation, I've found here - #827, and it does not give me enough confidence since it is swamped with other changes.

wzslr321:
I've received feedback on Slack, no need to review as of this moment :)

Issue #108, opened 2022-09-20
Access Modifiers


staycoolcall911:
### 1.5 Access Modifiers
Visibility inference is done with the following rules:
- Default visibility is `private` for all members. If modifiers are missing, the
  symbol is assumed private by the compiler and not exported.
- `pub` is to declare a symbol that is visible for and exported publicly.
- `protected` is to declare a symbol that is visible for and exported publicly
  but only for the class and its subclasses.
- `internal` is to declare a symbol that is visible for and exported publicly
  but only for the current compilation unit.
Accessing fields, members, or structured data is done with `.`.
Visibility modifiers can be applied to members of classes.
Mixing `protected` and `internal` is not allowed.

staycoolcall911:
The bare minimum is adding `pub` - let's see if that makes sense when we implement

staycoolcall911:
Another requirement (from @eladb): all fields must be private (access should be done via getter/setter)

monadabot:
Congrats! :rocket: This was released in [Wing 0.30.9](https://github.com/winglang/wing/releases/tag/v0.30.9).

Issue #1746, opened 2023-03-07
Write style guide / convention for error messages


Chriscbr:
Provide guidance on a recommended style / voice for our compiler's error messages.

Context: some languages like Rust and Go typically show errors in all-lowercase, while other languages like Swift and others start errors with capital letters. Most languages also don't end error messages with periods, but I'm not sure if it's universal

eladb:
@Chriscbr can you pick this up?

Chriscbr:
We discussed this issue briefly during the March 28 Wing community meeting ([link to slides I made](https://docs.google.com/presentation/d/14SFO7SrntYSLqYostC1gcaQPIjVZxOpbDkxpmsIwEE8/edit?usp=sharing)).

We said we should try to group all of the error messages in the compiler into a single file so that style guide notes/comments can be put right beside them. This will also let us make improvements like including unique error codes etc.

It was also suggested start out with avoiding periods, and using backticks for names of types/variables/keywords etc. Some folks liked errors starting with lowercase letters, and others liked them starting with uppercase letters. Nothing too surprising. 😅

PR #4133, opened 2023-09-10
feat(compiler): access modifiers

yoav-steinberg:
See #108 
This PR adds `public` and `protected` support to class members and while doing so makes all other members **private** by default. It also verifies interface method implementations must be `public`. It also includes support for JSII access modifiers. To get this working I did the following refactors:

* Cleaned up fields in the `SymbolEnv` and instead added a `kind` field distinguishing between 3 types of environments: a simple code block (`Scope`) a function/method/closure environemt (`Function`) and an environment used to store members of a type (`Type`). This was primarily needed so I can use the result of a symbol env lookup to figure out in what ancestor class a given member was defined. The added benefit was code clarity and sanity checks which, btw, found a non-related bug in super constructor type checking (see below). One piece of such clarity was the long awaited removal of `return_type` from `SymbolEnv`.
* Cleaned up the `ContextVisitor` which I needed to add to the `TypeChecker`. Here I removed the requirement to store the `init_env` in `push_class` which didn't make sense, and is probably a leftover of some iteration on our lifting code. I also cleaned up the `push_method` code to make more sense and now it's a more general `push_function` stack with a helper for when we're just interested in methods.
* I added some more stuff to `VisitorWithContext` trait, for easy "push -> do stuff or even return -> pop" pattern. Specifically I needed the `TypeChecker` to push function definitions and current statement safely.
* Use a `VisitorContext` in `TypeChecker` to track context so I know in what class I am for access modifier checking. This approach also provides a better way of tracking the current `return_type` so I could remove it from the env. This also removed all sorts of dirt from the `TypeChecker` struct (like `in_json`, etc.).
* The `type_check_statement` method was getting way to big, so I moved the contents of its match statements into separate methods.
* The sanity checks when creating new `SymbolEnv`s detected that we're creating a weird env in `type_check_super_constructor_against_parent_initializer`. I removed a large bunch of code there that didn't seem correct to me.
* The actual access modifier verification takes place inside `get_property_from_class_like` (**<= this is where you want to look**). This way I avoid cascading errors and "unhydrated" classes. To make this work for all cases I updated `resolve_referece` for static type references to also use this method instead of doing its own thing. I think this is cleaner.

Still left and possible candidate for followup PRs:
- [x] Tests for JSII imports with access modifiers
- [x] Handle method overriding rules
- [x] Support `public` for types (export types from module) -> out of scope, will be handled added by our module system.
- [x] Decide if we allow `public` fields: tracking issue #4180
- [x] Decide about `internal`: tracking issue #4156

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -3770,24 +3924,22 @@ impl<'a> TypeChecker<'a> {
...
+			SymbolEnvKind::Scope,
 			env.phase,
 			stmt.idx,
 		));
 		self.types.set_scope_env(statements, if_scope_env);
-		self.inner_scopes.push(statements);
+		self.inner_scopes.push((statements, self.ctx.clone()));
 	}
 
 	fn type_check_super_constructor_against_parent_initializer(
```
yoav-steinberg, line 3935: @hasanaburayyan please take a look at my changes here, just deleted some code that didn't make sense to me. Am I missing something?

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -10,6 +10,105 @@ exports[`access_hidden_namespace.w 1`] = `
...
+Test Files 1 failed (1)
+Duration <DURATION>"
+`;
+
+exports[`access_modifiers.w 1`] = `
+"error: Cannot access protected member protected_field of Foo
+   --> ../../../examples/tests/invalid/access_modifiers.w:64:9
+   |
+64 | log(foo.protected_field);
+   |         ^^^^^^^^^^^^^^^ Cannot access protected member protected_field of Foo
```
Chriscbr, line 23: Should we put names of user identifiers in quotes or backticks to signal they're not part of the error message? e.g.
```suggestion
   |         ^^^^^^^^^^^^^^^ Cannot access protected member "protected_field" of "Foo"
```

I don't think anyone has proposed a style guide yet (https://github.com/winglang/wing/issues/1746) but I think the compiler already does this for some other error messages

libs/wingc/src/type_check.rs
```diff
@@ -2944,764 +2967,890 @@ impl<'a> TypeChecker<'a> {
...
-				// preflight classes cannot be declared inside an inflight scope
-				// (the other way is okay)
-				if env.phase == Phase::Inflight && *phase == Phase::Preflight {
-					self.spanned_error(stmt, format!("Cannot declare a {} class in {} scope", phase, env.phase));
-				}
-				// Verify parent is a known class and get their env
-				let (parent_class, parent_class_env) = self.extract_parent_class(parent.as_ref(), *phase, name, env);
+			StmtKind::Class(ast_class) => {
+				tc.type_check_class(env, stmt, ast_class);
+			}
```
Chriscbr, line 3033: Great cleanup

libs/wingc/src/type_check.rs
```diff
@@ -4728,20 +4852,60 @@ impl<'a> TypeChecker<'a> {
...
-		if let LookupResult::Found(field, _) = lookup_res {
+		if let LookupResult::Found(field, lookup_info) = lookup_res {
 			let var = field.as_variable().expect("Expected property to be a variable");
+
+			// Determine the access type of the property
+			// Lookup the property in the class env to find out in which class (perhaps an ancestor) it was defined
+			let SymbolEnvKind::Type(property_defined_in) = lookup_info.env.kind else {
+				panic!("Expected env to be a type env");
+			};
+			// Check if the class in which the property is defined is the current class nesting
```
Chriscbr, line 4872: What does this sentence mean?

examples/tests/invalid/access_modifiers.w
```diff
@@ -0,0 +1,100 @@
...
+  public public_method() {}
+  protected protected_method() {}
+  private_method() {}
+
+  public static public_static_method() {}
+  protected static protected_static_method() {}
+  static private_static_method() {}
+}
+
+class Bar extends Foo {
```
Chriscbr, line 40: Can we add a test checking this works with more than one level of inheritance?

examples/tests/invalid/access_modifiers.w
```diff
@@ -0,0 +1,100 @@
+class Foo {
+  protected protected_field: str;
+  private_field: str;
+  public public_field: str;
```
Chriscbr, line 4: Can access modifiers be added to an interface? (add a valid / invalid unit test)

Chriscbr:
Looks really great so far!

libs/wingc/src/type_check.rs
hasanaburayyan: Will do, excited to review this 😁

libs/wingc/src/type_check.rs
```diff
@@ -4520,6 +4638,13 @@ impl<'a> TypeChecker<'a> {
 
 				let mut property_variable = self.resolve_variable_from_instance_type(instance_type, property, env, object);
 
+				// Validate access modifier
+				// if !property_variable.type_.is_unresolved() {
+				// 	// TODO: this isn't good enough becaues `instance_type` might be an array so we need to "actual" class_type (Array). This is done for us in
```
hasanaburayyan, line 4643: do we have a tracking issue?

hasanaburayyan (commented):
Love the cleanups, let a comment about a bug with super initializers.

libs/tree-sitter-wing/grammar.js
```diff
@@ -532,7 +532,7 @@ module.exports = grammar({
 
     async_modifier: ($) => "async",
 
-    access_modifier: ($) => choice("public", "private", "protected"),
+    access_modifier: ($) => choice("public", "protected"),
```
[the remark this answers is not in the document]
yoav-steinberg: Yes, it seems to me more strongly related to "compilation units" and I as such I'd like a word with @Chriscbr (our module guy) to better understand the design.

libs/wingc/src/type_check.rs
yoav-steinberg: oops, forgot to delete this comment. This is a leftover comment from an earlier iteration. Resolved.

libs/wingc/src/type_check.rs
yoav-steinberg: I changed to:
"Check if the class in which the property is defined is one of the classes we're currently nested in"
Hopefully that's clearer. I definitely struggled to correctly describe what I'm doing here (and I'm also not happy with the method name `self.ctx.current_class_nesting` 3 lines below).
What I'm doing here is I run up the stack of currently nested classes to see if the property access is from one of those classes. Here's an example to make it clearer:
```wing
class A {
  static foo() {} // This is a **private** static method.
  method() {
    class B {
        method() {
          A.foo() // I can access A's private stuff since I'm nested inside B which is nested inside A.
        }
    }
  }
}
```

libs/wingc/src/type_check.rs
```diff
@@ -3799,43 +3943,7 @@ impl<'a> TypeChecker<'a> {
 							.collect_vec()[0]
 							.1;
 
-						let class_initializer = &class_type
```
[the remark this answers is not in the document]
yoav-steinberg: Thanks for looking into this. I see the issue. I mixed things up. I'll try to fix it.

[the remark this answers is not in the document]
yoav-steinberg: I pushed a fix, the gist of is is that if we type check the `super()` call statement during the type checking of the `init` itself then there's no need for env acrobatics. If we try to do this in advance then your original logic holds, with a slight required change, but it seems a bit complex. Can you review this commit: https://github.com/winglang/wing/pull/4133/commits/e36efd38f4d89ebed6e94762a0ea44f739220b3f

libs/wingc/src/type_check.rs
yoav-steinberg: resolving since we've continued this discussion in another conversation below

examples/tests/invalid/access_modifiers.w
yoav-steinberg: Good callout. Anything implementing an interface must be `public`. I'll need to added verification and tests...

libs/wingc/src/type_check.rs
Chriscbr: Got it, yeah makes more sense now  - thanks!

examples/tests/invalid/access_modifiers.w
yoav-steinberg: https://github.com/winglang/wing/pull/4133/commits/d3d1c3c33f35fb43ed570f2881d5b8782821d5fa

libs/wingc/src/ast.rs
```diff
@@ -513,6 +518,14 @@ pub struct ClassField {
 	pub reassignable: bool,
 	pub phase: Phase,
 	pub is_static: bool,
+	pub access_modifier: AccessModifier,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum AccessModifier {
```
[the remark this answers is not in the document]
yoav-steinberg: Added: https://github.com/winglang/wing/issues/4156

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -10,6 +10,147 @@ exports[`access_hidden_namespace.w 1`] = `
...
+Test Files 1 failed (1)
+Duration <DURATION>"
+`;
+
+exports[`access_modifiers.w 1`] = `
+"error: Method \\"private_method\\" is not public but it's part of the \\"SomeInterface\\" interface
+   --> ../../../examples/tests/invalid/access_modifiers.w:40:3
+   |
+40 |   private_method() {}
+   |   ^^^^^^^^^^^^^^ Method \\"private_method\\" is not public but it's part of the \\"SomeInterface\\" interface
```
Chriscbr, line 23: optional - I'm not sure if this message makes it obvious why being part of the interface means it has to be public. Maybe we could reword this to `Interface members must be declared public` or `Method "foo" is private in "Foo" but not in "SomeInterface"`, or something like that.

```suggestion
40 |   private_method() {}
   |   ^^^^^^^^^^^^^^ Interface members must be declared public
```

libs/tree-sitter-wing/grammar.js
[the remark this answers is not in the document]
yoav-steinberg: Added issue and referenced it diags: #4156

examples/tests/invalid/access_modifiers.w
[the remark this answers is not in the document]
yoav-steinberg: please review: https://github.com/winglang/wing/pull/4133/commits/5e7c8c71c81e62a74293eb479c4d8821659d33c8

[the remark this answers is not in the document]
Chriscbr: LGTM

libs/wingc/src/type_check.rs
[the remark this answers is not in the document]
hasanaburayyan: 🤌

libs/wingsdk/src/.gen/versions.json
Chriscbr, line 1: I think this file is supposed to be gitignored now

Chriscbr (approved):

Issue #3096, opened 2023-06-27
Standard library: utility functions


staycoolcall911:
### Feature Spec

A number of simple utility functions are added to Wing's standard library as described in [the RFC](https://www.winglang.io/contributing/rfcs/stdlib).
This includes the following modules:
- [x] `math`
- [x] `fs`
- [x] `util`

`http` module was already previously added.
- [x] http 

### Use Cases

Support simple programming needs such as performing math or file-system operations.

### Implementation Notes

`util` module: `env`, `tryEnv` and `sleep` were done previously (see the source under https://github.com/winglang/wing/blob/main/libs/wingsdk/src/util/util.ts). Still need to add the rest of the methods defined in the RFC.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

revitalbarletz:
@marciocadev is working on the math module.

staycoolcall911:
The only thing left to close this one is adding a simple `fs` module as defined in the RFC [here](https://www.winglang.io/contributing/rfcs/stdlib).

staycoolcall911:
@marciocadev - took you off this issue, because you already added the `math` and `util` modules, but if you want to implement the `fs` module as well please reassign this to yourself. Thanks!

jianzs:
Hey, there. I am working on the `fs` module.

PR #4135, opened 2023-09-10
feat(sdk): add fs module

jianzs:
Solves #3096 

Support fs module. All file paths are POSIX file paths, and will be normalized to the target platform when performing operations.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * Convert the path to Windows format if running on Windows.
+ * @param p the path to convert
+ * @returns normalized path
+ */
+function localePath(p: string): string {
+  if (process.platform !== `win32`) {
+    return p;
+  }
+  return p.split(path.sep).join(path.win32.sep)
```
tsuf239, line 16: it won't work on windows since in windows `path.sep === path.win32.sep` (https://nodejs.org/api/path.html#pathsep)
I think  this will work for all os:
`return p.split(path.posix.sep).join(path.sep)`

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+ */
+export interface RemoveOptions {
+  /**
+   * When `true`, exceptions will be ignored if `path` does not exist.
+   * @default false
+   */
+  force?: boolean | undefined;
+  /**
+   * If an `EBUSY`, `EMFILE`, `ENFILE`, `ENOTEMPTY`, or
+   * `EPERM` error is encountered, Node.js will retry the operation with a linear
```
tsuf239, line 47: `Node.js`?

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+  retryDelay?: number | undefined;
+}
+
+/**
+ * Custom settings for removing directories
+ */
+export interface RemoveDirOptions {
+  /**
+   * If an `EBUSY`, `EMFILE`, `ENFILE`, `ENOTEMPTY`, or
+   * `EPERM` error is encountered, Node.js will retry the operation with a linear
```
tsuf239, line 74: `Node.js`?

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * The fs class is used for interacting with the file system.
+ * All file paths must be POSIX file paths (/ instead of \),
+ * and will be normalized to the target platform if running on Windows.
+ */
+export class Util {
+  /**
+   * Join all arguments together and normalize the resulting path.
```
tsuf239, line 96: Some of the docStrings have no @params or @returns can you please fix that?

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+   * @param dirpath The path to the directory you want to create.
+   */
+  public static mkdir(dirpath: string, opts?: MkdirOptions): void {
+    const normPath = localePath(dirpath);
+    fs.mkdirSync(normPath, opts);
+  }
+  /**
+   * Returns the created directory path.
+   * Generates six random characters to be appended behind a required prefix to create a unique temporary directory.
+   */
```
tsuf239, line 173: can you please adjust this doc string (the return to be on a `@returns` annotation under the description, add `@param`)

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+  public static mkdirTemp(prefix: string): string {
+    const dirpath = fs.mkdtempSync(prefix);
+    if (process.platform !== `win32`) {
+      return dirpath;
+    }
+    return dirpath.split(path.win32.sep).join(path.posix.sep);
```
tsuf239, line 179: maybe return `return dirpath.split(path.sep).join(path.posix.sep);` for all of them? (more "computational work" of one side, less future maintenance of the other )

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+  /**
+   * Returns the contents of the `path`.
+   */
+  public static readFile(filepath: string): string {
```
tsuf239, line 184: can you please add encoding options (and more)?

libs/wingsdk/test/fs/fs.test.ts
```diff
@@ -0,0 +1,83 @@
+import { test, expect } from "vitest";
```
tsuf239, line 1: Can you please add an example written in wing for each of the methods? (here: `examples/tests/sdk_tests/fs`)

libs/wingsdk/test/fs/fs.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  const dirpath = fs.mkdirTemp(prefix);
+  expect(dirpath).match(new RegExp(`^${prefix}`));
+  
+  fs.removeDir(dirpath);
+  const exist = fs.exists(dirpath);
+  expect(exist).toEqual(false);
+})
+
+test("write and read json", async () => {
+  const filename = `testfile-${randomStr()}.txt`
```
tsuf239, line 62: can it end with `.json`?

libs/wingsdk/test/fs/fs.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  const obj = fs.readJson(filename);
+  expect(obj).toEqual(jsonObj);
+  
+  fs.remove(filename);
+  const exist = fs.exists(filename);
+  expect(exist).toEqual(false);
+});
+
+test("write and read yaml", async () => {
+  const filename = `testfile-${randomStr()}.txt`
```
tsuf239, line 74: can it end with `.yaml`?

tsuf239 (commented):
Great work!! just a few tiny adjustments (mostly to the js docs) and an addition of an example written in wing, and you're there! 💫

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+import { InflightClient } from "../core";
+import { Json } from "../std";
+
+/**
+ * Convert the path to Windows format if running on Windows.
+ * @param p the path to convert
+ * @returns normalized path
+ */
+function localePath(p: string): string {
+  if (process.platform !== `win32`) {
```
eladb, line 13: nit:

```suggestion
  if (process.platform !== "win32") {
```

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * Custom settings for creating directory
+ */
+export interface MkdirOptions {
+  /**
+   * Indicates whether parent folders should be created.
+   * If a folder was created, the path to the first created folder will be returned.
+   * @default false
+   */
+  recursive?: boolean | undefined;
```
eladb, line 28: The `?` implies `| undefined`:

```suggestion
  recursive?: boolean;
```

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+  /**
+   * A file mode. If a string is passed, it is parsed as an octal integer. If not specified
+   * @default 0o777
+   */
+  mode?: string | number | undefined;
```
eladb, line 33: We don't support union types. We need to choose one.

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+   * @param p the path to evaluate.
+   */
+  public static exists(p: string): boolean {
+    const normPath = localePath(p);
+    return fs.existsSync(normPath);
+  }
+  /**
+   * Reads the contents of the directory.
+   */
+  public static readDir(dirpath: string): Array<string> {
```
eladb, line 149:
```suggestion
  public static readdir(dirpath: string): Array<string> {
```

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+    const normPath = localePath(dirpath);
+    return fs.readdirSync(normPath);
+  }
+  /**
+   * Reads the contents of the directory if the path exists, `undefined` otherwise.
+   */
+  public static tryReadDir(dirpath: string): Array<string> | undefined {
```
eladb, line 156:
```suggestion
  public static tryReaddir(dirpath: string): Array<string> | undefined {
```

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+   */
+  public static mkdir(dirpath: string, opts?: MkdirOptions): void {
+    const normPath = localePath(dirpath);
+    fs.mkdirSync(normPath, opts);
+  }
+  /**
+   * Returns the created directory path.
+   * Generates six random characters to be appended behind a required prefix to create a unique temporary directory.
+   */
+  public static mkdirTemp(prefix: string): string {
```
eladb, line 174: Rename to `mkdtemp` and if `prefix` is undefined, then default to: `fs.mkdtempSync(path.join(os.tmpdir(), "wing"))` or something like that.

```suggestion
  public static mkdtemp(prefix?: string): string {
```

It's very useful to be able to just call `fs.mkdtemp()` and get a temp directory.

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+   */
+  public static remove(p: string, opts?: RemoveOptions): void {
+    const normPath = localePath(p);
+    fs.rmSync(normPath, opts)
+  }
+  /**
+   * Removes directories. Returns `undefined`.
+   * @param dirpath The path to the directory you want to remove.
+   */
+  public static removeDir(dirpath: string, opts?: RemoveDirOptions): void {
```
eladb, line 269: Do we really need `removeDir` if we have `remove`?

eladb (changes requested):

jianzs:
Thank you all for your suggestions and comments, I will improve it as soon as possible

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+    const normPath = localePath(dirpath);
+    fs.rmdirSync(normPath, opts)
+  }
+  /**
+   * @internal
+   */
+  public static _toInflightType(): string {
+    return InflightClient.forType(__filename, this.name);
+  }
+}
```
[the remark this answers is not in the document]
jianzs: Perhaps simply propagating low-level errors to the user is sufficient? like example at `examples/tests/sdk_tests/fs/json.main.w` and `examples/tests/sdk_tests/fs/directory.main.w`.

examples/tests/sdk_tests/fs/json.main.w
```diff
@@ -0,0 +1,46 @@
...
+
+let filename = "test-json.json";
+let data = Json {
+    "foo": "bar", 
+    "arr": [1, 2, 3, "test", { "foo": "bar" }]
+};
+
+try {
+    fs.writeFile(filename, "invalid content");
+    fs.readJson(filename);
```
eladb, line 12: ❤️

examples/tests/sdk_tests/fs/json.main.w
```diff
@@ -0,0 +1,46 @@
...
+} catch e {
+    assert(regex.match("^Unexpected token", e) == true);
+}
+
+fs.writeJson(filename, data);
+assert(fs.exists(filename) == true);
+
+let obj = fs.readJson(filename);
+assert(Json.stringify(obj) == Json.stringify(data));
```
eladb, line 21:
```suggestion
assert(obj == data);
```

I suspect this should work, right @hasanaburayyan ?

examples/tests/sdk_tests/fs/path.main.w
```diff
@@ -0,0 +1,36 @@
...
+assert(result == "/a/b/c");
+
+result = fs.basename(from);
+assert(result == "d");
+
+result = fs.resolve(from, to);
+assert(result == "/a/b/e/f");
+
+test "inflight path conversion" {
+    let var result = fs.join(from, to);
```
eladb, line 22: @MarkMcCulloh @Chriscbr @yoav-steinberg any thoughts on how we can avoid the duplicated test? It's related to phase independence

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,292 @@
...
+  public static mkdir(dirpath: string, opts?: MkdirOptions): void {
+    const normPath = localePath(dirpath);
+    fs.mkdirSync(normPath, opts);
+  }
+  /**
+   * Create a temporary directory.
+   * Generates six random characters to be appended behind a required prefix to create a unique temporary directory.
+   * @param prefix The prefix for the directory to be created, default `wingtemp`.
+   * @returns The created directory path.
+   */
```
eladb, line 175: Thank god for this

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,292 @@
...
+   * @returns The JSON object contained in the file.
+   */
+  public static readJson(filepath: string): Json {
+    const text = Util.readFile(filepath);
+    return JSON.parse(text) as Json;
+  }
+  /**
+   * Get the content of the file and convert it to JSON if the path exists, `undefined` otherwise.
+   * @param filepath The file path of the JSON file.
+   * @returns The JSON object contained in the file, `undefined` otherwise.
```
eladb, line 222: Worth documenting that an invalid json will throw (and not return undefined)

libs/wingsdk/src/fs/fs.ts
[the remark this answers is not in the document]
eladb: I think it's totally sufficient for now

eladb (approved):
Great job @jianzs 🙏 love the thoroughness and high quality code and testing. I love the batteries included, feel free to add more capabilities here as you see fit. I've always found fs libraries to be too puristy and I think having a rich standard library makes a difference.

Ideally but definitely not required right now, if you call these methods from preflight they will be sync, and calling them from inflight should be async.. it's not possible to do right now but putting this thought on @yoav-steinberg 's radar...


BTW, we need a `process` library if you'd like to take the design. (spawn, shell, etc).

examples/tests/sdk_tests/fs/json.main.w
jianzs: I have attempted this approach, but it is not functioning as expected. I am unable to determine the underlying issue, even though the output of Json.stringify for these two objects is identical.

examples/tests/sdk_tests/fs/path.main.w
Chriscbr: IMO the most straightforward solution for the duplicate code is to write the test logic in an `unphased` function and then call the function in both preflight and inflight.

https://github.com/winglang/wing/pull/1711

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -287,13 +287,6 @@ error: \\"cloud\\" is already defined
   |       ^^^^^ \\"cloud\\" is already defined
 
 
-error: \\"fs\\" is not a built-in module
-  --> ../../../examples/tests/invalid/bring.test.w:6:1
-  |
-6 | bring fs;
-  | ^^^^^^^^^ \\"fs\\" is not a built-in module
```
Chriscbr, line 294: We can probably remove this line from examples/tests/invalid/bring.test.w

Chriscbr:
@jianzs Since we're introducing some functionality that deals with paths, I've added the `pr/e2e-full` label so that our continuous integration runs everything on Windows as well. It should take affect on any new commits added to the branch.

Chriscbr (approved):

examples/tests/sdk_tests/fs/json.main.w
jianzs: Weirdly, I am unable to reproduce the phenomenon using the 'extern' mechanism. Regardless of how I modify the programming approach, the assertion always evaluates to true.

In the fs sdk, I utilize the 'as' operator to convert the result of JSON.parse into a Json object. This may not be considered as best practice? I have yet to discover how to create a Json object in JavaScript.


Here is the code.
json.main.w
```
class JsJson {  
    extern "./json.js" pub static writeJson(obj: Json);
    extern "./json.js" pub static readJson(): Json;
}

let data1 = Json {
    "foo": "bar", 
    "arr": [1, 2, 3, "test", { "foo": "bar" }]
};

JsJson.writeJson(data1);
let data2 = JsJson.readJson();

test "json compare" {
    log(Json.stringify(data1));
    log(Json.stringify(data2));
    assert(data1 == data2);
}
```

json.js
```javascript
const fs = require("fs");

exports.writeJson = function (obj) {
    fs.writeFileSync("data.json", JSON.stringify(obj, null, 2));
}

exports.readJson = function () {
    const content = fs.readFileSync("data.json");
    return JSON.parse(content);
};
```

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,302 @@
...
+  winPath = winPath.replace(/^\\([A-Za-z])\\/, "$1:\\");
+  return winPath;
+}
+
+/**
+ * Convert the path to POSIX format.
+ * @param p The path to convert.
+ * @returns The Normalized path.
+ */
+function posixPath(p: string): string {
```
MarkMcCulloh, line 28: It's not really possible to convert windows path to posix path.

`C:\` is not that same as `/c/`, nor is it always the same as `/`.

If you're running on windows and you attempt to read `/`, I think a reasonable behavior of wing would be to just let windows handle it. Which means it'll read the default system drive afaik. If we do that then I don't think this function is needed.

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,302 @@
...
+ * @param p The path to convert.
+ * @returns The Normalized path
+ */
+function localePath(p: string): string {
+  if (process.platform !== "win32") {
+    return p;
+  }
+  let winPath = p.split(path.posix.sep).join(path.win32.sep);
+  // If the path starts from the root directory, the first segment represents a specific drive.
+  winPath = winPath.replace(/^\\([A-Za-z])\\/, "$1:\\");
```
MarkMcCulloh, line 19: On windows, the path `/c/` can resolve to `C:\c`

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,302 @@
...
+ * and will be normalized to the target platform if running on Windows.
+ */
+export class Util {
+  /**
+   * Join all arguments together and normalize the resulting path.
+   * @param paths The array of path need to join.
+   * @returns The resulting path after joining all the paths.
+   */
+  public static join(...paths: string[]): string {
+    return posixPath(path.join(...paths));
```
MarkMcCulloh, line 97: the SDK already has a function for normalizing path separators you can use, in the util file I believe.

```suggestion
    return normalPath(path.join(...paths));
```

jianzs: Thanks for the reminder. However, I wonder if it will work correctly when the path starts with a specified drive. The `mkdtemp` method will return a path with a specified drive. I tried using this method to convert 'C:\a\b\c\d', and the result was 'C:/a/b/c/d'. Does this behave as expected?

libs/wingsdk/src/fs/fs.ts
jianzs: Do you mean we should keep the drive in the posix-like path?

libs/wingsdk/src/fs/fs.ts
MarkMcCulloh: On windows `C:\a\b\c\d` and `C:/a/b/c/d` are equivalent as far as I know, so I'd say it works as expected if the goal is to at least attempt to normalize between windows and posix.
The normalization is most useful for relative paths of course, which I think will be the normal use-case within wing.

libs/wingsdk/src/fs/fs.ts
MarkMcCulloh: > Do you mean we should keep the drive in the posix-like path?

Yes, I think no attempt should be made convert absolute Windows paths into a _truly_ posix-compliant path. Converting `\` to `/` is useful since it helps with consistency across platforms and doesn't hurt functionality, but mangling the drive letter is not possible to handle automatically without possibly breaking something.

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,293 @@
...
+    return normalPath(path.resolve(...paths));
+  }
+  /**
+   * Check if the path exists.
+   * @param p The path to evaluate.
+   * @returns `true` if the path exists, `false` otherwise.
+   */
+  public static exists(p: string): boolean {
+    const normPath = localePath(p);
+    return fs.existsSync(normPath);
```
MarkMcCulloh, line 137: No need to normalize paths when calling into the actual filesystem, it's safe to assume the given path is a usable path.

```suggestion
    return fs.existsSync(p);
```

MarkMcCulloh: This feedback applies to other methods here as well. We should only normalize returned paths.

libs/wingsdk/package.json
```diff
@@ -105,6 +106,7 @@
     "esbuild-wasm": "^0.18.20",
     "express": "^4.18.2",
     "ioredis": "^5.3.2",
+    "js-yaml": "^4.1.0",
```
MarkMcCulloh, line 109: We actually already have a dependency on https://www.npmjs.com/package/yaml, could that be used instead?

MarkMcCulloh: Yeah we definitely want to minimize deps (even though there are quite a lot already).

Second output makes sense as a default, so this should work:

```js
yaml.stringify(data, { aliasDuplicateObjects: false })
```

jianzs: Oh! Finally! Thank you!

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * If the path exists, read the contents of the directory; otherwise, return `undefined`.
+   * @param dirpath The path to evaluate.
+   * @returns The contents of the directory if the path exists, `undefined` otherwise.
+   */
+  public static tryReaddir(dirpath: string): Array<string> | undefined {
+    if (!Util.exists(dirpath)) {
+      return undefined;
+    }
+    return Util.readdir(dirpath);
```
MarkMcCulloh, line 138: No need to call exists first. readdir can fail for other reasons so might as well make this more bulletproof.

```suggestion
    try {
      return Util.readdir(dirpath);
    } catch {
      return undefined;
    }
```

examples/tests/sdk_tests/fs/basic.main.w
```diff
@@ -0,0 +1,30 @@
+bring fs;
+
+let filename = "hello.txt";
+let data = "Hello, Wing!";
+
+fs.writeFile(filename, data);
```
MarkMcCulloh, line 6: All of these tests need to write to a tmp directory because we run these tests in AWS lambda where nothing outside of /tmp is writable.

(See here https://github.com/winglang/wing/actions/runs/6498033748/job/17648493122 where I ran the tests in AWS and it failed for fs)

MarkMcCulloh (changes requested):
Really great work, last change needed

examples/tests/sdk_tests/fs/basic.main.w
jianzs: Thank you for reviewing and providing assistance!!

examples/tests/sdk_tests/fs/basic.main.w
```diff
@@ -0,0 +1,32 @@
+bring fs;
+
+let filepath = "/tmp/hello-preflight.txt";
```
MarkMcCulloh, line 3: Should use fs.mkdtemp() here

libs/wingsdk/src/fs/fs.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * Custom settings for creating directory
+ */
+export interface MkdirOptions {
+  /**
+   * Indicates whether parent folders should be created.
+   * If a folder was created, the path to the first created folder will be returned.
+   * @default false
```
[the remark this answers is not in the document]
eladb: Wow pretty wild, but I am in.

PR #4147, opened 2023-09-11
chore(build): gitignore non-deterministic json files from `cdktf get`

MarkMcCulloh:
The output order of these json files sees to depend on the system. AFAIK they are not useful to commit (the version is already pinned in the root `cdktf.json` file) so let's just get rid of them.

@Chriscbr let me know if this makes sense, since you added the providers


Chriscbr (approved):
LGTM

Issue #4153, opened 2023-09-12
console tests are causing ci timeouts


polamoros:
### I tried this:

Runing the CI
https://github.com/winglang/wing/actions/runs/6000565194/job/16301782522?pr=3973

### This happened:

<img width="274" alt="image" src="https://github.com/winglang/wing/assets/5547636/3a2fbc03-5d01-4a77-a9d4-9c60a4f8ab83">

### I expected this:

To finish without a timeout.

### Is there a workaround?

rerun the CI

### Component

Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.30.5](https://github.com/winglang/wing/releases/tag/v0.30.5).

PR #4154, opened 2023-09-12
fix(console): tests are causing ci timeouts

polamoros:
Improve console tests to prevent CI timeouts caused by some tests randomly getting stuck.

resolves https://github.com/winglang/wing/issues/4153

### DIFF
--- a/apps/wing-console/console/app/playwright.config.ts
+++ b/apps/wing-console/console/app/playwright.config.ts
@@ -14,6 +14,7 @@ export default defineConfig({
   workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: [["html", { open: "never" }]],
+  timeout: 10_000,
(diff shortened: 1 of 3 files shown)


MarkMcCulloh:
image

Not sure if you saw, but a failed test still passes the build (it's just considered flakey). I think these kind of tests probably shouldn't allow for any flakiness

polamoros:
> image
> Not sure if you saw, but a failed test still passes the build (it's just considered flakey). I think these kind of tests probably shouldn't allow for any flakiness

The test was flagged as flaky because it timed out during the teardown process. (see [here](https://github.com/winglang/wing/actions/runs/6160127414/job/16716372097?pr=4154#step:9:2595))

Tests are usually flagged as flaky when they succeed after a retry.
I agree that we shouldn't allow flakiness, but if we remove the retries, we'll have many CI failures due to the console tests..

ainvoner (approved):

PR #4163, opened 2023-09-12
fix(sdk): fixing table methods

tsuf239:
## Checklist

image

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-tf-aws/table.ts
+++ b/libs/wingsdk/src/target-tf-aws/table.ts
@@ -56,7 +56,10 @@ export class Table extends ex.Table {
       throw new Error("tables can only be bound by tfaws.Function for now");
     }
 
-    if (ops.includes(ex.TableInflightMethods.INSERT)) {
+    if (
+      ops.includes(ex.TableInflightMethods.INSERT) ||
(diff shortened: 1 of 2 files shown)


libs/wingsdk/src/target-tf-aws/table.ts
```diff
@@ -56,7 +56,10 @@ export class Table extends ex.Table {
       throw new Error("tables can only be bound by tfaws.Function for now");
     }
 
-    if (ops.includes(ex.TableInflightMethods.INSERT)) {
+    if (
+      ops.includes(ex.TableInflightMethods.INSERT) ||
+      ops.includes(ex.TableInflightMethods.UPSERT)
+    ) {
```
Chriscbr, line 62: Can we add unit tests for this?

Chriscbr (approved):

libs/wingsdk/src/target-tf-aws/table.ts
tsuf239: how? (since there is no automatic way as far as I know to know which permissions are needed for each sdk method?) The wing SDK test caught it, we just need to run them on time....

PR #4164, opened 2023-09-12
feat(sdk): replace `cloud.Api` REST API with HTTP API for `tf-aws` target 

### LINKED ISSUE #4097: Replace REST API Gateway with HTTP API for tf-aws target
### Feature Spec

Using the [HTTP API](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api.html) rather than the [REST API](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-rest-api.html) for the `tf-aws` target will simplify the implementation and gets rid of some custom plumbing around 404 handlers and CORS. On top of that, it'll be [cheaper](https://cloudonaut.io/review-api-gateway-http-apis/) per request.

The HTTP API lacks a few things, most notably (see here for more [details](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html)):

- X-Ray
- Edge endpoints

X-Ray is probably the one which might be missed in the future, however it's not being used at the moment. Edge endpoints are just baked in Cloudfront capabilities which could be easily added.

### Use Cases

Generally I think, the HTTP should be the default choice, primarily for the ease of use and the more attractive pricing model. This would get rid of https://github.com/winglang/wing/issues/4091 and most of the follow up tasks of https://github.com/winglang/wing/pull/2904 by using baked in capabilities of the HTTP API.

### Implementation Notes

In [Slack](https://winglang.slack.com/archives/C048QCN2XLJ/p1693943660561259), the option of "implementations" for SDK resources was brought up - where there might be multiple implementations per resource / target. Don't have any details on this, perhaps someone else could elaborate on this.

### Component

SDK

### Community Notes

garysassano:
Closes #4097


libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -390,70 +390,105 @@ class WingRestApi extends Construct {
...
 
-  /**
-   * Creates a OpenApi extension object for the endpoint and handler
-   * @param handler Lambda function to handle the endpoint
-   * @returns OpenApi extension object for the endpoint and handler
-   */
   private createApiSpecExtension(handler: Function) {
     const extension = {
       "x-amazon-apigateway-integration": {
         uri: `arn:aws:apigateway:${this.region}:lambda:path/2015-03-31/functions/${handler.arn}/invocations`,
```
skorfmann, line 488: I believe this might need to change to `arn:aws:lambda:us-east-2:123456789012:function:HelloWorld` according to https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-open-api.html

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -390,70 +390,105 @@ class WingRestApi extends Construct {
...
         produce: () => {
-          const injectGreedy404Handler = (openApiSpec: OpenApiSpec) => {
-            openApiSpec.paths = {
-              ...openApiSpec.paths,
-              ...defaultResponse,
-            };
-            return openApiSpec;
-          };
-          return JSON.stringify(injectGreedy404Handler(props.apiSpec));
+          props.apiSpec = "random assignment so compiler doesn't complain";
```
skorfmann, line 463: ?

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -390,70 +390,105 @@ class WingRestApi extends Construct {
...
-      deploymentId: this.deployment.id,
+    this.stage = new Apigatewayv2Stage(this, "stage", {
+      apiId: this.api.id,
+      name: "$default",
+      autoDeploy: true,
     });
 
-    //should be exported from here, otherwise won't be mapped to the right token
-    this.url = this.stage.invokeUrl;
+    this.url = `https://${this.api.id}.execute-api.${this.region}.amazonaws.com`;
```
skorfmann, line 476: why not https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apigatewayv2_api#api_endpoint ?

skorfmann (commented):
just looked at the code an left a few comments. 

Also, I think it's important to get a clear direction for the topic of [feature mismatch](https://github.com/winglang/wing/issues/4097#issuecomment-1712186304) between Rest and HTTP, since this might be a blocker. @ekeren what's your take on active tracing (or the lack of it) in the context of an API Gateway, would this turn into a road block in the near future?

libs/wingsdk/src/target-tf-aws/api.ts
```diff
@@ -390,70 +390,105 @@ class WingRestApi extends Construct {
...
+    //         allowOrigins: ["*"],
+    //       },
+    //     }
+    //   : OAS30_BASE;
+
+    const apigwCorsConfig = props.cors
+      ? {
+          corsConfiguration: {
+            // renamed allowOrigin to allowOrigins
+            allowOrigins: props.cors.allowOrigins
```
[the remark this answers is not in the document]
garysassano: I also found it weird that AWS renamed the header, but I still prefered to have matching names.

libs/wingsdk/src/target-tf-aws/api.ts
garysassano: Didn't want to comment that, and compiler throws errors if you have unused stuff in the code.

garysassano: Some leftover code from previous tests.

libs/wingsdk/src/target-tf-aws/api.ts
garysassano: I updated the code. The `_addToSpec` method fails to update the spec of the ApiGw for some reasons.

garysassano: Btw, I've just tried to export an OAS30 and the format was indeed correct.

garysassano: Ok, I did some tests and both formats seem to work, so we can just use the lambda function ARN since that's shorter.

libs/wingsdk/src/target-tf-aws/api.ts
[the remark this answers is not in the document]
skorfmann: Given that Wing is intending to abstract the cloud resources, I wouldn't follow whatever the AWS platform is doing by default. At the moment there's also the `local` target and in the future gcp, azure and potentially more targets for the API resource

Other than that, since `allowOrigin` is already released, this would be a breaking change. Doesn't mean this can't change, but don't see a strong case to support this at the moment.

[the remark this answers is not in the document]
garysassano: Ok, I named it back to `allowOrigin`.

Issue #3897, opened 2023-08-20
Itermittant queue spec test permission failure


MarkMcCulloh:
### I tried this:

See https://github.com/winglang/wing/actions/runs/5917587426/job/16046054892

### This happened:

```
fail ┌ pop.tfaws » root/Test.NUImFLl7-D/env0/test:pop
     │ Error: Invoke failed with message: "Unhandled". Full error: "{"errorType":"Error","errorMessage":"assertion failed: first == \"Foo\"","trace":["Error: assertion failed: first == \"Foo\"","    at /var/task/index.js:52:23","    at $Closure1.handle (/var/task/index.js:53:15)","    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)","    at async exports.handler (/var/task/index.js:21513:10)"]}"
```

### I expected this:

Always pass

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
@MarkMcCulloh Is there an error in AWS related to permissions specifically?

MarkMcCulloh:
https://github.com/winglang/wing/actions/runs/6028244874/job/16355140792

```
[...]

PR #4173, opened 2023-09-13
fix(sdk): cloud.Queue is not FIFO by default

Chriscbr:
Related to https://github.com/winglang/wing/issues/3897

We have a flakey test in our CI that validates part of the functionality of `cloud.Queue`. It turns out, our test was assuming that the queue messages are received in FIFO order (first-in-first-out), which wasn't something we meant to guarantee (see the original issue about `queue.pop()` [here](https://github.com/winglang/wing/issues/1385)). This PR fixes the test, and updates the simulator implementation to reflect the fact that messages can be received out of order.

(Why not FIFO by default? For many distributed apps, the exact order of messages isn't critical, and relaxing this constraint makes it easier to handle higher throughputs of messages. That said, adding a `fifo` option to `cloud.Queue` would be reasonable as a future enhancement)

## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/docs/04-standard-library/01-cloud/queue.md
+++ b/docs/docs/04-standard-library/01-cloud/queue.md
@@ -20,6 +20,7 @@ sidebar_position: 1
 
 The `cloud.Queue` resource represents a data structure for holding a list of messages.
 Queues are typically used to decouple producers of data and the consumers of said data in distributed systems.
+Queues by default are not FIFO (first in, first out) - so the order of messages is not guaranteed.
 
 ## Usage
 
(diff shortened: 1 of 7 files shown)


staycoolcall911 (approved):

hasanaburayyan (approved):
Do you mind adding a quick snippet to https://www.winglang.io/docs/standard-library/cloud/queue about queues not being FIFO?

I was going to open a PR but wonder if we can just sneak it in here? I dont mind doing it separately either.

garysassano:
>That said, adding a fifo option to cloud.Queue would be reasonable as a future enhancement

Wouldn't it be better to add this option already? So we can reuse existing code instead of eliminating it.

PR #4176, opened 2023-09-14
fix(compiler): unable to use phase-independent types from jsii

### LINKED ISSUE #4102: can't instantiate JSII objects in inflight code
### I tried this:

running this code:

```ts
bring "aws-cdk-lib" as awscdk;
bring cloud;

// compiles
new awscdk.aws_lambda.Code();

  // Unknown symbol "$inflight_init"
  new awscdk.aws_lambda.Code();
});
```

### This happened:

got this error:

`Unknown symbol "$inflight_init"`

### I expected this:

to compile

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
Fixes #4102

`bring`ed ("brought?") JSII libs are effectively preflight bindings, so they need to go through lifting even when referencing a phase-indepenant type from them inflight. I feel like a utility for this could probably be hoisted into the sdk, but the approach here seems to work.

Here we rely on the module resolution logic being somewhat shared between preflight execution and bundling.


libs/wingc/src/lifting.rs
```diff
@@ -281,8 +282,28 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
...
+					NamespaceKind::JSII { fqn } => format!(
+						"{{ _toInflightType: () => `require(\"{fqn}\").{}` }}",
+						node
+							.fields
+							.iter()
+							.map(|f| f.name.clone())
+							.collect::<Vec<String>>()
+							.join(".")
+					),
+				}
```
yoav-steinberg, line 303: I'd really like us to find a way to base the "lifting" logic on the type and not comparing names. Then implement this generically in the SDK. Maybe when the JSifier outputs the call to `_toInflightType` is should call some SDK function instead `toLifterType` that'll check if this object inherits `std.Resouce` and in that case just call its `_toInflightType` and if not inline the `require(...)` above.

MarkMcCulloh: I wasn't sure what you mean by "comparing names". Like checking for the WINGSDK_ASSEMBLY_NAME? That's avoidable.

In any case, I've refactored to put more logic into the SDK. Unfortunately it doesn't look much better because SDK needs a lot of information passed to it. Lemme know what you think about it.

yoav-steinberg: By "comparing names" I meant checking for `WINGSDK_ASSEMBLY_NAME` and treating it differently than any other JSII lib. And I also meant having different kinds of namespaces (`NamespaceKind`) just for the purpose of capturing types inside that namespace differently.

I'll take a look at your last change so we can continue this discussion, sorry this review is taking a while.

libs/wingc/src/lifting.rs
```diff
@@ -281,8 +281,30 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
...
+							fqn = format!("{}/{}", fqn, root_namespace.name);
+						};
+						let type_path = node
+							.fields
+							.iter()
+							.map(|f| f.name.clone())
+							.collect::<Vec<String>>()
+							.join(".");
+						let udt = self.jsify_udt(&node);
+						format!("$stdlib.core.toLiftableModuleType({udt}, \"{fqn}\", \"{type_path}\")",)
```
yoav-steinberg, line 302: Since `toLiftableModuleType` returns the input type if it has a `toInflightType` method then why can't we call it for **all** UDTs (`node`s)? 
Can we change the code here to always calculate `fqn` and `type_path` then always call `toLiftableModuleType`? Then we don't need the `NamespaceKind`. 
Finally, perhaps, move this into `jsify_udt` to keep our JS generating code under the `jsify_*` prefix.

MarkMcCulloh: > Since toLiftableModuleType returns the input type if it has a toInflightType method then why can't we call it for all UDTs (nodes)?

Not all UDTs have a fqn representation right now. To be clear, fqn here must refer to something `require`-able. For namespaces in JSII libs this could be a npm package name or path, whereas for wing type it would have to be a path to a js file. I'm not sure we currently have a way to get the path to `require` a wing namespace, and it's not necessary since wing types always have toInflightType.

> move this into jsify_udt to keep our JS generating code under the jsify_* prefix.

Can do

libs/wingc/src/type_check.rs
```diff
@@ -273,6 +273,17 @@ pub struct Namespace {
...
+
+	/// Where we can resolve this namespace from
+	pub module_path: ResolveSource,
+}
+
+#[derive(Debug)]
+pub enum ResolveSource {
+	/// A wing file within the source tree for this compilation
+	WingFile,
+	/// Something like an external JSII module. This will will the spec for the module.
```
yoav-steinberg, line 285: >  This will will the spec for the module.

??

libs/wingc/src/type_check/lifts.rs
```diff
@@ -107,7 +107,10 @@ impl Lifts {
 			return;
 		}
 
-		let token = self.render_token(code);
+		let token = match lifted_thing {
+			Liftable::Expr(_) => self.render_token(code),
+			Liftable::Type(t) => self.render_token(&format!("{}", t)),
```
yoav-steinberg, line 112: Why do we now ignore the passed in `code` in this case?

yoav-steinberg (approved):

libs/wingc/src/type_check/lifts.rs
MarkMcCulloh: It resulted in unnecessarily verbose js where `$stdlib.core.toLiftableModuleType` is the code for the type.

libs/wingc/src/type_check.rs
MarkMcCulloh: idk what you mean, makes perfect sense

(jk, will fix)

PR #4194, opened 2023-09-17
chore(sdk): add Max-Age CORS header for `cloud.Api`

garysassano:
Adding missing CORS header.


libs/wingsdk/src/cloud/api.ts
```diff
@@ -147,6 +153,12 @@ type CorsOptionsResponseHeaders = {
    * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods
    */
   "Access-Control-Allow-Methods": string;
+
+  /**
+   * Indicates how long the results of a preflight request can be cached.
+   * developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
```
Chriscbr, line 159:
```suggestion
   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
```

libs/wingsdk/src/cloud/api.ts
```diff
@@ -46,6 +46,12 @@ export interface ApiCorsOptions {
    * @default - false
    */
   readonly allowCredentials?: boolean;
+
+  /**
+   * Number of seconds that the browser should cache preflight request results.
+   * @default - 300
+   */
+  readonly maxAge?: number;
```
Chriscbr, line 54: This would be a good use case for a `duration`, no?
```suggestion
  /**
   * Duration that the browser should cache preflight request results.
   * @default - 300 seconds
   */
  readonly maxAge?: Duration;
```

libs/wingsdk/src/cloud/api.ts
```diff
@@ -46,6 +46,12 @@ export interface ApiCorsOptions {
    * @default - false
    */
   readonly allowCredentials?: boolean;
+
+  /**
+   * How long the browser should cache preflight request results.
+   * @default - 300 seconds
+   */
+  readonly maxAge?: Duration;
```
Chriscbr, line 54: Can we update the unit tests to validate configuring this option works / the default works? https://github.com/winglang/wing/blob/b71cae19d036459e5fee1a6084e229bd34bc7a01/libs/wingsdk/test/target-sim/api.test.ts#L601-L663

I think since the value is a `Duration` you'll need to call `.seconds()` on it in the implementation, or something like that.

examples/tests/valid/website_with_api.main.w
```diff
@@ -11,7 +11,8 @@ let api = new cloud.Api(
     allowMethods: [cloud.HttpMethod.GET, cloud.HttpMethod.POST, cloud.HttpMethod.OPTIONS],
     allowHeaders: ["Content-Type"],
     allowCredentials: false,
-    exposeHeaders: ["Content-Type"]
+    exposeHeaders: ["Content-Type"],
+    maxAge: 600
```
Chriscbr, line 15:
```suggestion
    maxAge: 600s
```

libs/wingsdk/src/cloud/api.ts
garysassano: I added the test under the preflight request (`OPTIONS` method).

Chriscbr (approved):

Issue #2723, opened 2023-05-31
CLI / Compiler: Make it more obvious that Secrets are self managed by the user


skorfmann:
### Feature Spec

Something like this at compile time would be nice. Might get a bit noisy when doing a lot of compilations, but I think I'd error on the side of being too noisy to improve the new user experience for now.

```
wing compile -t tf-aws food.w
The following secrets are defined in the application:
- USER_LOCATION_NEW_3
- GOOGLE_PLACES_API_KEY_NEW_3
Please make sure to create them ahead of deployment, e.g. "aws secretsmanager create-secret --name my-api-key --secret-string 1234567890"  Find out more here: https://docs.winglang.io/resources/secret
```

### Use Cases

As a first step towards an improved UX around Secrets, it would be great to just list and print the referenced secrets at compile time. With the right messaging, this would make it obvious to the user that there's some action necessary ahead of plan / deploy time. Right now a user will only see an error when performing a `terraform plan / apply`:

```
➜  food.tfaws git:(main) ✗ terraform plan
╷
│ Error: Secrets Manager Secret "GOOGLE_PLACES_API_KEY_NEW_3" not found
│
│   with data.aws_secretsmanager_secret.root_WhereToEat_GOOGLEPLACESAPIKEYNEW3_994DAE8C,
│   on main.tf.json line 38, in data.aws_secretsmanager_secret.root_WhereToEat_GOOGLEPLACESAPIKEYNEW3_994DAE8C:
│   38:       },
│
╵
╷
│ Error: Secrets Manager Secret "USER_LOCATION_NEW_3" not found
│
│   with data.aws_secretsmanager_secret.root_WhereToEat_USERLOCATIONNEW3_A657CC68,
│   on main.tf.json line 47, in data.aws_secretsmanager_secret.root_WhereToEat_USERLOCATIONNEW3_A657CC68:
│   47:       }
│
```

The Simulator / Wing console is providing some feedback around this already:

![Screenshot 2023-05-31 at 11 15 09](https://github.com/winglang/wing/assets/136789/f0ccfc3b-a266-45ea-9940-cc3976ba4fbb)

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

staycoolcall911:
Nice enhancement! Should be easy to add

ekeren:
Keep

PR #4195, opened 2023-09-17
feat(sdk)!: `cloud.Secret` enhancement

### LINKED ISSUE #2726: SDK: Enhance Secret / Config Handling
### Feature Spec

Either change the existing implementation of `Secret` to use a SSM Parameter or alternative introduce a new Construct along the lines of `Config` (e.g. `Config.Parameter` / `Config.Secret`) and make it clear by documentation that these serve different use-cases.

### Use Cases

Right now https://docs.winglang.io/resources/secret are implemented as an AWS [Secrets Manager](https://docs.aws.amazon.com/secretsmanager/latest/userguide/intro.html) resource in the targets `tf-aws` and `awscdk`. It might be better off as an [SSM Parameter](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html).

#### The use case I was expecting for the Wing Secret construct:

- Handling Envs / Simple config values (debatable, might deserve a dedicated construct)
- Handling sensitive 3rd party configuration values (i.e. API Keys)

#### The use case I think the Wing Secret construct serves best right now:

- Handling Secrets for other provisioned resources (e.g. VMs, DBs, ...) where one would want capabilities like generating random values or automatically rotating credentials

The different use cases are reflected in the pricing of the resources. While the standard Parameter Store values are [free](https://aws.amazon.com/systems-manager/pricing/#Parameter_Store) (up to 10k per AWS account), the Secrets Manager value [costs](https://aws.amazon.com/systems-manager/pricing/) USD 0.40 / each per month.

### Implementation Notes

Here's a related Slack [conversation](https://winglang.slack.com/archives/C047QFSUL5R/p1685479604135819) 

### Component

SDK

### Community Notes

garysassano:
Closes #2726

### DIFF
--- a/libs/wingsdk/src/cloud/secret.ts
+++ b/libs/wingsdk/src/cloud/secret.ts
@@ -21,7 +21,7 @@ export interface SecretProps {
    *
    * @default - a new secret is provisioned with a generated name
    */
-  readonly name?: string;
+  readonly name: string;
 }
 
 /**
@@ -37,12 +37,12 @@ export abstract class Secret extends Resource {
   public static _newSecret(
     scope: Construct,
     id: string,
-    props: SecretProps = {}
+    props: SecretProps
   ): Secret {
     return App.of(scope).newAbstract(SECRET_FQN, scope, id, props);
   }
 
-  constructor(scope: Construct, id: string, props: SecretProps = {}) {
+  constructor(scope: Construct, id: string, props: SecretProps) {
     super(scope, id);
 
     Node.of(this).title = "Secret";
--- a/examples/tests/sdk_tests/secret/main.w
+++ b/examples/tests/sdk_tests/secret/main.w
@@ -1,0 +1,3 @@
+bring cloud;
+
+let secret = new cloud.Secret();
--- a/examples/tests/sdk_tests/secret/main.w
+++ b/examples/tests/sdk_tests/secret/main.w
@@ -1,3 +1,3 @@
 bring cloud;
 
-let secret = new cloud.Secret();
+let secret = new cloud.Secret(name: "test");
--- a/libs/wingsdk/src/target-awscdk/secret.ts
+++ b/libs/wingsdk/src/target-awscdk/secret.ts
@@ -19,7 +19,7 @@ export class Secret extends cloud.Secret {
   private readonly secret: ICdkSecret;
   private readonly arnForPolicies: string;
 
-  constructor(scope: Construct, id: string, props: cloud.SecretProps = {}) {
+  constructor(scope: Construct, id: string, props: cloud.SecretProps) {
     super(scope, id, props);
 
     if (props.name) {
--- a/libs/wingsdk/src/target-sim/secret.ts
+++ b/libs/wingsdk/src/target-sim/secret.ts
@@ -14,7 +14,7 @@ import { BaseResourceSchema } from "../testing/simulator";
  */
 export class Secret extends cloud.Secret implements ISimulatorResource {
   private readonly name: string;
-  constructor(scope: Construct, id: string, props: cloud.SecretProps = {}) {
+  constructor(scope: Construct, id: string, props: cloud.SecretProps) {
     super(scope, id, props);
 
     this.name =
--- a/libs/wingsdk/src/target-tf-aws/secret.ts
+++ b/libs/wingsdk/src/target-tf-aws/secret.ts
@@ -1,19 +1,11 @@
 import { TerraformOutput } from "cdktf";
 import { Construct } from "constructs";
 import { Function } from "./function";
-import { SsmParameter } from "../.gen/providers/aws/ssm-parameter";
+import { DataAwsSsmParameter } from "../.gen/providers/aws/data-aws-ssm-parameter";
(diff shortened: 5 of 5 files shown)


garysassano:
I think that the `cloud.Secret` interface needs a complete redesign because in its current state I cannot implement it correctly.

libs/wingsdk/src/cloud/secret.ts
```diff
@@ -79,15 +87,15 @@ export interface ISecretClient {
    * @Returns the secret value as string.
    * @inflight
    */
-  value(options?: GetSecretValueOptions): Promise<string>;
+  getValue(options?: ISecretGetValueOptions): Promise<string>;
```
eladb, line 90: Why did we rename this?

I think this should still be `value()`

libs/wingsdk/src/cloud/secret.ts
```diff
@@ -53,14 +58,17 @@ export abstract class Secret extends Resource {
...
+      SecretInflightMethods.GET_VALUE_JSON,
+    ];
   }
 }
 
 /**
  * Options when getting a secret value
  */
-export interface GetSecretValueOptions {
+export interface ISecretGetValueOptions {
```
eladb, line 71:
```suggestion
export interface SecretGetValueOptions {
```

This is a jsii struct, not an interface.

libs/wingsdk/src/cloud/secret.ts
```diff
@@ -22,6 +22,11 @@ export interface SecretProps {
    * @default - a new secret is provisioned with a generated name
    */
   readonly name?: string;
+
+  /**
+   * The secret's value.
+   */
+  readonly value: string;
```
eladb, line 29: No... we don't allow secret in the code.

eladb (changes requested):
Let's make sure we discuss these changes. Seems like the wrong way

libs/wingsdk/src/cloud/secret.ts
garysassano: Looking at all the other resources, we use verbs for defining the methods, so it made logically more sense to me. Having only `value` alone is something I would expect from a class attribute.

libs/wingsdk/src/cloud/secret.ts
garysassano: Ok, I thought that the current implementation was a temporary workaround; I didn't understand that it was intentional and that Wing shouldn't allow you to handle secrets in the code.

garysassano: Btw, if we decide that an user can only import existing secrets, then the `name` argument will be mandatory to specify which one you want (currently it is optional and a new secret is created if a name isn't provided).

@eladb Are you fine with this change?

skorfmann:
Continuing discussion from [slack](https://winglang.slack.com/archives/C048QCN2XLJ/p1694962090896659)

The original issue is about either adapting the the existing Secret resource or adding a new Config resource. So, I wouldn't fundamentally change the existing workflow. 

The value fetching itself is currently done [inflight](https://github.com/winglang/wing/blob/037fb0e6ace1c140cfab58a16d8bd0336fa89143/libs/wingsdk/src/shared-aws/secret.inflight.ts) which should not change.

When looking at the [secrets data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/secretsmanager_secret) it's not persisting any sensitive data in tf state. The [ssm data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ssm_parameter) in contrast is persisting in state. However, it has the [with_decryption](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ssm_parameter#with_decryption) attribute which could be set to `false` to avoid leaking in tf state. Still not ideal, and I'm wondering what these data sources are being used for anyway. I believe that only the `arn` is being used. So the only benefit we get seems to be some sort deploy time validation that the ARN exists.

One other aspect which crossed my mind since would be a hard breaking change for existing apps: Is there a concept of migrations for stateful resources? Or at least a way to let users know that there's action required on their end when upgrading?

Last but not least: The CLI based workflow of managing these values can certainly be improved, e.g. https://github.com/winglang/wing/issues/2723 but there's more headroom of course. Perhaps this change would make more sense together with a more elaborate CLI workflow to manage the SSM parameters? Would certainly be a huge improvement in terms of DX.

eladb:
I feel we need to continue the conversation over the issue about what are the requirements from this feature and how to implement them.

I am personally okay with using Secrets Manager in AWS for managing secrets, but maybe that's something we can add as a platform configuration (copying @hasanaburayyan who is working on defining the platform APIs right now).

For secrets I was also thinking we should allow storing them in `.env` file and/or environment variables for local development. Seems like that's what the cool kids are doing nowadays.

hasanaburayyan:
> but maybe that's something we can add as a platform configuration

Id say this is most definitely something that should be configurable through platforms. I mean everything should be able to be configured through the platforms.

I can imagine especially for secrets management a lot of orgs will be very opinionated on how they store secrets, wether its AWS Secrets Manager, Vault, Thycotic, etc..

PR #4197, opened 2023-09-17
feat(sdk): cloud.Function for target tf-gcp

### LINKED ISSUE #614: cloud.Function - GCP implementation

subh-cs:
Fixed #614 
This adds support of cloud function resource for target tf-gcp using cdktf

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,127 @@
...
+      environmentVariables: props.env ?? {},
+    });
+
+    // create the permissions
+    new CloudfunctionsFunctionIamMember(this, "Function_IAM_Member", {
+      project: app.projectId,
+      region: app.storageLocation,
+      cloudFunction: this.function.name,
+      role: "roles/cloudfunctions.invoker",
+      member: "allUsers",
```
skorfmann, line 111: I'm not really familiar with GCP, what does `allUsers` in this context mean?

subh-cs: > I'm not really familiar with GCP, what does `allUsers` in this context mean?

This is for public function and IAM entry for all users to invoke the function

For example if we do this
`member = "user:myFunctionInvoker@example.com"`
this will only allow a specific user to have the permission to invoke

You can take a look at this tf doc
[Public Access](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions_function.html#example-usage---public-function)
[Single User Access](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions_function.html#example-usage---single-user)

hasanaburayyan: We do not want to allow all users to invoke the function. Permissions to the function should be granted during the bind method.

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,127 @@
...
+    });
+  }
+
+  // TODO: implement
+  public _toInflight(): string {
+    throw new Error("Method not implemented.");
+  }
```
hasanaburayyan, line 118:
```suggestion
  public _toInflight(): string {
    throw new Error("cloud.Function cannot be used as an Inflight resource on GCP yet");
  }
```

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  public addPermission(_principal: string, _action: string): void {
+    throw new Error("Method not implemented.");
+  }
```
hasanaburayyan, line 122: I dont believe this is required
```suggestion

```

hasanaburayyan (changes requested):
My GCP account is not active, working on restoring to test his. I didnt complete a comprehensive review but left a few comments so far.

subh-cs:
> So I tried testing this in GCP and get the following error when creating the Function: `│ Error: Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.`
> 
> The bucket for the function was created just fine, but the function itself failed. I think this is because right now we require the users to provide `GOOGLE_PROJECT_ID` and `GOOGLE_STORAGE_LOCATION` env vars. But I think we will also require a `region` or `zone` to be provided for this.
> 
> Im tagging @Chriscbr for a sanity check.

@hasanaburayyan, sorry, I missed those parameters, just added them. 
So do we really need to pass them as env? 
```js
        // create the cloud function
        this.function = new CloudfunctionsFunction(this, "Default_Function", {
            name: ResourceNames.generateName(this, FUNCTION_NAME_OPTS),
            description: "This function was created by Wing",
            project: app.projectId,                  //<----------- GOOGLE_PROJECT_ID
            region: app.storageLocation,       //<----------- GOOGLE_STORAGE_LOCATION
            runtime: "nodejs16",
            availableMemoryMb: props.memory ?? 128,
            sourceArchiveBucket: this.bucket.name,
            sourceArchiveObject: this.bucketObject.name,
            entryPoint: "handler",
            triggerHttp: true,
            timeout: props.timeout?.seconds ?? 60,
            environmentVariables: props.env ?? {},
        });
```
        
Also I've removed the permission for invoking function from constructor and will add them within bind method later in this PR.

hasanaburayyan:
> > So I tried testing this in GCP and get the following error when creating the Function: `│ Error: Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.`
> > The bucket for the function was created just fine, but the function itself failed. I think this is because right now we require the users to provide `GOOGLE_PROJECT_ID` and `GOOGLE_STORAGE_LOCATION` env vars. But I think we will also require a `region` or `zone` to be provided for this.
> > Im tagging @Chriscbr for a sanity check.
> 
> @hasanaburayyan, sorry, I missed those parameters, just added them. So do we really need to pass them as env?
> 
> ```js
>         // create the cloud function
>         this.function = new CloudfunctionsFunction(this, "Default_Function", {
>             name: ResourceNames.generateName(this, FUNCTION_NAME_OPTS),
>             description: "This function was created by Wing",
>             project: app.projectId,                  //<----------- GOOGLE_PROJECT_ID
>             region: app.storageLocation,       //<----------- GOOGLE_STORAGE_LOCATION
>             runtime: "nodejs16",
>             availableMemoryMb: props.memory ?? 128,
>             sourceArchiveBucket: this.bucket.name,
>             sourceArchiveObject: this.bucketObject.name,
>             entryPoint: "handler",
>             triggerHttp: true,
>             timeout: props.timeout?.seconds ?? 60,
>             environmentVariables: props.env ?? {},
>         });
> ```
> 
> Also I've removed the permission for invoking function from constructor and will add them within bind method later in this PR.

Okay cool, I should checked. I was worried that `region` and `storageLocation` were different things. will retest.

Chriscbr:
> So I tried testing this in GCP and get the following error when creating the Function: `│ Error: Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.`
> 
> The bucket for the function was created just fine, but the function itself failed. I think this is because right now we require the users to provide `GOOGLE_PROJECT_ID` and `GOOGLE_STORAGE_LOCATION` env vars. But I think we will also require a `region` or `zone` to be provided for this.
> 
> Im tagging @Chriscbr for a sanity check.

Yeah I don't remember the exact details, but I want to say that when I implemented Bucket I wasn't sure if there was a standard list of GCP regions or not, hence the name `GOOGLE_STORAGE_LOCATION`.

Maybe if it's possible, we can let the user configure a single region for all resources (with an environment variable like `WING_GCP_REGION` or `WING_GCP_ZONE`) and use that value for all resources , including buckets, functions, etc.

subh-cs:
> > So I tried testing this in GCP and get the following error when creating the Function: `│ Error: Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.`
> > The bucket for the function was created just fine, but the function itself failed. I think this is because right now we require the users to provide `GOOGLE_PROJECT_ID` and `GOOGLE_STORAGE_LOCATION` env vars. But I think we will also require a `region` or `zone` to be provided for this.
> > Im tagging @Chriscbr for a sanity check.
> 
> Yeah I don't remember the exact details, but I want to say that when I implemented Bucket I wasn't sure if there was a standard list of GCP regions or not, hence the name `GOOGLE_STORAGE_LOCATION`.
> 
> Maybe if it's possible, we can let the user configure a single region for all resources (with an environment variable like `WING_GCP_REGION` or `WING_GCP_ZONE`) and use that value for all resources , including buckets, functions, etc.

@hasanaburayyan @Chriscbr I was wrong. region is different than storageLocation. I just added support for that by my recent commit. Also I've added tests for that as well. 
```js
this.function = new CloudfunctionsFunction(this, "Default_Function", {
      name: ResourceNames.generateName(this, FUNCTION_NAME_OPTS),
      description: "This function was created by Wing",
      project: app.projectId,
      region: app.region,    //-------------> GOGGLE_REGION
      runtime: "nodejs16",
      availableMemoryMb: props.memory ?? 128,
      sourceArchiveBucket: this.bucket.name,
      sourceArchiveObject: this.bucketObject.name,
      entryPoint: "handler",
      triggerHttp: true,
      timeout: props.timeout?.seconds ?? 60,
      environmentVariables: props.env ?? {},
    });
```
also I've added many test including this as well, which is passing successfully
```js

test("can read region from environment variable", () => {
  // GIVEN
  const props = {
    outdir: mkdtemp(),
    projectId: "projectId",
    region: undefined as any,
    storageLocation: "US",
    entrypointDir: __dirname,
  };
  const region = "us-central1";
  process.env.GOOGLE_REGION = region;
  let app: tfgcp.App;

  // THEN
  expect(() => (app = new tfgcp.App(props))).not.toThrow();
  expect(app!.region).toEqual(region);
});

```

hasanaburayyan:
Successfully tested. Will re-review shortly

image

libs/wingsdk/src/target-tf-gcp/app.ts
```diff
@@ -15,6 +16,12 @@ export interface AppProps extends CdktfAppProps {
    */
   readonly projectId: string;
 
+  /**
+   * The Google Cloud region, used for all resources.
+   * @see https://cloud.google.com/functions/docs/locations
+   */
+  readonly region: string;
+
```
hasanaburayyan, line 24: I just checked we are able to use the same region for both storage and functions, so maybe we can just combine these two properties.

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+    const randomId = new Id(this, "Id", {
+      byteLength: 4, // 4 bytes = 8 hex characters
+    });
+
+    // create the bucket
+    this.bucket = new StorageBucket(this, "Function_Bucket", {
+      name: bucketName + "-" + randomId.hex,
+      location: app.storageLocation,
+      project: app.projectId,
+    });
```
hasanaburayyan, line 78: Can we just use the Bucket resource to create this bucket? See https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-tf-azure/function.ts#L73

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+    // so we must generate a random string to append to the bucket name.
+
+    // The random string must be managed in Terraform state so that it doesn't
+    // change on every subsequent compile or deployment.
+    const randomId = new Id(this, "Id", {
+      byteLength: 4, // 4 bytes = 8 hex characters
+    });
+
+    // create the bucket
+    this.bucket = new StorageBucket(this, "Function_Bucket", {
```
hasanaburayyan, line 74:
```suggestion
    this.bucket = new StorageBucket(this, "FunctionBucket", {
```

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+      name: bucketName + "-" + randomId.hex,
+      location: app.storageLocation,
+      project: app.projectId,
+    });
+
+    // put the executable in the bucket as an object
+    this.bucketObject = new StorageBucketObject(
+      this,
+      "Function_Object_Bucket",
```
hasanaburayyan, line 83:
```suggestion
      "FunctionObjectBucket",
```

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+        source: asset.path,
+      }
+    );
+
+    // memory limits must be between 128 and 8192 MB
+    if (props?.memory && (props.memory < 128 || props.memory > 8192)) {
+      throw new Error(
+        "Memory must be between 128 and 8192 MB for GCP Cloud Functions"
+      );
+    }
```
hasanaburayyan, line 96: Lets check for this earlier so we can bail out early. (move it above the resource creations)

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // timeout must be between 1 and 540 seconds
+    if (
+      props?.timeout &&
+      (props.timeout.seconds < 1 || props.timeout.seconds > 540)
+    ) {
+      throw new Error(
+        "Timeout must be between 1 and 540 seconds for GCP Cloud Functions"
+      );
+    }
```
hasanaburayyan, line 106: Move this up as well

libs/wingsdk/src/target-tf-gcp/function.ts
hasanaburayyan: Is description required? If not lets remove.

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+      timeout: props.timeout?.seconds ?? 60,
+      environmentVariables: props.env ?? {},
+    });
+  }
+
+  public get functionName(): string {
+    return this.function.name;
+  }
+
+  // TODO: implement
```
hasanaburayyan, line 129:
```suggestion
  // TODO: implement with https://github.com/winglang/wing/issues/1282
```

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -0,0 +1,139 @@
...
+      props?.timeout &&
+      (props.timeout.seconds < 1 || props.timeout.seconds > 540)
+    ) {
+      throw new Error(
+        "Timeout must be between 1 and 540 seconds for GCP Cloud Functions"
+      );
+    }
+
+    // create the cloud function
+    this.function = new CloudfunctionsFunction(this, "Default_Function", {
```
hasanaburayyan, line 109:
```suggestion
    this.function = new CloudfunctionsFunction(this, "DefaultFunction", {
```

hasanaburayyan (changes requested):
@subh-cs 🔥 I think this is really close to shipping I added a few comments most of them are nits

libs/wingsdk/test/target-tf-gcp/function.test.ts
```diff
@@ -0,0 +1,143 @@
+import * as cdktf from "cdktf";
+import { test, expect } from "vitest";
+import { Function } from "../../src/cloud";
+import { Duration } from "../../src/std";
+import * as tfgcp from "../../src/target-tf-gcp";
+import { Testing } from "../../src/testing";
```
MarkMcCulloh, line 6:
```suggestion
import { Testing } from "../../src/simulator";
```

hasanaburayyan (approved):
@subh-cs thanks soo much for sticking this one and making the requrested changed. I know it was a bit of a pain to test but its a huge building block to getting GCP supported ❤️ 

I re-tested with changes and everything still works.

PR #4207, opened 2023-09-18
feat(cli): add `--test-filter <regex>` to `wing test`

### LINKED ISSUE #3086: `wing test` test name filter
### Feature Spec

When running `wing test --test-filter <regex>`, only the test names that match that pattern will be run

### Use Cases

- Iterate on specific tests
- Provides a way for systems like VSCode's Test API to run specific tests

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

garysassano:
Closes #3086


apps/wing/src/commands/test.ts
```diff
@@ -246,6 +247,21 @@ function noCleanUp(synthDir: string) {
   );
 }
 
+function filterTestsByRegex(tests: string[], regexString: string): string[] {
+  let regex = new RegExp(regexString);
+
+  return tests.filter((test) => {
+    // Extract test name from the string
```
eladb, line 254: can you add an example of the contents of the string in a comment?

eladb (changes requested):
missing test

apps/wing/src/cli.ts
```diff
@@ -160,6 +160,7 @@ async function main() {
     .option("-p, --plugins [plugin...]", "Compiler plugins")
     .option("-r, --rootId <rootId>", "App root id")
     .option("--no-clean", "Keep build output")
+    .option("--test-filter <regex>", "Run tests that match the provided regex pattern")
```
eladb, line 163: let's add an alias `-f` for this

apps/wing/src/cli.ts
```diff
@@ -159,6 +159,7 @@ async function main() {
     )
     .option("-p, --plugins [plugin...]", "Compiler plugins")
     .option("-r, --rootId <rootId>", "App root id")
+    .option("-f, --test-filter <regex>", "Run tests that match the provided regex pattern")
```
tsuf239, line 162:
```suggestion
    .option("-f, --test-filter <regex>", "Run tests that match the provided regex pattern within the selected entrypoint files")
```

apps/wing/src/commands/test.ts
```diff
@@ -254,10 +271,18 @@ async function testSimulator(synthDir: string, options: TestOptions) {
...
 
+  if (options.testFilter) {
+    testsToRun = filterTestsByRegex(filteredTests, options.testFilter);
+  } else {
+    testsToRun = filteredTests;
+  }
+
   // TODO: run these tests in parallel
-  for (const path of filteredTests) {
+  for (const path of testsToRun) {
```
tsuf239, line 285: it filters only the tests of the simulator, can you please put all the logic in `filterTestsByRegex`, and then call it in each of the targets? 
another option will be putting the logic in`testRunner.listTests()` or `pickOneTestPerEnvironment` (please pick the option that will require you to duplicate the least amount of code),
also, I suspect we randomize test names- can you please verify that?

apps/wing/src/cli.ts
tsuf239: (just because I find it confusing, since we can also run a few files at the same time)

apps/wing/src/commands/test.ts
```diff
@@ -26,6 +26,7 @@ const generateTestName = (path: string) => path.split(sep).slice(-2).join("/");
  */
 export interface TestOptions extends CompileOptions {
   clean: boolean;
+  testFilter?: string;
```
eladb, line 29:
```suggestion
  filter?: string;
```

Always prefer one word if it's clear from the context.

Also, can you please add docstrings?

eladb (dismissed):
still missing test for this

apps/wing/src/commands/test.ts
```diff
@@ -246,16 +247,31 @@ function noCleanUp(synthDir: string) {
...
+      // root/env0/test:<testName>
+      const testName = test.split(":").pop();
+      return testName ? regex.test(testName) : false;
+    });
+  }
+
+  return pickOneTestPerEnvironment(tests);
+}
+
 async function testSimulator(synthDir: string, options: TestOptions) {
```
MarkMcCulloh, line 264: The filtering logic needs to happen across all targets, not just the sim. Without code duplication ideally.

apps/wing/src/commands/test.ts
garysassano: Where exactly should I add docstrings?

apps/wing/src/commands/test.test.ts
```diff
@@ -35,6 +35,16 @@ describe("printing test reports", () => {
   });
 });
 
+describe("test options", () => {
+  test("--test-filter <regex>", () => {
+    const filteredTests = filterTests(EXAMPLE_UNFILTERED_TESTS, "get");
+
+    expect(filteredTests.length).toBe(2);
+    expect(filteredTests[0]).toBe("root/env0/test:get()");
+    expect(filteredTests[1]).toBe("root/env1/test:get:At()");
```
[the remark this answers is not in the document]
tsuf239: Ohhh I saw now that `filterTests` actually returns `pickOneTestPerEnvironment` Can it return just a list of strings? 
(for better unit-testing, being able to test both `filterTests` and `pickOneTestPerEnvironment` separately)

also I would test it with a regex contains : and / to make sure it doesn't catch those in the test names

then the call will look like this: `[test Runner, pickOneTestPerEnvironment(filterTests(tests, filter))];`
or:
```
const filteredTests = filterTests(tests, filter);
[test Runner, pickOneTestPerEnvironment(filteredTests)];
```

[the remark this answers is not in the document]
garysassano: @MarkMcCulloh Are you fine with this change? I would need to basically revert my [past change](https://github.com/winglang/wing/pull/4207/commits/e3280c0a9d10bc428837c5a5923a4b31af9c7c9c) that you asked me to do [here](https://github.com/winglang/wing/pull/4207#discussion_r1332069465).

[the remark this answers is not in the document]
tsuf239: yes :) It is an addition to Mark's comment, we're still not duplicating the code while keeping slightly different "logic components" separated for easier testing :)

garysassano:
Took me a while to figure out, but [this suggested change](https://github.com/winglang/wing/pull/4207#discussion_r1332011201) broke everything. I guess `filter` is a reserved word or it's already used somewhere else. Do you have any insight? @MarkMcCulloh

garysassano:
Took me a while to figure out, but [this suggested change](https://github.com/winglang/wing/pull/4207#discussion_r1332011201) broke everything. I guess `filter` is a reserved word or it's already used somewhere else. Do you have any insight? @MarkMcCulloh

tools/hangar/__snapshots__/test_corpus/sdk_tests/counter/set.test.w_test_sim.md
```diff
@@ -2,11 +2,10 @@
 
 ## stdout.log
 ```log
-pass ─ set.test.wsim » root/env0/test:set    
-pass ─ set.test.wsim » root/env1/test:key set
+pass ─ set.test.wsim (no tests)
```
MarkMcCulloh, line 5: Regression - tests are not running

apps/wing/src/commands/test.ts
```diff
@@ -246,16 +250,34 @@ function noCleanUp(synthDir: string) {
...
+      // Extract test name from the string
+      // root/env0/test:<testName>
+      const firstColonIndex = test.indexOf(":");
+      const testName = test.substring(firstColonIndex + 1);
+      return testName ? regex.test(testName) : false;
+    });
+  }
+
+  return filteredTests;
+}
```
MarkMcCulloh, line 268:
```suggestion
export function filterTests(tests: Array<string>, regexString?: string): Array<string> {
  if (regexString) {
    const regex = new RegExp(regexString);
    return tests.filter((test) => {
      // Extract test name from the string
      // root/env0/test:<testName>
      const firstColonIndex = test.indexOf(":");
      const testName = test.substring(firstColonIndex + 1);
      return testName ? regex.test(testName) : false;
    });
  } else {
    return tests;
  }
}
```

tools/hangar/__snapshots__/test_corpus/sdk_tests/counter/set.test.w_test_sim.md
garysassano: I found the bug: it's this [suggested change](https://github.com/winglang/wing/pull/4207#discussion_r1332011201). Reverting from `filter` to `testFilter` fixed it even though I'm not sure why.

MarkMcCulloh (approved):

PR #4210, opened 2023-09-18
feat(compiler): bring directories

Chriscbr:
This pull request adds support for bringing directories. Bringing a directory will make all types defined in the files in a directory accessible through the alias defined by the user:

```
bring "./aws" as aws;
``` 

If a symbol like "Bucket" is defined multiple times inside files at the same hierarchical level of "aws", an error will be raised saying that there are multiple definitions of "Bucket" inside the directory. If the directory contains subdirectories, it will be available through a child namespace inside `aws`. For example, `aws.storage` will contain types from all `aws/storage/*.w` files.

To achieve this, I had to figure out some way to model symbol lookups in a directory, where there might be multiple symbol environments that have to be looked up before finding a symbol. To that end, I changed `Namespace` so instead of storing a single single environment, it stores a vector of symbol environments, as this seemed like the least invasive change.

Future items:
- https://github.com/winglang/wing/issues/4276

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -266,7 +267,7 @@ pub struct Namespace {
 	pub name: String,
 
 	#[derivative(Debug = "ignore")]
-	pub env: SymbolEnv,
+	pub envs: Vec<SymbolEnvRef>,
```
Chriscbr, line 270: @yoav-steinberg FYI ☝️  - this is how I figured how to add support for multi-file imports, curious if you see any dangers down this implementation path...

libs/wingc/src/parser.rs
```diff
@@ -746,40 +777,76 @@ impl<'s> Parser<'s> {
...
-			if !source_path.is_file() {
-				return self.with_error(
-					format!("Cannot bring module \"{}\": not a file", source_path),
-					statement_node,
-				);
-			}
-			if source_path.ends_with(".main.w\"") {
+
+			// case: .w file
+			if source_path.ends_with(".main.w\"") || source_path.ends_with("main.w\"") {
```
Chriscbr, line 782: TODO: double check we have invalid e2e tests for this

libs/wingc/src/type_check.rs
```diff
@@ -5257,6 +5408,10 @@ where
 {
 	let (message, span) = match lookup_result {
 		LookupResult::NotFound(s) => (format!("Unknown symbol \"{s}\""), s.span()),
+		LookupResult::MultipleFound => (
+			format!("Ambiguous symbol \"{looked_up_object}\""),
+			looked_up_object.span(),
+		),
```
Chriscbr, line 5414: I don't think this error should appear, but I couldn't really be sure so I decided to leave it here just in case

docs/docs/03-language-reference.md
```diff
@@ -1771,6 +1771,24 @@ bring cloud; // from cloud bring * as cloud;
...
+
+It's also possible to import a directory as a module. The module will contain all
+types defined in the directory's files. If the directory has subdirectories,
+they will be available under the corresponding names.
+
+```TS
+bring "./my-module" as myModule;
+
+// from ./my-module/submodule/my-class.w
+new myModule.submodule.MyClass();
```
eladb, line 1789: ❤️

docs/docs/03-language-reference.md
```diff
@@ -1771,6 +1771,24 @@ bring cloud; // from cloud bring * as cloud;
...
 
+To import an individual Wing file as a module, you can specify its path relative
+to the current file:
+
+```TS
+bring "./my-module.w" as myModule;
+```
+
+It's also possible to import a directory as a module. The module will contain all
+types defined in the directory's files. If the directory has subdirectories,
```
eladb, line 1782: This should eventually be "all public types", correct?

I am 100% okay with this not included in the current PR (and I understand why), but definitely worth adding a "Future items" section to the PR description indicating this is the intention.

libs/wingc/src/file_graph.rs
```diff
@@ -44,6 +44,17 @@ impl FileGraph {
 		self.path_to_node_index.contains_key(path)
 	}
 
+	/// Returns a list of the direct dependencies of the given file
+	/// (does not include all transitive dependencies)
```
eladb, line 48: maybe add a comment indicating that `path` must be relative to to the root of the file graph. otherwise this will panic.

libs/wingc/src/jsify.rs
```diff
@@ -138,6 +142,7 @@ impl<'a> JSifier<'a> {
 		let mut output = CodeMaker::default();
 
 		let is_entrypoint_file = source_path == self.entrypoint_file_path;
+		let is_directory = source_path.is_dir();
```
eladb, line 145: should we validate that this is a relative path (without `..`), otherwise the graph won't include it, right?

libs/wingc/src/jsify.rs
```diff
@@ -175,6 +180,24 @@ impl<'a> JSifier<'a> {
 				"new $App({{ outdir: {}, name: \"{}\", rootConstruct: {}, plugins: {}, isTestEnvironment: {}, entrypointDir: process.env['WING_SOURCE_DIR'], rootId: process.env['WING_ROOT_ID'] }}).synth();",
 				OUTDIR_VAR, app_name, ROOT_CLASS, PLUGINS_VAR, ENV_WING_IS_TEST
 			));
+		} else if is_directory {
+			let directory_children = self.source_file_graph.dependencies_of(source_path);
```
eladb, line 184: maybe add a little comment explaining what the output is going to look like

libs/wingc/src/lsp/completions.rs
```diff
@@ -716,27 +716,21 @@ fn get_completions_from_namespace(
 ) -> Vec<CompletionItem> {
 	// If a namespace has a class named "Util", then its members can be accessed directly from
 	// the namespace as a syntactic sugar. e.g. "foo.bar()" is equivalent to "foo.Util.bar()"
-	let util_completions = match namespace.env.lookup_nested_str("Util", None) {
-		LookupResult::Found(kind, _) => match kind {
-			SymbolKind::Type(typeref) => {
+	let mut util_completions = vec![];
+	for ns_env in &namespace.envs {
+		if let LookupResult::Found(kind, _) = ns_env.lookup_nested_str("Util", None) {
```
eladb, line 721: Maybe while we here, move `"Util"` to some const

libs/wingc/src/parser.rs
```diff
@@ -163,41 +163,74 @@ pub fn parse_wing_project(
...
+				if path.is_dir() || path.extension() == Some("w") {
+					files_and_dirs.push(path);
+				}
+			}
+
+			// Sort the files and directories so that we always visit them in the same order
+			files_and_dirs.sort();
+
+			// Parse the file here (since it doesn't exist)
+			let mut tree_sitter_parser = tree_sitter::Parser::new();
```
eladb, line 203: What happens if we had an earlier `bring` statement for this file specifically?

```js
bring "./my-module/file.w" as a;
bring "./my-module" as b;
```

Would it parse `file.w` twice?

libs/wingc/src/parser.rs
```diff
@@ -163,41 +163,74 @@ pub fn parse_wing_project(
...
+			tree_sitter_parser.set_language(tree_sitter_wing::language()).unwrap();
+			let tree_sitter_tree = tree_sitter_parser.parse("", None).unwrap();
+			let ast = Scope::empty();
+			let dependent_wing_paths = files_and_dirs;
+
+			// Update our collections of trees and ASTs and our file graph
+			tree_sitter_trees.insert(file_or_dir_path.clone(), tree_sitter_tree);
+			asts.insert(file_or_dir_path.clone(), ast);
+			file_graph.update_file(&file_or_dir_path, &dependent_wing_paths);
```
eladb, line 212: DRY with lines 228-230

libs/wingc/src/type_check.rs
eladb: Since the symbol environment is a map, I don't think it is possible to find multiple symbols with the same name within an environment, no?

examples/tests/valid/subdir2/file1.w
```diff
@@ -0,0 +1,11 @@
+bring "./inner" as blah; // an alias that is not "inner"
+
+class Foo {
+  pub foo(): str {
+    return "foo";
+  }
+
+  pub checkWidget(widget: blah.Widget): num {
```
eladb, line 8: Perhaps add a static call to `bla.Widget.someStaticMethod()` because this example only references a type from `bla` without any runtime manifestation.

libs/wingc/src/type_check.rs
Chriscbr: Good question. It is a map, but the entries of the map aren't always individual types - they can also refer to namespaces (see the `SymbolKind` enum). In this PR I've updated namespaces so they now can point to multiple environments (to support looking up a symbol in a directory), hence why it's now possible to have multiple matches during a symbol lookup.

docs/docs/03-language-reference.md
Chriscbr: Yep you're right - will update

libs/wingc/src/jsify.rs
Chriscbr: Mmm -- I've gone ahead and added some validation to `parser.rs` for validating we're dealing with relative paths (it kind of feels like this kind of input validation should be handled long before jsification)

libs/wingc/src/lsp/completions.rs
Chriscbr: Done

libs/wingc/src/parser.rs
Chriscbr: OK - I tried refactoring it to a function: [`40a12c6` (#4210)](https://github.com/winglang/wing/pull/4210/commits/40a12c63d92d34c349157b716eb5ba4e489f8af9)

I'm not really sure this improves readability, but I might be misunderstanding your suggestion

libs/wingc/src/parser.rs
Chriscbr: It turns out this works out of the box! 🎉 In a nutshell, we're building a stack of paths that need to be parsed, yeah? When we're popping entries from the stack, there's a branch that checks if the path has already been parsed, in which case it's ignored.

examples/tests/valid/subdir2/file1.w
Chriscbr: Added

docs/docs/03-language-reference.md
```diff
@@ -1771,6 +1771,28 @@ bring cloud; // from cloud bring * as cloud;
...
+
+It's also possible to import a directory as a module. The module will contain all
+public types defined in the directory's files. If the directory has subdirectories,
+they will be available under the corresponding names.
+
+```TS
+bring "./my-module" as myModule;
+
+// from ./my-module/submodule/my-class.w
+new myModule.submodule.MyClass();
```
MarkMcCulloh, line 1789: I wonder if this will create a schism in how people do imports in wing. Should we only allow one way? I suspect people would gravitate towards directory imports since they're more powerful.

MarkMcCulloh (approved):

libs/wingc/src/parser.rs
eladb: I would have declared this as a local function because there's a lot of local state involved.

examples/tests/invalid/subdir/other.w
```diff
@@ -0,0 +1,2 @@
+bring "./inner" as inner;
+// Error: "Foo" is defined multiple times in "./inner"
```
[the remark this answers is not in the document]
eladb: This is something that came up during the design (@Chriscbr was concerned about the same thing). I propose that we try this approach for now and see how it works in the wild.

One of the principles we tries to maintain is that any .w file can be used as the root of a compilation graph. It explicitly references all files it depends on.

If we implicitly allow types to be referenced within a subdirectory, it means that the compiler would need to always compile all the files within the subdirectory (recursively).

As far as I understand this simplifies the LSP.

docs/docs/03-language-reference.md
eladb: I think it would be pretty annoying if I would have to create a directory for every module.

If we eventually decide to support implicit references across files in the same directory, then this would make more sense.

eladb (approved):

libs/wingc/src/type_check.rs
yoav-steinberg: My initial concern is that this has the potential to break the name uniqueness we're trying to have at a namespace level. In theory now a namespace might have two symbols with the same name each in a different env.
I'll dig into the implementation to see if I'm missing something here.

libs/wingc/src/parser.rs
```diff
@@ -735,54 +801,95 @@ impl<'s> Parser<'s> {
...
-			if !source_path.is_file() {
-				return self.with_error(
-					format!("Cannot bring module \"{}\": not a file", source_path),
-					statement_node,
-				);
-			}
-			if source_path.ends_with(".main.w\"") || source_path.ends_with(".test.w\"") {
+
+			// case: .w file
+			if source_path.ends_with("/main.w") || source_path.ends_with(".main.w\"") || source_path.ends_with(".test.w\"") {
```
yoav-steinberg, line 822: maybe:
```suggestion
			if source_path.ends_with("/main.w\"") || source_path.ends_with(".main.w\"") || source_path.ends_with(".test.w\"") {
```
Also why fail and not just silently ignore them. This error won't let you bring something that also has an entry point. Finally what about "test.w" files? Should we ignore them too?

libs/wingc/src/parser.rs
```diff
@@ -735,54 +801,95 @@ impl<'s> Parser<'s> {
...
-					format!(
-						"bring {} must be assigned to an identifier (e.g. bring \"foo\" as foo)",
-						module_name
-					),
-					statement_node,
-				)
-			};
+			if source_path.is_file() {
+				if source_path.extension() != Some("w") {
+					return self.with_error(
```
yoav-steinberg, line 834: silently ignore? what if there's a README.md file etc..

libs/wingc/src/type_check.rs
yoav-steinberg: Can't we design it in such a way that if a dir has multiple files in it and none of them are explicitly brought then the dir's namespace will have a single env aggregating all the types in all the files in that level of the dir?
Bringing one of the files explicitly will by definition create another namespace where we'll just add those types into again.

docs/docs/03-language-reference.md
Chriscbr: Directory imports are more powerful, but sometimes you may want to use types from sibling files, in which case importing the individual file is necessary because https://github.com/winglang/wing/issues/4276 is not supported yet

libs/wingc/src/type_check.rs
Chriscbr: I didn't have a chance to try that approach, though it sounds like it could also work. I think my only concern is that it would require copying all of the symbols from the other symbol environments into a new symbol environment (unless there is another way...?) -- and since that would mean there's no longer one source of truth for a given symbol (as it might be defined in multiple places), it could lead to bugs in the future. For example, maybe there could be issues with only a subset of files in the project are re-type-checked by the LSP for example (incremental compilation). But I'm not sure.

libs/wingc/src/parser.rs
Chriscbr: Oh I see. Looks like it's a little tricky to do here because of borrow checking details but I'll keep the suggestion in mind!

docs/docs/03-language-reference.md
```diff
@@ -1771,6 +1771,28 @@ bring cloud; // from cloud bring * as cloud;
...
+```
+
+The following features are not yet implemented, but we are planning to add them in the future:
+
+* Specify types as public using `pub` - see https://github.com/winglang/wing/issues/4294 to track.
+
```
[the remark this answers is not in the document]
eladb: yes, that's what I was thinking too.

libs/wingc/src/parser.rs
```diff
@@ -153,51 +153,102 @@ pub fn parse_wing_project(
...
-		files.add_file(&file_path, file_text.clone()).unwrap();
-		files.get_file(&file_path).unwrap();
+		// Handle directories specially
+		if file_or_dir_path.is_dir() {
+			// Collect a list of all files and subdirectories in the directory
+			let mut files_and_dirs = Vec::new();
+			for entry in fs::read_dir(&file_or_dir_path).expect("read_dir call failed") {
+				let entry = entry.unwrap();
+				let path = Utf8PathBuf::from_path_buf(entry.path()).expect("invalid utf8 path");
+				if path.is_dir() || path.extension() == Some("w") {
```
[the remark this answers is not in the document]
eladb: I think yes

examples/tests/valid/bring_local_dir.test.w
```diff
@@ -0,0 +1,19 @@
...
+
+// from subdir/file2.w
+let bar = new subdir.Bar();
+assert(bar.bar() == "bar");
+
+// from subdir/inner/widget.w
+let widget2 = new subdir.inner.Widget();
+assert(widget2.compute() == 42);
+
+assert(foo.checkWidget(widget2) == 1379);
```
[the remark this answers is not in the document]
Chriscbr: @yoav-steinberg In this example I believe they are the same type (they're referring to the same class from the same source file -- they're just being accessed in different ways). Did you mean some kind of test that two different classes named `Widget` (defined in different files) aren't accidentally treated as the same?

examples/tests/valid/struct_from_json.test.w
```diff
@@ -262,7 +262,7 @@ bring "./subdir/structs_2.w" as otherExternalStructs;
 
 struct MyStruct {
   m1: externalStructs.MyStruct;
-  m2: otherExternalStructs.MyStruct;
+  m2: otherExternalStructs.MyStruct2;
```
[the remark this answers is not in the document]
Chriscbr: OK, yeah you're right - I looked into it and it's safe revert this change.

The reason changed it was because I noticed the code was organized so `subdir/structs.w` had a struct named `MyStruct` AND `subdir/structs_2.w` has a struct named `MyStruct`. Since the structs have the same name, it's not possible to `bring "./subdir"` safely (there would be a name collision). But `struct_from_json.test.w` doesn't actually import the entire directory, so that error doesn't come up.

libs/wingc/src/ast.rs
```diff
@@ -428,7 +426,10 @@ pub struct Interface {
 pub enum BringSource {
 	BuiltinModule(Symbol),
 	JsiiModule(Symbol),
+	/// Refers to a relative path to a file
 	WingFile(Symbol),
+	/// Refers to a relative path to a file
```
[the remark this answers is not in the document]
Chriscbr: Thanks

libs/wingc/src/parser.rs
Chriscbr: Fixed (turns out we already have a helper function named `is_entrypoint_file` for this)

> Also why fail and not just silently ignore them. This error won't let you bring something that also has an entry point.

I'm not sure I'm understanding your suggestion. Basically this rule / error is enforcing that there are two types of Wing files, entrypoint files and module files, and you can only "bring" other module files. What's the behavior you're proposing?

I've added another case to `bring_local_self.test.w` that shows the error message since it looks like I missed this one

> Finally what about "test.w" files? 

I don't have a strong opinion on it, but I'd rather leave for another PR if it's ok

libs/wingc/src/parser.rs
Chriscbr: What would `bring "./README.md"` do?

Chriscbr: oh maybe you're asking about how we handle directories generally -- we just skip trying to parse any file that doesn't have an extension of .w

https://github.com/winglang/wing/blob/bd3003689b771cc7d81646a3839573c6db43bb83/libs/wingc/src/parser.rs#L193-L203

docs/docs/03-language-reference.md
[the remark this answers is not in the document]
Chriscbr: I went ahead and added a link to the issue about `internal` 👍

eladb:
Horray!

PR #4257, opened 2023-09-21
feat(cli): introduce default entrypoints

garysassano:
Wing CLI can now run the following commands without passing any entrypoint:
- `wing test` - looks for`*.test.w` files in current directory and tests them
- `wing compile` - looks for `main.w` file in current directory and compiles it
- `wing it` - looks for `main.w` file in current directory and runs it

### DIFF
--- a/apps/wing/package.json
+++ b/apps/wing/package.json
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@segment/analytics-node": "^1.1.0",
+    "@types/glob": "^8.1.0",
     "@wingconsole/app": "workspace:^",
     "@wingconsole/server": "workspace:^",
     "@winglang/compiler": "workspace:^",
@@ -40,6 +41,7 @@
     "commander": "^10.0.1",
     "compare-versions": "^5.0.3",
     "debug": "^4.3.4",
+    "glob": "^8.1.0",
     "nanoid": "^3.3.6",
     "open": "^8.4.2",
     "ora": "^5.4.1",
--- a/apps/wing/src/cli.ts
+++ b/apps/wing/src/cli.ts
@@ -6,7 +6,7 @@ import { collectCommandAnalytics } from "./analytics/collect";
 import { exportAnalytics } from "./analytics/export";
 import { optionallyDisplayDisclaimer } from "./analytics/disclaimer";
 import { currentPackage } from "./util";
-
+import { glob } from "glob";
 export const PACKAGE_VERSION = currentPackage.version;
 let analyticsExportFile: Promise<string | undefined>;
 
@@ -21,6 +21,13 @@ if (!SUPPORTED_NODE_VERSION) {
 
 function runSubCommand(subCommand: string) {
   return async (...args: any[]) => {
+    if (args[0] === "DEFAULT_TEST_ENTRYPOINTS") {
+      args[0] = glob.sync("*.test.w");
+      if (args[0].length === 0) {
+        throw new Error("No '.test.w' file found in current directory.");
+      }
+    }
+
     try {
       const exitCode = await import(`./commands/${subCommand}`).then((m) => m[subCommand](...args));
       if (exitCode === 1) {
@@ -119,7 +126,7 @@ async function main() {
     .command("run")
     .alias("it")
     .description("Runs a Wing program in the Wing Console")
-    .argument("[entrypoint]", "program .w entrypoint")
+    .argument("[entrypoint]", "program .w entrypoint", "main.w")
     .option("-p, --port <port>", "specify port")
     .option("--no-open", "Do not open the Wing Console in the browser")
     .hook("preAction", collectAnalyticsHook)
@@ -134,7 +141,7 @@ async function main() {
   program
     .command("compile")
     .description("Compiles a Wing program")
-    .argument("<entrypoint>", "program .w entrypoint")
+    .argument("[entrypoint]", "program .w entrypoint", "main.w")
     .addOption(
       new Option("-t, --target <target>", "Target platform")
         .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
@@ -151,7 +158,7 @@ async function main() {
     .description(
       "Compiles a Wing program and runs all functions with the word 'test' or start with 'test:' in their resource identifiers"
     )
-    .argument("<entrypoint...>", "all entrypoints to test")
+    .argument("[entrypoint...]", "all entrypoints to test", "DEFAULT_TEST_ENTRYPOINTS")
     .addOption(
       new Option("-t, --target <target>", "Target platform")
         .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
--- a/apps/wing/src/commands/run.test.ts
+++ b/apps/wing/src/commands/run.test.ts
@@ -26,11 +26,11 @@ test("wing it runs the only .w file", async () => {
   try {
     process.chdir(workdir);
 
-    writeFileSync("foo.w", "bring cloud;");
+    writeFileSync("foo.main.w", "bring cloud;");
 
-    await run();
+    await run("foo.main.w");
     expect(createConsoleApp).toBeCalledWith({
-      wingfile: resolve("foo.w"),
+      wingfile: resolve("foo.main.w"),
       requestedPort: 3000,
       hostUtils: expect.anything(),
       requireAcceptTerms: false,
@@ -47,8 +47,8 @@ test("wing it with no file throws error for a directory with more than 1 .w file
   try {
     process.chdir(workdir);
 
-    writeFileSync("foo.w", "bring cloud;");
-    writeFileSync("bar.w", "bring cloud;");
+    writeFileSync("foo.main.w", "bring cloud;");
+    writeFileSync("bar.main.w", "bring cloud;");
 
     await expect(run).rejects.toThrow("Please specify which file you want to run");
   } finally {
@@ -62,11 +62,11 @@ test("wing it with a file runs", async () => {
   try {
     process.chdir(workdir);
 
-    writeFileSync("foo.w", "bring cloud;");
+    writeFileSync("foo.main.w", "bring cloud;");
 
-    await run("foo.w");
+    await run("foo.main.w");
     expect(createConsoleApp).toBeCalledWith({
-      wingfile: resolve("foo.w"),
+      wingfile: resolve("foo.main.w"),
       requestedPort: 3000,
       hostUtils: expect.anything(),
       requireAcceptTerms: false,
@@ -80,7 +80,7 @@ test("wing it with a file runs", async () => {
 test("wing it with a nested file runs", async () => {
   const workdir = await mkdtemp(join(tmpdir(), "-wing-it-test"));
   const subdir = join(workdir, "subdir");
-  const filePath = join(subdir, "foo.w");
+  const filePath = join(subdir, "foo.main.w");
   const prevdir = process.cwd();
   try {
     process.chdir(workdir);
@@ -107,9 +107,9 @@ test("wing it with an invalid file throws exception", async () => {
   try {
     process.chdir(workdir);
 
-    writeFileSync("foo.w", "bring cloud;");
+    writeFileSync("foo.main.w", "bring cloud;");
 
-    await expect(run("bar.w")).rejects.toThrow("bar.w doesn't exist");
+    await expect(run("bar.main.w")).rejects.toThrow("bar.main.w doesn't exist");
   } finally {
     process.chdir(prevdir);
   }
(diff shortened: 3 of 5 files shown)


apps/wing/src/cli.ts
```diff
@@ -151,7 +158,7 @@ async function main() {
     .description(
       "Compiles a Wing program and runs all functions with the word 'test' or start with 'test:' in their resource identifiers"
     )
-    .argument("<entrypoint...>", "all entrypoints to test")
+    .argument("[entrypoint...]", "all entrypoints to test", "DEFAULT_TEST_ENTRYPOINTS")
```
eladb, line 161: What's this?

eladb (changes requested):
Missing tests and docs for the new behavior 

@garysassano please review our contribution guidelines. Every new features requires a test and updates to the relevant documentation.

apps/wing/src/cli.ts
garysassano: If you call `wing test` without passing any entrypoint, it would use `DEFAULT_TEST_ENTRYPOINTS` as the default entrypoint, which then gets resolved here: https://github.com/winglang/wing/pull/4257/files#diff-c5ea03c3011f3998e7d6a7bb7ff4a3a29f2035a6006edc2e64c9651afae78b8eR24-R29

This solution was needed in order to throw an error and point the user towards the right direction, otherwise he could be confused seeing that the command works and executes correctly despite testing 0 files.

A different approach could be to put all the logic inside a function if you don't like the approach of using a placeholder string there.

garysassano: Btw, I'm aware of the contribution guidelines. I opened this PR without updating the existing tests and docs just to see if you agreed on having this as a  feature.

Chriscbr:
> - `wing test` - looks for`*.test.w` files in current directory and tests them
> - `wing compile` - looks for `main.w` file in current directory and compiles it
> - `wing it` - looks for `main.w` file in current directory and runs it

Not sure if I'm in love with this design - when I run `wing test` without any arguments, I'd want all unit tests throughout my project to get executed. It shouldn't matter if its name is `main.w` or `storage.test.w` or `storage.w` (you can't define tests in non-entrypoint files right now, but that's a bug at the moment).

For `wing it` or `wing compile` without any arguments, if I just have a single file in my directory named `hello.main.w` or `hello.test.w`, I think it makes sense for it to be picked up automatically, no? All of these are valid names of entrypoint files.

apps/wing/src/commands/test.ts
```diff
@@ -29,6 +30,25 @@ export interface TestOptions extends CompileOptions {
 }
 
 export async function test(entrypoints: string[], options: TestOptions): Promise<number> {
+  // Initialize a Set to store unique file names
+  let expandedEntrypointsSet: Set<string> = new Set();
+
+  // Expand each entrypoint and add file names to the Set
+  for (const entrypoint of entrypoints) {
+    const expandedEntrypoint = globSync(entrypoint);
```
MarkMcCulloh, line 38: Use the async glob instead.
Also I think glob can take an array, so there may be no need to deal with the Set above (it probably deals with duplicates)

garysassano:
> Not sure if I'm in love with this design - when I run `wing test` without any arguments, I'd want all unit tests throughout my project to get executed. It shouldn't matter if my tests are in `main.w` or `storage.test.w` or `storage.w` (you can't define tests in non-entrypoint files right now, but that's a bug at the moment).
> 
> For `wing it` or `wing compile` without any arguments, if I just have a single file in my directory named `hello.main.w` or `hello.test.w`, I think it makes sense for it to be picked up automatically, no? All of these are valid names of entrypoint files.

@Chriscbr So you say to close this PR?

Chriscbr:
@garysassano I just took a look at the code - most of the updates and doc changes look good, I think we can probably tweak some of the globs and add unit tests and it should be ok 👍

apps/wing/src/cli.ts
```diff
@@ -151,7 +150,7 @@ async function main() {
     .description(
       "Compiles a Wing program and runs all functions with the word 'test' or start with 'test:' in their resource identifiers"
     )
-    .argument("<entrypoint...>", "all entrypoints to test")
+    .argument("[entrypoint...]", "all files to test, globs are supported", ["{*main,*.test}.w"])
```
MarkMcCulloh, line 153: We don't want to include things like sdfgsdfgmain.w

```suggestion
    .argument("[entrypoint...]", "all files to test, globs are supported", ["{main,*.main,*.test}.w"])
```

docs/docs/06-tools/01-cli.md
```diff
@@ -139,10 +152,10 @@ Usage:
 
 ```sh
 $ export CDK_STACK_NAME="my-project"
-$ wing compile --target awscdk app.w
+$ wing compile [entrypoint] --target awscdk
```
MarkMcCulloh, line 155: Why reorder?

docs/docs/06-tools/01-cli.md
```diff
@@ -167,12 +180,18 @@ The `wing test` command can be used to compile and execute tests in Wing applica
...
 ```
 
-This will compile each entrypoint, and for each entrypoint it will invoke all tests in the program.
+`[entrypoint...]` specifies the entrypoint list of files that will be compiled and tested. Valid entrypoint names are `main.w`, `<file>.main.w` and `<file>.test.w`.
+
+For example ([test_bucket.test.w](https://github.com/winglang/wing/tree/main/examples/tests/valid/test_bucket.test.w)):
+
+:::default entrypoint...
+
+It's possible to execute `wing test` without specifying any entrypoint, in which case the CLI looks for all files ending with `.test.w` in the current directory.
```
MarkMcCulloh, line 192:
```suggestion
Executing `wing test` without specifying any entrypoint looks for all entrypoint files in the current directory.
```

docs/docs/06-tools/01-cli.md
```diff
@@ -167,12 +180,18 @@ The `wing test` command can be used to compile and execute tests in Wing applica
...
-$ wing test ENTRYPOINT...
+$ wing test [entrypoint...]
 ```
 
-This will compile each entrypoint, and for each entrypoint it will invoke all tests in the program.
+`[entrypoint...]` specifies the entrypoint list of files that will be compiled and tested. Valid entrypoint names are `main.w`, `<file>.main.w` and `<file>.test.w`.
+
+For example ([test_bucket.test.w](https://github.com/winglang/wing/tree/main/examples/tests/valid/test_bucket.test.w)):
+
+:::default entrypoint...
```
MarkMcCulloh, line 190:
```suggestion
:::note Default Entrypoint(s)
```

apps/wing/src/commands/run.test.ts
```diff
@@ -20,56 +20,20 @@ vi.mock("@wingconsole/app", () => {
...
 
-    await run("foo.w");
+    await run("foo.main.w");
     expect(createConsoleApp).toBeCalledWith({
-      wingfile: resolve("foo.w"),
+      wingfile: resolve("foo.main.w"),
       requestedPort: 3000,
       hostUtils: expect.anything(),
-      requireAcceptTerms: false,
+      requireAcceptTerms: true,
```
MarkMcCulloh, line 36: undo these

docs/docs/06-tools/01-cli.md
```diff
@@ -194,9 +213,9 @@ test "get" {
 Now, if we run the following command:
 
 ```sh
-$ wing test test_bucket.w
-pass | test_bucket.w | root/test:get
-pass | test_bucket.w | root/test:put
+$ wing test test_bucket.test.w
+pass | test_bucket.testw | root/test:get
```
MarkMcCulloh, line 217:
```suggestion
pass | test_bucket.test.w | root/test:get
```

apps/wing/src/cli.ts
garysassano: According to @Chriscbr:
>It shouldn't matter if my tests are in `main.w` or `storage.test.w` or `storage.w` (you can't define tests in non-entrypoint files right now, but that's a bug at the moment).

Therefore I used `*.w` so that it accounts already for when the bug will be fixed.

docs/docs/06-tools/01-cli.md
garysassano: It looked better to me to follow the same order everywhere with the `[entrypoint]` soon after the command (`wing run|it` doesn't have an option for example). If you think that it was better the other way around, I can revert this change.

apps/wing/src/commands/run.test.ts
garysassano: By undoing this change, I'm unable to move on since the test returns `true`.

apps/wing/package.json
```diff
@@ -26,7 +26,7 @@
     "copy-root-assets": "cp ../../logo/demo.gif ./logo/demo.gif",
     "watch": "tsc --watch",
     "test": "vitest run",
-    "dev": "./bin/wing",
+    "dev": "pnpm build && ./bin/wing",
```
MarkMcCulloh, line 29:
```suggestion
    "dev": "./bin/wing",
```

docs/docs/06-tools/01-cli.md
MarkMcCulloh: I don't feel strongly about it

apps/wing/src/commands/test.ts
```diff
@@ -29,6 +30,16 @@ export interface TestOptions extends CompileOptions {
...
 export async function test(entrypoints: string[], options: TestOptions): Promise<number> {
+  const expandedEntrypoints = await glob(entrypoints);
+
+  if (expandedEntrypoints.length === 0) {
+    // Check if it's the default entrypoint
+    if (entrypoints[0] === "*.test.w") {
+      throw new Error("No '.test.w' files found in current directory.");
+    }
+    throw new Error("No matching files found in current directory.");
+  }
```
MarkMcCulloh, line 41: I think the error can be simplified just by spitting the requested globs back to the user

```suggestion
  if (expandedEntrypoints.length === 0) {
    throw new Error(`No matching files found in current directory. (${entrypoints.join(", ")})`);
  }
```

docs/docs/06-tools/01-cli.md
```diff
@@ -17,40 +17,46 @@ The CLI is distributed as an npm package which can be installed globally using:
...
+`[entrypoint]` is a valid entrypoint for a Wing program. An entrypoint can be either source code or compiled. If it's source code, the entrypoint refers to a file named `main.w` or ending with `.main.w`/`.test.w`, whereas if it's compiled the entrypoint refers to a directory whose name ends with `.wsim`.
 
-- `ENTRYPOINT.w` is an entrypoint for a Wing program (source code). In this case, Wing Console will
-  *watch for changes* and will automatically recompile your program when the source code change.
-  The entrypoint is optional if there's a single `.w` file in the working directory (in which case you
-  can just type `wing it` and it will run this file).
-- `PROGRAM.wsim` is the output of `wing compile -t sim`
+:::note Default Entrypoint
+
+It's possible to execute `wing run|it` without specifying any entrypoint, in which case the CLI looks for exactly one file named `main.w` or ending with `.main.w` in the current directory. If no files or more than one file are found, the CLI throws an error.
```
MarkMcCulloh, line 41: This can be shortened, no need to mention the error IMO

```suggestion
By default, `wing run|it` will look for exactly one file named `main.w` or ending with `.main.w`
```

docs/docs/06-tools/01-cli.md
```diff
@@ -17,40 +17,46 @@ The CLI is distributed as an npm package which can be installed globally using:
...
 ```
-$ wing compile --target <TARGET> <ENTRYPOINT.w>
-```
 
-The `ENTRYPOINT.w` specifies the entrypoint file to compile (e.g. `hello.w`).
+`[entrypoint]` specifies the entrypoint file to compile. A file is considered a valid entrypoint if its name is `main.w` or if it ends with `.main.w`/`.test.w`.
+
+:::note Default Entrypoint
+
+It's possible to execute `wing compile` without specifying any entrypoint, in which case the CLI looks for exactly one file named `main.w` or ending with `.main.w` in the current directory. If no files or more than one file are found, the CLI throws an error.
```
MarkMcCulloh, line 57: Same as above

```suggestion
By default, `wing compile` will look for exactly one file named `main.w` or ending with `.main.w`
```

docs/docs/06-tools/01-cli.md
```diff
@@ -17,40 +17,46 @@ The CLI is distributed as an npm package which can be installed globally using:
...
+
+:::
 
 The `--target` option is **required**, and specifies the target platform to compile for. The
 following targets are supported:
```
MarkMcCulloh, line 62:
```suggestion
The --target option specifies the target platform to compile for. The default target is `sim`.
The following targets are supported:
```

docs/docs/06-tools/01-cli.md
```diff
@@ -59,19 +65,26 @@ following targets are supported:
 * `tf-aws` - Terraform/AWS
 * `tf-azure` - Terraform/Azure
 
+:::note Default Target
+
+It's possible to execute `wing compile` without specifying any target, in which case the CLI defaults to `sim` target.
+
+:::
+
```
MarkMcCulloh, line 73: No need for this if above suggestion is accepted

```suggestion
```

MarkMcCulloh (approved):
Most of the suggestions here are for the docs, however I think they need a lot of work regardless of this PR. Thanks for keeping on this one, Gary!

apps/wing/package.json
```diff
@@ -21,7 +21,7 @@
   },
   "scripts": {
     "build": "tsc && pnpm copy-root-readme",
-    "eslint": "eslint --ext .ts src",
+    "eslint": "eslint --fix --ext .ts src",
```
MarkMcCulloh, line 24: I snuck this change into your PR as part of settling the merge conflict with a recent PR adding eslint to this project

PR #4281, opened 2023-09-26
feat(sdk): added tests to std.Range

revitalbarletz:
## Checklist

- [ ] Tests added (always)


examples/tests/sdk_tests/std/range.test.w
```diff
@@ -1,2 +1,52 @@
...
+assert(std.Range.of(1,5,false).join() == "1,2,3,4");
+
+test "rangeOfWithoutLastNumber" {
+   assert(std.Range.of(1,5,false).join() == "1,2,3,4");
+}
+
+assert(std.Range.of(1,5).join() == "1,2,3,4");
+
+test "rangeOfWithoutLastNumberDefault" {
+   assert(std.Range.of(1,5).join() == "1,2,3,4");
```
WeepingClown13, line 19: perhaps an `assert(std.Range.of(x, y).join() == std.Range.of(x, y, false).join());` too? this could confirm that the default boolean does not get changed accidentally.

WeepingClown13 (approved):

WeepingClown13:
@revitalbarletz I modified the pr title ever so slightly to satisfy the linter.

examples/tests/sdk_tests/std/range.test.w
```diff
@@ -1,2 +1,52 @@
...
+assert(i==5);
+
+test "forWithRange" { 
+   let var i = 0;
+   for x in 0..5 { i+=1; }
+   assert(i==5);
+}
+
+let var n = 0;
+for x in 0..=5 { n+=1; }
```
tsuf239, line 45: maybe another assertion for each iteration step?

examples/tests/sdk_tests/std/range.test.w
```diff
@@ -1,2 +1,52 @@
...
+
+test "rangeOfWithoutLastNumberDefault" {
+   assert(std.Range.of(1,5).join() == "1,2,3,4");
+}
+
+assert(std.Range.of(-3,3,true).join() == "-3,-2,-1,0,1,2,3");
+
+test "rangeOfWithNegativeRange" {
+   assert(std.Range.of(-3,3,true).join() == "-3,-2,-1,0,1,2,3");
+}
```
tsuf239, line 26: Maybe one more with a negative range and false at the last argument? (as well as decreasing one with false at the last argument)

examples/tests/sdk_tests/std/range.test.w
revitalbarletz: Why do you think it is required? for loop should be increasing by 1, I am not testing the for loop here but the range.

tsuf239 (approved):

tsuf239:
> You could add a range with the same number: `std.Range.of(x, x)`

good idea actually

PR #4297, opened 2023-09-26
fix(compiler): compile to cjs for open issue 3608

bronifty:
updates to wing compiler package json scripts to format output as cjs

 "compile": "tsup --format cjs",
 "watch": "tsup --watch --format cjs",

references issue https://github.com/winglang/wing/pull/3608 to address  https://github.com/winglang/wing/issues/3601

currently it is not fixed, but once we resolve it I will be happy to add a test case.

## Checklist

- [ ] Tests added (always)


staycoolcall911:
@bronifty - you had a whitespace before "fix(compiler)" which caused the PR title linter to fail. I removed it

MarkMcCulloh:
@bronifty Can you add some details to the PR description about this change? The title references "issue 3608", is that referring to https://github.com/winglang/wing/pull/3608?
Does that mean this addresses https://github.com/winglang/wing/issues/3601? If so, adding a test case for this would be helpful.

bronifty:
reopening as per Uri message

bronifty:
reopening as per Uri message

bronifty:
hi @MarkMcCulloh yes it's meant to address https://github.com/winglang/wing/issues/3601 . 
the resolution i applied here doesn't actually fix the issue, which is why i closed it. but i'm reopening this as per a message from @staycoolcall911  so that hopefully we can work together to find a resolution. and yes i'd be happy to build a test case for it.

MarkMcCulloh:
@bronifty Thanks for taking this on again. The root issue here is about what files the wing compiler output. The PR currently only adds `--format cjs` for `@winglang/compiler`, but that only changes the output for that library. In fact, I think it is already outputting cjs so this change shouldn't be needed.

The change needed would be to find any places we are outputting .js files and change it to .cjs. This will include changes in both the rust and typescript code. There will likely be several places to change. Good files to start in:
- https://github.com/winglang/wing/blob/main/libs/wingc/src/jsify.rs
  - This is where the compiler outputs js files
- https://github.com/winglang/wing/blob/main/libs/wingsdk/src/shared/bundling.ts
  - This is where we run esbuild, which takes in and outputs js files for inflights

bronifty:
@MarkMcCulloh i spoke with @yoav-steinberg and @hasanaburayyan and my understanding of the architecture is as below. I reverted the original change as per your suggestion and made some updates to jsify and compile.ts, the companion compilation files as well as bundling.ts where the emitted files are ingested. Anecdotally, these changes solved my issue and i can run wing files in the root of an ESM project like React with Vite without any issues after these changes. Also, I show one less error in my build (pnpm build including hangar tests) than the build for the commit directly preceding it. Also the current upstream / production repo's main branch has more errors in its build than this just as a sanity check to put this in perspective. I'll add all three results here as an fyi as well. 
wing-compiler3

bronifty:
prod-build
prev-build
my-build

bronifty:
The prod build looks like this:
prod-build

I merged this upstream main branch into a test branch with my latest changes and the solution is working. here is a screenshot of running wing in the root of an ESM project:
test-branch-merged-upstream-working

bronifty:
Here are the results of the latest build on prod main. Is this what it looks like for you as well?
prod-build

I could use some pointers on handling the merge of my forked repo's main branch with the upstream main as well because it looks like I am able to use a struct without modifying it as public, which would be a regression since the logic was updated at some point to reflect these entities as having private members by default. here is the result of my test of running the cli against my dev build which incorporated upstream main code on a test branch  
test-branch-merged-upstream-working

hasanaburayyan:
@bronifty Thanks for taking this on, Im starting to do a review on this. It looks like the branch needs rebased with main due to some merge conflicts. 

Also wanted to ask @Chriscbr should this change also included adding `"**/*.cjs"` the included globs for `wing pack`?

Chriscbr:
> Also wanted to ask @Chriscbr should this change also included adding `"**/*.cjs"` the included globs for `wing pack`?

Yep I think all JS files (.js, .cjs, .mjs) should be packed into Wing libraries 👍

bronifty:
all globbed up ✅ @hasanaburayyan @Chriscbr

bronifty:
pls reopen have made updates to get it merged

PR #4300, opened 2023-09-27
fix(compiler): incorrect variadic type/arity check

### LINKED ISSUE #3872: Passing array to variadic argument does not raise type checking error
### I tried this:

```js
bring cloud;

let b1 = new cloud.Bucket() as "b1";
let b2 = new cloud.Bucket() as "b2";
let b3 = new cloud.Bucket() as "b3";
let b4 = new cloud.Bucket() as "b4";
let b5 = new cloud.Bucket() as "b5";

queue.node.addDependency([b1, b2, b3, b4]);
```

### This happened:

No type checking error

### I expected this:

An error like `Expected type to be "num" but got "Array<num>"`

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

0.25.14

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

Lancear:
- Merged the 2 arity checks to always provide the full diagnostics
- Positional arity check seemed to have the correct min args, the other had variadic for example required
- The arg count for functions with variadic params was not based on the actual args -> always passes the check
- There were 2 loops for checking the variadic args, but
  - 1 passed the same type for actual and expected
  - 1 checked the types with the first variadic arg as the expected type instead of the parameter type
- Added support for optional in collection_item_type as optional variadics seem to be allowed (cloud.Queue.push)
  - Since variadics can be empty as well, I don't really understand why we have optional variadics though
- Switched the variadic args type check to validate_type to also allow sub types like everywhere else
- Added more detailed error messages for arity check
- Renamed arg variables that were from the function signature to param

Fixes #3872

## Checklist

- [x] Tests added (always)


.gitignore
```diff
@@ -28,3 +28,5 @@ target/
 
 # turbo
 .turbo/
+
+esbuild-kit
```
MarkMcCulloh, line 32: Is this needed?

libs/wingc/src/type_check.rs
```diff
@@ -2512,32 +2503,23 @@ impl<'a> TypeChecker<'a> {
...
+		let arg_count = (non_variadic_len)
+			+ (if variadic_index.is_some() && arg_list.pos_args.len() > non_variadic_len {
+				1
+			} else {
+				0
+			}) + (if arg_list.named_args.is_empty() { 0 } else { 1 });
+
 		if arg_count < min_args || arg_count > max_args {
 			let err_text = if min_args == max_args {
 				format!("Expected {} arguments but got {}", min_args, arg_count)
```
MarkMcCulloh, line 2525: I know you didn't change this part, but the case of `Expected 1 arguments` is very noticeable now :)

```suggestion
				format!("Expected {} argument(s) but got {}", min_args, arg_count)
```

.gitignore
Lancear: It keeps getting auto-generated when I build. Asked in dev if someone knows more about it. I'm on windows using wsl.

libs/wingc/src/type_check.rs
Lancear: true, nice catch!

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -365,11 +365,11 @@ Duration <DURATION>"
 `;
 
 exports[`bring_non_std_construct.test.w 1`] = `
-"error: Expected 1 positional argument(s) but got 0
+"error: Expected between 1 and 2 arguments but got 0
   --> ../../../examples/tests/invalid/bring_non_std_construct.test.w:8:1
   |
 8 | new cdktf.S3Backend();
-  | ^^^^^^^^^^^^^^^^^^^^^ Expected 1 positional argument(s) but got 0
+  | ^^^^^^^^^^^^^^^^^^^^^ Expected between 1 and 2 arguments but got 0
```
yoav-steinberg, line 372: The new error message here seems a bit misleading to me. We actually expect either 2 pos argument or 1 pos argument N named argument fitting `S3BackendConfig`'s fields.
Granted, the old message was misleading too.

Lancear: What would be a better error message in such a case? Maybe we can use a different message depending on whether the last param is a struct?

.gitignore
Lancear: not needed, just some issue in my setup. fixed. removing it now

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: How does this work in more complex cases? Suppose there are multiple named arguments:

```js
struct ComputeOptions {
  opt1: str;
  opt2: str;
  opt3: str?;
  opt4: str?;
}

let compute = (arg1: num, arg2: num, opts: ComputeOptions) => {
  // implementation
};
compute(123); // error (a)
compute(123, 456, opt2: "yes", opt1: "purple"); // ok!
```

What should the error be at (a)? The message we have today feels sort of incomplete:
`Expected 2 positional argument(s) but got 1`

One option we could consider is to provide the user a full list of the names of all missing arguments:
`Missing 1 positional argument ("arg2") and 2 named arguments ("opt1", "opt2")`
or
`Expected 2 positional arguments and 2 named arguments, but only received 1 positional argument`

Just brainstorming some ideas 😄 

----

BTW - in the future it should be possible to add "hints" to errors, so e.g. we could suggest what the correct syntax look like:

```
hint: I expected the function call to look like
    compute(arg1, arg2, opt1: exp1, opt2: exp2)
```

Lancear: Updated the error to give more details about variadic and named arguments. Also checking that not both a positional arg and named args are given for the last parameter. I like the idea of the hint but not sure if it belongs in this pr as the original goal was to fix variadic type checks and arity check and thats more related to named args and more of a nice addition than a fix.

examples/tests/invalid/function_call_variadic_min_arity.test.w
```diff
@@ -0,0 +1,7 @@
+bring cloud;
+
```
yoav-steinberg, line 2: no need to bring :cloud: 
```suggestion
```

examples/tests/valid/function_call_variadic_max_arity.test.w
```diff
@@ -0,0 +1,7 @@
+bring cloud;
+
```
yoav-steinberg, line 2: no need to bring :cloud: 
```suggestion
```

examples/tests/valid/function_call_variadic_max_arity.test.w
```diff
@@ -0,0 +1,7 @@
...
+let func = (n: num, b: bool, ...events: Array<Json>) => {
+  log("hello");
+};
+
+func(1, true, "a", "b", "c", "d"); // variadic args should count as 1 in arity check
```
yoav-steinberg, line 7: Since this is a valid test, lets actually verify it works
```suggestion
let func = (n: num, b: bool, ...events: Array<Json>) => {
  assert(events.at(-1) == "d");
};

func(1, true, "a", "b", "c", "d"); // variadic args should count as 1 in arity check
```

examples/tests/valid/function_call_variadic_sub_types.test.w
```diff
@@ -0,0 +1,9 @@
+bring cloud;
+
```
yoav-steinberg, line 2: no need to bring :cloud: 
```suggestion
```

examples/tests/valid/function_call_variadic_sub_types.test.w
```diff
@@ -0,0 +1,9 @@
...
+let func = (...events: Array<Json>) => {
+  log("hello");
+};
+
+let jsonStr = Json "str";
+
+func("str", jsonStr); // variadic args should allow sub types
```
yoav-steinberg, line 9: I'm not sure this is a case of "subtyping", isn't it our json implicit casting, @MarkMcCulloh ?
Just in case we can also test this with a actual class inheritance.

examples/tests/valid/function_call_variadic_sub_types.test.w
```diff
@@ -0,0 +1,9 @@
+bring cloud;
+
+let func = (...events: Array<Json>) => {
+  log("hello");
```
yoav-steinberg, line 4: lets add an assert here, cause why not?
```suggestion
  assert(events.at(0) == "str");
  assert(events.at(1) == "str");
```

libs/wingc/src/type_check.rs
```diff
@@ -2502,115 +2504,84 @@ impl<'a> TypeChecker<'a> {
...
+			if is_last_arg_struct {
+				let last_arg_type = last_arg.unwrap().typeref.maybe_unwrap_option();
+				self.validate_structural_type(&arg_list_types.named_args, &last_arg_type, call_span);
+			} else {
 				self.spanned_error(call_span, "No named arguments expected");
-				return Some(self.types.error());
-			}
-
-			self.validate_structural_type(&arg_list_types.named_args, &last_arg, call_span);
+			};
```
yoav-steinberg, line 2517: i think
```suggestion
			}
```

libs/wingc/src/type_check.rs
yoav-steinberg, line 1: A few questions
1. Where do we validate that the passed named args are of correct type to the corresponding struct fields? Do we have a test for this?
2. I couldn't find tests for the following error cases, which seem to be handled by the code:
  * "Expected {} positional argument(s) or named arguments for the last parameter but got {}"
  * "Expected {} positional argument(s) or named arguments for the last parameter + variadic args but got {}"
  * "Expected between {} and {} positional arguments but got {}"
  * "Expected between {} and {} positional arguments or named arguments for the last parameter + variadic args but got {}"
  * "Expected between {} and {} positional arguments + variadic args but got {}"

Lancear: 1. thats done within validate_structural_type which is called at the top of the function. Did not change that. I think I saw tests for it.
2. can add them. not sure if variadic + named args is possible though. just added it in case.

examples/tests/valid/function_call_variadic_sub_types.test.w
MarkMcCulloh: Yeah json and str (and other primitives) do not have a subtype relationship

libs/wingc/src/type_check.rs
yoav-steinberg: great, thanks, please try add what's missing (if there's something missing).

examples/tests/invalid/function_call_arity.test.w
```diff
@@ -0,0 +1,39 @@
...
+
+namedFunc(); // Expected {} positional arguments or named arguments for the last parameter but got {}
+
+let namedOptionalParamsFunc = (n: num, b: bool, s: str?, opts: Options) => {};
+
+namedOptionalParamsFunc(3); // Expected between {} and {} positional arguments or named arguments for the last parameter but got {}
+
+
+// variadic and named args cannot be combined atm
+let invalidCombinationFunc = (n: num, b: bool, ...variadic: Array<str>, opts: Options) => {};
```
yoav-steinberg, line 34: just to be clear this :point_up: give an error even without trying to call it?
`invalidCombinationFunc(1,2,"hi", prefix: "x", priority: 0);`

yoav-steinberg (approved):
looks good, thanks

examples/tests/invalid/function_call_arity.test.w
Lancear: yes as we have checks for variadics only allowed as the last parameter

Lancear:
Checking the following issue now, it seems to not recognize this function as being variadic 🤔 

```
FAIL  src/test_corpus/valid/function_variadic_arguments.test.w.test.ts > wing compile -t tf-aws
hangar:test: AssertionError:
hangar:test: error: Expected 0 positional argument(s) but got 1
hangar:test:    --> /home/lance/wing/examples/tests/valid/function_variadic_arguments.test.w:60:33
hangar:test:    |
hangar:test: 60 | subTypeFunc(new A("this is A"), new B("this is B")); // variadic args should allow sub types
hangar:test:    |                                 ^^^^^^^^^^^^^^^^^^ Expected 0 positional argument(s) but got 1
```

examples/tests/invalid/function_call_arity.test.w
Lancear: we dont seem to have that for the named args for structs, thats why i added a call for it

PR #4302, opened 2023-09-27
chore: improved benchmark reporting and gating

### LINKED ISSUE #428: Continuous Profiling
Added https://github.com/winglang/wing/pull/2139, some followups are needed:

- [ ] Vitest as a benchmarking platform adds too much overhead and has some sort of leak which increases time for each iteration of a benchmark. This makes it nearly unusable. We should move forward with a shell-based benchmarked like [hyperfine](https://github.com/sharkdp/hyperfine) instead.
- [ ] We need a way to compare benchmark runs and output the results

MarkMcCulloh:
Fixes #428

- In CI, the benchmarking will fail if any of the benches are slower than main by a percentage threshold
  - This threshold is currently set to a stupid number (50%). I'd like to ease it down to find the sweet spot over time
- Benchmarks should hopefully be more reliable
  - Running more iterations
  - Added a small sleep between each run, noticed that process handled leaked for a short time in-between cycles
  - (Note) None of the above effects the final runtime of CI since it still runs faster than hangar
- Added a comment to the PR with bench results and comparisons to main
  - This only happens for PRs from this repo (for now)
- Locally, compare to your last run
- Added a bench to get the version from the CLI
  - kinda works as a nice baseline

After this merges, I think it would be valuable to use the `wing -V` result as a baseline to determine if the threshold is being breached. Right now the variation of runs is still pretty bad. The last option left would be to get dedicated machines for benchmarks.


tools/hangar/src/benchmarking/table_report.ts
MarkMcCulloh, line 1: This file was moved from scripts/benchmark_json_to_table.mjs.

Other than switching it to typescript, it has barely changed.

tools/hangar/src/benchmarking/compare.ts
```diff
@@ -0,0 +1,135 @@
...
+      Math.round((newMean - oldMean) * 100) / 100;
+    differences[itemName].meanPercentDiff =
+      Math.round(((newMean - oldMean) / oldMean) * 10000) / 100;
+  }
+
+  const formatNumber = (num: number | typeof NaN | undefined, unit: string) => {
+    if (num === undefined || isNaN(num)) {
+      return "...";
+    }
+    return num.toFixed(2) + unit;
```
staycoolcall911, line 75: It applies to all the measurements we finally print to the table: I think we can keep it at ms, no reason for 1/100 of ms precision.

staycoolcall911 (commented):
We've been expecting this! Great!!

tools/hangar/src/benchmarking/compare.ts
MarkMcCulloh: Ok I changed ms to be rounded, but I kept 1 decimal of precision of the % (to prevent it from saying 0% change)

staycoolcall911 (approved):
LGTM

.github/workflows/build.yml
```diff
@@ -25,6 +25,7 @@ env:
   CARGO_TERM_COLOR: always
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
   TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+  SAME_REPO_PR: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
```
Chriscbr, line 28:
```suggestion
  IS_SAME_REPO_PR: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
```

Chriscbr (approved):
Nice work! A few non-blocking comments:

- Is showing all of the samples a bit verbose? (maybe we can just link to the build logs if someone wants to see the exact numbers)
- `totalTime` feels redundant (it's just `mean` * # of samples). Maybe we could show `variance` / `stdDev` instead
- Can we include plus/minus signs (+/-) in the `Change` column to signal whether the changes were increases or decreases?  Bonus points we can highlight them as green or red for easier-at-a-glance readability
- Super nit: move benchmark Wing files to `examples/tests/benchmarks` to make them easier to discover. (If benchmarking errors in CI, that should fail the build, so I'm not sure if they also need to be ran by the hangar)
- Can we add a section to the contributing guide about how to run the benchmarks locally (if it's possible)?

MarkMcCulloh:
> Super nit: move benchmark Wing files to examples/tests/benchmarks to make them easier to discover. (If benchmarking errors in CI, that should fail the build, so I'm not sure if they also need to be ran by the hangar)

Addressed all of the feedback except this. Having the benchmarks as a subset of `valid` is useful because it's shared node_modules with the rest of them. Also, I can see a usecase to have invalid benchmarks at some point.

Issue #4134, opened 2023-09-10
cloudwatch log groups should define retention period


skorfmann:
### I tried this:

Create and destroy any resource which will create log groups implicitly (e.g. a `cloud.Function`)

### This happened:

![Screenshot 2023-09-10 at 14 01 54](https://github.com/winglang/wing/assets/136789/a4e2c97f-6525-4612-822e-b36fdc61f403)

Log groups are created with `infinity` as retention period.

### I expected this:

Log groups to define a retention period of e.g. 7 days (any value will do).

### Component

SDK

### Wing Version

0.29.6

### Node.js Version

20.6.0

### Anything else?

tf cloudwatch logs can set a [skip destroy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group#skip_destroy), so that these will be kept around. Would probably be nice if there was an option if that's desired or not, but that's another issue.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
I think this issue may need some more discussion. I'm not sure if the change in #4303 to add `logRetentionDays` as an option in `cloud.Function` is the right design for the Wing Cloud Library, at least for now. We have a tenet that APIs for resources should be focused on functional behavior:

> Our APIs are designed around the functional aspects that developers care about for building and testing their applications. Implementations of resources in the SDK are guaranteed to be scalable, highly-available, and fault tolerant by default, so that developers do not need to customize security policies or scaling configuration within their application code. Operational aspects of resources should not leak into the core API surface area, except when they are essential to the functional behavior of the resource and the user's mental model. For example, while the timeout of a serverless function can be considered an operational detail, it is essential to the user's mental model of functions as an ephemeral, stateless resource that should not be used for long-running or stateful workloads. [[1]](https://www.winglang.io/contributing/rfcs/2023-01-20-wingsdk-spec#design-guidelines)

The duration a function's logs are stored for doesn't affect its functional behavior, nor the user's mental model of the resource, so it doesn't feel right to include it as part of our public API.

I think the better design here would be to set the default log retention to 30 days, and if it's needed, the user (or organization or platform team) can override the target-specific implementation of the resource through compiler plugins, [escape hatches](https://github.com/winglang/wing/issues/4161), or [platform providers](https://github.com/winglang/wing/issues/3620).

tsuf239:
[...]

PR #4303, opened 2023-09-27
fix(sdk): set log retention for cloud.Function (tf-aws and awscdk)

exoego:
Part of https://github.com/winglang/wing/issues/4134

Added `logRetentionDays` property to `cloud.Function`.
It is implemented for tf-aws and awscdk.

I skipped tf-gcp and tf-azure since the ticket does not mention and other cloud provider appears to have default retention:
- Cloud Logging: 30 days 
- Azure Monitor Application Insights: 90 days

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-awscdk/function.ts
```diff
@@ -36,6 +36,13 @@ export class Function extends cloud.Function implements IAwsFunction {
     // bundled code is guaranteed to be in a fresh directory
     const bundle = createBundle(this.entrypoint);
 
+    const logRetentionDays =
+      props.logRetentionDays === undefined
+        ? 30
+        : props.logRetentionDays < 0
+        ? undefined // Negative value means Infinite retention
```
tsuf239, line 43: Shouldn't we allow this as well? (what if a wing user does want infinite retention?)

tsuf239 (commented):
Is it possible add a wing test? (is there a way maybe an HTTP call to verify this property was set?)

tsuf239:
also, when I think of it now, maybe it will be good to add a "global" log retention value (let's say I want to set any of my resources to 1 week retention period), which probably should be in a different PR 
@skorfmann WDYT?

libs/wingsdk/src/target-awscdk/function.ts
exoego: Yes, it should allow infinite retention.
If a user specifies a negative value (e.g. `-1`), `logRetentionDays` will be `undefined`, which means infinite retention in CloudWatch.

tsuf239: oh I see now, nice!

exoego:
@tsuf239 Rebased into the main

tsuf239 (approved):
Nice!
image

PR #4314, opened 2023-09-27
fix(sdk): aws queue client does not delete popped messages

flyingImer:
## Checklist

- [X] Tests added (always)


libs/wingsdk/src/shared-aws/queue.inflight.ts
```diff
@@ -54,14 +55,25 @@ export class QueueClient implements IQueueClient {
...
-    const data = await this.client.send(command);
+    const data = await this.client.send(receiveCommand);
     if (!data.Messages) {
       return undefined;
     }
-    return data.Messages[0].Body;
+
+    const message = data.Messages[0];
+
+    if (message.ReceiptHandle) {
```
Chriscbr, line 69: What does it mean if there's no receipt handle on the received message / we can't delete it? (curious if it should be an error of some kind, or if we should log something...)

Chriscbr (approved):
LGTM

In the future I wonder if we should rework the Queue API to give the user a bit more control -- e.g. `pop` could return you a `Task` object that can be deleted from the queue with `task.complete()`

libs/wingsdk/src/shared-aws/queue.inflight.ts
```diff
@@ -54,14 +55,25 @@ export class QueueClient implements IQueueClient {
   }
 
   public async pop(): Promise<string | undefined> {
```
tsuf239, line 57: Was it tested on aws?

tsuf239 (approved):
Tests are passing!
image

libs/wingsdk/src/shared-aws/queue.inflight.ts
flyingImer: yes, marked it as WIP to update the sdk test, it's in now

flyingImer:
> LGTM
> 
> In the future I wonder if we should rework the Queue API to give the user a bit more control -- e.g. `pop` could return you a `Task` object that can be deleted from the queue with `task.complete()`

I actually had a couple of POCs on that track, but ended up dropping them. I *think* the current queue API is closer to the classic data structure interface, so I pivoted to match the simulator behavior.

However, I definitely agree with you on the user control story, especially after seeing messages like "queue is not FIFO" in the code base. Queue API might need to re-position itself a bit between classic queue surface and its characteristic of being distributed in the cloud.

libs/wingsdk/src/shared-aws/queue.inflight.ts
flyingImer: Actually, let me log it to start with

Chriscbr:
Ran the queue tests again, all passed 👍

Chriscbr (approved):

PR #4326, opened 2023-09-28
feat(compiler)!: remove `$` from string interpolations

yoav-steinberg:
use `"hello {name}"` instead of `"hello ${name}"`
Motivation: One less char to type, more ergonomic and the `$` char had no practical use. Prefixing a `$` seems like a leftover from languages that used `$` as a var prefix (bash/php/php...). In our context we're using an **expression** inside the string interpolation so we need a start and end token (the curly braces) but anything else seems redundant.

Cons:
 * If you're coming from JS world a bit less familiar, but this can also be treated as a pro since we're trying to remove stuff from JS that never made sense anyway.
 * This means that string representation of a JS objects and OpenAPI url paths need to be escaped/changed:
```wing
// now:
Json.parse("\{ \"a\": 1 }");
// was:
Json.parse("{ \"a\": 1 }");

// Update to cloud.Api: we use express js like syntax to specify params, so we don't need to escape the curly brackets
// now:
api.get("/users/:name", handler);
// was:
api.get("/users/{name}", handler);
```

## Checklist

- [x] Tests added (always)


eladb (approved):

examples/tests/sdk_tests/api/path_vars.test.w
```diff
@@ -20,14 +20,14 @@ let handler_two = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
...
 };
 
-api.get("/users/{name}", handler);
-api.get("/path/{name}", handler);
-api.get("/users/permission/{name}", handler);
-api.get("/path/{name}/{age}", handler_two);
+api.get("/users/\{name}", handler);
+api.get("/path/\{name}", handler);
+api.get("/users/permission/\{name}", handler);
+api.get("/path/\{name}/\{age}", handler_two);
```
yoav-steinberg, line 26: Note we now need to escape these. WDYT?

Chriscbr: Hypothetically we could address this by choosing a different syntax for the API resource: 
```js
api.get("/users/:name", handler);
api.get("/path/:name", handler);
api.get("/users/permission/:name", handler);
api.get("/path/:name/:age", handler_two);
```

yoav-steinberg: I think we're aiming to be OpenAPI compatible so any change to API path syntax will break that.
Another thing we can do is have a raw string syntax for these cases:
`r"I'm a raw string with no escaping so this \ and { can be used freely inside me"`
`r"/path/{var}"`

docs/docs/03-language-reference.md
MarkMcCulloh, line 1: There are many more `.md` files that reference ${}, but only this one is changed in the PR

examples/tests/sdk_tests/api/path_vars.test.w
MarkMcCulloh: I think needing to escape these is a dealbreaker. I'm cool with `r""` although typing an extra letter doesn't feel so nice.

MarkMcCulloh (changes requested):

examples/tests/sdk_tests/api/path_vars.test.w
eladb: Fuck

eladb: After second thought, I think we should consider @Chriscbr's idea to use the ":foo" syntax.

We want to be OpenAPI compatible, but that doesn't imply our route syntax must be the same. We can convert it to openapi downstream. No?

MarkMcCulloh: `:` is a valid character in URLs so I don't think we can get away with that one. Sadly, most useful characters are either valid URLs or otherwise reserved (like `%`, `#`, `[`, `]`)

Chriscbr: > `:` is a valid character in URLs so I don't think we can get away with that one.

I'm guessing this shouldn't be an issue since this is the syntax used in ExpressJS, right? If you want to represent the character `:` verbatim, I'm guessing you can escape it with `::` or something similar.

----

If the `:variable_name` syntax doesn't look Wingy, another alternative is to use angle brackets (it looks like this the syntax used by Python's [Flask](https://flask.palletsprojects.com/en/3.0.x/quickstart/) and [Django](https://docs.djangoproject.com/en/4.2/topics/http/urls/) frameworks).

```js
api.get("/users/<name>", handler);
api.get("/path/<name>", handler);
api.get("/users/permission/<name>", handler);
api.get("/path/<name>/<age>", handler_two);
```

I don't have a preference between the syntaxes. thoughts? @MarkMcCulloh @yoav-steinberg @eladb

ainvoner: I think needing to escape `{` won't be intuitive to Winglang users.

eladb: I vote for the ":field" syntax for api routes.

eladb (approved):

eladb:
@yoav-steinberg lets get this over the finish line.

We need to change cloud.API to use the `:foo` syntax

ekeren:
When we originally did the `api.get("/users/{name}", handler);` I was asking why not doing it express js style `api.get("/users/:name", handler);` and was answered by @eladb  that this is how the [openapi style](https://swagger.io/docs/specification/describing-parameters/) and we should stick to it. 

That answer made sense to me... 

Having said that, I don't have a strong objection against changing it to expressjs style

staycoolcall911:
I guess that concludes the discussion @yoav-steinberg - you can proceed and get that merged in.

eladb:
> was answered by @eladb that this is how the [openapi style(https://swagger.io/docs/specification/describing-parameters/) and we should stick to it.

I reserve the right to change my mind. If I balance the tradeoff between these two decisions, I prefer that we change to `:param`.

eladb (approved):

examples/tests/sdk_tests/api/path_vars.test.w
yoav-steinberg: done

tools/hangar/__snapshots__/test_corpus/valid/for_loop.test.w_compile_tf-aws.md
```diff
@@ -286,96 +286,96 @@ class $Root extends $stdlib.std.Resource {
       {((cond) => {if (!cond) throw new Error("assertion failed: postContinueHits == 0")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(postContinueHits,0)))};
     }
     {((cond) => {if (!cond) throw new Error("assertion failed: j == 3")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(j,3)))};
-    {console.log("---\nfor x in 0..0 { ... }")};
+    {console.log("---\nfor x in 0..0 \{ ... }")};
```
MarkMcCulloh, line 289: Shouldn't the `\` not be here here?

yoav-steinberg: good point, in the JS output we can remove this escaping, even though it doesn't matter much as far as I can tell.

MarkMcCulloh (approved):

skyrpex (approved):

Issue #4118, opened 2023-09-07
Assertion failed for code that works once compiled to JS


garysassano:
#### app.w
```
let j = Json { hello: 123, world: [1, 2, 3] };
assert(Json.values(j) == [Json 123, Json [1, 2, 3]]);
```
I run this command:
```sh
wing compile app.w
```

### This happened:

```
ERROR: assertion failed: Json.values(j) == [Json 123, Json [1, 2, 3]]

target/app2.wsim.961063.tmp/.wing/preflight.js:9
       super(scope, id);
       const j = ({"hello": 123,"world": [1, 2, 3]});
>>     {((cond) => {if (!cond) throw new Error("assertion failed: Json.values(j) == [Json 123, Json [1, 2, 3]]")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })((Object.values(j)),[123, [1, 2, 3]])))};
     }
   }
```

### I expected this:

I expected Wing assertion to work, because the compiled JS code doesn't throw any error when pasted into Node.js REPL.

#### preflight.js
```
const j = ({"hello": 123,"world": [1, 2, 3]});
{((cond) => {if (!cond) throw new Error("assertion failed: Json.values(j) == [Json 123, Json [1, 2, 3]]")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })((Object.values(j)),[123, [1, 2, 3]])))};
```

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

garysassano:
Found about this with @hasanaburayyan while doing some tests.

ekeren:
@Chriscbr FYI

Chriscbr:
It appears to be an issue with how Object.values and equality behave in a VM:

```
let code = `
const j = ({});
require('assert').deepStrictEqual(Object.values(j), []);

`

let vm = require("vm").runInNewContext(code, {
    require: require,
    // Object
});
console.log(vm);

```

ekeren:
I don't think it ever worked

Chriscbr:
I updated the example in my code above -- it's quite interesting that `{}.constructor.prototype === Object.prototype` works outside of the VM but not inside it.

[...]

PR #4336, opened 2023-09-29
feat(sdk): adding std.Map tests

WeepingClown13:
This is the first time I am writing tests, so admittedly my tests are not that versatile. But as I also believe that std tests should be straightforward, I do not totally mind it. Since `keys()` and `values()` methods of maps seems to return an array that fails to assert "is the same as another array of same elements" (which seems to be a [bug](https://github.com/winglang/wing/issues/4118)), I had to iterate over the array values and confirm it works as expected.

closes: #4363

## Checklist

- [ ] Tests added (always)


examples/tests/sdk_tests/std/map.test.w
```diff
@@ -15,3 +15,210 @@ let mvalues = m.values();
...
 assert(mvalues.at(1) == 99);
+
+
+let trilogies = { "PT" => 2, "OT" => 1, "ST" => 3 };
+assert(trilogies.size() == 3);
+
+let mutTrilogies = MutMap<num> { "PT" => 2, "OT" => 1, "ST" => 3 };
+assert(mutTrilogies.size() == 3);
+mutTrilogies.delete("ST"); // One could argue that it's better if this does not exist
+assert(mutTrilogies.size() == 2);
```
tsuf239, line 26: Can you test addition and reassignment too?

examples/tests/sdk_tests/std/map.test.w
```diff
@@ -15,3 +15,210 @@ let mvalues = m.values();
...
+}
+
+
+let forceUsers = { "sith" => ["malak", "vader", "bane"], "jedi" => ["sunrider", "bastila", "bindo"]};
+let userKeys = forceUsers.values();
+let valArr = [["malak", "vader", "bane"], ["sunrider", "bastila", "bindo"]];
+assert(userKeys.length == valArr.length);
+for i in 0..userKeys.length {
+    assert(userKeys.at(i) == valArr.at(i));
+}
```
tsuf239, line 97: it will compare two arrays of size 3, right?

examples/tests/sdk_tests/std/map.test.w
```diff
@@ -15,3 +15,210 @@ let mvalues = m.values();
...
+sithTriumvirate.delete("sion");
+assert(sithTriumvirate == MutMap<str> { "traya" => "lord of betrayal" });
+
+test "delete()" {
+    let sithTriumvirate = MutMap<str> { "traya" => "lord of betrayal", "nihilus" => "lord of hunger", "sion" => "lord of pain" };
+    sithTriumvirate.delete("nihilus");
+    assert(sithTriumvirate.size() == 2);
+    sithTriumvirate.delete("sion");
+    assert(sithTriumvirate == MutMap<str> { "traya" => "lord of betrayal" });
+}
```
tsuf239, line 224: I'd add a check for testing if calling `set`, `delete`, `clear` and `copy` on a non Mut Map throws an error too

tsuf239 (commented):
Awesome work! super thorough work 👏 👏

examples/tests/sdk_tests/std/map.test.w
WeepingClown13: Yeah it would. I originally wanted to do a direct assertion but that fails for some unknown reasons as I mentioned in the description.

examples/tests/sdk_tests/std/map.test.w
WeepingClown13: I actually tried one, I think. But the try-catch wasn't able to catch that error and it was instead failing. It was something like;
```
let a = MutMap<str> {};
let b = a.copy();
try {
    a.delete()
} catch error {
    ## not sure if catch error or just catch is the correct syntax but both fails
    throw("")
}
```

Now this, instead of catching the error, would result in a compilation error inside the try block. So I couldn't figure out how to do it. Any suggestions?

tsuf239: Right! 
that's why we got the invalid tests (tests in the `examples/tests/invalid` folder, that catch compilation errors)
I see that there is only one test for `set` there, can you add tests for the rest?  (it's in examples/tests/invalid/immutable_container_types.test.w, maybe we can change the name of the file)

examples/tests/sdk_tests/std/map.test.w
tsuf239: cool! can you please add a link to the bug?

WeepingClown13: It's already linked in the description, but [here](https://github.com/winglang/wing/issues/4118) it is. I guess I'll also add a direct reference to the bug number in the description.

tsuf239: I'm sorry, I meant as a comment in the code (so the next one to read/ copy this code will be aware of the issue)

revitalbarletz:
@tsuf239 can you approve if you think it is ready?

revitalbarletz:
@WeepingClown13 great work!!!

tsuf239 (approved):
all passed, great job!

WeepingClown13:
Oh thanks. I just added another commit to add a comment about the mentioned bug and finished building it but sure😂

Issue #4331, opened 2023-09-29
Implement `cloud.Bucket` inflight method `metadata` for AWS targets


Chriscbr:
```js
/**
 * Get the metadata of an object in the bucket.
 */
inflight metadata(key: str): ObjectMetadata;

/**
 * Metadata of a bucket object.
 */
struct ObjectMetadata {
  /** The size of the object in bytes. */
  size: num;
  /** The time the object was last modified. */
  lastModified: Datetime;
  /** The content type of the object, if it is known. */
  contentType: str?;
}
```

staycoolcall911:
Thanks for creating these issues @Chriscbr - this should also resolve #1686, right?

Chriscbr:
@staycoolcall911 It should get us most of the way there - we also need to allow specifying the mime type in `bucket.put()` as well

monadabot:
Congrats! :rocket: This was released in [Wing 0.37.2](https://github.com/winglang/wing/releases/tag/v0.37.2).

Issue #4330, opened 2023-09-29
Implement `cloud.Bucket` inflight method `metadata` for simulator target


Chriscbr:
```js
/**
 * Get the metadata of an object in the bucket.
 */
inflight metadata(key: str): ObjectMetadata;

/**
 * Metadata of a bucket object.
 */
struct ObjectMetadata {
  /** The size of the object in bytes. */
  size: num;
  /** The time the object was last modified. */
  lastModified: Datetime;
  /** The content type of the object, if it is known. */
  contentType: str?;
}
```

monadabot:
Congrats! :rocket: This was released in [Wing 0.37.2](https://github.com/winglang/wing/releases/tag/v0.37.2).

PR #4338, opened 2023-09-29
feat(sdk): implement cloud.Bucket inflight method metadata for AWS targets and Simulator

exoego:
Closes https://github.com/winglang/wing/issues/4331
Closes https://github.com/winglang/wing/issues/4330

This PR implements inflgiht method `Bucket.metadata` for AWS.
Other implementations are stubbed.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -345,6 +346,27 @@ export interface IBucketClient {
    * @inflight
    */
   publicUrl(key: string): Promise<string>;
+
+  /**
+   * Get the metadata of an object in the bucket.
```
Chriscbr, line 351:
```suggestion
   * Get the metadata of an object in the bucket. Throws if there is no object with the given key.
```

Chriscbr (commented):
Thanks for picking this up @exoego - this looks like a great start!

I think we need two more things:
1. Add a test written in Wing to our examples directory here: https://github.com/winglang/wing/tree/main/examples/tests/sdk_tests/bucket
2. I suspect we also need to specify what permissions are needed by `metadata` - check out this function: https://github.com/winglang/wing/blob/6a9ce9de4216f8f02e466c746b1e24f42713a5cc/libs/wingsdk/src/shared-aws/permissions.ts#L84

Chriscbr (changes requested):
See comments

libs/wingsdk/src/cloud/bucket.ts
exoego: I've addressed that in https://github.com/winglang/wing/pull/4338/commits/5cfe0cc417cfff1adea53e4dde4ce4ce1d4d06df
`@throws` annotation is used.

exoego:
@Chriscbr Request was addressed.

examples/tests/sdk_tests/bucket/metadata.test.w
```diff
@@ -0,0 +1,17 @@
+bring cloud;
+
+let b = new cloud.Bucket();
+
+test "metadata" {
+  b.put("test1.txt", "Foo");
+
+  assert(b.metadata("test1.txt").size == 3);
+  assert(b.metadata("test1.txt").contentType == nil);
```
Chriscbr, line 9: When I ran this test on my AWS account, I got an error because the contentType was not nil. To make sure the AWS and `sim` targets behave the same, I think we need the following:

- (a) `put` for AWS implementations should always specify a content type so it isn't automatically set by AWS - I think the AWS SDK already sets it to [application/octet-stream](https://github.com/aws/aws-sdk-js-v3/blob/fa961c2350b650a1c29cc09306f0e2e0b9eb6c4e/clients/client-s3/src/protocols/Aws_restXml.ts#L3320) by default but it's a good practice for us to set it
- (b) the `sim` implementation should track a content type for each file, such as inside a JavaScript `Map`

Chriscbr (changes requested):

examples/tests/sdk_tests/bucket/metadata.test.w
exoego: Ah, I thought this test is executed only in Simulator, but we can execute in actual clouds?
How should I execute this test file in AWS?
I've checked `05-development.md` but could not figure out.

exoego: In my personal opinion, there should be a `content-type` argument to `put` methods and let users to specify it.
It would be user-friendly if content-type is automatically guessed by Wing using a 3rd party library, but there should be an escape hatch in case such a library does not recognize content-type.

Chriscbr: Sorry for any confusion with our development guide. You can execute any test in AWS by running Wing's  `test` command with the `--target tf-aws` flag. For example `wing test --target tf-aws metadata.test.w`.

> In my personal opinion, there should be a `content-type` argument to `put` methods and let users to specify it. It would be user-friendly if Wing automatically guessed content-type using a 3rd party library, but there should be an escape hatch in case such a library does not recognize content-type.

Totally, I think both suggestions make sense to me. I didn't want to add additional scope to this PR, but both can be supported. Would you mind creating an issue on GitHub for these enhancements? 🙏

revitalbarletz:
Hey @exoego !!! How about joining our [slack](https://t.winglang.io/slack)? It is easier to address some questions over there and we have a great community :) so its totally win win !

examples/tests/sdk_tests/bucket/metadata.test.w
exoego: @Chriscbr 
Updated the PR. 
- https://github.com/winglang/wing/pull/4338/commits/350470283cb4344e4890eb50ef5d80607975465f Sim returns `application/octet-stream` until we allow custom content-type.
- https://github.com/winglang/wing/pull/4338/commits/34456836cdd784c006948586672b6847794ecbb5 Fixes AWS implementation. I've found that a generic `"403"` error is thrown instead of `NotFound` when `s3:ListBucket` (underlying action for `s3:ListObject(V2)`) is not permitted.

As far as I tested, the test now succeeds both on Simulator and TF-AWS.

exoego:
> How about joining our [slack](https://t.winglang.io/slack)

Thanks, but no, thanks.
I do not like that Google-unsearchable places (Discord or Slack) are used for questions and discussions in OSS community.

Chriscbr (approved):
Stellar work @exoego! 🎉

PR #4342, opened 2023-09-29
feat(cli): compile directories

Chriscbr:
The wing toolchain now supports compiling directories. For example:

```
$ wing compile .
```

When a directory is compiled, no preflight code is executed (this only happens when an entrypoint file like `main.w` is passed to `wing compile`).

Required for https://github.com/winglang/wing/pull/3938

## Checklist

- [x] Tests added (always)

### DIFF
--- a/apps/wing/src/commands/compile.test.ts
+++ b/apps/wing/src/commands/compile.test.ts
@@ -6,6 +6,7 @@ import { Target } from "@winglang/compiler";
 import { generateTmpDir } from "src/util";
 
 const exampleDir = resolve("../../examples/tests/valid");
+const exampleSmallDir = resolve("../../examples/tests/valid/subdir2");
 const exampleFilePath = join(exampleDir, "captures.test.w");
 
 describe(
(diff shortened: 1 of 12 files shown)


eladb (approved):

MarkMcCulloh:
This isn't really compilation, is it? I always figured `compile` for wing means preflight execution. Also why is a directory special? A directory and normal wing file are both modules, so the behavior of this command changing based on file vs directory is unexpected.

Chriscbr:
> This isn't really compilation, is it? I always figured compile for wing means preflight execution.

The motivation of the change comes from #3938 - the idea is `wing pack` will (1) compile your project (i.e. the current directory) and (2) package it into a sharable artifact. My thinking was that it would be weird if there wasn't a way to just do part (1) and compile your current project without packaging it.

> Also why is a directory special? A directory and non-entrypoint wing file are both modules, so the behavior of this command changing based on file vs directory is unexpected.

They're really treated the same, I didn't explain this in the PR description well. Compiling any unit of code that doesn't represent an entrypoint will not synthesize application artifacts (i.e. a Terraform manifest for the `tf-aws` target, or a `simulator.json` for the `sim` target), excluding tests in the future.

PR #4349, opened 2023-09-30
chore(repo): ready for implementation field in issue templates

staycoolcall911:
Adding another field to issue templates, to indicate whether this issue is ready to start implementation or whether some further discussion is needed first.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -37,16 +37,6 @@ body:
     label: "Is there a workaround?"
     placeholder: "What's the workaround to avoid this issue?"
 - type: dropdown
-  id: impl-ready
-  attributes:
-    label: "Ready for implementation?"
-    description: "Does this issue contain enough info to start implementation or does it need further discussion?"
-    options:
-    - "Yes"
-    - "No"
-  validations:
-    required: true
-- type: dropdown
   id: Component
   attributes:
     label: "Component"
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -37,6 +37,16 @@ body:
     label: "Is there a workaround?"
     placeholder: "What's the workaround to avoid this issue?"
 - type: dropdown
+  id: impl-ready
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -36,6 +36,16 @@ body:
   attributes:
     label: "Is there a workaround?"
     placeholder: "What's the workaround to avoid this issue?"
+- type: dropdown
```
eladb, line 39: Not sure we need this for bugs. Let's not overload too much if not needed.

.github/ISSUE_TEMPLATE/feature.yml
```diff
@@ -34,6 +34,16 @@ body:
     description: 
     placeholder: |
       Any thoughts regarding how this should be implemented?
+- type: dropdown
```
eladb, line 37: Perhaps instead of a user controlled field (which feels quite weak), we can do this:

1. Add a default label for all features that says "need discussion"
1. Add a note in the community notes section that says not to pick an item if it has that label.

Then, it is on us to make clear out this label when we groom the backlog (or schedule for discussion, request an RFC, etc).

Wdyt?

eladb (changes requested):

.github/ISSUE_TEMPLATE/feature.yml
staycoolcall911: I think most issues will not need this label, so it's better to do it the other way around: new issues are created without the label, and we add it if needed during triage.
That said, I'm also fine starting as you suggested.

eladb: I am not sure I agree. Feels like it would be healthy for all *feature requests* to have some kind of review by someone from the team before they are ready to be implemented, no?

staycoolcall911: I review everything anyway, so it's just a matter of adding the needs-discussion label or removing it.
Let's start with adding it automatically to alm issue. Fine by me

eladb (approved):

PR #4351, opened 2023-10-01
rfc: compute resources requirements

eladb:
An RFC which lists the requirements for compute resources in the Wing SDK

### DIFF
--- a/docs/contributing/999-rfcs/2023-10-01-compute-resources.md
+++ b/docs/contributing/999-rfcs/2023-10-01-compute-resources.md
@@ -1,0 +1,12 @@
+---
+title: "#4350 Compute Resources"
+description: Requirements for compute resources in the Wing Cloud Library
+---
+
+# Compute Resources - Requirements Specification
+
+- **Author(s)**: @eladb
+- **Submission Date**: 2023-10-01
+- **Stage**: Proposal
+- **Stage Date**: 2023-10-01
+


docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,252 @@
...
+
+1. EKS/GKE/AKS (P1)
+2. [ECS](https://aws.amazon.com/ecs/)
+3. Helm (basically Kubernetes `Deployment`, `Service` and `Ingress` resources)
+4. [fly.io](https://fly.io)
+5. [ControlPlane](https://github.com/controlplane-com/terraform-provider-cpln/blob/main/docs/resources/workload.md#nestedblock--options)
+
+The default behavior for `tf-aws` should be to create a dedicated EKS cluster for the app (P1) but
+users should be able to point to an existing EKS cluster through platform configuration options
+(P2).
```
MarkMcCulloh, line 148: Why kubernetes as a default? If you're targeting AWS, it would make sense to use their primary container orchestrator. ECS would be faster, simpler, cheaper, and still sufficiently powerful.

It probably seems like doing EKS/GKE/AKS is like killing 3 birds with 1 stone, but I'm not convinced there will be much shared work between the three.

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,252 @@
...
+Requirements are marked with a *Reqtag* so they will be easy to trace in subsequent design
+documents. Suggested priorities are marked inline as "(P1)", "(P2)" or "(P3)".
+
+### Containerized workloads (P1)
+
+> Reqtag: `c:workload`
+
+Wing must support defining **containerized workloads** implemented through user-defined container
+images. Images may either be pulled from a container registry or built from a `Dockerfile` within
+the project directory.
```
MarkMcCulloh, line 27: What's the relationship between "Workload", "Container", and "Function". Can I describe a container workload to run in a Function?

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,252 @@
...
+* Environment variables (P1)
+* Entrypoint command to override (if you wish to use a pre-built container image and just change the
+  entrypoint) (P1)
+* CPU and memory requirements (in some normalized model such as vCPUs, GBs) (P2)
+* GPUs (P2)
+* Liveness probe
+* Readiness probe (P1)
+* Sidecar containers (P2)
+* Volumes (P2)
+* Autoscaling (P1)
```
MarkMcCulloh, line 113: I feel like we're conflating containers and container orchestrators. Most of the time this document mentions containers I feel like it's talking about pods. Am I thinking about this wrong?

docs/contributing/999-rfcs/2023-10-01-workloads.md
eladb: The use case of running containers within serverless platforms is currently out of scope for this RFC, as well as using any other runtime within serverless functions. We'll scope that out separately.

I'll add a note.

As for the terminology, I'll add some definitions at the top.

docs/contributing/999-rfcs/2023-10-01-workloads.md
eladb: Kubernetes is by far the most common way to run containers on the cloud, both based on data and our conversations with customers.

Indeed, the cluster provisioning is very different across managed k8s platforms, but the actual k8s resources are going to be the same.

Having said that, like you said, it shouldn't be too hard to deploy workloads into ECS and I believe we will support this use case quite quickly.

I think the more common scenario will actually be to reference an existing cluster instead of provisioning one for each wing app, but we want wing standard platforms to work out of the box.

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Requirements
+
+Requirements are marked with a *Reqtag* so they will be easy to trace in subsequent design
+documents. Suggested priorities are marked inline as "(P1)", "(P2)" or "(P3)".
+
+### Containerized workloads (P1)
+
+> Reqtag: `c:workload`
+
```
eladb, line 24:
```suggestion

A "workload" describes a unit of scalable deployment. It's akin to a K8S's deployment + service + ingress, or ECS's task or fly.io's machine or ControlPlane's workload.
```

docs/contributing/999-rfcs/2023-10-01-workloads.md
MarkMcCulloh: > Indeed, the cluster provisioning is very different across managed k8s platforms, but the actual k8s resources are going to be the same.

I don't think this is true. In my experience, doing EKS properly means resources will have aws-specific things. This ranges from simple annotations and sidecars to setting up specific [controllers](https://github.com/kubernetes-sigs/aws-load-balancer-controller/).
It's been a few years and it looks like they've done some work to build in some of this to EKS, so that's good, but the manifests definitely won't be exactly the same.

I know this is just using k8s for the sake of k8s so there's no changing this path, just wanted to keep voicing my concerns. I think k8s is extremely hard to do correctly wrt security and resilience and it will be very easy for us to hand over a terribly configured resources since that's the default k8s experience. Think this is fine if the user wants to explicitly opt-in to a kubernetes experience, but if they just wanted to run a container then defaulting them to this is doing a disservice imo.

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,262 @@
...
+### Simulator implementation (P1)
+
+> Reqtag: `c:simulator`
+
+When using Docker images, the Wing simulator will need to take care of seamlessly building and/or
+pulling the container image and maintaining it's lifecycle and cleanup via a local docker
+installation.
+
+If docker is not installed, an error should provide useful instructions on how to install it.
```
hasanaburayyan, line 38: Since we plan to inform the user with useful instructions on how to install docker, should we also provide them with useful instructions on how to ensure they have permissions to push/pull from container registries (in the event of a permissions issue)? 

I.E. if they are using ECR/GCR/etc.., we (Wing) are able to configure permissions to ECS/K8/etc.. but for local testing that is on them to have a configured credentials (I assume the same way its on them to have docker installed)

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Hot reloading for fast local iteration (P1)
+
+> Reqtag: `c:hot-reloading`
+
+When developing a Wing application which references a locally defined container image, we would like
+to support live updates which to instantly reload the program when the source changes.
```
hasanaburayyan, line 45: > when the source changes

Is this a reference to the source code thats bundled in the container, the Dockerfile, or the local image tag? Maybe its a combination, just wanted to clarify how this hot-reload picture looks.

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,262 @@
...
+The following is a non-exhaustive list of capabilities we will need to support:
+
+* Expose network ports of various types (http, http2, grpc, ??) (P1)
+* Command line arguments to pass to the container (P1)
+* Environment variables (P1)
+* Entrypoint command to override (if you wish to use a pre-built container image and just change the
+  entrypoint) (P1)
+* CPU and memory requirements (in some normalized model such as vCPUs, GBs) (P2)
+* GPUs (P2)
+* Liveness probe
```
hasanaburayyan, line 110: I assume this is missing the P1 classification
```suggestion
* Liveness probe (P1)
```

hasanaburayyan (commented):
Still making my way through, left a few thoughts so far.

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Design Notes
+
+This section includes a few notes related to the design and implementation of these requirements.
+Consider them recommendations and pointers for the next phase of this project.
+
+### SDK resources
+
+I would recommend implementing this as a single "monolithic" resource called `cloud.Workload` which
+will expose all of the features described in this spec. This will produce a rich model for
```
Chriscbr, line 239: Checking if I understand this correctly - would `cloud.Service` be implemented on top of `cloud.Workload`? i.e. `cloud.Service` might package inflight code into a NodeJS dockerfile / image, and then pass that dockerfile / image to `cloud.Workload`?

Chriscbr (commented):
Exciting

docs/contributing/999-rfcs/2023-10-01-workloads.md
Chriscbr: > The use case of running containers within serverless platforms is currently out of scope for this RFC

Do you just mean within FaaS platforms, or outside of any managed container platforms? I've heard some folks call ECS serverless, but honestly even I get easily confused sometimes 😅

docs/contributing/999-rfcs/2023-10-01-workloads.md
eladb: Moved autoscaling into a separate topic. I would map "workload" to a combination of "service + deployment", not "pod".

docs/contributing/999-rfcs/2023-10-01-workloads.md
eladb: > doing EKS properly means resources will have aws-specific things

You are right, the manifests won't be exactly the same. There could be some specific additions to the K8S resources synthesized based on the target platform, but the general setup is going to be similar.

> I think k8s is extremely hard to do correctly wrt security and resilience and it will be very easy for us to hand over terribly configured resources since that's the default k8s experience

That's another good reason to dive in head first and not punt these challenges.

> defaulting them to this is doing a disservice imo

I see what you mean. We can discuss what makes sense as the default backend for `cloud.Workload` on AWS down the road (it's totally possible that ECS Fargate will make more sense). We need to prioritize one or the other for our MVP, and there's a good case for starting with Kubernetes, especially given there will be some reuse across providers.

I'll tweak the wording.

eladb: @chriscbr I added a "Cluster" section with some details. Let me know if this requires additional info.

docs/contributing/999-rfcs/2023-10-01-workloads.md
```diff
@@ -0,0 +1,262 @@
...
+   somehow emit typescript type definitions for development (another reason why the Wing compile
+   needs to perform the build).
+1. Call any Wing inflight API from other languages (P3): this is lower priority for now given the
+   technical complexity but it would be healthy to start thinking about the path forward here. This
+   could also be a multi-phased project where we start with something like a sidecar/child process
+   (with generated client bindings) and then compile our inflight clients to WebAssembly when the
+   ecosystem is mature enough (currently there is no AWS SDK for WebAssembly). 
+
+The bindings APIs will need to also allow the user to specify the qualification of the bind (e.g.
+the permission requirements) so that the relevant permissions can be created.
```
hasanaburayyan, line 193: Probably also needs to handle subnet specifications

eladb: I think that's more related to the platform implementation, no?

hasanaburayyan: I think platforms  are the solution to permissions and networking.

Though my thought process here is if we allow this bind for permissions why not allow easy network configs as well? 

I.E. someone who doesn’t want to write a platform for a small use case and just needs their containers to be added to a security group to access maybe some Redis cluster.

unless I’m thinking about this wrong

eladb: Perhaps... Let's see how this evolves and iterate.

hasanaburayyan (approved):

skyrpex (approved):

staycoolcall911:
@eladb - can merge that in?

Issue #4052, opened 2023-09-03
Do not allow ambiguous api paths


tsuf239:
### I tried this:

```
api.get("/{param}/{something}", handler);
api.get("/path/{something}", handler)
```

Those paths are ambiguous- since for the call `${api.url}/path/xxx` I can be directed to both of them.

### This happened:

The app failed only when tried to access one of the paths- during a test, and after deploying to aws...

### I expected this:

To throw a compilation error

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

garysassano:
Would it be possible to have the full code so that I can replicate it on my side?

garysassano:
Would it be possible to have the full code so that I can replicate the issue on my side?

tsuf239:
```
bring cloud;
let api = new cloud.Api();
let handler = inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
  return cloud.ApiResponse {
    status: 200,
    body: body
  };
});

api.get("/{param}/{something}", handler);
api.get("/path/{something}", handler)
```

tsuf239:
this is resolved by https://github.com/winglang/wing/pull/4352

PR #4291, opened 2023-09-26
fix(sdk): ambiguous api paths are allowed

gcfbn:
Fixes the issue https://github.com/winglang/wing/issues/4052 by adding a check if newly added path is ambiguous with paths already added to the API.

## Checklist

- [x] Title matches [Winglang's style guide](https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted)
- [x] Description explains motivation and solution
- [x] Tests added (always)
- [ ] Docs updated (only required for features)
- [ ] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Wing Cloud Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.

PR #4352, opened 2023-10-01
fix(sdk): ambiguous api paths are allowed

gcfbn:
Fixes the issue https://github.com/winglang/wing/issues/4052 by adding a check if newly added path is ambiguous with paths already added to the API. I've re-created the https://github.com/winglang/wing/pull/4291 PR, as it got stuck on merging main to the feature branch.

## Checklist

- [x] Tests added (always)


tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -250,6 +250,23 @@ error: Unknown symbol \\"this\\"
...
  
+Tests 1 failed (1)
+Test Files 1 failed (1)
+Duration <DURATION>"
+`;
+
+exports[`ambiguous_api_paths.test.w 1`] = `
+"ERROR: Endpoint for path '/test/{variable}' and method 'GET' is ambiguous - it conflicts with existing endpoint for path '/test/path'
+
+../../../examples/tests/invalid/target/test/ambiguous_api_paths.test.wsim.157309.tmp/.wing/preflight.js:38
```
tsuf239, line 261: We probably need to sanitize the number here: `ambiguous_api_paths.test.wsim.157309.tmp`
(next to this code, I think- https://github.com/winglang/wing/blob/8eab41aec4294c4299124c186f81868f587df591/tools/hangar/src/utils.ts#L53-L55)

since we generate a new temp directory for each build, and we are re-generating the snapshot it doesn't match, and it never matches because of the that number, it keeps re-building over and over (that's the monabot commits you see)

tsuf239:
the tests pass on aws!
image
I'll approve as soon as we can fix the test snapshot (let me know if any help is needed), canceling the build for now

tsuf239:
Another option I can suggest is to comment out the invalid test code and I'll open an issue and fix it at our end :)

gcfbn:
> Another option I can suggest is to comment out the invalid test code and I'll open an issue and fix it at our end :)

Alright, if you're fine with this approach, it works for me too.

gcfbn:
@tsuf239 Apparently it won't work for a file without the tests

tsuf239:
I see, so it's ok to remove this file, I'll add it back in a new PR with the path serialization fixed

tsuf239 (approved):

PR #4370, opened 2023-10-02
fix: cloud.Service cant capture api.url

### LINKED ISSUE #4369: `cloud.Service` can't capture `api.url`
### I tried this:

compiling this code:

```ts
bring cloud;

let api = new cloud.Api();

  log(api.url);
});
```

### This happened:

got this error:

```
eladcohen@Elads-w-MacBook-Pro tmp % wing compile main.w
ERROR: Tokens can only be bound by a Function for now

target/main.wsim.197650.tmp/.wing/preflight.js:37
         _registerBind(host, ops) {
           if (ops.includes("handle")) {
>>           $Closure1._registerBindObject(api.url, host, []);
           }
           super._registerBind(host, ops);
```

### I expected this:

code to compile

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

eladcon:
Tokens in simulator can also bind now to a `cloud.Service`

Fixes #4369 

## Checklist

- [X] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/std/resource.ts
+++ b/libs/wingsdk/src/std/resource.ts
@@ -14,11 +14,6 @@ export interface IInflightHost extends IResource {
    * Adds an environment variable to the host.
    */
   addEnvironment(name: string, value: string): void;
-
-  /**
-   * Returns the set of environment variables for this host.
-   */
-  get env(): Record<string, string>;
 }
 
 /**
--- a/libs/wingsdk/src/std/resource.ts
+++ b/libs/wingsdk/src/std/resource.ts
@@ -14,6 +14,11 @@ export interface IInflightHost extends IResource {
    * Adds an environment variable to the host.
    */
   addEnvironment(name: string, value: string): void;
+
+  /**
+   * Returns the set of environment variables for this host.
+   */
+  get env(): Record<string, string>;
 }
 
 /**
--- a/libs/wingsdk/src/target-sim/tokens.ts
+++ b/libs/wingsdk/src/target-sim/tokens.ts
@@ -63,10 +63,7 @@ export class SimTokens extends Tokens {
     switch (typeof value) {
       case "string":
         const envName = this.envName(value);
-        // the same token might be bound multiple times by different variables/inflight contexts
-        if (host.env[envName] === undefined) {
-          host.addEnvironment(envName, value);
-        }
+        host.addEnvironment(envName, value);
         break;
       default:
         throw new Error(`Unable to bind token ${value}`);
--- a/libs/wingsdk/src/target-sim/tokens.ts
+++ b/libs/wingsdk/src/target-sim/tokens.ts
@@ -1,4 +1,3 @@
-import { Function, Service } from "../cloud";
 import { Tokens } from "../core/tokens";
 import { IInflightHost, IResource } from "../std";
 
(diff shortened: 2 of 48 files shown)


libs/wingsdk/src/target-sim/tokens.ts
```diff
@@ -61,10 +60,6 @@ export class SimTokens extends Tokens {
    * Binds the given token to the host.
    */
   public bindValue(host: IInflightHost, value: any) {
-    if (!(host instanceof Function)) {
```
tsuf239, line 64: I wonder if it'll be better to remove this restriction only on the sim right now (because it was tested only for the sim) and to open an issue for the rest of the targets

libs/wingsdk/src/std/resource.ts
```diff
@@ -14,6 +14,11 @@ export interface IInflightHost extends IResource {
    * Adds an environment variable to the host.
    */
   addEnvironment(name: string, value: string): void;
+
+  /**
+   * Returns the set of environment variables for this host.
+   */
+  get env(): Record<string, string>;
```
tsuf239, line 21: This is nice :)  I'm just not sure if any resource can hold an `env`, maybe you can add it only to the container for now?

libs/wingsdk/src/target-sim/tokens.ts
tsuf239: maybe we can also narrow the restriction to NOT function and NOT a container? (instead of removing it completely?)

libs/wingsdk/src/target-sim/tokens.ts
```diff
@@ -61,8 +61,10 @@ export class SimTokens extends Tokens {
...
```
Chriscbr, line 64: I think it's ok to remove this check now since we recently added the capability in the SDK that all IInflightHost are expected to implement addEnvironment - which is all the tokenizer needs.

Chriscbr (approved):

libs/wingsdk/src/std/resource.ts
Chriscbr: Let's try to avoid adding direct `env` access in `IInflightHost` if it's possible for now. Accessing the environment variables be error prone since when you write `let x = host.env();` in userland, it's possible someone else will call `addEnvironment(...)` on the host afterwards, but then `x` will not reflect the new environment variable, leading to hard to find bugs.

libs/wingsdk/src/target-sim/tokens.ts
Chriscbr: Over here:
```js
        const envName = this.envName(value);
        // the same token might be bound multiple times by different variables/inflight contexts
        if (host.env[envName] === undefined) {
          host.addEnvironment(envName, value);
        }
```

What if we change `addEnvironment` so you can call it multiple times with the same key, as long as the value that is already stored is the same? I'm not sure but I wonder if this could solve the use case

```js
host.addEnvironment("foo", "bar");
host.addEnvironment("foo", "bar"); // OK
host.addEnvironment("foo", "baz"); // runtime error!
```

Chriscbr: https://github.com/winglang/wing/pull/4379

eladb (commented):
Nice

PR #4373, opened 2023-10-02
chore: rerun baseline benchmark as part of workflow

MarkMcCulloh:
Now the benchmark runs twice: Once for the latest version and once for the PR. They run on the same machine in the same job, so there are less variables to worry about.


staycoolcall911 (approved):
This is great @MarkMcCulloh! Elegant and smart change, making benchmarks very useful

PR #4376, opened 2023-10-02
feat(sdk): target tf-gcp bucket inflight client support

### LINKED ISSUE #1282: cloud.Bucket - GCP inflight client
### Community Note

### Feature Spec

Implement an inflight client for GCP Bucket. This is a client that would be used inflight whenever a Bucket is referenced inflight in JavaScript.

For reference, here is the implementation of the inflight client for the Azure bucket (https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-tf-azure/bucket.inflight.ts) (original PR: https://github.com/winglang/wing/pull/832). The PR shows how dependencies are added through `.projenrc.ts`, and unit tests were added using the jest library. Check out the [contributing guide](https://docs.winglang.io/contributors/) for more tips and pointers.

### Use Cases

Prepare `Bucket` so that it can be used inflight once `cloud.Function` is implemented for GCP.

### Implementation Notes

_No response_

### Component

SDK

subh-cs:
Fixes #1282 
This PR is adding support of GCP Bucket inflight client by
1. adding bucket membership to function resource.
2. adding all of the bucket methods from compatibility matrix.
3. adding unit tests by mocking Google Storage Bucket Resource at the test file.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-tf-gcp/bucket.ts
```diff
@@ -32,6 +33,11 @@ const BUCKET_NAME_OPTS: NameOptions = {
   includeHash: false,
 };
 
+enum StorageBucketPermissions {
+  READ = "roles/storage.objectViewer",
+  READWRITE = "roles/storage.objectAdmin",
```
wzslr321, line 38: https://cloud.google.com/iam/docs/understanding-roles#cloud-storage-roles

Isn't `objectAdmin` too much? I believe we should stick to the [Zero Trust](https://cloud.google.com/blog/topics/developers-practitioners/zero-trust-and-beyondcorp-google-cloud) model. I wonder if [objectUser](https://cloud.google.com/iam/docs/understanding-roles#cloud-storage-roles) would work here, though it doesn't have a note:
```
Lowest-level resources where you can grant this role:

Bucket
```
But it seems to have all the needed permissions. Can we apply it here?

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -20,6 +21,11 @@ const FUNCTION_NAME_OPTS: NameOptions = {
   case: CaseConventions.LOWERCASE,
 };
 
+enum StorageBucketPermissions {
```
wzslr321, line 24: The same permissions are defined [here](https://github.com/winglang/wing/pull/4376/files#diff-efe32a53e4d5756b6117f10699793545c18a00d1352c26bcc576b34884a60464R36), can we merge them to avoid duplicity?

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -108,6 +116,49 @@ export class Function extends cloud.Function {
...
+        {
+          bucket: bucket.bucket.name,
+          role: "roles/storage.objectViewer",
+          member: `projectViewer:${app.projectId}`,
+        }
+      );
+      this.bucketPermission.set(
+        `${bucket.bucket.name}-${StorageBucketPermissions.READ}`,
+        newReadPermission
+      );
```
wzslr321, line 144: nit: can add early return to remove `else if`

libs/wingsdk/src/target-tf-gcp/bucket.inflight.ts
wzslr321, line 1: I believe this file should be placed in `src/shared-gcp` directory to match other clouds

https://github.com/winglang/wing/blob/main/libs/wingsdk/src/shared-aws/bucket.inflight.ts

libs/wingsdk/src/target-tf-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,224 @@
...
+  private _public: boolean;
+  private bucketName: string;
+  private storage: Storage;
+  private bucket: Bucket;
+
+  constructor(
+    bucketName: string,
+    isPublic: boolean = false,
+    storage: Storage,
+    projectId: string = process.env.GCP_PROJECT_ID || "wingsdk-test"
```
wzslr321, line 15: I wonder if it can lead to conflicts, since we can also specify the projectId in [AppProps](https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-tf-gcp/app.ts#L17)

wzslr321:
Added a few comments to consider, looks awesome overall!

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -108,6 +116,49 @@ export class Function extends cloud.Function {
     );
   }
 
+  public addBucketPermission(
```
wzslr321, line 119: Can we somehow make this generic? Like in the azure target [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-tf-azure/function.ts#L198)

libs/wingsdk/src/target-tf-gcp/bucket.ts
subh-cs: Yes, I also believe "roles/storage.objectUser" will be perfect fit in this scenario. Thanks for pointing this out.

libs/wingsdk/src/target-tf-gcp/function.ts
subh-cs: I'm moving this to shared-gcp/bucket-inflight.ts and then importing it from there to bucket.ts and function.ts preflight resource file to avoid duplicity. Thanks!!
```js
import { StorageBucketPermissions } from "../shared-gcp/bucket.inflight";
```

libs/wingsdk/src/target-tf-gcp/function.ts
subh-cs: Are you suggesting this? 
```js
if(permission!==StorageBucketPermissions.READ && permission!==StorageBucketPermissions.READWRITE){
      throw new Error("Invalid permission");
      }
```
Then it would be only else
```js
else {
      let newReadWritePermission = new StorageBucketIamMember(
        this,
        "BucketPermissionToWrite",
        {
          bucket: bucket.bucket.name,
          role: "roles/storage.objectAdmin",
          member: `projectEditor:${app.projectId}`,
        }
      );
      this.bucketPermission.set(
        `${bucket.bucket.name}-${StorageBucketPermissions.READWRITE}`,
        newReadWritePermission
      );
    }
```

libs/wingsdk/src/target-tf-gcp/bucket.inflight.ts
subh-cs: In this scenario, it will accept the props first if it is passed. if not, then it will check the env for `GCP_PROJECT_ID` and if still it didn't find anything it will gonna use `wingsdk-test` as default.

I'm doing it this way because during the test at test/shared-gcp/bucket-inflight.test.ts we're defining a mockeStorage bucket instance without projectId

wzslr321: Sounds cool, thank you for the clarification :)

libs/wingsdk/src/target-tf-gcp/function.ts
subh-cs: yes, I also just noticed. instead of enum, I did this. 
```js
if (permission === StorageBucketPermissions.READ) {
      let newReadPermission = new StorageBucketIamMember(
        this,
        "BucketPermissionToRead",
        {
          bucket: bucket.bucket.name,
          role: permission,
          member: `projectViewer:${app.projectId}`,
        }
      );
      this.bucketPermission.set(
        `${bucket.bucket.name}-${permission}`,
        newReadPermission
      );
    }
```
I directly used the permissions props.

wzslr321: Great! :)

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -95,6 +113,25 @@ export class Function extends cloud.Function {
       timeout: props.timeout?.seconds ?? 60,
       environmentVariables: props.env ?? {},
     });
+
+    // apply permissions to bound resources
+    if (this.permissions) {
+      for (const [uniqueId, roleDefinitions] of this.permissions) {
+        const scopedResource = this.node.tryFindChild(uniqueId);
```
subh-cs, line 120: I'm wondering if I can find the resource this way. Need suggestions!!

wzslr321:
I am not sure about these permissions & bound resources. I believe it'd be best if someone from the core team would check this out.

libs/wingsdk/src/target-tf-gcp/function.ts
revitalbarletz: @tsuf239 any tips here?

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,238 @@
...
+  private storage: Storage;
+  private bucket: Bucket;
+
+  constructor(
+    bucketName: string,
+    isPublic: boolean = false,
+    storage: Storage,
+    projectId: string = process.env.GCP_PROJECT_ID || "wingsdk-test"
+  ) {
+    this._public = isPublic;
```
tsuf239, line 22: is there a way to make the bucket itself private?

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,238 @@
...
+      const [url] = await this.bucket.file(key).getSignedUrl({
+        action: "read",
+        expires: Date.now() + expiryTimeInSeconds * 1000,
+      });
+      return url;
+    } catch (error) {
+      throw new Error(`Failed to get signed URL. (key=${key})`);
+    }
+  }
+}
```
tsuf239, line 238: looks good!

libs/wingsdk/src/target-tf-gcp/bucket.ts
```diff
@@ -81,13 +86,60 @@ export class Bucket extends cloud.Bucket {
...
-    throw new Error("Method not implemented.");
+  public bind(host: IInflightHost, ops: string[]): void {
+    if (!(host instanceof GCPFunction)) {
+      throw new Error("buckets can only be bound by tfgcp.Function for now");
+    }
+    if (
+      ops.includes(cloud.BucketInflightMethods.GET) ||
+      ops.includes(cloud.BucketInflightMethods.LIST) ||
+      ops.includes(cloud.BucketInflightMethods.GET_JSON)
+    ) {
```
tsuf239, line 97: what about `exists`, `publicUrl`, `signedUrl` etc...?

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -12,14 +12,31 @@ import {
...
 
 const FUNCTION_NAME_OPTS: NameOptions = {
   maxLen: 32,
   disallowedRegex: /[^a-z0-9]+/g,
   case: CaseConventions.LOWERCASE,
 };
 
+export enum ResourceTypes {
+  BUCKET = "Bucket",
+  Function = "Function",
```
tsuf239, line 25: can you please make them all upper case?

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -12,14 +12,31 @@ import {
...
+}
+
+export enum ActionTypes {
+  StorageRead = "roles/storage.objectViewer",
+  StorageReadWrite = "roles/storage.objectUser",
+  FunctionInvoker = "roles/cloudfunctions.invoker",
+  FunctionViewer = "roles/cloudfunctions.viewer",
+}
```
tsuf239, line 33: same here- all of the SDK enums are uppercase

libs/wingsdk/src/target-tf-gcp/function.ts
tsuf239: @Chriscbr?

libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts
```diff
@@ -0,0 +1,320 @@
...
+  const BUCKET_NAME = "test-bucket";
+  const KEY = "test-file-1";
+  const VALUE = "test-content-1";
+
+  const storage = new MockStorage();
+
+  const client = new BucketClient(BUCKET_NAME, false, storage as any);
+  const res = await client.put(KEY, VALUE);
+
+  expect(res).toBe(undefined);
```
tsuf239, line 24: can you test if the right gcp function is called?

libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts
```diff
@@ -0,0 +1,320 @@
...
+  const BUCKET_NAME = "test-bucket";
+  const KEY = "test-file-1";
+  const VALUE = "test-content-1";
+
+  const storage = new MockStorage();
+  await storage.bucket(BUCKET_NAME).put(KEY, VALUE);
+
+  const client = new BucketClient(BUCKET_NAME, false, storage as any);
+  const res = await client.delete(KEY);
+
```
tsuf239, line 105: can you test if the gcp function is called/ the file is deleted?

libs/wingsdk/test/target-awscdk/__snapshots__/bucket.test.ts.snap
```diff
@@ -305,6 +305,7 @@ def submit_response(event: dict, context, response_status: str, error_message: s
           ],
         },
         "Runtime": "nodejs18.x",
+        "Timeout": 900,
```
tsuf239, line 308: where are those coming from?

tsuf239 (commented):
Great work! Did you test those?
I'll try and test it on my account soon...

libs/wingsdk/src/target-tf-gcp/function.ts
Chriscbr: @subh-cs can you share more details about what you are trying to implement? It looks like some of the code is duplicated with the logic in the `addPermission` method

libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts
subh-cs: How about doing it this way? 

For the test `"put object to bucket"`
```js
  const client = new BucketClient(BUCKET_NAME, storage as any);
  await client.put(KEY, VALUE);

  // check that the object was put by ensuring it exists
  const res = await client.exists(KEY);
  expect(res).toBe(true);
  ```
  
  For the test `"delete object from bucket"`
  ```js
  const client = new BucketClient(BUCKET_NAME, storage as any);
  await client.delete(KEY);

  // check that the object was deleted by ensuring it doesn't exist
  const res = await client.exists(KEY);

  expect(res).toBe(false);
  ```

File: libs/wingsdk/src/shared-gcp/bucket.inflight.ts (hunk 7 above)
Comment (subh-cs) [reply libs/wingsdk/src/shared-gcp/bucket.inflight.ts:22]: Yes there is. `this.bucket.makePrivate()` and `this.bucket.makePublic()` works well. But I don't think `IBucketClient` interface support any of related to this right now for making a bucket public or private. If it is something we need, I can definitely implement that. 

Right now I'm using this `isPublic` props just for unit testing.

File: libs/wingsdk/test/target-awscdk/__snapshots__/bucket.test.ts.snap (hunk 14 above)
Comment (subh-cs) [reply libs/wingsdk/test/target-awscdk/__snapshots__/bucket.test.ts.snap:308]: I'm not doing anything, trust me 😂
I just ran `pnpm test` inside wingsdk and getting this snapshot.

File: libs/wingsdk/src/shared-gcp/bucket.inflight.ts (hunk 7 above)
Comment (tsuf239) [reply libs/wingsdk/src/shared-gcp/bucket.inflight.ts:22]: I see, 
I guess it's implemented in `target-tf-gcp/bucket.ts`, 
is there a way to query that on the inflight client?

File: libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts (hunk 12 above)
Comment (tsuf239) [reply libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts]: yes :)

File: libs/wingsdk/test/target-awscdk/__snapshots__/bucket.test.ts.snap (hunk 14 above)
Comment (tsuf239) [reply libs/wingsdk/test/target-awscdk/__snapshots__/bucket.test.ts.snap:308]: @MarkMcCulloh do you have any idea?

File: libs/wingsdk/.projenrc.ts [hunk 15]
```diff
@@ -363,7 +366,6 @@ project.gitignore.addPatterns("src/.gen");
 
 project.preCompileTask.exec("cdktf get --force");
 
-project.package.file.addDeletionOverride("pnpm");
```
MarkMcCulloh, line 366: undo, the pnpm section of package.json isn't needed now

libs/wingsdk/test/target-awscdk/__snapshots__/bucket.test.ts.snap
MarkMcCulloh: My guess is that the lockfile was updated and the newer version of aws-cdk-lib added this default.

libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts
```diff
@@ -0,0 +1,341 @@
...
+  const storage = new MockStorage();
+
+  const client = new BucketClient(BUCKET_NAME, storage as any);
+  const res = await client.tryDelete(NON_EXISTENT_KEY);
+
+  expect(res).toBe(false);
+});
+
+// test("Given a bucket, when giving one of its keys, we should get its signed url", async () => {
```
tsuf239, line 314: can you please add the issue link here as well? (as well as in the permissions comment)

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/t.main.w_compile_tf-aws.md
```diff
@@ -0,0 +1,703 @@
+# [t.main.w](../../../../../../examples/tests/sdk_tests/bucket/t.main.w) | compile | tf-aws
+
```
tsuf239, line 2: this should be deleted, I think

tsuf239 (approved):
Great work! both @subh-cs and I were testing the methods on gcp, and they're working! 👏 👏 👏

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,257 @@
...
+
+export class BucketClient implements IBucketClient {
+  private bucketName: string;
+  private storage: Storage;
+  private bucket: Bucket;
+
+  constructor(
+    bucketName: string,
+    storage: Storage,
+    projectId: string = process.env.GCP_PROJECT_ID || "wingsdk-test"
```
Chriscbr, line 18:
```suggestion
    projectId: string = process.env.GCP_PROJECT_ID ?? "wingsdk-test"
```

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,257 @@
...
+  ) {
+    this.bucketName = bucketName;
+    this.storage = storage ? storage : new Storage({ projectId: projectId });
```
Chriscbr, line 21: How come there is a default initialization? Based on the constructor's type annotation (`storage: Storage`) it looks like it should always be defined

Chriscbr, line 21:
```suggestion
    this.storage = storage ? storage : new Storage({ projectId });
```

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,257 @@
...
+    this.bucket = this.storage.bucket(this.bucketName);
+  }
+
+  public async metadata(_key: string): Promise<ObjectMetadata> {
+    throw new Error("Method not implemented.");
+  }
+
+  // check if bucket is public or not from bucket metadata and set it to _public
+  public async isPublic(): Promise<boolean> {
```
Chriscbr, line 30:
```suggestion
  // check if bucket is public or not from bucket metadata and set it to _public
  private async isPublic(): Promise<boolean> {
```

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -0,0 +1,257 @@
...
+        `Failed to check if bucket is public. (bucket=${this.bucketName})`
+      );
+    }
+  }
+
+  /**
+   * Check if an object exists in the bucket
+   *
+   * @param key Key of the object
+   */
```
Chriscbr, line 45: @subh-cs I'm wondering if we can safely omit these docstrings for inflight clients (would this cause any build errors?)

If they are copied from `cloud/bucket.ts` then there will be a lot of duplication, and information will get out of date... (cc @tsuf239)
```suggestion
```

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -12,14 +12,31 @@ import {
...
+  BUCKET = "Bucket",
+  FUNCTION = "Function",
+}
+
+export enum ActionTypes {
+  STORAGE_READ = "roles/storage.objectViewer",
+  STORAGE_READ_WRITE = "roles/storage.objectUser",
+  FUNCTION_INVOKER = "roles/cloudfunctions.invoker",
+  FUNCTION_VIEWER = "roles/cloudfunctions.viewer",
+}
```
Chriscbr, line 33: Can we link to any docs for these?

libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts
Chriscbr, line 314: nit: If possible can we avoid including unused code? To save code for later I recommend saving it in a separate branch

libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts
Chriscbr, line 1: Clean tests! ✨

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
subh-cs: So should I only do this?
```js
    this.storage = storage;
```

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
subh-cs: I don't think these will cause any build error, and yes they are copied from `cloud/bucket.ts`. I'll removing them.

libs/wingsdk/src/target-tf-gcp/function.ts
subh-cs: I'm not sure about this as it is only GCP specific. Would love to know your thought on this

tsuf239: I guess Chris meant to put the link to the gcp docs related to the permission type  as a comment next to each

Issue #4290, opened 2023-09-26
unable to call a method inside inflight class constructor


eladcon:
### I tried this:

run this code:

```ts
inflight class T {
  extern "./util.js" pub static inflight ext(): str;

  pub inflight s: str;

  init() {
    this.s = T.ext();
  }
}

test "aa" {
  let t = new T();
  log("${t.s}");
}
```

### This happened:

```
eladcohen@Elads-w-MacBook-Pro tmp % wing test main.w
fail ┌ main.wsim » root/env0/test:aa
     │ Error: Failed to bundle function: Error: Build failed with 1 error:
     │ target/test/main.wsim/.wing/inflight.T-1.js:7:16: ERROR: "await" can only be used inside an "async" function
```

### I expected this:

test should run

### Is there a workaround?

make the field optional and mutable and move the call to another function

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

staycoolcall911:
This is not related just to `extern`: a regular inflight method also fails similarly:
```ts
inflight class T {
  static inflight ext(): str {
    return "s";
  }

  pub inflight s: str;

  init() {
    this.s = T.ext();
  }
}

test "aa" {
  let t = new T();
  log("${t.s}");
}
```

staycoolcall911:
Also fails when the method is not `static`.

yoav-steinberg:
This is an approach which was suggested by the team:
output inflight class js:
```js
      class C {
        async get_six() {
          return 6;
        }
        constructor() {
        }
        async $infilght_init() {
          await super.$infilght_init();
          this.my_num = (await this.get_six());
        }
      }
      const c = new C();
      c.$infilght_init();
      { ((cond) => { if (!cond) throw new Error("assertion failed: c.my_num == 6") })((((a, b) => { try { return require('assert').deepStrictEqual(a, b) === undefined; } catch { return false; } })(c.my_num, 6))) };
    }
```

monadabot:
Congrats! :rocket: This was released in [Wing 0.35.10](https://github.com/winglang/wing/releases/tag/v0.35.10).

PR #4390, opened 2023-10-03
fix(compiler): can't call methods in inflight class's `init`

yoav-steinberg:
see #4290
Compiler now treats `init`'s of inflight classes as an async function. This makes it possible to call other async functions (specifically the class's methods) in the init function like you would expect.

The PR also includes some minor naming and comment cleanups.

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify/snapshots/access_methods_and_properties_on_collections.snap
```diff
@@ -29,6 +29,8 @@ module.exports = function({ $_y_at_0__, $x_length }) {
       {((cond) => {if (!cond) throw new Error("assertion failed: x.length == 3")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })($x_length,3)))};
       {((cond) => {if (!cond) throw new Error("assertion failed: y.at(0) == \"hello\"")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })($_y_at_0__,"hello")))};
     }
+    async $inflight_init() {
+    }
```
yoav-steinberg, line 33: I'm considering removing all these empty `inflight_init`s, but the nicest way of doing so will be to correctly reflect they don't exist in the AST because they really don't exist.
I'll try to add that to the PR..

yoav-steinberg: Eventually added support for removing empty `inflight_init`s without modifying the AST because that just made the code too complex. Instead I just avoid executing it if it's not defined in the emitted JS object.

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_outside_inflight_closure.test.w_compile_tf-aws.md
```diff
@@ -10,7 +10,7 @@ module.exports = function({ $BinaryOperation }) {
       return $obj;
     }
     async handle() {
-      const op = new $BinaryOperation(10,20);
+      const op = (await (async () => {const o = new $BinaryOperation(); if ('$inflight_init' in o) { await o.$inflight_init(10,20); } return o; })());
```
Chriscbr, line 13: It might be possible to dress this up a little nicer by defining a helper method that is responsible for both `new`-ing the class and then performing the inflight initialization logic:

```suggestion
      const op = (await $BinaryOperation.$init(10, 20));
```

(basically, I think this means changing `$inflight_init` to be a static method. What do you think?)

examples/tests/valid/inflight_init.test.w
```diff
@@ -0,0 +1,34 @@
...
+    this.field1 = this.get_six();
+    this.field2 = f2;
+  }
+}
+
+inflight class FooChild extends Foo {
+  pub field3: num;
+
+  init() {
+    super(5);
```
eladb, line 19: Impressive!

libs/wingc/src/jsify/snapshots/capture_type_inflight_class_sibling_from_init.snap
```diff
@@ -39,11 +39,11 @@ module.exports = function({  }) {
       class Bar {
         async func() {
           x;
-          new Foo();
+          (await (async () => {const o = new Foo(); if ('$inflight_init' in o) { await o.$inflight_init(); } return o; })());
```
eladb, line 42:
```suggestion
           await new.Foo()?.$inflight_init();
```

Or something along those lines.

tools/hangar/__snapshots__/test_corpus/valid/class.test.w_compile_tf-aws.md
```diff
@@ -95,8 +95,8 @@ module.exports = function({  }) {
 ```js
 module.exports = function({ $A }) {
   class B extends $A {
-    constructor(sound) {
-      super(sound);
+    async $inflight_init(sound) {
+      await super.$inflight_init(sound);
```
eladb, line 99: What happens if super doesn't have an inflight_init function?

```suggestion
      await super?.$inflight_init?(sound);
```

(I really don't remember the syntax🤦‍♀️, but basically optional chaining to the max)

eladb (approved):
Pretty brave... interested to see how this works in the wild. @3p3r is going to be thrilled (although he wanted async constructors in preflight).

skyrpex:
From the issue description:

> Compiler now treats init's of inflight classes as an sync function.

I think it's meant to say `as an async function`, right?

tools/hangar/__snapshots__/test_corpus/valid/class.test.w_compile_tf-aws.md
yoav-steinberg: > What happens if super doesn't have an inflight_init function?

Pretty useless calling `super()` if there's no init in the parent, but still a bug. Fixed, nice catch.

eladb: What happens if my inflight wing class extends a normal JavaScript class? How would constructor chaining super calls work?

For example, I need to propagate arguments to the normal super ctor?

libs/wingc/src/jsify/snapshots/capture_type_inflight_class_sibling_from_init.snap
yoav-steinberg: This won't work :( because we need to return the new object and the optional chaining here might return `undefined`.

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_outside_inflight_closure.test.w_compile_tf-aws.md
yoav-steinberg: Implementing this is tricky because I'll need to detect references to `this` in the inflight init and special case them.
Having some internal stdlib function for `std.$instantial_inflight_type(Type, args)` that calls `new`, optionally calls `$inflight_init` and finally returns the object seems ideal to me. But this is yet another case of needing access to stdlib from inflight which we currently don't have..

libs/wingc/src/jsify/snapshots/capture_type_inflight_class_sibling_from_init.snap
yoav-steinberg: Resolving this, see open [conversation above](https://github.com/winglang/wing/pull/4390#discussion_r1348000429)

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_outside_inflight_closure.test.w_compile_tf-aws.md
yoav-steinberg: I cleaned this up a bit using `?.` optional chaining. I think it's good enough for now given the challenges above. WDYT?

tools/hangar/__snapshots__/test_corpus/valid/class.test.w_compile_tf-aws.md
eladb: I am not sure this is a very common scenario and maybe we decide that we don't allow it but consider this:

```js
bring "some-external-jsii-module" as e1;

inflight class MyClass extends e1.MyBase {
  init(foo: str) {
    super(foo);
    // bla
  }
}
```

In the above example, MyBase is a normal JavaScript class without an async $inflight_init.

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_outside_inflight_closure.test.w_compile_tf-aws.md
Chriscbr: > Implementing this is tricky because I'll need to detect references to this in the inflight init and special case them.

Ah, I see what you mean - if a class has code like this:

```js
inflight class A {
  init() {
    this.x = this.calculate();
  }
  calculate() {
    return 42;
  }
}
```

Then the JS code would have to be something like:

```js
class A {
  constructor() {}
  static $init(..args) {
    const $this = new A(..args);
    $this.x = $this.calculate();
  }
}
```

(`$this` could be also named `o`, or `self`, or whatever)

Doable but it's not trivial. I also agree `std.$createInflight(Type, args)` would also be a good solution to this. Yeah I'm cool with revisiting this once we have easy access to stdlib inflight 👍

tools/hangar/__snapshots__/test_corpus/valid/class.test.w_compile_tf-aws.md
Chriscbr: JSII libraries don't have inflight classes (for now?), but they can have phase-independent (i.e. non-construct) classes, and it should be possible to define an inflight class as a subclass of a phase-independent class once we have more phase-independent language capabilities, so the example sort of makes sense...

Chriscbr (approved):

tools/hangar/__snapshots__/test_corpus/valid/class.test.w_compile_tf-aws.md
yoav-steinberg: I'll look into supporting this. If it become too complex I'll open a separate issue since I think at this point this is a fringe case, and getting inflight inits to work in all other cases is higher priority.

yoav-steinberg: Look what error I got while writing a test for this:
image
So I'm resolving this since it's currently not supported anyway. I'll open an issue and possibly complete the job with another PR.

3p3r:
@ me when preflights are async! glad to see the progress.

PR #4393, opened 2023-10-03
feat(sdk): gcp implementation of ex.table - preflight

wzslr321:
Partially solves #1316 

Video showing deployment to GCP: https://streamable.com/dty9yd

It basically works, but I'd appreciate some guidance on how to make it production-ready.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-tf-gcp/table.ts
```diff
@@ -0,0 +1,102 @@
...
+      instanceName: instance.name,
+      columnFamily: columnsFamily,
+      project: app.projectId,
+    };
+
+    new BigtableTable(this, "Default", tableConfig);
+  }
+
+  public addRow(_key: string, _row: Json): void {
+    throw new Error("Method is not supported as preflight for GCP target.");
```
wzslr321, line 90: Google Cloud does not support adding a row without a client, therefore it is not possible with terraform at this stage. I believe it is not possible to implement this as a preflight method - only as an inflight.  Can someone tell me what I should do in this case?

libs/wingsdk/src/target-tf-gcp/table.ts
```diff
@@ -0,0 +1,102 @@
...
+      columnsFamily.push({ family: key });
+    }
+
+    let clusterId = props.clusterId;
+    if (!props.clusterId) {
+      clusterId = "default";
+    }
+
+    const autoscalingConfig: BigtableInstanceClusterAutoscalingConfig = {
+      minNodes: props.minNodes ?? 1,
```
wzslr321, line 60: I believe it is not the best way to specify props, since these props are needed only in GCP, but are available on more 'global' ex.Table. I am not sure how I should handle it.

libs/wingsdk/src/target-tf-gcp/table.ts
Chriscbr: I think throwing an error here is ok. 👍  In the future, if they add support in Terraform we can update the implementation. I think occasional gaps like this is normal since the cloud is still new.

Once we have support for the inflight methods of `Table`, we can work around this by making `addRow` use [OnDeploy](https://www.winglang.io/docs/standard-library/cloud/on-deploy) to create a table row at deployment time.

libs/wingsdk/src/target-tf-gcp/table.ts
Chriscbr: Good question! Yeah - it's not uncommon for clouds to have extra parameters on some of their resources. Options like `minNodes` and `maxNodes` are typically considered "non-functional" aspects of the resource API because they represent operational concerns that are usually independent from the business logic you are performing with the resource.

For the Wing SDK, the best practice right now is to set a reasonable default (usually a good starting point is whatever defaults the cloud provider recommends when you're configuring the resource through their web console). Then, if anyone needs different values, they can be overridden using a [compiler plugin](https://www.winglang.io/docs/tools/compiler-plugins). In the future, there might be additional ways to customize these options within the Wing language by casting `cloud.Table` into `gcp.Table` or other mechanics.

libs/wingsdk/src/target-tf-gcp/app.ts
```diff
@@ -59,6 +70,8 @@ export class App extends CdktfApp {
       );
     }
 
+    this.zone = props.zone ?? process.env.GOOGLE_ZONE;
```
wzslr321, line 73: Technically it is necessary to deploy `Table', but if it is not needed for other resources, so I do not raise an error here if it is undefined. Or maybe I should?

libs/wingsdk/src/target-tf-gcp/table.ts
```diff
@@ -0,0 +1,99 @@
...
+      columnFamily: columnsFamily,
+      project: app.projectId,
+    };
+
+    new BigtableTable(this, "Default", tableConfig);
+  }
+
+  public addRow(_key: string, _row: Json): void {
+    throw new Error(
+      "Method is not supported as a preflight for the GCP target."
```
Chriscbr, line 86: Since adding rows in preflight isn't supported, we should give the user a similar error message if they try to configure `initialRows`. We can add something like this to the constructor:

```js
if (props.initialRows) {
  throw new Error("prop initialRows is not supported for the GCP target.");
}
```

Chriscbr (changes requested):
This is almost ready to ship! Just a few small changes

libs/wingsdk/src/target-tf-gcp/app.ts
wzslr321: Yes, we can - The zones are part of the region, so there is always at least one - 'a'.  That is a good suggestion, thank you - I committed it :)

libs/wingsdk/src/target-tf-gcp/table.ts
wzslr321: Great suggestion - the user should know why the properties he specified have no effect. Thank you - I pushed a commit with it.

eladb (commented):
Exciting!

Chriscbr (approved):
Nicely done @wzslr321!

PR #4416, opened 2023-10-04
feat(compiler): error diagnostics v2

Chriscbr:
The Wing compiler now has the capability to provide more detailed information with each compiler error. I've implemented a few examples in this pull request to demonstrate the feature:

Example 1: showing the original symbol definition in "use before defined" errors:
Screenshot 2023-10-04 at 6 34 41 PM

Example 2: showing the previous definition in "symbol already defined" errors:
Screenshot 2023-10-04 at 6 36 01 PM

Example 3: showing the variable definition in "variable is not reassignable" errors:
Screenshot 2023-10-04 at 6 31 04 PM

These errors are also displayed as related information in the LSP with clickable links:
Screenshot 2023-10-04 at 6 21 01 PM

I haven't tested what happens when the annotations refer to other files, but in theory it should also work...

## Checklist

- [x] Tests added (always)


Chriscbr:
I tried out an alternative error diagnostic formatting library named `annotate-snippets` instead of the current one we're using, `codespan-reporting`. Here's a comparison of how some errors look:

---

codespan-reporting:
Screenshot 2023-10-04 at 6 56 51 PM

annotate-snippets:
Screenshot 2023-10-04 at 6 57 44 PM

---

codespan-reporting:
Screenshot 2023-10-04 at 7 01 23 PM

annotate-snippets:
Screenshot 2023-10-04 at 7 00 54 PM

----

I wasn't sure if it was much of an improvement so I left it out of the PR

libs/wingc/src/type_check.rs
```diff
@@ -3879,7 +3888,14 @@ impl<'a> TypeChecker<'a> {
 		let (var, var_phase) = self.resolve_reference(&variable, env);
 
 		if !var.type_.is_unresolved() && !var.reassignable {
-			self.spanned_error(variable, "Variable is not reassignable".to_string());
+			report_diagnostic(Diagnostic {
+				message: "Variable is not reassignable".to_string(),
+				span: Some(variable.span()),
+				annotations: vec![DiagnosticAnnotation {
+					message: "Variable is defined here (try adding \"var\" in front)".to_string(),
```
MarkMcCulloh, line 3895: I feel like annotations should be read highly contextually and not even require complete sentences. Maybe even no need to start with a capital letter:

```suggestion
					message: "defined here (try adding \"var\" in front)".to_string(),
```

WDYT?

libs/wingc/src/type_check.rs
```diff
@@ -5407,23 +5424,32 @@ fn lookup_result_to_type_error<T>(lookup_result: LookupResult, looked_up_object:
...
+		LookupResult::MultipleFound => TypeError {
+			message: format!("Ambiguous symbol \"{looked_up_object}\""),
+			span: looked_up_object.span(),
+			annotations: vec![],
+		},
+		LookupResult::DefinedLater(span) => TypeError {
+			message: format!("Symbol \"{looked_up_object}\" used before being defined"),
+			span: looked_up_object.span(),
+			annotations: vec![DiagnosticAnnotation {
+				message: format!("\"{}\" defined later here", looked_up_object),
```
MarkMcCulloh, line 5442:
```suggestion
				message: "defined later here".to_string(),
```

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -214,10 +214,16 @@ impl SymbolEnv {
 			return Err(TypeError {
 				span: symbol.span.clone(),
 				message: format!("Symbol \"{}\" already defined in this scope", symbol.name),
+				annotations: vec![DiagnosticAnnotation {
+					message: format!("previous definition of \"{}\"", symbol.name),
```
MarkMcCulloh, line 218:
```suggestion
					message: "previous definition".to_string(),
```

pnpm-lock.yaml
```diff
@@ -12938,7 +12938,7 @@ packages:
     dependencies:
       semver: 7.5.4
       shelljs: 0.8.5
-      typescript: 5.3.0-dev.20231003
+      typescript: 5.3.0-dev.20231004
```
MarkMcCulloh, line 12941: 🤔

libs/wingc/src/diagnostic.rs
```diff
@@ -187,9 +187,16 @@ impl PartialOrd for WingSpan {
...
 	pub message: String,
+	pub annotations: Vec<DiagnosticAnnotation>,
 	pub span: Option<WingSpan>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
+pub struct DiagnosticAnnotation {
+	pub message: String,
+	pub span: WingSpan,
+}
```
MarkMcCulloh, line 198: Should we have include the concept of hints in this model? Obviously you can put it in the message, but what if it were a separate field or something here. Might encourage others to add hints, and maybe they can be formatted differently.

Also, a possible future for this: Would be interesting to integrate hints with LSP code actions to fix errors. That way these fixes can be in the compiler itself and not rewritten for the language server.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2154,6 +2162,9 @@ error: Cannot call an optional function
 error: Variable is not reassignable
    --> ../../../examples/tests/invalid/optionals.test.w:53:3
    |
+50 | if let hi = hi {
+   |        -- Variable is defined here (try adding \\"var\\" in front)
+   .
```
MarkMcCulloh, line 2167: Worth double checking if these errors look fine in the console.

MarkMcCulloh (approved):
Nice. Had a thought on style but LGTM

pnpm-lock.yaml
Chriscbr: Yanked

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: It was not automatically updated with the new annotations. Working on a fix

libs/wingc/src/diagnostic.rs
Chriscbr: Sounds like a good future improvement

eladb:
Very cool!

PR #4427, opened 2023-10-05
feat(vscode): add underlined hints for error annotations

MarkMcCulloh:
Stealing rust's idea:

image

### DIFF
--- a/apps/wing/src/commands/lsp.ts
+++ b/apps/wing/src/commands/lsp.ts
@@ -144,11 +144,28 @@ export async function lsp() {
           }))
         );
 
+        // Add annotations as notes hinting back to the original diagnostic
+        const extraNotes = rd.annotations.map((a) =>
+          Diagnostic.create(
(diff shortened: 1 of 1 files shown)


apps/wing/src/commands/lsp.ts
```diff
@@ -144,11 +144,28 @@ export async function lsp() {
...
+          Diagnostic.create(
+            Range.create(a.span.start.line, a.span.start.col, a.span.end.line, a.span.end.col),
+            a.message,
+            DiagnosticSeverity.Hint,
+            undefined,
+            undefined,
+            [
+              {
+                location: Location.create(diagnosticUri, diag.range),
+                message: `(Source) ${diag.message}`,
```
Chriscbr, line 158: Feels like this doesn't have to be caps but I'm not strongly opinionated on it
```suggestion
                message: `(source) ${diag.message}`,
```

Chriscbr (approved):

PR #4434, opened 2023-10-05
feat(sdk): add `Function.invokeAsync()` for `sim`, `tf-aws` and `awscdk` targets

### LINKED ISSUE #1387: cloud.Function.invokeAsync()
https://www.winglang.io/contributing/rfcs/2023-01-20-wingsdk-spec#function

The signature of this method (in Wing) is:

```js
/**
 * Invoke the function asynchronously, meaning it will not wait for the function to finish its execution.
 * It returns as soon as the invoke request has been received.
 */
inflight invokeAsync(payload: str): void;
```

Example:

```js
bring cloud;
bring util;

let counter = new cloud.Counter();

  util.sleep(3s);
  counter.inc();
});

test "invokeAsync" {
  assert(counter.peek() == 0);
  fn.invokeAsync("");
  assert(counter.peek() == 0);
  util.waitUtil(inflight () => { return counter.peek() == 1; }, timeout: 10s);
  assert(counter.peek() == 1);
}
```

garysassano:
Closes #1387

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/cloud/function.ts
```diff
@@ -144,10 +147,16 @@ export abstract class Function extends Resource implements IInflightHost {
  */
 export interface IFunctionClient {
   /**
-   * Invoke the function asynchronously with a given payload.
+   * Invokes the function synchronously with a payload and waits for the result.
```
eladb, line 150:
```suggestion
   * Invokes the function with a payload and waits for the result.
```

libs/wingsdk/src/cloud/function.ts
```diff
@@ -144,10 +147,16 @@ export abstract class Function extends Resource implements IInflightHost {
...
    * @inflight
    */
   invoke(payload: string): Promise<string>;
+
+  /**
+   * Invokes the function asynchronously with a payload and returns immediately.
```
eladb, line 156:
```suggestion
   * Kicks off the execution of the function with a payload and returns immediately while the function is running. 
```

eladb (changes requested):
Missing sdk test and documentation update

Chriscbr:
@garysassano is this still something you're looking to work on? I wanted to check and see if there's anything I can do to help get this PR over the finish line

garysassano:
@Chriscbr I'm not sure if the `sim` implementation functions correctly, but from what I recall, the `tf-aws` one worked well when I tested it.

Chriscbr (changes requested):
The AWS implementation seems right to me. Looks like we need a test in the `sdk_tests` folder.

Let's also add an example of `invokeAsync` in the function's documentation [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/src/cloud/function.md)

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -41,9 +41,27 @@ export class FunctionClient implements IFunctionClient {
...
+      Payload: fromUtf8(JSON.stringify(payload)),
+      InvocationType: "Event",
+    });
+    const response = await this.lambdaClient.send(command);
+
+    if (response.StatusCode !== 202) {
+      throw new Error(
+        `Failed to enqueue event. Received status code: ${response.StatusCode}`
+      );
+    }
```
Chriscbr, line 59: Maybe we can log any extra error information available in case it fails. I found these fields listed in the docs [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-lambda/Interface/InvokeCommandOutput/)
```suggestion
    if (response.StatusCode !== 202) {
      console.error("Error: " + response.FunctionError);
      console.error(fromUtf8(response.Payload));
      throw new Error(
        `Failed to enqueue event. Received status code: ${response.StatusCode}`
      );
    }
```

Chriscbr (approved):

Chriscbr:
@garysassano I pushed a small update to your branch to update the docs

examples/tests/sdk_tests/function/invoke_async.test.w
```diff
@@ -0,0 +1,19 @@
...
+  log("log inside f1");
+}) as "f1";
+
+let f2 = new cloud.Function(inflight (input: str): str => {
+  log("log inside f2");
+}) as "f2";
+
+test "invoke" {
+  log("log inside test");
+  f1.invokeAsync(payload);
```
Chriscbr, line 17: Looks like there's an issue with test flakeyness (see https://github.com/winglang/wing/pull/4434/commits/ad97a8a2655cae868735fbaf43ab2b55dc0a09e4). I believe the issue is f1's code might not get run until after the test finished, but sometimes it will. If we want to guarantee "log inside f1" is printed in the test output, we should add something like `util.sleep(1s)` to the end of the test

garysassano:
Keep.

revitalbarletz:
Hey @garysassano, do you need any assistance here?

revitalbarletz:
@Chriscbr and @garysassano can you update this PR with your status?

Chriscbr:
@revitalbarletz I started working to update the PR on Friday - still want to give one more look over the tests before we merge this

libs/wingsdk/src/shared-azure/function.inflight.ts
```diff
@@ -28,6 +28,16 @@ export class FunctionClient implements IFunctionClient {
...
   }
 
+  /**
+   * Invokes the function asynchronously, passing the given payload as an argument.
+   * @returns immediately once the event has been handed off to AWS Lambda.
+   */
+  public async invokeAsync(payload: string): Promise<void> {
+    return Promise.reject(
+      `invokeAsync is not implemented: (payload=${payload})`
+    );
```
eladb, line 38:
```suggestion
    throw new Error("invokeAsync not implemented");
```

I think this is good enough, no?

eladb (approved):

libs/wingsdk/src/shared-azure/function.inflight.ts
garysassano: You must use the parameter or you won't be able to compile because of the linting warning.

Chriscbr: I guess we can try this
```suggestion
    payload;
    throw new Error("invokeAsync is not implemented");
```

PR #4435, opened 2023-10-06
fix(docs): fix `Function.invoke()` docs

garysassano:
`Function.invokeAsync()` still hasn't been implemented: #1387

### DIFF
--- a/libs/wingsdk/src/cloud/function.ts
+++ b/libs/wingsdk/src/cloud/function.ts
@@ -144,7 +144,7 @@ export abstract class Function extends Resource implements IInflightHost {
  */
 export interface IFunctionClient {
   /**
-   * Invoke the function asynchronously with a given payload.
+   * Invokes the function synchronously with a payload and waits for the result.
    * @inflight
    */
   invoke(payload: string): Promise<string>;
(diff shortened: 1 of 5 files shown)


libs/wingsdk/src/cloud/function.ts
```diff
@@ -144,7 +144,7 @@ export abstract class Function extends Resource implements IInflightHost {
  */
 export interface IFunctionClient {
   /**
-   * Invoke the function asynchronously with a given payload.
+   * Invokes the function synchronously with a payload and waits for the result.
```
MarkMcCulloh, line 147: I'm not sure it's worth mentioning async vs sync here. The function itself is not blocking, but that's implied by the return type.

```suggestion
   * Invokes the function with a payload and waits for the result.
```

MarkMcCulloh (approved):

libs/wingsdk/src/cloud/function.ts
MarkMcCulloh: the difference between this and `invokeAsync` is less about the async-ness of the function and more about the "waits for the result" part

PR #4436, opened 2023-10-06
feat(vscode): quickfix and autofix to insert semicolons

MarkMcCulloh:
https://github.com/winglang/wing/assets/1237390/67c8581d-f017-4caa-ac42-3e3b6dc5b688

Adds both a quick fix and an autofix for semicolon insertion.
It will only insert a semicolon if there is no other error present at the same position.

Misc:
- Refactored the contains* functions for WingSpan to avoid unneccesary .into()s


eladb:
@MarkMcCulloh what tool are you using for the screen recordings that has this cool zoom feature?

libs/wingc/src/lsp/code_actions.rs
```diff
@@ -0,0 +1,238 @@
...
+			..Default::default()
+		}));
+	}
+
+	action_list
+}
+
+fn get_fix_for_diagnostic(file: Url, diagnostic: Diagnostic) -> Option<CodeActionOrCommand> {
+	match diagnostic.message.as_str() {
+		"Expected ';'" => {
```
eladb, line 132: Seems fragile... maybe put into a const and reuse this const when the diagnostic is created.

eladb (approved):
Cool beans

MarkMcCulloh:
@eladb https://www.screen.studio/

libs/wingc/src/lsp/code_actions.rs
MarkMcCulloh: Can do. Ideally I'd like the quickfix generation to actually be part of generating the diagnostic so it's not separate, but figured that can wait

libs/wingc/src/lsp/code_actions.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        insta::with_settings!(
+          {
+            prepend_module_to_snapshot => false,
+            omit_expression => true,
+            snapshot_path => "./snapshots/code_actions",
+          }, {
+            insta::assert_yaml_snapshot!($name, {
+              "[].edit.changes" => insta::dynamic_redaction(|value, _path| {
+								// value is a hashmap with guid filename keys, change the keys to just be incrementing numbers
```
Chriscbr, line 193: Looks like some mixed formatting

Chriscbr (approved):

Issue #4442, opened 2023-10-06
`cloud.Domain` for `tf-aws` is missing the `iamCertificate` and `hostedZoneId` properties


skyrpex:
### I tried this:

The CI is trying to run the [aws-sdk-spec-test / domain](https://github.com/winglang/wing/actions/runs/6432089976/job/17467933233#logs) job.

### This happened:

Fails to run the test:

```
- Compiling domain/domain.main.w to tf-aws...

✖ Compiling domain/domain.main.w to tf-aws...

ERROR: 
  - 'iamCertificate' or 'acmCertificateArn' is missing from root/Test._qo49qv3Az/env0/my-domain
  - 'hostedZoneId' is missing from root/Test._qo49qv3Az/env0/my-domain

These are required properties of platform-specific types. You can set these values
either through '-v | --value' switches or '--values' file.

../../../../../tmp/-wing-compile-testp68oLu/target/test/domain.main.tfaws.979586.tmp/.wing/preflight.js:[42](https://github.com/winglang/wing/actions/runs/6432089976/job/17467933233#step:10:43)
         }
       }
>>     const d = this.node.root.newAbstract("@winglang/sdk.cloud.Domain",this,"my-domain",{ domainName: "example.com" });
       this.node.root.new("@winglang/sdk.std.Test",std.Test,this,"test:domain name",new $Closure1(this,"$Closure1"));
     }
```

### I expected this:

The CI test should work.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

MarkMcCulloh:
The original test was deleted in https://github.com/winglang/wing/pull/4447, but this issue is still relevant since we want a way to run tests with cloud.Domain

marciocadev:
The problem with running this test is that with each execution, the test changes the path, as shown in the example below:

> root/Test.gfvqV9U482/env0/my-domain
> root/Test.OcZ_aYTvRy/env0/my-domain

I can abstract part of the path when a test is performed by ignoring the values between `Test.` and the next `/`, and this would solve the snapshot generation.

I will try to resolve this, and I will open a new PR shortly.

garysassano:
I think #3749 may solve this issue.

tsuf239:
In the issue, it says that the sdk-spec-tests gh action doesn't have a way to add the --values while running, 
[...]

PR #4443, opened 2023-10-06
chore: skip aws-cdk-spec-test for `cloud.Domain`

skyrpex:
Related to #4442.


Chriscbr (approved):

eladb (approved):

MarkMcCulloh:
I would just delete the file for now.
If not, I think the tf-aws-test.yml workflow needs to be modified because I think it'll have an issue with this .skip file.

PR #4445, opened 2023-10-06
fix: windows fails if any drives are inaccessible

MarkMcCulloh:
This makes the behavior match between windows and non-windows (checking the drive first before adding preopen)

### DIFF
--- a/libs/wingcompiler/src/wingc.ts
+++ b/libs/wingcompiler/src/wingc.ts
@@ -103,8 +103,14 @@ export async function load(options: WingCompilerLoadOptions) {
       .map((value) => value.trim());
 
     for (const drive of drives) {
-      // drive will be something like "C:"
-      preopens[`${drive}/`] = `${drive}\\`;
+      try {
+        let actualPath = `${drive}\\`;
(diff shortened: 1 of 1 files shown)


libs/wingcompiler/src/wingc.ts
```diff
@@ -103,8 +103,14 @@ export async function load(options: WingCompilerLoadOptions) {
...
-      // drive will be something like "C:"
-      preopens[`${drive}/`] = `${drive}\\`;
+      try {
+        let actualPath = `${drive}\\`;
+        await fs.promises.access(actualPath, fs.constants.R_OK | fs.constants.F_OK);
+        // drive will be something like "C:"
+        preopens[`${drive}/`] = actualPath;
+      } catch {
+        // can't access the drive, don't bother preopening it
+      }
```
Chriscbr, line 113: nit: when a function throwing starts change the control flow of the rest of the code I feel like it's sometimes clearer to refactor into a function

```suggestion
      if (await isDriveAccessible(drive)) {
        // drive will be something like "C:"
        preopens[`${drive}/`] = `${drive}\\`;
      }
```

Chriscbr (approved):

PR #4455, opened 2023-10-07
fix(compiler)!: unexpected heuristic for inference across scopes

### LINKED ISSUE #4438: Unable to infer type, but it's previously inferred
### I tried this:

Given `main.w`:

```ts
bring cloud;

struct GitHubTokens {
  access_token: str;
  expires_in: num;
  refresh_token: str;
  refresh_token_expires_in: num;
  token_type: str;
  scope: str;
}

let exchangeCodeForTokens = inflight (code: str) => {
  return GitHubTokens.fromJson("");
};

let handler = inflight (request: cloud.ApiRequest) => {
  let tokens = exchangeCodeForTokens("code");
};
```

Run `wing compile main.w`.

### This happened:

```
error: Unable to infer type
   --> main.w:17:7
   |
17 |   let tokens = exchangeCodeForTokens("code");
   |       ^^^^^^ Unable to infer type
```

### I expected this:

It should infer the type, since it's already know if you hover on `exchangeCodeForTokens`.

### Is there a workaround?

Explicitely define the return type as `GitHubTokens`.

### Component

Compiler

### Wing Version

0.35.5

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
Fixes #4438

Due to:
- breadth-first type checking
- simplistic inference
- no re-checking (good)

Currently, this works:

```
let f = (arg) => {
  return arg.get("a");
};

if true {
  let x = f({ a: true });
}
```

but this does not work:

```
let f = () => {
  return { a: true };
};

if true {
  let x = f().get("a");
//        ^^^ don't know what the return type is yet because the closure above hasn't be checked
}
```

Because when first implementing inference, the sub-scopes of a scope were checked in reverse order. I thought this heuristic was kinda neat and figured it had legs. Now I'm thinking that was wrong. 

With this PR, the scope will be checked in normal order. Now the second case will work and the first case above will **not** work. It will require explicit annotation on the argument:

```
let f = (arg: Json) => {
  return arg.get("a");
};

if true {
  let x = f({ a: true });
}
```

BREAKING CHANGE: Inference now prioritizes earlier scopes rather than later ones. This previously allowed closures called from other scopes to leave off argument type annotations. [This](https://github.com/winglang/wing/blob/fcfc3462d37a063d96d49b4105e27466ff7793c8/examples/tests/invalid/inference.test.w#L106) is an example of a case that previously worked.

### DIFF
--- a/examples/tests/invalid/inference.test.w
+++ b/examples/tests/invalid/inference.test.w
@@ -103,12 +103,12 @@ if true {
 //             ^^^^^^^^^^^^^^^ Expected str, got num
 }
 
+// See https://github.com/winglang/wing/issues/4464 for details
 let unknownArg = (arg) => {
 //                ^^^ Unable to infer type
   return arg.get("a");
 //           ^^^ Property not found
 };
-
 if true {
   unknownArg({ a: true });
 }
--- a/examples/tests/invalid/inference.test.w
+++ b/examples/tests/invalid/inference.test.w
@@ -96,3 +96,19 @@ let badFunc: inflight (str): void = inflight (arg1: num) => {};
 // ^ Inferred type Array<str> conflicts with already inferred type Array<num>
 };
 
+
+let returnsNumber = () => { return 2; };
+if true {
+  let x: str = returnsNumber();
+//             ^^^^^^^^^^^^^^^ Expected str, got num
+}
+
+let unknownArg = (arg) => {
+//                ^^^ Unable to infer type
+  return arg.get("a");
+//           ^^^ Property not found
+};
+
+if true {
+  unknownArg({ a: true });
+}
--- a/libs/wingc/src/type_check/inference_visitor.rs
+++ b/libs/wingc/src/type_check/inference_visitor.rs
@@ -187,3 +187,32 @@ impl<'a> crate::visit_types::VisitType<'_> for InferenceVisitor<'a> {
 		}
 	}
 }
+
+#[derive(Default)]
+/// Deeply visits a type and finds any inferences.
+pub struct InferenceCounterVisitor {
+	/// Whether or not we found an inference during the entire visit
+	pub found_inference: bool,
+
+	/// Whether or not we found an inference during the entire visit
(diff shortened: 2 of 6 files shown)


libs/wingc/src/type_check/inference_visitor.rs
```diff
@@ -187,3 +187,32 @@ impl<'a> crate::visit_types::VisitType<'_> for InferenceVisitor<'a> {
 		}
 	}
 }
+
+#[derive(Default)]
+/// Deeply visits a type and finds any inferences.
+pub struct InferenceCounterVisitor {
```
MarkMcCulloh, line 193: ℹ️ I created this during some other experiments while researching the issue. It's not functionally different than the previous visitor for the use-case of looking for inferences. It's just more efficient and has no `mut`s anywhere.

examples/tests/invalid/inference.test.w
```diff
@@ -96,3 +96,19 @@ let badFunc: inflight (str): void = inflight (arg1: num) => {};
...
+
+let unknownArg = (arg) => {
+//                ^^^ Unable to infer type
+  return arg.get("a");
+//           ^^^ Property not found
+};
+
+if true {
+  unknownArg({ a: true });
+}
```
Chriscbr, line 114: If I understand correctly, this is an only an invalid in the sense that our inference system isn't strong enough to analyze it, not because there's a logical error/bug in the code, right?

If that's the case I don't know if it should be flagged as an invalid case given we might want to support it in the future. (Or maybe we can leave it and link to the issue about implementing better inference?)

Chriscbr (approved):

examples/tests/invalid/inference.test.w
MarkMcCulloh: Yep that's right.

I think there's value in showing this case, since we also want to validate behavior that shouldn't work. This ensures it doesn't change in unexpected ways. I'll add an issue and link to it though.

PR #4458, opened 2023-10-07
chore(compiler): update rust and wasi toolchain

MarkMcCulloh:
This was a difficult update. A newer version of rust exposed/caused an issue with our UnsafeRefs in some cases, causing memory corruption. I tried to highlight semi-functional changes

~Looks like the wasm size has increased by almost 2x. Not ideal, but not sure why or what to do. Based on benchmarks it looks like the perf is unchanged at least.~ Ok it turns out the new version of rust also somehow broke cargo-wasi and it no longer ran wasm-opt anymore. So now this PR also removes cargo-wasi and instead runs wasm-opt directly from the binaryen toolchain.
On the bright side, the install for this is much faster. The only other useful special thing that cargo-wasi did was run https://github.com/rustwasm/walrus to demangle names. Since we always do release builds, I'm not sure how valuable that is.

Updates:
- wasi-sdk: 19 -> 20
- rust: 1.67.1 -> 1.73.0
- cargo-wasi: Removed. Now using the `wasm32-wasi` target and wasm-opt directly
- tree-sitter: 0.20.7 -> 0.20.8
- All crates were updated
- All crates were set to use similar pinning as `^`, where patches are automatically installed


libs/wingc/src/type_check.rs
```diff
@@ -4918,18 +4928,14 @@ impl<'a> TypeChecker<'a> {
...
-					unsafe {
-						// We don't use the return value but need to call replace so it'll drop the old value
-						_ = std::mem::replace(&mut *mut_ptr, new_ref);
-					}
-					return self.resolve_reference(reference, env);
+
+					// Store this reference for later when we can modify the final AST and replace the original reference with the new one
+					self.types.type_expressions.insert(object.id, new_ref.clone());
+
+					return self.resolve_reference(&new_ref, env);
```
MarkMcCulloh, line 4938: ℹ️ This is the main change required in this PR to update rust.

The mem swapping was causing the reference data to be moved, but only in release builds (not only wasm32-wasi).

This replacement now happens after type-checking is done via `type_expressions`

libs/wingc/src/comp_ctx.rs
```diff
@@ -105,8 +102,9 @@ pub fn set_custom_panic_hook() {
 
 		report_diagnostic(Diagnostic {
 			message: format!(
-				"Compiler bug ({}) during {}, please report at https://www.winglang.io/contributing/start-here/bugs",
-				pi,
+				"Compiler bug ('{}' at {}) during {}, please report at https://www.winglang.io/contributing/start-here/bugs",
+				pi.payload().downcast_ref::<&str>().unwrap_or(&""),
+				pi.location().unwrap_or(&std::panic::Location::caller()),
```
MarkMcCulloh, line 107: The new Display for panicinfo changed in a recent rust version. IMO it didn't look good here, so here's a new look.

libs/wingc/src/diagnostic.rs
```diff
@@ -285,10 +276,17 @@ pub fn report_diagnostic(diagnostic: Diagnostic) {
...
+			Ok(json) => {
+				let bytes = json.as_bytes();
+				unsafe {
+					send_diagnostic(bytes.as_ptr(), bytes.len() as u32);
+				}
+			}
+			Err(err) => {
+				eprintln!("Error serializing diagnostic: {}", err);
+				// avoiding a panic here because we don't want to crash the panic handler
+			}
```
MarkMcCulloh, line 289: ℹ️  When having issue with memory corruption, this section would panic due to the json being invalid. This shouldn't normally happen, but just in case.

libs/wingc/src/type_check.rs
```diff
@@ -5756,9 +5763,10 @@ mod tests {
 
 	#[test]
 	fn function_subtyping_parameter_contravariance() {
-		let void = UnsafeRef::<Type>(&Type::Void as *const Type);
-		let string = UnsafeRef::<Type>(&Type::String as *const Type);
-		let opt_string = UnsafeRef::<Type>(&Type::Optional(string) as *const Type);
+		let void = UnsafeRef::<Type>(&Type::Void);
+		let string = UnsafeRef::<Type>(&Type::String);
+		let opt_string_type = Type::Optional(string);
+		let opt_string = UnsafeRef::<Type>(&opt_string_type);
```
MarkMcCulloh, line 5769: ℹ️ This change also exemplefies the issue regarding memory issues: In release mode, the pointer given by `UnsafeRef::<Type>(&Type::Optional(string) as *const Type)` becomes invalid later. 

The `Type::Optional(string)` part needs to be extracted, presumably so the data lives at its expected place longer.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2282,11 +2282,11 @@ Duration <DURATION>"
 `;
 
 exports[`panic.test.w 1`] = `
-"error: Compiler bug (panicked at 'User invoked panic', libs/wingc/src/type_check.rs:LINE:COL) during type-checking, please report at https://www.winglang.io/contributing/start-here/bugs
+"error: Compiler bug ('User invoked panic' at libs/wingc/src/type_check.rs:LINE:COL) during type-checking, please report at https://www.winglang.io/contributing/start-here/bugs
```
Chriscbr, line 2285: 👍

libs/wingii/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@winglang/wingii",
   "private": true,
   "scripts": {
-    "compile": "cargo wasi build --release",
+    "compile": "cargo build --target wasm32-wasi --release",
```
Chriscbr, line 5: Is there a 32-bit WASM and 64-bit WASM? 😱

libs/wingc/src/lib.rs
```diff
@@ -186,7 +187,7 @@ pub unsafe extern "C" fn wingc_compile(ptr: u32, len: u32) -> u64 {
 	if results.is_err() {
 		WASM_RETURN_ERROR
 	} else {
-		string_to_combined_ptr("winged it!".to_string())
```
Chriscbr, line 189: I didn't realize we had this lol

libs/wingc/src/comp_ctx.rs
```diff
@@ -12,22 +12,17 @@ use crate::diagnostic::{report_diagnostic, Diagnostic, WingSpan};
 /// Try to make these end with 'ing' (building, parsing, etc.) so they'll fit in
 /// in the context of a diagnostic message.
 /// See `CompilationContext::set` for more information.
-#[derive(EnumString, Display, PartialEq, Clone, Copy)]
+#[derive(EnumString, Display, PartialEq, Clone, Copy, Debug, Default)]
 #[strum(serialize_all = "kebab-case")]
 pub enum CompilationPhase {
+	#[default]
```
Chriscbr, line 18: TIL

Chriscbr:
@MarkMcCulloh when I tried building the PR locally I got an error

```
error[E0463]: can't find crate for `core`
 |
 = note: the `wasm32-wasi` target may not be installed
 = help: consider downloading the target with `rustup target add wasm32-wasi`
```

After I ran `rustup target add wasm32-wasi` it was fixed. Should we add this to the contributing guide?

Chriscbr (approved):
LGTM

libs/wingii/package.json
MarkMcCulloh: Only 32bit is really supported for now. As with most of wasm, 64bit is on its way and will probably take a while.

MarkMcCulloh:
> After I ran rustup target add wasm32-wasi it was fixed. Should we add this to the contributing guide?

Good catch, so now I added wasm32-wasi to the targets of the rust toolchain file. This should handle installing it. We already mention rustup in the contrib docs so this should be sufficient.

Issue #1414, opened 2023-02-01
Add jsii fixture library to monorepo


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

For maintainers and contributors to test out the Wing compiler's capabilities to import JSII libraries, instead of using an external library as done in https://github.com/winglang/wing/pull/1397, we should add a fixture JSII library to the monorepo where random classes, methods, types, etc. can all be added and used for testing purposes.

### Use Cases

Simplify testing cycle (avoid the need to publish a separate library or find an existing CDK library demonstrating particular aspects of the wing type system)

### Component

Compiler

yoav-steinberg:
I'm on this as part of some other issue..

monadabot:
Congrats! :rocket: This was released in [Wing 0.41.6](https://github.com/winglang/wing/releases/tag/v0.41.6).

PR #4465, opened 2023-10-08
feat(compiler): type-map JSII date to datetime

exoego:
Closes https://github.com/winglang/wing/issues/3569

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/type_check/jsii_importer.rs
+++ b/libs/wingc/src/type_check/jsii_importer.rs
@@ -94,7 +94,7 @@ impl<'a> JsiiImporter<'a> {
 				PrimitiveType::Boolean => self.wing_types.bool(),
 				PrimitiveType::Any => self.wing_types.anything(),
 				PrimitiveType::Json => self.wing_types.json(),
-				PrimitiveType::Date => self.wing_types.anything(), // TODO: https://github.com/winglang/wing/issues/3569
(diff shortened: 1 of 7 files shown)


libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -94,7 +94,7 @@ impl<'a> JsiiImporter<'a> {
 				PrimitiveType::Boolean => self.wing_types.bool(),
 				PrimitiveType::Any => self.wing_types.anything(),
 				PrimitiveType::Json => self.wing_types.json(),
-				PrimitiveType::Date => self.wing_types.anything(), // TODO: https://github.com/winglang/wing/issues/3569
+				PrimitiveType::Date => self.wing_types.datetime(),
```
MarkMcCulloh, line 97: This may not work as-is. I'm not super familiar with the JSII Date type, but based on reading the docs I'm assuming it'll either be a string or the actual js `Date` type. Either way, that's not directly compatible with the `std.Datetime` class we have.

MarkMcCulloh (changes requested):
To make sure this works, we'll have to add some tests to make sure this behaves as expected when consuming this type via JSII.
This would be a great use-case for https://github.com/winglang/wing/issues/1414, which I believe is being worked on right now.

exoego:
This was more complex than my thought

Issue #3914, opened 2023-08-21
Wing sdk tests #2 - allow more targets


tsuf239:
### Feature Spec

* Adding tf-azure target to the test command, 
* Finding a systematic way to skip tests for resources that are not yet implemented

### Use Cases

testing

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

tsuf239:
> Is there an `azure` target now, different from `tf-azure`?

they are the same, fixed the description

monadabot:
Congrats! :rocket: This was released in [Wing 0.39.14](https://github.com/winglang/wing/releases/tag/v0.39.14).

Issue #1371, opened 2023-01-30
cloud.Function - Azure inflight client


hasanaburayyan:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.
				

### Feature Spec

Implement an inflight client for Azure Function. This client will be used when a Function is referenced as an inflight in JavaScript.

### Use Cases

Will allow inflight code to invoke functions.

Should support
```typescript
let someFunction(inflight (msg: str): str => {return "foo bar" });

new cloud.Function(inflight (msg: str): str => {
	let resp = otherFunction.invoke(msg);	 // <---- This is the feature needed
	print("response: ${resp}")
});

```

### Component

SDK

hasanaburayyan:
Still searching for evidence that the SDK supports invoking another function. Documentation online seems to lean towards using http triggers for this 🤷 
https://azure.github.io/azure-sdk/#javascript

staycoolcall911:
Hey @hasanaburayyan, can you please add some wing code to this issue to emphasize what's the DoD for this issue (e.g. what would be supported once this issue is completed)?

hasanaburayyan:
@staycoolcall911 added. Im going to continue reading into the recommended architecture for invoking functions from other functions in Azure. GCP and AWS both have the ability to call functions in their SDK, for Azure its... special. Looks like there may be a way to do Http invocations if you can get the master key of the function.

Chris found this when doing some additional digging. https://learn.microsoft.com/en-us/azure/azure-functions/functions-manually-run-non-http 

For now I separated this from the resource implementation (https://github.com/winglang/wing/pull/1372) since it may take some brainstorming to apply the right abstraction. 

Id like to assign this to myself as I imagine Ill implement this in parallel with https://github.com/winglang/wing/issues/1376

staycoolcall911:
Thanks @hasanaburayyan - I assigned the issue to you and added to the next sprint.

hasanaburayyan:
dropping some references here

- Durable functions allow function chaining: https://build5nines.com/scaling-azure-functions-with-durable-functions/

tsuf239:
part of [#3914](https://github.com/winglang/wing/issues/3914)

monadabot:
Congrats! :rocket: This was released in [Wing 0.39.14](https://github.com/winglang/wing/releases/tag/v0.39.14).

Issue #4497, opened 2023-10-10
Add better auth level to Azure's function


tsuf239:
### Feature Spec

Currently, the function's auth level is "anonymous", which means everyone can invoke it, we need to limit that down.
https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=python-v2%2Cisolated-process%2Cnodejs-v3%2Cfunctionsv2&pivots=programming-language-javascript#http-auth

https://github.com/winglang/wing/blob/b684cfd371672fdf2ff4acd9e7c4bf74df1ecd8d/libs/wingsdk/src/target-tf-azure/function.ts#L105

         

### Use Cases

authentication and security

### Implementation Notes

please wait until [this](https://github.com/winglang/wing/issues/1371) is merged. 

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Issue #4494, opened 2023-10-10
implement `Function.Timeout` for azure


tsuf239:
### Feature Spec

The current implementation seems not to work after implementing the inflight function of Azure- the date was parsed incorrectly.
Please:
- Go over the documentation to see what are the limits of the function- and make sure to throw an error if needed
- Make sure that the props' timeout is parsed correctly
- Add more tests if needed

### Use Cases

Azure implementation of SDK targets

### Implementation Notes

Wait until [this](https://github.com/winglang/wing/issues/1371) is merged.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

PR #4500, opened 2023-10-10
feat(sdk): adding inflight function, application insights and test to tfazure

tsuf239:
## Checklist
fixes:  #1371
fixes: #3914 
image

Added (to tf-azure):
- inflight function client
 - application insights (a resource to collect logs, otherwise debugging is very inconvenient), 
 - test runner 

Also:
 - adjusted the test tf aws function to support any tf target
 - renamed the `WING_TEST_RUNNER_FUNCTION_ARNS` constant to `WING_TEST_RUNNER_FUNCTION_ARNS` to fit the adjustment (that changed around 20 snapshot files) 

All the code containing and tests containing files are on the first commit, the second commit is only snap changes.

Opened the following follow-up issues: 
#4494 
#4497

- [x] Tests added (always)


libs/wingsdk/src/target-tf-azure/function.inflight.ts
subh-cs, line 1: Can you move this under `shared-azure` to match other cloud providers?

See [this](https://github.com/winglang/wing/pull/4376#discussion_r1348248230) for better context.

libs/wingsdk/src/target-tf-azure/test-runner.inflight.ts
subh-cs, line 1: I believe this should be moved under `shared-azure` as well.

garysassano:
FYI, I opened a releated issue: #4514

I think this should have a high priority since it can dramatically improve the DevEx when working with inflight methods in Azure targets.

libs/wingsdk/src/shared-azure/function.inflight.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Invoke the function, passing the given payload as an argument.
+   *
+   * @returns the function returned payload and logs
+   */
+  public async invokeWithLogs(payload: string): Promise<[string, Trace[]]> {
+    console.error("invokeWithLogs is not implemented yet on tf-azure");
+    return [await this.invoke(payload), []];
+  }
```
Chriscbr, line 36: I think this method should throw an error instead of logging one. `TestRunner` for `tf-azure` can use `invoke` instead of `invokeWithLogs` if this method isn't supported

libs/wingsdk/src/target-tf-azure/app.ts
```diff
@@ -76,24 +82,50 @@ export class App extends CdktfApp {
       );
     }
 
-    new AzurermProvider(this, "azure", { features: {} });
+    new AzurermProvider(this, "azure", {
+      features: {
+        resourceGroup: { preventDeletionIfContainsResources: false },
```
Chriscbr, line 87: Maybe add a comment what this does / why it's added

libs/wingsdk/src/shared-azure/function.inflight.ts
Chriscbr, line 1: There's only one supported Azure target right now, so I think we can keep all of the clients in the `target-tf-azure` directory, just like we do with the simulator.

(rename to `libs/wingsdk/src/target-tf-azure/function.inflight.ts`)

libs/wingsdk/src/target-tf-azure/function.ts
```diff
@@ -155,6 +162,9 @@ export class Function extends cloud.Function {
         applicationStack: {
           nodeVersion: functionNodeVersion,
         },
+        applicationInsightsConnectionString:
+          this.applicationInsights.connectionString,
+        applicationInsightsKey: this.applicationInsights.instrumentationKey,
```
Chriscbr, line 167: (just for my own curiosity) What does this do?

Chriscbr (approved):

libs/wingsdk/src/shared-azure/function.inflight.ts
tsuf239: I did it to be able to use the same function between all tf targets- I'll add an issue to include traces in tf-azure

libs/wingsdk/src/target-tf-azure/function.ts
tsuf239: This enables logs and debugging to the function on Azure

libs/wingsdk/src/shared-azure/function.inflight.ts
tsuf239: I agree with you, but, Elad B. has liked a comment on a GCP PR suggesting moving the inflight clients to shared-gcp, and I understand that it'll save us some time when adding a target

libs/wingsdk/src/target-tf-azure/function.ts
Chriscbr: Got it - thanks! (Cloud providers always come up with such intuitive names 🙃)

Issue #3168, opened 2023-06-29
Running tests for tf-aws target gets slow beyond a few tests


skorfmann:
### I tried this:

<details><summary>Creating a few tests (11 in this case) and a minimal API example has a total runtime of something in the range of 8 - 16 minutes</summary>

```
// This is the import statement in Wing.
// Here we bring the Wing standard library that
// contains abstractions of popular cloud services.
bring cloud;

// This code defines a bucket as part of your app.
// When compiling to a specific cloud provider
// it will be substituted by an implementation for
// that cloud. I.e, for AWS it will be an S3 Bucket.

let api = new cloud.Api();

api.get("/hello", inflight(req: cloud.ApiRequest): cloud.ApiResponse => {
    return cloud.ApiResponse {
      status: 200,
      body: "Hello World!"
    };
});

test "Hello World" {
    assert(true);
}

test "Hello World 1" {
    assert(true);
}

test "Hello World 2" {
    assert(true);
}

test "Hello World 3" {
    assert(true);
}

test "Hello World 4" {
    assert(true);
}

test "Hello World 5" {
    assert(true);
}

test "Hello World 6" {
    assert(true);
}

test "Hello World 7" {
    assert(true);
}

test "Hello World 8" {
    assert(true);
}

test "Hello World 9" {
    assert(true);
}

test "Hello World 10" {
    assert(true);
}
```

</details>

Unfortunately, the `--debug` flag doesn't work properly for the target `tf-aws` (see https://github.com/winglang/wing/issues/2882)

See this repo for the full code:  https://github.com/skorfmann/wing-perf-issues-2949

#### Long Test Cycle

```
wing test -t tf-aws perf.w 
pass ─ perf.tfaws » root/Default/env0/test:Hello World
pass ─ perf.tfaws » root/Default/env1/test:Hello World 1
pass ─ perf.tfaws » root/Default/env10/test:Hello World 10
pass ─ perf.tfaws » root/Default/env2/test:Hello World 2
pass ─ perf.tfaws » root/Default/env3/test:Hello World 3
pass ─ perf.tfaws » root/Default/env4/test:Hello World 4
pass ─ perf.tfaws » root/Default/env5/test:Hello World 5
pass ─ perf.tfaws » root/Default/env6/test:Hello World 6
pass ─ perf.tfaws » root/Default/env7/test:Hello World 7
pass ─ perf.tfaws » root/Default/env8/test:Hello World 8
[...]

PR #4504, opened 2023-10-11
chore: skip std spec tests

tsuf239:
## Description 
Currently, there are plenty of tests in a single file of any std test, and our current test scope (which duplicates resources by the number of tests) isn't handling it well, so we often get timeouts and exceptions.
Since the std tests' code is identical both on aws and the sim (and tested in the sim via hangar) I think we can skip it until the [testing issue](https://github.com/winglang/wing/issues/3168) resolves.
link to test run: https://github.com/winglang/wing/actions/runs/6480615481

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,6 @@ on:
     branches:
       - main
       - dev
-      - tsuf/skip-std-tests
   workflow_dispatch: {}
 
 concurrency:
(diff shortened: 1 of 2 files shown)


eladb:
@tsuf239 can you please add a description to the PR with the rationale. It's important to keep track of the reason we are doing things...

eladb (approved):

Issue #4457, opened 2023-10-07
Support inflight classes extending phase independent classes 


yoav-steinberg:
Until we have full support for phase independent classes this means only extending JSII phase independent (non construct) classes:
```wing
bring "some-jsii" as jsii_lib;
inflight Foo extends jsii_lib.SomeClass { // SomeClass isn't a construct so it's considered phase independent
  init(arg1: num, arg2: num) {
    super(arg1); // Note that arg1 needs to be passed to the normal (non async) constructor since our parent doesn't have an async ctor
  }
}
```
See discussion here: https://github.com/winglang/wing/pull/4390#discussion_r1348418995

monadabot:
Congrats! :rocket: This was released in [Wing 0.41.6](https://github.com/winglang/wing/releases/tag/v0.41.6).

PR #4513, opened 2023-10-11
fix(compiler): support inheriting phase independent classes

yoav-steinberg:
See #4457
Enable inflight classes to inherit phase independent classes. Currently the only way to define a phase independent class is by bringing a class that isn't a construct from a JSII lib. So we now support inheriting from such classes.
To do this I needed the inflight class's `init`'s JS code to be split into a regular ctor that can call the phase independent's ctor  (`super(x,y,z)`) and an async `$inflight_init` that'll contain all the rest of the inflight's `init` code. This is relatively easy since we can assume that there's no `this` access before the the `super()` call and that the `super()` call is the first statement in the inflight init.
Two challenges:
* Arguments passed to the `super()` call need to also be available to the inflight init's code: To resolve this I made inflight init's of inflight classes a closure that's defined inside the regular ctor and has access to the ctor's args.
* Defining the the `$inflight_init` as a closure in the inflight object made it impossible to call the parent class's `$inflight_init`, so now if there's a need for it I back up the parent's class's `$inflight_init` as `super_$inflight_init` and use that to call the parent's async init.
 
In addition:
* Added JSII test fixture (see #1414) to be used to test inheriting a phase independent class. This can be used if other places and easily extended to be used in other JSII tests (instead of _jsii-code-samples_).
* Removed class phase form visitor ctx phase tracking. The VisitorContext struct tracks phase so we know the "current execution phase", a class's phase is more a default definition for members in the class and not a real indication of the execution phase. In other words we only need to update the current phase when stepping into a function/closure not when stepping into a class definition.

## Checklist

- [x] Tests added (always)


examples/jsii-fixture/lib/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1,222 @@
+{
+  "program": {
```
eladb, line 2: Probably no need to commit this file

eladb:
Perhaps add a few words in the PR description on the implementation strategy

eladb:
@yoav-steinberg this is not finished yet, right? I can't seem to find some of the new logic as well as the snapshots for the new tests. I am also not sure I understand how this change solves the issue.

yoav-steinberg:
> @yoav-steinberg this is not finished yet, right? I can't seem to find some of the new logic as well as the snapshots for the new tests. I am also not sure I understand how this change solves the issue.

It's almost there, there's still a test failure that might hint I missed something big..
I was counting on self-mut to update all snapshots and then go over them.

I'll definitely add a better description in tje PR, just wanted to get it in..

libs/wingc/src/jsify.rs
```diff
@@ -1116,67 +1196,69 @@ impl<'a> JSifier<'a> {
...
-
-		// default base class for preflight classes is `core.Resource`
-		let extends = if let Some(parent) = &class.parent {
-			format!(" extends {}", self.jsify_user_defined_type(parent, ctx))
-		} else {
-			format!(" extends {}", STDLIB_CORE_RESOURCE)
-		};
+			// default base class for preflight classes is `core.Resource`
+			let extends = if let Some(parent) = &class.parent {
+				format!(" extends {}", self.jsify_user_defined_type(parent, ctx)) //bg! we think we're inflight but...
```
Chriscbr, line 1234: ?

libs/wingc/src/jsify/snapshots/fails_when_inflight_class_tries_to_extend_preflight_class.snap
```diff
@@ -3,4 +3,3 @@ source: libs/wingc/src/jsify/tests.rs
 ---
 ## Errors
 Class "Derived" is an inflight class and cannot extend preflight class "Base" 2:35
-Cannot qualify access to a lifted type "Base" (see https://github.com/winglang/wing/issues/76 for more details) 2:35
```
Chriscbr, line 6: Possible regression?

examples/jsii-fixture/src/index.ts
Chriscbr, line 1: nice!! This will be super helpful for all kinds of testing for the jsii importer and other features

libs/wingc/src/jsify/snapshots/fails_when_inflight_class_tries_to_extend_preflight_class.snap
Chriscbr: I took a second look at the test case -- I'm guessing this is fine. We're already giving an error that an inflight class can't extend a preflight class, I'm not sure the second error adds any value.

Chriscbr (approved):

libs/wingc/src/jsify.rs
```diff
@@ -1057,6 +1070,73 @@ impl<'a> JSifier<'a> {
...
+				..
+			}) = body_scope.statements.iter().next()
+			{
+				let args = self.jsify_arg_list(&arg_list, None, None, ctx);
+				code.line(format!("super({args});"));
+
+				// If our parent's phase in inflight then backup a reference to the paren't inflight init to be used in the super ctor call
+				if parent_class_phase(ctx) == Phase::Inflight {
+					code.line(format!(
+						"this.super_{CLASS_INFLIGHT_INIT_NAME} = this.{CLASS_INFLIGHT_INIT_NAME};"
```
Chriscbr, line 1123:
```suggestion
						"this.{CLASS_SUPER_INFLIGHT_INIT_NAME} = this.{CLASS_INFLIGHT_INIT_NAME};"
```

libs/wingc/src/jsify.rs
```diff
@@ -1057,6 +1070,73 @@ impl<'a> JSifier<'a> {
...
+
+			for p in &func_def.signature.parameters {
+				if p.variadic {
+					parameter_list.push("...".to_string() + &p.name.to_string());
+				} else {
+					parameter_list.push(p.name.to_string());
+				}
+			}
+
+			let parameters = parameter_list.iter().map(|x| x.as_str()).collect::<Vec<_>>().join(", ");
```
Chriscbr, line 1107: Do we have a `jsify_parameters()` function available?

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_inside_inflight_closure.test.w_compile_tf-aws.md
```diff
@@ -16,8 +16,10 @@ module.exports = function({ $__parent_this_1_b }) {
...
           {((cond) => {if (!cond) throw new Error("assertion failed: this.field == \"value\"")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(this.field,"value")))};
         }
-        async $inflight_init() {
-          this.field = "value";
+        constructor(){
+          this.$inflight_init = async () => {
+            this.field = "value";
+          }
         }
       }
```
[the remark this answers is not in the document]
yoav-steinberg: It's a functional change. The new code enabled the inflight_init to access the arguments passed to the ctor threw regular closure capturing. This means the args are evaluated only once when passed to the ctor but can be used in both the ctor and async inflight init context:

```js
class Foo extends $jsii_fixture_JsiiClass {
  constructor(arg){
    super(arg);
    this.$inflight_init = async () => {
      // use arg
    };
  }
}

cont f = new Foo(some_function_call_with_side_effects());
```

Note that `some_function_call_with_side_effects` is only called once so the side effects happen only once. But we can use its result in both the ctor (pass them to `super`) and in the async inflight init.

libs/wingc/src/jsify.rs
yoav-steinberg: Oops, leftovers I'll clean up

libs/wingc/src/jsify.rs
yoav-steinberg: :+1: added

PR #4515, opened 2023-10-12
feat(sdk): add `contentType` option to `Bucket.put()`

### LINKED ISSUE #4512: Customize `contentType` of objects in `Bucket.put`
### Feature Spec

Current signature of `bucket.put`:

```js
/**
 * Put an object in the bucket.
 * @param key Key of the object.
 * @param body Content of the object we want to store into the bucket.
 * @inflight
 */
inflight put(key: str, body: str): void;
```

New signature of `bucket.put`:

```js
/**
 * Put an object in the bucket.
 * @param key Key of the object.
 * @param body Content of the object we want to store into the bucket.
 * @param options Additional options
 * @inflight
 */
inflight put(key: str, body: str, options: BucketPutOptions): void;

/**
 * Additional options for `Bucket.put`.
 */
struct BucketPutOptions {
  /**
   * The HTTP Content-Type of the object.
   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
   */
  contentType: str;
}
```

### Use Cases

This is a common option that can be configured for all objects on [AWS S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_HeadObject_section.html), [Azure Blob Storage](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.models.blobproperties.contenttype?view=azure-dotnet), and [Google Cloud Storage](https://cloud.google.com/storage/docs/metadata).

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

garysassano:
Closes #4512

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -23,6 +23,17 @@ export interface BucketProps {
   readonly public?: boolean;
 }
 
+/**
+ * Options for `Bucket.put`.
+ */
+export interface BucketPutProps {
```
eladb, line 29: We need to decide on a naming convention for these option structs. I think we normally call them `XxxOptions` if they are not initialization props.

@Chriscbr what do you think?

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -23,6 +23,17 @@ export interface BucketProps {
...
+  /**
+   * The HTTP Content-Type of the object.
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
+   */
+  readonly contentType: string;
```
eladb, line 34: This should be optional and have a default of auto-detecting the content type from the file extension.

docs/docs/04-standard-library/01-cloud/bucket.md
```diff
@@ -435,7 +435,7 @@ Returns a url to the given file.
 ##### `put` <a name="put" id="@winglang/sdk.cloud.IBucketClient.put"></a>
 
 ```wing
```
eladb, line 437: Missing sdk test

eladb (changes requested):

libs/wingsdk/src/cloud/bucket.ts
garysassano: I was also going to bring this up. Actually, @Chriscbr proposed to call the struct `BucketPutOptions`, but then I saw `SetConsumerProps` used for `Queue`, `ServiceOnStartProps` used for `Service`, etc. so I wanted to follow the convention. The only place where I introduced the word `Option` is inside `Json`.

libs/wingsdk/src/cloud/bucket.ts
garysassano: I checked the AWS docs and only the Java SDK has the built-in Content-Type autodetection (see [HERE](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ObjectMetadata.html#setContentType-java.lang.String-)). So I guess you want to implement this functionality using the [mime module](https://www.npmjs.com/package/mime)?

garysassano: If the user takes a `.jpg` file and renames it to `.txt`, should the autodetection be able to determine the actual file type based on its binary information or should it just perform a shallow check looking only at its extension?

garysassano: The problem with `mime` module is that it determines the MIME type based solely on the file extension, and it doesn't examine the file's binary contents.

eladb: > If the user takes a .jpg file and renames it to .txt, should the autodetection be able to determine the actual file type based on its binary information or should it just perform a shallow check looking only at its extension?

Just the extension. I don't see this as an issue. It's a very common practice to use the extension to determine the mime type.

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -23,6 +23,17 @@ export interface BucketProps {
   readonly public?: boolean;
 }
 
+/**
+ * Options for `Bucket.put`.
+ */
+export interface BucketPutProps {
+  /**
+   * The HTTP Content-Type of the object.
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
```
eladb, line 32: Missing "@default" documentation

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -59,11 +61,17 @@ export class BucketClient implements IBucketClient {
...
+    key: string,
+    body: string,
+    props?: BucketPutProps
+  ): Promise<void> {
     const command = new PutObjectCommand({
       Bucket: this.bucketName,
       Key: key,
       Body: body,
+      ContentType:
+        props?.contentType || mime.getType(key) || "application/octet-stream",
```
eladb, line 74:
```suggestion
        props?.contentType ?? mime.getType(key) ?? "application/octet-stream",
```

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -340,4 +350,14 @@ export class BucketClient implements IBucketClient {
       await this.s3Client.send(command);
     return region;
   }
+
+  // private async function getMimeType(buffer) {
```
eladb, line 354: Remove

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -57,10 +58,18 @@ export class BucketClient implements IBucketClient {
...
-      .upload(body, body.length);
+  public async put(
+    key: string,
+    body: string,
+    props?: BucketPutProps
+  ): Promise<void> {
+    const blobClient = this.containerClient.getBlockBlobClient(key);
+    const options = {
+      blobHTTPHeaders: {
+        blobContentType: props?.contentType,
```
eladb, line 69: Missing auto detect or does this happen automatically in azure? If so, better document.

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -278,9 +289,10 @@ export interface IBucketClient {
    * Put an object in the bucket.
    * @param key Key of the object.
    * @param body Content of the object we want to store into the bucket.
+   * @param options Additional options
    * @inflight
    */
-  put(key: string, body: string): Promise<void>;
+  put(key: string, body: string, options?: BucketPutProps): void;
```
eladb, line 296: What about the sim implementation? Where are we storing the content type there?

eladb (changes requested):

libs/wingsdk/src/cloud/bucket.ts
garysassano: Actually, I have no clue. The local filesystem of most OSs doesn't store the Content-Type of a file as metadata since it can infer the type from the file extension. So I'm not sure if Wing Simulator actually needs to handle that.

garysassano: An approach we could use is to create a metadata JSON file with the same name as the uploaded file but with an additional `.meta` extension.

eladb: I think we will need something like this because we would need a way to query the content type of an object.

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -250,15 +256,37 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
...
     const hash = this.hashKey(key);
     const filename = join(this._fileDir, hash);
     const dirName = dirname(filename);
+
     await fs.promises.mkdir(dirName, { recursive: true });
     await fs.promises.writeFile(filename, value);
+
+    const determinedContentType =
+      contentType ?? mime.getType(key) ?? "application/octet-stream";
+    await this.storeMetadata(filename, determinedContentType);
```
Chriscbr, line 288: Couldn't we store this in a simple `Map` or `Record<string, Metadata>`? The metadata is pretty small and it will probably be faster as well
```suggestion
    this.metadata[key] = { contentType: determinedContentType };
```

libs/wingsdk/src/target-sim/bucket.inflight.ts
Chriscbr, line 1: The `metadata` inflight method needs to be updated

Chriscbr (changes requested):

libs/wingsdk/src/target-sim/bucket.inflight.ts
garysassano: `metadata` is a method, so how can you assign a value to it?

libs/wingsdk/src/target-sim/bucket.inflight.ts
garysassano: I did a bit of refactoring, please tell me what you think.

garysassano: I did a bit of refactoring, it should be much cleaner now.

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -227,43 +234,52 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
    * @param key Key of the object.
    * @throws if the object does not exist.
    */
-  public async metadata(key: string): Promise<ObjectMetadata> {
+  public async getMetadata(key: string): Promise<ObjectMetadata> {
```
eladb, line 237: @Chriscbr I don't think we've established these API naming guidelines yet, but my preference is not to use `getXxx()` for getters, just `xxxx()`. What do you think?

```suggestion
  public async metadata(key: string): Promise<ObjectMetadata> {
```

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -227,43 +234,52 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
...
+      lastModified: Datetime.fromIso(filestat.mtime.toISOString()),
+      contentType: determinedContentType,
+    };
+
     this.objectKeys.add(key);
     await this.notifyListeners(actionType, key);
   }
 
-  private hashKey(key: string): string {
+  private getHashKey(key: string): string {
```
eladb, line 282:
```suggestion
  private hashKey(key: string): string {
```

eladb (changes requested):

libs/wingsdk/src/target-sim/bucket.inflight.ts
garysassano: For me it's actually confusing to not have verbs to define methods since that's what I've always seen as [common best practice](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods) in OOP. Besides, you have a `Bucket.get()` method to get the object, but then you don't want a `Bucket.getMetadata()` to get the object's metadata? At first I thought that was an attribute, not a method. I had the exact same issue when using the `cloud.Secret`, where I thought `value()` was an attribute and not a method.

garysassano: If you don't want to use getters altogether, then `Bucket.get()` should be renamed to `Bucket.object()` for consistency, which ends up becoming as confusing as `Secret.value()`.

Chriscbr: I tend to agree with @eladb here - I think we can leave the method as `metadata()` for now. Longer, precise names like "getObject()" or "getMetadata()" can sometimes bring more consistency to our APIs, but we have to balance it with the benefits of making our APIs convenient to read and write. In this case, I don't feel like changing `metadata()` to `getMetadata()` clarifies what the method does in a significant way. But it's definitely something we can revisit if a lot of users get stuck on this...

eladb (approved):

Chriscbr (approved):

PR #4523, opened 2023-10-12
feat(sdk)!: support Map.tryGet()

hasanaburayyan:
Includes the changes from: https://github.com/winglang/wing/pull/3771 that giovanni did.

Closes: https://github.com/winglang/wing/issues/3575

## Checklist

- [x] Tests added (always)

Co-authored-by: Giovanni Orciuolo <giovanni@orciuolo.it>
BREAKING CHANGE: `Map.get()` now throws if key does not exist.


libs/wingsdk/src/std/map.ts
```diff
@@ -39,7 +39,7 @@ export class Map {
    * If the value that is associated to the provided key is an object, then you will get a reference
    * to that object and any change made to that object will effectively modify it inside the map.
    *
-   * @macro ($self$)[$args$]
+   * @macro ((obj, key) => { if (obj[key] === undefined) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })($self$, $args$)
```
Chriscbr, line 42: I want to raise a flag that there's secretly a bug here, because if a user has a `Map<str?>`, then calling `get` with a valid key could return `nil`, and since we represent empty optionals with JavaScript's `undefined`, this would throw an error.

I think the bug might be beyond the scope of this PR - could you open an issue about this?

libs/wingc/src/jsify/snapshots/access_property_on_value_returned_from_collection.snap
```diff
@@ -16,15 +16,15 @@ source: libs/wingc/src/jsify/tests.rs
 ## inflight.$Closure1-1.js
 
 ```js
-module.exports = function({ $_s___hello___length }) {
+module.exports = function({ $__obj__key_______if__obj_key______undefined__throw_new_Error__Map_does_not_contain_key_____key______return_obj_key______s___hello___length }) {
```
Chriscbr, line 19: That's odd looking. 😅 If it works that's all good, though I don't think the compiler should be generating super long identifiers like this. Open an issue?

Chriscbr (approved):

examples/tests/sdk_tests/std/map.test.w
```diff
@@ -71,26 +71,26 @@ test "size()" {
 let greeting = { "hello" => "there!" };
 assert(greeting.get("hello") == "there!");
 
-let general: str? = greeting.get("grievous");
+let general: str? = greeting.tryGet("grievous");
```
eladb, line 74: Where are the `get` tests?

libs/wingsdk/src/std/map.ts
eladb: Interesting use case. One may argue that logically there is no difference between an optional without a value and no value at all.

As long as `has()` returns `false` in this case, I think this might be reasonable behavior.

All in all, out of scope (but let's open a tracking issue and see if this creates friction).

eladb (approved):

examples/tests/sdk_tests/std/map.test.w
hasanaburayyan: There we already a bunch of existing ones in: examples/tests/valid/container_types.test.w

eladb: The tests under `valid/` are not "official" tests. We should aim to have coverage for all of our SDK APIs under the `sdk_tests/` tree.

libs/wingsdk/src/std/map.ts
hasanaburayyan: Wild, yea great eye! Opened tracking issue: https://github.com/winglang/wing/issues/4525

examples/tests/sdk_tests/std/map.test.w
hasanaburayyan: Ah okay. I guess I was still under the impression everything in our `valid` was "official" but yea this file already has happy path tests for `.get` I added some unhappy path tests in the unofficial files Ill just move those

libs/wingsdk/src/std/map.ts
MarkMcCulloh: technically this should use `in`

```suggestion
   * @macro ((obj, key) => { if (!(key in obj)) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })($self$, $args$)
```

hasanaburayyan: > technically this should use `in`

🧠 yea that does make sense, it also squashes the bug. Incoming new commit

PR #4541, opened 2023-10-14
fix(sdk): set default values cloud.Queue (sim/aws)

### LINKED ISSUE #4521: `cloud.Queue` props have different default values on different targets
Looking at their implementations:
- https://github.com/winglang/wing/blob/43a880856a6c820fe7b58fbcfb860b99c2f574a9/libs/wingsdk/src/target-sim/queue.ts
- https://github.com/winglang/wing/blob/43a880856a6c820fe7b58fbcfb860b99c2f574a9/libs/wingsdk/src/target-tf-aws/queue.ts

The `sim` queue has a default timeout of 10 seconds and a default retention period of 1hr, while the AWS queue doesn't specify a default so it likely falls back to the AWS defaults of a 30 second timeout and 4 day retention period. I expected the default to be the same on both targets.

marciocadev:
Set the timeout to 30 seconds and the retention period to 1 hour for the `sim` and `aws` targets.

Closes #4521 

## Checklist

- [x] Tests added (always)


garysassano:
Default message retention period for an SQS queue is 4 days, so wouldn't it make sense to use that for AWS targets?

Also, since an SQS queue has the following range limits:
- `VisibilityTimeout`: 0s - 12h (default 30s)
- `MessageRetentionPeriod`: 1m - 14d (default 4d)

Wouldn't it be nice to put guardrails in the code? Or we delegate to the AWS API to throw errors if the value provided by the user is out of allowed range?

tsuf239 (commented):
seems ok to me, running the sdk tests now on aws :)

tsuf239:
> Wouldn't it be nice to put guardrails in the code? Or we delegate to the AWS API to throw errors if the value provided by the user is out of allowed range?

I tend to agree, but it would differ for each target (and has no limits in the sim target), I think we brought up those differences on one of our team times, but I can't remember what the suggested solution was. @Chriscbr @hasanaburayyan @staycoolcall911 Do you remember?

tsuf239 (approved):
all tests passes :) 
https://github.com/winglang/wing/actions/runs/6524353138/job/17715902625

Chriscbr:
@tsuf239 For numerical-like values that must be in a certain range, it's probably okay if we only throw an error based on the target you're compiling to, since the limits could vary a lot between AWS and Azure etc. For example, if you compile to AWS, the SDK will throw an error if your queue's retention period is longer than 14 days, and if you compile to Azure, it will only throw an error if your queue's retention period is longer than Azure's threshold, etc. The simulator doesn't need to have a limit.

tsuf239:
@marciocadev can you try to merge main in? I think it would resolve the failed test :)

Issue #4510, opened 2023-10-11
Enforce inflight APIs of preflight classes only accept/return serializable values


Chriscbr:
### Feature Spec

The Wing simulator does not support inflight methods on preflight classes that accept OR return serializable values. The reason is that inflight methods may be called other processes. When a separate process (e.g. a Docker container) invokes an inflight API, it must serialize the function call's arguments and send them over HTTP to the simulator process. Likewise, the result of the function call is serialized and sent over HTTP back to the requester process.

The following types should be serializable:
- `num`
- `str`
- `duration`
- `boolean`
- `Json` / `MutJson`
- `Map<T>` / `MutMap<T>` for serializable T
- `Array<T>` / `MutArray<T>` for serializable T
- `Set<T>` / `MutSet<T>` for serializable T
- structs where all fields are serialzable
- enums
- `T?` for serializable T
- `anything`

The following types are not serializable:
- classes
- functions

In the compiler this should be enforced:
- while type checking inflight methods on user-defined classes
- while importing JSII APIs

### Use Cases

Prevent Wing APIs that will not work with the `sim` target.

In the future we might be able to loosen this restriction through annotations or other mechanisms, for scenarios where an inflight method takes a non-serializable value but it does not need to perform any work over the wire, and/or the responsibility of serialization could be in a dedicated function etc.

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

staycoolcall911:
Closed a duplicate issue #6475, which added some reasoning behind this change:
>The motivation would be to add a constraint so that we can guarantee inflight methods on preflight classes can be safely called from other language runtimes.
We may also want to consider where we don't need to enforce this -- for example, perhaps private methods and fields can be allowed to use non-serializable types.

PR #4561, opened 2023-10-16
feat: `sim.State`

eladb:
A simple key/value in-memory state for the simulator with support for lazy tokens.

Consider a use case where there is an attribute of a simulated service that only gets resolved during initialization (e.g. the exposed port of a container). In order to create such resources, we need a way to obtain a lazy token that gets resolved during simulator initialization. It just so happens that we already have a mechanism like this, but it was not exposed as a public API.

Use the preflight method `token(key)` to obtain a token that can be used to reference the value of the state at runtime.

During deployment (i.e. `cloud.OnDeploy` or `cloud.Service` startup), you must call the inflight method `set(key, value)` to set the runtime value. The value will be available at runtime through the inflight method `get(key)` (or resolved as a token).

See tests for examples.

Inspired by #3793

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/state.inflight.ts
```diff
@@ -0,0 +1,31 @@
...
+  public async init(): Promise<Record<string, any>> {
+    return {};
+  }
+
+  public async cleanup(): Promise<void> {}
+
+  public async set(key: string, value: any): Promise<void> {
+    // once the property is set on the object, the token will be resolved.
+    Object.defineProperty(this, key, { value });
+    this.context.setResourceAttributes(this.path, { [key]: value });
```
Chriscbr, line 20: What is the difference between these lines? I'm not sure I understand what the `defineProperty` line does

libs/wingsdk/src/target-sim/state.ts
```diff
@@ -0,0 +1,90 @@
...
+    };
+  }
+}
+
+/**
+ * Inflight interface for `State`.
+ */
+export interface IStateClient {
+  /**
+   * Sets the runtime state of this object.
```
Chriscbr, line 68: ?
```suggestion
   * Sets the runtime state of this object. Can only be called once.
```

libs/wingsdk/src/target-sim/state.ts
Chriscbr, line 1: Maybe we can add a `state.md` file to go along with this

Chriscbr (approved):
I like the API! Looks pretty easy to use. I added some random comments and questions but none of them are blocking

libs/wingsdk/src/target-sim/state.ts
Chriscbr: On second thought I guess there are probably use cases for this... I got this example working:

```js
class QueueSim {
  state: sim.State;
  messagesKey: str;
  init() {
    this.state = new sim.State();
    this.messagesKey = "messages";
  }
  
  inflight init() {
    this.state.set(this.messagesKey, MutArray<str>[]);
  }

  inflight _getMessages(): MutArray<str> {
    return unsafeCast(this.state.get(this.messagesKey));
  }

  pub inflight push(message: str) {
    let messages = this._getMessages();
    messages.push(message);
  }

  pub inflight pop(): str {
    let messages = this._getMessages();
    return messages.pop();
  }
}

let queue = new QueueSim();

test "QueueSim" {
  queue.push("hello");
  queue.push("world");
  assert(queue.pop() == "world");
  assert(queue.pop() == "hello");
}
```

libs/wingsdk/src/target-sim/state.ts
```diff
@@ -0,0 +1,90 @@
...
+   * @param key The object's key
+   * @param value The object's value
+   */
+  set(key: string, value: any): Promise<void>;
```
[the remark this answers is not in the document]
Chriscbr: If we have inflight methods that get or return `any`, it will be harder to implement https://github.com/winglang/wing/issues/4510

Chriscbr:
Here's a random example I tried to dogfood this - modeling a simulated `cloud.Queue`:

```js
class QueueSim {
  state: sim.State;
  messagesKey: str;
  init() {
    this.state = new sim.State();
    this.messagesKey = "messages";
  }
  
  inflight init() {
    this.state.set(this.messagesKey, MutArray<str>[]);
  }

  inflight _getMessages(): MutArray<str> {
    return unsafeCast(this.state.get(this.messagesKey));
  }

  pub inflight push(message: str) {
    let messages = this._getMessages();
    messages.push(message);
  }

  pub inflight pop(): str {
    let messages = this._getMessages();
    return messages.pop();
  }

  pub inflight size(): num {
    let messages = this._getMessages();
    return messages.length;
  }
}

let queue = new QueueSim();

let checkSize = new cloud.Function(inflight () => {
  assert(queue.size() == 2);
});

test "QueueSim" {
  queue.push("hello");
  queue.push("world");
  checkSize.invoke("");
  assert(queue.pop() == "world");
  assert(queue.pop() == "hello");
}
```

When I run the test, I get this error. I realized the problem was I'm initializing `this.state` inside of an `inflight init`, which will be executed once per inflight host.

```
fail ┌ service.main.wsim » root/env0/test:QueueSim
     │ TypeError: Cannot redefine property: messages
     │     at Function.defineProperty (<anonymous>)
     │     at State.set (/Users/chrisr/dev/wing1/libs/wingsdk/lib/target-sim/state.inflight.js:15:16)
```

We've already discussed some of the potential confusion of `inflight init()` / whether we should get rid of it or change it some way, so okay.

Next I tried to fix my code by changing my helper function so it sets the state if it hasn't been initialized yet:

```js
  inflight _getMessages(): MutArray<str> {
    if this.state.get(this.messagesKey) == nil {
      this.state.set(this.messagesKey, MutArray<str>[]);
    }
    return unsafeCast(this.state.get(this.messagesKey));
  }
```

But then I got the error "Error: Key messages not found". `get()` doesn't like it if the key hasn't been set (which makes sense). I guess we should add a `tryGet` method, yeah?

I saw from your example that I could use a `cloud.Service`, but it felt like an oversized tool for the job (I just wanted to initialize the queue's array once -- I didn't care to model the queue as a service). Then I came up with the idea of using `cloud.OnDeploy` instead:

```js
  init() {
    this.state = new sim.State();
    this.messagesKey = "messages";

    new cloud.OnDeploy(inflight () => {
      this.state.set(this.messagesKey, MutArray<str>[]);
    });
  }
```

and then my example worked! 🎉

libs/wingsdk/src/target-sim/state.ts
```diff
@@ -0,0 +1,90 @@
...
+      props: {},
+      attrs: {},
+    };
+  }
+}
+
+/**
+ * Inflight interface for `State`.
+ */
+export interface IStateClient {
```
Chriscbr, line 66: Add tryGet()?

libs/wingsdk/src/target-sim/state.inflight.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  public async get(key: string): Promise<any> {
+    const value = this.context.resourceAttributes(this.path)[key];
+    if (value === undefined) {
+      throw new Error(`Key ${key} not found`);
```
Chriscbr, line 26:
```suggestion
      throw new Error(`Key "${key}" not found`);
```

libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -310,3 +311,12 @@ export interface DynamodbTableSchema extends BaseResourceSchema {
...
 }
+
+/** Schema for simulator.State */
+export interface StateSchema extends BaseResourceSchema {
+  readonly type: typeof SIMULATOR_STATE_TYPE;
+  readonly props: {};
+}
+
+/** Runtime attributes for cloud.Bucket */
+export interface BucketAttributes {}
```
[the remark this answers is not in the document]
eladb: Removed it

libs/wingsdk/src/target-sim/state.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @inflight `@winglang/sdk.sim.IStateClient`
+ */
+export class State extends Resource implements ISimulatorResource {
+  /**
+   * Returns a token that can be used to retrieve the value of the state after the simulation has
+   * run.
+   * @param key The object key retrieved through the inflight `state.get()`.
+   */
+  public token(key: string): string {
+    return simulatorAttrToken(this, key);
```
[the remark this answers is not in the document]
eladb: The current scheme might be tricky if the path is funky, you are right. I'll open an issue to revisit this, but prefer not to do it as part of this PR.

libs/wingsdk/src/target-sim/state.ts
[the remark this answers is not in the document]
eladb: OK, I'll scope it down to `Json` but this means it won't be possible to use this to store arbitrary objects (e.g. an `http.Server` object). 

Having said that, you are right that now since our simulator APIs are remote, we can't pass along this type of state anyway. It is is pretty important to implement https://github.com/winglang/wing/issues/4510 as soon as possible, even if we do it initially just as a API checker rule.

libs/wingsdk/src/target-sim/state.ts
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @inflight `@winglang/sdk.sim.IStateClient`
+ */
+export class State extends Resource implements ISimulatorResource {
+  /**
+   * Returns a token that can be used to retrieve the value of the state after the simulation has
+   * run.
+   * @param key The object key retrieved through the inflight `state.get()`.
+   */
+  public token(key: string): string {
```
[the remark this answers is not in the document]
eladb: yes, we should :-)

libs/wingsdk/src/target-sim/state.ts
eladb: Done

libs/wingsdk/src/target-sim/state.inflight.ts
eladb: The `defineProperty` was a residual from a previous implementation. Not needed whatsoever...

eladb:
@Chriscbr your example failed because there was a residual code to used `Object.defineProperty` to set the state on the object. I removed it, so now it doesn't fail.

But... Using `inflight init` for this doesn't make any sense at all because `inflight init` is executed *per client* and not once. This means, for example, that if there are multiple cloud functions interacting with your queue, each one will reset the queue messages :-(.

This goes back to something we discussed around this misleading syntax around (which have been tripping both you and me already in some examples). The use case for inflight init is to cache client-side state (not server-side), but something about syntax is very misleading.

PR #4572, opened 2023-10-17
chore: clean up obsolete snapshots

MarkMcCulloh:


MarkMcCulloh:
vitest normally handles this afaik, but because we are doing custom snapshotting we have to handle it ourselves (by writing something)

PR #4590, opened 2023-10-18
chore(sdk): force usage of mime v3 throughout repo

MarkMcCulloh:
We were already depending on `mime`, so this does not increase our dependency closure.

The version of `mime` that `express` depends on (via the `send` package) is very old and can easily conflict with other dependencies we may want to add. This forces our packages to always use v3 of `mime`, and patches `mime` v3 to expose the previously removed functions.


Chriscbr (approved):
🫡

PR #4605, opened 2023-10-19
feat(sdk): bring ui

Chriscbr:
Introducing the `ui` module to the SDK, a collection of APIs for adding visual interactions for your classes in the Wing Console! Here's an example of the API in action:

```js
bring cloud;
bring ui;

class WidgetService {
  data: cloud.Bucket;
  counter: cloud.Counter;
  init() {
    this.data = new cloud.Bucket();
    this.counter = new cloud.Counter();
    
    // a button lets you invoke any inflight function
    new ui.Button("Add widget", inflight () => { this.addWidget(); });
    
    // a field displays a labeled value, with optional refreshing
    new ui.Field(
      "Total widgets",
      inflight () => { return this.countWidgets(); },
      refreshRate: 5s,
    );
  }

  inflight addWidget() {
    let id = this.counter.inc();
    this.data.put("widget-${id}", "my data");
  }

  inflight countWidgets(): str {
    return "${this.data.list().length}";
  }
}

new WidgetService();
```

UI components aren't displayed in the Wing Console quite yet, but they can be accessed and called through the simulator API like so:

```js

async function main() {
    const simulator = new sdk.simulator.Simulator({ simfile: ... });
    await simulator.start();

    const uiComponents = simulator.getResourceUIComponents("root/Default/WidgetService");
    console.log("uiComponents:", JSON.stringify(uiComponents, null, 2));
    const buttonHandlerPath = uiComponents[0].handlerPath;
    const fieldHandlerPath = uiComponents[1].handlerPath;

    const buttonClient = simulator.getResource(buttonHandlerPath);
    await buttonClient.invoke();
    await buttonClient.invoke();

    const fieldClient = simulator.getResource(fieldHandlerPath);
    const result = await fieldClient.invoke();
    console.log("total widgets:", result);

    await simulator.stop();
}

main();
```

<details>
<summary>Console output</summary>

```
uiComponents: [
  {
    "kind": "button",
    "label": "Add widget",
    "handlerPath": "root/Default/WidgetService/ui.Button/Handler"
  },
  {
    "kind": "field",
    "label": "Total widgets",
    "handlerPath": "root/Default/WidgetService/ui.Field/Handler",
    "refreshRate": 5
  }
]
total widgets: 2
```

</details>

In addition to `ui.Field` and `ui.Button` mentioned above, you can also use `ui.Section` to group components together:

```js
let section = new ui.Section();
section.add(button1, button2);
```

TODO:
- [x] tests
- [x] docs

Future work:
- Allow inflight methods to be passed directly, i.e. `new ui.Button("label", this.myInflightMethod)` (requires compiler support)

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/ui/ui.ts
Chriscbr, line 1: Move these classes into separate files

libs/wingsdk/src/simulator/tree.ts
```diff
@@ -13,4 +13,21 @@ export class Tree {
   public rawData(): ConstructTree {
     return structuredClone(this.data);
   }
+
+  /**
+   * Returns the raw data for a specific construct node.
+   */
+  public rawDataForNode(path: string): ConstructTreeNode | undefined {
```
Chriscbr, line 20: Unit test

libs/wingsdk/src/ui/ui.ts
Chriscbr, line 1: Unit test (create an app with some UI components, and make an assertion on the generated `tree` information)

libs/wingsdk/src/ui/visual.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const VISUAL_COMPONENT_SYMBOL = Symbol.for(
+  "@winglang/sdk.ui.VisualComponent"
+);
+
+/**
+ * A visual component is used to customize the view of other classes in the Wing Console.
+ * This is a base class for all other visual components.
+ */
+export abstract class VisualComponent extends Resource {
```
Chriscbr, line 14: Naming suggestions welcome

libs/wingsdk/src/core/tree.ts
```diff
@@ -68,6 +69,37 @@ export interface DisplayInfo {
    * @default - no source information
    */
   readonly sourceModule?: string;
+
+  /**
+   * UI components to display for this resource.
+   * @default - no UI components
+   */
+  readonly uiComponents?: any[]; // UIComponent
```
eladb, line 77: Maybe just `ui`

```suggestion
  readonly ui?: any[]; // UIComponent
```

libs/wingsdk/src/core/tree.ts
```diff
@@ -68,6 +69,37 @@ export interface DisplayInfo {
...
+}
+
+/** @internal */
+export type UIComponent = UIField | UISection | UIButton;
```
eladb, line 81: Maybe also add `UILabel` and `UILink`?

libs/wingsdk/src/core/tree.ts
```diff
@@ -68,6 +69,37 @@ export interface DisplayInfo {
...
+export interface UIButton {
+  readonly kind: "button";
+  readonly label: string;
+  readonly handlerPath: string;
+}
+
+/** @internal */
+export interface UISection {
+  readonly kind: "section";
+  readonly label: string | undefined;
```
eladb, line 101:
```suggestion
  readonly label?: string;
```

nit picking

libs/wingsdk/src/core/tree.ts
```diff
@@ -68,6 +69,37 @@ export interface DisplayInfo {
...
+  readonly label: string;
+  readonly handlerPath: string;
+  readonly refreshRate: number | undefined;
+}
+
+/** @internal */
+export interface UIButton {
+  readonly kind: "button";
+  readonly label: string;
+  readonly handlerPath: string;
```
eladb, line 95: "Path" is an ambiguous anyway so I would just call this `handler` and add some docstring that explains this is the node path of a cloud function handler.

libs/wingsdk/src/core/tree.ts
```diff
@@ -160,12 +192,27 @@ function synthDisplay(construct: IConstruct): DisplayInfo | undefined {
     return;
   }
   const display = Node.of(construct);
-  if (display.description || display.title || display.hidden) {
+
+  // generate ui data only based on direct children
+  const uiComponents: UIComponent[] = [];
+  if (
+    App.of(construct)._target === "sim" &&
```
eladb, line 199: Why is this condition needed?

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -400,6 +400,18 @@ export class Simulator {
     return config;
   }
 
+  /**
+   * Obtain a resource's visual interaction components.
+   * @returns An array of UIComponent objects
+   */
+  public getResourceUIComponents(path: string): any {
```
eladb, line 407:
```suggestion
  public getResourceUI(path: string): any[] {
```

libs/wingsdk/src/ui/ui.ts
```diff
@@ -0,0 +1,253 @@
...
+    handler: IFieldHandler,
+    props: FieldProps = {}
+  ) {
+    super(scope, id);
+
+    this.label = label;
+    this.refreshRate = props.refreshRate?.seconds;
+
+    if (App.of(this)._target === "sim") {
+      this.fn = App.of(scope).newAbstract(
```
eladb, line 127: Maybe use Function.newFunction()

libs/wingsdk/src/ui/ui.ts
```diff
@@ -0,0 +1,253 @@
...
+    label: string,
+    handler: IFieldHandler,
+    props: FieldProps = {}
+  ) {
+    super(scope, id);
+
+    this.label = label;
+    this.refreshRate = props.refreshRate?.seconds;
+
+    if (App.of(this)._target === "sim") {
```
eladb, line 126: Eventually this won't be only for sim I believe

libs/wingsdk/src/ui/visual.ts
eladb: Sometimes it's called `Widget`

eladb (approved):
Wow wow wow!! Yes!

This is going to be so powerful!

libs/wingsdk/src/core/tree.ts
Chriscbr: Mmm-  I was thinking of skipping these for non-sim targets since Wing Console doesn't work on production apps today, so it would just be creating lots of extra resources. But I suppose perhaps it's a preoptimization - I'll take it out the condition for now

libs/wingsdk/src/core/tree.ts
Chriscbr: What would be the API for UILabel and UILink?

libs/wingsdk/src/ui/section.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Adds a visual component to the section. The components
+   * will be rendered in the order they were added.
+   */
+  public add(...components: Array<VisualComponent>): void {
+    for (const component of components) {
+      if (component._newParent !== undefined) {
+        throw new Error(
+          `Cannot add ${component.node.path} to ${this.node.path} because it already belongs to ${component._newParent}.`
+        );
+      }
```
Chriscbr, line 62: this is a little hack. Basically, I thought it be nice to avoid boilerplate and allow folks to define buttons and labels and other stuff at a top level, in case they don't want to use `Section`:

```js
let button1 = new ui.Button();
let button2 = new ui.Button();
```

But the issue is if you do use a section:

```js
let section = new ui.Section();
section.add(button1, button2);
```

then "button1" and "button2" would now get rendered in both places (as a child of Section, and at the top level). To work around this, I added a field to keep track of whether each component was added to a grouping component.

PR #4608, opened 2023-10-20
feat(cli): add support for `.env` files

### LINKED ISSUE #4595: Support dotenv files from the Wing CLI
### Feature Spec

Wing uses [dotenv](https://github.com/motdotla/dotenv) to load additional environment variables from the following files in your current working directory:

```
.env                # loaded in all cases
.env.local          # loaded in all cases, ignored by git
.env.[mode]         # only loaded in specified mode (compile, run|it, test, lsp)
.env.[mode].local   # only loaded in specified mode, ignored by git
```

These environment variables are available in Wing code using `util.env` and `util.tryEnv`.

### Use Cases

Environment files are an industry standard and a convenient way to configure your local application while working on it.

### Implementation Notes

_No response_

### Component

CLI

### Community Notes

garysassano:
Closes #4595

## Checklist

- [x] Tests added (always)


skyrpex:
Nice job! You may want to take a look at how Vite does it: https://github.com/vitejs/vite/blob/4cedcdc99ca8ad38a67f46cf94dd3be65a567955/packages/vite/src/node/env.ts#L4.

They use `dotenv-expand` which allows reusing env variables within the `.env` files, which is quite common.

garysassano:
@skyrpex
1) Are base env files (`.env`/`.env.local`) and mode env files (`.env.[mode]` /`.env.[mode].local`) mutually exclusive, or could both be loaded at the same simultaneously? In that case, what should be the behaviour for env variables that appear on multiple files? Should there be a precedence order?
2) What should be the behaviour for env variables that already exist in your shell but you also specified in an env file? Should we have something like `--env-overwrite` as global option or it's just the standard behaviour to always overwrite?

garysassano:
@skyrpex
1) Are base env files (`.env`/`.env.local`) and mode env files (`.env.[mode]` /`.env.[mode].local`) mutually exclusive, or could both be loaded simultaneously? In that case, what should be the behaviour for env variables that appear on multiple files? Should there be a precedence order?
2) What should be the behaviour for env variables that already exist in your shell but you also specified in an env file? Should we have something like `--env-overwrite` as global option or it's just the standard behaviour to always overwrite?

staycoolcall911:
@garysassano when this is ready for review, can you please ask a review from @skyrpex?

garysassano:
@staycoolcall911 It's already working, there's just a problem with hangar not being able to load the dotenv files since it runs from a different directory.

staycoolcall911:
I see @garysassano, thanks for the update. Please let us know if you need additional assistance with fixing that.

garysassano:
@staycoolcall911 I tryed to copy the dotenv files to `tools/hangar` but didn't fix the problem. I'm not actually sure how this could be solved.

staycoolcall911:
Ok, we'll get you some help. Might take a couple of days

apps/wing/src/cli.ts
```diff
@@ -178,12 +128,18 @@ async function main() {
       void import("./commands/upgrade").then((m) => m.checkForUpdates());
     }
   }
+  function loadEnv(cmd: Command) {
+    loadEnvVariables({
+      mode: cmd.args[0],
```
garysassano, line 133: Do `wing it` and `wing run` load from distinct environment files even if it's the same mode?

MarkMcCulloh: Good point, I oversimplified it

docs/docs/06-tools/01-cli.md
```diff
@@ -222,3 +222,24 @@ $ wing pack
 This will compile your current Wing directory, and bundle it as a tarball that can be published to [GitHub packages](https://github.com/features/packages) or [npm](https://www.npmjs.com/).
 
 See [Libraries](../05-libraries.md) for more details on packaging and consuming Wing libraries.
+
+## Environment Variables
+
+Wing uses `dotenv` to load additional environment variables from files in your current working directory. These environment variables can be accessed in Wing code using `util.env` and `util.tryEnv`.
```
Chriscbr, line 228: Should we maybe adjust the documentation to avoid assuming prior knowledge about the npm library? I see the phrase "dotenv" used but it's not clear to me if it's some well-defined standard (if it is then I'm cool with just linking that). It might be a better user experience if we explain how it works and how a `.env` file looks. The other upside is it keeps our API promise more stable in case we change to using a library other than `dotenv`, or rewrite the CLI in Rust etc.

```suggestion
For development and testing, Wing can automatically read environment variables from `.env` files in your current working directory. These environment variables can be accessed in Wing code using `util.env` and `util.tryEnv`.
```

apps/wing/src/cli.ts
garysassano: I think you actually did a great job. Thanks to you, I could find a simpler way to load the env variables and make sure that also using `wing it` sets the right mode.

MarkMcCulloh (approved):

Issue #4503, opened 2023-10-10
Show resolve paths as hints in compiler errors


Chriscbr:
### Feature Spec

When an error refers to a relative path (in a bring or extern statements):

```
error: Cannot find module "./blah.w"
  --> main.w:2:7
  |
2 | bring "./blah.w" as another;
  |       ^^^^^^^^^^ Cannot find module "./blah.w"
```

We would like to show the error the path that the compiler resolved it to as a hint, maybe something like:

```
error: Cannot find module "./blah.w"
  --> main.w:2:7
  |
2 | bring "./blah.w" as another;
  |       ^^^^^^^^^^ Cannot find module "./blah.w"
  = note: Expected to find a file at "/path/to/blah.w"
```

### Use Cases

Better debugging experience

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

PR #4609, opened 2023-10-20
fix(compiler): phase mismatch not stated explicitly

### LINKED ISSUE #2230: inflight/preflight type mismatch error ergonomic improvements
### I tried this

Implement `IFunctionHandler` but forgot to indicate that `handle` is `inflight`.
Used this code:
```ts
bring cloud;

class Handler impl cloud.IFunctionHandler {
  pub handle(event: str) {

  }
}
```
Instead of this (forgot to add the `inflight` modifier on the `handle` method):
```ts
bring cloud;

class Handler impl cloud.IFunctionHandler {
  pub inflight handle(event: str) {

  }
}
```

### I expected this:

If the types only differ in their inflight / preflight - the error should say it explicitly

### Instead, this happened
```sh
error: Expected type to be "inflight (event: str): void", but got "preflight (event: str): void" instead
  --> wing/main.w:3:7
  |
3 | class Handler impl cloud.IFunctionHandler {
  |       ^^^^^^^ Expected type to be "inflight (event: str): void", but got "preflight (event: str): void" instead
```

### Is there a workaround?

N/A

### Component

Compiler

### Wing Version

0.33.1

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

Lancear:
Closes #2230 

- Adds the phase mismatch hint when none of the expected types match the phase that was given

## Checklist

- [x] Tests added (always)


Lancear:
will add tests tomorrow

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -793,14 +793,14 @@ Duration <DURATION>"
 `;
 
 exports[`cloud_function_expects_inflight.test.w 1`] = `
-"error: Expected type to be \\"inflight (event: str): void\\", but got \\"preflight (name: str): str\\" instead
+"error: Expected phase to be preflight, but got inflight instead
```
Chriscbr, line 796: Looks like the error message got swapped around - it should say that the expected phase is "inflight"

examples/tests/invalid/phase_mismatch.test.w
```diff
@@ -0,0 +1,13 @@
...
+
+let my_func2 = (callback: (num): num) => {  };
+my_func2(inflight (n: num): num => { return n * 2; });
+
+interface IMyHandlerInterface1 {
+  inflight handle(n:  num): num;
+}
+
+let my_func3 = (callback: IMyHandlerInterface1) => {  };
+my_func3((n: num): num => { return n * 2; });
```
Chriscbr, line 12: What happens if the phases don't match and the argument types are incompatible? (can we add a test?)

examples/tests/invalid/phase_mismatch.test.w
Chriscbr, line 1: What happens if the phase mismatch is not the "top-most" error? for example:

```js
bring cloud;

let accepts_callback = (callback: (inflight (str): str): void): void => {
  callback(inflight (msg: str): str => {
    return "Hi " + msg + "!";
  });
};

let callback = (inner_fn: ((str): str): void): void => {};

accepts_callback(callback); // error
```

On the last line, the expected type is `preflight (inflight (str): str): void` but the function passed was type `preflight (preflight (str): str): void`.

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr, line 796: I feel a bit mixed on this new error because we're subtracting information away from the user. IMO we should try to find some way to highlighting the phase mismatch while still showing the expected and found types.

Two ideas that come to mind:

1) add on to the beginning of the error

```
error: Mismatched function phase; expected type to be "inflight (event: str): void", but got "preflight (name: str): str" instead
```

2) add the information as a separate error hint (see https://github.com/winglang/wing/issues/4503):

```
error: Expected type to be "inflight (event: str): void", but got "preflight (name: str): str" instead
  <code excerpt>
  hint: Expected phase to be inflight, but got preflight instead
```

What do you think about these alternatives?

(cc @MarkMcCulloh about error message style)

Chriscbr (changes requested):

tools/hangar/__snapshots__/invalid.ts.snap
Lancear: nice catch

Lancear: I think this would work great as a hint to the existing type mismatch error 👍

examples/tests/invalid/phase_mismatch.test.w
Lancear: currently it would result in a type mismatch. phase mismatch is currently checked as "would the type match when we dont look at the phase"  instead of checking if the phases mismatch. Especially when we go the hint route we could simply check if the phase mismatch and add the hint, would make the check also easier.

tools/hangar/__snapshots__/invalid.ts.snap
staycoolcall911: Hint sounds great.

examples/tests/invalid/phase_mismatch.test.w
Lancear: changed this to always show the hint when the expected types include no function with the given phase, thus also if the argument types mismatch. added a test for this.

examples/tests/invalid/phase_mismatch.test.w
Lancear: only checked if top-most error atm. added this case to the tests

tools/hangar/__snapshots__/invalid.ts.snap
Lancear: switched it to a hint like the others around it, now looking into the cli/lsp changes that need to be done for #4503

Lancear: added hints to diagnostics and tried to update the cli and lsp accordingly

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2335,11 +2339,13 @@ Duration <DURATION>"
...
 exports[`nil.test.w 1`] = `
-"error: Expected type to be \\"str\\", but got \\"nil\\" instead (hint: to allow \\"nil\\" assignment use optional type: \\"str?\\")
+"error: Expected type to be \\"str\\", but got \\"nil\\" instead
   --> ../../../examples/tests/invalid/nil.test.w:3:14
   |
 3 | let x: str = nil;
-  |              ^^^ Expected type to be \\"str\\", but got \\"nil\\" instead (hint: to allow \\"nil\\" assignment use optional type: \\"str?\\")
+  |              ^^^ Expected type to be \\"str\\", but got \\"nil\\" instead
+  |
+  = hint: to allow \\"nil\\" assignment use optional type: \\"str?\\"
```
Chriscbr, line 2348: Thank you for updating these other spots in the code as well!

Chriscbr (approved):
Great work @Lancear! LGTM

polamoros (approved):

PR #4613, opened 2023-10-21
chore(cli): fix argument description for `wing run`

garysassano:
Aligning argument description for `wing run` to the docs.

### DIFF
--- a/apps/wing/src/cli.ts
+++ b/apps/wing/src/cli.ts
@@ -123,7 +123,7 @@ async function main() {
     .command("run")
     .alias("it")
     .description("Runs a Wing program in the Wing Console")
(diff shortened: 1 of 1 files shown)


apps/wing/src/cli.ts
```diff
@@ -123,7 +123,7 @@ async function main() {
     .command("run")
     .alias("it")
     .description("Runs a Wing program in the Wing Console")
-    .argument("[entrypoint]", "program .w entrypoint")
+    .argument("[entrypoint]", "program .w or .wsim entrypoint")
```
eladb, line 126: I am not sure this is something we want to (or need to) support long term so I wouldn't advertise it.

PR #4614, opened 2023-10-21
fix(compiler): optional params allowed before non-optionals

### LINKED ISSUE #3129: Function arguments: compiler should throw an error if optional arguments appear before non-optionals
The [language reference](https://www.winglang.io/docs/language-guide/reference#1614-function-arguments) defines the following behavior:

Non-optional arguments can only be used before all optional arguments:

```TS
let myFun = (a: str, x: num?, y: str): void => { /* ... */ };
//-----------------------------^^^^^^ ERROR: cannot declare a non-optional argument after an optional
```

However this currently compiles without an error. We should add a check in the type checker that enforces the described behavior.

Lancear:
Closes #3129

As variadic args are kind of optional by default (would simply result in an empty collection), I allowed those after optional parameters. Not sure if a similar case exists with structs with all optional fields 🤔 

## Checklist

- [x] Tests added (always)


examples/tests/invalid/function_type.test.w
```diff
@@ -12,3 +12,6 @@ interface IFace {
   inflight my_method2(x: num);
 //         ^^^^^^^^^^^^^^^^^^ Expected method return type
 }
+
```
eladb, line 15: Could you add please a positive tests for the variadic case?

Lancear: yes, will do

Lancear: added

MarkMcCulloh (approved):

PR #4602, opened 2023-10-19
feat(cli): adding unsupported resource indication to tests

tsuf239:
`wing test -t tf-azure  examples/tests/sdk_tests/function/*.test.w`
image

`wing test -t tf-azure  examples/tests/sdk_tests/counter/*c.test.w`
image

`wing test -t tf-gcp  examples/tests/sdk_tests/counter/*c.test.w`
image

## Checklist
 fixes #4573 
Detects unsupported resources only, working now to support an indication for unsupported methods too
- [x] Title matches [Winglang's style guide](https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted)
- [x] Description explains motivation and solution
- [ ] Tests added (always)
- [ ] Docs updated (only required for features)
- [ ] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Wing Cloud Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.

PR #4624, opened 2023-10-22
feat(cli): unsupported inflight method indication

tsuf239:
## Checklist
partially fixes: #4602 
Right now indicates only Azure's bucket inflight methods. Please let me know what you think regarding the implementation and I'll the rest of the clients for the rest of the targets. 
**adding tests soon**

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/std/resource.ts
+++ b/libs/wingsdk/src/std/resource.ts
@@ -1,6 +1,7 @@
 import { Construct, IConstruct } from "constructs";
 import { Duration } from "./duration";
 import { App } from "../core";
+import { NotImplementedError } from "../core/errors";
 import { liftObject } from "../core/internal";
 import { log } from "../shared/log";
 import { Node } from "../std";
@@ -174,6 +175,10 @@ export abstract class Resource extends Construct implements IResource {
   }
 
   private readonly onLiftMap: Map<IInflightHost, Set<string>> = new Map();
+  /**
+   * @internal
+   */
+  protected _clientClass?: any; // to assign the client class
 
   /** @internal */
   public abstract _getInflightOps(): string[];
@@ -227,9 +232,12 @@ export abstract class Resource extends Construct implements IResource {
     // For each operation, check if the host supports it. If it does, register the binding.
     const supportedOps = [...(this._getInflightOps() ?? []), "$inflight_init"];
     for (const op of ops) {
-      if (!supportedOps.includes(op)) {
-        throw new Error(
-          `Resource ${this.node.path} does not support inflight operation ${op} (requested by ${host.node.path})`
+      if (
+        !supportedOps.includes(op) ||
+        (this._clientClass && !this._clientClass.prototype.hasOwnProperty(op))
+      ) {
+        throw new NotImplementedError(
+          `Resource ${this.node.path} does not support inflight operation ${op} (requested by ${host.node.path}).\nIt might not be implemented yet.`
         );
       }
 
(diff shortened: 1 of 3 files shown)


libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -5,16 +5,10 @@ import {
...
-  ObjectMetadata,
-  BucketPutOptions,
-  BucketDeleteOptions,
-  BucketSignedUrlOptions,
-} from "../cloud";
+import { IBucketClient, BucketPutOptions, BucketDeleteOptions } from "../cloud";
 import { Json } from "../std";
 
-export class BucketClient implements IBucketClient {
+export class BucketClient implements Partial<IBucketClient> {
```
eladb, line 11: It feels sketchy to use partial here. I think an explicit throw from each non implemented method is more readable

libs/wingsdk/src/target-tf-azure/bucket.ts
```diff
@@ -52,6 +54,7 @@ export class Bucket extends cloud.Bucket {
   public readonly storageContainer: StorageContainer;
   private readonly public: boolean;
   private readonly storageAccount: StorageAccount;
+  protected _clientClass = BucketClient;
```
eladb, line 57: Why is this needed?

libs/wingsdk/src/std/resource.ts
```diff
@@ -225,11 +230,16 @@ export abstract class Resource extends Construct implements IResource {
...
-      if (!supportedOps.includes(op)) {
-        throw new Error(
-          `Resource ${this.node.path} does not support inflight operation ${op} (requested by ${host.node.path})`
+      if (
+        op !== "$inflight_init" &&
+        (!supportedOps.includes(op) ||
+          (this._clientClass &&
+            !this._clientClass.prototype.hasOwnProperty(op)))
+      ) {
+        throw new NotImplementedError(
```
eladb, line 241: Why do we need a specific exception type?

eladb (changes requested):

libs/wingsdk/src/target-tf-azure/bucket.ts
tsuf239: When going over the bound host operations, we check if this client class implements all of the used operations. This keeps us from saving a dedicated list of non-implemented methods (that will have to be updated manually for each change- had to maintain). This approach introduces 0 maintenance- in the price of having the client class implement `Partial<ClientInterface>` and having the client class specified here. (I added tests that can show easily how it's done)

libs/wingsdk/src/std/resource.ts
tsuf239: it gets caught by the cli, this way we know if to show failed, or unsupported :)

libs/wingsdk/src/target-tf-azure/bucket.ts
tsuf239: if you have another idea, please let me know :)

tsuf239: another option I can think of is to implement a different list of `_getInflightOps` in each of the targets (keeping the parent's one abstract maybe), which still have some maintenance

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -54,23 +54,7 @@ export abstract class Bucket extends Resource {
...
-      BucketInflightMethods.PUBLIC_URL,
-      BucketInflightMethods.EXISTS,
-      BucketInflightMethods.TRY_GET,
-      BucketInflightMethods.TRY_GET_JSON,
-      BucketInflightMethods.TRY_DELETE,
-      BucketInflightMethods.SIGNED_URL,
-      BucketInflightMethods.METADATA,
-    ];
-  }
+  public abstract _getInflightOps(): string[];
```
eladb, line 57: Maybe rename to `_supportedOps(): string[]`?

eladb (changes requested):

PR #4654, opened 2023-10-23
feat(sdk): support multiple test simultaneously on awscdk

### LINKED ISSUE #3750: Add support for running more than one test simultaneously on awscdk
### Feature Spec

following https://github.com/winglang/wing/pull/3723
it seems like the test mechanism for awscdk doesn't support running multiple tests at once, 
for example, when running: `wing test -t awscdk examples/tests/sdk_tests/function/*.w ` (runs 4 tests at once)
it compiles well, but then throws many of those errors:
```
Command failed: cdk deploy --require-approval never --ci true -O ./output.json --app . 

 ❌  stack failed: AlreadyExistsException: ChangeSet cdk-deploy-change-set cannot be created due to a mismatch with existing attribute ClientToken
    at Request.extractError (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:46430)
    at Request.callListeners (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:89985)
    at Request.emit (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:89433)
    at Request.emit (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:196191)
    at Request.transition (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:189743)
    at AcceptorStateMachine.runTo (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:154615)
    at /opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:154945
    at Request.<anonymous> (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:190035)
    at Request.<anonymous> (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:196266)
    at Request.callListeners (/opt/homebrew/lib/node_modules/aws-cdk/lib/index.js:321:90153) {
  code: 'AlreadyExistsException',
  time: 2023-08-09T09:28:51.969Z,
  requestId: 'dd2caf07-0b6a-402a-801a-e660215766ae',
  statusCode: 400,
  retryable: false,
  retryDelay: 109.94840449062337
}
```

### Use Cases

testing

### Implementation Notes

_No response_

### Component

CLI

### Community Notes

marciocadev:
When creating a stack with the AWS CDK, it's necessary to set the environment variable CDK_STACK_NAME. This variable is causing a problem when trying to execute multiple tests simultaneously.

The solution I found was to capture this variable and concatenate it with the filename. This way, each stack will be named for deployment when running tests in the format:

CDK_STACK_NAME + "-" + file name

Closes #3750 

## Checklist

- [ ] Tests added (always)


apps/wing/src/commands/test/test.ts
```diff
@@ -255,10 +255,18 @@ async function testAwsCdk(synthDir: string, options: TestOptions): Promise<std.T
...
+      if (match) {
+        // remove special characters from file name
+        stackName += "-" + match[1].replace(/[^a-zA-Z0-9]/g, "");
+      }
+
       const testArns = await awsCdkOutput(
         synthDir,
         ENV_WING_TEST_RUNNER_FUNCTION_IDENTIFIERS_AWSCDK,
-        process.env.CDK_STACK_NAME!
+        stackName
```
tsuf239, line 269: what if we have two files of the same name? (in different folders) would it be better to randomize it as we do to the rootId? (we can also use only valid letters in the randomization, keeping it simpler :) )

marciocadev: Tsuf, the problem is that I cannot use a randomized value because I need to obtain this value in two different places, both in `wingsdk/target-awscdk/app` and in `wing/command/test`

A second problem is that if I structure it like this:

image

The `env.test.w` is tested, but the tests in the `folder` do not run independently of the target (tf-aws, awscdk).

tsuf239: Maybe randomize the stack-name on the test command, then pass it as an env var?
https://github.com/winglang/wing/blob/64d830eae5a4c8c2d6588c7c879c3fd88df6be9c/apps/wing/src/commands/test/test.ts#L99 (like we did to the rootId here)
https://github.com/winglang/wing/blob/54f03f7c0be88bf49b3e4e61bcb7196d3b04b9ed/libs/wingcompiler/src/compile.ts#L127-L129 (here it's added to the `process.env` of the context we run the preflight on)
and be put here as a prop: https://github.com/winglang/wing/blob/add9ab68be3aa49c13944a4148f5229e85ddf83f/libs/wingc/src/jsify.rs#L188

tsuf239: yes, it can work :) 
we can also hash the absolute path to the wing entry file

tsuf239 (approved):
image
a few multiple tests are passing! awesome job!

PR #4657, opened 2023-10-23
fix(compiler): jsii import use-before-defined error

### LINKED ISSUE #4575: Can not bring two extern files that bring utils
### I tried this:

```
// main.w

bring "./class1.w" as class1;
bring "./class2.w" as class2;
```

```
// class1.w

bring cloud;
bring ex;
bring util;
```

```
// class2.w

bring util;

class Class2 {
  init() {
    let nanId = util.nanoid();
  }
}
```

### This happened:

```
Failed to compile.

error: Symbol "util" used before being defined
  --> apps/wing-console/console/app/demo/class2.w:5:17
  |
5 |     let nanId = util.nanoid();
  |                 ^^^^ Symbol "util" used before being defined
  |
  --> apps/wing-console/console/app/demo/class1.w:3:7
  |
3 | bring util;
  |       ---- defined later here
```

### I expected this:

to work

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

Chriscbr:
Fixes #4575

This was a tricky bug to pin down. I initially thought it might be an issue with our symbol lookup code, but it turns out it was working fine - it instead there was an issue in our symbol environment construction. In `class2.w`, the `util` symbol had the was being stored with the span and statement index information associated with `class1.w`.

The root cause of this seems to be that after the first `bring util` statement, some of the import information is stored in `JsiiImportSpec`, which then got reused when `bring util` was called a second time. To fix this, I updated the logic we perform for bring-jsii statements so that it only reuses an existing `JsiiImportSpec` if the `alias` it has stored is exactly the same (both the string and its position in the source code).

I'm not sure if storing a separate `JsiiImportSpec` for every unique bring statement has any performance drawbacks but I didn't have enough context to easily refactor the `alias` and `import_statement_idx` fields out of `JsiiImportSpec`:

https://github.com/winglang/wing/blob/392aa4bdce6868ad01ac5df674a7decb0a823f21/libs/wingc/src/type_check/jsii_importer.rs#L51-L69

Definitely interested if there are ways to simplify this though.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/turbo.json
+++ b/libs/wingc/turbo.json
@@ -11,7 +11,7 @@
       ]
     },
     "test": {
-      "dependsOn": ["compile", "@winglang/sdk#compile", "examples-valid#topo", "examples-invalid#topo"]
+      "dependsOn": ["compile", "@winglang/sdk#compile", "examples-valid#topo", "examples-invalid#topo", "jsii-fixture#compile"]
     },
     "dev": {
       "dependsOn": ["@winglang/sdk#compile"]
--- a/libs/wingc/turbo.json
+++ b/libs/wingc/turbo.json
@@ -11,7 +11,7 @@
       ]
     },
     "test": {
-      "dependsOn": ["compile", "@winglang/sdk#compile", "examples-valid#topo", "examples-invalid#topo", "jsii-fixture#compile"]
+      "dependsOn": ["compile", "@winglang/sdk#compile", "examples-valid#topo", "examples-invalid#topo"]
     },
     "dev": {
       "dependsOn": ["@winglang/sdk#compile"]
--- a/apps/wing/src/commands/pack.ts
+++ b/apps/wing/src/commands/pack.ts
@@ -25,9 +25,7 @@ export async function pack(options: PackageOptions = {}): Promise<string> {
   // check that the library compiles to the "sim" target
   console.log('Compiling to the "sim" target...');
 
-  // TODO: workaround for https://github.com/winglang/wing/issues/4431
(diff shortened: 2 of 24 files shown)


libs/wingc/turbo.json
```diff
@@ -11,7 +11,7 @@
       ]
     },
     "test": {
-      "dependsOn": ["compile", "@winglang/sdk#compile", "examples-valid#topo", "examples-invalid#topo"]
+      "dependsOn": ["compile", "@winglang/sdk#compile", "examples-valid#topo", "examples-invalid#topo", "jsii-fixture#compile"]
```
MarkMcCulloh, line 14: Are you sure this is needed? `examples-valid#topo` is already dependent on `jsii-fixture#compile`

MarkMcCulloh (approved):
Nice catch 🎉 

> I'm not sure if storing a separate JsiiImportSpec for every unique bring statement has any performance drawbacks but I didn't have enough context to easily refactor the alias and import_statement_idx fields out of JsiiImportSpec

It shouldn't be big perf impact (the underlying jsii data is shared via `TypeSystem`), but yeah that whole system needs a refactor with multi-file projects in mind.

eladb (approved):

libs/wingc/turbo.json
Chriscbr: Ah good catch - I think I might have just ran the wrong commannd...

PR #4633, opened 2023-10-22
feat(sdk): add `Bucket.metadata()` for `tf-gcp` target

garysassano:
Closes #4332

## Checklist

- [x] Title matches [Winglang's style guide](https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted)
- [x] Description explains motivation and solution
- [x] Tests added (always)
- [ ] Docs updated (only required for features)
- [ ] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Wing Cloud Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.

PR #4663, opened 2023-10-24
fix(sdk): bucket `put()` and `putJson()` don't set right content-type for `tf-gcp` target

garysassano:
Fixing the issue described [here](https://github.com/winglang/wing/pull/4633#issuecomment-1777107824).


libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -47,20 +49,34 @@ export class BucketClient implements IBucketClient {
...
+  public async put(
+    key: string,
+    body: string,
+    opts?: BucketPutOptions
+  ): Promise<void> {
+    const options = {
+      contentType:
+        (opts?.contentType ?? mime.lookup(key)) || "application/octet-stream",
+    };
+    await this.bucket.file(key).save(body, options);
```
Chriscbr, line 67: Can we add update the unit tests [here](https://github.com/winglang/wing/blob/main/libs/wingsdk/test/shared-gcp/bucket.inflight.test.ts)?

I think we essentially want a test that checks if given someone called `put()` in our `Bucket` class (the one defined in this file) without specifying a `contentType`, then `save()` should have been called with "application/octet-stream" in its options.

garysassano: Yeah, I think that makes sense. I honestly was surprised to see that the Node.js SDK for GCP automatically infers the type for known extensions, but not for unknown ones. Yet, if you upload files manually to Cloud Storage from the Google Cloud Console, it automatically sets the type to `application/octet-stream` if it doesn't recognize the file type.

garysassano: Yeah, I think that makes sense. I honestly was surprised to see that the Node.js SDK for GCP automatically sets the type for known extensions, but not for unknown ones. Yet, if you upload files manually to Cloud Storage from the Google Cloud Console, it automatically sets the type to `application/octet-stream` if it doesn't recognize the file type.

garysassano: Apparently you cannot `vi.spyOn()` on the methods that are called internally by `BucketClient`, so the only solution I see for adding the test would be to merge #4633 so that you can call `put()` followed by `metadata()` to check that the Content-Type is indeed `application/octet-stream` when uploading an object with unknown extension.

garysassano:
Keep.

revitalbarletz:
@garysassano can you update what is the status of this PR?

garysassano:
@revitalbarletz It's functional as of now, but @tsuf239 preferred to hold off on merging it until #4905 is complete.

garysassano:
@tsuf239 Could you approve this? I need it to finish #4633 otherwise the test fails.

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -53,20 +55,34 @@ export class BucketClient implements IBucketClient {
...
+   * @param key Key of the object
+   * @param body Json object
+   */
   public async putJson(key: string, body: Json): Promise<void> {
-    try {
-      await this.put(key, JSON.stringify(body, null, 2));
-    } catch (error) {
-      throw new Error(`Failed to put JSON object. (key=${key})`);
-    }
+    await this.put(key, JSON.stringify(body, null, 2), {
```
revitalbarletz, line 83: Why are we removing the try/catch logic?

garysassano: My rationale was probably to make it consistent with the [AWS implementation](https://github.com/winglang/wing/blob/693ee4b3a995a9c305479b32c10e0a87013ff125/libs/wingsdk/src/shared-aws/bucket.inflight.ts#L77). I believe error handling is not necessary here because the operation involves either creating a new object or overwriting an existing one.

garysassano: To make it consistent, we might consider introducing methods like `tryPut()` and `tryPutJson()`, akin to `tryGet()` and `tryGetJson()`, for situations where we need to manage failures directly.

garysassano: For the sake of consistency, we might consider introducing methods like `tryPut()` and `tryPutJson()`, akin to `tryGet()` and `tryGetJson()`, for situations where we need to manage failures directly.

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
tsuf239: Sounds strange,
is it because they are private? (you can easily bypass by `@ts-expect-error`, there are other solutions on the internet as well)
or there is another problem?

tsuf239 (commented):
I wonder if it will be ok to merge this branch to the metadata one, since you own both of them, so we can test everything is working properly before using it to main, WDYT?

garysassano:
@tsuf239 I don't get it. What sounds strange?

FYI, I've a lready merged this branch to metadata's branch (see #4633), run `wing test `on it and everything works.

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
Chriscbr: Got it - yeah this change makes sense to me. Unifying error messages to be more similar between targets could be a nice project in the future but the current code looks like it's discarding the thrown `error` which is not great for debugging, so I'm okay with just letting any errors bubble up

garysassano:
I'm closing this PR since the fix was merged into #4633 a few days ago.

PR #4665, opened 2023-10-24
fix(vscode): invalid completions for types that look like builtins

MarkMcCulloh:
Previously, text like `String.` would always show completions as if it were `str.`, which is incorrect. Additionally, I noticed that completions for types would not show up inside the brackets of types like `Array<>` so that's fixed as well.

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -1651,7 +1651,7 @@ impl Types {
 		}
 	}
 
-	/// Given an unqualified type name of a builtin type, return the full type info.
+	/// Given a builtin type, return the full class info from the standard library.
 	///
 	/// This is needed because our builtin types have no API.
 	/// So we have to get the API from the std lib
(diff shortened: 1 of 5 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -1651,18 +1651,44 @@ impl Types {
...
 	/// So we have to get the API from the std lib
 	/// but the std lib sometimes doesn't have the same names as the builtin types
 	/// https://github.com/winglang/wing/issues/1780
 	///
 	/// Note: This doesn't handle generics (i.e. this keeps the `T1`)
-	pub fn get_std_class(&self, type_: &str) -> Option<(&SymbolKind, symbol_env::SymbolLookupInfo)> {
-		let type_name = fully_qualify_std_type(type_);
+	pub fn get_std_class(&self, type_: &TypeRef) -> Option<(&SymbolKind, symbol_env::SymbolLookupInfo)> {
+		let type_name = match &**type_ {
+			Type::Number => "Number",
```
eladb, line 1664: As a future refactor, would it be possible to avoid the hard coding of all std types in the compiler and simply use the type information from the `std` namespace in the sdk?

@yoav-steinberg ?

eladb (approved):

libs/wingc/src/type_check.rs
MarkMcCulloh: Long time ago I made [this issue](https://github.com/winglang/wing/issues/1780) and we've also talked about this during a team meeting.
Would love this refactor but it's definitely not a low-hanging fruit.

PR #4666, opened 2023-10-24
feat(cli): add `--retry [retries]` to `wing test`

### LINKED ISSUE #3796: adding a `retry` option to the cli test command
### Feature Spec

Adding a `retry` option to the cli test command to re-run the failed tests for a specific amount of times.

### Use Cases

Mostly for automated tests that tend to fail over environmental factors rather than the code that is tested. It can save plenty of time, since now we're repeating the command, it runs again all the tests- including the successful ones.

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

garysassano:
Closes #3796

## Checklist

- [x] Tests added (always)


tools/hangar/__snapshots__/test_corpus/valid/print.test.w_test_sim.md
```diff
@@ -8,6 +8,8 @@ pass ┌ print.test.wsim » root/env0/test:log1
      │ inflight log 1.1
      └ inflight log 1.2
 pass ┌ print.test.wsim » root/env1/test:log2
+     │ inflight log 1.1
+     │ inflight log 1.2
```
MarkMcCulloh, line 12: Looks like the logs are getting messed up now that the tests are running in parallel.

Maybe keep them running serially for this PR and work on parallelizing on a different one.

apps/wing/src/cli.ts
```diff
@@ -191,6 +191,11 @@ async function main() {
       "-o, --output-file <outputFile>",
       "File name to write test results to (file extension is required, supports only .json at the moment)"
     )
+    .addOption(
+      new Option("-R, --retry [retries]", "Number of times to retry failed tests")
+        .preset(3)
```
MarkMcCulloh, line 196: Not sure if we want this as a default. During development iteration, test failures will be normal and retying them will just lead to noise.

garysassano: What do you mean "as a default"? That's only if you use the `--retry` option in the command.

garysassano: I could probably add an extra test for when the `--retry` parameter isn't passed, to check that it actually doesn't retry failed tests if you want.

garysassano: If you want, I could probably add an extra test for when the --retry parameter isn't passed, to check that it actually doesn't retry failed tests.

garysassano: If you want, I could probably add an extra test for when the `--retry` parameter isn't passed, to check that it actually doesn't retry failed tests.

garysassano: To sum it up:
- `wing test [...endpoint]` = 0 retries
- `wing test [...endpoint] --retry` = 3 retries
- `wing test [...endpoint] --retry x` = x retries

MarkMcCulloh: Gotcha, I misread it, yeah then that makes sense

apps/wing/src/commands/test/test.ts
```diff
@@ -29,11 +29,23 @@ const generateTestName = (path: string) => path.split(sep).slice(-2).join("/");
...
+   */
   outputFile?: string;
-  /** String representing a RegEx used for test filtering. */
+  /**
+   * String representing a regex pattern used to selectively filter which tests to run.
+   */
   testFilter?: string;
+  /**
+   * How many times failed tests should be retried.
+   * @default 3
```
MarkMcCulloh, line 46: In the context of calling test(), the default is actually 0
```suggestion
   * @default 0
```

MarkMcCulloh (approved):

apps/wing/src/cli.ts
garysassano: For the sake of completeness, I've added a test for when the option is missing.

apps/wing/src/commands/test/test.ts
garysassano: You are right, if you call `test()` directly the default value would be indeed `0`. It's Commander that sets the value to `3` when calling the method through it.

garysassano: You are right, if you call `test()` directly the default value would be indeed `0`. It's Commander that sets the value to `3` or `x` when calling the method through it.

garysassano: Actually, there's no default value. If you call `test()` directly, the `retry` variable would be undefined. It's Commander that sets the value but that doesn't exist without it.

garysassano: It's the reason why I manually set the value to `0` here: https://github.com/winglang/wing/pull/4666/files#diff-cc35dcb1d186df3226515f587ca678d51e6c8e1af1ef38684e312d30bbcab5daR286

Issue #4667, opened 2023-10-25
Access clients of user-defined resources via simulator API


Chriscbr:
### Feature Spec

Today, the [simulator API](https://www.winglang.io/docs/concepts/simulator) only allows you to access a client for a preflight object if the object's class was implemented in the SDK:

```js
import { testing } from '@winglang/sdk';

const sim = new testing.Simulator({ simfile: "hello.wsim" });
const fn = sim.getResource("root/Default/cloud.Function");
const response = await fn.invoke("hello!");

```

If you try accessing the client corresponding to a class written in Wing, the simulator API will throw an error saying it's not available.

We would like to lift this restriction so that clients of all resources in the app can be accessed.

### Use Cases

For Wing Console features

### Component

Compiler, SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

ainvoner:
This feature also enables us to meet our basic requirements here:
https://github.com/winglang/wing/issues/4493

Chriscbr:
It looks like this isn't as easy as I first thought. My initial approach was to codegen `toSimulator()` for every preflight class, so that every resource in a Wing app gets listed in`simulator.json` when `wing compile -t sim` is run. That part is straightforward.

Now, how does the simulator know how to run a user defined resource? One idea is to have the simulator config point to the JavaScript code for that resource's inflight client. For example, if I have an app with a user-defined `Store` class and SDK-defined `Bucket` class, I can store a new field called `simulationPath` that refers to JS code for the client:

```json
    {
      "type": "@winglang/sdk.cloud.Bucket",
      "path": "root/Default/Store/cloud.Bucket",
      "simulationPath": "/Users/chrisr/dev/wing1/libs/wingsdk/lib/target-sim/bucket.inflight.js",
      "props": {
        "public": false,
        "initialObjects": {},
        "topics": {}
      },
      "attrs": {}
    },
    {
      "type": "Store",
      "simulationPath": "/Users/chrisr/dev/wing-test/target/example.main.wsim/.wing/store_c8ce3f27.js",
      "path": "root/Default/Store",
      "props": {},
      "attrs": {}
    },
```

[...]

Issue #4556, opened 2023-10-16
implement `cloud.Domain` for `sim` target


eladcon:
As we want to provide a Domain for simulator use cases, one of the options is to use a tunnels service. Such a service could be written in wing (see [POC](https://github.com/eladcon/wing-tunnels)) and give public access through a dedicated subdomain (e.g. mydomain.tunnels.wingcloud.io) to local services including `cloud.Api` or `cloud.Website`.

ainvoner:
@eladcon I guess we can close this one since it was implemented.
can you link the ngrok issue and close this one?

PR #4668, opened 2023-10-25
chore(sdk): refactor to store type information in simulator.json

Chriscbr:
Initial steps towards #4667.

When you compile to the `sim` target, info about the app's resources is stored in `simulator.json`. This PR updates the file to include metadata about each resource's type, such as the simulation's source code location. This eliminates an ugly factory class that we've had to update each time someone adds a new resource to the simulator. 🙌

Since the Wing Console and SDK are happily living together in our monorepo and packaged together as part of the CLI, I've also cleaned up some cross-library type information so there's just one fully-qualified name constant associated with each SDK resource instead of two (see libs/wingsdk/src/target-sim/schema-resources.ts for the removal).

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/domain.ts
```diff
@@ -16,4 +17,13 @@ export class Domain extends cloud.Domain {
...
   }
+
+  public toSimulator(): BaseResourceSchema {
+    return {
+      type: cloud.DOMAIN_FQN,
+      path: this.node.path,
+      props: {},
+      attrs: {},
+    };
+  }
```
MarkMcCulloh, line 28: Domain isn't currently supported in the sim, right?  Should this throw or something?

libs/wingsdk/test/core/__snapshots__/connections.test.ts.snap
```diff
@@ -56,10 +56,88 @@ return class Handler {
...
         },
-        "type": "wingsdk.cloud.Function",
+        "type": "@winglang/sdk.cloud.Function",
       },
     ],
     "sdkVersion": "0.0.0",
+    "types": {
+      "@winglang/sdk.cloud.Api": {
+        "className": "Api",
+        "sourcePath": "<ABSOLUTE PATH>/api.inflight.js",
```
MarkMcCulloh, line 66: Will this need to be resolvable across machines? If so it'd be useful for this to be a module-relative path instead. I think it's fine to iterate on this but just wanted to check to see how critical the portability of simulator.json is.

I guess we could eventually just plop an entire JSII manifest in here haha.

MarkMcCulloh (approved):

libs/wingsdk/test/core/__snapshots__/connections.test.ts.snap
Chriscbr: > Will this need to be resolvable across machines?

I think with the changes we have planned to allow the simulator running across multiple processes, there's still an assumption of a "master" process (the one which started the simulation) that is responsible for initializing all of the resources -- at the moment I think that's the only process that needs to resolve these paths. But it's a good question, I think we we'll revise this format many times in the future

libs/wingsdk/src/target-sim/domain.ts
Chriscbr: I wasn't sure - I see we have an open issue for adding support [here](https://github.com/winglang/wing/issues/4556), but it looks like the class was already added to the `target-sim` folder, and we already have docs about the simulator implementation on our [docs](https://www.winglang.io/docs/standard-library/cloud/domain#target-specific-details). @marciocadev should this be removed?

Chriscbr: In any case it doesn't seem like `Domain` has any inflight methods, so I'm okay leaving in some of this boilerplate for now for

Issue #3620, opened 2023-07-26
Wing Custom Platform Providers


eladb:
### Feature Spec

The concepts of *targets* and *compiler plugins* in Wing are now merged into a single concept called "platform library".

A Wing platform library is an npm module that adheres to a certain interface and determines how Wing programs will be synthesized and the concrete implementations for Wing cloud library resources.

Wing is shipped with a set of default platforms: `tf-aws`, `tf-azure`, `awscdk`, etc. But users can create custom platforms based on their team cloud requirements.

A wing platform library can control:

1. The concrete implementation of Wing Cloud Library resources (basically what "targets" used to determine).
2. Hook into the various compilation states and completely customize the synthesized output at different stages of the compilation (basically what "compiler plugins" used to determine).

A platform is an npm library that implements the following API:

```ts
interface IPlatform {
  newInstance?(type: string, scope: Construct, id: string, props: any): any;
  newApp?(): App;
  // ...current plugin API
}
```

When compiling a Wing program, you can specify any number of platform libraries using the `-p`
option. The platforms will get applied in *reverse order*, which means it is possible to stack them
together.

For example:

```console
$ npm i @winglang/tf-aws
$ npm i @winglang/tf-eks-k8s
$ wing compile -p tf-aws -p tf-eks-k8s main.w
```

The `tf-aws` platform implements the cloud APIs for AWS/Terraform. In this platform the default
implementation for `cloud.Service` is ECS/Fargate. However, the `tf-eks-k8s` platform overrides this
behavior (by implementing `newInstance()` for the `cloud.Service` type) to deploy cloud services on
EKS instead.

Platforms can also be composed together through code:

```ts
import { IPlatform } from '@winglang/sdk';
import aws from '@winglang/tf-aws';
import { EksService } '@winglang/tf-eks-k8s';

class MyPlatform implements IPlatform {
  public newInstance(type: string, scope: Construct, id: string, props: any): any {
    if (type == Service.TYPE) {
      return new EksService(type, scope, id, {
        ...props,

        // ...customize EKS...
      });

      return aws.newInstance(type, scope, id, props);
    }
  }
}
```

We should also add a way to configure platforms. For example, the `tf-aws` platform creates a VPC by default in certain cases (e.g. when using Redis), but it makes a lot of sense to allow users to connect to an existing VPC.

I think this can be achieved through some simple config file and/or command line switches that will be passed through to the platform provider.

### Use Cases

Allow full extensibility "below the abstraction line"

### Component

[...]

Issue #1474, opened 2023-02-07
Enable Plugins to Explicitly State Compatibility 


hasanaburayyan:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Want to allow plugins to explicitly state their compatibility since it seems to be a potentially common pattern that plugins are target specific. (tf-aws, tf-azure, tf-gcp, sim)

```java
exports.compatibileTargets = ["tf-aws", "sim"]

exports.validate = function(config) {
	// ...
}
```

if the plugin is not compatible with the target in a given context, then it could be ignores with a warning message:
"Skipping plugin x as it is not compatible with target y"

If a plugin does not explicitly state its compatibility maybe we just treat it as a universal plugin (if such thing exists :-) ) 

### Use Cases

- Allow plugin writers to explicitly state compatibility
- Gives users the insight into whether their plugins will have any effect on the resulting compilation.

hasanaburayyan:
This was not meant to be closed by plugin example PR. The examples only show how the compatibility might be declared.

monadabot:
Congrats! :rocket: This was released in [Wing 0.5.75](https://github.com/winglang/wing/releases/tag/v0.5.75).

hasanaburayyan:
Im going to repurpose this issue for Platform Compatibility

PR #4669, opened 2023-10-25
feat(platforms)!: support wing platforms

hasanaburayyan:
### New Features:
- `--platform` cli option accepts multiple platforms (can be BUILTIN platform, .js file, or path to lib folder)
- Wing Platforms: Allow use of builtin platforms such as sim , tf-aws , tf-azure , tf-gcp , awscdk (for now) or path to custom Platform

### BREAKING CHANGES:
- `--target` cli option has been removed
- `--plugins` cli option has been removed
- Wing Compiler Plugin System - Deprecated in place for new platform API
- Wing Targets - Deprecated in place for new platform API

### Summary
In this PR, the Wing Platform API is introduced. This allows users to make use of Platforms that are builtin and shipped with the wing installation, or make use of custom Platforms as described in this issue: https://github.com/winglang/wing/issues/3620

The big gist of this is the removal of the compiler plugin system and the need to run plugins within a sandbox vm, the removal of a single `target` in favor for multiple platforms that can be combined together for `ULTRA EXTENSIBILITY` !!!

#### IMPORTANT IMPLEMENTATION NOTE
Platforms are evaluated in the order in which they are provided, and ONLY the first platform to provide an implementation of `newApp()` will have its App returned. 

## TODO:
- [x] Update SOOOOOO many docs to reflect the removal of Plugins and Targets (probably keep some historical record)
- [x] Refactor CLI analytics to send data about platforms (instead of target)
- [x] Do another inspection to ensure all references to targets and plugins in code are references for platforms
- [x] Create followup issues for `newInstance()` and compatibility matching (will repurpose https://github.com/winglang/wing/issues/1474)

Closes: https://github.com/winglang/wing/issues/3620

## Checklist

- [x] Tests added (always)


hasanaburayyan:
@ekeren I need to as part of this PR update the CLI analytics to send data about Wing Platforms instead of the target. Since the user can provide custom platform targets (though unlikely could contain sensitive data) should we just report back builtin platform selections, and if its something else just report "custom-platform"

libs/wingsdk/src/platform/platform.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface IPlatform {
+  /**
+   * The model the platform is built for
+   *
+   * @example "sim"
+   */
+  readonly model: string;
+
+  // TODO: implement newInstance
+  // newInstance?(type: string, scope: Construct, id: string, props: any): any;
```
hasanaburayyan, line 16: Theres some additional complexity to get this API to work, due to the rapid development on Wing and the sheer size of the required changes for platforms I want to introduce this initial implementation so I can get out of rebase HELL

libs/wingsdk/src/platform/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./platform";
+export * from "./platform-manager";
```
hasanaburayyan, line 2: I dont have a great reason for putting these in their own namespace, might be worth just including it in core

tools/hangar/src/platform.test.ts
```diff
@@ -1,34 +1,35 @@
...
-    const targetDir = path.join(pluginsDir, "target", "main.tfaws");
+  describe("AWS target platform", () => {
+    const args = ["compile"];
+    const platforms = ["tf-aws"]
+    const targetDir = path.join(platformsDir, "target", "main.tfaws");
 
     test("permission-boundary.js", async () => {
-      const plugin = path.join(pluginsDir, "permission-boundary.js");
+      const plugin = path.join(platformsDir, "permission-boundary.js");
+      platforms.push(plugin);
```
Chriscbr, line 23: nit (avoid mutating variables shared between tests)
```suggestion
    const basePlatforms = ["tf-aws"]
    const targetDir = path.join(platformsDir, "target", "main.tfaws");

    test("permission-boundary.js", async () => {
      const plugin = path.join(platformsDir, "permission-boundary.js");
      const platforms = [...basePlatforms, plugin];
```

libs/wingsdk/src/target-tf-gcp/platform.ts
```diff
@@ -0,0 +1,16 @@
...
+export class TFGCPPlatform implements IPlatform {
+  /** Platform model */
+  public readonly model = "tf-gcp";
+
+  public newApp?(appProps: any): any {
+    return new App(appProps);
+  }
+}
+
+export const Platform = TFGCPPlatform;
```
Chriscbr, line 16: I think this can be
```suggestion
export class Platform implements IPlatform {
  /** Platform model */
  public readonly model = "tf-gcp";

  public newApp?(appProps: any): any {
    return new App(appProps);
  }
}
```

libs/wingsdk/src/platform/platform.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * New App Method
+   *
+   * @param appProps the props to pass to the app
+   */
+  newApp?(appProps: any): App;
```
Chriscbr, line 23:
```suggestion
  newApp?(appProps: AppProps): App;
```

libs/wingcompiler/src/compile.ts
```diff
@@ -281,7 +314,7 @@ async function runPreflightCodeInVm(
  * @param plugins list of plugin paths (absolute or relative)
  * @returns list of absolute plugin paths or relative to cwd, joined by ";"
  */
-function resolvePluginPaths(plugins: string[]): string {
+function resolvePlatformPaths(plugins: string[]): string {
```
Chriscbr, line 317: Update docstring

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+    const loadedPlatform = require(pathToRead);
+    if (!loadedPlatform || !loadedPlatform.Platform) {
+      console.error(`Failed to load platform from ${pathToRead}`);
+      return;
+    }
+
+    this.platformInstances.push(new loadedPlatform.Platform());
+  }
+
+  public createPlatformInstances() {
```
Chriscbr, line 40:
```suggestion
  private createPlatformInstances() {
```

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+  private readonly appProps: AppProps;
+  private readonly platformInstances: IPlatform[] = [];
+
+  constructor(options: PlatformManagerOptions) {
+    this.appProps = options.appProps;
+    this.platformPaths = options.platformPaths ?? [];
+    this.appProps;
+  }
+
+  private addPlatformPath(platformPath: string) {
```
Chriscbr, line 24: nit/optional
```suggestion
  private loadPlatformPath(platformPath: string) {
```

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -0,0 +1,85 @@
+import { basename, join } from "path";
+import { IPlatform } from "./platform";
+import { App, AppProps, SynthHooks } from "../core";
+
+interface PlatformManagerOptions {
+  readonly appProps: AppProps;
+  readonly platformPaths?: string[];
```
Chriscbr, line 7: We should note here that each value is either a path OR the name of a built-in platform

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      postSynthesize: [],
+      validate: [],
+    };
+
+    let firstAppFound = false;
+    this.platformInstances.forEach((instance) => {
+      if (instance.newApp && !firstAppFound) {
+        appCall = instance.newApp;
+        firstAppFound = true;
+      }
```
Chriscbr, line 61: Can we add a test for what happens if you pass two platforms that both implement `newApp`? (eg `wing compile --platform tf-aws --platform tf-gcp`)

Chriscbr (commented):
Looks awesome Hasan! 🔥 🔥 🔥  Just adding some small comments

ekeren:
> @ekeren I need to as part of this PR update the CLI analytics to send data about Wing Platforms instead of the target. Since the user can provide custom platform targets (though unlikely could contain sensitive data) should we just report back builtin platform selections, and if its something else just report "custom-platform"

Lets report the name of the platform and update the analytics page that this value is reported

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.createPlatformInstances();
+
+    let appCall: (props: AppProps) => App | undefined;
+    let synthHooks: SynthHooks = {
+      preSynthesize: [],
+      postSynthesize: [],
+      validate: [],
+    };
+
+    let firstAppFound = false;
```
hasanaburayyan, line 56: After some discussion with the team, It probably makes for a better model to expect that the first platform that is passed in MUST contain an App, instead of just searching for the first instance of one. 

This precedence should also help make the compatibility model easier to implement.

libs/wingsdk/src/core/app.ts
```diff
@@ -59,6 +59,32 @@ export interface AppProps {
...
+  readonly synthHooks?: SynthHooks;
+}
+
+/**
+ * Hooks for the synthesis process.
+ */
+export interface SynthHooks {
+  /**
+   * Hooks to be called before synthesizing the app.
+   */
```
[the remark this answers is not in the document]
hasanaburayyan: Right now the existing platforms call these in order of how they were received via command line but technically that order is the responsibility of the App so I dont think we should document some sort of guarantee. 

I could imagine an organization potentially being opinionated about what sort of hooks it allows.

apps/wing/src/cli.ts
```diff
@@ -140,11 +139,11 @@ async function main() {
     .description("Compiles a Wing program")
     .argument("[entrypoint]", "program .w entrypoint")
     .addOption(
-      new Option("-t, --target <target>", "Target platform")
-        .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
-        .default("sim")
+      new Option(
+        "-t, --platform <platform...>",
```
eladb, line 143: can this also be `-p`?

eladb (approved):

apps/wing/src/commands/compile.test.ts
```diff
@@ -45,7 +45,7 @@ describe(
     test("should be able to compile the SDK capture primitives test to awscdk", async () => {
       process.env.CDK_STACK_NAME = "compile-test-stack";
       const artifactDir = await compile(exampleFilePath2, {
-        target: Target.AWSCDK,
+        platform: [BuiltinPlatform.AWSCDK],
```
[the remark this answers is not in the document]
hasanaburayyan: Yes that is right, I have a an open issue for it https://github.com/winglang/wing/issues/4693

apps/wing/src/cli.ts
```diff
@@ -139,11 +138,7 @@ async function main() {
     .command("compile")
     .description("Compiles a Wing program")
     .argument("[entrypoint]", "program .w entrypoint")
-    .addOption(
-      new Option("-t, --target <target>", "Target platform")
-        .choices(["tf-aws", "tf-azure", "tf-gcp", "sim", "awscdk"])
-        .default("sim")
-    )
+    .addOption(new Option("-t, --platform <platform...>", "Target platform").default(["sim"]))
```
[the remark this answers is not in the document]
hasanaburayyan: That is what I had planed on, the pros of leaving it `-t` is most people wont even notice the breaking change 😁 `wing compile -t tf-aws` will still work the same.

Though `-p` is more sensical, so Ill get it fixed up before taking PR out of draft.

docs/docs/02-concepts/03-platform-providers.md
```diff
@@ -0,0 +1,235 @@
...
+id: platform-providers
+description: Wing platform providers
+keywords: [Wing platform providers, provisioning engines, multi-cloud]
+---
+
+When working with the Wing programming language, an integral part of the compilation process is the use of platform providers. In essence, platform providers specify how and where your application is deployed. They determine both the cloud environment and the provisioning engine that the code will be deployed with.
+
+## Platforms
+
+You can view the list of available builtin platform providers with the `wing compile --help` command. Here is an example of the output:
```
Chriscbr, line 12: What's the difference between a platform and a platform provider?

This section is titled Platforms but it starts talking about platform providers so I was left a little confused

(also I'm wondering if the article title should be simplified to "Platforms")

hasanaburayyan: There is not a difference, one just sounds fancier 😅

No the real answer is I have been using them synonymously but there might be a distinction between a general platform which might just be a small platform that adds an aspect to a construct tree and a platform provider that must be the first platform that offers an App?

I dunno, need to shop around the verbiage. For now I think maybe just simplify the terms to just be "platform" I wrote most of this doc in a frenzied stupor 

@Chriscbr WDYT?

Chriscbr: I like the idea of using platform everywhere. We could also maybe say a "platform provider" is the implementing code for a platform in the Wing toolchain, if it seems useful to distinguishing between platform as a concept and platform as an artifact / shareable thing. But yeah even I'm not 100% sure 😅

apps/wing/src/cli.ts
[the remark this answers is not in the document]
eladb: Let's support both

docs/docs/02-concepts/03-platform-providers.md
eladb: I actually think that a more accurate term is "platform provider" because this is basically just an adapter that connects wing apps to a platform.

But I think it's okay to omit the "provider" in most cases.

Basically I am saying the current terminology feels right.

hasanaburayyan:
Taking out of draft and into ready for review in order to get full/e2e tests running. I have a few more comments to address but I think its just about ready

apps/wing/src/cli.ts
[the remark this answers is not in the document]
hasanaburayyan: Ah I totally misread your comment I thought is said change it to `-p` now I realize you said alias, yea thats a great idea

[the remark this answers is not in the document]
hasanaburayyan: Though it looks like commander does not support multiple shorthands for options. Commands can be aliased, but not seeing anything for options. (without doing something hacky)

apps/wing/src/cli.ts
eladb: ok, let's leave as `-t`

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
+---
+title: Platform
+id: platform
+description: Wing platforms
+keywords: [Wing platform, provisioning engines, multi-cloud]
```
eladb, line 5:
```suggestion
title: Platforms
id: platforms
description: Wing platforms
keywords: [platforms, targets, target, platform, aws, gcp, azure, sim, terraform, cloudformation, pulumi, provisioning engines, multi-cloud]
```

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+---
+
+When working with the Wing programming language, an integral part of the compilation process is the use of platform. In essence, platform specify how and where your application is deployed. They determine both the cloud environment and the provisioning engine that the code will be deployed with.
+
+## Platforms
+
+You can view the list of available builtin platform with the `wing compile --help` command. Here is an example of the output:
+
+```sh
+mywing compile --help
```
eladb, line 15:
```suggestion
wing compile --help
```

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+
+Arguments:
+  entrypoint                    program .w entrypoint
+
+Options:
+  -t, --platform <platform...>  Target platform provider (builtin: sim, tf-aws, tf-azure, tf-gcp, awscdk) (default: ["sim"])
+  -h, --help                    display help for command
+```
+
+In the Options section above, there are several builtin platform that are shipped with Wing, such as `sim`, `tf-aws`, `tf-azure`, `tf-gcp`, and `awscdk`.
```
eladb, line 28:
```suggestion
Wing is shipped with several builtin platforms such as `sim`, `tf-aws`, `tf-azure`, `tf-gcp`, and `awscdk`, but it is also possible to create and use [custom platforms] to fully control how Wing resources are deployed to the cloud.
```

Add a link to [custom platforms] below.

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+
+These providers contain a combination of provision engine and cloud environment in their names, we refer to these as the platform models (which is discussed in more detail below). The only exception is `sim`, which is a special platform for testing and simulating applications locally.
+
+### Specifying Multiple Platforms
+
+You may have noticed that the `--platform` option accepts multiple arguments. This means you can specify multiple platforms to compile your application to. For example, if you wanted to compile your application using multiple platforms
+
+```sh
+wing compile app.main.w --platform tf-aws platform-foo platform-bar
```
eladb, line 37: Don't you need to specify `-p` multiple times in this case?

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+Options:
+  -t, --platform <platform...>  Target platform provider (builtin: sim, tf-aws, tf-azure, tf-gcp, awscdk) (default: ["sim"])
+  -h, --help                    display help for command
+```
+
+In the Options section above, there are several builtin platform that are shipped with Wing, such as `sim`, `tf-aws`, `tf-azure`, `tf-gcp`, and `awscdk`.
+
+These providers contain a combination of provision engine and cloud environment in their names, we refer to these as the platform models (which is discussed in more detail below). The only exception is `sim`, which is a special platform for testing and simulating applications locally.
+
+### Specifying Multiple Platforms
```
eladb, line 32: I would put this below "custom platforms".

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Custom Platforms
+
+Wing's platform architecture is not just limited to the built-in platforms that come with the language; it's extensible. This means you can create custom platforms tailored to your needs, whether to support a unique cloud provider, introduce additional optimization layers, or integrate with specific enterprise systems.
+
+### Why might you want to create a custom platform?
+
+There are many reasons why you might want to create a custom platform. Here are a few examples:
+
+- Specialized Requirements: PErhaps your organization utilizes a particular cloud provider that isn't supported by Wing, or maybe you've built your own cloud infrastructure. Custom platforms allow you to tailor the deployment process precisely to your needs.
```
eladb, line 98:
```suggestion
- Custom infrastructure requirements: Wing's built-in platforms are opinionated about how various resources are implemented. With custom platforms, you can fully control the infrastructure configuration of each cloud resource based on the needs, policies and constraints of your team.
```

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+
+- Enhanced Security: Some organizations have stringent security requirements. With custom platforms, you can embed additional security checks, audits, or encryption layers to suit these needs.
+
+- Optimizations: Your organization may have developed optimization strategies that can help reduce costs or improve performance when deploying applications. Integrating these strategies into a custom platform can make them a seamless part of the deployment process.
+
+### Creating a custom platform
+
+Developing a custom platform requires understanding and adhering to the IPlatform interface. This ensures that your custom platform can be integrated smoothly with the Wing compilation process.
```
eladb, line 106:
```suggestion
Developing a custom platform requires understanding and adhering to the `IPlatform` interface. This ensures that your custom platform can be integrated smoothly with the Wing compilation process. 
```

docs/docs/02-concepts/03-platforms.md
```diff
@@ -0,0 +1,235 @@
...
+
+queue.setConsumer(inflight (msg: str) => {
+  invocationCounter.inc();
+});
+
+new cloud.Function(inflight ()=> { 
+  // push a message to queue
+  queue.push("m");
+  // sleep according to model 
+  if util.env("WING_MODEL") == "sim" {
```
eladb, line 78: I am thinking that perhaps we should use the term "target" instead of "model". WDYT? It would work and might feel a bit more natural, especially in this context.

libs/wingcompiler/src/compile.ts
```diff
@@ -106,20 +139,21 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
-  const synthDir = resolveSynthDir(targetdir, wingFile, options.target, testing);
+  const synthDir = resolveSynthDir(targetdir, wingFile, model, testing);
   log?.("synth dir: %s", synthDir);
   const workDir = resolve(tmpSynthDir, ".wing");
   log?.("work dir: %s", workDir);
 
   // TODO: couldn't be moved to the context's since used in utils.env(...)
   // in the future we may look for a unified approach
-  process.env["WING_TARGET"] = options.target;
+  process.env["WING_MODEL"] = model;
```
eladb, line 152:
```suggestion
  process.env["WING_TARGET"] = model;
```

Let's keep this `WING_TARGET` for now. `WING_MODEL` doesn't feel right when I am looking at this here.

eladb (changes requested):
@hasanaburayyan this is amazing work.

When reading this pull request, I am having second thoughts about the term "model". Something feels off and I am thinking that perhaps we should have just used the term "target" to represent this idea.

It would also reduce some of the breaking changes (e.g. `WING_TARGET`).

How hard would it be to change this?

docs/docs/02-concepts/03-platforms.md
hasanaburayyan: Nope, commander variadic commands just allow you to keeep space separated. Which is kinda nice and kinda not because it makes it need to come after the entrypoint file. I think I’d prefer needing multiple -p but I haven’t found a way that works

docs/docs/02-concepts/03-platforms.md
hasanaburayyan: I feel like there might be some more exploring that needs done here, I agree in this context "target" sounds better, then in others I feel like "model" is a more accurate term. That said, Im in favor of going back to target for now

hasanaburayyan:
@eladb I found myself agreeing and reverted `model` back to `target`. I still think there might be some more thinking needed around this terminology but this probably makes the most sense for now.

examples/tests/valid/platforms/replace-tf-bucket.js
```diff
@@ -0,0 +1,18 @@
...
+    if (type === BUCKET_FQN) {
+      // This is just a simple example the bucket being returned would not be
+      // valid for lifting
+      return new aws.s3Bucket.S3Bucket(scope, "MyReplacedBucket", {
+        bucket: "my_replaced_bucket"
+      });
+
+    }
+  }
+}
```
eladb, line 18:
```suggestion
exports.Platform = {
  target: "tf-aws",

  newInstance(type, scope, id, props) {
    if (type === BUCKET_FQN) {
      // This is just a simple example the bucket being returned would not be
      // valid for lifting
      return new aws.s3Bucket.S3Bucket(scope, "MyReplacedBucket", {
        bucket: "my_replaced_bucket"
      });

    }
  }
}
```

Do we really need the class there?

examples/tests/valid/platforms/replicate-s3.js
```diff
@@ -128,6 +125,9 @@ class ReplicateS3Aspect {
     }
 }
 
-exports.preSynth = function (app) {
-  cdktf.Aspects.of(app).add(new ReplicateS3Aspect());
-};
\ No newline at end of file
+exports.Platform = class ReplicateS3 {
```
eladb, line 128: What's the benefit of requiring that the export name will be Platform? Why not just make this the default export?

```suggestion
module.exports = class ReplicateS3 {
```

examples/tests/valid/platforms/tf-backend.js
```diff
@@ -29,21 +29,21 @@
...
-    case "s3":
-      return configureS3Backend(config);
-    case "azurerm":
-      return configureAzurermBackend(config);
-    case "gcs":
-      return configureGcsBackend(config);
-    default:
-      throw new Error(`Invalid value for TF_BACKEND: ${process.env.TF_BACKEND} (valid values: s3, azurerm, gcs)`)
+exports.Platform = class TFBackend {
+  target = "tf-*";
```
eladb, line 33: Nice! Is this tested?

eladb (approved):
Looks great

examples/tests/valid/platforms/tf-backend.js
hasanaburayyan: The TFBackend or the `tf-*`?

In the case of the the TFBackend, yes this has a test. 

As for the compatibility bit, no I have an open issue for enforcing compatibility: https://github.com/winglang/wing/issues/1474

examples/tests/valid/platforms/replicate-s3.js
hasanaburayyan: So I believe that the benefit is consistency across platforms written in just pure JS vs TS. If your platform is written in typescript then somewhere you will probably have:

```ts
export class Platform {
	target: "sim"

	// etc...
}
```

This allows the platform manager to just create an instance of the platform based on the convention of having a class exported called Platform. So whether its loading it in from a builtin plugin or a custom one through a sandbox, its able to just rely on the existence of the `Platform` export

Maybe the convention has some draw backs, but It felt like something that was not too crazy and easy to test for.

apps/wing/src/commands/test/test.ts
```diff
@@ -100,8 +101,9 @@ export async function test(entrypoints: string[], options: TestOptions): Promise
 }
 
 async function testOne(entrypoint: string, options: TestOptions) {
+  const model = process.env.WING_TARGET; // TODO: try to just call method
```
Chriscbr, line 104:
```suggestion
  const target = process.env.WING_TARGET; // TODO: try to just call method
```

docs/docs/02-concepts/03-platforms.md
Chriscbr, line 1: These docs are great

libs/wingcompiler/src/compile.ts
```diff
@@ -58,15 +72,12 @@ export interface CompileOptions {
 function resolveSynthDir(
   outDir: string,
   entrypoint: string,
-  target: Target,
+  model: string,
```
Chriscbr, line 75:
```suggestion
  target: string,
```

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1,7 +1,11 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`access_hidden_namespace.test.w 1`] = `
-"error: Unknown symbol \\"core\\"
+"- Compiling invalid/access_hidden_namespace.test.w to tf-aws...
+
+✖ Compiling invalid/access_hidden_namespace.test.w to tf-aws...
+
+error: Unknown symbol \\"core\\"
```
Chriscbr, line 8: I wonder why these extra logs are ending up in our snapshots

tools/hangar/src/error.test.ts
```diff
@@ -5,7 +5,8 @@ import { runWingCommand } from "./utils";
 
 errorWingFiles.forEach((wingFile) => {
   test(wingFile, async ({ expect }) => {
-    const args = ["test", "-t", "sim"];
+    const platforms = ["tf-aws"];
+    const args = ["test"];
```
Chriscbr, line 9: Regression (changed from "sim" to "tf-aws")?

Chriscbr (approved):

tools/hangar/src/error.test.ts
hasanaburayyan: son of gun, good catch

tools/hangar/__snapshots__/invalid.ts.snap
hasanaburayyan: The regression you found was causing this. Just pushed up changes that fixed it.

examples/jsii-fixture/dist/js/jsii-fixture@0.0.0.jsii.tgz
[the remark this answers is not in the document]
hasanaburayyan: Damn that pesky tarball, I just removed again

polamoros (approved):

PR #4679, opened 2023-10-26
chore(repo): reduce snapshot output noise

Chriscbr:
In the repo we have several different kinds of end-to-end tests for validating different parts of the Wing toolchain work harmoniously.

E2E tests added to the `valid`, `invalid`, and `error` tests folders are usually associated with new compiler features and bug fixes (changes that may affect parsing, type checking, LSP features, etc.). To make it easy to see how code generation is affected by changes, we store snapshots of the JavaScript produced by each of these tests (along with other outputs) - you can see an example [here](https://github.com/winglang/wing/blob/22a07719a928074eaee0b3e314acc17869e44531/tools/hangar/__snapshots__/test_corpus/valid/capture_reassigable_class_field.test.w_compile_tf-aws.md).

E2E tests in the `sdk_tests` are used to check the behavior of the built-in APIs in the Wing standard library, including built-in types and cloud resources, and ensuring they have same behavior across different platforms. How the compiler parses, type checks, or generates JavaScript isn't the main focus of these tests. To reduce noise when reviewing compiler changes (where hundreds of snapshots may be updated), this PR removes the JavaScript snapshots associated with the `sdk_tests`.

## Checklist

- [ ] Tests added (always)


tools/hangar/src/generated_test_targets.ts
```diff
@@ -65,7 +69,8 @@ export async function compileTest(
 export async function testTest(
   sourceDir: string,
   wingFile: string,
-  env?: Record<string, string>
+  env?: Record<string, string>,
+  includeJavaScriptInSnapshots: boolean = true
```
yoav-steinberg, line 73: doesn't seem to be used

yoav-steinberg (approved):

hasanaburayyan (approved):

eladb:
Makes sense

PR #4563, opened 2023-10-17
feat(sdk): add `Bucket.copy()` for `sim` and `aws` targets

garysassano:
Closes #4395
Closes #4396

## Checklist

- [x] Title matches [Winglang's style guide](https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted)
- [x] Description explains motivation and solution
- [x] Tests added (always)
- [ ] Docs updated (only required for features)
- [ ] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Wing Cloud Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.

PR #4690, opened 2023-10-27
feat(sdk): add `Bucket.copy()` for `tf-azure` target

### LINKED ISSUE #4397: Implement `cloud.Bucket` inflight method `copy` for `tf-azure` target
```js
/**
 * Copy an object to a new location in the bucket. If the destination object
 * already exists, it will be overwritten.
 *
 * @throws Will throw if the `src` object doesn't exist.
 */
inflight copy(src: str, dst: str): void;
```

garysassano:
Closes #4397

## Checklist

- [x] Tests added (always)


tsuf239:
will be reviewed after merging in the copy wing test in  https://github.com/winglang/wing/pull/4563

tsuf239:
Can you please merge in main?

garysassano:
Waiting for #4691 to get merged so I run the test in Azure.

garysassano:
Waiting for #4691 to get merged to be able to test on Azure.

garysassano:
Ready to ship.

azure-test-pass

garysassano:
Ready to 🚢

azure-test-pass

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -264,16 +264,23 @@ export class BucketClient implements IBucketClient {
...
    * @param srcKey The key of the source object you wish to copy.
    * @param dstKey The key of the destination object after copying.
    * @throws if `srcKey` object doesn't exist.
    */
   public async copy(srcKey: string, dstKey: string): Promise<void> {
-    return Promise.reject(
-      `copy is not implemented: (srcKey=${srcKey}, dstKey=${dstKey})`
-    );
+    const srcBlobUrl = this.containerClient.getBlobClient(srcKey).url;
+    const dstBlobClient = this.containerClient.getBlockBlobClient(dstKey);
```
tsuf239, line 275: what is the difference between `getBlobClient` to `getBlockBlobClient`?

garysassano: `BlobClient` is a generic client that works for all [blob types](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction#blobs), so it's generally used for read operations where you don't know the type of the blob you are accessing.

`BlockBlobClient` is a specialized client that works only for block blobs. This means that you must be sure that the blob on which you are operating is of the correct type. As you can imagine, type-specific clients have a wider set of methods for manipulating a particular blob type.

tsuf239: and why are we sure that the dest key is a block blob type?

garysassano: @tsuf239 Because you create the destination blob. You don't know what you are gonna read, but you definitely know what you want to write.

garysassano: @tsuf239 Because you are the one creating the destination blob. Basically, don't know what you are gonna read, but you definitely know what you want to write.

garysassano: @tsuf239 Because you are the one creating the destination blob. Basically, you don't know what you are gonna read, but you definitely know what you want to write.

libs/wingsdk/src/target-tf-azure/bucket.inflight.ts
```diff
@@ -264,16 +264,23 @@ export class BucketClient implements IBucketClient {
...
+
+    try {
+      await dstBlobClient.syncCopyFromURL(srcBlobUrl);
+    } catch (error) {
+      throw new Error(
+        `Unable to copy. Source object does not exist (srcKey=${srcKey}).`
```
eladb, line 281: Is that the only error. I would chain the original error message to the string:

```suggestion
        `Unable to copy. ${error.message} (srcKey=${srcKey}).`
```

eladb (approved):

PR #4697, opened 2023-10-27
feat(sdk)!: expose simulated resources via HTTP server (attempt 2)

Chriscbr:
Reroll of the changes in #4105. Differences from last time:
- using the built-in `http` module instead of `express`
- added a small load test (examples/tests/sdk_tests/bucket/load_test.test.w)
- adjusted some TypeScript import statements to reduce bundle sizes
- added extra debug logging in the simulator for showing produced bundle sizes
- handle an additional edge case - simulator HTTP server gracefully returns 500 (instead of crashing) if it receives a request to a resource that has been cleaned up after the simulation shut-down sequence

----

This PR evolves the Wing simulator so that every simulation spins up an HTTP server. This HTTP server accepts requests to call inflight methods on resources, making it possible for simulated resources to interact across multiple processes on the same machine (such as from containers running on your system).

BREAKING CHANGE: The rawClient() method on ex.Redis has been removed. This API was infrequently used and was provided without any type information. If you have a use case for this API, let us know!

## Checklist

- [ ] Tests added (always)


libs/wingsdk/test/target-sim/__snapshots__/bucket.test.ts.snap
```diff
@@ -17,18 +17,21 @@ exports[`bucket on event creates 3 topics, and sends the right event and key in
   "Publish (message=a).",
   "Sending message (message=a, subscriber=sim-6).",
   "Put (key=a).",
+  "Bundled code (697326 bytes).",
```
Chriscbr, line 20: On one hand this info feels useful, though I think it might cause a lot more noisy snapshots... maybe I'll hide behind the DEBUG flag

eladb (approved):

PR #4707, opened 2023-10-29
chore(sdk, repo): adding tf-azure as a spec test target

tsuf239:
## Description
- adding tf-azure as a spec test target
- fixing azure function and bucket permissions
- all tests are passing! 
     https://github.com/winglang/wing/actions/runs/6696481710

## Checklist

- [ ] Tests added (always)


MarkMcCulloh:
These tests take quite a while, looks like it's mainly because each test file has to go through the tf apply/destroy cycle. 
As part of this PR, what if we reduced the number of test files for most modules? Especially ones like `math`, which has way too many.

tsuf239:
> These tests take quite a while, looks like it's mainly because each test file has to go through the tf apply/destroy cycle. As part of this PR, what if we reduced the number of test files for most modules? Especially ones like `math`, which has way too many.

I'm not sure it will help, as long as the test scope is still duplicating tests- since we'll have more tests per file, I can skip it- like I did for the std, if it helps (since it's the same js that is running on the sim)

.github/workflows/tf-aws-test.yml
```diff
@@ -23,7 +30,11 @@ env:
   MANUAL: ${{ github.event_name == 'workflow_dispatch' }}
   REPO: ${{ inputs.repo || github.repository }}
   REF: ${{ inputs.ref || github.ref }}
+  TARGET: ${{ inputs.target || 'all' }}
   PNPM_VERSION: "8.6.3"
+  AZURE_LOCATION: "East Us"
```
hasanaburayyan, line 35: nit:
```suggestion
  AZURE_LOCATION: "East US"
```

.github/workflows/tf-aws-test.yml
```diff
@@ -54,19 +65,31 @@ jobs:
...
   test-tf-aws:
     needs: setup
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       max-parallel: 10
-      matrix: ${{ fromJson(needs.setup.outputs.tests) }}
-    name: ${{ matrix.test.name }}
+      matrix:
+        test: ${{ fromJson(needs.setup.outputs.tests) }}
```
hasanaburayyan, line 90: TIL

examples/tests/sdk_tests/schedule/init.test.w
```diff
@@ -5,6 +5,8 @@ skip: true
 
 bring cloud;
 
+new cloud.Schedule( rate: 5m ) as "s0";
+
```
hasanaburayyan, line 9: What does this test?

libs/wingsdk/src/target-tf-azure/function.ts
```diff
@@ -182,27 +182,32 @@ export class Function extends cloud.Function {
       }) as any,
     });
 
-    // Add permissions to read function code
-    new RoleAssignment(this, `ReadLambdaCodeAssignment`, {
+    // Apply permissions from bound resources
+    for (const key of this.permissions?.keys() || []) {
```
hasanaburayyan, line 186: permissions is no longer optional, so I think we can remove
```suggestion
    for (const key of this.permissions.keys() || []) {
```

libs/wingsdk/src/target-tf-azure/function.ts
```diff
@@ -182,27 +182,32 @@ export class Function extends cloud.Function {
...
+      const scopedRoleAssignment = this.permissions?.get(
+        key
+      ) as ScopedRoleAssignment;
```
hasanaburayyan, line 189: Do we need to cast?
```suggestion
      )
```

libs/wingsdk/src/target-tf-azure/bucket.ts
```diff
@@ -77,7 +77,7 @@ export class Bucket extends cloud.Bucket {
     this.storageContainer = new StorageContainer(this, "Bucket", {
       name: storageContainerName,
       storageAccountName: this.storageAccount.name,
-      containerAccessType: this.public ? "public" : "private",
+      containerAccessType: this.public ? "blob" : "private",
```
hasanaburayyan, line 80: Just want to make sure I understand. `blob` allows access to the blobs in the container but unlike `public` does not give access to metadata and properties on container?

hasanaburayyan (approved):
@tsuf239 LGTM, I left a few questions/comments

libs/wingsdk/src/target-tf-azure/bucket.ts
tsuf239: There are no `public` unfortunately- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_container.html#container_access_type
do you think `container` will be better than blob?

examples/tests/sdk_tests/schedule/init.test.w
tsuf239: In this file we are testing the schedule constructor, (since we have a few ways to init it, and we want to avoid regression)- 
I've added a valid one on the top of the file since it is complaining about an empty configuration file if not added (it does the same with the awscdk target)

MarkMcCulloh (approved):

examples/tests/sdk_tests/schedule/init.test.w
hasanaburayyan: 👍

PR #4741, opened 2023-10-31
feat(cli, sdk): resource usage report

### LINKED ISSUE #4644: "translate" passing tests to supported resources, methods and props

tsuf239:
## Description
closes #4644 
As a part of the matrix automation process, we need to figure out what are the methods and properties that are used in each test. I came out with the following approach:
- collect the inflight methods from the `_addOnLift`
- collect the preflight methods and properties by adding a proxy to the Resource class
- save the collected methods list to a file in the output dir (currently, it may change in the future)

Please let me know what you think :) 

TODO:
- [ ] add tests
- [ ] prefix "internal" prelight properties and methods with a "_"
 
## Checklist

- [ ] Tests added (always)


MarkMcCulloh:
These tests take quite a while, looks like it's mainly because each test file has to go through the tf apply/destroy cycle. 
As part of this PR, what if we reduced the number of test files for most modules? Especiall ones like `math` which have way too many.

libs/wingsdk/src/std/resource.ts
```diff
@@ -176,6 +176,33 @@ export abstract class Resource extends Construct implements IResource {
 
   private readonly onLiftMap: Map<IInflightHost, Set<string>> = new Map();
 
+  constructor(scope: Construct, id: string) {
```
Chriscbr, line 179: Can we implement this without adding a dependency on `std.Resource`? We have some plans to deprecate this class (see https://github.com/winglang/wing/issues/3717)

Chriscbr (changes requested):

libs/wingsdk/src/std/resource.ts
tsuf239: we'll need to attach it to whatever class that replaces that- it has to be a parent class for all of the resources (even if it will be empty besides this constructor) should I add a comment?

Chriscbr: > we'll need to attach it to whatever class that replaces that

You can safely assume every preflight class extends `constructs.Construct`.

Maybe it's possible to inject this logic in the App's `new` function? see here: https://github.com/winglang/wing/blob/cac0b1d1cf9d9c864da1f8ade48e1eeaf3720814/libs/wingsdk/src/core/app.ts#L203-L218

tsuf239:
Way too many conflicts. closing that one and opening #5243 instead

Issue #4066, opened 2023-09-03
"Show tests" button in Wing Console is not explained


Chriscbr:
### I tried this:

If your Wing program does not have any `test` statements, then clicking the button does not do anything, which confused some first-time users.

### This happened:

Nothing

### I expected this:

Maybe the button should hide when my application does not have any tests, or maybe it should have a mouse-over hint, etc.

### Component

Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.43.1](https://github.com/winglang/wing/releases/tag/v0.43.1).

PR #4752, opened 2023-11-01
fix(console): show tests button is visible when there are no tests

ainvoner:
resolves: https://github.com/winglang/wing/issues/4066

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing-console/console/ui/src/services/use-tests.ts
+++ b/apps/wing-console/console/ui/src/services/use-tests.ts
@@ -1,9 +1,13 @@
+import { inferRouterOutputs } from "@trpc/server";
+import { Router } from "@wingconsole/server";
 import { useContext, useEffect, useState } from "react";
 
 import { TestItem, TestStatus } from "../shared/test-item.js";
 import { TestsContext } from "../tests-context.js";
 
 import { trpc } from "./trpc.js";
+
+type RouterOutput = inferRouterOutputs<Router>;
 
 export const useTests = () => {
   const [testList, setTestList] = useState<TestItem[]>([]);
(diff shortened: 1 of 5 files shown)


apps/wing-console/console/ui/src/services/use-tests.ts
```diff
@@ -26,10 +28,8 @@ export const useTests = () => {
   });
 
   useEffect(() => {
-    if (!testListQuery.data) {
-      return;
-    }
-    setTestList(testListQuery.data);
+    setTestList(testListQuery.data || []);
+    setTestsExists(!!testListQuery.data && testListQuery.data.length > 0);
   }, [testListQuery.data]);
```
skyrpex, line 33: I'd rather keep the early return but use two different effects. It's more readable to me.

polamoros:
I don't really like the current solution for the UI.

Maybe we should add an empty state with a link to the docs. [(link)](https://www.winglang.io/docs/concepts/tests)
image

Or we could simplify the Tests sections to something like:
image

apps/wing-console/console/ui/src/ui/no-tests.tsx
```diff
@@ -0,0 +1,27 @@
...
+        "flex flex-col text-2xs px-3 py-2 font-mono items-center",
+      )}
+    >
+      <div>There are no tests.</div>
+      <div>
+        <span>Learn how to add tests </span>
+        <a
+          className={"text-sky-500 hover:text-sky-600"}
+          href={"https://www.winglang.io/docs/concepts/tests"}
+          target={"_blank"}
```
polamoros, line 19:
```suggestion
          className="text-sky-500 hover:text-sky-600"
          href="https://www.winglang.io/docs/concepts/tests"
          target="_blank"
```

polamoros (approved):

Issue #1980, opened 2023-04-04
cloud.Function timeout not working in simulator - vm.runInContext doesn't seem to kill the function


tsuf239:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### I tried this:

run a cloud function with a timeout set to 1s, that is calling a timeout of 1.5s.
```
bring cloud;

let c = new cloud.Counter();

resource Helper {
  init () {}
  extern "./sleep.js" inflight sleep(time_ms: num);
}

let helper = new Helper();

let f = new cloud.Function(inflight () => {
  // call sleep
  helper.sleep(1500);
  c.inc();
}, cloud.FunctionProps {timeout: 1s});

// doesn't work yet- see issue: 
new cloud.Function(inflight () => {
try{
  f.invoke("");
} catch (err) {}
  assert(c.peek() == 0);
}) as "test:function timeout";
```

### I expected this:

The process that the function is running on stops, and an error is thrown.

### Instead, this happened:

The function exceeded the timeout. No error was thrown.
it happened because of vm.runInContext doesn't "wait" for the inflight code to run, but executes it synchronically (without "await"). We suspect it runs on a different process and is not interrupted by the timeout completion.

### Component:

SDK

Chriscbr:
Based on this issue https://github.com/nodejs/node/issues/3020 it seems like the behavior we're seeing is unavoidable as long as we're using Node's `vm` capabilities.

Related: https://github.com/winglang/wing/issues/4725

monadabot:
Congrats! :rocket: This was released in [Wing 0.59.43](https://github.com/winglang/wing/releases/tag/v0.59.43).

PR #4760, opened 2023-11-01
fix(sdk): aws queue approxSize

marciocadev:
The `approxSize` on AWS only filters `ApproximateNumberOfMessages`, but this should not be the only filter.

* `ApproximateNumberOfMessages`: Returns the approximate number of messages available for retrieval from the queue.

* `ApproximateNumberOfMessagesDelayed`: Returns the approximate number of messages in the queue that are delayed and not available for reading immediately. This can happen when the queue is configured as a delay queue or when a message has been sent with a delay parameter.

* `ApproximateNumberOfMessagesNotVisible`: Returns the approximate number of messages that are in flight. Messages are considered to be in flight if they have been sent to a client but have not yet been deleted or have not yet reached the end of their visibility window.

However, the `ApproximateNumberOfMessagesDelayed`, `ApproximateNumberOfMessagesNotVisible`, and `ApproximateNumberOfMessagesVisible` metrics may not achieve consistency until at least 1 minute after the producers stop sending messages. This period is required for the queue metadata to reach eventual consistency.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/shared-aws/queue.inflight.ts
```diff
@@ -51,10 +51,22 @@ export class QueueClient implements IQueueClient {
   public async approxSize(): Promise<number> {
     const command = new GetQueueAttributesCommand({
       QueueUrl: this.queueUrl,
-      AttributeNames: ["ApproximateNumberOfMessages"],
+      AttributeNames: [
```
revitalbarletz, line 54: Sweet!

libs/wingsdk/src/shared-aws/queue.inflight.ts
```diff
@@ -51,10 +51,22 @@ export class QueueClient implements IQueueClient {
...
+        "ApproximateNumberOfMessages",
+        "ApproximateNumberOfMessagesNotVisible",
```
Chriscbr, line 56: Including the inflight messages in the count sounds ok to me. Let's note it in the API docstring for `approxSize` (in cloud/queue.ts):

```
Retrieve the approximate number of messages in the queue.
This include messages that are still being processed by other queue clients.
```

Can we add a test for this? For example, if I have a queue consumer that takes 2 minutes to process a message, and I check on the queue's size after 1 minute, the queue's size should still be 1. This might require a change in the simulator.

marciocadev: Sure, let me write a specific approxSize test

marciocadev: @Chriscbr 

I think I find a problem on queue simulator look at this

https://github.com/winglang/wing/blob/104b550ba1aeb050171a5f561b2a92d80b23c474/libs/wingsdk/src/target-sim/queue.inflight.ts#L20

I don't think that the subscribers to a queue should be an array.
If a queue is consumed by an array of subscribers, it makes its behavior similar to a topic, don't you think?

marciocadev: This test is complicated to perform in the simulator because the `processMessages` method of the Queue class in the simulator removes messages from the message list for processing.

Even when avoiding the removal of messages, the problem persists since the execution of the function is done asynchronously. This means that the sleep placed within the function does not prevent the `processMessages` processing from continuing and successfully completes the processing, removing the messages.

This behavior also prevents a function from being interrupted by a `timeout` in the simulator, and essentially, all error handling for the invocation of the `processMessages` function is never executed

I believe that this issue regarding the asynchronous behavior in the simulator is mapped in this issue https://github.com/winglang/wing/issues/1980

Chriscbr: I see... #1980 seems important for this

Chriscbr:
@marciocadev I see the PR is marked as `fix`, is there an bug caused because the SDK is only counting `ApproximateNumberOfMessages` today?

I'm wondering if we should hold off on this change until it's possible to support it on both `sim` and AWS platforms since otherwise it will otherwise introduce other kinds of bugs

Chriscbr:
@marciocadev For housekeeping purposes I'll close this PR for now as it seems to be stale. Let me know if you need it reopened.

PR #4802, opened 2023-11-05
fix(sdk): revamp `Bucket.signedUrl()` to handle both GET and PUT actions

### LINKED ISSUE #4031: Add upload operation to `cloud.Bucket.signedUrl`
### Feature Spec

This issue is related to #1383.

The `cloud.Bucket.signedUrl` method was originally designed with only the download operation in mind, but a signed url could actually allow either the download or the upload operation to a bucket.

Because of this, we should decide on which approach we want to take:
- Add an extra `command` parameter which specifies if you want to create a `GET` or `PUT` signed url
- Split the method into two separate methods called something like `signedUrlGet` and `SignedUrlPut`

### Use Cases

Allow both downloading and uploading objects to a bucket through a signed url.

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

garysassano:
Closes #4031

## Checklist

- [x] Tests added (always)


garysassano:
It would make sense to separate the `action` from the `opts` since it's an essential parameter for generating a presigned URL. Making `action` a standalone parameter would improve the method's clarity. What do you think? @Chriscbr 

```ts
bucket.signedUrl("dir/foo.txt", "GET", { duration: 300s })
```

revitalbarletz:
@staycoolcall911 can we add an assignee here?

staycoolcall911:
@revitalbarletz - the team usually doesn't review `draft` PRs, unless specifically asked. Let's wait for the PR to mature out of `draft` status or for a review to be requested.
(that's unless you were asked for a review on a different way - in that case, happy to add a reviewer of course)

garysassano:
@staycoolcall911 I was the one specifically asking for the review. Before I fix this, I'd like confirmation if [this method signature](https://github.com/winglang/wing/pull/4802#issuecomment-1797852685) is ok or if it should be modified. I personally believe the action should be explicit and not optional.

garysassano:
Additionally, I think that in order to effectively test this, it might be necessary to first have #5167 merged. As far as I'm aware, there isn't an alternative Wing library capable of performing a task similar to `curl -X PUT -T <FILE_PATH> <SIGNED_URL>` for testing the created presigned URLs.

staycoolcall911:
@Chriscbr - could you please take a look at the method signature Gary suggests and let him know whether he should proceed or make changes? Thank you!
https://github.com/winglang/wing/pull/5217#issuecomment-1860319158

Chriscbr:
> It would make sense to separate the `action` from the `opts` since it's an essential parameter for generating a presigned URL. Making `action` a standalone parameter would improve the method's clarity. What do you think? @Chriscbr
> 
> ```ts
> bucket.signedUrl("dir/foo.txt", "GET", { duration: 300s })
> ```

I think even if we make the access required, I'd prefer leaving it in `opts` since it lets anyone reorder the options, and the label makes the code a little more readable to anyone unfamiliar with the API:

```ts
bucket.signedUrl("dir/foo.txt", access: "GET", duration: 300s)
```

garysassano:
@Chriscbr Should be ready to 🚢 now. Can you take a look?

examples/tests/sdk_tests/bucket/signed_url.test.w
```diff
@@ -1,29 +1,72 @@
...
+test "signedUrl GET" {
+  let KEY = "tempfile.txt";
+  let VALUE = "Hello, Wing!";
+
+  bucket.put(KEY, VALUE);
+
+  let getSignedUrl = bucket.signedUrl(KEY);
+
+  // Download file from private bucket using GET presigned URL
+  let output = util.shell("curl \"{getSignedUrl}\"");
```
tsuf239, line 17: interesting, does it work the same on Windows, Linux and Mac?

libs/ts4w/dist/index.d.ts
```diff
@@ -0,0 +1,31 @@
+import { Construct } from 'constructs';
```
tsuf239, line 1: what is this doing here?

examples/tests/sdk_tests/bucket/signed_url.test.w
```diff
@@ -1,29 +1,72 @@
...
-    let signedUrl = testBucket.signedUrl("file1.txt");
-    assert(http.get(signedUrl).body ==  "Foo"); 
-  } 
+test "signedUrl GET" {
+  let KEY = "tempfile.txt";
+  let VALUE = "Hello, Wing!";
+
+  bucket.put(KEY, VALUE);
+
+  let getSignedUrl = bucket.signedUrl(KEY);
```
tsuf239, line 14: can you add another one when specifying the `action: "GET"` option?

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -252,14 +252,34 @@ export interface BucketDeleteOptions {
...
+  PUT = "PUT",
+}
+
 /**
  * Options for `Bucket.signedUrl()`.
  */
 export interface BucketSignedUrlOptions {
   /**
-   * The duration for the signed url to expire
+   * The duration for the signed URL to expire.
```
tsuf239, line 274: what is the default?

examples/tests/sdk_tests/bucket/signed_url.test.w
```diff
@@ -1,29 +1,72 @@
...
+  });
+}
+
+test "signedUrl PUT" {
+  let KEY = "tempfile.txt";
+  let VALUE = "Hello, Wing!";
+
+  let tempDir = fs.mkdtemp();
+  let tempFile = fs.join(tempDir, KEY);
+  fs.writeFile(tempFile, VALUE);
```
tsuf239, line 47: the tmp file appears in the PR, can you delete the tmp dir afterward?

testfile.txt
tsuf239, line 1: Can you please remove?

tsuf239 (commented):
Somehow there are many `libs/ts4w/dist...` files here, can you please update from main?

libs/wingsdk/src/cloud/bucket.ts
garysassano: I didn't want to specify it there since it would be different based on the cloud provider:
- AWS is 15m.
- GCP is 1h.
- Azure forces you to specify it.

I think we can streamline it to be 15m for all of them.

examples/tests/sdk_tests/bucket/signed_url.test.w
garysassano: That was actually manually added by me at first. The `tempFile` automatically gets deleted as it's placed inside a `tempDir` (temp dirs are cleaned up at the end of tests as part of the test routine).

examples/tests/sdk_tests/bucket/signed_url.test.w
garysassano: All tests passed so it looks like that.

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -343,37 +344,50 @@ export class BucketClient implements IBucketClient {
...
+        throw new Error(`Invalid action: ${opts?.action}`);
     }
-    const expiryTimeInSeconds: number = options?.duration?.seconds || 86400;
-    const command: GetObjectCommand = new GetObjectCommand({
-      Bucket: this.bucketName,
-      Key: key,
+
+    // Generate the presigned URL
+    const signedUrl = await getSignedUrl(this.s3Client, command, {
+      expiresIn: opts?.duration?.seconds ?? 900,
```
tsuf239, line 387: Can you please set it to follow the default?

libs/wingsdk/src/cloud/bucket.ts
tsuf239: @Chriscbr FYI

Chriscbr: I feel like we should have the same defaults across targets unless there's an exceptional reason for one of them (e.g. related to `sim` being local). 15min or 1hr sounds alright as a default

libs/wingsdk/src/shared-aws/bucket.inflight.ts
garysassano: It already follows the default (15m).

libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -252,14 +252,35 @@ export interface BucketDeleteOptions {
...
+export enum BucketSignedUrlAction {
+  /**
+   * Represents a GET request for a presigned URL, allowing read access for an object in the bucket.
+   */
+  GET = "GET",
+  /**
+   * Represents a PUT request for a presigned URL, allowing write access for an object in the bucket.
+   */
+  PUT = "PUT",
+}
```
Chriscbr, line 267: For naming, I'm wondering if we should use a higher level mental model that doesn't refer to low-level HTTP verbs, e.g.
```
enum BucketSignedUrlAccess {
  READ,
  WRITE,
}
```

For example, in my head my coworker doesn't give me "put" access to their google docs spreadsheet, they give me "write" access, so it may be more common / easier to understand for anyone new to signed urls. We could still mention the HTTP verbs in the docstrings.

What do you think @garysassano @tsuf239? (I'm also totally cool with keeping the current naming)

Chriscbr (approved):

Chriscbr (approved):
Let's ship it 👍

libs/wingsdk/src/cloud/bucket.ts
garysassano: @Chriscbr  Actually, [I originally named those DOWNLOAD and UPLOAD](https://github.com/winglang/wing/pull/4802/commits/e008284160bc0c5121af29c393a980372e4a632d#diff-27f918bd13c7dacebbb394122669f25c09bf03e2b1f3eff4da3a4c487c0db0e1R267-R270), since I thought those are two actions any user would be familiar with, even without any knowledge about the underlying HTTP methods used for the operations.

garysassano: I'm going to restore the old naming, as I think that effectively conveys the purpose of the generated presigned URL, which is to allow you to either download or upload a file to a private bucket.

PR #4817, opened 2023-11-06
fix(sdk): website can't reference url in simulator

### LINKED ISSUE #2719: SDK: Website can't reference url in simulator
### I tried this:

```
bring cloud;

let w = new cloud.Website(path: "./foo");
w.addJson("config.json", { websiteUrl: w.url });
```

Try it in the [playground](https://play.winglang.io/?code=YnJpbmcgY2xvdWQ7CgpsZXQgdyA9IG5ldyBjbG91ZC5XZWJzaXRlKHBhdGg6ICIuL3dlYnNpdGUiKTsKdy5hZGRKc29uKCJjb25maWcuanNvbiIsIHsgd2Vic2l0ZVVybDogdy51cmwgfSk7) as well.

Taken from [here](https://github.com/winglang/research/blob/258006b78e7150b2de568fb0048d66920f001797/dogfooding/where-to-eat/food.w#L103) 

### This happened:

```
Failed to start simulator.

Error: Tried to resolve token "${root/Default/cloud.Website#attrs.url}" but resource root/Default/cloud.Website has no attribute url defined yet. Is it possible root/Default/cloud.Website needs to take a dependency on root/Default/cloud.Website?
```

It compiles to target `tf-aws` without an error (as long as the folder is present, hence doesn't work in the playground)

### I expected this:

It should let me reference the base url of the website

### Is there a workaround?

_No response_

### Component

Compiler, SDK

### Wing Version

0.17.44

### Wing Console Version

Version 0.2.154 (0.2.154)

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

Chriscbr:
Fixes #2719

Updates how we resolve tokens in the simulator to make the simulator token format a little more distinctive, and so tokens in the middle of strings are resolved.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -643,30 +681,19 @@ export class Simulator {
...
+      let resolveCount = 0; // prevent infinite loop, just in case
+      while (resolveCount < 100) {
+        resolveCount += 1;
+        const matches = obj.match(SIMULATOR_TOKEN_REGEX);
+        if (!matches) {
+          break;
+        }
+        for (const match of matches) {
+          const resolved = this.tryResolveToken(match);
+          if (resolved === undefined) {
```
tsuf239, line 693: what if a property is set to wing's `nil`?

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -643,30 +681,19 @@ export class Simulator {
...
-        if (rest.startsWith("attrs.")) {
-          const attrName = rest.slice(6);
-          const attr = config?.attrs[attrName];
-
-          // we couldn't find the attribute. this doesn't mean it doesn't exist, it's just likely
-          // that this resource haven't been started yet. so return `undefined`, which will cause
-          // this resource to go back to the init queue.
-          if (!attr) {
+      let resolveCount = 0; // prevent infinite loop, just in case
+      while (resolveCount < 100) {
```
tsuf239, line 685: Does each token have 100 tries to be resolved? that are executed one after the other?

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: good point, thanks for catching this. I think I was assuming the token always resolves to a string, but we don't really know that for sure, so I should probably model this a bit differently. Working on a fix

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: I updated this to avoid this while loop hack

eladb (approved):
I think the title should be a bit more general, something like `simulator doesn't resolve tokens in substrings` or something like that...

Issue #247, opened 2022-10-10
Add support for "punning" sugar in map/struct literals


MarkMcCulloh:
[punning](https://reasonml.github.io/docs/en/jsx#punning)

```js
const cool = "beans"
const spaghetti = {
	cool, // sugar for `cool: cool`
}
```

monadabot:
Congrats! :rocket: This was released in [Wing 0.75.4](https://github.com/winglang/wing/releases/tag/v0.75.4).

PR #4827, opened 2023-11-07
rfc: set syntax proposal

Chriscbr:
The current syntax for declaring sets in Wing uses curly braces, and allows the type annotation to be omitted if it can be inferred:
```js
let items = {1, 2, 3}; // Set<num>
let names = MutSet<str>{"alice", "bob", "charlie"};
```

This syntax matches the mathematical syntax for [sets](https://en.wikipedia.org/wiki/Set_(mathematics)) which also typically use curly braces, and it makes it convenient to create immutable set constants.

But the syntax has some drawbacks:
1. An empty set always requires a type annotation, or else `{}` will be interpreted as `Json` - this can be surprising to new users.
2. The set syntax collides with the desire for [punning in struct/Json literals](https://github.com/winglang/wing/issues/247), a feature many users have requested where the syntax `{ foo, bar }` would represent an object whose keys "foo" and "bar" are supplied by the "foo" and "bar" variables in the current scope.

This PR proposes changing the set literal syntax to use square brackets instead:

```js
let items = Set<num>[1, 2, 3];
let names = MutSet<str>["alice", "bob", "charlie"];
```

The syntax is inspired by Swift's [`Set` syntax](https://developer.apple.com/documentation/swift/set), which also uses square brackets, and [recent enhancements in C# 12](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions) to support simpler collection expressions.

While this implies square brackets will not always refer to `Array` or `Json` arrays, in practice it would be okay since sets are used much less commonly than arrays in most Wing programs, and the `Set` annotation would always be required if the collection refers to a set.

In the future we may be able to simplify the syntax further by allowing the element type to be omitted (`Set [1, 2, 3]`) and/or specify the set in [lowercase](https://github.com/winglang/wing/issues/1468) (`set [1, 2, 3]`).

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/docs/03-language-reference.md
+++ b/docs/docs/03-language-reference.md
@@ -80,22 +80,22 @@ import TOCInline from '@theme/TOCInline';
 
 #### 1.1.2 Container Types
 
-| Name          | Extra information                     |
-| ------------- | ------------------------------------- |
-| `Set<T>`      | set type (set of unique items)        |
-| `Map<T>`      | map type (key-value with string keys) |
-| `Array<T>`    | variable size array of a certain type |
-| `MutSet<T>`   | mutable set type                      |
-| `MutMap<T>`   | mutable map type                      |
-| `MutArray<T>` | mutable array type                    |
+| Name          | Extra information                               |
+| ------------- | ----------------------------------------------- |
+| `Array<T>`    | variable size array of a certain type           |
+| `Map<T>`      | map type (key-value with string keys)           |
+| `Set<T>`      | set type (unordered collection of unique items) |
+| `MutArray<T>` | mutable array type                              |
+| `MutMap<T>`   | mutable map type                                |
+| `MutSet<T>`   | mutable set type                                |
 
 > ```TS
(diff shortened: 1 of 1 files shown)


MarkMcCulloh:
I like this, it makes sense to me that `[]` is a _collection_ literal rather than specifically an _array_ literal

WeepingClown13:
Isn't the empty set behaviour more of a global one? I don't have many references but Python also needs explicit `set()` constructor for an empty set because `{}` is interpreted as an empty dictionary. There are occasions when I wanted set to have something similar but I wouldn't say it ever confused me.

As for the other part, it makes sense. Though personally I'd prefer Set<str>(a, b, c) over Set<str>[a, b, c] I think.

eladb (approved):

eladb (approved):
I believe we only want to update the language reference once this feature lands, no? Perhaps this update needs to go into the language spec RFC?

Issue #3647, opened 2023-07-30
compile the same code twice generates different archive.zip files


ekeren:
### I tried this:

I recompiled the same codebase twice and got a diff in the main.tf.json file

```ts
bring cloud;
bring util;


let b = new cloud.Bucket();
b.onCreate(inflight (name: str) => {
  log("file was created with name ${name}");
});

let  f = new cloud.Function(inflight (m: str) => {
  b.put("file.txt", m);
});

```

### This happened:

The `assets/cloudFunction_Asset_*/*.zip` files are different

See this diff output comparing  `index.js` file from multiple compilations
```sh  
<   "target/simple.tfaws.300228.tmp/.wing/inflight.$Closure2.js"(exports2, module2) {
---
>   "target/simple.tfaws.342499.tmp/.wing/inflight.$Closure2.js"(exports2, module2) {
```

### I expected this:

Although this issue doesn't create a correctness issue. It does create a lot of clatter in a single `terraform plan` command output after doing some minimal change 

### Component

Compiler

### Wing Version

0.24.55

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skyrpex:
Looks like the cause is the temporary path used to build the javascript assets:

![image](https://github.com/winglang/wing/assets/1077520/ed590e56-5b9b-4fce-9130-c549a8e28d99)

skyrpex:
Maybe this `target/main.*.*.tmp/` path can be removed from all generated code?

skorfmann:
To add some colour to this:

![Screenshot 2023-11-07 at 11 15 47](https://github.com/winglang/wing/assets/136789/cd5c4bd9-7a88-4e6c-85a6-1e6ee43a5f96)

Out of ~ 700 lines there are actually 10 relevant ones in a relatively small wing application with a static website and a few api handlers

skorfmann:
Yes, much better! 🎉

PR #4831, opened 2023-11-07
feat(compiler): wing sourcemaps and stacktraces

### LINKED ISSUE #1201: Wing/JavaScript source maps
### Community Note

### Feature Spec

Wing now reports inflight errors with direct references to where in the original Wing source code the error was thrown from.
This has been made possible by generating source maps for Wing libraries and intermediate preflight code. Recall:

- The Wing compiler operates in two steps (generating JavaScript code, and then executing it).
- Source maps are annotations added to JavaScript files that map information in the JavaScript file to which lines in the original source file it originated from.

By generating source maps, the Wing compiler can trace any line of executed JavaScript code back to the Wing code associated with it, which can be used to produce more relevant error messages and a better debugging experience.

### Use Cases

Preflight error messages (see #1034), and debugging preflight code

### Implementation Notes

_No response_

### Component

Compiler

MarkMcCulloh:
Closes #1201

### Summary

Adds sourcemap generation to the compiler. These sourcemaps are used for stacktrace rewriting in both preflight and inflight. This means that errors can now show wing source files for errors rather than the compiled JS.

Misc:
- Inflight bundles now have minified whitespace by default. This was necessary if we want to remove comments, because the bundle result cannot be modified without breaking the sourcemaps
- Console log changes (see below)

### Limitations

- The granularity of soucemaps is currently limited in many cases. To avoid spending too much time rewriting jsify.rs, not all emitted js uses the most exact span possible. This is easy to iterate on.
- Sourcemaps are enabled by default. This can cause slowdowns and may not be the desired behavior
- The "names" part of sourcemaps is not used. This is normally very important if you do any sort of name mangling, but that rarely happens in our current compiler.
- esbuild messes up the sourcemaps
  - The "hack" portion of the bundling logic is not good, but is required because otherwise the paths are wrong
  - Sources are duplicated. This doesn't break anything, but it's inefficient
  - source content is embedded in the sourcemap. This is convenient but not a good idea

### Examples

preflight:
image

inflight: (notice also that it's going through a library file and an extern)
image

### Wing Console Changes

The logs show text as-is now, which means it also shows newlines. Only the first line is shown, requiring expansion otherwise:

image
image


libs/wingsdk/src/shared/bundling.ts
```diff
@@ -47,17 +57,41 @@ export function createBundle(entrypoint: string, outputDir?: string): Bundle {
...
+  // sourcemap hacks for winglang:
+  // source paths in sourcemap are incorrect and need to be fixed
+  const sourcemapData = JSON.parse(
+    new TextDecoder().decode(esbuild.outputFiles[0].contents)
+  );
+  // unrandomize the sourceRoot
+  sourcemapData.sourceRoot = normalPath(sourcemapData.sourceRoot).replace(
+    /\.\d+\.tmp\/\.wing\//g,
+    "/.wing/"
+  );
```
MarkMcCulloh, line 72: ℹ️ Along with some other changes in this file, this probably fixes #3647. I didn't added anything to verify that though

eladb:
Congrats! very impressive work

hasanaburayyan:
freaking dope! 🔥  still making my way through this one.

I wonder if we should open some steam of consciousness (issue/discussion) about performance concerns and expectations around behavior in the cloud (i.e Cloudwatch Logs)

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -213,7 +213,14 @@ export class Function extends cloud.Function implements IAwsFunction {
...
       environment: {
-        variables: Lazy.anyValue({ produce: () => this.env }) as any,
+        variables: Lazy.anyValue({
+          produce: () => ({
+            ...this.env,
+            // enable source maps
+            NODE_OPTIONS:
+              (this.env.NODE_OPTIONS ?? "") + "--enable-source-maps",
+          }),
+        }) as any,
```
MarkMcCulloh, line 223: WRT @hasanaburayyan's comments about perf, I would be fine removing this to work on an alternate solution later. The impact of `--enable-source-maps` is not negligible, especially with aws-sdk-sized sourcemaps.

I think the ideal scenario is that functions should make the sourcemap available as a preflight value, and the CLI and/or TestRunner can be responsible for pulling in sourcemaps as needed. This also makes sense if you want to export your sourcemaps for other usecases (like uploading to your error tracking system).

apps/wing/vitest.config.ts
```diff
@@ -4,6 +4,7 @@ export default defineConfig({
   test: {
     // Some tests use process.chdir which is not supported in vitest's multi-thread mode
     threads: false,
+    update: true,
```
Chriscbr, line 7: Should we be consistent about this for all of the packages in the repo?

Chriscbr (commented):
I haven't had a chance to review the Rust changes yet, but here's an initial pass. Looks great so far

libs/wingsdk/src/target-tf-aws/function.ts
MarkMcCulloh: With this flag, when V8 generates a stacktrace it automatically takes care of remapping the file positions based on the available sourcemaps.

libs/wingsdk/src/shared/bundling.ts
```diff
@@ -35,8 +43,11 @@ export function createBundle(entrypoint: string, outputDir?: string): Bundle {
       "@winglang/sdk": SDK_PATH,
     },
     minify: false,
+    sourcemap: "external",
     platform: "node",
     target: "node18",
+    write: false,
+    minifyWhitespace: true,
```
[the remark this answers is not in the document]
MarkMcCulloh: Unfortunately we can use esbuild plugins (because we use the sync version of the API), so our options of doing this are slim.

I did just think of something to try though: I can go back to the same as before, except instead of deleting the lines and I leave them empty. Sourcemap encodings are sensitive on a per-line basis.

> I wonder if there's any way to leave the comments but instead ensure that they're the same across compilations (i.e. not pointing to temporary directories)?

I forgot that's why we originally removed comments, haha. It's actually possible my other change in this file (using `absWorkingDir`) might've fixed that anyways. I'll check it out.

tools/hangar/__snapshots__/error.ts.snap
```diff
@@ -19,15 +18,13 @@ Duration <DURATION>"
...
-       super($scope, $id);
->>     const a = ((args) => { if (isNaN(args)) {throw new Error(\\"unable to parse /\\"\\" + args + \\"/\\" as a number\\")}; return parseInt(args) })(\\"123a\\");
-     }
-   }
-
+"[ERROR] unable to parse \\"123a\\" as a number
+../../../examples/tests/error/num_from_str.test.w:3:14
+1| // Note that this test has to be alone because it needs to compile successfully and fail at preflight.
+2| // If it is run with other tests, subsequent failures will be ignored in snapshot.
+3| let a: num = num.fromStr(\\"123a\\");
```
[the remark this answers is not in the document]
MarkMcCulloh: Sure, I didn't originally because I couldn't reuse codespan-reporting, but I can make it more similar.

apps/wing/vitest.config.ts
MarkMcCulloh: Whoops, should not have left that in. The consistent thing to do would be to update the package.json script, so I'll do that

libs/wingsdk/src/util/enhanced-error.ts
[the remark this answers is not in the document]
MarkMcCulloh: Unfortunetly not, the compiler diagnostic formatting is still duplicated

libs/wingsdk/src/shared/bundling.ts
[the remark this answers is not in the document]
MarkMcCulloh: Ok I changed it to not minify, and I confirmed that my other changes got rid of the tmp dir in the path, so now no post-processing is needed on the bundle

eladb (approved):

libs/wingc/src/jsify/codemaker.rs
```diff
@@ -1,36 +1,125 @@
...
-			self.lines.push((self.indent, subline.into()));
+		for mut line_data in new_code.lines.drain(..) {
+			line_data.indent += self.indent;
+			if line_data.mappings.is_empty() {
+				if let Some(source) = self.source.as_ref() {
+					line_data.mappings.push((0, source.clone()))
+				}
+			}
+			self.push_line(line_data);
 		}
```
Chriscbr, line 123: (optional) Another possible way to write this

```suggestion
		for mut line_data in new_code.lines.drain(..) {
            let new_line_data = LineData {
                line: line_data.line,
                indent: line_data.indent + self.indent,
                mappings: line_data.mappings.unwrap_or((0, source.clone())),
            };
			self.push_line(new_line_data);
		}
```

libs/wingc/src/jsify.rs
```diff
@@ -772,29 +827,33 @@ impl<'a> JSifier<'a> {
 	}
 
 	fn jsify_statement(&self, env: &SymbolEnv, statement: &Stmt, ctx: &mut JSifyContext) -> CodeMaker {
+		let mut code = CodeMaker::with_source(&statement.span);
+
 		CompilationContext::set(CompilationPhase::Jsifying, &statement.span);
 		ctx.visit_ctx.push_stmt(statement.idx);
-		let code = match &statement.kind {
+		// code.push_original_span(statement.span.clone());
```
Chriscbr, line 834:
```suggestion
```

libs/wingc/src/jsify/codemaker.rs
```diff
@@ -42,8 +131,47 @@ impl CodeMaker {
...
+					.mappings
+					.push((last_line.line.len() - first_new_line.line.len() + offset, span));
+			}
+
+			if add_another_map {
+				if let Some(last_original_mapping) = last_original_mapping {
+					last_line.mappings.push((last_line.line.len(), last_original_mapping.1));
+				}
+			}
+		}
```
Chriscbr, line 170: This section looks important, can we add a few comments?

libs/wingc/src/jsify/codemaker.rs
```diff
@@ -153,26 +318,79 @@ mod tests {
...
+		code.append("1");
+		code.append("2");
+		code.append("\n3");
+
+		assert!(code.lines.len() == 2);
+		assert_eq!(code.to_string(), "12\n3");
+	}
+
+	#[test]
+	fn sourcemap_default() {
```
Chriscbr, line 337: What's the behavior if you try to get a sourcemap when there's no span data for any of the lines? Does it produce an empty source map? Let's add a test

Chriscbr (approved):

libs/wingc/src/jsify/codemaker.rs
MarkMcCulloh: commented and fixed a bug 👍

libs/wingc/src/jsify/codemaker.rs
MarkMcCulloh: By default you at least must have the source file, so each line will be mapped to the file (no source line/char). Added test.

libs/wingc/src/jsify/codemaker.rs
MarkMcCulloh: Rewrote it to be closer to your suggestion, although because `source` is an Option it had to be a little different

libs/wingc/src/jsify/codemaker.rs
```diff
@@ -1,36 +1,125 @@
...
+	indent: IndentAmount,
+	source: Option<WingSpan>,
+}
+
+pub type IndentAmount = usize;
+pub type CharacterOffset = usize;
+
+#[derive(Default, Clone, Debug)]
+pub struct LineData {
+	pub line: String,
```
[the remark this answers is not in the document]
MarkMcCulloh: added doc strings and included this info

skyrpex (approved):

skyrpex:
@MarkMcCulloh this is terrific! 👏🏻

PR #4837, opened 2023-11-08
fix(console): avoid unnecessary re-renders

### LINKED ISSUE #4834: fix(console): map view re-renders while zooming
Using the React devtools, we can see that the map view is re-rendered for every zoom change. It can be fixed.

skyrpex:
Uses `useCallback`, `useMemo` and `memo` where possible to avoid unnecessary re-renders.

Fixes #4834.


ainvoner (changes requested):
@skyrpex please reduce the usage of memo in pure ui components. It should be the responsibility of the using components to prevent unneeded renders

skyrpex:
> @skyrpex please reduce the usage of memo in pure ui components. It should be the responsibility of the using components to prevent unneeded renders

Done. Ready for review.

ainvoner (approved):

Issue #4847, opened 2023-11-09
Improve API of `function.invoke()`


eladb:
### Feature Spec

The signature of:

```js
function.invoke(payload: str): void
```

Should change to:

```js
function.invoke(payload: Json?): Json?
```

### Use Cases

* `Json` offers the ability to pass in and out structured data.
* Optionality improves ergonomics. If the inflight handler doesn't call `return` the value is `nil`.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Chriscbr:
Done

PR #4867, opened 2023-11-09
feat(sdk)!: response.body is str

Chriscbr:
Currently a lot of Wing code that deals with API responses has to perform extra checking to handle `body` fields that may be `nil`. For example:

```js
let response = http.get(...);
let foo = Foo.parseJson(response.body ?? "");
log(foo.bang);
```

Notice above how we have to write `response.body ?? ""` since `response.body` is typed as `str?`.

Since the `http` module calls the [text()](https://developer.mozilla.org/en-US/docs/Web/API/Response/text) method in JavaScript which always returns a `string`, it seems as though it ought to be safe to change the type of `body` to `str` and provide a better DX in the standard library:

```js
let response = http.get(...);
let foo = Foo.parseJson(response.body); // changed
log(foo.bang);
```

Another example with a HEAD request: 
```js
test "head request" {
  let resp = http.fetch("https://news.ycombinator.com", {
    method: http.HttpMethod.HEAD,
  });
  assert(resp.body == "");
}
```

## Checklist

- [x] Tests added (always)


garysassano:
I guess this is related to #4847 ?

examples/proposed/url-shortener.main.w
```diff
@@ -11,14 +11,14 @@ class UrlShortener {
...
-  inflight getId(url: str): str {
+  pub inflight getId(url: str): str {
     let id = this._makeId();
     this.lookup.put(id, url);
     return id;
   }
 
   // Get the url for the given id. Returns nil if the id is invalid.
-  inflight getUrl(id: str): str? {
+  pub inflight getUrl(id: str): str? {
```
MarkMcCulloh, line 21: Why the modifier change?

MarkMcCulloh (approved):
LGTM
I believe this should be a breaking change

examples/proposed/url-shortener.main.w
Chriscbr: Ah - this is code from the proposed examples folder. It's not getting compiled by CI but I just came around to updating the code

Chriscbr:
> I guess this is related to #4847 ?

That issue requires a separate change since it pertains to the Function class

eladb (approved):

PR #4869, opened 2023-11-09
fix(console): remove top padding for the root node

skyrpex:
There was a top padding in the root node of the map view that wasn't removed.


skyrpex:
Actually, we need to keep it for the rest of the nodes.

skyrpex:
It's not prioritary for now... Will tackle later.

Issue #4842, opened 2023-11-08
[RFC] Typescript experience for Wing


MarkMcCulloh:
### Design Tenets

- "Wing first" - We optimize for authoring libraries in winglang and consuming them from TypeScript. No need to optimize for a TypeScript authoring experience.
- We will design the TypeScript API to be as close to the the winglang experience as possible, because we assume we designed the winglang to be the ideal experience.
- Users will use the `wing` CLI to compile, test and run Wing applications written in TypeScript.
- DX is based on the construct programming model used by CDKs (`new T(scope, id)`).
- "TypeScript native" - Wing for TypeScript won't require require customizations or extensions to the TypeScript compiler. In the future we might consider adding "sugar" or "helpers" using such mechanisms, but the base experience shouldn't require this.
- The Wing IDE extension is not needed to use Wing for TypeScript.

### Feature Spec

The Wing toolchain now supports TS/JS entrypoints. Commands like `wing compile`, `wing test`, and `wing it` can take a Typescript or Javascript file. This allows you to get the benefit of the Wing ecosystem (beautiful abstractions, testing, simulator, Wing Console, and Wing Cloud) without committing to winglang (yet!)

_(example contains a proposal API for discussion, very much subject to change)_

```ts
// main.ts
import { cloud, std, expect, inflight } from "@winglang/sdk";

// Final App depends on platform used
const app = new cloud.App();

// Familiar construct-base DX
const bucket = new cloud.Bucket(app, "Bucket");
const api = new cloud.Api(app, "Api");
const message = "Hi";

// System for modeling `inflight` functions
// Will need to explore mechanisms to lift preflight objects
const func = new cloud.Function(app, "Function", inflight({ bucket, api, message }, async (ctx, event) => {
  // can access inflight methods via the ctx context
  await ctx.bucket.put("stuff", event);

  // can reference preflight fields
  console.log(ctx.api.url);

  // error: basic static analysis can catch references to preflight value without ctx
  console.log(api.url);

  // can reference static data
  console.log(ctx.message);
}))

// The explicit lifting ({ func, bucket }) may be avoidable in a future version
new std.Test(app, "can call function", inflight({ func, bucket }, async (ctx) => {
  await ctx.func.invoke();
  assert(await ctx.bucket.exists("stuff"));
})

app.synth();
```

```shell
# All valid commands
wing compile main.ts
wing it main.ts
wing test main.ts
wing test -t tf-aws main.ts
```

## MVP

[...]

PR #4871, opened 2023-11-09
feat(sdk)!: concrete resource classes

MarkMcCulloh:
Part of the effort of https://github.com/winglang/wing/issues/4842

Will allow you to directly do `new cloud.Bucket` in TS directly and still get the benefit of the App factory.

Changed it so that you can no longer instantiate abstract classes, since there is no need for it anymore.


libs/wingsdk/src/cloud/api.ts
```diff
@@ -219,6 +209,10 @@ export abstract class Api extends Resource {
   protected corsOptions?: ApiCorsOptions;
 
   constructor(scope: Construct, id: string, props: ApiProps = {}) {
+    if (new.target === Api) {
+      return App.of(scope).newAbstract(API_FQN, scope, id, props);
+    }
+
     super(scope, id);
```
MarkMcCulloh, line 216: @eladb @Chriscbr
This seems to work... Am I missing something? I feel like I remember some issues with this approach back in the polycon days

Chriscbr: I tried racking my brain but I couldn't think of anything concrete... curious if @eladb thinks of anything.

I'm wondering if someone tries instantiating this construct in TypeScript outside of a Wing `App` (e.g. inside of a plain cdktf `App`), maybe we can give a nicer error (e.g. "This is an abstract class, so it can only be instantiated in the scope of a Wing App")

libs/wingsdk/src/cloud/api.ts
```diff
@@ -173,23 +173,13 @@ export type CorsHeaders = {
...
-   */
-  public static _newApi(
-    scope: Construct,
-    id: string,
-    props: ApiProps = {}
-  ): Api {
-    return App.of(scope).newAbstract(API_FQN, scope, id, props);
-  }
-
+export class Api extends Resource {
```
Chriscbr, line 176: Even though these are no longer abstract in TypeScript, I'm wondering if we should still track their "abstract"-ness in some way, e.g. with an `@abstract` flag in the docstring

Chriscbr (approved):
LGTM

eladb (approved):
ship it

libs/wingsdk/src/ex/dynamodb-table.ts
```diff
@@ -83,7 +75,14 @@ export abstract class DynamodbTable extends Resource {
...
   /** @internal */
-  public abstract _supportedOps(): string[];
+  public _supportedOps(): string[] {
+    throw new Error("proxy");
+  }
+
+  /** @internal */
+  public _toInflight(): string {
+    throw new Error("proxy");
+  }
```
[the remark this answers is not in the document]
MarkMcCulloh: Agreed, not sure how to resolve it until then (other than ensuring all methods are tested, which will catch it).

I started writing an additional case for wing-api-checker to help with this, but then realized it doesn't have any info from the target-specific classes.

libs/wingsdk/src/cloud/api.ts
MarkMcCulloh: Added, both for classes and members that are "abstract"

PR #4938, opened 2023-11-15
feat(sdk): expose aws resources

marciocadev:
This pull request enables access to the primary AWS resources within the Wing Resources.

For instance, if you need to access the AWS Bucket object created when defining a `cloud.Bucket`, you can do so as demonstrated below.

```ts (wing)
bring cloud;
bring aws;

if let b = aws.Bucket.from(bucket) {
    let bucketArn = b.arn();
}
```

| Wing | AWS | Interface | Method |
|-|-|-:|:-|
| Api | API Gateway (RestApi) | IAwsApi | arn() |
| Api | API Gateway (RestApi) | IAwsApi | restApiId() |
| Api | API Gateway (RestApi) | IAwsApi | restApiName() |
| Api | API Gateway (RestApi) | IAwsApi | stageName() |
| Api | API Gateway (RestApi) | IAwsApi | invokeUrl() |
| Api | API Gateway (RestApi) | IAwsApi | deploymentId() |
||
| Bucket | S3 (Bucket) | IAwsBucket | arn() |
| Bucket | S3 (Bucket) | IAwsBucket | bucketName() |
||
| DynamodbTable | DynamoDB (Table) | IAwsTable | arn() |
| DynamodbTable | DynamoDB (Table) | IAwsTable | tableName() |
||
| Function | Lambda (Function) | IAwsFunction | arn() |
| Function | Lambda (Function) | IAwsFunction | functionName() |
||
| Queue | SQS (Queue) | IAwsQueue | arn() |
| Queue | SQS (Queue) | IAwsQueue | queueName() |
| Queue | SQS (Queue) | IAwsQueue | queueUrl() |
||
| Table | DynamoDB (Table) | IAwsTable | arn() |
| Table | DynamoDB (Table) | IAwsTable | tableName() |
||
| Topic | SNS (Topic) | IAwsTopic | arn() |
| Topic | SNS (Topic) | IAwsTopic | topicName() |
||
| Website | S3 (Bucket) | IAwsBucket | arn() |
| Website | S3 (Bucket) | IAwsBucket | bucketName() |

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/shared-aws/api.ts
```diff
@@ -0,0 +1,40 @@
+import { IResource } from "../std";
+
+/**
+ * A shared interface for AWS api.
+ */
+export interface IAwsApi {
+  /**
+   * Get internal AWS ApiGateway
+   */
+  innerAwsApi(): any;
```
eladb, line 10: as much as it might be tempting to simply expose these raw resources, i think that's not a very useful api because the consumer needs too much of a deep knowledge about the internal anatomy. for example, awscdk and cdktf would have a totally different internal api.


i am wondering if a better way is for each resource expose some explicit attributes. for example, for the API resource on AWS, I would expose things like `restApiId`, `stageName`, `deploymentId`, etc.

what do you think?

eladb (changes requested):
Please do not expose the inner resources as any. Instead expose specific attributes of the AWS resource in a strongly-typed manner.

libs/wingsdk/src/shared-aws/api.ts
marciocadev: I understand. 
Personally, exposing the raw resource makes it easier to work with L2 constructs from aws-cdk, but it makes the library less understandable for those without experience with CdkTF and aws-cdk.

I will make changes to expose the ids and other relevant information. 
In the case of CdkTF and aws-cdk, the naming is not always the same, so I will choose the one that I find more readable. 

For example, between `functionArn` and `arn`, I will opt for `arn` since the resource type will already be defined.
So, to obtain the function arn the code could be like this (for both cdktf and aws-cdk).

```ts
bring cloud;
bring aws;

if let b = aws.Bucket.from(bucket) {
    let functionArn = b.arn();
}
```

eladb: i actually think that `function.arn` is nicer than `function.functionArn`, but please make a revision with some naming suggestion and i'll review.

libs/wingsdk/src/shared-aws/api.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * A helper class for working with AWS apis.
+ */
+export class Api {
+  /**
+   * If the api is an AWS RestApi, return a helper interface for
+   * working with it.
+   * @param api The cloud.Api.
+   */
+  public static from(api: IResource): IAwsApi | undefined {
```
eladb, line 47: I think this should be more strongly typed:

```suggestion
  public static from(api: cloud.Api): IAwsApi | undefined {
```

eladb (changes requested):
see some comments and apply across the board

libs/wingsdk/src/shared-aws/api.ts
marciocadev: Basically, I used the `IResource` because I use the `IAwsBucket` for the `Bucket` and the `Website`, and took advantage of `Bucket.from` to check if it is an AWS resource.

However, I can create a specific one for the Website

eladb: Let's do it. The stronger the type the better.

libs/awscdk/src/bucket.ts
```diff
@@ -230,6 +231,14 @@ export class Bucket extends cloud.Bucket {
   private envName(): string {
     return `BUCKET_NAME_${this.node.addr.slice(-8)}`;
   }
+
+  public get arn(): string {
```
eladb, line 235: Why are these needed?

libs/awscdk/src/on-deploy.ts
```diff
@@ -21,7 +21,7 @@ export class OnDeploy extends cloud.OnDeploy {
     const awsFn = fn as AwsFunction;
 
     let trigger = new Trigger(this, "Trigger", {
-      handler: awsFn._function,
```
eladb, line 24: Maybe revert this change. Not sure it's needed.

libs/wingsdk/src/shared-aws/counter.ts
```diff
@@ -0,0 +1,37 @@
+import { cloud } from "..";
+
+/**
+ * A shared interface for AWS Counter.
+ */
+export interface IAwsCounter {
+  /**
+   * AWS Dynamodb arn
+   */
+  readonly arn: string;
```
eladb, line 10:
```suggestion
  readonly dynamoTableArn: string;
```

libs/wingsdk/src/shared-aws/counter.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * AWS Dynamodb name
+   */
+  readonly tableName: string;
```
eladb, line 15:
```suggestion
  readonly dynamoTableName: string;
```

libs/wingsdk/src/shared-aws/function.ts
```diff
@@ -5,6 +5,16 @@ import { IInflightHost } from "../std";
  * A shared interface for AWS functions.
  */
 export interface IAwsFunction {
+  /**
+   * AWS Function arn
+   */
+  readonly arn: string;
```
eladb, line 11:
```suggestion
  readonly functionArn: string;
```

eladb (changes requested):
Sorry for the back and forth. Looking at this holistically, i think should use the full names (xxxArn) after all.

libs/awscdk/src/bucket.ts
marciocadev: I didn't understand, Elad.

Do you want me not to expose the bucket arn?

Because I find it interesting to expose it.

Eventually, someone might want to create a library that uses Event Bridge or Step Function and wishes to connect directly to a bucket. Exposing the ARN would make it easy to do that

eladb: My mistake. Ignore my comment. I didn't realize it's a property...

eladb (approved):
Ship it

PR #4958, opened 2023-11-15
chore: remove some ceremony from bug template

MarkMcCulloh:
- Removed the "component" section. 
  - IMO it's not the responsibility of a bug reporter to know how maintainers categorize issues. This is better suited for labels at our discretion.
- Moved the "Anything Else" to be next to the rest of the actual bug report.
  - This additional context is likely more useful than the version information and putting it before that increases the chances of it being filled in
  - Removed `## Misc` since it didn't really matter without the "Anything Else"
- Added whitespace to make it consistent
- Fixed minor typo

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -6,10 +6,12 @@ body:
   attributes:
     value: |
        :pray: Thanks for taking the time to fill out this bug report! Feel free to ping us on [Wing Slack](https://t.winglang.io/slack) if you have any questions or need help.
+
 - type: markdown
   attributes:
     value: |
        ## Bug Report
+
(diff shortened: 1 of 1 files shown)


eladb (approved):

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -6,58 +6,48 @@ body:
...
 - type: textarea
   id: instead-what-happened
   attributes:
     label: "This happened:"
-    placeholder: "What happend instead of what you've expected?"
+    placeholder: "What happened instead of what you've expected?"
   validations:
     required: true
+
 - type: textarea
```
eladb, line 31: I think we can make this optional. In many cases, a simple error message in the "actual" section is sufficient.

Issue #2853, opened 2023-06-08
Inflight closure inside of for loop causes compilation error


ekeren:
### I tried this:

```ts (wing)
bring cloud;

let q0 = new cloud.Queue() as "a";
let q1 = new cloud.Queue() as "b";

let queues = [q0, q1];

for q in queues {
  q.addConsumer(inflight (m:str): void => {

  });
}
```

### This happened:

```
Failed to compile.

Error: Command failed: wing compile /private/tmp/a/app.w -t sim
Error: ERROR: There is already a Construct with name '$Closure5' in $Root [Default]

hint: Every preflight object needs a unique identifier within its scope. You can assign one as shown:

> new cloud.Bucket() as "MyBucket";

For more information, see https://docs.winglang.io/concepts/resources

private/tmp/a/target/app.wsim.543993.tmp/.wing/preflight.js:345
         class $Closure5 extends $stdlib.std.Resource {
           constructor(scope, id, ) {
>>           super(scope, id);
             this._addInflightOps("handle");
             this.display.hidden = true;

    at compile ([/Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/dist/commands/compile.js:238:15](vscode://file//Users/eyalkeren/.nvm/versions/node/v18.12.1/lib/node_modules/winglang/dist/commands/compile.js:238:15))
```

### I expected this:

to compile and work

### Is there a workaround?

yes, but it wont be DRY

### Component

Compiler

### Wing Version

0.19.3

### Anything else?

I wonder if this `$closure` resource even need to have an ID, ( it is not really on terraform) 

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

ekeren:
Was discussed in team time: @yoav-steinberg , @ekeren , @ShaiBer , @skyrpex , @Chriscbr,  @MarkMcCulloh 

It was proposed by @MarkMcCulloh that instantiating the `$closure` class we should always do 
` new $Closure() as "$Closure:${global.incClosureCounter()}" `

ekeren:
keep

ekeren:
@staycoolcall911 , I promoting this to p1

We should fix this

ekeren:
Oh, it is already P1

Chriscbr:
As a workaround you can define the inflight closure outside of the for loop:

```js
bring cloud;

let q0 = new cloud.Queue() as "a";
[...]

PR #4993, opened 2023-11-17
feat(sdk): inflights are not required to be resources

MarkMcCulloh:
## Huh?

The primary goal of this PR is to reduce the input required to create an inflight function in TS (See https://github.com/winglang/wing/issues/4842) without necessarily overhauling the compiler (yet). Ideally, the minimum information required for an inflight is simply the code itself. However, because inflights are currently modeled as resources, they require a scope and id.

So the first change was to make a new non-resource interface, IInflight, encompassing the inflight contract. The most important part of this contract is that inflights must be liftable, a behavior currently unique to resources and certain other primitives. So I extracted the lift-related functions from IResource and slapped them on the new ILiftable (which both IInflight and IResource now extend).

But that created a new problem: lifting itself also currently requires a scope. The only usage of the scope was to be able to resolve tokens. This did not seem like a good enough reason to require scope, so I changed token resolution to be more of a global concern rather than a tree-level concern. This is dangerous, but it's mostly only dangerous when you try to deal with tokens in a multi-app scenario, which would be dangerous with our current approach anyways. So this is something we'll have to add some extra handling for eventually anyways.

## Results

The primary outcome of this can be seen in the SDK unit tests, where the `Testing.makeHandler()` now only requires the code and (optional) bindings. This is basically 1 or 2 steps away from an ideal TS experience.

## But wait nothing changed in winglang

The original purpose of representing inflights as resources was to ease the implementation of lifting in the compiler and generally unify the logic of inflights between inflight closures and inflight methods of preflight classes. This hasn't changed in this PR. Luckily, the class instance emitted by the wing compiler for inflights still satisfies IInflight. It just has some extra hidden resource stuff that is simply unused. Assuming this PR is wanted, I will do a followup to change the compiler as well. This will be a more complicated change and I think it's useful to basically get the backend working first.

## Changes

- `Testing.makeHandler` now takes only code text and bindings. 9 billion tests were updated for this contract. `convertBetweenHandlers` changed similarly
- TokenResolvers are now globally registered and not tied to specific apps
- wingc adds a _hash private field to inflight closure resource classes to match the new IInflight (just an md5 hash)
- Many of the resources that deduped functions based on `addr` now do so with `_hash` instead
- Removed many occurrences of `this.node.id` used in resource ids when it's not necessary. The only time this should be necessary is if the resources is being created in the scope of this.node.scope instead
- Added a `Counter` concept to help with the many places that we want to add a subresource many times and a simple incrementing number will suffice for uniqueness and clarity
  - This was needed because the inflight `addr` was often relied upon to make this unique, but that will no longer be viable. I think it's better this way anyways


libs/wingsdk/src/core/tokens.ts
```diff
@@ -1,33 +1,45 @@
...
+/**
+ * Represents values that can only be resolved after the app is synthesized.
+ * Tokens values are captured as environment variable, and resolved through the compilation target token mechanism.
+ */
+export abstract class Tokens {
+  public static addResolver(resolver: ITokenResolver) {
+    this._resolvers.push(resolver);
+  }
+
+  public static getTokenResolver(value: any): ITokenResolver | undefined {
```
eladb, line 29: Add some docstrings here to describe api and behavior

libs/wingsdk/src/core/tokens.ts
```diff
@@ -1,33 +1,45 @@
...
    */
-  public abstract onLiftValue(host: IInflightHost, value: any): void;
+  onLiftValue(host: IInflightHost, value: any): void;
+}
+
+/**
+ * Represents values that can only be resolved after the app is synthesized.
+ * Tokens values are captured as environment variable, and resolved through the compilation target token mechanism.
+ */
+export abstract class Tokens {
```
eladb, line 24: Why is this abstract? Seems like all methods are static.

You can use a private constructor to protect against instantiation if that was the intent.

libs/wingsdk/src/shared-tf/app.ts
```diff
@@ -34,7 +34,7 @@ export abstract class CdktfApp extends App {
   public readonly terraformManifestPath: string;
   public readonly outdir: string;
   public readonly isTestEnvironment: boolean;
-  public readonly _tokens: Tokens;
```
eladb, line 37: What's the motivation to keep this field if tokens are now global?

libs/wingsdk/src/shared/misc.ts
```diff
@@ -72,3 +73,15 @@ export async function runDockerImage({
 
   return { hostPort };
 }
+
+var _id = 0;
+export function inflightId(inflight?: IInflight) {
```
eladb, line 78: I am wondering if perhaps we can just hash the contents of the inflight to generate a more stable id.

Using a global incrementing number means that multiple apps that run in the same process (eg for testing) will result in identifiers that are not deterministic (depends on the order of execution of the tests).

Alternatively, can we get rid of the id altogether somehow?

libs/wingsdk/src/std/resource.ts
```diff
@@ -18,13 +18,24 @@ export interface IInflightHost extends IResource {
...
 
 /**
- * Abstract interface for `Resource`.
- * @skipDocs
+ * TODO Docs
+ */
+export interface IInflight extends ILiftable {
+  /**
+   * Tracks the unique ID of this inflight object.
+   * This is not hash-based and will be different for two inflight instances with the same code
```
eladb, line 26: ☝️

libs/wingsdk/src/std/resource.ts
```diff
@@ -144,10 +161,10 @@ export abstract class Resource extends Construct implements IResource {
         }
 
         // if the object is a resource (i.e. has a "lift" method"), register a lifting between it and the host.
-        if (isResource(obj)) {
+        if (IsLiftable(obj) && "_addOnLift" in obj) {
```
eladb, line 164:
```suggestion
        if (isLiftable(obj) && "_addOnLift" in obj) {
```

libs/wingsdk/src/std/resource.ts
```diff
@@ -357,19 +374,14 @@ export interface OperationAnnotation {
   };
 }
 
-function isResource(obj: any): obj is Resource {
-  return isIResourceType(obj.constructor);
-}
-
-function isIResourceType(t: any): t is new (...args: any[]) => IResource {
+function IsLiftable(t: any): t is new (...args: any[]) => ILiftable {
```
eladb, line 377:
```suggestion
function isLiftable(t: any): t is new (...args: any[]) => ILiftable {
```

libs/wingsdk/src/std/resource.ts
```diff
@@ -174,6 +191,19 @@ export abstract class Resource extends Construct implements IResource {
...
+  /**
+   * "Lifts" a value into an inflight context. If the value is a resource (i.e. has a `_toInflight`
+   * method), this method will be called and the result will be returned. Otherwise, the value is
+   * returned as-is.
+   *
+   * @param value The value to lift.
+   * @returns a string representation of the value in an inflight context.
+   * @internal
+   */
+  protected static _lift(value: any): string {
```
eladb, line 203: Maybe instead just make `liftObject()` available as a static method somewhere under `std` (`std.Inflight.liftObject()`).

Doesn't seem like there's too much value in exposing it from every resource in wing

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -62,6 +62,7 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     actionType: BucketEventType,
     key: string
   ): Promise<void> {
+    // console.log(this.topicHandlers[actionType]);
```
eladb, line 65: Remove comments

eladb (approved):
Looks like a worthy direction to explore. May the gods of our unit test and type system guide you safely to a better place.

(Gut feeling is that the non determinism of inflight identifiers could bite us in the ass, but I am willing and p give this a go).

Also, please don't forget the "wing first" tenet. Eventually our typescript api needs to be produced based on wing declarations.

libs/wingsdk/src/shared/misc.ts
MarkMcCulloh: This id is stable, or at least it's as stable (in wing) as our current id mechanism for inflight.

> if perhaps we can just hash the contents of the inflight to generate a more stable id.

The problem with using a content-based id is that we (currently) need to be able to uniquely identify inflights even if they have the same content.

> Using a global incrementing number means that multiple apps that run in the same process (eg for testing) will result in identifiers that are not deterministic

Yes and no. Due to all of this being synchronous and the preflight sandboxing we do, this might still work as expected. Additionally, for testing I think it's reasonable to expect isolation for globals like this (like how it is deterministically working in the wingsdk unit tests now).
I understand the concern though and I would prefer figure out an alternative solution.

> Alternatively, can we get rid of the id altogether somehow?

Maybe, I'm not sure. Part of my remaining TODOs is figuring out what to do with this thing, and getting rid of it would be my favorite option.

libs/wingsdk/src/core/tokens.ts
MarkMcCulloh: Just some messy get-it-working code based on the existing system, I have it as a todo in the description

libs/wingsdk/src/shared-tf/app.ts
MarkMcCulloh: There is none, just haven't reworked the tokens to make sense in that context

ekeren:
@MarkMcCulloh , is this suppose to resolve https://github.com/winglang/wing/issues/2853

MarkMcCulloh:
@ekeren Not yet, although it's a step in that direction. This changes how the SDK works but the compiler will still be creating resources. This will need to change, but it is useful to change it here first (and it's backwards compatible)

libs/wingsdk/test/target-tf-aws/api.test.ts
```diff
@@ -68,7 +68,7 @@ test("api with GET routes with common prefix", () => {
 
   // THEN
   expect(tfResourcesOfCount(output, "aws_api_gateway_rest_api")).toEqual(1);
-  expect(tfResourcesOfCount(output, "aws_lambda_function")).toEqual(2);
+  expect(tfResourcesOfCount(output, "aws_lambda_function")).toEqual(1);
```
MarkMcCulloh, line 71: API previously deduped inflights only if there were the exact same instance (referentially)

With _hash, we can understand when two inflight are equal in a non-referential way. So now this test actually dedupes these functions.

libs/wingsdk/src/core/counter.ts
```diff
@@ -0,0 +1,59 @@
+import { IConstruct } from "constructs";
+import { App } from "./app";
+
+const COUNTERS_SYMBOL = Symbol.for("@winglang/sdk.core.Counter");
+
+/**
+ * A class that generates tree-unique sequenced IDs for constructs.
+ */
+export class Counters {
```
MarkMcCulloh, line 9: The name of this obviously conflicts with the `Counter` resource, so I'm happy to hear some naming suggestions.

some other ideas:

IdGenerator
IdIncrementor
IdEnumerator

libs/wingc/src/jsify.rs
```diff
@@ -1371,6 +1372,11 @@ impl<'a> JSifier<'a> {
 			code.append(" {");
 			code.indent();
 
+			// TODO Hack to make sure closures match the IInflight contract from wingsdk
+			if class_type.is_closure() {
```
eladb, line 1376: is this still needed?

libs/wingsdk/src/core/counter.ts
eladb: i think we don't need an additional dedicated type for this. just plug it as a method of `App` and keep the state in there:

```js
App.of(this).makeId(scope, prefix)
```

Something like this...

libs/wingsdk/src/shared/convert.ts
```diff
@@ -1,52 +1,37 @@
...
 
-    constructor(theScope: Construct, theId: string, handler: IResource) {
-      super(theScope, theId);
-      this.handler = handler;
-      Node.of(this).hidden = true;
-      this.args = args;
-    }
+  const newHandler = {
+    ...baseHandler,
+    _hash: createHash("md5").update(newCode).digest("hex"),
```
eladb, line 27: do we still need the hash?

libs/wingsdk/src/std/resource.ts
```diff
@@ -143,11 +169,11 @@ export abstract class Resource extends Construct implements IResource {
...
         }
 
-        // if the object is a resource (i.e. has a "lift" method"), register a lifting between it and the host.
-        if (isResource(obj)) {
+        // if the object is a liftable resource (i.e. also has a "_addOnLift" method"), register a lifting between it and the host.
+        if (isLiftable(obj) && "_addOnLift" in obj) {
           // Explicitly register the resource's `$inflight_init` op, which is a special op that can be used to makes sure
           // the host can instantiate a client for this resource.
-          obj._addOnLift(host, [...ops, "$inflight_init"]);
+          (obj as any)._addOnLift(host, [...ops, "$inflight_init"]);
```
eladb, line 176: why do we need `as any` here?

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -274,9 +274,11 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     value: string,
     contentType?: string
   ): Promise<void> {
+    // console.log(this.objectKeys);
```
eladb, line 277: clean up

libs/wingsdk/src/target-tf-aws/queue.ts
```diff
@@ -69,8 +67,9 @@ export class Queue extends cloud.Queue implements IAwsQueue {
     );
 
     const fn = new Function(
-      this.node.scope!, // ok since we're not a tree root
-      `${this.node.id}-SetConsumer-${hash}`,
+      // ok since we're not a tree root
+      this.node.scope!,
```
eladb, line 71: i am curious why this is not just `this`...

can you raise an issue to follow up on this?

libs/wingsdk/src/core/counter.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  public createId(scope: IConstruct, id: string): string {
+    const key = `${scope.node.addr}|${id}`;
+    const counter = this._counters[key];
+    if (counter !== undefined) {
+      return `${id}${this._counters[key]++}`;
+    } else {
+      this._counters[key] = 1;
+      return `${id}0`;
+    }
```
eladb, line 57:
```suggestion
    this._counters[key] = this._counters[key] ?? 0;
    return `${id}_${this._counters[key]++}`;
```

(also add `_` between prefix and counter.

libs/wingsdk/src/core/counter.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Generate a unique ID for the given scope and id. The newly generated ID is
+   * guaranteed to be unique within the given scope.
+   * It will have the form '<id><n>', where '<id>' is the given id and '<n>' is a
+   * increasing sequence of integers starting from '0'.
+   *
+   * @param scope The scope for which this id is generated.
+   */
+  public createId(scope: IConstruct, id: string): string {
```
eladb, line 49:
```suggestion
  public createId(scope: IConstruct, prefix: string): string {
```

eladb (changes requested):
a few suggestions

libs/wingsdk/src/target-tf-aws/queue.ts
MarkMcCulloh: to avoid cyclic dependencies I believe

libs/wingsdk/src/std/resource.ts
MarkMcCulloh: TS doesn't know that _addOnLift is callable here, only that it is a property of obj. I could add a check to see that `_addOnLift` but it seemed unnecessary since it shouldn't happen. I can try changing this block to just do `if (obj instanceof Resource)` instead.

libs/wingsdk/src/shared/convert.ts
MarkMcCulloh: Yeah, there are several resources that need a way to identify inflights so that they can be deduped. (e.g. API is able to reuse cloud.Functions across routes)

libs/wingsdk/src/core/counter.ts
MarkMcCulloh: Not sure I agree with having `_`. It would be very surprising to use this for non-alpha characters, so you already have a natural separator. As a result, having _ just feels like adding noise (especially because we all use `-` in IDs)

libs/wingsdk/src/core/counter.ts
MarkMcCulloh: Is this important enough to add to the API of App?

I could also get rid of the class and make a global function `makeId(scope, prefix)` that gets the app of the scope and stores the data there the same way with a symbol.

libs/wingc/src/jsify.rs
MarkMcCulloh: See https://github.com/winglang/wing/pull/4993#discussion_r1402488116

libs/wingsdk/src/core/counter.ts
eladb: ok, no problem

libs/wingsdk/src/shared/convert.ts
eladb: ok

libs/wingsdk/src/core/counter.ts
eladb: yes, i think it is. it could be useful for creating identifiers in user-defined classes. totally see this as a general facility of wing apps.

libs/wingsdk/test/target-tf-aws/api.test.ts
eladb: are you sure the hash is safe vis-a-vis captures? can we really assume that two inflight closures are the same just based on their code?

MarkMcCulloh: The "code" includes captures, so yes it should be safe

libs/wingsdk/src/core/counter.ts
MarkMcCulloh: Added

Issue #5009, opened 2023-11-20
`wing pack` didn't discover #5008


eladb:
### I tried this:

see https://github.com/winglang/wing/issues/5008

### This happened:

`wing pack` succeeded but consumers failed.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.49.0](https://github.com/winglang/wing/releases/tag/v0.49.0).

Issue #5008, opened 2023-11-20
unable to bring directories with `-`, `.` and other non symbolic names


eladb:
### I tried this:

```sh
npm i @winglibs/containers
```

`main.w`:

```js
bring containers;
```

then:

```sh
wing compile
```

### This happened:

```
runtime error: Unexpected token '-'
  --> main.w:1:1
1 | bring containers;
  | ^
at  (/Users/eladb/playground/containers-test/main.w:1:1)
--------------------------------- ORIGINAL STACK TRACE ---------------------------------
Error: Unexpected token '-'
    at runPreflightCodeInVm (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/compiler/dist/index.js:468:11)
    at async Object.compile (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/node_modules/@winglang/compiler/dist/index.js:401:5)
    at async Object.compile (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/dist/commands/compile.js:61:16)
    at async Command.<anonymous> (/Users/eladb/.volta/tools/image/packages/winglang/lib/node_modules/winglang/dist/cli.js:47:30)
```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

eladb:
additional research shows that the culprit the names of the docker image directories in the container test directory: `my-app`, `microservices-consumer`, `microservices-producer` which contain `-`.

trying to create a minimal reproduction.

eladb:
[repro-5008.zip](https://github.com/winglang/wing/files/13413031/repro-5008.zip)

`./test/foo-bar/dummy.txt`:

```txt
nothing
```

`./main.w`

```js
bring "./test" as t;
```

eladb:
This also happens if there's a directory the starts with `.` (e.g. `.bunny`) in inside the library.

monadabot:
Congrats! :rocket: This was released in [Wing 0.49.0](https://github.com/winglang/wing/releases/tag/v0.49.0).

Issue #4929, opened 2023-11-14
Error while bringing a directory that contains a subdirectory with a leading dot


skyrpex:
### I tried this:

Given:

- `main.w`:
  ```ts
  bring "./test" as test;
  ```

- `test/.subdir/`: just an empty directory

Then:

- `wing compile`

### This happened:

```
ERROR: Unexpected token '.'

target/main.wsim.983855.tmp/.wing/preflight.js:6
   const $wing_is_test = process.env.WING_IS_TEST === "true";
   const std = $stdlib.std;
>> const prob = require("./preflight.test-2.js")({ $stdlib });
   class $Root extends $stdlib.std.Resource {
     constructor($scope, $id) {
```

### I expected this:

I expected it to compile without errors

### Component

SDK

### Wing Version

0.46.5

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.49.0](https://github.com/winglang/wing/releases/tag/v0.49.0).

PR #5017, opened 2023-11-20
fix(compiler): unable to bring modules with non-symbolic directory names

hasanaburayyan:
## Summary
1. Skip bringing directories where no wing file is found
2. Report a diagnostic error if directory contains non-symbolic character

Closes: https://github.com/winglang/wing/issues/5008
Closes: https://github.com/winglang/wing/issues/5009
Closes: https://github.com/winglang/wing/issues/4929

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/valid/bring_local_dir.test.w
+++ b/examples/tests/valid/bring_local_dir.test.w
@@ -1,5 +1,6 @@
 bring "./subdir2/inner/widget.w" as w;
 bring "./subdir2" as subdir;
+bring "./subdir-3" as hasNoWingFiles;
 
 let widget1 = new w.Widget();
 assert(widget1.compute() == 42);
--- a/examples/tests/valid/bring_local_dir.test.w
+++ b/examples/tests/valid/bring_local_dir.test.w
@@ -1,6 +1,5 @@
 bring "./subdir2/inner/widget.w" as w;
 bring "./subdir2" as subdir;
-bring "./subdir-3" as hasNoWingFiles;
 
 let widget1 = new w.Widget();
 assert(widget1.compute() == 42);
--- a/examples/tests/valid/subdir-3/blah.txt
+++ b/examples/tests/valid/subdir-3/blah.txt
@@ -1,2 +1,0 @@
-even though this directory name is invalid for bringing we will
-ignore it since there are no wing files in it
--- a/examples/tests/valid/subdir-3/blah.txt
+++ b/examples/tests/valid/subdir-3/blah.txt
@@ -1,0 +1,2 @@
+even though this directory name is invalid for bringing we will
+ignore it since there are no wing files in it
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -2,6 +2,7 @@ use camino::{Utf8Component, Utf8Path, Utf8PathBuf};
 use indexmap::{IndexMap, IndexSet};
 use itertools::Itertools;
 use phf::{phf_map, phf_set};
+use regex::Regex;
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::{fs, str, vec};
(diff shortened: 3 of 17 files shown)


examples/tests/valid/bring_local_dir.test.w
```diff
@@ -1,5 +1,6 @@
 bring "./subdir2/inner/widget.w" as w;
 bring "./subdir2" as subdir;
+bring "./subdir-3" as hasNoWingFiles;
```
eladb, line 3: this should fail, no? (because there are no .w files there)

libs/wingc/src/parser.rs
```diff
@@ -260,6 +261,42 @@ fn parse_wing_file(
...
+		} else if path.extension() == Some("w") {
+			return true;
+		}
+	}
+	false
+}
+
+fn contains_non_symbolic(str: &str) -> bool {
+	// uses the same regex pattern from grammar.js for valid identifiers
+	let re = Regex::new(r"^([A-Za-z_$][A-Za-z_$0-9]*|[A-Z][A-Z0-9_]*)$").unwrap();
```
eladb, line 282:
```suggestion
	let re = Regex::new(r"^([A-Za-z_][A-Za-z_0-9]*|[A-Z][A-Z0-9_]*)$").unwrap();
```

let's not allow `$` in directory names

examples/tests/valid/subdir-3/blah.txt
```diff
@@ -0,0 +1,2 @@
+even though this directory name is invalid for bringing we will
```
eladb, line 1: we should check that something like:

```
good_dir/my.w
good_dir/bad-dir/nothing.txt
```

then:

```js
bring "./good_dir" as g;
```

should work (because the `bad-dir` is ignored).

tools/hangar/__snapshots__/test_corpus/valid/bring_local_dir.test.w_compile_tf-aws.md
```diff
@@ -249,6 +250,17 @@ module.exports = function({ $stdlib }) {
 //# sourceMappingURL=preflight.subdir2-5.js.map
 ```
 
+## preflight.subdir3-6.js
```
eladb, line 253: this file feels like it shouldn't exist because there are no exports here (because there are no .w files in this directory).

eladb (changes requested):

eladb:
please change the title to reflect the *bug* not the solution. (in this case `unable to bring modules with non-symbolic directory names).

examples/tests/valid/bring_local_dir.test.w
hasanaburayyan: I thought we just want to skip directories with no .w files, rather than error on them

eladb: if i explicitly try to `bring` a directory that doesn't have any `.w` files i think it should be an error.

in this case, there are two issues:

1. the directory doesn't have any `.w` file.
2. if it had a `.w` file, it would have been an error because the name is invalid.

so by definition This Code Is Bad™️

hasanaburayyan: Ah I see where I confused myself. So this will need to be re-worked to only ignore sub directories of the brought directory. Otherwise the brought directory should always contain some `.w` file.

Thanks for setting me straight!

eladb: yes, this seems like a reasonable direction.

eladb:
@hasanaburayyan if by any chance this is easy to fix together with this fix, then it will be awesome: https://github.com/winglang/wing/issues/5009

i think we need to somehow run the `bring "./directory"` logic when running `wing pack`. 

i think @Chriscbr had some ideas.

examples/tests/valid/subdir-3/blah.txt
hasanaburayyan: Yupp, this scenario is now tested

hasanaburayyan:
@eladb Yea I think it would be simple enough to put both of these fixes in together. Like you said should just need to run this same logic when running `wing pack` the difference being its in the typescript so will need to replicate the checks, which is simple enough.

I need to run out for a few hours to a dinner party, Ill try adding this in tonight when Im back before I re-request another look.

eladb (approved):

PR #5021, opened 2023-11-21
fix(sdk): add missing `module` object to platform vm context

skyrpex:
CommonJS scripts can export via `exports` or `module.exports`. The latter was missing from the VM created by the platform manager.


hasanaburayyan (approved):
checks out! 🙇

eladb (approved):

Issue #4830, opened 2023-11-07
production observability 


skorfmann:
### Feature Spec

With the new release of Wing, each SDK resource comes with platform native observability options baked in. This makes it dramatically easier to observe, trace and understand the full system once deployed to one of the supported platfroms. 

### Use Cases

- make it easy to spot errors without having to trawl through logs 
- analyse performance 
- analyse request / event flow though the system
- correlate logs with request across different resources 

### Implementation Notes

potentially related to https://github.com/winglang/wing/issues/833 but I would expect that one can compile and deploy a wing application and observe it with cloud platform native tooling (e.g. xray) as well 

also slightly related to https://github.com/winglang/wing/issues/4829

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Issue #4125, opened 2023-09-09
simple queue handler / cloud function takes longer than expected


skorfmann:
### I tried this:

```
bring cloud;
bring util;

let bucket = new cloud.Bucket();
let queue = new cloud.Queue();

queue.setConsumer(inflight (message) => {
  bucket.put("wing.txt", "Hello, ${message}");
}, timeout: 3s);

test "Hello, world!" {
  queue.push("world!");

  let found = util.waitUntil(() => {
    log("Checking if wing.txt exists");
    return bucket.exists("wing.txt");
  });

  assert("Hello, world!" == bucket.get("wing.txt"));
}
```

[see example](https://github.com/winglang/examples/blob/13eb1adc24779885bf4bc975a2b2aa52f930efee/examples/hello-wing/Readme.md)

`wing test -t tf-aws main.w`

### This happened:

```
wing test --no-clean -t tf-aws main.w                                                                                                           <aws:monada-examples>
✔ Compiling main.w to tf-aws...

✔ terraform init

✔ terraform apply

✔ Setting up test runner...

✔ Running tests...

fail ┌ main.tfaws » root/Test.hiAsf5bKhU/env0/test:Hello, world!
     │ Error: Invoke failed with message: "Unhandled". Full error: "{"errorMessage":"2023-09-09T12:22:38.143Z 44cc9698-e6b2-436b-9691-016c01f6396c Task timed out after 30.07 seconds"}"
     │     at FunctionClient.invokeWithLogs (/Users/sebastian/.nvm/versions/node/v20.6.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/shared-aws/function.inflight.js:50:19)
     │     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
     │     at async TestRunnerClient.runTest (/Users/sebastian/.nvm/versions/node/v20.6.0/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/shared-aws/test-runner.inflight.js:28:41)
     │     at async /Users/sebastian/.nvm/versions/node/v20.6.0/lib/node_modules/winglang/dist/commands/test.js:314:30
     │     at async withSpinner (/Users/sebastian/.nvm/versions/node/v20.6.0/lib/node_modules/winglang/dist/util.js:55:24)
     │     at async testTfAws (/Users/sebastian/.nvm/versions/node/v20.6.0/lib/node_modules/winglang/dist/commands/test.js:311:25)
[...]

Issue #3410, opened 2023-07-13
Enable Custom esbuild config


skorfmann:
### Feature Spec

Wing now supports customizing the esbuild config for bundling inflight functions. This enables bundling functions with custom requirements, such as WebAsssembly modules or using custom esbuild plugins.

### Use Cases

- enable usage of custom esbuild plugins
- enable usage of custom loaders for  WASM modules or similar formats 

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skorfmann:
Just ran into the error described here https://answers.netlify.com/t/lambda-functions-fastify-error-cannot-find-module-string-decoder/40219/12 in one of the examples

skyrpex:
A related use case is running `cloud.Service` during `sim` and accessing modules such as `vite`, where the default esbuild bundling config fails.

eladb:
> ```
> external_node_modules = ["readable-stream"]
> ```


@skorfmann would it make sense to always build with this setting when bundling for aws lambda?

skyrpex:
Dependencies such as the `fsevents` dependency should also never be bundled AFAIK.

skorfmann:
This would help with using jsx in the Typescript SDK as well, e.g. something like this https://github.com/winglang/examples/pull/45

Chriscbr:
See also the bundling options offered by CDK here: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html

One design option could be to have some way to configure bundling on a per `cloud.Function` basis:

```js
new cloud.Function(handler,
  env: {
    "KEY": "VALUE",
  },
  bundling: {
    external: ["fsevents"],
    minify: false,
  }
);
```

Another option is to have some way to configure bundling app-wide:

```js
// inside the platform or in Wing, both can be supported
std.bundling.configure(
  external: @app.target == "aws" ? ["fsevents", "aws-sdk"] : ["fs-events"],
  minify: false,
);
```

When we expose an API for serializing an inflight into a string, bundling options would also be exposed for that naturally (this would be the lowest level).

@eladb thoughts?

eladb:
Good idea - I think it might be sufficient as an initial step just support this at the app level.

Only thing, I'd try to avoid a global and expose it on via the `@app`:

```js
@app.bundling.configure(...)
```

PR #5025, opened 2023-11-22
feat(sdk): enable xray

skorfmann:
This aims to do the following things:

- [x] wrap the aws sdk clients with [xray](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-awssdkclients.html)
- [ ] instrument outgoing [http](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-httpclients.html)

For `http`, I haven't found the right way to inject the necessary code. Tests are failing with something like this:

```
 FAIL  test/target-tf-aws/api.test.ts > api with GET routes with different prefix
Error: Failed to bundle function: Error: Build failed with 1 error:
api-onrequest-c5395e41_c8f26cae.js:2:24: ERROR: Could not resolve "aws-xray-sdk"
 ❯ createBundle src/shared/bundling.ts:54:11
 ❯ new Function src/target-tf-aws/function.ts:84:32
 ❯ Api.addInflightHandler src/target-tf-aws/api.ts:258:12
 ❯ Api.getExistingOrAddInflightHandler src/target-tf-aws/api.ts:219:17
 ❯ Api.addHandler src/target-tf-aws/api.ts:201:19
 ❯ Api.httpRequests src/target-tf-aws/api.ts:78:21
 ❯ Api.get src/target-tf-aws/api.ts:100:10
 ❯ test/target-tf-aws/api.test.ts:83:7
     81|   const inflight2 = Testing.makeHandler(app, "Handler2", INFLIGHT_CODE);
     82|
     83|   api.get("/hello/foo", inflight);
       |       ^
     84|   api.get("/foo/bar", inflight2);
```

Also, I'm not quite sure if the current approach for `http` instrumentation is the right way. I think in another project I brought in the xray sdk via a Lambda layer and excluded it from bundling. But I don't think this would help with the failing tests. Thoughts / ideas?

Related to https://github.com/winglang/wing/issues/4830

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -235,6 +236,25 @@ export class Function extends cloud.Function implements IAwsFunction {
...
+   * @returns the function code lines as strings
+   */
+  protected _getCodeLines(handler: IFunctionHandler): string[] {
+    const inflightClient = handler._toInflight();
+    const lines = new Array<string>();
+
+    lines.push('"use strict";');
+    lines.push(`const AWSXRay = require('aws-xray-sdk');`);
+    lines.push(`AWSXRay.captureHTTPsGlobal(require('http'));`);
+    lines.push("exports.handler = async function(event) {");
```
skorfmann, line 251: that's the part I'm not sure about.

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -235,6 +236,25 @@ export class Function extends cloud.Function implements IAwsFunction {
...
+   * @param handler IFunctionHandler
+   * @returns the function code lines as strings
+   */
+  protected _getCodeLines(handler: IFunctionHandler): string[] {
+    const inflightClient = handler._toInflight();
+    const lines = new Array<string>();
+
+    lines.push('"use strict";');
+    lines.push(`const AWSXRay = require('aws-xray-sdk');`);
+    lines.push(`AWSXRay.captureHTTPsGlobal(require('http'));`);
```
eladb, line 250: Would this also capture `fetch` which we are using to implement the wing `http` module?

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -235,6 +236,25 @@ export class Function extends cloud.Function implements IAwsFunction {
...
+    lines.push("exports.handler = async function(event) {");
+    lines.push(`  return await (${inflightClient}).handle(event);`);
```
eladb, line 252:
```suggestion
    lines.push(`  return (${inflightClient}).handle(event);`);
```

Not needed

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -235,6 +236,25 @@ export class Function extends cloud.Function implements IAwsFunction {
     this.addEnvironment("WING_FUNCTION_NAME", name);
   }
 
+  /**
+   * @internal
+   * @param handler IFunctionHandler
+   * @returns the function code lines as strings
+   */
+  protected _getCodeLines(handler: IFunctionHandler): string[] {
```
eladb, line 244: Where is this called from?

eladb (commented):
I don't have any idea what's the issue you are seeing with bundling

libs/wingsdk/src/target-tf-aws/function.ts
skorfmann: hm, good point - https://github.com/aws/aws-xray-sdk-node/issues/531 - no, it's not. They apparently recommend to use some open telemetry sdk. Will check.

skorfmann: looks interesting https://aws-otel.github.io/

Also, seems to be less intrusive and more generic. Will see if that's really just pluggable via ENV, layer and permissions.

eladb: yes! is it possible to use OT for everything instead of directly xray? that would be a much better and general solution.

skorfmann:
did some research yesterday, it's possible to add open telemetry without changes and get some basic instrumentation going. 

Screenshot 2023-12-04 at 23 23 10

However, the segments for AWS SDK calls are missing. It looks like we're [bundling](https://github.com/winglang/wing/blob/0ea8670963539fcf1b13a6a9c32e72179080e7d7/libs/wingsdk/src/shared/bundling.ts#L32-L53) the `aws sdk v3`, which apparently gets in the way of auto-instrumenting sdk calls to services like s3 or ddb. 

There were attempts to exclude it from bundling already https://github.com/winglang/wing/pull/4152 which led to drastically increased cold starts https://github.com/winglang/wing/issues/4125 - Perhaps there's a way to get around the performance issues. Will give it a quick try.

btw: that's a case where https://github.com/winglang/wing/issues/3410 would be super helpful to quickly iterate and upstream once it's working.

skorfmann:
oh, and there's a tax on the cold start times for adding the otel agent. Roughly in the range of 0.5 - 1s

Chriscbr:
> oh, and there's a tax on the cold start times for adding the otel agent. Roughly in the range of 0.5 - 1s

Oof... any way to avoid this?

skorfmann:
> for reference, here's a recent issue in the aws skd v3 repo talking about the cold start issues with unbundled sdk. ~ 1s added at 1024mb [aws/aws-sdk-js-v3#5382](https://github.com/aws/aws-sdk-js-v3/issues/5382) - so I'm assuming it gets way worse with less memory.

There's a pretty good blog post about this by https://twitter.com/astuyve 

https://aaronstuyvenberg.com/posts/aws-sdk-comparison

skorfmann:
some more links around the aws open telemetry stuff

- build a custom layer https://github.com/open-telemetry/opentelemetry-lambda
- the original design proposal for that https://github.com/open-telemetry/opentelemetry-lambda/blob/main/docs/design_proposal.md
- and the repo for the layers published by AWS itself https://github.com/aws-observability/aws-otel-lambda

eladb:
😞

skorfmann:
for wing cloud, we're using a custom platform to define the open telemetry config, which works ok. As said above, it adds substantial cold start times though.

So, perhaps an opt-in option for functions would work?

Issue #4977, opened 2023-11-17
Unable to use a variable as a map key


eladb:
### I tried this:

```js
let x = "mykey";
let map = { x => "hello" };
```

### This happened:

```
Failed to compile.

error: Unknown parser error
  --> wing/main.w:3:15
  |
3 | let map = { x => "hello" };
  |               ^ Unknown parser error


error: Expected type to be "num", but got "str" instead
  --> wing/main.w:3:13
  |
3 | let map = { x => "hello" };
  |             ^ Expected type to be "num", but got "str" instead


error: Expected type to be "num", but got "str" instead
  --> wing/main.w:3:18
  |
3 | let map = { x => "hello" };
  |                  ^^^^^^^ Expected type to be "num", but got "str" instead

```

### I expected this:

no error

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

yoav-steinberg:
Supporting this means we'll need to implement the map literal as a loop of insertions into the js object. And if two expressions/variables evaluate to the same thing the last one in the literal will be applied:
```wing
let x = "mykey";
let y = "mykey";
let map = { x => "hello", y => "world" };
log(map.get("mykey")); // logs "world"
```

monadabot:
Congrats! :rocket: This was released in [Wing 0.50.4](https://github.com/winglang/wing/releases/tag/v0.50.4).

PR #5046, opened 2023-11-24
feat(compiler): allow map keys to be any string expression

yoav-steinberg:
see #4977
Now any expression that evaluates to a string can be used as as a map key:
```wing
let m = { "string_lit" => 1, "{1+1}" => 2, some_func() => 3, some_var => 4};
```
If two expression eval to the same string the last one is applied.

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -2372,21 +2372,26 @@ impl<'a> TypeChecker<'a> {
...
+					// 	if let JsonDataKind::Fields(ref mut fields) = kind {
+					// 		fields.insert(
+					// 			key.clone(),
+					// 			SpannedTypeInfo {
+					// 				type_: t,
+					// 				span: value.span(),
+					// 			},
+					// 		);
+					// 	}
+					// }
```
yoav-steinberg, line 2387: @MarkMcCulloh can you explain this? Not sure what to do here..

libs/wingc/src/jsify/snapshots/access_property_on_value_returned_from_collection.snap
```diff
@@ -78,7 +78,10 @@ class $Root extends $stdlib.std.Resource {
         super._registerOnLift(host, ops);
       }
     }
-    const s = ({"hello": "123"});
```
eladb, line 81: Instead of this imperative code, can we use this syntax:

```js
const s = {
 ["expression" + "can" + "go" + "here"]: "123"
};
```

eladb (changes requested):
i am wondering if we can use the `{ [expr]: key }` syntax instead of the closure

libs/wingc/src/jsify/snapshots/access_property_on_value_returned_from_collection.snap
yoav-steinberg: OMG computed object properties, JS is just crazy. I'll fix.

eladb (approved):

eladb (approved):

libs/wingc/src/type_check.rs
MarkMcCulloh: This allows you to treat Map<Json> basically as a Json itself, so it keeps information about each field stored. In theory this would allow this to work implicitly:

```
struct A {
  a: str;
}

let jsonVal = Json "blah";
let x = { "a" => jsonVal };

let y: A = x;
//^ This casting works
```

However... this does not actually seem to work right now and I doubt anyone cares about it. So I'd say go ahead and delete it and it can be addressed another day

docs/docs/03-language-reference.md
```diff
@@ -92,7 +92,7 @@ These don't affect the value of the number or how they are printed:
 | Name          | Extra information                               |
 | ------------- | ----------------------------------------------- |
 | `Array<T>`    | variable size array of a certain type           |
-| `Map<T>`      | map type (key-value with string keys)           |
+| `Map<T>`      | map type (key-value with string keys, keys may be any expression evaluating to a string)|
```
MarkMcCulloh, line 95: IMO the syntax feels off for a couple reasons. For one, maps look similar to json objects but have widely different semantics. I feel like that will throw people off. This is more a point about how I think we should rid of the special map syntax, but if not we should at least avoid surprises:

```
let a = "stuff";
{ a => 2, b => 2 };
{ a: 2, b: 2 };
```

If we ever add punning, this incongruity would get even worse.


Additionally, allowing arbitrary expressions is way too much freedom in something that should be simple.
I can understand wanting string interpolation here, which already would allow for arbitrary expressions, but at least it signals to the user that they should keep their expression brief (like they are probably doing in all interpolations)

libs/wingc/src/type_check.rs
yoav-steinberg: Ok, that's what I thought it was supposed to do but it didn't work so I asked. I'll remove it for now.

docs/docs/03-language-reference.md
yoav-steinberg: I might be nitpicking but the above example should be:
```wing
let a = "stuff";
{ a => 2, "b" => 2 };
{ a: 2, b: 2 };
```
I think the motivation for moving to `=>` was to better differentiate between the two.
Anyway this seems like a language design issue. @eladb WDYT?

MarkMcCulloh: > I might be nitpicking but the above example should be:

The fact that this mistake was so easy kinda proves the point (or I'm just dumb)

I'm fine with this PR going through and moving the bigger discussion to another time because I think we should just remove the map (and set) literals anyway. Maps are not important enough to have their own literal when we can support implicit conversion from Json -> Map or explicit initialization ala structs:

```
let a = Map<num> { a: 1 } // This doesn't work right now, but it could
let b: Map<num> = { a: 1 } // This already works
```

yoav-steinberg: I see, so part of that future discussion can be how do we provide a good syntax for expression json object keys (better than `{ [ expr ]: value}`, maybe just support string interpotations in keys is good enough for json.

PR #5061, opened 2023-11-26
test: parse decimal number string

olivernybroe:
Add a test for parsing decimal number strings to num.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/sdk_tests/std/number.test.w
+++ b/examples/tests/sdk_tests/std/number.test.w
@@ -12,7 +12,9 @@ test "fromJson" {
 // fromStr
 
 assert(num.fromStr("42") == 42);
+assert(num.fromStr("42.42") == 42.42);
 
 test "fromStr" {
   assert(num.fromStr("888") == 888);
+  assert(num.fromStr("42.42") == 42.42);
 }


eladb (approved):

staycoolcall911:
Hey @olivernybroe looks like this PR is failing - does it work when you test it locally? I'm available and happy to assist in fixing this

MarkMcCulloh:
@olivernybroe The fix for this should be super quick if you want to throw it in this PR:

https://github.com/winglang/wing/blob/main/libs/wingsdk/src/std/number.ts#L28

change `parseInt` here to `new Number`

libs/wingsdk/src/std/number.ts
```diff
@@ -25,7 +25,7 @@ export class Number {
   /**
    * Parse a number from string.
    *
-   * @macro ((args) => { if (isNaN(args)) {throw new Error("unable to parse \"" + args + "\" as a number")}; return parseInt(args) })($args$)
+   * @macro ((args) => { if (isNaN(args)) {throw new Error("unable to parse \"" + args + "\" as a number")}; return new Number(args) })($args$)
```
MarkMcCulloh, line 28: My bad, JS hates me. This fails under strict equality because this is actually creating a different object than a number. I hate that there are 9000 ways to do this conversion. Anyways, this *should* work.

```suggestion
   * @macro ((args) => { if (isNaN(args)) {throw new Error("unable to parse \"" + args + "\" as a number")}; return Number(args) })($args$)
```

olivernybroe: Oh totally forgot to go back into this and review if it failed or passed. Thanks for suggesting fix!

PR #5074, opened 2023-11-28
fix: cannot call methods from base interfaces

### LINKED ISSUE #5073: Cannot call methods from base interfaces
### I tried this:

```js
bring expect;

interface IFoo {
  foo(): str;
}

interface IBar extends IFoo { }

class Baz impl IBar {
  pub foo(): str { return "foo"; }
}

let baz: IBar = new Baz();
expect.equal(baz.foo(), "foo");
                 ^^^ Cannot access static property "foo" from instance
```

### This happened:

```
Cannot access static property "foo" from instance
```

### I expected this:

_No response_

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

eladb:
Fixes #5073 by setting inherited members to non static (it was a mistake I am assuming).

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/valid/inheritance_interface.test.w
+++ b/examples/tests/valid/inheritance_interface.test.w
@@ -1,0 +1,18 @@
+bring expect;
+
+interface IFoo {
+  foo(): str;
+}
+
+interface IBar extends IFoo {
+  bar(): str;
+}
+
+class Baz impl IBar {
+  pub foo(): str { return "foo"; }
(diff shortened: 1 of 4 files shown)


staycoolcall911 (approved):
Yeah that seems like a mistake. Extended interface's members should not be considered static

PR #5095, opened 2023-11-29
feat: `cloud.Endpoint`

### LINKED ISSUE #5096: cloud.Endpoint
The `cloud.Endpoint` resource represents a publicly accessible endpoint of the app.

The following resources define endpoints automatically:

1. `cloud.Website`
2. `cloud.Api`
3. `containers.Workload`
4. `websockets`

Any other future sdk or user defined resources that expose a port.

As a convenience, endpoints should be emitted as Terraform outputs when compiling to terraform, or CFN outputs when compiling to CFN.

The Simulator resource should expose an API for listing all application endpoints.

eladcon:
Adds `cloud.Endpoint` to represent endpoints in apps. `cloud.Api` and `cloud.Website` now use endpoints to export the url property. Outputs the endpoints for the relevant platform.  

Closes #5096 

## Checklist

- [X] Tests added (always)
- [X] Docs updated (only required for features)


docs/docs/04-standard-library/01-cloud/endpoint.md
```diff
@@ -0,0 +1,72 @@
...
+
+The `cloud.Endpoint` resource holds a URL and outputs it.
+
+## Usage
+
+```ts playground
+bring cloud;
+
+let endpoint = new cloud.Endpoint("https://example.com");
+```
```
Chriscbr, line 26: I'm not sure I understand the reasoning for renaming this resource from `cloud.Output` to `cloud.Endpoint`. Can you elaborate @eladcon @eladb?

As a user, when I see `new cloud.Endpoint("http://example.com")`, as a user it looks as though the app is going to perform some work to expose a resource (file, api, socket, etc.) on the networked address, just like how `cloud.Api` does. But that doesn't seem to be how the resource works.

The intent of the original `cloud.Output` API was to _expose simple data as the **output** of a Wing program_. I think this is a more natural mental model. Developers are familiar with the concepts of inputs and outputs. The name "output" will also make the resource easier to pick up for users coming from other tools (see CloudFormation Outputs, Pulumi outputs, Terraform outputs, etc.). I'm not sure if there's enough differentiated functionality here to justify deviating from the existing norms.

Chriscbr (changes requested):

docs/docs/04-standard-library/01-cloud/endpoint.md
eladb: An endpoint is more than just an output. Our use case is to be able to show endpoints in the console, expose them from preview environments and from local dev machines, maybe bind them to domains, etc.

So they need to be explicitly modeled.

I am not against also introducing `cloud.Output` (now or later) and define an output for all endpoints, but it's not required for the endpoint use case.

I hope that makes sense.

eladb: Created https://github.com/winglang/wing/issues/5096 to cover endpoints and renamed back the original issue to Output

docs/docs/04-standard-library/01-cloud/endpoint.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Target-specific details
+
+### Simulator (`sim`)
+
+The sim implementation of `cloud.Endpoint` outputs the endpoint URL.
```
eladb, line 32: Outputs where?

I think we also need to add something like `sim.listEndpoints()`

docs/docs/04-standard-library/01-cloud/endpoint.md
```diff
@@ -0,0 +1,72 @@
...
+
+### AWS (`tf-aws`)
+
+The AWS implementation of `cloud.Endpoint` uses a [Terraform Output](https://developer.hashicorp.com/terraform/language/values/outputs).
+
+### AWS (`awscdk`)
+
+🚧 Not supported yet
```
eladb, line 40: CFN output

libs/wingsdk/src/cloud/endpoint.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export const ENDPOINT_FQN = fqnForType("cloud.Endpoint");
+
+/**
+ * A cloud Endpoint
+ * @abstract
+ */
+export class Endpoint extends Resource {
+  /** @internal */
+  protected _url!: string;
```
eladb, line 16: Why does this need to be protected?

libs/wingsdk/src/target-sim/api.ts
```diff
@@ -23,6 +23,8 @@ export class Api extends cloud.Api implements ISimulatorResource {
 
   constructor(scope: Construct, id: string, props: cloud.ApiProps = {}) {
     super(scope, id, props);
+
+    new cloud.Endpoint(this, "Endpoint", this.url);
```
eladb, line 27: I think this should be at the cloud.Api level, so it applies to all targets.

libs/wingsdk/src/target-tf-aws/endpoint.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * AWS implementation of `cloud.Endpoint`.
+ */
+export class Endpoint extends cloud.Endpoint {
+  constructor(scope: Construct, id: string, url: string) {
+    super(scope, id, url);
+
+    new TerraformOutput(this, "Url", {
+      value: this._url,
```
eladb, line 16:
```suggestion
      value: this.url,
```

Any reason to access the internal?

libs/wingsdk/src/target-tf-aws/endpoint.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+  }
+
+  /** @internal */
+  public onLift(host: IInflightHost, ops: string[]): void {
+    if (!(host instanceof Function)) {
+      throw new Error("apis can only be bound by tfaws.Function for now");
```
eladb, line 23: "apis"?

libs/wingsdk/src/target-tf-aws/website.ts
```diff
@@ -153,6 +153,8 @@ export class Website extends cloud.Website implements aws.IAwsWebsite {
     }
 
     this._url = `https://${distribution.domainName}`;
+
+    new cloud.Endpoint(this, "Endpoint", this.url);
```
eladb, line 157: Same note. Should go at the cloud.Website level.

eladb (approved):

docs/docs/04-standard-library/01-cloud/endpoint.md
Chriscbr: Ok - thanks for explaining. Yes I think separating them (and/or having each Endpoint create an Output) sounds ok to me.

I'm curious to see how these resources that only have a semantic use in the context of Wing Cloud or Wing Console  end up evolving... (the resources in `bring ui` are another example of this)

Chriscbr (approved):

docs/docs/04-standard-library/01-cloud/endpoint.md
eladb: Geeking out...

They have a semantic context, period. Wing Cloud uses these semantics to implement an experience, but any tool can leverage these semantics in the future of course.

docs/docs/04-standard-library/01-cloud/endpoint.md
```diff
@@ -0,0 +1,72 @@
+---
```
staycoolcall911, line 1: The `.md` file should be created in the https://github.com/winglang/wing/blob/main/libs/wingsdk/src/cloud/ folder, just like the other cloud resources

Issue #5087, opened 2023-11-29
`wing test` doesn't work for `tf-azure` platform


garysassano:
### I tried this:

```
test "foo" {
  throw("error");
}
```
`wing test foo.test.w --platform tf-azure`

### This happened:

```sh
✔ Compiling foo.test.w to tf-azure...

✔ terraform init

✔ terraform apply

✔ Setting up test runner...

⠋ Running tests...
✔ Running tests...

pass ─ foo.test.tfazure.787795 » root/Test.FYTLBvIFw3/env0/test:foo
✔ terraform destroy

 
 
Tests 1 passed (1)
Test Files 1 passed (1)
Duration 4m48.68s
```

### I expected this:

I expected test to fail due to error being thrown.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.51.8](https://github.com/winglang/wing/releases/tag/v0.51.8).

Issue #5080, opened 2023-11-28
Azure storage buckets are leaking in sdk tests


eladb:
### I tried this:

Our builds have started failing

### This happened:

```
creating Azure Storage Account "testfyuod5tercc8617b2e": storage.AccountsClient#Create: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="MaxStorageAccountsCountPerSubscriptionExceeded" Message="Subscription *** already contains 250 storage accounts with Standard Dns endpoints in location eastus and the maximum allowed is 250."
```

### I expected this:

I think buckets are leaking on azure

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

tsuf239:
I don't think it's solved yet

monadabot:
Congrats! :rocket: This was released in [Wing 0.51.8](https://github.com/winglang/wing/releases/tag/v0.51.8).

PR #5110, opened 2023-12-03
fix(sdk): azure test should fail on error

tsuf239:
## Description 
Fixes: https://github.com/winglang/wing/issues/5087
it should fix: #5080 too

I noticed that we're not reporting the error message when the test fails #5111

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/shared/bundling.ts
```diff
@@ -35,6 +35,7 @@ export function createBundle(entrypoint: string, outputDir?: string): Bundle {
     outdir: originalEntrypointDir,
     sourceRoot: originalEntrypointDir + "/",
     absWorkingDir: originalEntrypointDir,
+    keepNames: true,
```
Chriscbr, line 38: Maybe add a comment why this is needed

Chriscbr (approved):

Issue #4727, opened 2023-10-30
an excessive await is added when jsifying macro 


tsuf239:
### I tried this:

checked out to this branch: `revitalb/array-at-fix`
in this branch, Revital fixed the macro `array.at` to throw an error in the case of invalid input
run `pnpm wing test examples/tests/sdk_tests/std/string.test.w --no-clean`

### This happened:

the following error appears: 
```
fail ┌ string.test.wsim » root/env8/test:split()     
     │ Error: Failed to bundle function: Error: Build failed with 1 error:
     │ examples/tests/sdk_tests/std/target/test/string.test.wsim/.wing/inflight.$Closure9-1.js:10:274: ERROR: "await" can only be used inside an "async" function
     │     at ....
```
this code:
```
test "split()" {
  assert("hello;wing".split(";").at(0) == "hello");
....
}
```
is jsified wrongly to 
```
      {((cond) => {if (!cond) throw new Error("assertion failed: \"hello;wing\".split(\";\").at(0) == \"hello\"")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(((obj, args) => { if (args[0] < 0 || args[0] >= (await "hello;wing".split(";")).length) throw new Error("Index out of bounds"); return obj[args[0]]; })((await "hello;wing".split(";")), [0]),"hello")))};
```

### I expected this:

To pass with no errors.

The await is coming from here:
https://github.com/winglang/wing/blob/main/libs/wingc/src/jsify.rs#L421-L424

(verified by removing it- and the issue resolved)

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
Closed by #4704

PR #4704, opened 2023-10-29
fix(sdk): Array at will now throw an exception when i is out of bounds.

revitalbarletz:
## Checklist

- [ ] Title matches [Winglang's style guide](https://www.winglang.io/contributing/start-here/pull_requests#how-are-pull-request-titles-formatted)
- [ ] Description explains motivation and solution
- [ ] Tests added (always)
- [ ] Docs updated (only required for features)
- [ ] Added `pr/e2e-full` label if this feature requires end-to-end testing

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Wing Cloud Contribution License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md)*.

PR #5120, opened 2023-12-04
fix(sdk): awaiting client object causes error in sim target

Chriscbr:
I was debugging an error with @yoav-steinberg, and we had an issue while augmenting the lifting system to get the following code to work:

```js
let b = [new cloud.Bucket()];
test "foo" {
  b.at(0);
}
```

We found that the generated JavaScript looked like this:

```js
await b.at(0);
```

Ordinarily this would no-op (since calling "await" on non-promises usually does nothing). But in the Wing simulator, the bucket client is a special proxy object that automatically translates all method calls into HTTP requests. When determining if a value is a promise or not, the JS runtime tries to call "then" on objects (see https://stackoverflow.com/questions/55262996/does-awaiting-a-non-promise-have-any-detectable-effect), and this ended up causing an error instead of the no-op behavior.

The first major bug was that trying to call a non-existent method on the proxy object was crashing the simulator server, so that's fixed.

Beyond that, to accommodate the "no-op await" use case, I've added some special case handling so if the the client contacts the server, trying to call `then()`, and the server responds saying the method does not exist, then the client object will return itself back. Of course, we probably shouldn't generate `await b.at(0)` since `at` is a synchronous method on `Array`, but I don't think this does any harm.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/shared/sandbox.ts
+++ b/libs/wingsdk/src/shared/sandbox.ts
@@ -42,6 +42,10 @@ export class Sandbox {
       sandboxConsole[level] = (...args: any[]) => {
         const message = util.format(...args);
         this.options.log?.(false, level, message);
+        // also log to stderr if DEBUG is set
+        if (process.env.DEBUG) {
+          console.error(message);
+        }
       };
     }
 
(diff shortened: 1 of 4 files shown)


MarkMcCulloh (approved):

staycoolcall911:
This is great @Chriscbr. I believe it also fixes #4727, which is blocking this PR: #4704

Issue #3593, opened 2023-07-24
Cannot reference an inflight value within a preflight expression, even though it's valid


Chriscbr:
### I tried this:

```js
bring math;

let ALPHANUMERIC_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";

class IdGenerator {
  // Generate a random 5-character string
  inflight _makeId(): str {
    let var id = "";
    for i in 0..5 {
      let randomIndex = math.floor(math.random() * ALPHANUMERIC_CHARS.length);
      id = id + ALPHANUMERIC_CHARS.at(randomIndex);
    }
    return id;
  }
}
```

### This happened:

```
error: Cannot reference an inflight value from within a preflight expression
   --> id_gen.w:10:39
   |
10 |       id = id + ALPHANUMERIC_CHARS.at(randomIndex);
   |                                       ^^^^^^^^^^^ Cannot reference an inflight value from within a preflight expression
```

### I expected this:

No error

### Is there a workaround?

Move the constant inside of the inflight method:

```js
bring math;

class IdGenerator {
  // Generate a random 5-character string
  inflight _makeId(): str {
    let ALPHANUMERIC_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
    let var id = "";
    for i in 0..5 {
      let randomIndex = math.floor(math.random() * ALPHANUMERIC_CHARS.length);
      id = id + ALPHANUMERIC_CHARS.at(randomIndex);
    }
    return id;
  }
}
```

### Component

Compiler

### Wing Version

0.24.46

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.54.14](https://github.com/winglang/wing/releases/tag/v0.54.14).

PR #5158, opened 2023-12-06
fix(compiler): can't pass inflight arg to phase independent method on lifted object

yoav-steinberg:
see #3593
When the type checker needs to assign a concrete phase to a phase independent property instead of using the object's phase we now take into account cases where we need the property to be inflight even if the object is a lifted preflight instance. If the phase independent method receives any inflight inputs it'll need to be executed inflight and therefore treated as an inflight method. This means that if the object is an inflight object **or** if the property is a method call and any of the arguments passed are inflight expressions then the property will be treated as an inflight property.

### Mechanism for lifting non resource std lib classes to inflight
Once I did the above change we encountered problem when accessing some basic std types inflight (Duration/Datetime) because now only part of the type was lifted instead of the entire expressions:
```wing
let x = datetime.systemNow();
test "mytest" { 
  x.toIso() // Now only `x` is lifted and `toIso()` is executed inflight
}
```
The problem is that `Datetime` objects weren't being lifted. To fix this I added a mechanism where we can mark std classes as `ILiftable` even if they aren't a resource. This means that when used the std lib code is bundled into inflight and the relevant object (`Datetime`/`Duration`) is serialized into inflight and deserialized inflight and can be used freely.
This also cleans up special case lifting code out of `_liftObject` which had some special cases for `Duration` which were broken anyway (for example when trying to access `milliseconds` field in a `Duration`).

### Better error when accessing preflight members of inflight (lifted) object:
`Can't access preflight member \"{property}\" on inflight instance of type \"{instance_type}\"`
Previously `lifted_object.preflight_field` would pass type checking and error duing the lifted code with a more general (can't mix inflight expression in preflight expression) error.

### Calling phase independent static methods with inflight args
This is now also supported and will result in the phase independent method to be executed inflight.

### Binary expression / string interpolation evaluation returns correct expr phase
Before something like `a + b` would either return the phase of `a` or `b` or `Phase::Independent` (depending on the type of `a` and `b` and which operator was used). This was pretty random. We now return the appropriate phase based on the phases of `a` and `b` using a new `combine_phases` function. This was required so when such expressions are passed as arguments to a method we need to know if the resulting expression can be lifted or includes inflight components in it.

### More correct errors on failure to qualify a lift
We now report a diagnostic when we encounter access to a property of an **inflight** expression that evaluates to an object of a preflight class type. This is more correct than reporting a qualification error on any non-property reference to a preflight object, which wasn't good enough anymore given the other changes in this PR. Details in this comment: https://github.com/winglang/wing/pull/5158#discussion_r1435802072

## Checklist

- [x] Tests added (always)


libs/wingc/src/lifting.rs
```diff
@@ -197,6 +197,20 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
 				return;
 			}
 
+			// Inflight expressions that evaluate to a preflight type are currently unsupported because
+			// we can't determine exactly which preflight object is being accessed and therefore can't
+			// qualify the original lift expression.
```
Chriscbr, line 202: Great comment

libs/wingc/src/lifting.rs
```diff
@@ -197,6 +197,20 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
...
+			if expr_phase == Phase::Inflight && expr_type.is_preflight_class() && v.ctx.current_property().is_some() {
+				report_diagnostic(Diagnostic {
+					message: format!(
+						"Reference to lifted preflight object of type \"{expr_type}\" of unknown origin, can't qualify the lift (see https://github.com/winglang/wing/issues/76 for details)"
```
Chriscbr, line 206: I'm a little worried this error message might be difficult to understand. Is there any way to make this clearer? "unknown origin" seems cryptic

libs/wingc/src/type_check.rs
```diff
@@ -1865,399 +1883,416 @@ impl<'a> TypeChecker<'a> {
...
+		&mut self,
+		exp: &Expr,
+		env: &mut SymbolEnv,
+		expr_visit_info: ExprVisitInfo,
+	) -> (TypeRef, Phase) {
+		CompilationContext::set(CompilationPhase::TypeChecking, &exp.span);
+
+		self.curr_expr_info.push(expr_visit_info);
+
+		//let (mut t, phase) = self.type_check_exp_helper(&exp, env);
```
Chriscbr, line 1899: Remove

libs/wingc/src/type_check.rs
```diff
@@ -2266,326 +2301,347 @@ impl<'a> TypeChecker<'a> {
...
+			// // Othewise, if any of the expressions are preflight then the result is preflight
+			// (Phase::Preflight, _) | (_, Phase::Preflight) => Phase::Preflight,
+			// // Otherwise the result is independent
+			// (Phase::Independent, Phase::Independent) => Phase::Independent,
+
+			// Otherwise, if any of the expressions are independent then the result is independent
+			(Phase::Independent, _) | (_, Phase::Independent) => Phase::Independent,
+			// Othersize the result is preflight
+			(Phase::Preflight, Phase::Preflight) => Phase::Preflight,
 		}
```
Chriscbr, line 2645: Move this outside of `impl TypeChecker` since it's a pure function?

libs/wingc/src/type_check.rs
```diff
@@ -5076,7 +5141,12 @@ impl<'a> TypeChecker<'a> {
 			.map(|_| base_udt)
 	}
 
-	fn resolve_reference(&mut self, reference: &Reference, env: &mut SymbolEnv) -> (VariableInfo, Phase) {
+	fn resolve_reference(
+		&mut self,
+		reference: &Reference,
+		env: &mut SymbolEnv,
+		calee_with_inflight_args: bool,
```
Chriscbr, line 5148:
```suggestion
		callee_with_inflight_args: bool,
```

libs/wingc/src/type_check.rs
```diff
@@ -5145,9 +5215,16 @@ impl<'a> TypeChecker<'a> {
 
 				let mut property_variable = self.resolve_variable_from_instance_type(instance_type, property, env, object);
 
-				// if the object is `this`, then use the property's phase instead of the object phase
+				// Try to resolve phase independent propertie's actual phase
```
Chriscbr, line 5218:
```suggestion
				// Try to resolve phase independent property's actual phase
```

libs/wingc/src/type_check.rs
```diff
@@ -1865,399 +1883,416 @@ impl<'a> TypeChecker<'a> {
...
-								);
+					Literal::Number(_) => (self.types.number(), Phase::Independent),
+					Literal::Boolean(_) => (self.types.bool(), Phase::Independent),
+				},
+				ExprKind::Binary { op, left, right } => {
+					let (ltype, ltype_phase) = self.type_check_exp(left, env);
+					let (rtype, rtype_phase) = self.type_check_exp(right, env);
+
+					// Resolve the phase
+					let phase = Self::combine_phases(ltype_phase, rtype_phase);
```
Chriscbr, line 1924: Do we need to combine phases for array literal / collection literal expressions?

libs/wingc/src/type_check.rs
```diff
@@ -1730,9 +1733,23 @@ pub struct TypeChecker<'a> {
 
 	is_in_mut_json: bool,
 
+	// Stack of extra information we need to keep track of while visiting nested
+	// expressions during type checking.
+	curr_expr_info: Vec<ExprVisitInfo>,
```
Chriscbr, line 1738: (random thought) I'm wondering if it's also possible to implement this by leveraging the fact that every `Expr` has an `id` field -- so you could store a lookup table that maps from IDs to the extra metadata structs. You can see this approach used for `type_for_expr`, `json_literal_casts`, and `type_expressions`.

(All that said I'm not sure if it would result in any less code)

libs/wingc/src/jsify/tests.rs
Chriscbr, line 1: Does this PR fix #3593? If so maybe we can add it to the test cases, here or in the `examples` folder

Chriscbr (approved):
Super cool fix! I added some feedback and nits, but it looks great overall

libs/wingc/src/jsify/tests.rs
MarkMcCulloh: Maybe this issue as well https://github.com/winglang/wing/issues/3456

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -3009,11 +3009,11 @@ Duration <DURATION>"
...
 exports[`resource_captures.test.w 1`] = `
-"error: Cannot qualify access to a lifted object of type \\"Bucket\\" (see https://github.com/winglang/wing/issues/76 for more details)
-   --> ../../../examples/tests/invalid/resource_captures.test.w:13:13
+"error: Reference to lifted preflight object of type \\"Bucket\\" of unknown origin, can't qualify the lift (see https://github.com/winglang/wing/issues/76 for details)
+   --> ../../../examples/tests/invalid/resource_captures.test.w:15:5
    |
-13 |     let b = this.bucket;
-   |             ^^^^^^^^^^^ Cannot qualify access to a lifted object of type \\"Bucket\\" (see https://github.com/winglang/wing/issues/76 for more details)
+15 |     b.put(\\"hello\\", \\"world\\");
+   |     ^ Reference to lifted preflight object of type \\"Bucket\\" of unknown origin, can't qualify the lift (see https://github.com/winglang/wing/issues/76 for details)
```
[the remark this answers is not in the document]
yoav-steinberg: Previously qualification errors where reported when you had a preflight object referenced without any property:
```wing
let b = new cloud.Bucket();
test "test" {
  let x = b; // We're referencing b without any qualification so we used to report an error here.
  x.list(); // Even though we can't handle this because we don't know who x is this wasn't an error.
}
```
This seemed to me more of a workaround than what we really want. The reason is that in the code above assigning to `x` is the real problem, and referencing `b` without any method isn't strictly a problem because we don't need to set up any permissions when no method is being used.
Now the approach is:
```wing
let b = new cloud.Bucket();
test "test" {
  let x = b; // This is fine, but when you'll try to use x you'll get an error since we don't do data flow analysis
  x.list(); // I don't know who `x` is so we report an error
}
```
I actually **had** to make this change because of this scenario:
```wing
let a = [new cloud.Bucket()];
test "test" {
  a.at(0).list(); // `a.at(0)` is lifted, and qualified with `list()` - all's great!
  let i = 0;
  a.at(i).list(); // new: `a` is lifted, `at(i)` is evaluated inflight, `list()` is a qualification on an `Bucket` but we don't know which because it's an inflight expression.
}
```
In the `at(i)` example above, without the new error message we simply run fine on the sim and run into a permission error on AWS. I needed the more explicit and correct logic of "you're trying to access an inflight expression that evals to a preflight type which we can't trace to the origin".

> Is it basically that it's ok to lift a bucket of unknown origin as long as you're only using its preflight fields (not calling inflight methods, or eventually getter functions perhaps) on it?

Not really. If it's on unknown origin, then you're not lifting you're referencing a lift:
```wing
// inflight
let x = this.bucket; // `this.bucket` is lifted. `x` is an inflight variable referencing a lift (a client).
log(x.node.id); // here we're trying to access preflight members on a lift (a client). This should be an error!
```
I added another fix to this PR to be explicit about the error above, even though it doesn't compile anyway because it fails JSification. `x.node.id` is considered a preflight expression because `node.id` are preflight fields. So we try to lift the expression and fail because it contains a sub-expression that's inflight `x`. This produces this error:

`Cannot reference an inflight value from within a preflight expression`

I now added a more explicit check for this scenario which fails earlier during type checking:

`Can't access preflight member "node" on inflight instance of type "T"`

libs/wingc/src/lifting.rs
yoav-steinberg: Here's another go:

Expression of type {T} references an unknown lift (unknown preflight object), can't qualify the lift (see...)

Is this any better?

tools/hangar/__snapshots__/invalid.ts.snap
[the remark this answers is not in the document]
Chriscbr:
```
test "test" {
  a.at(0).list(); // `a.at(0)` is lifted, and qualified with `list()` - all's great!
  let i = 0;
  a.at(i).list(); // new: `a` is lifted, `at(i)` is evaluated inflight, `list()` is a qualification on an `Bucket` but we don't know which because it's an inflight expression.
}
```

That the compiler can now distinguish these two kinds of expressions is really impressive

libs/wingc/src/lifting.rs
Chriscbr: Sounds better. I'm wondering if we should define a term like "capabilities" in our docs to collectively describe all of these permissions and environment variables and such that are granted to inflight hosts by preflight objects. So your error message could be tweaked to:

`Expression of type {T} references an unknown preflight object, can't qualify its capabilities`

Or another alternative:

`Unable to determine the capabilities needed to use preflight object of type {T} inflight`

tools/hangar/__snapshots__/invalid.ts.snap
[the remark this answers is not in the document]
yoav-steinberg: > That the compiler can now distinguish these two kinds of expressions is really impressive

I was wracking my brain to get this right. I hope I did. The gist of it is that `x.method(args)` meas that when `method` is executed it needs access to x **and** args. So if either is an inflight expression `method` must be executed inflight. Otherwise we can go either way. 

Then I needed to strike some balance that makes sense when "going either way". At some point I had the following **inflight** code JSify to a no-op: `assert(some_lifted_str.length == 1);` Because the assert itself is phase independent and the args aren't inflight we can lift the entire `assert` (void) call (But that seemed to extreme, the assertion is evaluated preflight during lifting!).

libs/wingc/src/jsify/tests.rs
yoav-steinberg: Fixes both, I'll add the test.

libs/wingc/src/type_check.rs
yoav-steinberg: Good point. The code for these literals currently adopts the phase of the env (and doesn't default to `Phase::Independent`). This means that this will generate and error:
```wing
// preflight
//inflight
assert([bucket].at(0).list().length == 0);
//     ^ this is considered an inflight expr, but we can fix this to be a preflight expr and then it can be lifted
```
I'll open a tracking issue for this, because it doesn't seem critical.
The critical issue is if we can produce an inflight expr that the compile won't detect as inflight and then it might try to lift something that it can't. That's why I created the `combile_pahses` logic. To deal with this case:
```
// inflight
preflight_array.at(0 + inflight_var);
//                 ^^^^^^^^^^^^^^^^ this was considered before as phase independent and then we didn't know that `at` can't be lifted here.
```

yoav-steinberg: See #5359

libs/wingc/src/lifting.rs
yoav-steinberg: Makes sense to use "capabilities", for now I'll take the `Expression of type...` suggestion above. Hope that's good enough.

libs/wingc/src/type_check.rs
yoav-steinberg: This is a slightly different case since we only need the information while we're doing the type checking. Ideally I should have added it to `VisitContext` since it's actually context information used during the AST visiting process of the type checking. But it seemed to "type-checking" specific to generalize it into the `VisitContext`. I'll keep it as is for now.

libs/wingsdk/src/core/lifting.ts
```diff
@@ -1,6 +1,7 @@
 import { getTokenResolver } from "./tokens";
-import { Duration } from "../std/duration";
-import { IResource } from "../std/resource";
+//import { Duration } from "../std/duration";
```
MarkMcCulloh, line 2:
```suggestion
```

tools/hangar/__snapshots__/test_corpus/valid/resource_captures.test.w_compile_tf-aws.md
```diff
@@ -338,7 +338,6 @@ class $Root extends $stdlib.std.Resource {
               $__arr__index_______if__index___0____index____arr_length__throw_new_Error__Index_out_of_bounds____return_arr_index______this_arrayOfStr__1_: ${$stdlib.core.liftObject(((arr, index) => { if (index < 0 || index >= arr.length) throw new Error("Index out of bounds"); return arr[index]; })(this.arrayOfStr, 1))},
               $__obj__key_______if____key_in_obj___throw_new_Error__Map_does_not_contain_key_____key______return_obj_key______this_mapOfNum___k1__: ${$stdlib.core.liftObject(((obj, key) => { if (!(key in obj)) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })(this.mapOfNum, "k1"))},
               $__obj__key_______if____key_in_obj___throw_new_Error__Map_does_not_contain_key_____key______return_obj_key______this_mapOfNum___k2__: ${$stdlib.core.liftObject(((obj, key) => { if (!(key in obj)) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })(this.mapOfNum, "k2"))},
-              $_this_myOptStr_______: ${$stdlib.core.liftObject((this.myOptStr ?? ""))},
```
MarkMcCulloh, line 341: Why did the `?? ""` get removed here?

yoav-steinberg: Note that this:
```js
 $_this_myOptStr_______: ${$stdlib.core.liftObject((this.myOptStr ?? ""))},
```
was replaced by this:
```js
$this_myOptStr: ${$stdlib.core.liftObject(this.myOptStr)},
```
Previously the type_checker, pretty arbitrarily, evaluated `??` expressions to the phase of the left expression. So `??` expression this inflight code:
```wing
  pub inflight testCaptureOptional() {
    assert(this.myOptStr ?? "" == "myOptString");
  }
```
was evaluated to a preflight expression and therefore lifted the entire expression. This worked, but it's a bug, imagine this case:
```wing
  pub inflight testCaptureOptional() {
    let x = "";
    assert(this.myOptStr ?? x == "myOptString");
  }
```
Now the expression lifting will fail because there's an inflight variable inside of the preflight expression.
Our new code fixes this by correctly evaluating the binary `??` expression's phase. Since `""` is phase independent we the entire expression is phase independent can we only need to lift the inner `this.OptStr` part of it.

Chriscbr:
woot! Great work

PR #5163, opened 2023-12-07
fix(sdk): unnecessary stderr from `docker inspect`

### LINKED ISSUE #5116: don't show docker inspect stderr
When pulling images the sim sdk will perform `docker inspect` to check if the requested image is available locally, If it is not, it will `docker pull` it.

When the image is not available, the following output will be sent to stderr:
```
Error: No such object: redis@sha256:e50c7e23f79ae81351beacb20e004720d4bed657415e68c2b1a2b5557c075ce0
```

We should not show this stderr, because it is misleading the user that there is a problem although it is a happy flow

_Originally posted by @ekeren in https://github.com/winglang/wing/pull/5112#discussion_r1413465005_

garysassano:
Closes #5116

### DIFF
--- a/libs/wingsdk/src/shared/misc.ts
+++ b/libs/wingsdk/src/shared/misc.ts
@@ -27,8 +27,10 @@ export async function runCommand(cmd: string, args: string[]): Promise<any> {
   const raw = await new Promise((resolve, reject) => {
     execFile(cmd, args, (error, stdout, stderr) => {
       if (error) {
-        console.error(stderr);
-        reject(error);
+        if (!(cmd === "docker" && args.includes("inspect"))) {
+          console.error(stderr);
+          reject(error);
+        }
       }
       resolve(stdout);
     });
--- a/libs/wingsdk/src/shared/misc.ts
+++ b/libs/wingsdk/src/shared/misc.ts
@@ -27,8 +27,8 @@ export function normalPath(path: string) {
  * Just a helpful wrapper around `execFile` that returns a promise.
  */
 export async function runCommand(cmd: string, args: string[]): Promise<any> {
-    const { stdout } = await execFilePromise(cmd, args);
-    return stdout;
+  const { stdout } = await execFilePromise(cmd, args);
+  return stdout;
 }
 
 export interface runDockerImageProps {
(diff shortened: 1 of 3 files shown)


libs/wingsdk/src/shared/misc.ts
```diff
@@ -27,8 +27,10 @@ export async function runCommand(cmd: string, args: string[]): Promise<any> {
   const raw = await new Promise((resolve, reject) => {
     execFile(cmd, args, (error, stdout, stderr) => {
       if (error) {
-        console.error(stderr);
-        reject(error);
+        if (!(cmd === "docker" && args.includes("inspect"))) {
+          console.error(stderr);
+          reject(error);
+        }
```
MarkMcCulloh, line 33: Because this is a general utility function, we probably shouldn't hardcode specific cases like this.
The root issue here is that this function has a side-effect (`console.error`). It's probably be better to remove that and instead return a more useful result. Then it can be the consumers responsibility to log or not.

This whole function could just be a simple wrapper around `promisify(execFile)` ([doc](https://nodejs.org/api/util.html#utilpromisifyoriginal)). The promise version of execFile has everything we could need in both the successful and rejected cases.

libs/wingsdk/src/shared/misc.ts
```diff
@@ -24,16 +27,14 @@ export function normalPath(path: string) {
...
-  });
-  return raw;
+  try {
+    const { stdout } = await execFilePromise(cmd, args);
+    return stdout;
+  } catch (error) {
+    if (error instanceof Error) {
+      throw error;
+    }
+  }
```
MarkMcCulloh, line 37: Not sure there's value in rethowing since the current callers already are already handling errors

```suggestion
    const { stdout } = await execFilePromise(cmd, args);
    return stdout;
```

libs/wingsdk/src/shared/misc.ts
```diff
@@ -24,16 +27,14 @@ export function normalPath(path: string) {
  * Just a helpful wrapper around `execFile` that returns a promise.
  */
 export async function runCommand(cmd: string, args: string[]): Promise<any> {
-  const raw = await new Promise((resolve, reject) => {
-    execFile(cmd, args, (error, stdout, stderr) => {
-      if (error) {
-        console.error(stderr);
```
MarkMcCulloh, line 30: Since we're no longer calling console.error, we should make sure we aren't breaking any existing usages of runComand. i.e. if they were expecting this to write the error we'll need to ensure the error still get's written somewhere

MarkMcCulloh (approved):

PR #5164, opened 2023-12-07
feat(sdk): simulator state

Chriscbr:
This PR adds the capability for the Wing simulator to store resource state across compilations. This means if you performed a lot of work to fill a `cloud.Bucket` with objects or values in `cloud.Counter`, it will be remembered when you recompile and reload the application as long as the resource is still in your app. This will be extended to other resources like Queue and Workload in the future.

Here's a screen recording - notice how even as the app is updated with new resources or different inflight code, the state of resources doesn't go away.

https://github.com/winglang/wing/assets/5008987/a877e0f1-8085-4264-bf32-1ac7c971af4c

Notes:
- The cloud.TestRunner resource is now only added to a Wing app's construct tree when it is compiled in test mode (with `wing test`).
- The Wing Console now runs two copies of the app, one compiled in the default mode, and one compiled with tests. The resource graph only shows the default app, but the tests panel shows a list of all tests from the app compiled with the simulator, and clicking any of the tests will run them with fresh state every time.

Future work:

- https://github.com/winglang/wing/issues/5197
- https://github.com/winglang/wing/issues/5198

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -412,6 +423,14 @@ export class Simulator {
     return treeData.display?.ui ?? [];
   }
 
+  private getResourceStateDir(resourcePath: string) {
+    const hash = createHash("sha256")
```
eladb, line 427: For later - We should include the resource `addr` in the simulator.json so there won't be a need to calculate a hash every time.

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -589,8 +608,18 @@ export class Simulator {
...
     const resourceObject = new ResourceType(resolvedProps, context);
 
-    // go ahead and initialize the resource
-    const attrs = await resourceObject.init();
+    const resourceStateDir = this.getResourceStateDir(resourceConfig.path);
+    let attrs: any;
+    if (existsSync(resourceStateDir)) {
+      attrs = await resourceObject.init(resourceStateDir);
+    } else {
+      // if the resource state directory doesn't exist, initialize the resource without state
```
eladb, line 616: I think the robustness should come from the resource implementation and not here so I would always make the directory and pass it down to the resource, and let the resource decide on the behavior if there is no state yet.

This will force the resource to handle partial state better.

It will also allow resources to continuously persist state if they want but that means they need to know the statedir upon initialization even if it's initially empty.

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -857,14 +886,21 @@ export interface ISimulatorResourceInstance {
   /**
    * Perform any async initialization required by the resource. Return a map of
    * the resource's runtime attributes.
+   * @param dir The directory to load the resource's state from, if any.
    */
-  init(): Promise<Record<string, any>>;
+  init(dir: string): Promise<Record<string, any>>;
```
eladb, line 891:
```suggestion
  init(statedir: string): Promise<Record<string, any>>;
```

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -857,14 +886,21 @@ export interface ISimulatorResourceInstance {
...
    * Stop the resource and clean up any physical resources it may have created
    * (files, ports, etc).
    */
   cleanup(): Promise<void>;
+
+  /**
+   * Save the resource's state to the given directory.
+   * @param dir The directory the resource should create and save its state into, if it has any.
+   */
+  save(dir: string): Promise<void>;
```
eladb, line 903:
```suggestion
  save(statedir: string): Promise<void>;
```

libs/wingsdk/src/std/test.ts
```diff
@@ -43,8 +43,18 @@ export class Test extends Resource {
     Node.of(this).title = "Test";
     Node.of(this).description = "A cloud unit test.";
 
-    if (App.of(this).isTestEnvironment || App.of(this)._target === "sim") {
```
eladb, line 46: I wonder if this could be encapsulated:

```js
App.of(this)._testRunner?.addTestFunction(this, "Handler", inflight, props);
```

And basically keep track of everything inside the runner.

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -300,6 +347,6 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
   }
 
   private hashKey(key: string): string {
-    return crypto.createHash("sha512").update(key).digest("hex");
+    return crypto.createHash("sha512").update(key).digest("hex").slice(-16);
```
eladb, line 350: Why?

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -42,7 +44,39 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
     return this._fileDir;
   }
 
-  public async init(): Promise<BucketAttributes> {
+  public async init(dir?: string): Promise<BucketAttributes> {
```
eladb, line 47: Wouldn't it make more sense to simply use the state directory as the backing store of the bucket and then we persist continuously instead of snapshots?

libs/wingsdk/src/target-sim/counter.inflight.ts
```diff
@@ -25,6 +27,13 @@ export class Counter implements ICounterClient, ISimulatorResourceInstance {
 
   public async cleanup(): Promise<void> {}
 
+  public async save(dir: string): Promise<void> {
```
eladb, line 30: - instead of in memory, can we just store the values all the time into the state directory and have that be the source of truth?

Much more robust

libs/wingsdk/test/core/__snapshots__/connections.test.ts.snap
```diff
@@ -142,19 +134,6 @@ return class Handler {
   "tree.json": {
     "tree": {
       "children": {
-        "cloud.TestRunner": {
```
eladb, line 145: Nice!

eladb (changes requested):
Looks good. See a few notes.

I am okay to get this over the finish line and then think what's the best way to support things like keeping containers alive (the additional requirement we discussed yesterday).

libs/wingsdk/src/target-sim/bucket.inflight.ts
Chriscbr: Updated

libs/wingsdk/src/target-sim/bucket.inflight.ts
Chriscbr: Just an optimization for error readability. I don't think there's a super high chance of files colliding in simulator (besides there are also limits to how long file names can be on different operating systems) - we can always tweak later if needed

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: OK

libs/wingsdk/src/core/synth-roots.ts
```diff
@@ -0,0 +1,26 @@
+import type { App, AppProps } from "./app";
+import { Node, TestRunner } from "../std";
+
+export function synthRoots(app: App, props: AppProps) {
```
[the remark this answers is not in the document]
Chriscbr: Let's try it 👍  (I initially wanted this logic somewhere on `core.App` but got into some crazy cyclic dependencies error... it didn't occur me to put it on TestRunner :-) )

libs/wingsdk/src/std/test.ts
Chriscbr: Added

apps/wing/src/commands/test/test.test.ts
```diff
@@ -158,10 +155,10 @@ describe("output-file option", () => {
     expect(testName).toBe("test.test.w");
     expect(writeResultsSpy.mock.calls[0][2]).toBe(outputFile);
 
-    expect(writeFileSpy).toBeCalledTimes(2);
-    const [filePath, output] = writeFileSpy.mock.calls[1];
-    expect(filePath).toBe("out.json");
-    expect(JSON.parse(output as string)).toMatchObject(OUTPUT_FILE);
```
Chriscbr, line 164: This test was getting flakey because it's spying on all `writeFile` calls, including those made by simulator resources (not really related to the CLI logic), so I refactored it a bit

libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -217,3 +223,21 @@ function transformRoutePath(route: string): string {
   // route validation is done in the preflight file
   return route.replace(/{/g, ":").replace(/}/g, "");
 }
+
+// express v4 doesn't natively handle async request handlers, so we need to
+// wrap them in a middleware function
+function asyncMiddleware(
```
Chriscbr, line 229: The only change in this file was adding this wrapper to resolve an eslint warning from a new rule I added

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -35,9 +35,11 @@ export class Queue
   }
 
   public async cleanup(): Promise<void> {
-    clearInterval(this.intervalId);
+    await this.processLoop.stop();
   }
```
Chriscbr, line 39: I thought there might be some flakiness caused by the current polling implementation, because when `cleanup()` calls `clearInterval()`, it doesn't wait for the current `processMessages` call to finish, meaning the simulator could end up in a limbo state. To that end I tried adding a hand-written poller that let the `cleanup()` function wait for the loop to finish before the simulator continues cleaning up other resources.

libs/wingsdk/test/target-sim/__snapshots__/table.test.ts.snap
```diff
@@ -831,14 +721,7 @@ exports[`tryGet row 2`] = `
...
       {
-        "attrs": {},
-        "path": "root/cloud.TestRunner",
-        "props": {
-          "tests": {},
-        },
-        "type": "@winglang/sdk.std.TestRunner",
-      },
-      {
+        "addr": "c8e507baf046ceb6108a9fe64c08a39ccc39ffe11c",
```
Chriscbr, line 724: These addresses are stable across compiles / builds since they're just the construct node.addr

Chriscbr:
> I am okay to get this over the finish line and then think what's the best way to support things like keeping containers alive (the additional requirement we discussed yesterday).

Container support hasn't escaped my mind yet. Some ideas on my mind:
- augment `cloud.Service()` to have a more simulator specific API for deeper customization of start/stop behavior (e.g. passing you the state directory so you can read and write to it). It could also be a dedicated resource, like `sim.Service`
- as an interim solution, create a `sim.Container` resource in the SDK that automatically handles this use case (can be used by cloud.Workload under the hood). Later we can deprecate this

apps/wing-console/console/server/src/utils/compiler.ts
```diff
@@ -44,6 +44,7 @@ export const createCompiler = ({
       await events.emit("compiling");
       const simfile = await wing.compile(wingfile, {
         platform,
+        testing: true,
       });
```
Chriscbr, line 48: If the app is compiled without `testing: true`, then the console doesn't see any tests...

Chriscbr:
Adding `do-not-merge` label since there are still some DX issues in the console that need to be resolved

apps/wing-console/console/server/src/router/test.ts
```diff
@@ -17,16 +17,18 @@ const listTests = (simulator: Simulator): Promise<string[]> => {
   return testRunner.listTests();
 };
 
+const reloadSimulator = async (simulator: Simulator, logger: ConsoleLogger) => {
+  logger.log("Reloading simulator...", "console", {
+    messageType: "info",
```
eladb, line 22:
```suggestion
    messageType: "verbose",
```

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -76,10 +81,15 @@ export interface IWithTraceProps {
  */
 export interface ISimulatorContext {
   /**
-   * This directory where the compilation output is
+   * Thes directory where the compilation output is
```
eladb, line 84:
```suggestion
   * The directory where the compilation output is
```

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -613,6 +633,7 @@ export class Simulator {
   private createContext(resourceConfig: BaseResourceSchema): ISimulatorContext {
     return {
       simdir: this.simdir,
+      statedir: join(this.statedir, resourceConfig.addr),
```
eladb, line 636: Move the logic to a helper and reuse

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -865,6 +886,11 @@ export interface ISimulatorResourceInstance {
    * (files, ports, etc).
    */
   cleanup(): Promise<void>;
+
+  /**
+   * Save the resource's state.
```
eladb, line 891:
```suggestion
   * Save the resource's state into the state directory.
```

As a convenience, I am wondering if it would be helpful to pass the state directory here so resource implementers won't have to carry it from init...

libs/wingsdk/src/std/test-runner.ts
```diff
@@ -23,6 +25,47 @@ export interface TestRunnerProps {}
  * @abstract
  */
 export class TestRunner extends Resource {
+  /**
+   * Instantiate one or more copies of a tree inside of an app based
+   * on how many isolated environments are needed for testing.
+   * @internal
+   */
+  public static _createTree(app: App, Root: any) {
```
eladb, line 33: See comment above, maybe this can just be the ctor

libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -14,35 +14,52 @@ import {
...
+    this._metadata = new Map();
   }
 
   public async init(): Promise<BucketAttributes> {
+    const fileDirExists = await exists(this._fileDir);
+    if (!fileDirExists) {
+      await fs.promises.mkdir(this._fileDir, { recursive: true });
+    }
+
+    if (fs.existsSync(join(this.context.statedir, METADATA_FILENAME))) {
```
eladb, line 49: Mixing sync and async fs calls on purpose?

libs/wingsdk/src/target-sim/queue.inflight.ts
eladb: Draining :-)

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -146,6 +149,7 @@ export class Queue
           sourceType: QUEUE_FQN,
           timestamp: new Date().toISOString(),
         });
+        // TODO: replace with await fnClient.invokeAsync()
```
eladb, line 152: 👍

eladb (approved):
Few minor notes, looks great!

libs/wingsdk/src/target-sim/queue.inflight.ts
Chriscbr: It was a fun function to write with the help of ChatGPT 😄

libs/wingsdk/src/target-sim/bucket.inflight.ts
Chriscbr: ah no, just result of some weird debugging rabbit holes - will fix :)

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: 👍

libs/awscdk/src/app.ts
```diff
@@ -113,9 +108,8 @@ export class App extends core.App {
     this.synthed = false;
     this.isTestEnvironment = props.isTestEnvironment ?? false;
     registerTokenResolver(new CdkTokens());
-    this.testRunner = new TestRunner(this, "cloud.TestRunner");
 
-    this.synthRoots(props, this.testRunner);
+    TestRunner._createTree(this, props.rootConstruct);
```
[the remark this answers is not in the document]
Chriscbr: I'm not sure if this works since when the app isn't compiled as a test, the test runner isn't created

apps/wing-console/console/server/src/router/test.ts
Chriscbr: Updated

eladb (approved):

apps/wing-console/console/ui/src/ui/map-controls.tsx
```diff
@@ -36,26 +36,6 @@ export const MapControls = ({}: MapControlsProps) => {
...
-                  ) : (
-                    <EyeIcon className="w-4 h-4" />
-                  )}
-                  <span className="text-xs px-0.5">
-                    {showTests ? "Hide tests" : "Show tests"}
-                  </span>
-                </div>
-              </ToolbarButton>
-            </div>
-          )}
```
Chriscbr, line 58: FYI @skyrpex I removed this from the toolbar since it no longer does anything right now (the app being shown in the main panel is the app compiled without tests, so there are no tests to show)

eladb: 👍 never liked this button

eladb (approved):

polamoros (approved):
The changes done in the console app look good to me 👍

PR #5166, opened 2023-12-09
feat(compiler)!: only allow struct declarations at the top-level of a file

### LINKED ISSUE #5135: runtime error when defining an inflight struct
### Decided

The discussion on this issue led to the following decision:
Declaring a struct can only be done on the top-level of a program. If a struct is defined elsewhere the compiler should throw an exception: "structs must be declared at the top-level of a program".

### I tried this:

[playground](https://www.winglang.io/play/?code=dABlAHMAdAAgACIAbQBhAGkAbgAiACAAewAKACAAIABzAHQAcgB1AGMAdAAgAFAAZQByAHMAbwBuACAAewAKACAAIAAgACAAbgBhAG0AZQA6ACAAcwB0AHIAOwAKACAAIAB9AAoACgAgACAAbABlAHQAIABwACAAPQAgACAAUABlAHIAcwBvAG4ALgBmAHIAbwBtAEoAcwBvAG4AKAB7AG4AYQBtAGUAOgAgACIAZQB5AGEAbAAiAH0AKQA7AAoAIAAgAGEAcwBzAGUAcgB0ACgAcAAuAG4AYQBtAGUAIAA9AD0AIAAiAGUAeQBhAGwAIgApADsACgB9AA%3D%3D)

```ts
test "main" {
  struct Person {
    name: str;
  }

  let p =  Person.fromJson({name: "eyal"});
  assert(p.name == "eyal");
}
```

### This happened:

`wing test`

```
runtime error: Person is not defined
```

### I expected this:

to failed on compilation  with `"structs must be declared at the top-level of a program".`

### Is there a workaround?

move the Person definition to preflight

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

autow0rk:
Closes #5135 

Check if a Struct is being created in an inner scope, and throw an error to tell users to create the Struct at the top-level of a program.

- [ ] Tests added (always)

BREAKING CHANGE: structs may only be defined at the top level of a wing file


tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1700,6 +1700,16 @@ error: Unexpected return value from void function. Return type annotations are r
...
 
 
+error: Structs must be declared at the top-level of a program: Struct 'SS' is not defined at the top-level
+   --> ../../../examples/tests/invalid/inference.test.w:88:3
+   |  
+88 | /   struct SS {
+89 | |     a: Array<num>;
+90 | |     b: Array<str>;
+91 | |   }
+   | \\\\---^ Structs must be declared at the top-level of a program: Struct 'SS' is not defined at the top-level
```
MarkMcCulloh, line 1710: To make sure we're testing the right things, could you adjust https://github.com/winglang/wing/blob/main/examples/tests/invalid/inference.test.w#L88 to remove this error (moving the struct outside the closure)

libs/wingc/src/type_check.rs
```diff
@@ -3128,6 +3128,19 @@ impl<'a> TypeChecker<'a> {
 		for (inner_scope, ctx) in inner_scopes {
 			let scope = unsafe { &*inner_scope };
 			self.ctx = ctx;
+			for statement in &scope.statements {
```
MarkMcCulloh, line 3131: This logic feels a little out of place here and adds mostly unnecessary iteration. It should be possible to do the necessary check inside `type_check_struct` by verifying `env.parent` is `None`

If that doesn't work, it could be valuable enough to introduce something like `is_top_level` to ctx instead.

MarkMcCulloh (changes requested):
Could you add to https://github.com/winglang/wing/blob/main/examples/tests/invalid/structs.test.w to verify that a couple scenarios fail as expected 🙏

libs/wingc/src/type_check.rs
autow0rk: Moved the typechecking to be at the struct level and there weren't any issues

tools/hangar/__snapshots__/invalid.ts.snap
autow0rk: The inference test was updated to remove that error, and new struct tests were added to verify that structs can't be made in preflight or inflight contexts.

examples/tests/valid/structs.test.w
```diff
@@ -70,4 +70,16 @@ test "struct definitions are phase independant" {
...
+  struct PreflightStruct {
+    name: str;
+  }
+};
+
+let y = inflight (): str => {
+  struct InflightStruct {
+    name: str;
+  }
+};
```
MarkMcCulloh, line 85: This file is in the `valid` test folder, so it should only contain passing tests. These failing cases should be moved to `examples/tests/invalid/structs.test.w`

MarkMcCulloh (changes requested):

examples/tests/valid/structs.test.w
autow0rk: Moved the test cases to the proper ```invalid``` folder

examples/tests/invalid/structs.test.w
```diff
@@ -45,4 +45,18 @@ struct Showtime extends Razzle, Dazzle {}
...
+    name: str;
+  }
+};
+          
+let y = inflight (): str => {
+  struct InflightStruct {
+      // ^ Structs must be declared at the top-level of a program: Struct InflightStruct is not defined at the top-level
+    name: str;
+  }
+};
```
MarkMcCulloh, line 62: x was already declared before which caused an unrelated/distracting error here, so we should rename these

```suggestion
let inlinePreflightStruct = (): str => {
  struct PreflightStruct {
      // ^ Structs must be declared at the top-level of a program: Struct PreflightStruct is not defined at the top-level
    name: str;
  }
};
          
let inlineInflightStruct = inflight (): str => {
  struct InflightStruct {
      // ^ Structs must be declared at the top-level of a program: Struct InflightStruct is not defined at the top-level
    name: str;
  }
};
```

libs/wingc/src/type_check.rs
```diff
@@ -3472,6 +3472,15 @@ impl<'a> TypeChecker<'a> {
 		access: &AccessModifier,
 		env: &mut SymbolEnv,
 	) {
+		// Structs can't be defined in preflight or inflight contexts, only at the top-level of a program
+		if let Some(_) = env.parent {
+			self.spanned_error(
+				name,
+				format!(
+					"Structs must be declared at the top-level of a program: Struct '{name}' is not defined at the top-level"
```
MarkMcCulloh, line 3480: Avoiding repetitive wording, also changed program to file

```suggestion
					"struct {name} must be declared at the top-level of a file"
```

libs/wingc/src/type_check.rs
```diff
@@ -3472,6 +3472,13 @@ impl<'a> TypeChecker<'a> {
 		access: &AccessModifier,
 		env: &mut SymbolEnv,
 	) {
+		// Structs can't be defined in preflight or inflight contexts, only at the top-level of a program
+		if let Some(_) = env.parent {
+			self.spanned_error(
+				name,
+				format!("struct {name} must be declared at the top-level of a file"),
```
Chriscbr, line 3479:
```suggestion
				format!("struct \"{name}\" must be declared at the top-level of a file"),
```

Chriscbr (approved):

MarkMcCulloh (approved):

PR #5167, opened 2023-12-10
feat(sdk): add `shell()` and `exec()` to util module

### LINKED ISSUE #4990: util.shell()
### Feature Spec

See https://github.com/winglang/winglibs/blob/main/containers/utils.js

### Use Cases

Execute shell commands

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

garysassano:
Closes #4990

## Checklist

- [x] Tests added (always)


Chriscbr (changes requested):
Looks like a great so far @garysassano! Could you please add some tests in https://github.com/winglang/wing/tree/main/examples/tests/sdk_tests/util ? We should include examples of each utility function working in preflight and inflight if possible.

libs/wingsdk/src/util/util.ts
```diff
@@ -57,6 +205,74 @@ export interface NanoidOptions {
...
+        stderr: "",
+        status: 0,
+      };
+    } catch (error: any) {
+      return {
+        stdout: error.stdout?.toString() ?? "",
+        stderr: error.stderr?.toString() ?? "",
+        status: error.status ?? -1,
+      };
+    }
```
Chriscbr, line 257: Let's see if we can use `exec` asynchronously... something like this perhaps

```js
import { exec as nodeExec } from "child_process";
import { promisify } from "util";

const exec = util.promisify(nodeExec);
const { stdout, stderr } = await exec(command, opts);
```

garysassano: Originally, I attempted to use a promisified version of Node.js's `child_process.exec()`, but I was unable to get the error code, so I eventually decided to switch to using `execSync()` instead.

libs/wingsdk/src/util/util.ts
```diff
@@ -57,6 +208,75 @@ export interface NanoidOptions {
...
+   */
+  public static async exec(
+    program: string,
+    args: Array<string>,
+    opts?: ExecOptions
+  ): Promise<Output> {
+    const command = `${program} ${args.join(" ")}`;
+
+    try {
+      const { stdout, stderr } = await execPromise(command, opts);
```
Chriscbr, line 249: There's a small issue - Wing's `shell()` function is intended to run the user's command in a shell, but the `exec()` function should not. To run the program directly, I think we need to use [execFile](https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback) instead of exec. Can you try this and see if it's possible to get the tests passing?

execFile will allow you to pass the program and args directly (no need to concatenate the strings together)

garysassano: Essentially, the current `exec.test.w` should be renamed to `shell.test.w`. As for `execFile()`, what kind of program could we run with it in a test?

examples/tests/sdk_tests/util/exec.test.w
```diff
@@ -0,0 +1,24 @@
...
+  expect.equal(output.stdout, "Hello, Wing!");
+  expect.equal(output.stderr, "");
+  expect.equal(output.status, 0);
+}
+
+test "exec() with invalid program" {
+  let program = "no-such-program";
+  let args = [""];
+
+  let output = util.exec(program, args);
```
eladb, line 19: This should throw, no?

examples/tests/sdk_tests/util/shell.test.w
```diff
@@ -0,0 +1,59 @@
...
+  
+  expect.equal(output.stdout, "Hello, Wing!");
+  expect.equal(output.stderr, "");
+  expect.equal(output.status, 0);
+}
+
+test "shell() with invalid command" {
+  let command = "no-such-command";
+
+  let output = util.shell(command);
```
eladb, line 18: Should throw by default and we can add an option to disable `throw: false`

examples/tests/sdk_tests/util/shell.test.w
```diff
@@ -0,0 +1,59 @@
...
+
+  expect.equal(output.stdout, "");
+  expect.equal(output.stderr, "/bin/sh: 1: no-such-command: not found\n");
+  expect.equal(output.status, 127);
+}
+
+test "shell() with explicit non-zero exit status" {
+  let command = "exit 1";
+
+  let output = util.shell(command);
```
eladb, line 28: Should throw

libs/wingsdk/src/util/util.ts
```diff
@@ -1,9 +1,161 @@
...
+  /**
+   * This stream will be ignored. This is the equivalent of attaching the stream to /dev/null.
+   */
+  NULL = "ignore",
+}
+
+/**
+ * Base command options.
+ */
+export interface CommandOptions {
```
eladb, line 33: Add an option to control whether a non zero exit code results in an exception or not

libs/wingsdk/src/util/util.ts
```diff
@@ -1,9 +1,161 @@
...
+  /**
+   * Wait for the process to finish and return its output.
+   * Calling this method multiple times will return the same output.
+   */
+  public async wait(): Promise<Output> {
+    return new Promise((resolve, reject) => {
+      let stdout = "";
+      let stderr = "";
+
+      this.childProcess.stdout?.on("data", (data) => (stdout += data));
```
eladb, line 145: Feels like these should added earlier, not just if I wait.

eladb (changes requested):

eladb:
The most common use case should be to:

```js
let stdout = util.shell("command");
```

And have this command throw an exception if the command fails for any reason, including non-zero exit code.

libs/wingsdk/src/util/util.ts
```diff
@@ -1,9 +1,158 @@
...
 import { v4 } from "uuid";
 import { InflightClient } from "../core";
 import { Duration, IInflight } from "../std";
 
+const execPromise = promisify(nodeExec);
+
+/**
+ * Describes what to do with a standard I/O stream for a child process.
+ */
+export enum Stdio {
```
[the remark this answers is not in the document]
garysassano: That's what @Chriscbr defined in its [API design](https://github.com/winglang/wing/issues/4990#issuecomment-1843344472).

libs/wingsdk/src/util/util.ts
```diff
@@ -1,9 +1,161 @@
...
+  }
+
+  /**
+   * The underlying native child process object representing the spawned process.
+   */
+  private childProcess: ReturnType<typeof spawn>;
+  /**
+   * The child's OS-assigned process ID.
+   */
+  public readonly id: number;
```
eladb, line 118:
```suggestion
  public readonly pid: number;
```

libs/wingsdk/src/util/util.ts
```diff
@@ -57,6 +209,77 @@ export interface NanoidOptions {
...
+   * @returns The standard output of the shell command.
+   * @throws An error if the shell command execution fails.
+   */
+  public static async shell(
+    command: string,
+    opts?: ShellOptions
+  ): Promise<String> {
+    try {
+      const { stdout } = await execPromise(command, opts);
+      return stdout.toString();
```
eladb, line 225: This should also throw if exit code is non zero

libs/wingsdk/src/util/util.ts
garysassano: I think this should apply only to `ShellOptions`, since both `exec()` and `spawn()` return an `Output` struct which contains the error code inside, so it would be up to the user whether to throw or not based on the returned `Output.status`.

examples/tests/sdk_tests/util/exec.test.w
Chriscbr: Only the `shell()` API automatically throws on error. `exec()` lets the user decide what they want to do based on the command's output.

examples/tests/sdk_tests/util/exec.test.w
```diff
@@ -0,0 +1,64 @@
...
+
+  expect.equal(output.stdout, "");
+  expect.equal(output.stderr, "");
+  expect.equal(output.status, "ENOENT");
```
Chriscbr, line 24: This looks like a bug -- the `status` should be a number, not a string. It's odd that this is compiling

examples/tests/sdk_tests/util/shell.test.w
```diff
@@ -0,0 +1,82 @@
...
+  expect.equal(output, "tempfile.txt\n");
+}
+
+test "shell() with throw option" {
+  let command = "exit 1";
+  let opts = {throw: false};
+
+  let output = util.shell(command, opts);
+
+  expect.equal(output, 1);
```
Chriscbr, line 81: The output of `util.shell()` is a string, no?

examples/tests/sdk_tests/util/shell.test.w
```diff
@@ -0,0 +1,82 @@
...
+  let assertThrows = (expected: str, block: (): void) => {
+    let var error = false;
+    try {
+      block();
+    } catch actual {
+      assert(actual.contains(expected));
+      error = true;
+    }
+    assert(error);
+  };
```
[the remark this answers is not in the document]
garysassano: That's a thing in many test files. If you define `assertThrows()` outside the test blocks, then you would get the following error in Wing:
```
Cannot call into preflight phase while inflight
```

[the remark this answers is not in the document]
Chriscbr: @garysassano If you move it outside of `test` then you'll need to annotate the function as inflight for it to work:

```js
let assertThrows = inflight (expected: str, block: (): void) => { ... };
```

examples/tests/sdk_tests/util/exec.test.w
garysassano: I thought the same, but that's what is actually being returned as exit status.

examples/tests/sdk_tests/util/shell.test.w
garysassano: It's actually weird that this doesn't throw since expect.equal implements node's `deepStrictEqual()`.

garysassano: It's actually weird that this doesn't throw since `expect.equal() `implements node's `deepStrictEqual()`.

garysassano: It's actually weird that this doesn't throw since `expect.equal` implements node's `deepStrictEqual()`.

garysassano: It's actually weird that this doesn't throw since `expect.equal()` implements node's `deepStrictEqual()`.

garysassano:
@Chriscbr I think it would be better to remove `spawn()` from this PR, what you think?

examples/tests/sdk_tests/util/exec.test.w
eladb: I have to say that I think that a non-zero exit code and an invalid program are not the same situation. I would argue that for `exec()`, the non-zero exit code should be a "success" (e.g. return the exit code in `status`) and ENOENT should be an exception.

Chriscbr: I see, it sounds like you're saying that the host machine not being able to find the program (like `no-such-program`) is a different error case than a program running and returning a non-zero exit code. Throwing an exception for the first situation makes sense to me. 👍

@garysassano perhaps we can add special handling that checks if the error code is "ENOENT" and throws in that one case?

Chriscbr:
@garysassano I'm ok with taking out `spawn()` to reduce the scope, sure 👍

examples/tests/sdk_tests/util/exec.test.w
garysassano: @Chriscbr Should be ready to 🚢 now.

examples/tests/sdk_tests/util/shell.test.w
```diff
@@ -0,0 +1,117 @@
...
+  } else {
+    command = "echo %WING_TARGET%";
+  }
+  let opts = {
+    inheritEnv: true,
+  };
+
+  let output1 = util.shell(command);
+  let output2 = util.shell(command, opts);
+
```
Chriscbr, line 85: Wondering if we can add something like:

```
let target = util.env("WING_TARGET");
assert(output1.includes(target) == false);
assert(output2.includes(target) == true);
```

Chriscbr (approved):
Awesome addition! 🎉

eladb (approved):

Issue #4972, opened 2023-11-16
Can't early return in a `cloud.Service`


skyrpex:
### I tried this:

```ts
bring cloud;

new cloud.Service(inflight () => {
  return;
});
```

### This happened:

```
Failed to compile.

error: Expected type to be "inflight (): void?", but got "void" instead
  --> wing/main.w:7:3
  |
7 |   return;
  |   ^^^^^^^ Expected type to be "inflight (): void?", but got "void" instead
```

### I expected this:

To be able to early return.

### Is there a workaround?

Yes, change it to `return nil;`

### Wing Version

0.47.13

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
I'm not sure this is a bug. The compiler error is saying that it expects an optional value to be returned, and you're not returning any value. The fix is to return a function, or to return nil:

```js
new cloud.Service(inflight () => {
  return nil;
});
```

Maybe we can improve the error message instead?

eladb:
@Chriscbr I actually tend to agree that void can be considered a subtype of any T?. It's pragmatic and useful and I don't think I see any potential harm, no?

Chriscbr:
Adding some additional context to my earlier comment - here's an example of some Wing code highlighting the difference between `void` and `nil`:

```js
bring cloud;

let f = (x: str): void => {
  log(x);
};

let foo = f("hi"); // error: Cannot assign expression of type "void" to a variable
```

eladb:
@Chriscbr I am not sure I fully understand the example, in this example, the compiler infers the type of `foo` to be `void` and as such it can emit this error.

I guess this might be a bit odd:

```js
let foo: str? = f("hi");
assert(foo == nil);
```

But I am not sure this has a soundness problem... Am I missing something?

Chriscbr:
In my mind there are at least a few reasons why the distinction is useful... but I'll let you be the judge. 👇 

1. The ability to distinguish functions that do and don't return a value seems handy as an API design tool. When you write a function that doesn't return a value, you're self-documenting that "hey, this function performs a side effect".  Changing void-returning functions to be assignable to variables negates this.
[...]

PR #5183, opened 2023-12-12
feat(compiler): early return in `cloud.Service()`

autow0rk:
Addresses #4972 

Adds early returning in `cloud.Service()` by considering void to be a subtype of the optional<t> type in the typechecker.

The transpiled code functions properly with 'return;' in cloud.Service's handler function:

```js
async handle() {
  return;
}
```

**Not sure how this could be fixed from the WingSDK side, so fixing it from the typechecker side was my solution.**

- [ ] Tests added (always)

### DIFF
--- a/examples/tests/valid/optionals.test.w
+++ b/examples/tests/valid/optionals.test.w
@@ -204,3 +204,13 @@ if let s1 = str1 {
 } elif let s2 = str2 {
   assert(true);
 }
+
+let functionReturningNothing = (): void? => {
+  return;
+}
+
+if let nothing = functionReturningNothing {
+  assert(false);
+} else {
+  assert(true);
+}
--- a/examples/tests/valid/optionals.test.w
+++ b/examples/tests/valid/optionals.test.w
@@ -205,7 +205,7 @@ if let s1 = str1 {
   assert(true);
 }
 
-let functionReturningNothing = (): void? => {
+let functionReturningNothing = (): str? => {
   return;
 }
 
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -756,9 +756,9 @@ impl Subtype for Type {
 			(Self::Nil, Self::Optional(_)) => {
 				// Nil is a subtype of Optional<T> for any T
 				true
-			},
+			}
 			(Self::Void, Self::Optional(_)) => {
-				// Void (not specifying a type) is a subtype of Optional<T> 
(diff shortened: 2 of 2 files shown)


eladb:
Makes sense!

libs/wingc/src/type_check.rs
```diff
@@ -757,6 +757,11 @@ impl Subtype for Type {
 				// Nil is a subtype of Optional<T> for any T
 				true
 			}
+			(Self::Void, Self::Optional(_)) => {
```
eladb, line 760: Can you please add a test?

autow0rk: Added a test, let me know if it needs to be changed or in a different file.

examples/tests/valid/optionals.test.w
```diff
@@ -204,3 +204,13 @@ if let s1 = str1 {
 } elif let s2 = str2 {
   assert(true);
 }
+
+let functionReturningNothing = (): void? => {
```
eladb, line 208: i thought the whole idea is that `void` is implicitly an optional, no?

```suggestion
let functionReturningNothing = (): void => {
```

eladb (changes requested):

examples/tests/valid/optionals.test.w
autow0rk: My idea was, if you want to check that returning ```void``` (doing ```return;```) in a function is ok in a function that expects an optional value (```void?```, ```num?```, etc.), then this test would verify that. 

Technically ```void?``` and ```void``` lead to the same case in ```is_subtype_of()``` being triggered, but I thought this might be easier to read and make the idea more explicit. Let me know if this should be changed though and I'll update the PR.

eladb: I see, so let's change this to `num?` or `str?` instead.

eladb (approved):

eladb (commented):
LGTM. @Chriscbr can you take a look?

Chriscbr:
@eladb please see the comment I added to the original issue:

https://github.com/winglang/wing/issues/4972#issuecomment-1854687189

eladb:
@Chriscbr I actually tend to agree that `void` can be considered a subtype of any `T?`. It's pragmatic and useful and I don't think I see any potential harm, no?

autow0rk:
Closing because discussion in the original issue led to another solution

PR #5203, opened 2023-12-13
chore: story 16 - wing for typescript

eladb:
User story for the initial version of Wing for TypeScript.

### DIFF
--- a/docs/contributing/03-stories/story-16.md
+++ b/docs/contributing/03-stories/story-16.md
@@ -1,6 +1,6 @@
 # User Story 16 - Wing for TypeScript
 
-> **Status**: Last Call
+> **Status**: Approved
 
 [Anders](https://en.wikipedia.org/wiki/Anders_Hejlsberg) loves TypeScript, but he also loves the amazing experience Wing offers for building cloud applications.
 
--- a/docs/contributing/03-stories/story-16.md
+++ b/docs/contributing/03-stories/story-16.md
@@ -97,10 +97,10 @@ pub class Bucket {
   pub addFile(key: str, path: str, encoding?: str): void { return this.inner.addFile(key, path, encoding); }
   pub addObject(key: str, body: str): void               { return this.inner.addObject(key, body); }
 
-  pub onCreate(fn: IBucketEventHandler, opts?: BucketOnCreateOptions): void { return this.onCreate(fn, opts); }
-  pub onDelete(fn: IBucketEventHandler, opts?: BucketOnDeleteOptions): void { return this.onDelete(fn, opts); }
-  pub onUpdate(fn: IBucketEventHandler, opts?: BucketOnUpdateOptions): void { return this.onUpdate(fn, opts); }
-  pub onEvent(fn: IBucketEventHandler, opts?: BucketOnEventOptions): void   { return this.onEvent(fn, opts); }
+  pub onCreate(fn: IBucketEventHandler, opts?: BucketOnCreateOptions): void { return this.inner.onCreate(fn, opts); }
+  pub onDelete(fn: IBucketEventHandler, opts?: BucketOnDeleteOptions): void { return this.inner.onDelete(fn, opts); }
+  pub onUpdate(fn: IBucketEventHandler, opts?: BucketOnUpdateOptions): void { return this.inner.onUpdate(fn, opts); }
+  pub onEvent(fn: IBucketEventHandler, opts?: BucketOnEventOptions): void   { return this.inner.onEvent(fn, opts); }
 
   pub inflight copy(srcKey: str, dstKey: str): void               { return this.inner.copy(srcKey, dstKey); }
   pub inflight delete(key: str, opts?: BucketDeleteOptions): void { return this.inner.delete(key, opts); }
--- a/docs/contributing/03-stories/story-16.md
+++ b/docs/contributing/03-stories/story-16.md
@@ -2,24 +2,34 @@
 
 > **Status**: Draft
 
-Anders loves TypeScript, but he also loves the amazing experience Wing offers for building cloud applications.
+[Anders](https://en.wikipedia.org/wiki/Anders_Hejlsberg) loves TypeScript, but he also loves the amazing experience Wing offers for building cloud applications.
 
-It's Anders' lucky day because Wing Cloud had just announced that very soon TypeScript developers are able to start leveraging Wing libraries from the comfort of
+It's Anders' lucky day because Wing Cloud had just announced that very soon TypeScript developers will be able to leverage any Wing library from the comforts of
 their favorite programming language.
 
-He got an invite for a pre-release. The invite explains that initially, there is only support for `cloud.Bucket` and `cloud.Function` and for consuming [winglibs](https://github.com/winglang/winglibs).
+Anders just got an invite for a pre-release of Wing for TypeScript (code name "wing4ts"). The invite explains that in this pre-release, there is only support for `cloud.Bucket` and `cloud.Function` from the Wing Cloud Library and for consuming [winglibs](https://github.com/winglang/winglibs).
 
-The example looks like this:
-
-First, install some libraries:
+First, install the Wing CLI:
 
 ```sh
-$ npm i winglang wingsdk @winglibs/cloud @winglibs/checks
+npm i winglang
+```
+
+Now, install the Wing for TypeScript library (temporary name):
+
+```sh
+npm i wing4ts
+```
+
+Now, install some winglibs:
+
+```sh
+$ npm i @winglibs/cloud @winglibs/checks
 ```
 
 ```ts
 // main.ts
-import { inflight, App } from "wingsdk";
+import { inflight, App } from "wing4ts";
 import * as cloud from "@winglibs/cloud";
 import { Check } from "@winglibs/checks";
 
@@ -96,5 +106,8 @@ pub class Bucket {
   pub inflight delete(key: str, opts?: BucketDeleteOptions): void { return this.inner.delete(key, opts); }
 
   // ... you got the point
+
+  pub grantPut(target: std.InflightHost): void { /* ??? */ }
+  // ...
 }
 ```
--- a/docs/contributing/03-stories/story-16.md
+++ b/docs/contributing/03-stories/story-16.md
@@ -65,3 +65,36 @@ $ wing compile -t tf-aws main.ts
 ```
 
 And enjoy life!
+
+## Notes
+
+The `@winglibs/cloud` library is a Wing library that looks like this:
+
+```js
+bring cloud;
+
+pub struct BucketProps {
+  public: bool?;
+}
+
+pub class Bucket {
+  inner: cloud.Bucket;
+
+  new(props: BucketProps?) {
+    this.inner = new cloud.Bucket(props);
+  }
+
+  pub addFile(key: str, path: str, encoding?: str): void { return this.inner.addFile(key, path, encoding); }
+  pub addObject(key: str, body: str): void               { return this.inner.addObject(key, body); }
+
+  pub onCreate(fn: IBucketEventHandler, opts?: BucketOnCreateOptions): void { return this.onCreate(fn, opts); }
+  pub onDelete(fn: IBucketEventHandler, opts?: BucketOnDeleteOptions): void { return this.onDelete(fn, opts); }
+  pub onUpdate(fn: IBucketEventHandler, opts?: BucketOnUpdateOptions): void { return this.onUpdate(fn, opts); }
+  pub onEvent(fn: IBucketEventHandler, opts?: BucketOnEventOptions): void   { return this.onEvent(fn, opts); }
+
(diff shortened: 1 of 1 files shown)


docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,67 @@
...
+First, install some libraries:
+
+```sh
+$ npm i winglang wingsdk @winglibs/cloud @winglibs/checks
+```
+
+```ts
+// main.ts
+import { inflight, App } from "wingsdk";
+import * as cloud from "@winglibs/cloud";
```
staycoolcall911, line 23: What does the @winglibs/cloud library look like?

eladb: It is a simple shim that exports a Bucket and Function classes with an API that mimics the Wing Cloud Library and proxies everything to the "real" resources.

This is the proposed mechanism to initially expose Wing Cloud Library resources to TypeScript users.

Eventually we will rewrite the entire SDK in winglang and then this additional layer won't be needed.

For the very first release, we can just expose these two types as an exercise. Then we can easily add the rest of the cloud library.

eladb: @MarkMcCulloh @staycoolcall911 added a section with some example.

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,100 @@
...
+
+First, install some libraries:
+
+```sh
+$ npm i winglang wingsdk @winglibs/cloud @winglibs/checks
+```
+
+```ts
+// main.ts
+import { inflight, App } from "wingsdk";
```
MarkMcCulloh, line 22: Is this intentionally not `@winglang/sdk`?

```suggestion
import { inflight, App } from "@winglang/sdk";
```

I can see the benefit of a new package specifically for TS, but not sure it's an initial requirement.

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+And enjoy life!
+
+## Notes
+
+The `@winglibs/cloud` library is a Wing library that looks like this:
+
+```js
+bring cloud;
```
MarkMcCulloh, line 74: I know what you mean here, although technically `bring cloud` also means "bring cloud from wingsdk".

I guess the compiler should read `cloud` here as always referring to the internal one first, so I assume this won't be a problem

docs/contributing/03-stories/story-16.md
MarkMcCulloh, line 22: Assuming this is the same wingsdk that's used elsewhere, which version of the SDK takes priority if the node_modules version and the CLI version conflict?

This problem technically already exists but will be more prevalent when users are commonly installing the SDK themselves. I believe currently the CLI version takes precedence, but I'm guess that would have to change. WDYT?

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,100 @@
...
+const check = new Check(app, "Check", inflight({ message, fn, bucket, data }, async (ctx, event) => {
+  await fn.invoke();
+  const actual = await bucket.get("hello.txt");
+
+  console.log(`actual = ${actual}, expected = ${data}`);
+
+  if (actual !== ctx.data) {
+    throw new Error("check failed");
+  }
+}));
```
MarkMcCulloh, line 47: This is definitely going to be an easy mistake to make (mostly the `data` one, which wouldn't have a type error)

```suggestion
const check = new Check(app, "Check", inflight({ fn, bucket, data }, async (ctx, event) => {
  await ctx.fn.invoke();
  const actual = await ctx.bucket.get("hello.txt");

  console.log(`actual = ${actual}, expected = ${ctx.data}`);

  if (actual !== ctx.data) {
    throw new Error("check failed");
  }
}));
```

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,100 @@
...
+const app = new App();
+
+const data = "world";
+const bucket = new cloud.Bucket(app, "Bucket");
+
+const fn = new cloud.Function(app, "Function", inflight({ bucket, data }, async (ctx, event) => {
+  await ctx.bucket.put("hello.txt", ctx.data);
+});
+
+bucket.grantPut(fn);
```
MarkMcCulloh, line 36: implementation note: This seems fine with me and the idea of automatically codegenerating these based on available inflight methods makes sense

MarkMcCulloh (approved):

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,100 @@
+# User Story 16 - Wing for TypeScript
+
+> **Status**: Draft
+
+Anders loves TypeScript, but he also loves the amazing experience Wing offers for building cloud applications.
+
+It's Andres' lucky day because Wing Cloud had just announced that very soon TypeScript developers are able to start leveraging Wing libraries from the comfort of
```
Chriscbr, line 7: His name seems to be shape shifting haha

docs/contributing/03-stories/story-16.md
Chriscbr: The mistakes could be easy to make, though it's not too unusual for some of these to exist in frameworks. It reminds me a little bit of how in React you're not allowed to conditionally call hooks.

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,100 @@
...
+fn.grantInvoke(check);
+bucket.grantGet(check);
+
+app.synth();
+```
+
+Now, you can open the Wing Console:
+
+```sh
+$ wing run main.ts
```
Chriscbr, line 58: I almost forgot we had a `run` command, I've been using `it` for so long... 😆

docs/contributing/03-stories/story-16.md
eladb, line 7:
```suggestion
It's Anders' lucky day because Wing Cloud had just announced that very soon TypeScript developers are able to start leveraging Wing libraries from the comfort of
```

docs/contributing/03-stories/story-16.md
eladb: Yes, it is intentional - this is *not* the internal `@winglang/sdk` library which we have today. It's a new library designed to support the TypeScript experience.

I am proposing to separate these layers so that we don't couple our TypeScript experience with the JSII library we use as the underlying implementation for winglang's standard library.

They are not the same thing in my mind, and I hope that at some point we will completely deprecate `@winglang/sdk` and rewrite it winglang.

eladb: As mentioned above, this is *not* `@winglang/sdk`, it's a new library designed for TypeScript users.

I think your concern still holds though but as you observed, we already have this issue, so I think we should discuss this outside the scope of the TypeScript experience.

docs/contributing/03-stories/story-16.md
eladb: yes that will be awesome.

I would implement these manually for now in `@winglibs/cloud` but I think we will need to expose some APIs from the standard library to be able to explicitly add bindings/permissions/lifting operations.

@Chriscbr see above.

docs/contributing/03-stories/story-16.md
eladb: Good point. The intention here *is* to refer to the standard library (internal) module and not to the winglib.

But as you say, since in winglang the internal modules take precedence, then in TypeScript this kind of makes sense.

I am wondering if eventually this is where we will put the code of the SDK (once we start writing it in Wing).

docs/contributing/03-stories/story-16.md
eladb: Fixed.

docs/contributing/03-stories/story-16.md
eladb: added some pseudo code in the winglib code.

docs/contributing/03-stories/story-16.md
eladb: Changed the name to a temporary name `wing4ts` so that it's clearer that this is not the same as `@winglang/sdk`.

MarkMcCulloh: > I am proposing to separate these layers so that we don't couple our TypeScript experience with the JSII library we use as the underlying implementation for winglang's standard library.

Being in the same library doesn't necessarily couple them, although I would argue `App` maybe should be coupled since it will already do the exact same thing we already have to deal with in wing compiled output.
From a maintenance perspective, it's very useful if the codepath for the compiled ts and the jsified wing are similar. **As long as it doesn't harm or dictate the wing experience**, of course.

`inflight` is special, sure, but an entire library just for that seems a bit overkill. What if we just have a specific barrel import until the day comes where `@wingland/sdk` is deleted (which I'm assuming will take a while)

```
import { inflight, App } from "@winglang/sdk/typescript"; // or node or js or whatever
```

Where App can just be a re-export of the existing App.

---

In the end it's not a big deal, I just figured a new package can wait, but if you feel strongly about it being separate then I'm cool with that.

eladb: Awesome. I think this separation will give us more flexibility to evolve the TypeScript experience independently of how the winglang standard library is implemented. I prefer TypeScript users not to be exposed to all the horrors within the internal implementation, and only have a nice and clean API to work with that's optimized for them.

docs/contributing/03-stories/story-16.md
MarkMcCulloh: I'll use this thread to talk about a circular dependency that I've been mentioning.

Currently, The implementations of `std.Test` use `cloud.Function`. `std.Test` is built into wing and is effectively a language feature, so it makes sense that it would have to stay in an internal library. If `@winglang/sdk` no longer had `cloud.Function`, this internal library would depend on `@winglibs/cloud` instead. This means the compiler itself would not be able to compile `@winglibs/cloud` (or at best it could not use `test "" {}`)

If `@winglibs/cloud` is a simple proxy to `@winglang/sdk` then this isn't an issue, but it's a hurdle to think about later.

eladb: > If @winglibs/cloud is a simple proxy to @winglang/sdk then this isn't an issue, but it's a hurdle to think about later.

Cool :-)

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,113 @@
...
+    this.inner = new cloud.Bucket(props);
+  }
+
+  pub addFile(key: str, path: str, encoding?: str): void { return this.inner.addFile(key, path, encoding); }
+  pub addObject(key: str, body: str): void               { return this.inner.addObject(key, body); }
+
+  pub onCreate(fn: IBucketEventHandler, opts?: BucketOnCreateOptions): void { return this.onCreate(fn, opts); }
+  pub onDelete(fn: IBucketEventHandler, opts?: BucketOnDeleteOptions): void { return this.onDelete(fn, opts); }
+  pub onUpdate(fn: IBucketEventHandler, opts?: BucketOnUpdateOptions): void { return this.onUpdate(fn, opts); }
+  pub onEvent(fn: IBucketEventHandler, opts?: BucketOnEventOptions): void   { return this.onEvent(fn, opts); }
```
skyrpex, line 103:
```suggestion
  pub onCreate(fn: IBucketEventHandler, opts?: BucketOnCreateOptions): void { return this.inner.onCreate(fn, opts); }
  pub onDelete(fn: IBucketEventHandler, opts?: BucketOnDeleteOptions): void { return this.inner.onDelete(fn, opts); }
  pub onUpdate(fn: IBucketEventHandler, opts?: BucketOnUpdateOptions): void { return this.inner.onUpdate(fn, opts); }
  pub onEvent(fn: IBucketEventHandler, opts?: BucketOnEventOptions): void   { return this.inner.onEvent(fn, opts); }
```

docs/contributing/03-stories/story-16.md
```diff
@@ -0,0 +1,113 @@
...
+  pub onDelete(fn: IBucketEventHandler, opts?: BucketOnDeleteOptions): void { return this.inner.onDelete(fn, opts); }
+  pub onUpdate(fn: IBucketEventHandler, opts?: BucketOnUpdateOptions): void { return this.inner.onUpdate(fn, opts); }
+  pub onEvent(fn: IBucketEventHandler, opts?: BucketOnEventOptions): void   { return this.inner.onEvent(fn, opts); }
+
+  pub inflight copy(srcKey: str, dstKey: str): void               { return this.inner.copy(srcKey, dstKey); }
+  pub inflight delete(key: str, opts?: BucketDeleteOptions): void { return this.inner.delete(key, opts); }
+
+  // ... you got the point
+
+  pub grantPut(target: std.InflightHost): void { /* ??? */ }
```
tsuf239, line 110: I wonder if we can somehow reuse the way we grant permissions in wing- keeping those ops in a list and call `_registerOnLift` later, to keep the implementation as close to wing's and avoid duplications of logic pieces that needs to be maintained

eladb: We decided to start manually and iterate

PR #5208, opened 2023-12-13
fix(sdk): valid api url

meirdev:
Hi!

There is a problem with the current api validate path, a path like this: `/:param/test` is considered invalid.

I fixed and simplified the Regex.

### DIFF
--- a/examples/tests/valid/api_valid_path.test.w
+++ b/examples/tests/valid/api_valid_path.test.w
@@ -65,3 +65,4 @@ testValidPath("/test/segment1/:param1/segment2?query1=value1?query2=value2");
 testValidPath("/test/segment1/segment2?query=value1&query2=value2");
 testValidPath("/test/path.withDots");
 testValidPath("/test/path/.withDots/:param/:param-dash/x");
+testValidPath("/:param/test");
--- a/libs/wingsdk/src/cloud/api.ts
+++ b/libs/wingsdk/src/cloud/api.ts
@@ -383,11 +383,7 @@ export class Api extends Resource {
    * @internal
    */
   protected _validatePath(path: string) {
-    if (
-      !/^(\/[a-zA-Z0-9_\-\.]+(\/\:[a-zA-Z0-9_\-]+|\/[a-zA-Z0-9_\-\.]+)*(?:\?[^#]*)?)?$|^(\/\:[a-zA-Z0-9_\-]+)*\/?$/g.test(
-        path
-      )
-    ) {
+    if (!/^((\/:[\w-]+|\/[\w\.-]+)*(\?[^#]*)?)?\/?$/g.test(path)) {
       throw new Error(
         `Invalid path ${path}. Url parts can only contain alpha-numeric chars, "-", "_" and ".". Params can only contain alpha-numeric chars and "_".`
       );


ekeren:
Thanks @meirdev!!! 

@tsuf239 should we also have an sdk_test here?

ekeren:
@meirdev , following your article https://meirdev.wordpress.com/2023/12/12/wing-%d7%a9%d7%a4%d7%aa-%d7%aa%d7%9b%d7%a0%d7%95%d7%aa-%d7%9c%d7%a2%d7%a0%d7%9f/

Do you mind opening an issue about the application that wasn't able to compile?

Chriscbr (approved):

meirdev:
@ekeren

A minimal example:

```
bring cloud;

let api = new cloud.Api();

api.get("/:id", inflight (req) => {
    let id = req.vars.get("id");

    return {
        body: "id: {id}",
        status: 200,
    };
});
```

It compiles, but fails when I try to apply the tf files:

```
Error: creating API Gateway REST API (hjnq3g5xqd) specification: BadRequestException: Errors found during import:
│       Unable to create resource at path '/{proxy+}': A sibling ({id}) of this resource already has a variable path part -- only one is allowed
│ Additionally, these warnings were found:
│       Parse issue: attribute info is missing
│       Required "info" property is missing from the document root.
│ 
│   with aws_api_gateway_rest_api.cloudApi_api_2B334D75,
│   on main.tf.json line 72, in resource.aws_api_gateway_rest_api.cloudApi_api_2B334D75:
│   72:       }
```

Here is the JSON the program generated:

```json
{
    "openapi": "3.0.3",
    "paths": {
        "/{id}": {
            "get": {
                "operationId": "get-:id",
                "responses": {
                    "200": {
                        "description": "200 response",
                        "content": {}
                    }
                },
                "parameters": [
                    {
                        "name": "id",
                        "in": "path",
                        "required": true,
                        "schema": {
                            "type": "string"
                        }
                    }
                ],
                "x-amazon-apigateway-integration": {
                    "uri": "arn:aws:apigateway:${data.aws_region.Region.name}:lambda:path/2015-03-31/functions/${aws_lambda_function.cloudApi_get_id_0_456E8BD7.arn}/invocations",
                    "type": "aws_proxy",
                    "httpMethod": "POST",
                    "responses": {
                        "default": {
                            "statusCode": "200"
                        }
                    },
                    "passthroughBehavior": "when_no_match",
                    "contentHandling": "CONVERT_TO_TEXT"
                }
            }
        },
        "/{proxy+}": {
            "x-amazon-apigateway-any-method": {
                "produces": [
                    "application/json"
                ],
                "x-amazon-apigateway-integration": {
                    "type": "mock",
                    "requestTemplates": {
                        "application/json": "\n {\"statusCode\": 404}\n "
                    },
                    "passthroughBehavior": "never",
                    "responses": {
                        "404": {
                            "statusCode": "404",
                            "responseParameters": {
                                "method.response.header.Content-Type": "'application/json'"
                            },
                            "responseTemplates": {
                                "application/json": "{\"statusCode\": 404, \"message\": \"Error: Resource not found\"}"
                            }
                        },
                        "default": {
                            "statusCode": "404",
                            "responseParameters": {
                                "method.response.header.Content-Type": "'application/json'"
                            },
                            "responseTemplates": {
                                "application/json": "{\"statusCode\": 404, \"message\": \"Error: Resource not found\"}"
                            }
                        }
                    }
                },
                "responses": {
                    "404": {
                        "description": "404 response",
                        "headers": {
                            "Content-Type": {
                                "type": "string"
                            }
                        }
                    }
                }
            }
        }
    }
}
```

The compiler always adds the "/{proxy+}" part and this causes a conflict with the variable that the user chooses.

I think the problem is in this line:

https://github.com/winglang/wing/blob/693ee4b3a995a9c305479b32c10e0a87013ff125/libs/wingsdk/src/target-tf-aws/api.ts#L325

It should be:

```typescript
const defaultResponse = props.cors ? API_CORS_DEFAULT_RESPONSE(props.cors) : {};
```

PR #5226, opened 2023-12-15
feat(compiler): enhance diagnostics for invalid access errors

Chriscbr:
Before:

Screenshot 2023-12-14 at 5 22 59 PM

After:

Screenshot 2023-12-14 at 5 22 18 PM

Misc:
1. Adding this diagnostic uncovered an issue where if a diagnostic annotation references a span from a JSII library (for example, if you try accessing a protected member of a JSII class), then the CLI crashes since the spans don't point to real files on disk. To fix that, I've removed JSII spans from the compiler AST for now.
2. Adding this diagnostic also uncovered an issue where if too many compiler errors are emitted, then they might not all get printed because the CLI calls `process.exit()`, which forces the process to exit before all I/O operations finish (see https://stackoverflow.com/a/37592669/8605016). To fix this, I've changed several of these places in the code to set the process's exit code with `process.exitCode = 1`.
3. The changes in (2) uncovered an issue where if the simulator is unable to spin up a resource like in https://github.com/winglang/wing/blob/main/examples/tests/invalid/unresolved_state.test.w, then the simulator server wouldn't get cleaned up. I fixed this by adding try/catch handling to this area of the simulator code.
4. In the debugging process I added an eslint rule for catching incorrect promise usage to the `wing` package.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/apps/wing/src/commands/test/test.ts
+++ b/apps/wing/src/commands/test/test.ts
@@ -305,7 +305,8 @@ async function testTf(synthDir: string, options: TestOptions): Promise<std.TestR
   const { clean, testFilter, retry, platform = [BuiltinPlatform.SIM] } = options;
 
   try {
-    if (!isTerraformInstalled(synthDir)) {
+    const installed = await isTerraformInstalled(synthDir);
+    if (!installed) {
       throw new Error(
         "Terraform is not installed. Please install Terraform to run tests in the cloud."
       );
--- a/apps/jsii-docgen/src/cli.ts
+++ b/apps/jsii-docgen/src/cli.ts
@@ -58,5 +58,5 @@ export async function main() {
 
 main().catch((e) => {
   console.error(e);
-  process.exit(1);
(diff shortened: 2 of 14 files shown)


apps/wing/src/commands/test/test.ts
```diff
@@ -305,7 +305,8 @@ async function testTf(synthDir: string, options: TestOptions): Promise<std.TestR
   const { clean, testFilter, retry, platform = [BuiltinPlatform.SIM] } = options;
 
   try {
-    if (!isTerraformInstalled(synthDir)) {
+    const installed = await isTerraformInstalled(synthDir);
+    if (!installed) {
```
Chriscbr, line 309: It's cool that the new eslint rule caught this

MarkMcCulloh (approved):
Nice catches

eladb (approved):

ainvoner (approved):

PR #5230, opened 2023-12-16
chore(build): upgrade github actions to artifacts v4

garysassano:
Upgrading GitHub Actions to [Artifacts v4](https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/) for performance improvements.

### DIFF
--- a/.github/workflows/tf-aws-test.yml
+++ b/.github/workflows/tf-aws-test.yml
@@ -120,7 +120,7 @@ jobs:
       ##################### install a local wing version #########################
       - name: Download Dist Artifacts
         if: ${{ env.LOCAL_BUILD ==  'true' }}
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: dist
           path: dist
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,7 +91,7 @@ jobs:
           SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY }}
 
       - name: Upload Artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: dist
           path: dist/*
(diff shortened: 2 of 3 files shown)


tsuf239:
from the [link](https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/):
```
This version of the artifact actions include performance improvements and new features, however there are also key differences from previous versions that may require updates to your workflows.

- Artifacts will be scoped to a job rather than a workflow. This allows the artifact to become immediately available to [download from the API](https://docs.github.com/rest/actions/artifacts#download-an-artifact) after being uploaded.
- Multiple jobs in the same run cannot upload to the same artifact.
- A single job can upload a maximum of 10 artifacts.
- Artifacts v4 is not cross-compatible with previous versions. For example, an artifact uploaded using v3 cannot be used with actions/download-artifact@v4.
```

@MarkMcCulloh , I am worried about the first two points, do you think it will break something?

.github/workflows/tf-aws-test.yml
MarkMcCulloh, line 1: Could you also update action-download-artifact in the mutation workflow to v3?

MarkMcCulloh (approved):
Thanks @garysassano, and good callout @tsuf239. There's only one part I'm concerned about:

> Artifacts will be scoped to a job rather than a workflow. 

This might be a problem, but I'm not sure. We use https://github.com/dawidd6/action-download-artifact for the mutation job. TBH I'd rather give this a shot and monitor the mutation job. If it fails we'll roll this back for now.

staycoolcall911:
@MarkMcCulloh - do you know why mergify failed?

MarkMcCulloh:
The merge queue PR failed here https://github.com/winglang/wing/actions/runs/7253027111/job/19758792949?pr=5248

Normally I would say it's a transient error, but I'm a little concerned that it failed on this exact step that was updated. I'll update this PR and try it again.

PR #5243, opened 2023-12-18
feat(sdk): resource usage report

### LINKED ISSUE #4644: "translate" passing tests to supported resources, methods and props

tsuf239:
## Description
closes #4644 
As a part of the matrix automation process, we need to figure out what are the methods and properties that are used in each test. I came out with the following approach:
- collect the inflight methods from the `_addOnLift`
- collect the preflight methods and properties by adding a proxy to the Resource class
- save the collected methods list to a file in the output dir (currently, it may change in the future)

### Implemented by adding a new platform (@hasanaburayyan FYI)

Please let me know what you think :) 

TODO:
- [x] Add tests
- [ ] ~Change platform documentation (to reflect the change in the newInstance method)~
- [ ] Prefix "internal" prelight properties and methods with a "_"
 
## Checklist

- [ ] Tests added (always)


tsuf239:
Can be tested with: 
`pnpm wing test -t sim -t libs/compatibility-spy/lib examples/tests/sdk_tests/<path to test>.test.w --no-clean`
(after running `pnpm compile` in `libs/compatibility-spy`)

libs/compatibility-spy/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (!type) {
+      return undefined;
+    }
+
+    return new Proxy(new type(scope, id, ...args), {
+      get: (target, prop: string | Symbol) => {
+        if (
+          typeof prop === "string" &&
+          !prop.startsWith("_") &&
```
hasanaburayyan, line 39: nit: Since we do this check in both the get and set, maybe just create 
```typescript
const isPrivateProp = (prop: string | Symbol) => // ...
```

libs/compatibility-spy/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const PARENT_PROPERTIES: Set<string> = new Set([
+  "node",
+  "onLiftMap",
+  ...Object.getOwnPropertyNames(Construct),
+  ...Object.getOwnPropertyNames(Construct.prototype),
+]);
+
+export class Platform implements IPlatform {
+  public readonly target = "compatibility-spy";
```
hasanaburayyan, line 15: Right now we dont have any compatibility checking in the Platform system but this platform would be agnostic so we can just use a "*" (since I plan to have that support in the future)
```suggestion
  public readonly target = "*";
```

libs/compatibility-spy/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    for (const key of this._usageContext.keys()) {
+      context[key] = Array.from(this._usageContext.get(key) ?? []);
+    }
+
+    writeFileSync(
+      resolve(app.outdir, "usage_context.json"),
+      JSON.stringify(context, null, 2)
+    );
+  }
+}
```
hasanaburayyan, line 103: Maybe we can add some tests to ensure we dont end up breaking this with future changes to the Platform system. 

Perhaps using hangar.

libs/compatibility-spy/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@winglang/compatibility-spy",
```
hasanaburayyan, line 2: Will need to add this to the packages for publishing in github workflow

https://github.com/winglang/wing/blob/main/.github/workflows/build.yml#L388

hasanaburayyan: Also if we do any hangar testing with this platform will need to update `tools/hangar/src/package.setup.ts` and `tools/hangar/package.json`

libs/wingsdk/src/core/app.ts
```diff
@@ -283,8 +290,32 @@ export abstract class App extends Construct implements IApp {
...
+
+  /**
+   * Synthesize the root construct if one was given. If this is a test environment, then
+   * we will synthesize one root construct per test. Otherwise, we will synthesize exactly
+   * one root construct.
+   *
+   * @param props The App props
+   * @param testRunner The test runner
+   */
+  protected synthRoots(props: AppProps, testRunner: TestRunner) {
```
hasanaburayyan, line 305: Where is this used?

hasanaburayyan (commented):
@tsuf239 I love that you did this as a Platform!! I left a few comments please let me know your thoughts, I think this is nearly ready for shipping

libs/wingsdk/src/platform/platform.ts
```diff
@@ -15,12 +15,19 @@ export interface IPlatform {
...
    * @param props props to pass to the resource
    */
-  newInstance?(type: string, scope: Construct, id: string, props: any): any;
+  newInstance?(
+    app: Construct,
+    type: string,
+    scope: Construct,
+    id: string,
+    props: any
+  ): any;
```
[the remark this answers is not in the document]
tsuf239: Cool! I just wanted to check if it's ok first before breaking stuff :) will change the title and the documentation

libs/wingsdk/src/core/app.ts
tsuf239: resolved a conflict in the wrong way, thanks!

libs/compatibility-spy/src/index.ts
tsuf239: that's a good idea,
I tried and it's messing with `prop` type (since we're not specifying `typeof prop === "string" &&` on the same line)
image

libs/wingsdk/src/platform/platform.ts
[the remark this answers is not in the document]
hasanaburayyan: Oh I was not thinking about this originally but I think we can just use the `scope` arg to do this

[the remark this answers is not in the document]
hasanaburayyan: Actually maybe Im crazy, let me look at this a bit more 😁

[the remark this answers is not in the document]
tsuf239: you're right! `App.of(scope)`

tsuf239:
Hi @hasanaburayyan :)
I've added both vitests and hangar tests, please let me know if there any more comments

tools/generate-workspace/src/cli.ts
```diff
@@ -1,36 +1,18 @@
...
-  "@winglibs/testfixture",
-  "bump-pack",
-  "construct-library",
-  "generate-workspace",
-  "hangar",
-  "jsii-fixture",
-  "ts-fixture",
-  "ts4w",
-];
+const ignoreList = ["@wingconsole/design-system", "@wingconsole/error-message", "@wingconsole/eslint-plugin", "@wingconsole/tsconfig", "@wingconsole/use-loading", "@wingconsole/use-persistent-state", "@winglang/platform-awscdk", "@winglang/compatibility-spy", "@winglang/tree-sitter-wing", "@winglang/wingc", "@winglang/wingii", "@winglibs/testfixture", "bump-pack", "construct-library", "generate-workspace", "hangar", "jsii-fixture", "ts-fixture", "ts4w"];
```
hasanaburayyan, line 7: Was this collapse intentional?

tools/hangar/__snapshots__/compatibility-spy.ts.snap
```diff
@@ -0,0 +1,72 @@
...
+  ]
+}"
+`;
+
+exports[`dec.test.w 1`] = `
+"{
+  \\"TestRunner\\": [
+    \\"findTests\\"
+  ],
+  \\"Counter\\": [
```
hasanaburayyan, line 19: 🤌

hasanaburayyan (approved):

tools/generate-workspace/src/cli.ts
tsuf239: prettier did it, the file wasn't changed since we added it probably

tsuf239 (commented):
it seems like a lifting refactor PR done by Chris is interfering with this PR:
image
I'll merge it first to avoid more mess, and then I'll open another PR for a fix
(@hasanaburayyan FYI)

PR #5247, opened 2023-12-18
feat: use typescript with wing cli and worker thread for preflight

### LINKED ISSUE #3678: running multiple compilations simultaneously in the same node process
### I tried this:

```
import path from "path";

import * as wing from "@winglang/compiler";

const wingfile1 = path.resolve("file1.w");
const wingfile2 = path.resolve("file2.w");

const compile = async (wingfile: string) => {
  const simfile = await wing.compile(wingfile, {
    target: wing.Target.SIM,
  });
  return simfile;
};

const testCompile = async () => {
  compile(wingfile1);
  compile(wingfile2);
  compile(wingfile1);
  compile(wingfile2);
};

testCompile();
```

file1.w
```
bring cloud;

let bucket = new cloud.Bucket() as "test1_bucket";
let queue = new cloud.Queue() as "test1_queue";
let api = new cloud.Api() as "test1_api";
```

file2.w
```
bring cloud;

let bucket = new cloud.Bucket() as "test2_bucket";
```

### This happened:

the `target/file2.wsim` to contain the `file1` compilations files.
image

[target.zip](https://github.com/winglang/wing/files/12245074/target.zip)

### I expected this:

The compiler to compile the correct file. 

### Is there a workaround?

no

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

MarkMcCulloh:
#### General Vibe
- This is an MVP of getting TS working via the wing cli (`compile`, `run`, and `test`). This is basically only usable for preflight
- I wanted to change as little as possible in the SDK to avoid wasted work once resources are moved to winglibs
- Does not address inflight stuff yet

The entrypoint TS experience is different than the existing issue/story:

```ts
import { main } from "ts4wing";

main((app) => {
  // stuff goes here
})
```

This was due to our usage of a "root" construct for testing isolation. The wing cli needs to have control over the instantiation of resources, so `new App()` is not really feasible because we don't want users to actually attach stuff at the app-level. It may be possible to get it working with `new App()` but the hacks needed didn't seem worth it.
An added benefit of this approach is that the user no longer needs to call `app.synth()`

Anecdotally, I have built an awscdk-based internal framework at a previous job that looked like this for a similar reason (to duplicate stuff across Stages) and it worked pretty well.

#### Indirect Changes
- Made sure everything matches their `@types/node` version
- **preflight execution now happens in a worker thread instead of a VM**
  - VM context did not allow data to be provided past the initially loaded file.
  - In a VM, process.env is useful but pollutes the parent process
  - The overloaded "require" was only useful for the initial file. Now I added a shim that overloads require everywhere to make sure the SDK is loaded properly

Fixes #3678


libs/ts4wing/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * wing(app => {
+ *   new cloud.Bucket(app, "Bucket");
+ * });
+ * ```
+ *
+ * @param fn Define your application using the provided root construct.
+ *           Note that this function may be called multiple times when used with `wing test`.
+ */
+export function wing(fn: (root: Construct) => void, props: WingAppProps = {}) {
```
MarkMcCulloh, line 30: I considered calling this `app`, but I was concerned that people would expect it to be a construct rather than a function. Also I think people may want to call the root construct `app`

examples/tests/sdk_tests/util/env.test.w
```diff
@@ -5,7 +5,7 @@ let NOT_ACTUAL_ENV = "__NOT_ACTUAL_ENV_SHOULD_FAIL__";
 /**
  * env()
  */
-assert(util.env("PATH").length > 0);
+assert(util.env("WING_TARGET") != "");
```
MarkMcCulloh, line 8: Using a worker thread broke this on windows. Not sure why, but the variables below should also come from the CLI process so I'm not too concerned about a huge regression here.

examples/ts-fixture/src/main.ts
MarkMcCulloh, line 1: A 🧵 for the discussion on this function approach

(The name ts4wing is temporary, and eventually cloud will be moved to winglibs)

libs/ts4wing/src/compiler.ts
```diff
@@ -0,0 +1,44 @@
...
+    console.error(diagnostic.messageText);
+  }
+  if (results.emitSkipped) {
+    throw new Error("TS compilation failed");
+  }
+
+  // get the last .js file emitted, this should be the entrypoint
+  // TODO Is this really the best way to do this?
+  const emittedFiles = results.emittedFiles?.filter((f) => f.endsWith(".js"));
+  const emittedFile = emittedFiles?.[emittedFiles.length - 1];
```
MarkMcCulloh, line 37: The emittedFiles prop is empty unless the hidden listEmittedFiles is used, so it's clearly an internal-ish. It's kinda crazy there isn't a better API for this. If someone knows a good way lemme know

Chriscbr:
```js
import { wing } from "ts4wing";

wing((app) => {
  // stuff goes here
})
```

Interesting... the API reminds me a little bit of React's [createRoot](https://react.dev/reference/react-dom/client/createRoot) / root.render().

examples/tests/sdk_tests/util/env.test.w
eladb: Not sure I understand. Wouldn't we expect `PATH` to be defined?

Also/ why worker threads?

examples/ts-fixture/src/main.ts
```diff
@@ -0,0 +1,6 @@
+import { wing } from "ts4wing";
+import { cloud } from "@winglang/sdk";
+
+wing((app) => {
```
eladb, line 4: What's the benefit of using a function here over the more conventional App class?

I am okay to try this but it feels a bit weird and I am not sure exactly what's the motivation of breaking the CPM specifically for the app.

libs/ts4wing/README.md
```diff
@@ -0,0 +1,17 @@
+# ts4wing - Experimental TypeScript experience for Wing
+
+```ts
+// main.ts
+import { wing } from "ts4wing";
```
eladb, line 5: Maybe we can call this something even more cryptic given this is temporary:

```suggestion
import { wing } from "ts4w";
```

libs/ts4wing/src/compiler.ts
```diff
@@ -0,0 +1,44 @@
+import { join } from "path";
+
+export interface WingTSOptions {
+  workDir: string;
+  entrypoint: string;
+}
+
+export async function compileTypescriptForWing(options: WingTSOptions) {
```
eladb, line 8:
```suggestion
export async function compile(options: WingTSOptions) {
```

I think there's sufficient ambient context (the package name)

libs/ts4wing/src/compiler.ts
```diff
@@ -0,0 +1,44 @@
...
+  const ts = (await import("typescript")).default;
```
eladb, line 9: Curious what's the benefit of using an async import here

libs/ts4wing/src/compiler.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  for (const diagnostic of results.diagnostics) {
+    console.error(diagnostic.messageText);
+  }
+  if (results.emitSkipped) {
+    throw new Error("TS compilation failed");
+  }
+
+  // get the last .js file emitted, this should be the entrypoint
+  // TODO Is this really the best way to do this?
```
eladb, line 35: Maybe use the name to find it?

libs/ts4wing/src/index.ts
```diff
@@ -0,0 +1,74 @@
+import { std, platform } from "@winglang/sdk";
+import { Construct } from "constructs";
+
+/**
+ * Properties for a Wing app.
+ */
+export interface WingAppProps {
```
eladb, line 7:
```suggestion
export interface AppProps {
```

There's already ambient semantic context, so no need to add "wing" here.

libs/ts4wing/src/index.ts
eladb: How about `main`?

```js
main(app => {

  // ...

});
```

libs/ts4wing/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    constructor(scope: Construct, id: string) {
+      super(scope, id);
+      fn(this);
+    }
+  }
+
+  const platformPaths = ((s) => (!s ? [] : s.split(";")))(
+    process.env.WING_PLATFORMS
+  );
+  const outdir = process.env.WING_SYNTH_DIR;
```
eladb, line 56: Seems like this is a duplicated code from the generated wing code. Maybe we can reuse in the future

libs/wingcompiler/src/compile.ts
```diff
@@ -152,84 +151,120 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
-      (wingc.exports.memory as WebAssembly.Memory).buffer,
-      data_ptr,
-      data_len
-    );
-    const data_str = new TextDecoder().decode(data_buf);
-    errors.push(JSON.parse(data_str));
-  }
+  // Do TS compilation
+  if(entrypoint.endsWith(".ts")) {
+    const ts4wing = await import("ts4wing/dist/compiler.js")
```
eladb, line 189: Very fragile. Why not just export it properly ?

eladb, line 189: For readability I would recommend extracting this big fork into two functions.

libs/wingcompiler/src/compile.ts
```diff
@@ -152,84 +151,120 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
-  }
-  if (!compileSuccess) {
-    // This is a bug in the user's code. Print the compiler diagnostics.
-    throw new CompileError(errors);
+To use Wing with TypeScript files, you must install "ts4wing" as a dependency of your project.
+npm i ts4wing
+`);
+    });
+
+    preflightEntrypoint = await ts4wing.compileTypescriptForWing({
```
eladb, line 199:
```suggestion
    preflightEntrypoint = await ts4wing.compile({
```

libs/wingcompiler/src/compile.ts
```diff
@@ -9,6 +7,7 @@ import { copyDir, normalPath } from "./util";
 import { existsSync } from "fs";
 import { BuiltinPlatform } from "./constants";
 import { CompileError, PreflightError } from "./errors";
+import { Worker } from "worker_threads";
```
eladb, line 10: Can you elaborate why is this needed?

(Every time I touched worker threads in Node I got bit in the ass at some point.). Node is not well designed for multi threading...

eladb (approved):
Ship it!

examples/ts-fixture/src/main.ts
MarkMcCulloh: If you look at an existing preflight.js, you'll see:

```js
// ... stuff

  constructor($scope, $id) {
    super($scope, $id);
    // the actual app goes here, using `this` as a scope
  }
}

const $PlatformManager = new $stdlib.platform.PlatformManager({platformPaths: $platforms});
const $APP = $PlatformManager.createApp({ outdir: $outdir, name: "test", rootConstruct: $Root, isTestEnvironment: $wing_is_test, entrypointDir: process.env['WING_SOURCE_DIR'], rootId: process.env['WING_ROOT_ID'] });
$APP.synth();
```

The important part here is that constructs are by default created in the $Root construct scope rather than the app. This assumption is used elsewhere, like in testing to duplicate the root. I couldn't figure out a way to do this with `new App()`. Even if that `App` is not actually an App but a child of one, we would still need to execute all the code in the entire TS code to duplicate stuff. This felt icky and not worth keeping the constructor-based model for. 

The function is a clear signal that "anything inside here has a lifecycle managed by wing"

libs/ts4w/src/internal.ts
```diff
@@ -0,0 +1,43 @@
+import { join } from "path";
+
+export interface WingTSOptions {
```
eladb, line 3:
```suggestion
export interface CompileOptions {
```

nit

libs/wingcompiler/src/compile.ts
eladb: Thanks for the context. God speed

tools/hangar/__snapshots__/test_corpus/valid/api_cors_default.test.w_compile_tf-aws.md
```diff
@@ -289,126 +289,118 @@ module.exports = function({ $apiDefaultCors_url, $expect_Util, $http_HttpMethod,
...
+        })())
+      `;
+    }
+    _supportedOps() {
+      return [...super._supportedOps(), "handle", "$inflight_init"];
+    }
+  }
+  class $Closure2 extends $stdlib.std.Resource {
+    _hash = require('crypto').createHash('md5').update(this._toInflight()).digest('hex');
+    constructor($scope, $id, ) {
```
eladb, line 327: would have been nice if these snapshots wouldn't have changed. it's a bit hard to tell why they did and it reduces my confidence in this pull request.

eladb (approved):

tools/hangar/__snapshots__/test_corpus/valid/api_cors_default.test.w_compile_tf-aws.md
MarkMcCulloh: I can undo it. I made the change to test reusing the `main` function for winglang codegen as well.

MarkMcCulloh: undid those changes, it wasn't crucial for this PR

libs/ts4w/tsup.config.ts
Chriscbr, line 1: How's this used?

libs/wingcompiler/src/compile.ts
```diff
@@ -136,100 +137,89 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
+  };
+
+  let compilerModuleDir = nearestNodeModules(__dirname);
+  let sdkModuleDir = nearestNodeModules(join(require.resolve("@winglang/sdk"), "..", "..", ".."));
+  let wingNodeModules = nearestNodeModules(wingDir);
+  let workDirNodeModules = nearestNodeModules(workDir);
 
-  tempProcess.env["WING_SYNTH_DIR"] = tmpSynthDir;
-  tempProcess.env["WING_NODE_MODULES"] = wingNodeModules;
+  const nodePaths = [compilerModuleDir, sdkModuleDir, wingNodeModules, workDirNodeModules];
```
Chriscbr, line 174: I'd add a comment explaining why we need to look up in all of these directories / why the default node module resolution isn't sufficient

libs/wingcompiler/src/compile.ts
```diff
@@ -136,100 +137,89 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
-    const data_buf = Buffer.from(
-      (wingc.exports.memory as WebAssembly.Memory).buffer,
-      data_ptr,
-      data_len
-    );
-    const data_str = new TextDecoder().decode(data_buf);
-    errors.push(JSON.parse(data_str));
-  }
+    if (os.platform() === "win32") {
+      // In worker threads on Windows, environment variables on case-sensitive.
```
Chriscbr, line 212: Windows sure is something 🙃

libs/wingcompiler/src/compile.ts
```diff
@@ -136,100 +137,89 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
...
-  }
+      if ("Path" in preflightEnv) {
+        preflightEnv.PATH = preflightEnv.Path;
+        delete preflightEnv.Path;
+      }
+    }
 
-  if (isEntrypointFile(entrypoint)) {
-    await runPreflightCodeInVm(workDir, wingDir, tempProcess, log);
+    await runPreflightCodeInVm(preflightEntrypoint, wingDir, preflightEnv, log);
```
Chriscbr, line 222:
```suggestion
    await runPreflightCodeInWorkerThread(preflightEntrypoint, wingDir, preflightEnv, log);
```

libs/wingcompiler/src/compile.ts
```diff
@@ -259,57 +250,116 @@ function isEntrypointFile(path: string) {
...
+        env,
+      });
+      worker.on("error", reject);
+      worker.on("exit", (code) => {
+        if (code === 0) {
+          resolve(undefined);
+        } else {
+          reject(new Error(`Worker stopped with exit code ${code}`));
+        }
+      });
```
Chriscbr, line 359: Where do worker error logs go? I guess they inherit the stderr of the process it's created in, so if preflight code crashes the stack trace should still get printed to the CLI like we have today?

Chriscbr (approved):

libs/wingcompiler/src/compile.ts
MarkMcCulloh: If there's an uncaught exception, the "error" event gets raised and rejects from this promise.
stderr is inherited, so that works the same as when we used a vm.

The non-zero exit code here would only happen if someone explicitly called `process.exit(non-zero)` or something similar.

libs/ts4w/tsup.config.ts
MarkMcCulloh: the compile script runs this to bundle the library. We use this tool for a few other projects in the repo.

libs/wingcompiler/src/compile.ts
MarkMcCulloh: Figured out that all we need is to make sure the correct wingsdk is loaded, so I simplified things with a hack here (also has a comment) https://github.com/winglang/wing/pull/5247/files#diff-0c9cb50eae5ecfdb23433d8f48fb3d99a717429999343b0574c4766865569464R325

PR #5266, opened 2023-12-19
feat(sdk): add global secondary index to DynamoDB table

meirdev:
The GSI is a very important thing in the design of a data table in DynamoDB. Especially if you work with the [single table](https://www.alexdebrie.com/posts/dynamodb-single-table/) method


libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -312,6 +351,10 @@ export interface DynamodbTableSchema extends BaseResourceSchema {
      * Range key for this table.
      */
     readonly rangeKey?: string;
+    /**
+     * The GSI for the table.
+     */
+    readonly globalSecondaryIndex?: DynamodbTableGlobalSecondaryIndex[];
```
tsuf239, line 357: can `DynamodbTableGlobalSecondaryIndex` be imported from `libs/wingsdk/src/ex/dynamodb-table.ts`?

tsuf239 (commented):
Can you please add tests to the `libs/wingsdk/test/target-sim/dynamodb-table.test.ts`, and `libs/wingsdk/test/target-tf-aws/dynamodb-table.test.ts`  and/or `libs/wingsdk/test/shared-aws/dynamodb-table.inflight.test.ts`?

libs/wingsdk/src/target-sim/schema-resources.ts
meirdev: I added to `libs/wingsdk/test/target-tf-aws/dynamodb-table.test.ts`.

It looks like the tests in `libs/wingsdk/test/target-sim/dynamodb-table.test.ts` will have to test exactly what I've already tested in `examples/tests/sdk_tests/dynamodb-table/global-secondary-index.test.w`, is it still relevant to add there too?

tsuf239: yes, the `.w` test files run on the cloud before publishing a version, and the `.ts` tests run on each PR, keeping us covered and preventing regression issues 💪

examples/tests/sdk_tests/dynamodb-table/global-secondary-index.test.w
```diff
@@ -0,0 +1,93 @@
...
+    globalSecondaryIndex: [{
+        name: "TypeIndex",
+        hashKey: "type",
+        projectionType: "ALL"
+    }],
+) as "blog2";
+
+test "Global secondary index" {
+    for i in 1..11 {
+        let var id = util.nanoid();
```
tsuf239, line 42: I'm not sure how stable is that, imagine a random id starts with "zz" while the last id starts with "az" (or "00" and "0z" for the first one) 
is there a way to exclude the chars "0" and "z" from the nano id? another option is to create an array of constants (will ensure that the test will run the same  way each time)

examples/tests/sdk_tests/dynamodb-table/global-secondary-index.test.w
```diff
@@ -0,0 +1,93 @@
...
+
+        // Make the first item entered last, and the last item first
+        if i == 1 {
+            id = "0" + id;
+        } elif i == 10 {
+            id = "z" + id;
+        }
```
tsuf239, line 49: Doesn't 0 appear before z? The first one will be entered first, and the last one last

libs/wingsdk/src/target-sim/schema-resources.ts
meirdev: Thank you for the clarification 👍, I added that one too.

examples/tests/sdk_tests/dynamodb-table/global-secondary-index.test.w
meirdev: I changed the test, I entered the records manually to avoid any unexpected situation.

examples/tests/sdk_tests/dynamodb-table/global-secondary-index.test.w
```diff
@@ -0,0 +1,100 @@
...
+        projectionType: "ALL"
+    }],
+) as "blog2";
+
+test "Global secondary index" {
+    let idCreated = [
+        {id: "zuegksw", createdAt: 1703412771},
+        {id: "dirnfhw", createdAt: 1703413971},
+        {id: "pdkeruf", createdAt: 1703422171},
+        {id: "azjekfw", createdAt: 1703432071},
```
tsuf239, line 45: Just a suggestion, if you can make those values easier to estimate by the human eye (shorter for example) it would be great for example {id: "a", createdAt: 10}

tsuf239 (commented):
Nice work! thanks a lot for the contribution 💪 
I started an automated test process on the cloud- https://github.com/winglang/wing/actions/runs/7378073397
I can approve this PR as soon as this passes 🚀

tsuf239:
Can you please fix?
image

meirdev:
> Can you please fix? 

I think it happened because of the [fix](https://github.com/winglang/wing/pull/5371) I made (which was reverted once again), because I removed the error page so the API is completely empty.

tsuf239:
ok, running the test again :)
https://github.com/winglang/wing/actions/runs/7386718249

tsuf239:
image

meirdev:
done. (and finally I learned that I can use `wing test -t tf-aws` to run the test in the cloud).

tsuf239 (approved):
Amazing! 
(the relevant tests passed on the workflow toohttps://github.com/winglang/wing/actions/runs/7407446079)

tsuf239:
@skyrpex it seems like the console playwright test failed, do you maybe know why?

tsuf239 (approved):

PR #5267, opened 2023-12-19
docs(examples): add important examples to for-loop

meirdev:
Hi!

I encountered two examples for which I could not find documentation on the site (even [here](https://www.winglang.io/docs/language-reference#26-for)), I think it is important to mention them at least in the examples

### DIFF
--- a/docs/docs/07-examples/04-flow-controls.md
+++ b/docs/docs/07-examples/04-flow-controls.md
@@ -36,6 +36,16 @@ for value in iterable {
 for value in 0..10 {
   log("{value}");
 }
+
+// prints numbers in reverse order from 10 to 0
+for value in 10..-1 {
+    log("{value}");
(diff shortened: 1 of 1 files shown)


Chriscbr (approved):
Good idea 👍  thanks @meirdev!

NathanTarbert:
Nice addition @meirdev 👍 Thank you!

docs/docs/07-examples/04-flow-controls.md
```diff
@@ -36,6 +36,16 @@ for value in iterable {
 for value in 0..10 {
   log("{value}");
 }
+
+// prints numbers in reverse order from 10 to 0
+for value in 10..-1 {
```
eladb, line 41:
```suggestion
for value in 10..=0 {
```

PR #5278, opened 2023-12-20
feat: consume wing libraries from javascript/typescript

MarkMcCulloh:
#### Changes
- no need for a dummy "main" in wing package.json anymore
- `wing pack` now includes jsify output (no preflight execution)
  - This is probably the more controversial change. This doesn't change how winglang imports other winglang packages, which are still compiled from source every time. We can probably explore a custom module loader at some point, but this approach is simpler especially when we start generating .d.ts
- jsify output for libraries no longer exports a function, now exports a simple object
  - added getters for lazy evaluation of subdirs. This isn't really a big deal in terms of perf, just figured _why not_

#### Limitations
- No types
  - A dummy .d.ts is added for now
- ~No extern~
  - ~extern paths in js are absolute, so it only works right now if the lib was packed on your system~
  - Fixed in https://github.com/winglang/wing/pull/5361


apps/wing/src/commands/pack.test.ts
```diff
@@ -136,6 +136,45 @@ describe("wing pack", () => {
...
   });
 
+  it("can consume a Wing project from JS", async () => {
+    // GIVEN
+    const outdir = await generateTmpDir();
+    // copy everything to the output directory to sandbox this test
+    await exec(`cp -r ${goodFixtureDir}/* ${outdir}`);
+    process.chdir(outdir);
+
+    // WHEN
```
eladb, line 146: Is it possible to also add an example of a javascript program that uses a wing library?

eladb (commented):
yey!

apps/wing/src/commands/pack.test.ts
MarkMcCulloh: Currently, the compiler explicitly expects a `.ts` entrypoint. It could technically allow `.js` but I'm not sure there's much value adding and testing that yet.

eladb (approved):
Ship it dude!

Issue #5210, opened 2023-12-13
api path: collision with /{proxy+}


meirdev:
### I tried this:

A minimal example:

```
bring cloud;

let api = new cloud.Api();

api.get("/:id", inflight (req) => {
    let id = req.vars.get("id");

    return {
        body: "id: {id}",
        status: 200,
    };
});
```

### This happened:

It compiles, but fails when I try to apply the tf files:

```
Error: creating API Gateway REST API (hjnq3g5xqd) specification: BadRequestException: Errors found during import:
│       Unable to create resource at path '/{proxy+}': A sibling ({id}) of this resource already has a variable path part -- only one is allowed
│ Additionally, these warnings were found:
│       Parse issue: attribute info is missing
│       Required "info" property is missing from the document root.
│ 
│   with aws_api_gateway_rest_api.cloudApi_api_2B334D75,
│   on main.tf.json line 72, in resource.aws_api_gateway_rest_api.cloudApi_api_2B334D75:
│   72:       }
```

### I expected this:

API route in `/:id`.

### Anything else?

Here is the JSON the program generated:

```json
{
    "openapi": "3.0.3",
    "paths": {
        "/{id}": {
            "get": {
                "operationId": "get-:id",
                "responses": {
                    "200": {
                        "description": "200 response",
                        "content": {}
                    }
                },
                "parameters": [
                    {
                        "name": "id",
                        "in": "path",
                        "required": true,
                        "schema": {
                            "type": "string"
                        }
                    }
                ],
                "x-amazon-apigateway-integration": {
                    "uri": "arn:aws:apigateway:${data.aws_region.Region.name}:lambda:path/2015-03-31/functions/${aws_lambda_function.cloudApi_get_id_0_456E8BD7.arn}/invocations",
                    "type": "aws_proxy",
                    "httpMethod": "POST",
                    "responses": {
                        "default": {
                            "statusCode": "200"
                        }
                    },
                    "passthroughBehavior": "when_no_match",
                    "contentHandling": "CONVERT_TO_TEXT"
                }
            }
        },
        "/{proxy+}": {
            "x-amazon-apigateway-any-method": {
                "produces": [
                    "application/json"
                ],
                "x-amazon-apigateway-integration": {
                    "type": "mock",
                    "requestTemplates": {
                        "application/json": "\n {\"statusCode\": 404}\n "
                    },
                    "passthroughBehavior": "never",
                    "responses": {
                        "404": {
                            "statusCode": "404",
[...]

PR #5282, opened 2023-12-21
fix(sdk): remove /{proxy+} from path and add OPTIONS

meirdev:
Fix for: #5210

Add OPTIONS method instead of default handler with `{proxy+}`.

I can't write a _real_ test because I depend on how aws responds to changes, so I tried with small code:

```wing
bring cloud;

let api = new cloud.Api(cors: true);

api.get("/:id", inflight (req) => {
    let id = req.vars.get("id");

    return {
        body: "get id: {id}",
        status: 200,
    };
});

api.post("/:id", inflight (req) => {
    let id = req.vars.get("id");

    return {
        body: "post id: {id}",
        status: 200,
    };
});
```

```html
<html>
  <head>
    <script>
      fetch(
        "https://amol7a6ij3.execute-api.us-east-1.amazonaws.com/prod/page-id",
        {
          method: "POST",
          headers: {
            "Content-Type": "application/json",
          },
          body: "test",
        }
      ).then((res) => {
        res.text().then((data) => {
          console.log(data);
        });
      });
    </script>
  </head>

  <body>
  </body>
</html>
```

You can see the OPTIONS request that was sent when I entered the page:

Screenshot 2023-12-21 at 9 55 18


eladb (approved):
👏

staycoolcall911:
@meirdev - can you please resolve the conflicts on this PR? It should then merge automatically

yoav-steinberg:
@meirdev I reverted this since this fails on aws:

`apps/wing/bin/wing test -t tf-aws examples/tests/sdk_tests/api/404.test.w`

If you need help recreating it on AWS try to ping me on slack.

staycoolcall911:
@meirdev just to add some more details on why your PR was reverted:
First off, this is completely not your fault as a contributor. It's on us to improve our process.

We have a full test suite under `examples/tests/sdk_tests` which runs wing on the local simulator as well as on the cloud, to verify functional alignment between all platforms supported by wing (e.g. `sim`, `tf-aws`, etc.). 
Currently these tests run post merge to `main` and the one indicated by Yoav failed.
Please let us know if you have any questions on how to fix this test. Thank you!

meirdev:
It's OK,

The problem is in the difference between the AWS and the simulator, in AWS they return a 403 by default and the simulator returns a 404.

My fix allows the user to set what they choose (the default with the `{proxy+}` caused the problem I mentioned in the issue).

Should I change the simulator that will return 403 or delete the test, (or my fix)?

tsuf239:
@meirdev, thanks for your contribution! 
we should always aim to match the behaviour between the sim and the cloud platforms :) 
so, 
- we should keep the test!
- if it makes sense to return 403 ("access to the requested resource is forbidden") we should change the simulator too- in this case I suspect it happens because of missing permissions (but I admit I haven't checked closely... ), so I guess checking that would be the first step :)

meirdev:
@tsuf239 API Gateway REST API returns "Missing Authentication Token" (403) in our case because "The API request is made to an operation or resource that doesn't exist" ([source](https://repost.aws/knowledge-center/api-gateway-authentication-token-errors)).

I can change the test from 404 to 403, and change `libs/wingsdk/src/target-sim/api.inflight.ts`:

```typescript
  public async addEventSubscription(
    subscriber: string,
    subscriptionProps: EventSubscription
  ): Promise<void> {
    const routes = (subscriptionProps as any).routes as ApiRoute[];
    routes.forEach((r) => {
      const s = {
        functionHandle: subscriber,
        method: r.method,
        path: r.path,
      };
      this.routes.push(s);
      this.populateRoute(s, subscriber);
    });

    // return 403 by default (as AWS does)
    this.app.use(async (
      req: express.Request,
      res: express.Response,
    ) => {
      req;
      res.sendStatus(403);
    });
  }
```

tsuf239:
@meirdev I see. 
It looks like it happens because of not using `API_CORS_DEFAULT_RESPONSE` anymore, is it possible to change it back? (in a way it'll include your change while preserving the 404 response?)
if needed, you can maybe ask @skorfmann for help :)

meirdev:
I found a solution, it is possible to change the responses of the error pages:

```typescript
/**
 * Custom error when the resource doesn't exist
 */
const GATEWAY_404_RESPONSE = {
  "x-amazon-apigateway-gateway-responses": {
    "MISSING_AUTHENTICATION_TOKEN": {
      "statusCode": "404",
      "responseTemplates": {
        "application/json": "{\"message\": $context.error.messageString }",
      },
    },
  }
}

class WingRestApi extends Construct {
...
  
    constructor(
...
    ) {
...
              return {
                ...openApiSpec,
                ...GATEWAY_404_RESPONSE,
              }
...
      });
```

tsuf239:
@meirdev looks good :) 
Can you please open a new PR with the fix?

PR #5290, opened 2023-12-21
fix(sdk): sometimes aws function cannot be added to vpc

eladb:
If a function didn't have policy statements before, then the IAM statements required so that it can be added to the VPC are not added because `policyStatements` might be undefined.

The fix is to make sure `policyStatements` is either defined or an empty array.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-tf-aws/function.ts
+++ b/libs/wingsdk/src/target-tf-aws/function.ts
@@ -127,6 +127,8 @@ export class Function extends cloud.Function implements IAwsFunction {
       role: this.role.name,
       policy: Lazy.stringValue({
         produce: () => {
+          this.policyStatements = this.policyStatements ?? [];
+          
           // If there are subnets to attach then the role needs to be able to
           // create network interfaces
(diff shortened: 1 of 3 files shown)


libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -127,6 +127,8 @@ export class Function extends cloud.Function implements IAwsFunction {
       role: this.role.name,
       policy: Lazy.stringValue({
         produce: () => {
+          this.policyStatements = this.policyStatements ?? [];
```
MarkMcCulloh, line 130: The rest of this function no longer needs `this.policyStatements?.` or `this.policyStatements ?? []`

MarkMcCulloh (approved):

libs/wingsdk/src/target-tf-aws/function.ts
eladb: yes, you are right.

PR #5306, opened 2023-12-22
chore(compiler)!: clean up jsification

### LINKED ISSUE #4364: Error: `Unknown symbol "init"` when creating an object of type `std.Range`
### I tried this:

main.w
```
bring cloud;

let r = new std.Range(1, 5);
```
shell
```sh
wing compile main.w
```

### This happened:

```sh
error: Unknown symbol "init"
```

### I expected this:

I expected the code to compile normally.

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

Chriscbr:
Improve the quality of our JavaScript code generation so that when you write code like this in Winglang:

```js
assert(x.length == 3);
```

Instead of producing hairballs like this:

```js
{((cond) => {if (!cond) throw new Error("assertion failed: x.length == 3")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })($x_length,3)))};
```

The compiler instead produces something more readable:

```js
$helpers.assert($helpers.eq($x_length,3), "x.length == 3");
```

Closes #4364

BREAKING CHANGE: `std.Range` has been removed. Use `for x in from..to {}` and `for x in from..=to {}` instead.

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify/snapshots/base_class_with_fields_preflight.snap
```diff
@@ -64,6 +66,7 @@ const $platforms = ((s) => !s ? [] : s.split(';'))(process.env.WING_PLATFORMS);
 const $outdir = process.env.WING_SYNTH_DIR ?? ".";
 const $wing_is_test = process.env.WING_IS_TEST === "true";
 const std = $stdlib.std;
+const $helpers = $stdlib.helpers;
```
eladb, line 69: I am wondering if perhaps it would make more sense to put the helpers in `std`

eladb (approved):

libs/wingc/src/jsify/snapshots/base_class_with_fields_preflight.snap
Chriscbr: Totally, yeah I think that's the next step. Eventually we can have a place named std which just has shared utilities and types (like Datetime) that should be included in every inflight code bundle. Anything that doesn't need to be bundled like std.Test / TestRunner can be moved elsewhere perhaps

MarkMcCulloh (commented):
I forgot to post this review a few days ago, figured I'd still add it

PR #5328, opened 2023-12-25
fix(cli): `wing pack` does not respect `-o` for custom output filename

### LINKED ISSUE #4796: `wing pack` doesn't respect `-o`
### I tried this:

```sh
wing pack -o foo.tgz
```

### This happened:

Output was still the original name

### I expected this:

That the output will be `foo.tgz`

### Is there a workaround?

_No response_

### Component

CLI

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

garysassano:
Closes #4796

### DIFF
--- a/apps/wing/src/cli.ts
+++ b/apps/wing/src/cli.ts
@@ -207,7 +207,7 @@ async function main() {
   program
     .command("pack")
     .description("Package the current directory into an npm library (gzipped tarball).")
-    .addOption(new Option("-o --out-file <filename>", "Output filename"))
+    .addOption(new Option("-o --outfile <filename>", "Output filename"))
     .hook("preAction", collectAnalyticsHook)
     .action(runSubCommand("pack"));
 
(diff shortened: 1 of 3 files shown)


apps/wing/src/cli.ts
```diff
@@ -207,7 +207,7 @@ async function main() {
   program
     .command("pack")
     .description("Package the current directory into an npm library (gzipped tarball).")
-    .addOption(new Option("-o --out-file <filename>", "Output filename"))
+    .addOption(new Option("-o --outfile <filename>", "Output filename"))
```
MarkMcCulloh, line 210: Could you add a test?

apps/wing/src/cli.ts
```diff
@@ -207,7 +207,7 @@ async function main() {
   program
     .command("pack")
     .description("Package the current directory into an npm library (gzipped tarball).")
-    .addOption(new Option("-o --out-file <filename>", "Output filename"))
```
MarkMcCulloh, line 210: This fixes the issue because the interface name doesn't match, right? Could we change the interface instead of this? IMO I like out-file better, so maybe the JS side needs to be outFile.

MarkMcCulloh (approved):

PR #5340, opened 2023-12-26
chore(docs): fixing broken links in the docs

staycoolcall911:
The Algolia tool notified us of some broken links in our documentation. This PR fixes those links.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/contributing/998-archived/01-compile-targets.md
+++ b/docs/contributing/998-archived/01-compile-targets.md
@@ -33,7 +33,7 @@ These targets contain a combination of provisioning engine and cloud environment
 
 ## Portability
 
-The [Wing Cloud Library](/docs/category/cloud-library) has several classes representing abstracted cloud resources, whose APIs are not specific to a single cloud provider.
+The [Wing Cloud Library](/docs/category/cloud) has several classes representing abstracted cloud resources, whose APIs are not specific to a single cloud provider.
 This allows you to write code that can be deployed to any supported cloud environment or provisioning engine.
 
 Here's an example of a portable code snippet:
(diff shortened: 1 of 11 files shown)


docs/contributing/998-archived/01-compile-targets.md
```diff
@@ -91,7 +91,7 @@ In this example, we want to sleep briefly for the Simulator target and for 30 se
 
 Compiler plugins serve to extend the capabilities of the Wing compiler. They offer customization for the compilation output, such as infrastructure definitions. For instance, a compiler plugin can be employed to enforce a specific tagging convention on all resources within the resulting infrastructure.
```
MarkMcCulloh, line 92:
```suggestion
Platforms extend the capabilities of the Wing compiler. They offer customization for the compilation output, such as infrastructure definitions. For instance, a platform can enforce a specific tagging convention on all resources within the resulting infrastructure.
```

docs/contributing/998-archived/01-compile-targets.md
```diff
@@ -91,7 +91,7 @@ In this example, we want to sleep briefly for the Simulator target and for 30 se
...
 
-Learn more about compiler plugins [here](https://www.winglang.io/docs/tools/compiler-plugins)
+Learn more about compiler plugins [here](https://www.winglang.io/contributing/archived/compiler-plugins)
```
MarkMcCulloh, line 94:
```suggestion
Learn more about platforms [here](https://www.winglang.io/docs/concepts/platforms)
```

MarkMcCulloh (approved):

docs/contributing/998-archived/01-compile-targets.md
staycoolcall911: This is part of an archived doc about compilation targets, so I don't think we should update the text in it. The reason I'm still updating the link is because it is broken and we have an automatic tool that notifies us about broken links, so if I won't fix this link it will add noise.

docs/contributing/998-archived/01-compile-targets.md
staycoolcall911: Same remark as above. This is an archived doc.

ainvoner (approved):

PR #5354, opened 2023-12-28
fix(compiler): enforce super constructor call in derived class constructors

### LINKED ISSUE #4277: Missing error when derived class does not call `super` in initializer
### I tried this:

```js
class Foo {
  new(name: str) {}
}

class Bar extends Foo {
  new() {}
}
```

### This happened:

It compiles

### I expected this:

A type checking error, saying that `Bar`'s initializer needs a `super` call since it is a derived class

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

0.33.3

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

fynnfluegge:
Closes #4277 

This PR adds an error to compilation if  `super()` was not called in a derived class constructor.
One last point is questionable. If you take a look at `__snapshots__/invalid.ts.snap` the error is also thrown if `super()` is not the first statement in a constructor. This is tricky to solve.
Screenshot 2023-12-28 at 01 10 12

Further I got a weird change in `test_corpus/valid/optionals.test.w_compile_tf-aws.md` after I added the now required `super()` calls to `optionals.test.w`. This is not related to my changes in `parser.rs` but I am wondering though.

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify/snapshots/base_class_with_fields_preflight.snap
```diff
@@ -98,7 +99,7 @@ class $Root extends $stdlib.std.Resource {
     }
     class Derived extends Base {
       constructor($scope, $id, ) {
-        super($scope, $id);
+        super($scope, $id, );
```
eladb, line 102: Can this be avoided? Not critical if it's a hassle

eladb (commented):
Nice job. I'll let someone from the compiler team to finalize the review.

libs/wingc/src/jsify/snapshots/base_class_with_fields_preflight.snap
fynnfluegge: I was wondering about his change, it was generated after I added `super()` to the inflight constructor and I ran e2e tests. I don't know if this is critical, but seems suspicious to me, so I pushed it 🙂
Screenshot 2023-12-28 at 11 16 45

PR #5375, opened 2023-12-31
fix(sdk): `util.shell` on windows displays the console window

garysassano:
Aligning `util.shell()` to `util.exec()` and removing `shell: false` since that's the default value anyway.

image

### DIFF
--- a/libs/wingsdk/src/util/util.ts
+++ b/libs/wingsdk/src/util/util.ts
@@ -141,6 +141,7 @@ export class Util {
     opts?: ShellOptions
   ): Promise<String> {
     const shellOpts = {
+      windowsHide: true,
       cwd: opts?.cwd,
       env:
         opts?.inheritEnv === true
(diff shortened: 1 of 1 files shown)


eladb (changes requested):
Can you explain why these changes are needed in the PR description and in comments inside the code?

Also- missing test

garysassano:
The PR is in response to #5353 and I explained why the changes in the OP.

The tests already exist, it's just about removing the `shell: false` option which is redundant and aligning the `shell() `method to the `exec(`) method for Windows OS by adding the `windowsHide: true` option there as well.

libs/wingsdk/src/util/util.ts
```diff
@@ -183,9 +184,8 @@ export class Util {
     opts?: ExecOptions
   ): Promise<Output> {
     const execOpts = {
-      cwd: opts?.cwd,
       windowsHide: true,
-      shell: false,
```
eladb, line 188: Enabling shell has security implications, so we should avoid it for `exec()`.

I think `shell` should be `false` for `exec()` and `true` for `shell()`.

garysassano: You cannot set `shell: true` for `shell()` since it's a `string` option.

image

But we can keep the explicit `shell: false` for `exec()` if you prefer despite the fact that it's the default setting.

eladb (approved):

PR #5379, opened 2024-01-01
feat(sdk): add print function to create console output without a trialing newline

mortax:
This PR implements the real solution to the issue raised in discussion https://github.com/winglang/wing/discussions/1772:
* Add `print()` to generate console output without a trailing newline

Note that I have not included `println()` since it would be the same as `log()`.


examples/tests/valid/log.test.w
```diff
@@ -0,0 +1,15 @@
+bring cloud;
+
+
+log("preflight log");
```
eladb, line 4: Can you move this to test under sdk_tests?

eladb (changes requested):
I am not convinced that this use case entitles a whole new built in function. We are trying to minimize those as much as possible.

Maybe something like `fs.stdout.write("text")` work?

examples/tests/valid/log.test.w
mortax: sure. interestingly, this was already in `valid` but you're right, `sdk_tests` makes much more sense.

mortax: > Maybe something like fs.stdout.write("text") work?

yes, that would certainly work (i have been doing the equivalent). thought i completely understand the goal to minimize, i think any language that supports basic output (as Wing does with `log()`) should do the same without a newline (just MHO).

eladb: How about?

```js
log("foo", newline: false);
```

eladb: Just a bit of context: logging in wing doesn't have the same semantic meaning as printing to standard error, standard output in traditional programming, languages. 

There's also a big difference between preflight and inflight.

In preflight logs are basically part of the compilation output while in inflight logs are emitted from the runtime code, executed on the cloud in various computation platforms, such as lambda containers, etc.

mortax: ah, yes. that makes sense. let me try reworking this. i prefer the notion of _adding_ `newline:` rather than `log` vs. `print`, which are unrelated.

mortax: the bad news (for me) is that i'm not 100% sure how to implement an _optional_ `newline` parameter. making the new parameter optional is required to ensure all existing code works as is. because `log` is defined in `lib.rs` (as opposed to TypeScript helpers, e.g. `libs/wingsdk/src/math/math.ts`). it _seems_ like `FunctionParameter` in `wingc/src/ast.rs` would need to be extended to allow a default value (for an optional parameter) -- but i may be out in left field. let me know if there is any prior art i can review.

mortax: i gave it a try (spending most of today messing with it) but didn't have any real success.

eladb: @mortax really appreciate the efforts!

@Chriscbr or @yoav-steinberg will be able to help.

alternatively - if you are comfortable with implementing this as a standard library function in the meantime, so that you are unblocked, then go ahead and we can discuss the built-in support in `log()` separately.

Chriscbr: It looks like you're on the right track @mortax. To get the `newline` parameter working, I think you'll need to add a new struct type. The reason's that in Wing, named parameters are implemented by providing a struct as the last parameter in any function or constructor. So it will be as if we defined the `log` function like

```
let log = (message: str, opts: LogOptions) => { ... }
```

but callers can use it any of these ways:

```
log("foo");
// or
log("foo", newline: false);
// or
log("foo", LogOptions { newline: false });
```

I'd suggest maybe going about this by adding a struct through the Wing SDK, which is implemented as a TypeScript interface. Here's an example of another one:

https://github.com/winglang/wing/blob/2ca491b1965254c1fdea93b7439b52f986b85798/libs/wingsdk/src/std/json.ts#L3-L9

`std` would be a good place for this. Once the interface is added there, you can update the code where builtins are added. (It was in `lib.rs` but it was recently moved to `type_check.rs` - look for the `add_builtins` function).

Let me know that helps or if you run into any issues!

mortax: thanks. i'll try to give this a try tomorrow.

PR #5399, opened 2024-01-01
fix: invalid snapshots when removing multiple abs paths in same line

yoav-steinberg:
The code sanitizer had a faulty regex to detect absolute paths which swallowed multiple abs paths and other double quote strings when they were in the same line.
The result are fixed snapshots.

## Checklist

- [ ] Tests added (always)


staycoolcall911 (approved):
Thanks for the quick fix!

PR #5409, opened 2024-01-02
feat(sdk): add `bucket.rename()` for `tf-gcp` platform

### LINKED ISSUE #4662: Implement `cloud.Bucket` inflight method `rename()` for `tf-gcp` target
### Feature Spec

```js
/**
 * Move an object to a new location in the bucket. If an object already exists
 * at the destination key, it will be overwritten.
 *
 * @throws Will throw if the `src` object doesn't exist.
 */
inflight rename(src: str, dst: str): void;
```

### Use Cases

See above.

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

garysassano:
Closes #4662

## Checklist

- [x] Tests added (always)


garysassano:
image

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
```diff
@@ -54,9 +54,14 @@ export class BucketClient implements IBucketClient {
...
-      `rename is not implemented: (srcKey=${srcKey}, dstKey=${dstKey})`
-    );
+    if (srcKey === dstKey) {
+      throw new Error(
+        `Renaming an object to its current name is not a valid operation (srcKey=${srcKey}, dstKey=${dstKey}).`
+      );
+    }
+
+    await this.copy(srcKey, dstKey);
+    await this.delete(srcKey);
```
tsuf239, line 64: Do we need to put it in a try-catch? 
@Chriscbr WDYT?

tsuf239 (approved):
Well done!

libs/wingsdk/src/shared-gcp/bucket.inflight.ts
garysassano: Why would you do that? You already check that the object exists so the copy operation cannot fail. You could call `tryDelete()` instead of `delete()` if you wanted, but that seems unecessary.

garysassano: Why would you do that? You already check that the object exists so the copy method cannot fail. You could call `tryDelete()` instead of `delete()` if you wanted, but I think that you should actually throw the error.

PR #5411, opened 2024-01-02
feat: nodeof built-in function

### LINKED ISSUE #2027: API for reflecting on the construct tree (`nodeof`)
### Feature Spec

The `nodeof` keyword can be used to obtain a reference to the [`constructs.Node`](https://docs.aws.amazon.com/cdk/api/v2/docs/constructs.Node.html) object behind any Wing resource (or construct).

The `Node` class offers a rich API for reflecting on the construct tree. For example, the `id` property returns the identifier of the construct, the `path` property returns the full path, the `parent` property returns the parent construct, etc.

### Use Cases

Being able to reflect on construct metadata is useful for advanced scenarios.

See [this] as an example.

[this]: (https://github.com/winglang/research/blob/0e0b0a5d384824cd62925705985bc19b788c910c/dogfooding/sdk-in-wing/bucket.tf-aws.w#L325)

### Workaround

There is currently a workaround that allows accessing the node like this:

You will need to explicitly install the `constructs` library:

```sh
npm i constructs
```

And then, use the `Node.of()` static method to obtain the node:

```js

resource Foo {
  init() {}
}

let f = new Foo();
let node = constructs.Node.of(f);
log(node.path);
// root/Default/Foo
```

### Component

Compiler, SDK

### Community Notes

Chriscbr:
Closes #2027

This PR adds a `nodeof()` function, a handy built-in for reflecting on the application tree (aka construct tree) in preflight.

## Checklist

- [x] Tests added (always)


libs/wingc/src/nodeof_transform.rs
```diff
@@ -0,0 +1,77 @@
...
+
+		let ExprKind::Reference(Reference::Identifier(Symbol { ref name, .. })) = callee.kind else {
+			return Expr {
+				kind: ExprKind::Call {
+					callee: CalleeKind::Expr(callee),
+					arg_list,
+				},
+				id: expr.id,
+				span: expr.span,
+			};
```
Chriscbr, line 45: For some reason just returning `fold::fold_expr(self, expr)` in some of these early-return branches causes me to get borrow checking errors... I'm not sure if there's a better way to avoid this boilerplate

libs/wingc/src/parser.rs
```diff
@@ -129,6 +129,12 @@ static RESERVED_WORDS: phf::Set<&'static str> = phf_set! {
 	"Array",
 	"MutArray",
 
+	// wing global functions
+	"log",
+	"assert",
+	"unsafeCast",
+	"nodeof",
+
```
Chriscbr, line 137:
```suggestion
```

libs/wingc/src/ast.rs
```diff
@@ -309,6 +309,7 @@ pub enum UtilityFunctions {
 	Log,
 	Assert,
 	UnsafeCast,
+	Nodeof,
```
staycoolcall911, line 312: Can you please add `NodeOf` to the language reference doc?

libs/wingc/src/nodeof_transform.rs
```diff
@@ -0,0 +1,77 @@
...
+			_ => {
+				return Expr {
+					kind: ExprKind::Call { callee, arg_list },
+					id: expr.id,
+					span: expr.span,
+				};
+			}
+		};
+
+		let ExprKind::Reference(Reference::Identifier(Symbol { ref name, .. })) = callee.kind else {
```
MarkMcCulloh, line 37: Does changing this line help with the inability to use `fold::fold_expr(self, expr)`? Probably not but I figured I'd shoot my shot.

```suggestion
		let ExprKind::Reference(Reference::Identifier(Symbol { name, .. })) = &callee.kind else {
```

libs/wingc/src/ast.rs
```diff
@@ -328,6 +330,7 @@ impl Display for UtilityFunctions {
 			UtilityFunctions::Log => write!(f, "log"),
 			UtilityFunctions::Assert => write!(f, "assert"),
 			UtilityFunctions::UnsafeCast => write!(f, "unsafeCast"),
+			UtilityFunctions::Nodeof => write!(f, "nodeof"),
```
MarkMcCulloh, line 333: Not a blocker, but `nodeof` rather than `nodeOf` feels weird since it's not camelcase. I understand why it can kinda feel right when you compare it to something like `typeof` from js, but the distinction there is that `typeof` is an operator whereas `nodeOf` is written as a function.

libs/wingc/src/lib.rs
```diff
@@ -273,15 +274,49 @@ pub fn type_check(
...
+				variadic: false,
+			}],
+			return_type: std_node,
+			phase: Phase::Preflight,
+			js_override: None, // implementation replaced by nodeof_transform
+			docs: Docs::with_summary("Obtain the tree node of a preflight resource."),
+		}),
+		scope,
+		types,
+	);
```
MarkMcCulloh, line 316: Instead of doing this here and refactoring `add_jsii_module_to_env`, would it be reasonable to put this `add_builtin` with the others in the typechecker and then do the transform after tc is complete?

libs/wingc/src/ast.rs
Chriscbr: I'm open to either, WDYT @eladb?

eladb: I prefer `nodeof`.

libs/wingc/src/ast.rs
Chriscbr: Done

libs/wingc/src/nodeof_transform.rs
Chriscbr: (removed this whole file - resolving)

libs/wingsdk/src/helpers.ts
```diff
@@ -24,3 +37,7 @@ export function range(start: number, end: number, inclusive: boolean) {
   }
   return iterator();
 }
+
+export function nodeof(construct: Construct): Node {
+  return Node.of(construct);
```
MarkMcCulloh, line 42: Does this work to avoid extra guff bundled when inflights use helpers?

```suggestion
  const Node = require("./std/node").Node;
  return Node.of(construct);
```

MarkMcCulloh (approved):

libs/wingsdk/src/helpers.ts
Chriscbr: Just tried it out, it looks like the bundle size is the same both ways

libs/wingc/src/lib.rs
Chriscbr: Done 👍

PR #5415, opened 2024-01-03
chore(sdk): add test and docs for batch usage of ulid

skorfmann:
this adds docs (and a test) for the fact that ulid works with a 1ms precision only, which might cause inconsistent ordered ids for multiple calls to `ulid` at the exact same time. The generated ids are still unique, but might not reflect the order of creation.

## Checklist

- [ ] Tests added (always)


examples/jsii-fixture/dist/js/jsii-fixture@0.0.0.jsii.tgz
Chriscbr, line 1: Remove

docs/docs/04-standard-library/util/api-reference.md
```diff
@@ -256,6 +256,13 @@ util.ulid(options?: UlidOptions);
 
 Generates universally unique lexicographically sortable identifier.
 
+This works with a precision of 1ms, which might not be sufficient for some cases where the strict order of
+creation is a desired property - e.g. batch creation of resources. In these cases, the generated IDs will
+be still unique (can generate 1.21e+24 unique ULIDs per millisecond) but might not be strictly reflecting the order of creation.
+To mitigate this, you could sleep for 1ms between creation of resources. However, in distributed
+systems where the strict order of creation is a requirement, this is not sufficient
+and you should consider using a different ID generation strategy.
```
Chriscbr, line 264: Out of curiosity are you still using `ulid` for your use case or did you switch to another library/solution? (just trying to gauge usage)

examples/tests/valid/ulid.test.w
```diff
@@ -0,0 +1,35 @@
+bring util;
+bring expect;
+
+inflight class Sortable {
+  pub static extern "./sort.js" sortStringListAsc(arr: Array<str>): Array<str>;
+  pub static extern "./sort.js" sortStringListDesc(arr: Array<str>): Array<str>;
```
Chriscbr, line 6: Maybe we can add a note to remove these in the future
```suggestion
  // TODO: https://github.com/winglang/wing/issues/4998
  pub static extern "./sort.js" sortStringListAsc(arr: Array<str>): Array<str>;
  pub static extern "./sort.js" sortStringListDesc(arr: Array<str>): Array<str>;
```

Chriscbr (approved):

examples/tests/valid/ulid.test.w
```diff
@@ -0,0 +1,35 @@
...
+}
+
```
staycoolcall911, line 8: Makes sense to test it in preflight as well, right?

staycoolcall911 (approved):

docs/docs/04-standard-library/util/api-reference.md
skorfmann: not using it right now, but would personally use this for most of the cases where I want to generate ids for records.

examples/tests/valid/ulid.test.w
skorfmann: just tried it, but ran into issues with other tooling to force a sleep. so just added a simple preflight test.

- https://github.com/winglang/wing/pull/5417
- https://github.com/winglang/wing/issues/5418

PR #5425, opened 2024-01-05
fix: `util.shell()`, `util.sleep()` and `util.spawn()` do not work in preflight

### LINKED ISSUE #5418: util.sleep is only available inflight
### I tried this:

use `util.sleep` in preflight 

### This happened:

only available inflight 

### I expected this:

use it preflight 

### Is there a workaround?

was trying to use util.shell - but can't due to https://github.com/winglang/wing/pull/5417

### Anything else?

_No response_

### Wing Version

main branch

### Node.js Version

18.7

### Platform(s)

MacOS

### Community Notes

### LINKED ISSUE #5420: util shell does not work in preflight
### I tried this:

```
class PreflightTest {
  pub output: str;

  new() {
    this.output = util.shell("echo Hello, Wing!");
  }
}

let preflightTest = new PreflightTest();

test "preflight" {
  expect.equal(preflightTest.output, "Hello, Wing!\n");
}
```

### This happened:

```
runtime error: Unable to lift object of type Promise
  --> examples/tests/sdk_tests/util/shell.test.w:1:1
1 | bring util;
  | ^
```

### I expected this:

shell out in preflight 

### Is there a workaround?

use nodejs external 

### Anything else?

extracted from https://github.com/winglang/wing/pull/5417/

### Wing Version

0.54.21

### Node.js Version

18.7

### Platform(s)

MacOS

### Community Notes

Chriscbr:
Currently, `util.shell` does not behave as expected when called in preflight code because the underlying implementation returns a promise. There were a couple possible solutions for this:

- We could restrict the function as inflight-only, but this wouldn't address use cases for performing shell operations in preflight.
- We could change the implementation to use synchronous Node.js APIs, but this would harm performance of inflight code.

Neither of these address the issue fully. Instead, I've opted to allow unphased functions in the SDK to have both preflight and inflight implementations. The compiler is then able to use its existing phase information to emit the name of either the preflight or inflight method implementation when producing JavaScript code.

In the process of writing this, I also:
- added phase-specific implementations for the `fs` module
- added a preflight implementation of `util.sleep` that relies on the fact that we currently execute preflight code inside a worker thread

I'm not sure whether the second change is one that we want to keep but it's cool that it works.

Fixes #5418
Fixes #5420

## Checklist

- [x] Tests added (always)


examples/tests/sdk_tests/util/shell.test.w
```diff
@@ -3,10 +3,11 @@ bring expect;
 bring fs;
 
 class Util {
-  extern "./util.js" pub static inflight platform(): str;
+  extern "./util.js" pub static inflight platformInflight(): str;
+  extern "./util.js" pub static platformPreflight(): str;
 }
 
-let assertThrows = inflight (expected: str, block: (): void) => {
+let assertThrowsPreflight = (expected: str, block: (): void) => {
```
Chriscbr, line 10: Lots of duplication in this file, but it'll all get consolidated in due time

libs/wingsdk/src/util/util.ts
```diff
@@ -259,9 +270,19 @@ export class Util {
...
+   * @inflightImpl __sleep__inflight
    */
-  public static async sleep(delay: Duration): Promise<void> {
+  public static sleep(delay: Duration): void {
+    Atomics.wait(
+      new Int32Array(new SharedArrayBuffer(4)),
+      0,
+      0,
+      delay.seconds * 1000
+    );
```
MarkMcCulloh, line 281: When I first saw the Atomics API I got excited because I assumed it was a great hack for synced-async, and then came the sadness of the big limitation.

It's cool that this works now even though I think we'll eventually get off of worker threads and have to use a stinky execSync for sleep.

libs/wingsdk/src/util/util.ts
```diff
@@ -135,32 +141,37 @@ export class Util {
    * @param opts `ShellOptions`, such as the working directory and environment variables.
    * @returns The standard output of the shell command.
    * @throws An error if the shell command execution fails or returns a non-zero exit code.
+   * @inflightImpl __shell__inflight
```
MarkMcCulloh, line 144: nit: this reminded me of python so it looked kinda weird 🌚 Doesn't really matter but maybe a more direct convention?

```suggestion
   * @inflightImpl shell$$inflight
```

I assume a leading underscore was needed because it's marked as `@internal`, but I would also suggest just making it `private` instead. Thanks to the magic of JS, it's still accessible so it should still work.

libs/wingsdk/src/util/util.ts
```diff
@@ -337,5 +358,25 @@ export class Util {
...
+  private static prepareShellOpts(
+    opts: ShellOptions = {}
+  ): NodeExecSyncOptions & NodeExecOptions {
+    return {
+      cwd: opts.cwd,
+      env:
+        opts.inheritEnv === true
+          ? { ...process.env, ...opts.env }
+          : { ...opts.env },
+    };
```
garysassano, line 371: Missing `windowsHide: true`

eladb (approved):
Looks good. See my thoughts about a possible alternative approach.

libs/wingc/src/type_check.rs
```diff
@@ -801,6 +801,8 @@ pub struct FunctionSignature {
 	/// - `$args$`: the arguments passed to this function call
 	/// - `$args_text$`: the original source text of the arguments passed to this function call, escaped
 	pub js_override: Option<String>,
+	/// During jsify, calls to this function will be replaced with this other function name
+	pub inflight_impl: Option<String>,
```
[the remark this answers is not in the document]
Chriscbr: I think you're onto something...

Chriscbr, line 805: I'd like to try avoiding this new machinery by reusing the `_toInflightType` methods we already support

libs/wingsdk/src/util/util.ts
```diff
@@ -259,9 +270,19 @@ export class Util {
...
    */
-  public static async sleep(delay: Duration): Promise<void> {
+  public static sleep(delay: Duration): void {
+    Atomics.wait(
+      new Int32Array(new SharedArrayBuffer(4)),
+      0,
+      0,
+      delay.seconds * 1000
+    );
+  }
```
MarkMcCulloh, line 282: With https://github.com/winglang/wing/pull/6157, this approach will no longer work.
I think the only "reasonable" way left would be doing it via a shell.

PR #5434, opened 2024-01-12
feat: use a single `.bak` instead of generating `.tmp` directories every time

### LINKED ISSUE #4988: Revisit the technique to use a `.tmp` directory
It has a few issues:

1. It's impossible to embed paths to the output directory in the synthesized terraform because the path will change.
2. It leaves tons of junk.

One possible solution is to simply delete the old one and create a new one. This means that if synthesis fails, we end up with a corrupted target directory. That's okay I think.

If we insist on transactionality, instead of creating a new temp directory every time, we can rename the existing one to `.bak` and then create a new one with the final name. If creation fails, we can revert the `.bak` to the original name.

MarkMcCulloh:
Fixes #4988

When starting compilation, if there is already an existing synth directory, it will be renamed to "{dirname}.bak". Files that are not part of the generated output (e.g. terraform state files) will be copied from that backup before running the rest of the process.

To facilitate this, a new `.manifest` json file is created after compilation that lists all the files generated.

This PR looks like a bigger diff than it really is because I added a `try {} finally {}` block around most of the compilation process.

Couple thoughts:
- The current method will always replace a previous backup with a new one, regardless of whether the compilation failed or not. I went back and forth on whether this is important or not
- The main synth dir will no longer always be valid. I think it would be interesting to move (or just delete) failed compilations, but this causes issues with sourcemaps.
- I think there could be more use cases for other data in the .manifest, which is why it's named a little bit generically. It's definitely still not part of the public contract though, which is why it's a dotfile

### DIFF
--- a/apps/wing/src/commands/compile.test.ts
+++ b/apps/wing/src/commands/compile.test.ts
@@ -28,9 +28,9 @@ describe(
       expect(files.length).toBeGreaterThan(0);
       expect(files).toMatchInlineSnapshot(`
         [
+          ".manifest",
           ".wing",
           "connections.json",
-          ".manifest",
           "simulator.json",
           "tree.json",
         ]
@@ -63,9 +63,9 @@ describe(
       expect(files.length).toBeGreaterThan(0);
       expect(files).toMatchInlineSnapshot(`
         [
+          ".manifest",
           ".wing",
           "connections.json",
-          ".manifest",
           "simulator.json",
           "tree.json",
         ]
@@ -136,9 +136,9 @@ describe(
         expect(files.length).toBeGreaterThan(0);
         expect(files).toMatchInlineSnapshot(`
           [
+            ".manifest",
             ".wing",
             "connections.json",
-            ".manifest",
             "simulator.json",
             "tree.json",
           ]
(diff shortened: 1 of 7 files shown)


apps/wing/src/commands/pack.test.ts
```diff
@@ -192,6 +192,7 @@ describe("wing pack", () => {
 
     expect(Object.keys(tarballContents).sort()).toMatchInlineSnapshot(`
       [
+        "$lib/.manifest",
```
Chriscbr, line 195: Maybe we can omit this if it's not needed for TS usage?

apps/wing/src/commands/compile.test.ts
Chriscbr, line 1: Can we add a test that validates the flow of retaining non-generated files? e.g. "If I compile an app, write a .tfstate file into /target, then recompile the app, then the .tfstate file should still be there" If we already have such a test then I think we're all good 👍

Chriscbr (approved):

apps/wing/src/commands/compile.test.ts
MarkMcCulloh: There is indeed a test for that already (you added it ~9 months ago 🌚)

I only added to that test to verify the manifest file does not contain the added file.

eladb:
awesome!

Issue #4324, opened 2023-09-28
captured field is undefined even when initialized


eladcon:
### I tried this:

`wing it` this code:
```ts
bring cloud;
bring util;

class T {
  pub val: str;
  init() {
    this.val = "123";
  }
}

class P {
  pub t: T;

  init() {
    new cloud.Service(inflight () => {
      this.onstart();
    });
      
    this.t = new T();
  }

  inflight onstart() {
    log("${this.t?.val}");
  }
}

let p = new P();
```

### This happened:

log output is `undefined` even though `this.t` is initialized

### I expected this:

either get a compiler error that im changing a value after it was already captured or  log `123`

### Is there a workaround?

change the order of the statements

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
Interesting bug 🤔 

I'm not sure if there's a super concise explanation for why it's happening, but the root cause lies with how we bind preflight objects like `T` to inflight functions. When `new Service()` is initialized, the SDK starts stringing together associations between objects, such as that the inflight closure requires access to `this`. But the way things work now, we eagerly evaluate the transitive associations, including the association that "this" (which refers to `P`) has a method named "onstart" that depends on `this.t.val`. Since `T` hasn't been initialized, it becomes a runtime error.

I reckon there's probably a way to add an extra step of lazy evaluation inside the SDK to make this work..

ekeren:
@Chriscbr is this the same thing?

The following code:
```ts
class A {
  var connectFn: inflight(str): void;
  new() {
    this.connectFn = inflight () => {
      log("placeholder");
    };
    let connectFnProxy = inflight (s: str) => {
      log("wrapper");
      return this.connectFn(s);
    };

    new cloud.Service(inflight () => {
      connectFnProxy("service ");
    });
  }
  
  pub onConnect(handler: inflight(str): void): void {
    this.connectFn = handler;
  }
  pub inflight justForCapture(){
    // 
  }
}

let a = new A() as "a2";
a.onConnect(inflight (s:str) => {
  log("actual");
});


test "a" {
  a.justForCapture();
}
```
will output 
```
wrapper
placeholder
```

vs expected 
```
wrapper
actual
```

ekeren:
right now we have a undocumented workaround 

```
bring cloud;
class A impl std.IResource{
  var connectFn: inflight(str): void;
  new() {
[...]

PR #5435, opened 2024-01-12
chore(compiler): refactor lifting

Chriscbr:
Refactor some of how the "lifting" code is managed within the SDK so that instead of having two sets of methods for registering lifts (`_registerOnLift`) and then performing actual lifting (`onLift`), we just have one sets of methods for lifting (`onLift`, `onLiftType`). All lifting logic now happens after the construct tree has finished creation, during `preSynthesize`. 

I think will help address some issues I had writing a fix for https://github.com/winglang/wing/issues/4324.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/std/resource.ts
```diff
@@ -111,16 +102,50 @@ export interface IResource extends IConstruct, IHostedLiftable {
...
-   * operations. A type being lifted to a host means that that type's static members
-   * will be lifted to the host.
+   * A hook called by the Wing compiler once for each inflight host that needs to
+   * use this type inflight. The list of requested inflight methods
+   * needed by the inflight host are given by `ops`.
    *
+   * This method is commonly used for adding permissions, environment variables, or
+   * other capabilities to the inflight host.
+   */
+  public static onLiftType(host: IInflightHost, ops: string[]): void {
```
MarkMcCulloh, line 112: I feel like we can just go ahead and call this `onLift` as well. Lifting a class (i.e. a function) in the end isn't really different than lifting anything else.

Chriscbr: The issue I had was JSII doesn't allow instance and static members to have the same name :S but otherwise I agree

Chriscbr: Maybe I can just remove this from the abstract class

Chriscbr: I think it's possible but it'll require adding some more codegen, in any case I'm thinking to defer it for now if that's ok

libs/wingsdk/src/std/resource.ts
```diff
@@ -111,16 +102,73 @@ export interface IResource extends IConstruct, IHostedLiftable {
...
-   * operations. A type being lifted to a host means that that type's static members
-   * will be lifted to the host.
+   * A hook called by the Wing compiler once for each inflight host that needs to
+   * use this type inflight. The list of requested inflight methods
+   * needed by the inflight host are given by `ops`.
+   *
+   * This method is commonly used for adding permissions, environment variables, or
+   * other capabilities to the inflight host.
+   */
+  public static onLiftType(host: IInflightHost, ops: string[]): void {
```
MarkMcCulloh, line 112: Could this be marked internal/private? The addition of a new public static method on all resources is a bit distracting IMO.

MarkMcCulloh: Static inflights feel more like "inflight closures trapped in a namespace", and inflight closures don't have the option of adding associated capabilities either. Given that, it feels a little strange to expose this lifecycle hook to users, but you're right that there wouldn't be another place to do it.

MarkMcCulloh (approved):

libs/wingsdk/src/std/resource.ts
```diff
@@ -137,7 +185,7 @@ export abstract class Resource extends Construct implements IResource {
    *
    * @internal
    */
-  public static _registerOnLiftObject(
+  public static _onLiftObject(
```
eladb, line 188: Does this need to be on `Resource` and internally exposed or can this just be a private utility function somewhere?

eladb (approved):

libs/wingsdk/src/std/resource.ts
Chriscbr: Yeah it's true - maybe there's some incidental complexity here from the language's OO model. But I think it might be not too bad, we'll see

libs/wingsdk/src/std/resource.ts
```diff
@@ -111,16 +102,73 @@ export interface IResource extends IConstruct, IHostedLiftable {
...
+   * _onLiftObject(foo, host, ["other1", "other2"]);
+   * _onLiftObject(bar, host, ["other1", "other3"]);
+   * ```
    * @internal
    */
-  public static _registerOnLift(host: IInflightHost, ops: string[]): void {
-    // Do nothing by default
-    host;
-    ops;
+  public static _onLiftMatrix(
```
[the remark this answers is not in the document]
Chriscbr: Good point, yes this is as good of a time as ever to move these

libs/wingsdk/src/std/resource.ts
Chriscbr: Moved

libs/wingc/src/jsify.rs
```diff
@@ -1789,17 +1789,28 @@ impl<'a> JSifier<'a> {
...
-		for (method_name, method_qual) in lift_qualifications {
-			bind_method.open(format!("if (ops.includes(\"{method_name}\")) {{"));
-			for (code, method_lift_qual) in method_qual {
-				let ops_strings = method_lift_qual.ops.iter().map(|op| format!("\"{}\"", op)).join(", ");
 
-				bind_method.line(format!(
-					"{class_name}._registerOnLiftObject({code}, host, [{ops_strings}]);",
-				));
+		if !lift_qualifications.is_empty() {
+			bind_method.open(format!("{class_name}._onLiftMatrix(host, ops, {{"));
```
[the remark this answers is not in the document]
Chriscbr: Done

PR #5445, opened 2024-01-15
rfc: console sign in process

ainvoner:
this rfc describes the rationale and requirements for the Wing Console sign in process

## Checklist

- [ ] Tests added (always)


skyrpex:
So, we only want a single computer to do this sign in process once, right? If so, we can store a `signedIn` boolean along with the anonymous analytics ID and skip the process next time.

skyrpex:
I'd say to also share the GitHub/WingCloud user ID with the analytics engine, so we don't lose data.

ainvoner:
> So, we only want a single computer to do this sign in process once, right? If so, we can store a `signedIn` boolean along with the anonymous analytics ID and skip the process next time.

Yes, this is actually the last line of the RFC, we can just save a boolean indication instead of the token itself - sounds better!
you can update the RFC

ainvoner:
> I'd say to also share the GitHub/WingCloud user ID with the analytics engine, so we don't lose data.

We are already doing it today in Wing Cloud. We can do the same for Wing Console. But it means that we need to save this data on the user machine since we can't use browser local storage only
In Wing Cloud we are sending:
```js
export interface UserIdentity {
  id: string;
  name: string;
  email: string;
}
```

skyrpex:
> > I'd say to also share the GitHub/WingCloud user ID with the analytics engine, so we don't lose data.
> 
> We are already doing it today in Wing Cloud. We can do the same for Wing Console. But it means that we need to save this data on the user machine since we can't use browser local storage only In Wing Cloud we are sending:
> 
> ```js
> export interface UserIdentity {
>   id: string;
>   name: string;
>   email: string;
> }
> ```

I don't think we need to save anything at all (other than a boolean). We just need to send it to the analytics engine once.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+Events generated by signed in users will include the user's GitHub account email as an integral part of their user identity.
+In contrast, Wing CLI users are initially anonymous, and all analytics events originating from the CLI are sent anonymously with a unique identifier.
+However, when a user signs in to Wing Console, the unique identifier associated with CLI analytics events will be associated with the signed in user's identity, allowing us to link CLI analytics data to specific user accounts
+
+## Technical details
+
+Since we would like to have a single user entity in Wing, the process of signing in to Wing Console will use the same authentication mechanism as Wing Cloud.
+In fact, the user will actually be singing in to Wing Cloud and be redirected back to Wing Console without noticing.
+
+In order to make sure a sign in user won't need to sign in again, we will save the uses' authentication token in user's machine.
```
skorfmann:
```suggestion
In order to make sure a sign in user won't need to sign in again, we will save the uses' authentication token in user's machine.

## Next Steps

Add value for the user in exchange for signin up. Things which could be considered:

- Deploy to wing.cloud directly from the console. This could upload the working directory to a Bucket from which wing.cloud could deploy. The deployment status and the url would be shown in the local console.
- Offer a solution similar to ngrok, to make the locally running instance available for remote services (e.g. webhooks) or users (e.g. demo purposes)
```

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+
+New users of Wing Console will be required to sign in with their GitHub account before having access to the Wing Console application.
+
+Another requirement is to have only one user entity in Wing. This means that a user in Wing Console is also a user in Wing Cloud and vice versa.
+
+## User Experience
+
+### New Users
+
+New users that will `wing it` will see a sign in dialog blocking the Wing Console UI. The rationale behind this decision is that we want to show the user a teaser of Wing Console while asking them to sign in.
```
skorfmann: It should be made clear, that everything is still running locally. The user might get confused and believe that with the sign up, stuff will be running remotely. Questions like "will I expose my highly confidential code by doing this" should be addressed.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Support of multiple users on the same machine
+
+There would be no support for multiple users on the same machine. 
+
+## User identity for Analytics  
+
+Events generated by signed in users will include the user's GitHub account email as an integral part of their user identity.
+In contrast, Wing CLI users are initially anonymous, and all analytics events originating from the CLI are sent anonymously with a unique identifier.
+However, when a user signs in to Wing Console, the unique identifier associated with CLI analytics events will be associated with the signed in user's identity, allowing us to link CLI analytics data to specific user accounts
```
skorfmann: What about logging out locally?

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+This document describes the Wing Console Sign In process and rationale.
+
+## Background
+
+As we seek to find **Sales and Marketing Opportunities**, it has become increasingly important to establish a more direct and meaningful connection with our users.
+One of the ways to achieve this is by implementing a sign-in flow within the Wing Console application.
+
+## Requirements
+
+New users of Wing Console will be required to sign in with their GitHub account before having access to the Wing Console application.
```
skorfmann, line 22: The more enterprisier the audience, the more likely it'll be that people might not have a Github account. Any chance to track this somehow?

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+- **Author(s)**: @ainvoner
+- **Submission Date**: 2024-01-15
+- **Stage**: Proposal
+- **Stage Date**: 2024-01-15
+
+This document describes the Wing Console Sign In process and rationale.
+
+## Background
+
+As we seek to find **Sales and Marketing Opportunities**, it has become increasingly important to establish a more direct and meaningful connection with our users.
```
eladb, line 17:
```suggestion
As we seek to be able to understand how wing is used, it has become increasingly important to establish a more direct and meaningful connection with our users.
```

skyrpex:
So, the idea is to use some sort of OAuth with Wing Cloud and the Wing Console (which is localhost). There's security concerns about this:

- If we allow OAuth with localhost from Wing Cloud, it means any localhost app can trick you into logging in to Wing Cloud and gather whatever info we return: a token, a GitHub user, an email, etc
- Localhost is usually HTTP, not HTTPS, so those credentials can be stolen through Wi-Fi

In order to minimize the security issues, we should consider:

- Use HTTPS for the Console in localhost (so nobody can sniff the network to stole data)
- Use a (wing cloud) hosted version of the Console (if there's network connection)
- During the OAuth-like exchange, send a new analytics ID that is already connected to the Wing Cloud user. The Console would need to store this analytics ID and use it further on

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
eladb: How hard is it to add additional providers? If not too difficult then maybe we should.

Alternatively, we can add a "I don't have a GitHub account" link which will skip sign in and we can track analytics.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+
+New users of Wing Console will be required to sign in with their GitHub account before having access to the Wing Console application.
+
+Another requirement is to have only one user entity in Wing. This means that a user in Wing Console is also a user in Wing Cloud and vice versa.
+
+## User Experience
+
+### New Users
+
+New users that will `wing it` will see a sign in dialog blocking the Wing Console UI. The rationale behind this decision is that we want to show the user a teaser of Wing Console while asking them to sign in.
```
eladb, line 30: What's the relationship to the EULA?

Maybe the license agreement is a button from the sign up screen:

> By using the Wing Console you agree to these [terms and conditions]() 

or something like this.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+
+Once users will be signed in with their GitHub account they will be able to use Wing Console.
+
+### Singed In Users
+
+Signed in users to Wing Console will not need to sign in again. They will be able to use Wing Console as usual.
```
eladb, line 36: Worth mentioning that the user is per machine so any time the console is opened, if there was already a logged in user, it will be used.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Offline
+
+If the user is offline (no internet connection) they will be able to use Wing Console as usual without the need to sign in.
```
eladb, line 40: I am assuming this is an issue only for the first sign up, right?

eladb (approved):

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
skyrpex: Exactly.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
skyrpex: Yes, the relationship of the analytics ID and the Wing Cloud user will only happen once. We won't require this process again anymore.

skorfmann:
> * If we allow OAuth with localhost from Wing Cloud, it means any localhost app can trick you into logging in to Wing Cloud and gather whatever info we return: a token, a GitHub user, an email, etc

Any application running on the user machine could read the token we persist on disk (same goes for fly, aws, ...). Not saying that this is a good thing, but more or less the reality for a lot of dev tools. Here's a more detailed Oauth Device flow described https://rockset.com/blog/go-secure-command-line-using-oauth2-device-authorization/ (what e.g. AWS is using as well)

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
ainvoner: if we will allow to skip sign - no one will sign in.
I think, for initial phase, support only GitHub (we can add additional sign in providers later).

revitalbarletz: +1

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
```diff
@@ -0,0 +1,75 @@
...
+
+"By using the Wing Console you agree to these [terms and conditions](#)"
+
+### Singed In Users
+
+Signed in users to Wing Console will not need to sign in again. They will be able to use Wing Console as usual.
+
+### Offline
+
+If the user is offline (no internet connection) they will be able to use Wing Console as usual without the need to sign in.
```
revitalbarletz, line 46: Assuming they already signed in right?

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
skyrpex: We would need to do some refactor in Wing Cloud to allow other providers. It'd take some time right now.

> Alternatively, we can add a "I don't have a GitHub account" link which will skip sign in and we can track analytics.

That's what we do now, right? Just anonymous analytics.

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
skyrpex: Who knows about this legal stuff?

docs/contributing/999-rfcs/2024-01-15-console-sign-in.md
skyrpex: If they are offline, they won't be able to sign in anyway so we should allow them to just use the Console directly.

skyrpex:
> > * If we allow OAuth with localhost from Wing Cloud, it means any localhost app can trick you into logging in to Wing Cloud and gather whatever info we return: a token, a GitHub user, an email, etc
> 
> Any application running on the user machine could read the token we persist on disk (same goes for fly, aws, ...). Not saying that this is a good thing, but more or less the reality for a lot of dev tools. Here's a more detailed Oauth Device flow described https://rockset.com/blog/go-secure-command-line-using-oauth2-device-authorization/ (what e.g. AWS is using as well)

Yeah, exactly. I thought that to make it as safe as possible, we could generate a new analytics ID (which is already tied to the Wing Cloud user) and pass it as the OAuth token. It would be used by the Console to report analytics.

Any malicious apps getting this analytics ID won't be able to do any damage.

PR #5453, opened 2024-01-16
fix: captured field is undefined even when initialized

### LINKED ISSUE #3103: Lifted reassignable variable does not take on last calculated value
### I tried this:

```js
bring cloud;

let var x = 5;

let i1 = inflight (m: str) => {
  log("${x}");
};

let f1 = new cloud.Function(i1) as "f1";

x = 6;

test "test" {
  f1.invoke("");
}
```

### This happened:

Running `wing test example.w` will log "5"

### I expected this:

I expected it to log "6" since that is the last value assigned to `x`.

Note that if I move `x = 6` to be before the definition of `f1`, then it does print 6, but if I print it anytime after (including after the test) it prints 5.

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

0.22.18

### Wing Console Version

_No response_

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

### LINKED ISSUE #4324: captured field is undefined even when initialized

Chriscbr:
Fixes #4324
Fixes #3103

We have gone full lazy in the Wing compiler. This PR builds off of the changes in https://github.com/winglang/wing/pull/5435 so that instead of inflight hosts like `cloud.Function` calling `onLift()` on the handler during preSynth, now it is also the case that they call `toInflight()` on the handler during presynth. This fixes the bug in #4324 by ensuring `toInflight` is called after all[1] user-level preflight code has executed, so the inflight JavaScript code is generated after any changes to class fields have been made.

I think this change should hold water architecturally since calling `toInflight()` and `onLift()` are currently the only responsibilities performed during `preSynthesize()`. If users were allowed to add their own logic to preSynthesize (which would allow class fields to be mutated further etc.) then there could be new bugs introduced, but we should try avoiding supporting this if possible.

[1] with the exception of code inside `onLift` hooks

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify.rs
```diff
@@ -1458,7 +1458,7 @@ impl<'a> JSifier<'a> {
 
 			// TODO Hack to make sure closures match the IInflight contract from wingsdk
 			if class_type.is_closure() {
-				code.line("_hash = require('crypto').createHash('md5').update(this._toInflight()).digest('hex');")
+				code.line("_hash = $helpers.closureId();")
```
Chriscbr, line 1461: Not sure if $helpers is the right place since it's only needed in preflight. I'll move it to stdlib.core

libs/wingc/src/jsify.rs
```diff
@@ -1458,7 +1458,7 @@ impl<'a> JSifier<'a> {
 
 			// TODO Hack to make sure closures match the IInflight contract from wingsdk
 			if class_type.is_closure() {
-				code.line("_hash = require('crypto').createHash('md5').update(this._toInflight()).digest('hex');")
+				code.line(format!("_hash = {STDLIB_CORE}.closureId();"))
```
MarkMcCulloh, line 1461: `_hash` is currently being used to deduplicate handlers in various resources (like Api, Topic, etc.). So 2 inflights are "equal" for this purpose if their hashes are equal even if they are technically different instances.
The SDK tests are still passing because they're still generating hashes.

If we're going to change to this implementation (which I'm cool with), we should rename this field to `_id` or something and make sure we are using it consistently in the SDK tests. We also need to note that it's a global sequential id and shouldn't be used for anything expected to be stable between compilations. This is both because reordering the source changes it and because this id is globally shared in a possibly non-deterministic way.

Chriscbr: Got it, yeah I see the discrepancy now. Updating the name and the tests

MarkMcCulloh: btw why was this changed in this PR? The other changes don't seem to conflict with the hash generation

Chriscbr: I've added a longer comment in the code about this 😄

libs/wingc/src/jsify/snapshots/no_capture_of_identifier_from_inner_scope.snap
```diff
@@ -52,7 +52,7 @@ class $Root extends $stdlib.std.Resource {
   constructor($scope, $id) {
     super($scope, $id);
     class $Closure1 extends $stdlib.std.Resource {
-      _hash = $stdlib.core.closureId();
+      _id = $stdlib.core.closureId();
```
eladb, line 55: tip: just to avoid this change polluting your diff you could submit a quick PR to just change this and then rebase.

examples/tests/valid/mutation_after_class_init.test.w
```diff
@@ -0,0 +1,38 @@
...
+let q = new Queue();
+let c = new cloud.Counter();
+q.setConsumer(inflight (m: str) => {
+  log("received {m}");
+  c.inc();
+});
+
+test "hi" {
+  q.push("message1");
+  util.sleep(5s);
```
eladb, line 35: idea:

```js
sim.Queue.from(q).drain();
```

libs/awscdk/src/function.ts
```diff
@@ -54,6 +65,25 @@ export class Function extends cloud.Function implements IAwsFunction {
       architecture: Architecture.ARM_64,
       logRetention: logRetentionDays,
     });
+
+    // hack: accessing private field from aws_lambda.AssetCode
+    // https://github.com/aws/aws-cdk/blob/109b2abe4c713624e731afa1b82c3c1a3ba064c9/packages/aws-cdk-lib/aws-lambda/lib/code.ts#L266
+    const asset: Asset = (code as any).asset;
```
eladb, line 71: maybe add something defensive here:

```js
if (!asset.assetPath) {
  throw new Error("AWS CDK 'Asset' class no longer has an 'assetPath' property"); 
}
```

libs/awscdk/src/function.ts
```diff
@@ -54,6 +65,25 @@ export class Function extends cloud.Function implements IAwsFunction {
...
+
+  /** @internal */
+  public _preSynthesize(): void {
+    super._preSynthesize();
+
+    // produce an inflight code bundle using the latest information, including all
+    // changes made to captured variables/resources after the constructor
+    const bundle = createBundle(this.entrypoint);
+
+    // copy files from bundle.directory to this.assetPath
```
eladb, line 83: maybe rename is better/quicker?

libs/wingc/src/jsify.rs
```diff
@@ -1458,7 +1458,7 @@ impl<'a> JSifier<'a> {
 
 			// TODO Hack to make sure closures match the IInflight contract from wingsdk
 			if class_type.is_closure() {
-				code.line("_hash = require('crypto').createHash('md5').update(this._toInflight()).digest('hex');")
+				code.line(format!("_id = {STDLIB_CORE}.closureId();"))
```
eladb, line 1461: perhaps resist the temptation to rename `_hash` to `_id` or move to a different commit because this pollutes all snapshots and really hard to review

eladb (approved):
👑

libs/wingsdk/src/std/resource.ts
```diff
@@ -29,10 +29,32 @@ export interface IInflightHost extends IResource {
...
+   *   }
+   * });
+   *
+   * arr.push("world");
+   * ```
+   *
    * @internal
    */
-  _hash: string;
+  _id: number;
```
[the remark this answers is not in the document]
MarkMcCulloh: `fingerprint` and `hash` both imply a content-derived and stable value IMO, `id` isn't perfect but it better avoids that implication

libs/wingsdk/src/shared/convert.ts
```diff
@@ -14,18 +13,17 @@ export function convertBetweenHandlers(
...
-new (require("${normalPath(
-    newHandlerClientPath
-  )}")).${newHandlerClientClassName}({ handler: ${handlerClient}, args: ${JSON.stringify(
-    args
-  )} })`;
-
   const newHandler = {
     ...baseHandler,
-    _hash: createHash("md5").update(newCode).digest("hex"),
+    _id: baseHandler._id,
```
MarkMcCulloh, line 18: Shouldn't this generate a new id? Technically this is a new inflight

libs/wingc/src/jsify.rs
Chriscbr: oops. I'll try to keep in mind for the future

libs/wingsdk/src/cloud/function.ts
```diff
@@ -68,6 +68,8 @@ export class Function extends Resource implements IInflightHost {
 
     super(scope, id);
 
+    this.handler = handler;
```
yoav-steinberg, line 71: Why are we setting `this.handler` twice? (here and ~10 lines down again)

Chriscbr: Whoops! leftover from a merge conflicts, thanks for the catch!

examples/tests/valid/mutation_after_class_init.test.w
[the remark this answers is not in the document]
Chriscbr: Yeah - checking for a side effect that the non-dummy handler worked is the challenge here. I think I can change this to use the `util.waitUntil` utility to make it faster though.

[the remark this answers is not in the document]
yoav-steinberg: you can use two counters one to make sure the function was called (used in both the dummy and real handler) and one to make sure the real handler was called (used only in the real one).

libs/wingsdk/src/shared/convert.ts
Chriscbr: Fixed

eladb (approved):

Issue #3115, opened 2023-06-28
expected function to return type when trying to define an optional function


ekeren:
### I tried this:

```ts
struct Options {
  mappingFunction: ((num, str): str)? ; 
}
```

### This happened:

```sh
Failed to compile.

error: Expected function return type
  --> test.w:2:20
  |
2 |   mappingFunction: ((num, str): str)? ; 
  |                    ^^^^^^^^^^^^^^^^^ Expected function return type


error: Compiler bug (panicked at 'called `Result::unwrap()` on an `Err` value: ()', libs/wingc/src/parser.rs:910:34) during parsing, please report at https://docs.winglang.io/contributing/bugs
  --> test.w:1:1
  |  
1 | / struct Options {
2 | |   mappingFunction: ((num, str): str)? ; 
3 | | }
  | \-^ Compiler bug (panicked at 'called `Result::unwrap()` on an `Err` value: ()', libs/wingc/src/parser.rs:910:34) during parsing, please report at https://docs.winglang.io/contributing/bugs


```

### I expected this:

mappingFunction to be an optional function

### Component

Compiler

### Wing Version

0.22.27

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Issue #4296, opened 2023-09-26
Unable to declare a function signature that returns an optional function


eladb:
### I tried this:

I am trying to define a function signature of an inflight function that returns an optional inflight function.

```js
let fn: inflight (): (inflight (): void)? = inflight () => {};
```

### This happened:

VSCode:

```
Compiler bug (panicked at 'called `Result::unwrap()` on an `Err` value: ()', libs/wingc/src/parser.rs:1367:34) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
```

Console:

```
error: Expected function return type
  --> /Users/eladb/Desktop/test.main.w:5:22
  |
5 | let fn: inflight (): (inflight (): void)? = inflight () => {};
  |                      ^^^^^^^^^^^^^^^^^^^ Expected function return type


error: Compiler bug (panicked at 'called `Result::unwrap()` on an `Err` value: ()', libs/wingc/src/parser.rs:1367:34) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
  --> /Users/eladb/Desktop/test.main.w:5:1
  |
5 | let fn: inflight (): (inflight (): void)? = inflight () => {};
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Compiler bug (panicked at 'called `Result::unwrap()` on an `Err` value: ()', libs/wingc/src/parser.rs:1367:34) during parsing, please report at https://www.winglang.io/contributing/start-here/bugs
```

### I expected this:

To work or to have some other syntax for this. 

### Is there a workaround?

Type aliases would have been a potential workaround, but they are not supported yet (#3350).

### Component

Language Design, Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

yoav-steinberg:
Note that the above syntax:
```wing
let fn: inflight (): (inflight (): void)? = inflight () => {};
```
assumes a type annotation can be surrounded by parenthesis, which is wrong. What happens is that the compiler assumes the parenthesis indicate another function type which receives the given function as an argument, but then it crashes because there's no return type specifier.
I fixed the crash in #5450.

We need to consider supporting surrounding a type annotation with parenthesis to resolve this.

eladb:
> We need to consider supporting surrounding a type annotation with parenthesis to resolve this.

👍

[...]

PR #5455, opened 2024-01-17
fix(compiler): can't define optional function types

yoav-steinberg:
See #4296
We can now define optional function types via support for parenthesis around type annotations:
```wing
// Optional function type
var let fn: (():void)? = () => { };
fn = nil;
// Crazy parenthesis type annotation
let x: (((((((num))))))) = 1;
```

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -2552,23 +2552,11 @@ impl<'s> Parser<'s> {
 	}
 }
 
-/// Get actual child by field name when multiple childs exist because of extra nodes.
+/// Get actual child by field name when multiple children exist because of extra nodes.
 /// It should be safe to use this instead of `node.get_child_by_field_name`
 /// in cases where there's doubt.
 fn get_actual_child_by_field_name<'a>(node: Node<'a>, field_name: &str) -> Option<Node<'a>> {
-	let mut cursor = node.walk();
-
-	for child_node in node.children_by_field_name(field_name, &mut cursor) {
-		if child_node.is_extra() {
-			continue;
-		}
-		if !child_node.is_named() {
-			continue;
-		}
-
-		return Some(child_node);
-	}
-	None
+	get_actual_children_by_field_name(node, field_name).into_iter().next()
 }
 
 /// Get actual children by field name when the children might contain extra nodes.
@@ -2585,9 +2573,7 @@ fn get_actual_children_by_field_name<'a>(node: Node<'a>, field_name: &str) -> Ve
 		if !child_node.is_named() {
 			continue;
 		}
-		if let Some(child_node) = get_actual_child_by_field_name(child_node, field_name) {
-			children.push(child_node);
-		}
+		children.push(child_node);
 	}
 	children
 }
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -494,7 +494,7 @@ impl<'s> Parser<'s> {
 
 	fn get_child_field<'a>(&'a self, node: &'a Node<'a>, field: &str) -> DiagnosticResult<Node> {
 		let checked_node = self.check_error(*node, field)?;
-		let child = checked_node.child_by_field_name(field);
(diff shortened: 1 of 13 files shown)


eladb (approved):
Thank you lord

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2755,6 +2741,20 @@ error: Cannot call an optional function
    | ^^^^^^^^^^^^^^^^ Cannot call an optional function
 
 
+error: Expected type to be \\"(preflight (num): void)?\\", but got \\"preflight (x: str): void\\" instead
+    --> ../../../examples/tests/invalid/optionals.test.w:102:53
+    |
+102 | let functionWithOptionalFuncParam1: ((num):void)? = (x: str):void => {};
+    |                                                     ^^^^^^^^^^^^^^^^^^^ Expected type to be \\"(preflight (num): void)?\\", but got \\"preflight (x: str): void\\" instead
```
Chriscbr, line 2748: Unrelated to the PR changes but it feels a bit weird that some of the type signatures are formatted with parameter names, and some of the type signatures are formatted without type names

```
preflight (num): void
vs
preflight (x: num): void
```

libs/wingc/src/parser.rs
```diff
@@ -2562,6 +2552,32 @@ impl<'s> Parser<'s> {
 	}
 }
 
+/// Get actual child by field name when multiple children exist because of extra nodes.
+/// It should be safe to use this instead of `node.get_child_by_field_name`
+/// in cases where there's doubt.
+fn get_actual_child_by_field_name<'a>(node: Node<'a>, field_name: &str) -> Option<Node<'a>> {
```
Chriscbr, line 2558: Is the description saying that multiple children can have the same field name? I'm not sure I understand how the behavior of this function is different from `calling node.child_field_by_name(name)` (maybe a unit test would make it clearer)

Chriscbr (approved):

Chriscbr:
@yoav-steinberg Can you check if this fixes https://github.com/winglang/wing/issues/3115?

tools/hangar/__snapshots__/invalid.ts.snap
yoav-steinberg: I also noticed this, I'll add a new issue

libs/wingc/src/parser.rs
yoav-steinberg: I'll give a detailed explanation to make sure I understand correctly :) 
The tree-sitter API is indeed confusing. A node in tree-sitter can be a named node or unnamed node depending on how the grammar was defined. Also a node can have a **field** name (not to be confused with a named node).
* `node.get_children_by_field_name(field_name)` - returns all (can be multiple) children with the specified field name. e.g. in our grammar an interface definition can have multiple children with the "extends" field name. These children can be **unnamed** nodes too (usually not that useful). e.g. in our grammar "extends" nodes also include nodes for the commas separating the extend types: `interface I extends A,B {}` will have three sibling nodes with the field name "extends": `A`,`,`,`B`.
* `node.get_child_by_field_name(field_name)` - will return the first (or `None`) of the children returned by `get_children_by_field_name` above. This is very useful in places in the grammar where we can assume only a single child. Like in a class "extends" (vs an interface "extends").
* `get_actual_children_by_field_name(node, field_name)` - Will return all (can be multiple) children that are **named**, this filters out the the comma nodes in the example above.
* `get_actual_child_by_field_name(node, field_name)` - Will return the the fist (or `None`) of the children returned by `get_actual_children_by_field_name`, this is the most useful function since it filters both on a field name and named nodes and assumes a single result. In most places in the grammar this should work. e.g. a binary expression node `a + b` has a single **named** `operator` node with a field name called `op`.

yoav-steinberg: Added a unit test for `get_actual_children_by_field_name`

Issue #5383, opened 2024-01-01
`+=` isn't working for strings


tsuf239:
### I tried this:

```
let var message = "Yay! another user registered to the raffle!\n";
message  += "*email:* {email} *name:* {name}\n";
```

### This happened:

<img width="758" alt="image" src="https://github.com/winglang/wing/assets/39455181/f92b08a7-d93e-47e1-94a0-58be99d47e6d">

### I expected this:

to just work :) 

### Is there a workaround?

```
message = message + "*email:* {email} *name:* {name}\n";

```

### Anything else?

Discovered while working on this project: https://github.com/winglang/wing-raffle

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

lausuarez02:
Hey i would like to take this one but, where exactly should i tackled this issue? folder/architecture wise

staycoolcall911:
You got it @lausuarez02 :)
Here's how Chris added string concat with + a while ago: #2429 
It should give you a hint how to start. 
You should have some relevant guides on our doc site, such as this one: https://www.winglang.io/contributing/start-here/development#how-do-i-make-changes-to-the-wing-grammar

Warkanlock:
@lausuarez02 are you going to take this one? I might be able to do it during my weekend

revitalbarletz:
@Warkanlock I just assigned this issue to Ignacio, how about joining our slack channel and continue this discussion there https://t.winglang.io/slack ?

revitalbarletz:
@Warkanlock its all yours :)

PR #5456, opened 2024-01-17
feat(compiler): add support for `string` on `+=` for assignaments

Warkanlock:
Related to #5383

## Description

Based on #5383 request, we wanted to add the `+=` operator support to `string` types, so we are now able to achieve the following and compile `.wing` files correctly

```rust
let var initial = "hey";
initial += "you!";
```


examples/tests/invalid/types_strings_arithmetic.test.w
```diff
@@ -8,4 +8,9 @@ let strExample = "Hello World";
 let numExample = 2;
 
 let e3 = "{strExample}!" * numExample;
-// ERR   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Cannot multiply a string with a number
\ No newline at end of file
+// ERR   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Cannot multiply a string with a number
+
+let a = 2;
+let b = "2";
+let b += a;
```
yoav-steinberg, line 15: You probably wanted the third assignment to update `b` and note re-declare it.
```suggestion
let a = 2;
let var b = "2";
b += a;
```

examples/tests/valid/std_string.test.w
```diff
@@ -18,6 +18,10 @@ assert("hello" + " world" == "hello world");
 assert(
 "hello {"funky"}
  world" == "hello funky\n world");
+
+let var initial = "hey, ";
+initial += "you";
+assert(initial == "hey, you");
```
yoav-steinberg, line 24: :clap:

libs/wingc/src/lib.rs
```diff
@@ -276,6 +276,7 @@ pub fn compile(
 	// Wing files, then move on to files that depend on those, and repeat)
 	for file in &topo_sorted_files {
 		let mut scope = asts.remove(file).expect("matching AST not found");
+
```
yoav-steinberg, line 279:
```suggestion
```
Lets keep the diff with only relevant changes

libs/wingc/src/parser.rs
```diff
@@ -1924,6 +1924,7 @@ impl<'s> Parser<'s> {
 	fn build_reference(&self, reference_node: &Node, phase: Phase) -> DiagnosticResult<Expr> {
 		let actual_node = reference_node.named_child(0).unwrap();
 		let actual_node_span = self.node_span(&actual_node);
+
```
yoav-steinberg, line 1927:
```suggestion
```

libs/wingc/src/parser.rs
```diff
@@ -1983,6 +1984,7 @@ impl<'s> Parser<'s> {
 		let expression_span = self.node_span(exp_node);
 		CompilationContext::set(CompilationPhase::Parsing, &expression_span);
 		let expression_node = &self.check_error(*exp_node, "expression")?;
+
```
yoav-steinberg, line 1987:
```suggestion
```

libs/wingc/src/type_check.rs
```diff
@@ -4247,12 +4248,22 @@ impl<'a> TypeChecker<'a> {
...
+			}
+			_ if var.type_.is_subtype_of(&self.types.number())
+				&& matches!(kind, AssignmentKind::AssignIncr | AssignmentKind::AssignDecr) =>
+			{
+				self.validate_type(exp_type, self.types.number(), value);
+				self.validate_type(var.type_, self.types.number(), variable);
+			}
+			_ => {
+				self.validate_type(exp_type, var.type_, value);
+			}
```
yoav-steinberg, line 4265: The logic seems right, but using `match` here when everything is prefixed with a `_` seems cumbersome.
Try changing to a simple `if .. else if ..` chain:
```rust
if matches!(kind, AssignmentKind::AssignIncr) {
  self.validate_type_in(expr_type, &[self.types.number(), self.types.string()], value);
  self.validate_type_in(var.type_, &[self.types.number(), self.types.string()], value);
} else if matches!(kind, AssignmentKind::AssignDecr) {
  self.validate_type(expr_type, self.types.number(), value);
  self.validate_type(var.type_, self.types.number(), value);
}
```
or something similar

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -3659,7 +3659,21 @@ Duration <DURATION>"
...
+
+
+error: Unexpected 'identifier'
+   --> ../../../examples/tests/invalid/types_strings_arithmetic.test.w:15:10
+   |
+15 | let b += a;
+   |          ^ Unexpected 'identifier'
+
+
+error: Binary operator '+' cannot be applied to operands of type 'num' and 'str'; only (num, num) and (str, str) are supported
```
yoav-steinberg, line 3676: this is a good hint that the test isn't doing what we expected it to do

libs/wingc/src/type_check.rs
Warkanlock: I was on the same line before making this change (second commit); you are right, I'll rollback

Warkanlock: In addition, thanks for the validate_type_in; I didn't know it's existence

tools/hangar/__snapshots__/invalid.ts.snap
Warkanlock: Is there an easy way to generate this `hangar/*.snap` files from the CLI? 

is it ``pnpm turbo test`` sufficient?

yoav-steinberg:
```
cd tools/hangar
turbo test
```
But that still takes a pretty long time. There's probably a quicker way, but I'm not sure, try the dev channel on slack.
You can also just run the compiler on the specific test and see the output:
```
cd libs/wingc
cargo run --example compile -- ../../examples/tests/invalid/something.test.w
```

Warkanlock: thanks!

yoav-steinberg (approved):
:partying_face:

NathanTarbert:
Thanks, @Warkanlock for your contribution! ❤️

Issue #4791, opened 2023-11-03
Unable to extend jsii classes, such as`cdk8s.Chart` in non-entrypoint files


eladb:
### I tried this:

I am trying to write a module (non-entrypoint) with the following code:

```js
bring "cdk8s" as cdk8s;

class Foo extends cdk8s.Chart { }
```

### This happened:

```
ERROR: Cannot read properties of undefined (reading 'node')

test/target/test/containers.test.wsim.202680.tmp/.wing/preflight.js:6
   const $wing_is_test = process.env.WING_IS_TEST === "true";
   const std = $stdlib.std;
>> const containers = require("./preflight.containers-12.js")({ $stdlib });
   const expect = $stdlib.expect;
   const http = $stdlib.http;

--------------------------------- STACK TRACE ---------------------------------
/Users/eladb/code/containers/test/target/test/containers.test.wsim.202680.tmp/.wing/preflight.workload-10.js:70
  class _Chart extends (this.node.root.typeForFqn("cdk8s.Chart") ?? cdk8s.Chart) {
                             ^

TypeError: Cannot read properties of undefined (reading 'node')
    at module.exports (/Users/eladb/code/containers/test/target/test/containers.test.wsim.202680.tmp/.wing/preflight.workload-10.js:70:30)
    at module.exports (/Users/eladb/code/containers/test/target/test/containers.test.wsim.202680.tmp/.wing/preflight.tfaws-11.js:5:45)
    at module.exports (/Users/eladb/code/containers/test/target/test/containers.test.wsim.202680.tmp/.wing/preflight.containers-12.js:6:49)
    at evalmachine.<anonymous>:7:59
    at Script.runInContext (node:vm:135:12)
    at Object.runInContext (node:vm:283:6)
    at runPreflightCodeInVm (/Users/eladb/code/containers/node_modules/@winglang/compiler/dist/index.js:464:8)
    at async Object.compile (/Users/eladb/code/containers/node_modules/@winglang/compiler/dist/index.js:401:5)
    at async compile (/Users/eladb/code/containers/node_modules/winglang/dist/commands/compile.js:63:16)
    at async testOne (/Users/eladb/code/containers/node_modules/winglang/dist/commands/test/test.js:115:22)
    at async testFile (/Users/eladb/code/containers/node_modules/winglang/dist/commands/test/test.js:77:39)
    at async Promise.all (index 0)
[...]

PR #5459, opened 2024-01-18
fix(compiler): unable to extend jsii classes in non entrypoint

hasanaburayyan:
Have some more changes to make in TODO, but wanted to let E2E run in CI

## TODO:
- [ ] Some tidy up
- [ ] add some more tests to show extending jsii classes in non-entrypoint
- [ ] Doc updates to IPlatform API

Closes: https://github.com/winglang/wing/issues/4791

### Highlights
- Removed `new()`, `tryNew()` methods from `core/App` This is in an effort to remove some overloaded responsibilities from Apps.
- Concrete class instantiation is now determined by the Platform Manager
- Moved `typeForFqn` to `IPlatform` API, this now allows platforms to override type -> fqn resolution
- Creates a `platform_manager.js` generated file that is required by each preflight class

## Summary

In this PR I have refactored the way that we handle resolving concrete types, previously we have kinda overloaded the responsibilities of the `App` class. Not only was it responsible for synthesizing an app but also for determining concrete types via the `new()` and `tryNew()` methods, also Apps would manage resolving types from fqns via `typeFromFqn()` method.

These changes remove the `new()`, `tryNew()`, and `typeForFqn()` methods from the App class. Now all that responsibility is pawned off onto the Platforms and PlatformManager. 

Notable changes in the snapshots are that we no longer call `scope.node.root.new` now these calls are routed through the platform manager `$PlatformManager.new` the same for `scope.node.root.typeForFqn` is now `$PlatformManager.typeForFqn`

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -74,33 +131,43 @@ export class PlatformManager {
...
-
-    let newInstanceOverrides: any[] = [];
+    return appCall!({
+      ...appProps,
+      _platformManager: this,
+      synthHooks: {
+        preSynthesize: this.hooks.preSynth,
+        postSynthesize: this.hooks.postSynth,
+        validate: this.hooks.validate,
+      },
```
hasanaburayyan, line 141: I really wanted to remove the responsibility for calling the Synth hooks from the App class that way we can enforce the call order contract at this level vs the App level, but Im going to save that for a future iteration.

libs/wingsdk/src/core/app.ts
```diff
@@ -225,8 +213,13 @@ export abstract class App extends Construct implements IApp {
     id: string,
     ...args: any[]
   ): any {
-    // next delegate to "tryNew", which will allow derived classes to inject
-    const instance = this.tryNew(fqn, scope, id, ...args);
+    // uses the Back Reference for the platform manager to determine which what type to create
+    const instance = this._platformManager?._tryNewInstance(
```
Chriscbr, line 217: If `this._platformManager` is undefined, then the method will always throw an error. Can we change `this._platformManager` to be a required field?

libs/wingsdk/src/shared-tf/app.ts
```diff
@@ -62,25 +61,7 @@ export abstract class CdktfApp extends App {
...
-
-    (cdktfApp as any).newAbstract = (
-      fqn: string,
-      scope: Construct,
-      id: string,
-      ...args: any[]
-    ) => this.newAbstract(fqn, scope, id, ...args);
-
-    (cdktfApp as any).typeForFqn = (fqn: string) => this.typeForFqn(fqn);
+    // (cdktfApp as any).typeForFqn = (fqn: string) => this.typeForFqn(fqn);
```
Chriscbr, line 64: Nice cleanup!

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -2,25 +2,43 @@ import { readFileSync } from "fs";
...
   readonly platformPaths?: string[];
 }
 
+interface PlatformHooks {
+  newInstance: any[];
+  typeForFqn: any[];
+  preSynth: any[];
+  postSynth: any[];
+  validate: any[];
+}
```
Chriscbr, line 21: Docstrings

libs/wingsdk/test/target-sim/api.test.ts
```diff
@@ -25,6 +26,8 @@ const INFLIGHT_CODE_WITH_CONTENTTYPE_RESPONSE_HEADER = `async handle(req) { retu
 // Handler that responds to a request without a response body
 const INFLIGHT_CODE_NO_BODY = `async handle(req) { return { status: 200 }; }`;
 
+const _platformManager = new PlatformManager({ platformPaths: ["sim"] });
```
Chriscbr, line 29: Can this be removed?

tools/hangar/src/unsupported.test.ts
```diff
@@ -30,7 +30,7 @@ test("unsupported resource in target", async ({ expect }) => {
     },
   });
 
-  expect(sanitizeOutput(result.stderr)).toMatchInlineSnapshot('"runtime error: A Google Cloud region must be specified through the GOOGLE_REGION environment variable."');
+  expect(sanitizeOutput(result.stderr)).toMatchInlineSnapshot('"runtime error: Class extends value undefined is not a constructor or null"');
```
Chriscbr, line 33: Uh oh

Chriscbr (commented):
Looks great so far

tools/hangar/src/unsupported.test.ts
hasanaburayyan: Ah yea this is probably because a test is not using the createApp helpers.

sooooo many snapshots to dig through 😂

libs/wingsdk/test/target-sim/api.test.ts
hasanaburayyan: Yupp, this needs to use the helper methods 😅 good call

PR #5472, opened 2024-01-18
feat: generate `.d.ts` and support inflights in typescript

MarkMcCulloh:
Example usages of the APIs:

```ts
import { inflight, lift, main, std } from "ts4w";
import * as winglib from "@winglibs/testfixture";

main((app) => {
  const store = new winglib.Store(app, "Store");

  store.onSet(
    // lifting is optional, can just have an inflight closure
    inflight(async ({}, message) => {
      console.log("onSet", message);
    })
  );

  new std.Test(
    app,
    "Test",
    lift({ store })
      // granting is optional, full permissions are given by default
      .grant({ store: ["set"] })
      .inflight(async ({ store }) => {
        await store.set("wing");
      })
  );
});

```

### Type generation limitations

- Resource classes/interfaces from wingsdk cannot be re-exported. e.g. cloud resources can't be extended in public types. Luckily none of the winglibs I checked do this, but this restriction will eventually be an issue.
- Only public members of classes are listed. This is mostly related to the above point, as most private/protected fields don't work anyways. This is only really relevant for allowing users to extend classes in TS, which isn't necessary for MVP
- The inflight classes are generated alongside the preflight classes. This does not break anything but it's very distracting and misleading, especially since you can't actually instantiate inflight classes yourself right now.
- Inflight classes don't actually need to be classes, they can just be interfaces. Until the concept of instantiating inflight classes is better fleshed out.

### Misc

Had to remove any typescript testing from hangar. A better test infra needs to be built out for this case. We are still running the tests in the typescript fixture, but only on linux.

### `import cloud`

This PR also adds support for consuming wingsdk `cloud` and `ex` with full lifting/inflight support

image

### Followups

- The ts4w package needs to be given a real name
- docs docs docs
- tests, mostly a fixture system to handle the gnarly chain of typescript requiring a packaged wing library
- https://github.com/winglang/wing/issues/5514
- https://github.com/winglang/wing/issues/5513


apps/wing/src/commands/pack.test.ts
```diff
@@ -197,20 +197,25 @@ describe("wing pack", () => {
...
         "$lib/.wing/preflight.subdir-4.js",
         "$lib/.wing/preflight.subdir-4.js.map",
+        "$lib/.wing/preflight.util-2.d.ts",
         "$lib/.wing/preflight.util-2.js",
         "$lib/.wing/preflight.util-2.js.map",
         "LICENSE",
         "README.md",
         "enums.w",
         "package.json",
+        "store.test.w",
```
MarkMcCulloh, line 218: packing should exclude all entrypoint files

MarkMcCulloh: Done

examples/wing-fixture/package.json
```diff
@@ -8,14 +8,23 @@
...
   },
+  "scripts": {
+    "compile": "wing compile .",
+    "test": "wing test"
+  },
   "license": "MIT",
   "engines": {
     "node": ">=v18.0.0"
   },
+  "main": "target/wing-fixture.wsim/.wing/preflight.js",
```
MarkMcCulloh, line 19: Adding this allows this library to be used as a symlinked dependency. Not a completely supported usecase, but makes testing consumption from ts much easier

apps/wing/src/commands/pack.ts
```diff
@@ -143,6 +146,11 @@ export async function pack(options: PackageOptions = {}): Promise<string> {
     // add "wing" top-level field
     pkgJson.wing = true;
 
+    if (!pkgJson.peerDependencies) {
+      pkgJson.peerDependencies = {};
+    }
+    pkgJson.peerDependencies["@winglang/sdk"] = "*";
+
```
MarkMcCulloh, line 153: Forgot I left this in, need to double check if this is necessary. I ran into issues but I was also foolishly testing with pnpm...

examples/wing-fixture/package.json
eladb: TIL

examples/wing-fixture/store.test.w
```diff
@@ -0,0 +1,8 @@
+bring "./store.w" as store;
+
+let s = new store.Store();
+s.onSet(inflight (message) => {log(message);});
+
+test "t" {
+  s.set("hi");
+}
```
eladb, line 8: assert?

examples/wing-fixture/store.w
```diff
@@ -2,15 +2,25 @@ bring cloud;
...
+    this.handlers = MutArray<inflight (str): void> [];
   }
 
   extern "./util.js" pub static makeKey(name: str): str;
   extern "./util.js" pub static inflight makeKeyInflight(name: str): str;
 
-  inflight set(message: str) {
+  pub inflight set(message: str) {
     this.data.put("data.txt", myutil.double(message));
+    for handler in this.handlers {
```
eladb, line 18: ❤️

libs/awscdk/src/on-deploy.ts
```diff
@@ -17,7 +17,7 @@ export class OnDeploy extends cloud.OnDeploy {
   ) {
     super(scope, id, handler, props);
 
-    let fn = new cloud.Function(this, "Function", handler, props);
+    let fn = new cloud.Function(this, "Function", handler as cloud.IFunctionHandler, props);
```
eladb, line 20: Is this ok?

MarkMcCulloh: Yeah, structurally the types match. typescript would yell otherwise, and we'd have to use our `convertBetweenHandlers` function. 
I'm guessing it wasn't needed before because `handler`'s type was just an empty interface that extended IInflight.

libs/ts4w/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+    process.env.WING_PLATFORMS
+  );
+  const outdir = process.env.WING_SYNTH_DIR;
+  const name = props.name ?? "main";
+  const rootConstruct = $Root;
+  const isTestEnvironment = process.env.WING_IS_TEST === "true";
+  const entrypointDir = process.env.WING_SOURCE_DIR!;
+  const rootId = process.env.WING_ROOT_ID;
+
+  const $PlatformManager = new platform.PlatformManager({ platformPaths });
```
eladb, line 63: any way to reuse this with the stuff wing generates?

libs/ts4w/src/utility_types.ts
eladb, line 1: nit: rename to `utility-types.ts`

eladb (approved):
Great job man!

libs/wingsdk/src/ex/redis.ts
```diff
@@ -14,6 +15,9 @@ export const REDIS_FQN = fqnForType("ex.Redis");
  * @abstract
  */
 export class Redis extends Resource {
+  /** @internal */
+  public [INFLIGHT_SYMBOL]?: IRedisClient;
```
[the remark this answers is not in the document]
MarkMcCulloh: Yeah it's still needed, because this field does not survive into the `.jsii`

libs/ts4w/src/inflight.ts
```diff
@@ -0,0 +1,205 @@
...
+        [K in keyof TOperations &
+          keyof TLifted]: TOperations[K] extends (infer TGrantedOps extends keyof TLifted[K])[]
+          ? PickNonFunctions<TLifted[K]> & Pick<TLifted[K], TGrantedOps>
+          : TLifted[K];
+      },
+      ...args: Parameters<TFunction>
+    ) => ReturnType<TFunction>
+  ): Inflight<TFunction> {
+    return {
+      _id: closureId(),
```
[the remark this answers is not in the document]
MarkMcCulloh: Not sure I understand. `closureId()` is already giving us reuse.
_id itself is not used again in this file

libs/ts4w/src/main.ts
MarkMcCulloh: Yeah, although I'll leave that for follow-up (will create issue)

libs/wingsdk/src/cloud/website.ts
```diff
@@ -52,6 +53,9 @@ export interface WebsiteDomainOptions {
  * @abstract
  */
 export class Website extends Resource implements IWebsite {
+  /** @internal */
+  public [INFLIGHT_SYMBOL]?: IWebsiteClient;
```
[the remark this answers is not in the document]
MarkMcCulloh: It would be nice to do something similar to the _toInflightType test, but I don't see an easy way to do it for this. The difference here is that this doesn't exist at runtime, so runtime tests can't validate the existence.

We'd have to write tests that operate on the types. Vitest does have a system for this, but we'd still have to manually enumerate types we want to check.

IMO this would be perfect for an eslint rule instead, but I'll make an issue for that as a followup

examples/wing-fixture/store.test.w
MarkMcCulloh: Added an assert to the `onSet` handler

apps/wing/src/commands/pack.ts
MarkMcCulloh: So I took this out, but I had to add the same thing to ts4w at least.

This is going to be a big issue I think: winglibs have a version of wingsdk that they expect, the cli has a version of wingsdk, and now the typescript framework also needs to be linked to a version.

For now I'll keep the problem isolated to the typescript side and say that "the version installed with the CLI is the version used", so the peer dependency is just "*" to let the winglang's version take over.

PR #5475, opened 2024-01-19
feat(sdk): implement `entries()` method inside Map

### LINKED ISSUE #5005: map.entries()
### Feature Spec

Iterate over maps using `map.entries()`:
```ts
let map = { "foo" => "hello" };

for x in map.entries() {
  log(x.key);
  log(x.value);
}
```

### Use Cases

Iterate a map.

### Implementation Notes

Implement in https://github.com/winglang/wing/blob/main/libs/wingsdk/src/std/map.ts

### Component

SDK

### Community Notes

Warkanlock:
Related to #5005

## Description

Fixes #5005, we wanted to add the `entries()` method to `Map` instances (*and `MutMap`). This essentially returns an iterator and from that, via a macro, an array of the form [{ key, value }] for each element within the `Map`.

```ts
let map = { "foo" => "hello" };

for x in map.entries() {
  log(x.key);
  log(x.value);
}
```

Now returns 

> foo
> hello


Warkanlock:
Given the limitations of JSII (https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/) I didn't find yet a possible workaround to avoid type-checking in this particular scenario. So far, it's getting impossible to use parameterized values through JSII (even if this approach works on TS)

My initial approach was to use the generic within the `keyof` type operator and then list all the possible types related within the object.

```ts
const elements = {
  a: 1,
  b: "string",
  c: null,
};

function entries<T>(obj: T): { key: keyof T; value: T[keyof T] }[] {
  return Object.keys(obj as object).map((key): { key: keyof T; value: T[keyof T] } => ({
    key: key as keyof T,
    value: obj[key as keyof T],
  }));
}

const entriesElements = entries(elements);
const firstElement = entriesElements[0];

console.log(firstElement.key); // a
console.log(firstElement.value); // 1
```

However, while effective in the wild, this approach is (_I'm not 100% sure since I'm not an expert and I don't have the deep insights of the project_) not possible in JSII. 

Essentially, these two errors are making it impossible:

```
error JSII1003: Only string-indexed map types are supported
error JSII1006: Generic types are unsupported because semantics are not uniform in target languages.
```

Any thoughts here? 

CC @hasanaburayyan (we were discussing through Slack about this specific thing)

libs/wingsdk/src/std/map.ts
```diff
@@ -6,6 +6,16 @@ import { Array } from "./array";
...
 import { InflightClient } from "../core";
 
+/**
+ * Array entry representation
+ */
+export interface ArrayEntry {
+  /** The entry key */
+  readonly key: string;
+  /** The entry value */
+  readonly value: any;
```
hasanaburayyan, line 16: Hey @Warkanlock thanks so much for the detailed description and deep diving research to try and find something that works. 

I agree with you that given the current limitations that we cannot implement this feature without the use of `any` as the value type. However, I think since there exists a less convenient work around at the moment (iterate on `keys()` and get type safe values that way) that we should hold off for the moment until we can implement this in a typesafe way.

That said I want to allow others to chime in here as well. - thoughts? @Chriscbr, @MarkMcCulloh, @eladb, @yoav-steinberg

Chriscbr: I think I tend to agree the the API feels like it's less useful if all the entries you receive are typed as `any`. I'd be in favor of documenting a workaround for use cases where you want to iterate over keys and values together.

Warkanlock: I agree with both; also, we need to keep in mind that JSII does not allow generics in this specific scenario, so I proposed the following:

1. Remove the `.key` and `.value` API and stick with the Javascript native signature. [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries)
2. This will allow us to keep the value at least as it is without the need to type it

**_However_**, playing around with the `Map` type, I realized there's no way to init a map other than the type of inception. At least I'm trying to init a `Map` from the playground (latest version available) 

E.g.:
```wing
let map = { "foo" => 1, "bar" => "test" };
                                                      ^^^ Expected type to be "num", but I got "str" instead
```

Having said that, we can "assume" that the type of the `Map` to be initialized and returned should have the same value type as the first element of the `Map`. From that we can ensure type-safety at cost of having only one type per map (not modified in the scope of this pull-request though)

Let me know if I'm losing some context in the middle since it's my first PR to the repo, and I might be confused by now

Warkanlock: More information on this issue,

Typescript, by default, does not guarantee `type safety' for each element, even on the native `Object.entries()`. [The guarantee the type from a list of types](https://github.com/microsoft/TypeScript/blob/v4.6.2/lib/lib.es2017.object.d.ts#L34-L44). 

So, technically, in case we want to keep as close to the native javascript interface as possible (and to maintain the behaviour of entries), I don't think we will be able to do that without some trade-off in the middle.

An example:
```ts
const elements = {
    "1": "Hydrogen",
    "2": "Helium",
    "3": 12,
} 
```

You will get a type as:

```ts
signature(...ArrayLike<string | number>): [string, string | number][]
```

Which is technically the behavior indicated in the snippet I submitted above. However, it's virtually impossible to implement generics using JSII (perhaps some typing magic will do the trick, but the compilation process will fail as soon as a `Generic` is introduced).

I'm glad to hear your opinions on this since I'm new to the codebase, although, at this point, I think it's worth asking a question: Is it better not to have the function at all than to have it and not have it with the guarantee of type safety?

Warkanlock: @hasanaburayyan I know that this problem is not that serious, although it would be nice to see if we will continue to find a possible solution or if we will have to make trade-offs to put it in place.

Do you have any thoughts on this?

hasanaburayyan: @Warkanlock sorry I missed the notification from your message 😅

To your question:

> Is it better not to have the function at all than to have it and not have it with the guarantee of type safety?

In my personal opinion (fwiw), its better to not allow the function. Especially since there is a work around for achieving iterating through a map. Type safety is pretty fundamental to Wing and if we allow users to break the type safety it needs to be through a "Opt-In" path. Having a standard type expose a function that breaks typesafety, sounds like it may set users up for a pretty complicated debug experience.

That said, its been a little while since I have last looked into this issue of enhancing Wing Generics, right now we achieve some basic generic functionality by using a marker class called `T1` then the compiler does a hydration process replacing these types. Though it seems we are getting close to really needing to dive into a more sophisticated solution. 

> Having said that, we can "assume" that the type of the Map to be initialized and returned should have the same value type as the first element of the Map. From that, we can ensure type safety at cost of having only one type per map (not modified in the scope of this pull request though)

So yes the observation is correct, and intentional. In fact all Container types in Wing are homogenous by design. The only type that differs from this rule is Json which can be heterogenous.

Warkanlock:
Thanks to @MarkMcCulloh's changes, this PR is now type-safe at interface level so I requested a new review

examples/tests/invalid/map_entries.test.w
```diff
@@ -0,0 +1,2 @@
+let map = { 1 => "foo", 2 => "bar" };
```
hasanaburayyan, line 1: can we add an invalid test to ensure type safety. something like:
```js
let map = { "foo": 1, "bar": 2 };

let entries = map.entries();

let someStr: str = entries.at(0).value;
```

libs/wingsdk/src/std/map.ts
```diff
@@ -6,6 +6,18 @@ import { Array } from "./array";
...
+ * Array entry representation
+ *
+ * @typeparam T1
+ */
+export interface ArrayEntry {
+  /** The entry key */
+  readonly key: string;
+  /** The entry value */
+  readonly value: T1;
+}
```
hasanaburayyan, line 19: I love that this works

libs/wingsdk/src/std/map.ts
```diff
@@ -107,6 +119,17 @@ export class Map {
   public values(): Array {
     throw new Error("Macro");
   }
+
+  /**
+   * Returns the entries from the array.
```
hasanaburayyan, line 124:
```suggestion
   * Returns the entries from the map.
```

libs/wingsdk/src/std/map.ts
```diff
@@ -247,4 +270,15 @@ export class MutMap {
...
```
hasanaburayyan, line 275:
```suggestion
   * Returns the entries from the map.
```

examples/tests/invalid/map_entries.test.w
hasanaburayyan, line 1: This will result in a lot more errors than just `^ Expected type to be "str", but got "num" instead` since maps expect keys to be strings. 

If thats not intentional then lets change it to
```suggestion
let map = { "1" => "foo", "2" => "bar" };
```

hasanaburayyan (commented):
This is soo cool that it works, I added a few comments nothing major. 
Thanks for seeing this one through!!

libs/wingsdk/src/std/map.ts
Warkanlock: It's great indeed!

examples/tests/invalid/map_entries.test.w
Warkanlock: Sure! I'll put it in place!

examples/tests/invalid/map_entries.test.w
```diff
@@ -0,0 +1,7 @@
+let map = { "1" => "foo", "2" => "bar" };
+            ^ Expected type to be "str", but got "num" instead
+
+// check typesafety
+let mapUnsafe = { "foo" => 1, "bar" => 2 };
+let someStr: str = mapUnsafe.entries().at(0).value;
+                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected type to be "str", but got "num" instead
```
MarkMcCulloh, line 7: The first line shouldn't have an error, and the error annotation needs to be in a comment

```suggestion
let map = { "1" => "foo", "2" => "bar" };

let mapUnsafe = { "foo" => 1, "bar" => 2 };
let someStr: str = mapUnsafe.entries().at(0).value;
//                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected type to be "str", but got "num" instead
```

MarkMcCulloh (approved):

examples/tests/invalid/map_entries.test.w
Warkanlock: Sorry for the first comment; I accepted a suggestion that contained that change, and I was on my mobile so I could not read it well 👯

PR #5478, opened 2024-01-20
fix(sdk): inflight clients are not cached

eladb:
Reuse the handler's inflight client across `cloud.Function` invocations, as long as they are running in the same sandbox.

This enables common FaaS optimizations such as caching database connections, etc.

To test this and improve the fidelity of our simulator, I've changed the behavior of `cloud.Function` in the simulator to reuse the function sandbox across invocations and just call `handle()` every time.

## Checklist

- [x] Tests added (always)


yoav-steinberg:
Haven't reviewed yet, but one thing to keep in mind is that cold/warm lambda invocation is an implementation detail.
To correctly simulate this we might want to randomly decide if we want to reuse the sandbox or create a new one on each invocation.

libs/wingsdk/src/cloud/function.ts
```diff
@@ -115,10 +115,13 @@ export class Function extends Resource implements IInflightHost {
   protected _getCodeLines(handler: IFunctionHandler): string[] {
     const inflightClient = handler._toInflight();
     const lines = new Array<string>();
+    const client = "$handler";
 
     lines.push('"use strict";');
+    lines.push(`var ${client} = undefined;`);
     lines.push("exports.handler = async function(event) {");
-    lines.push(`  return await (${inflightClient}).handle(event);`);
+    lines.push(`  ${client} = ${client} ?? (${inflightClient});`);
```
yoav-steinberg, line 123: This means that the the `inflight new` will only be executed once per cold start. Since cold starts aren't strongly defined, we can say the the inflight instance will be created and execute its `inflight new` at least once per `cloud.Function` instance, but possibly multiple times and always before your handler is executed. We might want to document this.

yoav-steinberg (commented):
I think this is pretty elegant. 
I'd like to better doc the behavior an perhaps consider modeling the cold invocation into the `could.Function` API.

libs/wingsdk/src/cloud/function.ts
eladb: yes, that's the behavior in AWS Lambda...

I'll add to the docs.

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -34,30 +35,43 @@ export class Function implements IFunctionClient, ISimulatorResourceInstance {
...
+  /**
+   * Creates a single instance of the function sandbox and returns it.
+   *
+   * Subsequent calls will return the same instance to mimick the behavior
+   * in the cloud where a function container may be reused across multiple
+   * invocations.
+   *
+   * @returns The simulator sandbox instance.
+   */
+  private getCreateSandbox(): Sandbox {
```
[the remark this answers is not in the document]
eladb: Yes you are right, but feels like over design at the moment. I'd say that the "surprising" behavior is that the sandbox is reused, so I think it's enough that our simulator reuses the sandbox and people design their code to work with that.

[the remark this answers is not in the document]
eladb: Added docs

docs/docs/04-standard-library/cloud/function.md
```diff
@@ -50,11 +49,45 @@ new cloud.Function(inflight () => {
...
 
-  longTask.invokeAsync();
+  longTask.invokeAsync("");
   log("task started");
 }) as "Invoke Me";
 ```
 
+## Function container reuse
+
+Most cloud providers will opportunistically reuse the function's container in additional invocations. It is possible
```
yoav-steinberg, line 59:
```suggestion
Most cloud providers will opportunistically reuse the function's container in subsequent invocations. It is possible
```

yoav-steinberg (approved):

examples/tests/valid/inflight_handler_singleton.test.w
Chriscbr, line 1: Should we also have a test to check that if there are two logically `cloud.Function` resources, then they're always going to have separate `Foo` clients?

eladb: You got it: https://github.com/winglang/wing/pull/5503

PR #5484, opened 2024-01-21
fix(compiler): multiline string

meirdev:
fixes #4375

I also encountered a problem with a string of multilines (#4375), it seems that the reason is that the compiler does not escape the newline character when it is a string that does not contain a variable in it.

not works:
```wing
log("hello 
world");
```

code:
```js
...
console.log("hello 
world");
...
```

works:
```wing
let foo = "";
log("{foo}hello 
world");
```

code:
```js
...
const foo = "";
console.log(String.raw({ raw: ["", "hello \nworld"] }, foo));
...
```

after the fix:
```js
console.log("hello\nworld");
```

PR #5486, opened 2024-01-21
chore: turbo diff check failing for transitive dependencies

MarkMcCulloh:
This script is responsible for checking what projects/tasks changed in CI. There are cases where certain transitive dependencies don't bubble up all the way. Importantly, this caused the e2e to be ignored in a PR like this https://github.com/winglang/wing/pull/5484

This change now very explicitly lists out and checks transitive deps. Honestly it's pretty naive but at the scale we're dealing with it's not a big deal.

The break may be a result of updating turbo. The list returned from the turbo cli previously always seemed to be in topological order, and it wasn't for the above PR.


tools/bump-pack/src/turbo-diff.ts
```diff
@@ -38,6 +38,18 @@ import { setOutput } from "@actions/core";
...
+  package: string;
+  taskId: string;
+  directory: string;
+  cache: {
+    status: "HIT" | "MISS";
+  };
+  dependencies: string[];
+  inputs: { [key: string]: any };
+};
+type TaskChangeMap = { [key: string]: { cached: boolean; changes: boolean } };
```
Chriscbr, line 51: Can we include a comment maybe where this data structure is coming from? It looks as though the data is coming from a foreign API of some kind

tools/bump-pack/src/turbo-diff.ts
Chriscbr, line 1: (not blocking) Any parts of the code here that you think could be easily unit tested? (unless you think this script is something we'll throw out soon...)

Chriscbr (approved):

tools/bump-pack/src/turbo-diff.ts
MarkMcCulloh: Most of it could, I'll something

> unless you think this script is something we'll throw out soon

sadly not, unless we ditch either bundled deps or pnpm

PR #5493, opened 2024-01-22
feat(sdk): bucket get range

eladcon:
It is now possible to get a range of an object through `bucket.get`

## Checklist

- [X] Tests added (always)


libs/wingsdk/src/target-sim/bucket.inflight.ts
```diff
@@ -136,14 +137,30 @@ export class Bucket implements IBucketClient, ISimulatorResourceInstance {
...
+
+          let length = stat.size;
+          if (options?.end !== undefined) {
+            length = options.end + 1;
+          }
+
+          const buffer = Buffer.alloc(length - start);
+          await file.read(buffer, 0, length - start, start);
+          await file.close();
+          return buffer.toString("utf8");
```
Chriscbr, line 163: Buffer will allow invalid UTF-8 strings to be parsed:

```js
Buffer.from([240, 160, 174, 183]).toString("utf8") // 𠮷
Buffer.from([240, 160, 174]).toString("utf8") // �
```

Would it be better to throw an error if the bytes are invalid? I believe TextDecoder would handle this for us:

```js
new TextDecoder("utf8", { fatal: true }).decode(new Uint8Array([240, 160, 174, 183])) // 𠮷
new TextDecoder("utf8", { fatal: true }).decode(new Uint8Array([240, 160, 174])) // error thrown
```

examples/tests/sdk_tests/bucket/get.test.w
Chriscbr, line 1: Let's add a test where an error should be thrown due to invalid start/end values

e.g.
```
test "get range of an object" {
  b.put("test.txt", "𠮷");

  assert(b.get("test.txt", start: 0, end: 3) == "𠮷");
  assertThrows(b.get("test.txt", start: 0, end: 2), "Bytes are not valid UTF-8");
}
```

Chriscbr (changes requested):
I didn't even know S3 supported this! TIL

It looks like the Range API is supported by Azure Blob Storage (https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob?tabs=microsoft-entra-id) and GCP (https://cloud.google.com/storage/docs/xml-api/reference-headers#range) so it is safe to include in our higher-level `cloud.Bucket` abstraction 👍

PR #5499, opened 2024-01-22
feat(sdk): bundle functions eagerly in simulator

### LINKED ISSUE #5031: Playground performance: running local tests takes too long
### I tried this:

Running a simple test on the Playground took 8 seconds.
Second run was much quicker (2 seconds).

This is the code I used:
```ts
bring cloud;

let hello_world = inflight () => {
  bucket.put("hello.txt", "Hello, World!");
};

let f = new cloud.Function(hello_world);
                                      
test "some test" {
  f.invoke("some text");
}
```

### This happened:

On the playground:
image

### I expected this:

It should be quicker.
Running `wing test` on VSCode terminal took 1 second.
image

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

0.48.10

### Node.js Version

18.16.0

### Platform(s)

MacOS

### Community Notes

Chriscbr:
Fixes #5031

Sometimes [laziness is a virtue](https://github.com/winglang/wing/pull/5453). Other times (like now), eagerness is a virtue. It just depends on the situation. 🤓 

To improve interactivity in the Wing Console and Wing Playground, this PR updates the local simulator so that it eagerly bundles all functions while starting a simulation, instead of the previous behavior of waiting until the first time a function is invoked. This means that in most cases, by the time you try interacting with any `cloud.Function`, the code will probably have finished bundling in the background, and so clicking "Invoke" will only take as much time as it takes to actually execute the inflight code.

Since bundled code can have side effects at the top-level (for example, any code that runs inside `inflight new() {}` blocks), we still only load the bundled code into the VM upon the first time a function invocation is made, to guarantee that you get FaaS cold-start-like behavior. I've updated one of our tests (inflight_handler_singleton.test.w) to demonstrate this.

If you happen to click on the "invoke" button on a function before it has finished bundling, it's possible you might still have to wait a little bit extra - but even then it should be a speedup. 🏃

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/shared/sandbox.ts
+++ b/libs/wingsdk/src/shared/sandbox.ts
@@ -14,7 +14,9 @@ export interface SandboxOptions {
 
 export class Sandbox {
   private loaded = false; // "true" after first run (module is loaded into context)
+  private createBundlePromise: Promise<void>;
   private entrypoint: string;
+  private code: string | undefined;
   private readonly options: SandboxOptions;
   private readonly context: any = {};
 
@@ -22,6 +24,7 @@ export class Sandbox {
     this.entrypoint = entrypoint;
     this.options = options;
     this.context = this.createContext();
+    this.createBundlePromise = this.createBundle();
   }
 
   private createContext() {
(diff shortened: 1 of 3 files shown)


libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -92,33 +95,45 @@ export class Sandbox {
...
 
     this.loaded = true;
   }
 
   public async call(fn: string, ...args: any[]): Promise<any> {
-    await this.loadBundleOnce();
+    // wait for the bundle to finish creation
+    await this.createBundlePromise;
+
+    // load the bundle into context on the first run
```
eladb, line 133: Very good call.

So basically you implemented lambda provisioned concurrency ;-)

eladb (approved):

PR #5522, opened 2024-01-24
fix(compiler)!: interface methods can be declared with no return type

### LINKED ISSUE #5454: interface methods can be declared with no return type
### I tried this:

From: examples/tests/invalid/function_type.test.w
```wing
// interface methods with no return type
interface IFace {
  my_method(x: num);
//^^^^^^^^^^^^^^^^^ Expected method return type
  inflight my_method2(x: num);
//         ^^^^^^^^^^^^^^^^^^ Expected method return type
}
```

### This happened:

It compiles

### I expected this:

Error: Expected method return type

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

Chriscbr:
Fixes #5454 by raising an error when interface methods are declared without return types.

In addition to requiring return types on interfaces, I also made them required on extern function statements.

The language reference is a ambiguous and doesn't explicitly say return types are optional on class methods. It's probably a good idea to require return types there as well, but I imagine this is more controversial since some might argue it's okay to omit them and let the compiler infer them from function bodies. To avoid controversy I've left the existing behavior as is.

BREAKING CHANGE: Interface methods now require return types. This was required in earlier versions of Wing (v0.18.x) but due to regressions it was not enforced, so out of caution this change has been labeled as potentially breaking.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/dtsify/mod.rs
+++ b/libs/wingc/src/dtsify/mod.rs
@@ -416,8 +416,8 @@ pub interface Interface {
 }
 
 pub interface ClassInterface {
-	addHandler(handler: inflight (str): str);
-	inflight bar();
+	addHandler(handler: inflight (str): str): void;
+	inflight bar(): void;
 }
 
 pub class ParentClass impl ClassInterface {
--- a/libs/wingc/src/parser.rs
+++ b/libs/wingc/src/parser.rs
@@ -1541,15 +1541,24 @@ impl<'s> Parser<'s> {
 	) -> DiagnosticResult<(Symbol, FunctionSignature)> {
 		let name = interface_element.child_by_field_name("name").unwrap();
 		let method_name = self.node_symbol(&name)?;
-		let func_sig = self.build_function_signature(&interface_element, phase)?;
+		let (func_sig, ret_type_inferred) = self.build_function_signature(&interface_element, phase)?;
+		if ret_type_inferred {
+			self.add_error("Expected function return type", &interface_element);
+		}
 		Ok((method_name, func_sig))
 	}
 
-	fn build_function_signature(&self, func_sig_node: &Node, phase: Phase) -> DiagnosticResult<FunctionSignature> {
+	fn build_function_signature(
+		&self,
+		func_sig_node: &Node,
+		phase: Phase,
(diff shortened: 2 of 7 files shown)


libs/wingc/src/dtsify/mod.rs
```diff
@@ -416,8 +416,8 @@ pub interface Interface {
 }
 
 pub interface ClassInterface {
-	addHandler(handler: inflight (str): str);
-	inflight bar();
+	addHandler(handler: inflight (str): str): void;
+	inflight bar(): void;
```
MarkMcCulloh, line 420: Should interfaces and classes be consistant here? Class methods allow ommiting the `void`

libs/wingc/src/parser.rs
```diff
@@ -1460,11 +1460,8 @@ impl<'s> Parser<'s> {
 			}
 			match interface_element.kind() {
 				"method_signature" => {
-					let method_name = self.node_symbol(&interface_element.child_by_field_name("name").unwrap());
-					let func_sig = self.build_function_signature(&interface_element, phase);
-					match (method_name, func_sig) {
-						(Ok(method_name), Ok(func_sig)) => methods.push((method_name, func_sig)),
-						_ => {}
+					if let Ok((method_name, func_sig)) = self.build_interface_method(interface_element, Phase::Preflight) {
```
yoav-steinberg, line 1463: I suspect a hard coded `Phase::Preflight` here is wrong. What happens if we define an interface in an inflight context. Shouldn't the methods be implicitly inflight:
```wing
let f = inflight () = {
  interface MyIf { ... }; // All methods should be implicitly inflight
  class MyClass implements MyIf { ... }; // Because MyIf can only be used inflight
  let c = new MyClass();
  c.inflight_method();
};
```
If I'm right here, maybe add a test case.

yoav-steinberg (approved):
:clap:

libs/wingc/src/parser.rs
Chriscbr: Mmm I think you're right.

I reckon it's probably OK to disallow interfaces at the top level, but I'll open a separate issue for it

PR #5528, opened 2024-01-25
chore: compatibility-spy build fails on PRs from forks

MarkMcCulloh:
The recent addition of the compatibility-spy lib broke fork workflows. This is because it was attempting to compile during the e2e matrix, but it fails because the e2e matrix is not actually meant to rebuild things. This accidentally works on PRs from winglang/wing because we use turbo remote caching, so the `turbo compile` command pulled from the remote cache.

This PR instead utilizes @winglang/compatibility-spy as a package in hangar, avoiding any re-compilation. While there I also cleaned up the tsconfig and changed it to use tsup (since it was installed anyways).

### DIFF
--- a/libs/wingsdk/.projenrc.ts
+++ b/libs/wingsdk/.projenrc.ts
@@ -413,7 +413,7 @@ new JsonFile(project, "cdktf.json", {
 });
 project.gitignore.addPatterns("src/.gen");
 
-project.preCompileTask.exec("cdktf get --force");
+project.preCompileTask.exec("cdktf get");
 
 project.package.file.addDeletionOverride("pnpm");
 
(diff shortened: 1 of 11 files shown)


libs/wingsdk/.projenrc.ts
```diff
@@ -413,7 +413,7 @@ new JsonFile(project, "cdktf.json", {
 });
 project.gitignore.addPatterns("src/.gen");
 
-project.preCompileTask.exec("cdktf get --force");
+project.preCompileTask.exec("cdktf get");
```
MarkMcCulloh, line 416: This was included by accident, but honestly I'm tired of this thing anyways

tsuf239 (approved):
Perfect! thanks a lot 🙏 🙏

PR #5541, opened 2024-01-25
feat!: native esbuild for bundling

MarkMcCulloh:
BREAKING CHANGE: core.App.for was removed. @winglang/sdk is no longer usable in a browser, and may also no longer work in certain esoteric OS/platform combos (requires native esbuild)


libs/wingsdk/src/core/app.ts
```diff
@@ -92,27 +92,6 @@ export abstract class App extends Construct implements IApp {
...
-        const cannotFindModule = e.message.split("\n")[0];
-        throw new Error(
-          `${cannotFindModule}. The target "${target}" requires this module to be installed globally (using "npm i -g").`
-        );
-      }
-
-      throw new Error(`Unknown compilation target: "${target}": ${e.message}`);
-    }
-  }
-
```
MarkMcCulloh, line 115: This is dead code and the require(`../target-${target}/app`) was causing warnings in newer esbuild

libs/wingsdk/package.json
```diff
@@ -140,7 +140,7 @@
     "@types/aws-lambda",
     "cdktf",
     "cron-parser",
-    "esbuild-wasm",
+    "esbuild",
```
MarkMcCulloh, line 143: I think bundledDeps still run postInstall scripts, let's see!

MarkMcCulloh: Ok so they do run, but don't seem to like the fact that they have already run before when initially bundled. We do a different approach

eladb (approved):
yey!

apps/wing/package.json
```diff
@@ -69,7 +69,6 @@
     "@typescript-eslint/eslint-plugin": "^6.7.4",
     "@typescript-eslint/parser": "^6.7.4",
     "bump-pack": "workspace:^",
-    "esbuild": "^0.17.19",
```
MarkMcCulloh, line 72: idk why this was here

Chriscbr:
Nice benchmark improvements

eladb:
> Nice benchmark improvements

Where can we see the improvement? Is our benchmark suite reflecting it because I am not sure I fully understand the output.

yoav-steinberg:
image
Seems pretty significant :rocket:
Would be interesting to understand why some of the benchmarks have significant improvements while others don't.

MarkMcCulloh:
> Would be interesting to understand why some of the benchmarks have significant improvements while others don't.

None of the sim compilations changed since they don't bundle anything. For tf-aws:

Only 1 benchmark (hello_world.test.w) has an inflight that interacts with a cloud resource. This means the bundling is expensive for tf-aws because of the aws SDK.

The other two benchmarks with bundling (`functions_1.test.w` and `functions_10.test.w`) don't use any resources, so no clients to bundle.

> Is our benchmark suite reflecting it because I am not sure I fully understand the output.

@eladb Any feedback about the output? Too noisy or missing details? I think we're definitely missing more cases to test, but I'm hoping I can make this more legible

image

eladb:
> @eladb Any feedback about the output? Too noisy or missing details? I think we're definitely missing more cases to test, but I'm hoping I can make this more legible

Some thoughts:

1. Let's add another test that explicitly stresses the bundling path.
2. The red block here shows a degradation of 1ms (1.48%) - seems like a very low threshold. Perhaps we can change the threshold to +/- 10% so that there will be less false alarms.
3. Perhaps it makes sense to only display the tests that crossed the thresholds and not the ones that are within the standard deviation (or put them last).

Issue #4725, opened 2023-10-30
Run `cloud.Function` outside of NodeJS vm sandbox


MarkMcCulloh:
### Feature Spec

When invoking a `cloud.Function` in the simluator, an isolated worker/process is spawned to do the work. This should work/feel similar to simply running the inflight code in a new nodejs process.

### Use Cases

Provides a more realistic experience of running JS like you might experience on serverless platforms.

Will fix bugs relating to using VM sandboxes like:
- https://github.com/winglang/wing/issues/1980
- https://github.com/winglang/wing/issues/4131
- https://github.com/winglang/wing/issues/4118

### Implementation Notes

Depends on https://github.com/winglang/wing/pull/4697

The two primary options are
- Create an entirely new child process to invoke the `cloud.Function` by just running `node` with arguments
- (ideal, imo) Use https://nodejs.org/api/worker_threads.html as a means to isolate work. While this does not give us process isolation, it should isolate everything else. We can then use SharedArrayBuffer to communicate with the simulator by default (very fast), falling back to HTTP otherwise.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

garysassano:
Maybe another option: [WinterJS](https://wasmer.io/posts/announcing-winterjs-service-workers)

garysassano:
Could this possibly address the problem I'm currently facing with `Function.invokeAsync()`? I can't test this on the simulator because by the time the function finishes executing, the `wing test` has already printed the results and exited. Given that everything runs within the single-threaded Node.js Event Loop, it's quite challenging.

MarkMcCulloh:
> Maybe another option: [WinterJS](https://wasmer.io/posts/announcing-winterjs-service-workers)

In the future I think we'd be interested in exploring different runtimes (and/or allowing easy choosing), but to start with we're focusing on maximizing nodejs compatibility.

> Could this possibly address the problem I'm currently facing with Function.invokeAsync()?

Probably not, sounds more like a Promise is dangling somewhere.

monadabot:
Congrats! :rocket: This was released in [Wing 0.59.43](https://github.com/winglang/wing/releases/tag/v0.59.43).

PR #5547, opened 2024-01-25
[WIP] feat(sdk): use workers for inflight code

Chriscbr:
Implementing one approach from https://github.com/winglang/wing/issues/4725

## Checklist

- [ ] Tests added (always)


Chriscbr:
Currently the blocker is this:

https://github.com/winglang/wing/blob/fc49999f762333b53def09b278db63ffe506b509/libs/wingsdk/src/target-sim/service.inflight.ts#L63

Our API for `cloud.Service` expects the inflight function to return another inflight function, which means a JavaScript function returning a closure. Function closures can't be serialized between worker threads (and also the API wouldn't hold water in the cloud!) so it probably needs to be reworked.

cc @eladb

Chriscbr:
My other concern with this approach is it seems to undo a bunch of our work when it comes to stack trace source mapping. For example, with this Wing code:

```js
bring cloud;
bring util;

let foo = inflight () => {
  throw "uh oh!";
};

test "inflight stuff" {
  foo();
}
```

Running `wing test` yields:

```
$ wing test main.w
fail ┌ main.wsim » root/env0/test:inflight stuff
     │ runtime error: uh oh!
     │ at handle ([worker eval]:1158:17)
     │ at $obj ([worker eval]:1153:65)
     │ at handle ([worker eval]:1134:17)
     │ at handler ([worker eval]:1186:25)
     └ at ([worker eval]:1191:20)
 
 
Tests 1 failed (1)
Test Files 1 failed (1)
Duration 0m0.71s
```

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -109,55 +45,121 @@ export class Sandbox {
...
+    console.error(shim);
+
+    // currently, a fresh worker is used for every invocation
+    // it could be better to keep the worker alive and reuse it
+    // but this requires additional work to make sure logs between invocations
+    // are not mixed up, and timeouts are handled correctly
+    const worker = new Worker(shim, {
+      env: this.options.env,
+      eval: true,
+    });
```
MarkMcCulloh, line 90: To fix sourcemaps, a more typical way to do this is to use a preload script instead of an eval'd shim:

```ts
const worker = new Worker(this.entrypoint, {
  env: this.options.env,
  // create a local file called sandbox-shim.ts with all the extra stuff needed
  // This file will be executed first, then the entrypoint
  execArgv: ["-r", require.resolve("./sandbox-shim")],
});
```

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -109,55 +45,121 @@ export class Sandbox {
...
+console.info = (...args) => $parent.postMessage({ type: 'info', args });
+console.warn = (...args) => $parent.postMessage({ type: 'warn', args });
+console.error = (...args) => $parent.postMessage({ type: 'error', args });
+process.exit = (exitCode) => $parent.postMessage({ type: 'exit', exitCode });
+Object.defineProperty(globalThis, "__dirname", {
+  get: () => { throw new Error("__dirname cannot be used within bundled cloud functions"); },
+});
+Object.defineProperty(globalThis, "__filename", {
+  get: () => { throw new Error("__filename cannot be used within bundled cloud functions"); },
+});
```
MarkMcCulloh, line 70: Doesn't esbuild already fail if we have any __dirname or __filename?

eladb: This is actually a major limitation we have today. There's a good chance that existing user code will depend on these.

Any thoughts on how to address this?

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -109,55 +45,121 @@ export class Sandbox {
...
+  }
+});
+      `;
+    console.error(shim);
+
+    // currently, a fresh worker is used for every invocation
+    // it could be better to keep the worker alive and reuse it
+    // but this requires additional work to make sure logs between invocations
+    // are not mixed up, and timeouts are handled correctly
+    const worker = new Worker(shim, {
```
eladb, line 87: This would be a regression. I think we need to retain and reuse one worker for all invocations in order to simulate lambda host reuse (https://github.com/winglang/wing/pull/5478).

eladb (changes requested):
Intuitively I still prefer to spin up a child process or a container per cloud.function and interact with it across the process boundary. Not sure what the benefit of workers are here, eif we use the same container on the cloud.

libs/wingsdk/src/shared/sandbox.ts
Chriscbr: I'm not sure whether this is an optimization that holds water. See #5549

Chriscbr:
re: Workers vs separate node process -- I did a naive benchmark (scripts below) and it seemed at least superficially like their performance was roughly the same. So I can't think of any strong reasons against it off the top of my head

<details>

```js
// bench-workers.js
const { Worker } = require('worker_threads');

const numThreads = 10; // Number of worker threads to spawn
let threadsStarted = 0;
const start = process.hrtime.bigint();

for (let i = 0; i < numThreads; i++) {
    const worker = new Worker(`
        const { parentPort } = require('worker_threads');
        parentPort.postMessage('Worker thread started');
    `, { eval: true });

    worker.on('message', () => {
        threadsStarted++;
        if (threadsStarted === numThreads) {
            const end = process.hrtime.bigint();
            console.log(`All ${numThreads} worker threads started in ${(end - start) / BigInt(1e6)} ms`);
        }
    });
}

// bench-process.js
const { fork } = require('child_process');

const numProcesses = 10; // Number of child processes to spawn
let processesStarted = 0;
const start = process.hrtime.bigint();

for (let i = 0; i < numProcesses; i++) {
    const child = fork('./childProcess.js');

    child.on('message', (msg) => {
        if (msg === 'Child process started') {
            processesStarted++;
            if (processesStarted === numProcesses) {
                const end = process.hrtime.bigint();
                console.log(`All ${numProcesses} child processes started in ${(end - start) / BigInt(1e6)} ms`);
            }
        }
    });
}

// childProcess.js
process.send('Child process started');
```

</details>

libs/wingsdk/src/shared/sandbox.ts
Chriscbr: @MarkMcCulloh From what I can tell esbuild doesn't reject code with __dirname, it just silently leaves it (leading to possibly wrong runtime behavior).

It looks like it could be possible to support it through an esbuild plugin? https://github.com/evanw/esbuild/issues/859

Chriscbr:
Closing in favor of #5554

PR #5550, opened 2024-01-26
feat: improved sourcemaps with relative paths and better relative handling for extern

### LINKED ISSUE #5457: can't use winglibs from subdirectory
### I tried this:
using the `fifoqueue` winglib which has `extern` functions 

```wing
// ./dir/main.w
bring fifoqueue;

let q = new fifoqueue.FifoQueue();

test "" {
  q.push("x", groupId: "y");
}
```

```js
// ./package.json
{
  ...
  "dependencies": {
    "@winglibs/fifoqueue": "^0.0.2"
  }
}
```

### This happened:

```
/private/tmp/a3/node_modules/@winglibs/fifoqueue/aws.js must be a sub directory of /private/tmp/a3/dir
   --> node_modules/@winglibs/fifoqueue/fifo-queue.aws.w:78:3
   |
78 |   extern "./aws.js" static inflight _push(queueUrl: str, message: str, groupId: str); 
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ /private/tmp/a3/node_modules/@winglibs/fifoqueue/aws.js must be a sub directory of /private/tmp/a3/dir
```

### I expected this:

_No response_

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

MarkMcCulloh:
### Changes
- fixes #5457
  - Partially fixed, at least
  - The error was added as a dumb limitation of the relative path calculation, which is now fixed
  - However, this error was also used to prevent the creation of wing libraries that reference extern paths outside the library itself. This limitation is unnecessary for entrypoint files, so those cases now work
- Sourcemap source paths are now relative
  - They were previously emitted as absolute for reasons I don't remember, but it seems I got it working this time
  - This should also fix sourcemaps for typescript
- when bundling, avoid encoding the bundle output and just write the raw bytes to disk
- Omit wingsdk and node internals from pretty stack traces
- Removed parens from stack trace, not convinced it is useful
- Properly handle sourcemaps for inflight clients in the sim. Previously these were useless as the context was lost between the client/server

### Before

image

### After

image


examples/tests/invalid/lib/extern_above.w
```diff
@@ -0,0 +1,4 @@
+class Foo1 {
+  extern "../../valid/external_js.js" static getGreeting(name: str): str;
+       //^ must be a sub directory of the entrypoint
+}
```
MarkMcCulloh, line 4: TODO for this PR:
This tests the case for directories, need to add a successful test case for an entrypoint (like the linked issue)

libs/wingsdk/src/util/enhanced-error.ts
```diff
@@ -65,15 +66,27 @@ export async function prettyPrintError(
...
+  st = await st
+    .clean()
+    .filter((item) => !item.native)
+    // strip node internals
+    .filter((item) => !item.file.startsWith("node:"))
+    // strip wingsdk
+    .filter((item) => !normalPath(item.file).includes("/libs/wingsdk/src/"))
+    // special: remove the handler wrapper (See `cloud.Function` entrypoint for where this comes from)
+    .filter((item) => !normalPath(item.file).match(/\.wing\/handler_\w+\.js$/))
+    .withSourcesAsync();
```
Chriscbr, line 78: I wonder if there's any way to add some test cases for these... maybe we can add some kind of `stack_traces.test.w` to the error or invalid directory with examples of different kinds of stack traces (for cases where the root error comes from a winglib, or from Wing SDK, or from CDKTF, or from constructs, or from `assert` or `bring expect;`, or from the AWS SDK, etc.). Just random ideas :-)

Chriscbr (approved):

libs/wingsdk/test/target-sim/__snapshots__/topic.test.ts.snap
```diff
@@ -157,7 +157,6 @@ exports[`topic publishes messages to multiple subscribers 1`] = `
   "Publish (message=Alpha).",
   "Sending message (message=Alpha, subscriber=sim-0).",
   "InvokeAsync (payload=\\"Alpha\\").",
-  "Received Alpha",
```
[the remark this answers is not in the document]
MarkMcCulloh: changed the test to actually assert for the test case rather than just snapshot

Issue #4781, opened 2023-11-02
`sim.LocalService`


eladb:
### Feature Spec

A new helper for implementing local services in the simulator:

```js
bring util;
bring cloud;
bring sim;

class LocalService {
  protected state: sim.State;

  init() {
    if util.env("WING_TARGET") != "sim" {
      throw "${std.Node.of(this).path} is only available in the simulator";
    }

    this.state = new sim.State();
    new cloud.Service(inflight () => { 
      this.onStart(); 
      return () => { this.onStop(); };
    });
  }

  protected inflight onStart() {}
  protected inflight onStop() {}
}
```

### Use Cases

Usage:

```js
class Bucket extends LocalService {

  pub inflight put(key: str, body: str) {
    fs.writeFile(this.pathForKey(key), body);
  }

  pub inflight get(key: str): str {
    return fs.readFile(this.pathForKey(key));
  }

  protected inflight onStart() {
    this.state.set("dir", fs.mkdtemp());
  }

  protected inflight onStop() {
    fs.remove(this.pathForKey(), recursive: true);
  }

  inflight pathForKey(key: str?): str {
    return "${this.state.get("dir").asStr()}/${key ?? ""}";
  }
}

let b = new Bucket();

test "put and get" {
  b.put("foo", "bar");
  assert(b.get("foo") == "bar");
}
```

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Chriscbr:
See also discussion and API ideas from https://github.com/winglang/wing/issues/5305

eladb:
Closing as a duplicate of #5305

PR #5554, opened 2024-01-26
fix(compiler): equality and external libraries bugs in simulator

### LINKED ISSUE #1980: cloud.Function timeout not working in simulator - vm.runInContext doesn't seem to kill the function

### LINKED ISSUE #4118: Assertion failed for code that works once compiled to JS

### LINKED ISSUE #4131: extern functions behave differently in Wing simulator
### I tried this:

I made a Wing project with these two files, and ran `wing test`:
```js
// main.w
class Util {
  extern "./superjson.js" static inflight dostuff();
}

test "superjson" {
  Util.dostuff();
}

// superjson.js
const superjson = require("superjson");

exports.dostuff = function() {
    const string = superjson.stringify(
        { handle: 'sim-3', method: 'dec', args: [ undefined, 'my-key' ] }
    );
    console.log(string);
}
```

### This happened:

This is printed by the test:
```
{"json":{"handle":"sim-3","method":"dec","args":[null,"my-key"]},"meta":{"referentialEqualities":{}}}
```

### I expected this:

I expected the output to match the output when I run `superjson` in a Node REPL:

```
Welcome to Node.js v18.14.1.
Type ".help" for more information.
> require("superjson").stringify({ handle: 'sim-3', method: 'dec', args: [ undefined, 'my-key' ] })
'{"json":{"handle":"sim-3","method":"dec","args":[null,"my-key"]},"meta":{"values":{"args.0":["undefined"]}}}'
```

Notice that the "meta" value contains an annotation saying that args.0 needs to be parsed as "undefined" instead of null. This annotation is not added when running in Wing simulator.

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

0.29.11

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

### LINKED ISSUE #4725: Run `cloud.Function` outside of NodeJS vm sandbox

### LINKED ISSUE #4792: `expect.equal(json1, json2)` fails with `Values have same structure but are not reference-equal`
### I tried this:

The `expect.equal()` method is failing because of https://github.com/winglang/wing/issues/4118

### This happened:

Something like this:

```
     │ AssertionError [ERR_ASSERTION]: Values have same structure but are not reference-equal:
     │ 
     │ {
     │   producer_result: {
     │     result: 12
     │   }
     │ }
     │ 
     │     at Util.equal (/var/folders/vm/cm444l3124j_vh_8rkhz6kqm0000gn/T/wing-bundles-XhI7xo/index.js:5449:26)
     │     at $Closure1.handle (/var/folders/vm/cm444l3124j_vh_8rkhz6kqm0000gn/T/wing-bundles-XhI7xo/index.js:34:38)
     └     at async exports.handler (/var/folders/vm/cm444l3124j_vh_8rkhz6kqm0000gn/T/wing-bundles-XhI7xo/index.js:5685:10)
 
```

### I expected this:

To work

### Is there a workaround?

stringify the objects

### Component

SDK

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

@skorfmann until we fix the VM issue that's the root cause, let's just change the expect.equal library to use `JSON.stringify()` in case it compares to objects, ok?

### Community Notes

Chriscbr:
This pull request fixes a range of funky JavaScript issues we've had running inflight code in the Wing simulator by switching internally from running JS code in a `vm` to running JS code using Node.js child processes.

The main issues with `vm` are:
- several third-party JavaScript dependencies behave differently when executed inside a `vm` and outside. This may be related to differences in vm behavior as described in https://github.com/nodejs/node/issues/28823
- we cannot easily stop/kill code executing in a `vm` process, making it impossible to simulate cloud.Function timeouts

Fixes #1980
Fixes #4131
Fixes #4118
Fixes #4792

Closes #4725

## Checklist

- [x] Tests added (always)


MarkMcCulloh (commented):
Very exciting! This PR supercedes https://github.com/winglang/wing/pull/5547 right?

btw, on the note of worker vs child process: One of the big optimization options with workers is that you can share memory between them and the host, so we can use SharedArrayBuffer as a way to do fast (and still serialized) communication.

Might be an interesting future optimization, but cp makes sense now too

libs/wingsdk/src/shared/stream-processor.ts
```diff
@@ -0,0 +1,41 @@
+import { Readable } from "stream";
+import { StringDecoder } from "string_decoder";
+
+export function processStream(
```
Chriscbr, line 4: Docstring

libs/wingsdk/src/target-sim/service.inflight.ts
```diff
@@ -5,23 +5,23 @@ import {
...
+  private readonly sandbox: LegacySandbox;
   private running: boolean = false;
   private onStop?: IServiceStopHandlerClient;
 
   constructor(props: ServiceSchema["props"], context: ISimulatorContext) {
     this.context = context;
     this.entrypoint = resolve(context.simdir, props.sourceCodeFile);
     this.autoStart = props.autoStart;
-    this.sandbox = new Sandbox(this.entrypoint, {
+    this.sandbox = new LegacySandbox(this.entrypoint, {
```
Chriscbr, line 24: Open an issue about this (cloud.Service can't use the updated Sandbox because the API tries sending a function closure across HTTP)

Chriscbr:
@MarkMcCulloh yep, I think this can supersede the other PR assuming the approach makes sense to y'all.

> btw, on the note of worker vs child process: One of the big optimization options with workers is that you can share memory between them and the host, so we can use SharedArrayBuffer as a way to do fast (and still serialized) communication.

Interesting, yeah I see how that could give us some efficiency gains. It would be interesting to do some profiling on wing programs to see how much time is spent computing vs on IPC/HTTP etc.

libs/wingsdk/src/target-sim/service.inflight.ts
eladb: Any thoughts on how to address this?

examples/tests/valid/inflight_handler_singleton.test.w
```diff
@@ -36,11 +40,21 @@ test "single instance of Foo" {
...
   expect.equal(z, "100-fn2"); // fn2 should have a separate instance
 
-  // the simulator intentionally reuses the sandbox across invocations
-  // but we can't trust that this will always happen on the cloud
-  if sim {
-    expect.equal(y, "101");
-    expect.equal(z, "100-fn2"); // fn2 should have a separate instance
-    log("client has been reused");
-  }
+  // y could be 100 or 101 depending on whether the execution environment
```
eladb, line 43: 👍

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -12,8 +14,138 @@ export interface SandboxOptions {
...
+
+  constructor(entrypoint: string, options: SandboxOptions = {}) {
+    this.entrypoint = entrypoint;
+    this.options = options;
+    this.createBundlePromise = this.createBundle();
+  }
+
+  public async cleanup() {
+    for (const timeout of this.timeouts) {
+      clearTimeout(timeout);
```
eladb, line 46: what happens if there's a `setInterval` somewhere?

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -12,8 +14,138 @@ export interface SandboxOptions {
...
+
+  private async createBundle() {
+    const workdir = await mkdtemp(path.join(tmpdir(), "wing-bundles-"));
+
+    // wrap contents with a shim that handles the communication with the parent process
+    // we insert this shim before bundling to ensure source maps will work correctly
+    let contents = (await readFile(this.entrypoint)).toString();
+
+    // log a warning if contents includes __dirname or __filename
+    if (contents.includes("__dirname") || contents.includes("__filename")) {
```
eladb, line 58: this will only catch refs to `__dirname` and `__filename` at the root module, no required modules.

libs/wingsdk/src/target-sim/service.inflight.ts
Chriscbr: Probably https://github.com/winglang/wing/issues/4781

examples/tests/valid/inflight_handler_singleton.test.w
```diff
@@ -36,11 +40,21 @@ test "single instance of Foo" {
...
+  let n = foo.inc();
+  util.sleep(3s);
+  assert(n == foo.get());
+}) as "incAndCheck";
+
+test "Foo state is not shared between concurrent function invocations" {
+  // start two invocations of fn, staggering them by 1 second
+  incAndCheck.invokeAsync("");
+  util.sleep(1s);
+  incAndCheck.invoke("");
```
eladb, line 59: nit:

```suggestion
  incAndCheck.invokeAsync();
  util.sleep(1s);
  incAndCheck.invoke();
```

eladb (approved):
Looking forward for this one!

libs/wingsdk/src/shared/sandbox.ts
Chriscbr: explain?

libs/wingsdk/src/shared/sandbox.ts
Chriscbr: I'm aware (but don't know of any other solutions)

tools/hangar/__snapshots__/error.ts.snap
Chriscbr, line 1: Fixing...

libs/wingsdk/src/shared/bundling.ts
```diff
@@ -72,13 +72,13 @@ export function createBundle(
...
-    sourcemapData.sourceRoot = normalPath(sourcemapData.sourceRoot);
+    sourcemapData.sourceRoot = normalPath(resolve(sourcemapData.sourceRoot));
   }
 
   for (const [idx, source] of Object.entries(
     sourcemapData.sources as string[]
   )) {
-    sourcemapData.sources[idx] = normalPath(source);
+    sourcemapData.sources[idx] = normalPath(resolve(source));
   }
```
Chriscbr, line 82: Without this change, source paths were getting messed up on my system, causing the files to not be found - I was getting stack traces like this:

```
fail ┌ inflight_stacktraces.test.wsim » root/env0/test:assert
     │ child process.argv: [
     │   '/Users/chrisr/.volta/tools/image/node/20.11.0/bin/node',
     │   '/var/folders/q3/0yxxhjr95vx4yfkk3lvy6dj40000gn/T/wing-bundles-gefQdo/index.js'
     │ ]
     │ Error: assertion failed: false
     │ at /private/Users/chrisr/dev/wing1/examples/tests/error/inflight_stacktraces.test.w:7:3
     │ at handler /private/Users/chrisr/dev/wing1/examples/tests/error/target/test/inflight_stacktraces.test.wsim/.wing/handler_c8b175d2.sandbox.js:19:25
     └ at /private/Users/chrisr/dev/wing1/examples/tests/error/target/test/inflight_stacktraces.test.wsim/.wing/handler_c8b175d2.sandbox.js:24:19
```

I might be doing something wrong though

Chriscbr: Hmm, this only fixes it when I run via the compiled Wing CLI and not when it's run by the hangar.

tools/hangar/__snapshots__/error.ts.snap
```diff
@@ -26,7 +26,9 @@ exports[`inflight_stacktraces.test.w 1`] = `
      │   | test \\"assert\\" {
      │ 7 |   assert(false);
      │   |   ^
-     └ at <ABSOLUTE>/inflight_stacktraces.test.w:7:3
+     │ at <ABSOLUTE>/inflight_stacktraces.test.w:7:3
+     │ at handler <ABSOLUTE>/handler_c8b175d2.sandbox.js:18:25
+     └ at <ABSOLUTE>/handler_c8b175d2.sandbox.js:23:19
```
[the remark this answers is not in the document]
Chriscbr: Ah, yeah I think that's a good callout, I don't see a reason why we need to show these lines (at least for the default output)

Issue #5414, opened 2024-01-03
Throwing does not bypass return typecheck


MarkMcCulloh:
### I tried this:

https://www.winglang.io/play/?code=bABlAHQAIAB4ADoAIAAoACkAOgAgAG4AdQBtACAAPQAgACgAKQAgAD0APgAgAHsACgAgACAAdABoAHIAbwB3ACAAIgBuAG8AdAAgAGkAbQBwAGwAZQBtAGUAbgB0AGUAZAAiADsACgB9ADsA

```wing
let x: (): num = () => {
  throw "not implemented";
};
```

### This happened:

Compiler error, A function whose return type is "num" must return a value.

### I expected this:

The compiler should detect that that function never returns

### Is there a workaround?

Add a fake return after the `throw`

### Anything else?

Proposal for fix: The type checker should detect a `throw` statement at the top-level of a function. If detected, the function should not have this error regardless of the existence of a return statement anywhere in the functino. This correctly covers the common case without adding false negatives.

```wing
let x: (): num = () => {
  if true {
    return 0;
    // Without data flow analysis, the compiler does not know for sure this will happen
  }

  throw "not implemented";
  // There are only two possible cases here: Either the function will return or it will throw.
  // Either way, `A function whose return type is "num" must return a value` is satisfied
};
```

```wing
let x: (): num = () => {
  if true {
    return 0;
  }

  if true {
    throw "not implemented";
    // This case should still not work without DFA
  }
};
```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
Sounds like a reasonable fix 👍 

For anyone that wants to pick it up, here's where the current logic is: https://github.com/winglang/wing/blob/6437741d19632ae502fc335441c895cd94729db3/libs/wingc/src/type_check.rs#L3198-L3210

Warkanlock:
I'll try to tackle this issue this weekend!

staycoolcall911:
Completed in #5555

PR #5555, opened 2024-01-27
feat(compiler): bypass throw at parent level and disregard return type

Warkanlock:
Based on #5414 

## Description

Added a bypass check at the return type level when a throw is detected in the init statement. In addition, I implemented a more generic visitor iterator that allows us to visit and collect not only the return type of the throw type and potentially more statements. 

Before:
```
error: A function whose return type is "num" must return a value.
  --> wing/main.w:1:24
  |  
1 |   let x: (): num = () => {
  | /------------------------^
2 | |   throw "not implemented";
3 | | };
  | \-^ A function whose return type is "num" must return a value
```

Now this code executed well and it has also tests to support this claim


examples/tests/valid/bypass_return.test.w
```diff
@@ -0,0 +1,28 @@
...
+test "return when a number is found even if a throw exists" {
+    let x: (): num = () => {
+
+        if true {
+            return 1;
+        }
+
+        throw "not implemented";
+    };
+}
```
MarkMcCulloh, line 28: These tests are at the type-checker level, so no need to use tests.
Also, the last test isn't really different from the second one since we don't do any sort of data flow analysis.

```suggestion
let simpleThrow: (): num = () => {
  throw "not implemented";
};

let throwWithConditional: (): num = () => {
  // this should not be executed
  if false {
    return 1;
  }

  throw "not implemented";
};
```

examples/tests/invalid/bypass_return.test.w
MarkMcCulloh, line 1: I think some additional cases could be added here. One I can think if so to verify that we still ensure if you (optionally) return the correct type even if a throw happens at the top level


```wing
let wrongReturnInScope = (): num => {
  if false {
    return "a";
         //^^^ Expected type to be "num", but got "str" instead
  }

  throw "error";
}
```

or, if the throw happens in a lower scope we still require a return

```wing
let missingReturn = (): num => {
  if true {
    throw "error";
  }
}
//^ Expected to return "num"
```

MarkMcCulloh (changes requested):
Haven't had a chance to look at the implementation yet, just had some notes on the tests.

btw when writing tests in wing, use 2 spaces for indentation. Sadly we don't have any autoformatting for that yet 😢

examples/tests/invalid/bypass_return.test.w
Warkanlock: Thanks, Mark, for the suggestions; I most definitely need to change my Vim config to have 2 tabs instead of 4 whenever I spend time working on wing files.

Are there any plans for the linter of the wing? We might be able to do it pretty fast if we already have the grammar in place

In regards to the testing suite, good to point out that there's no need to test inside a test for a type-checker validation (I thought it was a good practice inside Wing to use it always, my bad)

MarkMcCulloh: > Are there any plans for the linter of the wing?

Yeah absolutely! The goal is for the wing toolchain to be very batteries-included, so linter/formatter would be included (eventually)

> I thought it was a good practice inside Wing to use it always, my bad

In normal wing code that's definitely true, but this corpus of wing files is a little special since it's specifically written as test cases for the compiler

examples/tests/invalid/bypass_return.test.w
```diff
@@ -0,0 +1,20 @@
...
+
+  throw "error";
+}
+
+let missingReturn = (): num => {
+  if true {
+    throw "error";
+  }
+}
+//^ Expected to return "num"
```
Chriscbr, line 20: Can we add a test case where the function body is empty? I believe it should still raise an error in this scenario

libs/wingc/src/type_check.rs
```diff
@@ -3317,11 +3317,19 @@ impl<'a> TypeChecker<'a> {
 				self.update_known_inferences(&mut return_type, &scope.span);
 			}
 
-			let mut has_return_stmt_visitor = HasReturnStatementVisitor::default();
```
Chriscbr, line 3320: I think the code for `HasReturnStatementVisitor` can be removed

Chriscbr (commented):
Looks promising!

libs/wingc/src/type_check.rs
Warkanlock: Sure! I'll do, I didn't want to mess with old code but if I have this confirmation I'll proceed

examples/tests/invalid/bypass_return.test.w
Warkanlock: Excellent; let me know if there's anything I can help with in the linter!

MarkMcCulloh, line 1: In CI this file is compiling successfully when it should have errors

Chriscbr (approved):

staycoolcall911:
Hey @MarkMcCulloh - do you still want to give it another review before this goes in? Your changes-requested are currently blocking this from merging, so just want to understand if there is still anything missing in this PR. Thanks!

MarkMcCulloh (approved):

Issue #4923, opened 2023-11-13
Bang (!) operator


Chriscbr:
### Feature Spec

The bang operator can be used to tersely unwrap an optional value, throwing an error if it is nil. This is notably _different_ from TypeScript's bang operator (`!`), which silently allows code to continue running if is undefined or null.

For example:

```js
struct Person {
  name: str;
  age: num;
}

let person = Person.tryParseJson(req.body)!;
log("${person.name} is ${person.age} years old.");
```

In the example above, `Person.tryParseJson` returns `Person?` because parsing the string as Json and converting the Json to Person may fail. If it fails, the "!" indicates that the code will automatically throw an error.

`foo!` is syntax sugar for `foo ?? throws "Unexpected nil"` (tracking issue: https://github.com/winglang/wing/issues/3858).

### Use Cases

See example above

### Component

Language Design, Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

eladb:
I think the example should use `tryParseJson()` no?

Chriscbr:
Fixed - thanks

meirdev:
I think I can implement it, is it ok to send PR?

eladb:
This is implemented, no?

Chriscbr:
I believe so - closing

PR #5556, opened 2024-01-27
feat(compiler): bang operator

meirdev:
I added the bang operator as described in #4923.

`throw` is a statement, so `x!` cannot be compiled to: `x ?? throw 'nil'`, It is also not enough to use block: `x ?? { throw 'nil' }` (syntax error). The only way is to use a function: `x ?? (() => throw 'nil')()`, and that's how I implemented it.

More precisely, `x!` becoming:

```
(x??(()=>{throw new Error("Unexpected nil");})())
```


libs/wingc/src/type_check.rs
```diff
@@ -2114,6 +2114,15 @@ impl<'a> TypeChecker<'a> {
 							}
 							(self.types.bool(), phase)
 						}
+						UnaryOperator::OptionalUnwrap => {
+							if !type_.is_option() {
+								self.spanned_error(unary_exp, format!("Expected optional type, found \"{}\"", type_));
```
Chriscbr, line 2119:
```suggestion
								self.spanned_error(unary_exp, format!("'!' expects an optional type, found \"{}\"", type_));
```

libs/wingc/src/jsify.rs
```diff
@@ -709,6 +709,14 @@ impl<'a> JSifier<'a> {
 						// We use the abstract inequality operator here because we want to check for null or undefined
 						new_code!(expr_span, "((", js_exp, ") != null)")
 					}
+					UnaryOperator::OptionalUnwrap => {
+						new_code!(
+							expr_span,
+							"(",
+							js_exp,
+							"??(()=>{throw new Error(\"Unexpected nil\");})())"
```
Chriscbr, line 717: We can make the output a little bit cleaner to read by adding a helper function to do the unwrapping: https://github.com/winglang/wing/blob/main/libs/wingsdk/src/helpers.ts

The code generation can rely on `$helpers` global being available in every scope, so instead of producing:

```
(x??(()=>{throw new Error("Unexpected nil");})())
```

Maybe it can produce something like:

```
$helpers.unwrap(x)
```

Chriscbr (changes requested):
This looks great @meirdev! Just a few small comments

libs/wingc/src/jsify.rs
meirdev: I thought about doing this at first, then I saw many built-in functions written this way even if they are more complicated and longer (especially in `std`).

For example `string.at`:

```typescript
// @macro ((args) => { if ($args$ >= $self$.length || $args$ + $self$.length < 0) {throw new Error("index out of bounds")}; return $self$.at($args$) })($args$)
```

Can be written:

```typescript
// helpers.ts

export function stringAt(obj: string, index: number): string | never {
    if (index >= obj.length || index + obj.length < 0) {
        throw new Error("index out of bounds")
    };
    return obj.at(index)!;
}

// std/string.ts

// @macro $helpers.stringAt($self$, $args$)
```

Is there a reason why they are not written that way?

Currently, the compiled code can get bloated if you write even simple things:

```wing
let s = "abcdef";

s.at(0);
s.at(1);
s.at(2);
s.at(3);
s.at(4);
s.at(5);
```

Complied to:

```javascript
const s = "abcdef";
((args) => { if (0 >= s.length || 0 + s.length < 0) {throw new Error("index out of bounds")}; return s.at(0) })(0);
((args) => { if (1 >= s.length || 1 + s.length < 0) {throw new Error("index out of bounds")}; return s.at(1) })(1);
...
```

Compare to:

```javascript
const s = "abcdef";
$helpers.stringAt(s, 0);
$helpers.stringAt(s, 1);
...
```

115/24 chars = ~500% smaller, and only for one function.

MarkMcCulloh: > Is there a reason why they are not written that way?

I think it's just because they predate $helpers being added. Also the example you gave is a macro, none of which use $helpers yet. Considering these are basically compiler codegen, it makes sense that they should use them too.

libs/wingsdk/src/helpers.ts
```diff
@@ -45,3 +45,10 @@ export function nodeof(construct: Construct): Node {
 export function normalPath(path: string): string {
   return path.replace(/\\+/g, "/");
 }
+
+export function unwrap<T>(value: T): T | never {
+  if (value) {
+    return value;
+  }
+  throw new Error("Unexpected nil");
```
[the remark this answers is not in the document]
meirdev: Oops, good catch, thanks. 🙏  (I now understand the comment about `UnaryOperator::OptionalTest`).

PR #5557, opened 2024-01-27
feat(platforms): support platform parameters

hasanaburayyan:
This PR introduces has two main changes:
- The ability to now define platform specific parameters as a JsonSchema object.
- A change in the `tf-aws` platform that will now allow users to specify their own `vpc` when creating resources that need to be deployed behind a vpc (right now just redis)

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -99,8 +106,19 @@ export class PlatformManager {
...
+      hook(registrar);
+    });
 
-    return appCall!({ ...appProps, synthHooks, newInstanceOverrides }) as App;
+    // synth the registrar before returning the app, this way
+    // the inputs are registered before the app is synthesized
+    registrar.synth();
+    
+    const app = appCall!({ ...appProps, synthHooks, newInstanceOverrides, platformParameterRegistrar: registrar }) as App;
+    return app;
```
hasanaburayyan, line 121: Want to see if I can get this registrar construct into the App tree,  ran into issues when first playing with this, and got frustrated

apps/wing/src/cli.ts
```diff
@@ -168,7 +168,7 @@ async function main() {
     )
     .option("-r, --rootId <rootId>", "App root id")
     .option("-v, --value <value>", "Platform-specific value in the form KEY=VALUE", addValue, [])
-    .option("--values <file>", "Yaml file with Platform-specific values")
+    .option("--values <file>", "File with Platform-specific values", "wing.yaml")
```
hasanaburayyan, line 171: The default file might be better served as `platform.yaml`

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -0,0 +1,76 @@
+import { Construct } from "constructs";
+import { PlatformParameter } from "./platform-parameter";
+import { loadPlatformSpecificValues } from "./util";
+
+/**
+ * Parameter Registrar
+ *
+ * This construct is used to register and validate platform parameters.
+ */
+export class ParameterRegistrar extends Construct {
```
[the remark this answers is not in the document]
hasanaburayyan: 🤦 damn! I didnt even think about using JSON Schema to model these platform parameters. 

Honestly, I dont think it would be too difficult and I should be able to model all the same features such as `choices`, and `dependencies` using it, Ill play with it this morning.

libs/wingsdk/src/platform/platform.ts
```diff
@@ -12,6 +12,11 @@ export interface IPlatform {
    */
   readonly target: string;
 
+  /**
+   * The schema for the parameters the platform accepts
+   */
+  readonly parameters?: any;
```
hasanaburayyan, line 18: I want to type this using @types/json-schema but having jsii issues (still looking into it)

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * @internal
+   */
+  public _preSynthesize() {
+    // TODO: Rather than loop through each schema we should be able to generate a single schema using allOf,
+    //      but I couldn't get this to work, nesting all the schemas under a single allOf was causing some
+    //      undesired behaviors, I suspect it has to do with the jsonschema library we're using.
+    //      Its a little dated and only supports draft-07, so we may want to consider switching to json-schema (https://www.npmjs.com/package/json-schema)
```
hasanaburayyan, line 118: Im going to continue to play with replacing the jsonschema library we are using, it only supports up to draft-07 and if we are going to have users building platforms and defining parameter schemas, then it makes sense to support the latest json schema features (mainly I want the new dependencies refactors)

But I still want to get eyes on this PR in the meantime

hasanaburayyan:
@eladb I went down the intermediate step of defining platform parameters in Jsonschema. Let me know what you think. I still need to update the platform docs, but I figured I would wait until I get some thoughts about the design, and if it makes sense to keep things this way.

hasanaburayyan:
@marciocadev I also requested your review incase you want to see any of these changes (since you built the initial platform specific code)

marciocadev:
Man, I don't even know what to say, the use of JSON schema to create the required parameters was brilliant

Ship it

I love it ❤️❤️❤️

apps/wing/src/cli.ts
```diff
@@ -168,7 +168,7 @@ async function main() {
     )
     .option("-r, --rootId <rootId>", "App root id")
     .option("-v, --value <value>", "Platform-specific value in the form KEY=VALUE", addValue, [])
-    .option("--values <file>", "Yaml file with Platform-specific values")
+    .option("--values <file>", "File with Platform-specific values")
```
marciocadev, line 171:
```suggestion
    .option("--values <file>", "File with Platform-specific values (TOML|YAML|JSON)")
```

I think it would be good to specify the expected file types

libs/wingsdk/src/platform/util.ts
```diff
@@ -0,0 +1,101 @@
...
+ * }
+ *
+ * @returns a JSON object with all platform-specific values
+ */
+export function loadPlatformSpecificValues() {
+  const cliValues = createValuesObjectFromString(process.env.WING_VALUES ?? "");
+
+  if (
+    process.env.WING_VALUES_FILE === undefined ||
+    process.env.WING_VALUES_FILE === ""
```
marciocadev, line 73:
```suggestion
    process.env.WING_VALUES_FILE === ""
```

It's not a big deal, but I believe that WING_VALUES_FILE will never be undefined

marciocadev (changes requested):

libs/wingsdk/src/shared-aws/domain.ts
```diff
@@ -17,15 +17,48 @@ export class Domain extends cloud.Domain {
   constructor(scope: Construct, id: string, props: cloud.DomainProps) {
     super(scope, id, props);
 
-    const values = getPlatformSpecificValues(
-      this,
-      "iamCertificate||acmCertificateArn",
-      "hostedZoneId"
+    const registrar = App.of(scope).platformParameterRegistrar;
```
Chriscbr, line 20: (optional) Shorten name maybe
```suggestion
    const parameters = App.of(scope).platformParameters;
```

libs/wingsdk/src/target-tf-aws/platform.ts
```diff
@@ -8,6 +8,41 @@ export class Platform implements IPlatform {
...
+          },
+          vpc_id: {
+            type: "string",
+          },
+          private_subnet_id: {
+            type: "string",
+          },
+          public_subnet_id: {
+            type: "string",
+          },
```
Chriscbr, line 30: Is there a way to add comments or descriptions to the properties?

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -0,0 +1,163 @@
+import Ajv from "ajv";
+import { Construct } from "constructs";
+import { loadPlatformSpecificValues } from "./util";
+
+/**
+ * Parameter Registrar
+ *
+ * This class is used to register and lookup parameter values.
+ */
+export class ParameterRegistrar extends Construct {
```
Chriscbr, line 10: Does this need to be a construct? Maybe it's possible to avoid it...

If it does let's at least make hide it in the Wing console, I think there's an API like this available
```
std.Node.of(this).hidden = true;
```

tools/hangar/src/platform.test.ts
```diff
@@ -38,6 +39,134 @@ describe("Multiple platforms", () => {
...
+          speakNoEvil: {
+            type: "boolean"
+          },
+        },
+        allOf: [
+          {
+            "$comment": "If see no evil, then we need to know if we hear or speak evil",
+            if: { properties: { seeNoEvil: { const: true } } },
+            then: { required: ["hearNoEvil", "speakNoEvil"] }
+          },
```
Chriscbr, line 64: Creative test case 😄

libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -164,16 +166,54 @@ export class App extends CdktfApp {
   /**
    * Returns the VPC for this app. Will create a new VPC if one does not exist.
    */
-  public get vpc(): Vpc {
+  public get vpc(): Vpc | DataAwsVpc {
```
Chriscbr, line 169: nit: Consider returning an interface here that defines the shared functionality of `Vpc` and `DataAwsVpc` (to make the return type Wing-friendly)

libs/wingsdk/src/shared-aws/domain.ts
```diff
@@ -17,15 +17,48 @@ export class Domain extends cloud.Domain {
...
+
+    // Domain requires parameters from the user, so we need to add the parameter schemas to the registrar
+    let s = {
```
Chriscbr, line 23: nit
```suggestion
    let jsonSchema = {
```

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (this.parameterValueByPath[path] === undefined) {
+      // attempt to read the value from the raw parameters, then cache it
+      this.parameterValueByPath[path] = resolveValueFromPath(
+        this._rawParameters,
+        path
+      );
+    }
+
+    return this.parameterValueByPath[path];
+  }
```
Chriscbr, line 40: What happens if the parameter isn't found at the path (is this possible)?

libs/wingsdk/src/shared-aws/domain.ts
```diff
@@ -17,15 +17,48 @@ export class Domain extends cloud.Domain {
...
+
+    const iamCertificate = registrar.getParameterValue(
+      `${this.node.path}/iamCertificate`
+    );
+    const acmCertificateArn = registrar.getParameterValue(
+      `${this.node.path}/acmCertificateArn`
+    );
+    const hostedZoneId = registrar.getParameterValue(
+      `${this.node.path}/hostedZoneId`
     );
```
Chriscbr, line 57: I wonder if it's possible to do schema validation when the user calls `addParameterSchemaAtPath`? This could prevent bugs where the user puts a value of the wrong kind in their parameters file (like a list instead of a string), and this code and any code that depends on the values would run without errors

libs/wingsdk/src/platform/util.ts
```diff
@@ -0,0 +1,101 @@
...
+    switch (fileExtension) {
+      case ".yaml":
+      case ".yml":
+        return yaml.parse(data);
+      case ".json":
+        return JSON.parse(data);
+      case ".toml":
+        return toml.parse(data);
+      default:
+        throw new Error(`Unsupported file extension: ${fileExtension}`);
```
Chriscbr, line 97:
```suggestion
        throw new Error(`Unsupported file extension: ${fileExtension} (expected .yml, .json, or .toml)`);
```

libs/wingsdk/src/platform/util.ts
```diff
@@ -0,0 +1,101 @@
...
+ *  "foo": {
+ *    "bar": "123",
+ *    "baz": "789"
+ *  }
+ * }
+ *
+ * @returns a JSON object with all platform-specific values
+ */
+export function loadPlatformSpecificValues() {
+  const cliValues = createValuesObjectFromString(process.env.WING_VALUES ?? "");
```
Chriscbr, line 69:
```suggestion
  const cliValues = parseValuesObjectFromString(process.env.WING_VALUES ?? "");
```

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -0,0 +1,163 @@
...
+
+(hint: make sure to use --values to provide the required parameters file)
+        `
+      );
+    }
+  }
+}
+
+/** @internal */
+export function resolveValueFromPath(
```
Chriscbr, line 146: Can we add a docstring or comment for this function?

Chriscbr (approved):
really well done! I've added some minor comments, but it looks solid overall to me :-)

libs/wingsdk/src/target-tf-aws/platform.ts
hasanaburayyan: Yes good call, the standard way in Json schema is to use "$comment" Il add

libs/wingsdk/src/target-tf-aws/app.ts
hasanaburayyan: I did not initially see a shared interface between VPC and DataAwsVpc in cdktf like I would have expected in aws cdk, but let me check again, making this Wing-Friendly is a great point

libs/wingsdk/src/platform/parameter-registrar.ts
hasanaburayyan: So in this case it would return undefined, which if the parameter was an optional parameter than that should be expected, however if its required then during preSynth there would be a missing required parameter error

libs/wingsdk/src/target-tf-aws/app.ts
Chriscbr: Ah gotcha - yeah if there's no interface provided, you might have to create a new one. If this is internal I'm ok with leaving it as is

libs/wingsdk/src/platform/parameter-registrar.ts
hasanaburayyan: Im leveraging the fact that it exists as a construct to utilize preSynth call to validate all schemas against the input values. I think it makes it easier to aggregate all schemas, and then run an `allOf` validation to get a full list of missing parameters, and return that to users.

docs/docs/02-concepts/03-platforms.md
```diff
@@ -235,4 +235,48 @@ exports.Platform = class MyPlatform {
...
+        type: "string",
+      },
+    },
+    "$comment": "Here in an allOf we can define multiple conditions that must be met for the schema to be valid",
+    allOf: [
+      {
+        if: { properties: { replicateAllBuckets: { const: true }} },
+        then: { required: ["bucketsToReplicate"] },
+      }
+    ]
```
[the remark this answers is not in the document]
hasanaburayyan: Yea its pretty sick

libs/wingsdk/src/shared-aws/domain.ts
hasanaburayyan: It could, though from a dx perspective to the wing user I think that would result in them getting partial validation errors. Since right now validation happens all at once in an `allOf` block during preSynth we can return a full back to the user of all their parameter issues, rather than show them partial, have them fix then get the rest of the missing issues.

Chriscbr: Maybe eventually we can refactor `getParameterValue` to return a token of some kind

PR #5559, opened 2024-01-28
fix(compiler): lifting globals from inflight classes defined in preflight fails

### LINKED ISSUE #2730: Unable to capture preflight object from within an inflight class
### I tried this:

```js
bring cloud;

let b = new cloud.Bucket();
let myConst = "bang bang";

inflight class Foo {
  uploadToBucket(k: str, value: str) {
    b.put(k, value);
  }
}

test "inflight class captures preflight resource" {
  let f = new Foo();
  f.uploadToBucket("hello.txt", "world");
}
```

### This happened:

```
     │ Error: Missing environment variable: BUCKET_HANDLE_ed7b9545
```

### I expected this:

To work

### Is there a workaround?

_No response_

### Component

Language Design

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

yoav-steinberg:
Fixes #2730
The wing compiler always generated dummy preflight classes for inflight classes defined preflight. This means such inflight classes have a lift map and therefore already include all the lift qualification information needed when their mehtods are used inflight. The problem was that our lift visitor ignored references to such classes because it looks for **preflight** expressions to lift, but references to inflight classes aren't preflight expressions.
This PR finds such references and creates a special lift with correct qualifications using dummy singleton preflight object representing the relevant inflight class. To get this working the following changes where made:
* Each inflight class **defined in preflight** gets a unique id
* The generated JS code for each such class stores a map from these unique ids to the generated js preflight class of that inflight class.
* Each generated preflight file (wing module) has this map, and when it's imported by another module its map is merged into the importer's map.
* The root construct contains a global map from all the unique ids to all the inflight types defined preflight. To achieve this I had to make the imported js files local to the construct tree, so they are now imported in the root object's (entry point object) ctor and not globally.
* The base `Resource` class has a satatic `_singleton()` method
* The lift visitor, when encountering a reference to an inflight class instance (defined preflight) will qualify the lift in a similar manner to preflight class instances, but it'll use the `_singleton()` method to get a preflight expression representing that object.

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -445,6 +442,32 @@ pub trait ClassLike: Display {
...
+		// Having a preflight state means that different instances of the class
+		// might access different preflight data. Such instances need to be lifted
+		// into inflight individually. Non preflight class instances don't need to be
+		// lifted and can therefore be instantiated inflight.
+		if self
+			.fields(true)
+			.any(|(_, vi)| vi.kind == VariableKind::InstanceMember && vi.phase == Phase::Preflight)
+		{
+			return Phase::Preflight;
+		}
```
Chriscbr, line 457: I'm curious to understand better how come the class phase is determined this way (why it's not determined based on the phase in the parsing phase). Maybe there's an example that would help illustrate this?

Also, could this calculation be done once inside `type_check_class` instead of calculating it each time?

yoav-steinberg: I'm scrapping this behavior. See https://winglang.slack.com/archives/C048QCN2XLJ/p1706977736744229
TLDR: bad DX to have the class's phase calculated instead of relying on the `inflight` annotation (or lack thereof).

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_capture_preflight_object.test.w_compile_tf-aws.md
```diff
@@ -40,3 +546,245 @@ $APP.synth();
...
+## preflight.file1-3.cjs
+```cjs
+"use strict";
+const $stdlib = require('@winglang/sdk');
+const std = $stdlib.std;
+const $helpers = $stdlib.helpers;
+const $extern = $helpers.createExternRequire(__dirname);
+let $preflightTypesMap = {};
+const blah = require("./preflight.inner-2.cjs");
+Object.assign($preflightTypesMap, blah.$preflightTypesMap);
```
Chriscbr, line 558: Is it possible for assignments to shadow each other? It might be worth asserting against this at runtime somehow just to prevent future bugs for when other are trying to make changes to this code in the future

libs/wingc/src/jsify/snapshots/qualify_inflight_type_refrencing_preflight_instance.snap
```diff
@@ -0,0 +1,209 @@
...
+        `;
+      }
+      get _liftMap() {
+        return ({
+          "handle": [
+            [$helpers.nodeof(this).root.$preflightTypesMap[143]._singleton(this,"InflightC_singleton_143"), ["foo"]],
+            [InflightC, []],
+          ],
+          "$inflight_init": [
+            [$helpers.nodeof(this).root.$preflightTypesMap[143]._singleton(this,"InflightC_singleton_143"), []],
```
Chriscbr, line 194: It might be possible to shorten this to something like
```suggestion
            [$helpers.inflightSingleton(this,"InflightC_singleton", 143), []],
```

libs/wingc/src/type_check.rs
```diff
@@ -1814,6 +1845,9 @@ pub struct TypeChecker<'a> {
 	ctx: VisitContext,
 }
 
+/// Class counter, used to generate unique names class types
+static CLASS_COUNTER: AtomicUsize = AtomicUsize::new(0);
```
Chriscbr, line 1849: (idea) maybe start the counter at 1 since it looks like all imported JSII classes will have a UID of 0

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_capture_preflight_object.test.w_compile_tf-aws.md
```diff
@@ -40,3 +546,245 @@ $APP.synth();
...
+    `;
+  }
+  get _liftMap() {
+    return ({
+      "$inflight_init": [
+      ],
+    });
+  }
+}
+module.exports = { $preflightTypesMap, Foo };
```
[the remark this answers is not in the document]
yoav-steinberg: I'm aiming to get a single `$preflightTypesMap` with **all** the (inflight) classes defined preflight. To achieve this each local `$preflightTypesMap` aggregates the inner ``$preflightTypesMap`s it imports. This doesn't violate wing's export rules because this map isn't being used by application code. It's only being used in the lift mpas, where we need access to all types. Note that `$preflightTypesMap` is indexed by a unique id per type, so there shouldn't be any naming conflicts here. The top comment to this PR tries to explain the mechanics of this.

libs/wingc/src/jsify/snapshots/qualify_inflight_type_refrencing_preflight_instance.snap
```diff
@@ -0,0 +1,209 @@
...
+            });
+            if (client.$inflight_init) { await client.$inflight_init(); }
+            return client;
+          })())
+        `;
+      }
+      get _liftMap() {
+        return ({
+          "handle": [
+            [$helpers.nodeof(this).root.$preflightTypesMap[143]._singleton(this,"InflightC_singleton_143"), ["foo"]],
```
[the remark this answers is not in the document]
yoav-steinberg: If you call `new InflightClass()` inside an inflight static method there shouldn't be any difference. We don't allow passing static methods a entry points to inflight code, only inflight closures that are transformed into a `handle` instance method. So any `new InflightClass()` is always in the context of some instance.
Can you perhaps give me an example of what you mean?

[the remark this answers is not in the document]
Chriscbr: Ah I was looking at how there's `$helpers.nodeof(this)` used here (requires a scope) -- but the code is for `_liftMap` which is a special method so it should be fine

libs/wingc/src/type_check.rs
yoav-steinberg: :+1: for consistency I changed the JSII classes to also have a unique id

yoav-steinberg: > 👍 for consistency I changed the JSII classes to also have a unique id

Reverted that since it caused snapshot noise. I think we can avoid the class UID altogether by using the (currently optional) FQN field of classes. This seems ideal to me, but I'll need to talk to you to find how to implement this given you're the "bring" expert.

libs/wingc/src/jsify/snapshots/qualify_inflight_type_refrencing_preflight_instance.snap
yoav-steinberg: cleaned this up a bit

tools/hangar/__snapshots__/test_corpus/valid/inflight_class_capture_preflight_object.test.w_compile_tf-aws.md
yoav-steinberg: This should never happen due to the type unique ids.
:+1: Added a runtime check for this.

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2597,13 +2597,6 @@ error: Cannot create preflight class "PreflightClass" in inflight phase
    |   ^^^^^^^^^^^^^^^^^^^^
 
 
-error: Cannot qualify access to a lifted type "PreflightClass" (see https://github.com/winglang/wing/issues/76 for more details)
-   --> ../../../examples/tests/invalid/inflight_class_created_in_preflight.test.w:19:7
-   |
-19 |   new PreflightClass();
-   |       ^^^^^^^^^^^^^^
```
Chriscbr, line 2604: Nice to get rid of this duplicate error

wt
Chriscbr, line 1: Remove?

libs/wingsdk/src/helpers.ts
```diff
@@ -177,10 +187,43 @@ export function bringJs(
     Object.entries(require(moduleFile)).filter(([k, v]) => {
       // If this is the preflight types array then update the input object and skip it
       if (k === preflightTypesObjectName) {
+        // Verify no key collision (should never happen)
+        Object.keys(v as object).forEach((key) => {
+          if (key in outPreflightTypesObject)
+            throw new Error(`Key collision (${key}) in preflight types map`);
```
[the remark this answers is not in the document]
yoav-steinberg: :+1:  I'm so rusty :)

PR #5563, opened 2024-01-29
feat: use `this` for lifting in typescript instead of ctx

MarkMcCulloh:
I wanted to experiment with this, and since I got it working pretty quickly, I decided to put up a PR for discussion.

Changes the lifting/inflight from this:

```ts
import { lift, main, cloud } from "ts4w";

main((app) => {
  const bucket = new cloud.Bucket(app, "B1");
  new cloud.Function(
    app,
    "F1",
    lift({ bucket }).inflight(async ({ bucket }, event) => {
      await bucket.put("hi", "stuff");
      return event;
    })
  );
});
```

to this:

```ts
import { lift, main, cloud } from "ts4w";

main((app) => {
  const bucket = new cloud.Bucket(app, "B1");
  new cloud.Function(
    app,
    "F1",
    lift({ bucket }).inflight(async function (event) {
      await this.bucket.put("hi", "stuff");
      return event;
    })
  );
});
```

Full typing remains intact, with `this` now being typed the same way `ctx` previously was.

Pros:
- function arguments are as-is, no longer shifted by the first argument
  - I think users will appreciate the 1to1 match of function signatures
  - This is especially useful for inflights with no bindings, which previously required an empty first arg
- When using `function(){}`, it won't be possible to use `this` thinking the value is something else

Cons:
- inflights with lifts require the use of the more verbose `function(){}` syntax rather than `()=>{}` syntax
  - if you use an arrow function, `this` will not correct typing so it shouldn't be possible to use it incorrectly.
  - The distraction between these two js function syntaxes will be pretty opaque for most users

Note: Not marking this as a breaking change since ts isn't really released yet.

### DIFF
--- a/examples/ts-fixture/src/main.ts
+++ b/examples/ts-fixture/src/main.ts
@@ -8,8 +8,8 @@ main((app) => {
     "F1",
     lift({ bucket })
       .grant({ bucket: ["put"] })
-      .inflight(async ({ bucket }) => {
-        await bucket.put("hi", "stuff");
+      .inflight(async function () {
+        await this.bucket.put("hi", "stuff");
         console.log("hi from function");
         return "hi";
       })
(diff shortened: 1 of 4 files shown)


eladb:
Personally, I would have kept `ctx` for now because it's more explicit and less magical but bottom line - it's your call.

MarkMcCulloh:
Thanks for the feedback @eladb, I decided to drop this for now.
There's an interesting nugget here but holistically there's something off.

Issue #5540, opened 2024-01-25
creating an rfc for testing


tsuf239:
### Use Case

following 25/01/24 team time - we can only make wing's testing framework great if someone owns it, I'll collect the relevant issues and try my best to create a solid rfc out of them 💪 

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

ekeren:
Here are my two cents:
- `test` is for preflight code blocks, this keyword is only available in `*.test.w` files and it creates an isolated environment that can run checks
- `check` (or maybe `monitor`) is for inflight code, nested inside a `test` block or inside any `*.main.w` file

test runs in isolation (a test doesn't interfere another test) while `check` run on the same INFA, and there is no guarantee on order and parallelism .

```wing
// *.main.w
bring cloud;
bring util;
bring expect;

let b = new cloud.Bucket();

check "bucket put and get" {
  // check doesn't run in isolation, so a unique file is required
  let file = util.uuidv4(); 
  b.put(file, "1234");
  expect.equal(b.get(file), "1234");
}

check "list bucket" {
  let file = util.uuidv4(); 
  b.put(file, "1234");
  let files = b.list(); // will return a list of more then this file
  assert(files.contains(file));
}

check "this is a bad test" {
  // this check will fail in some cases, depending which check will run first
  expect.equal(b.list().length, 0);
}
```

```wing 
// *.test.w
bring cloud;
bring util;
bring expect;

test "empty bucket" {
  let b = new cloud.Bucket();
  
  check "list is empty" {
    expect.equal(b.list().length, 0);
  }
} 

test "bucket addObject" {
  let b = new cloud.Bucket();
  let file = "file";
  b.addObject(file ,"xyz");
  check "list return 1" {
    expect.equal(b.list().length, 1);
    expect.equal(b.list().at(0), file);
  }
  check "add object returns right content" {
    expect.equal(b.get(file), "xyz");
  }
} 

```

PR #5576, opened 2024-01-30
rfc: wing tests

tsuf239:
fixes: https://github.com/winglang/wing/issues/5540

## Checklist

- [ ] Tests added (always)


docs/contributing/999-rfcs/2024-01-29-tests.md
```diff
@@ -0,0 +1,300 @@
...
+
+}
+```
+
+Each test run in separation, and the environment (the collection of resources available in the file) is duplicated per test. Since this is the only way available now, the user cannot re-use resources between tests. The test are running on parallel- one on each cloud.Function and each reports separately to the test runner.
+
+#### Suggested:
+
+We'll add another scope (and have two in total- outer scope, and inner scope) for the following structure:
+//TODO: names
```
tsuf239, line 41: I'd love to have some help here :) 
@ekeren suggested `test` for outerScope and `check`/`monitor` for innerScope
I feel like it's not clear enough, but I'm not sure :)

docs/contributing/999-rfcs/2024-01-29-tests.md
```diff
@@ -0,0 +1,300 @@
+---
```
tsuf239, line 1: This is an initial version, and I haven't filled some of the fields, @winglang/maintainers feel free to commit/ comment suggestions and solutions

docs/contributing/999-rfcs/2024-01-29-tests.md
```diff
@@ -0,0 +1,300 @@
...
+
+This rfc lists a few features that improve test-writing in wing, and strengthen wing's test framework.
+
+1. [scopes](#1-scopes)
+2. [lifecycle hooks](#2-lifecycle-hooks)
+3. [integration with external libs](#3-integration-with-external-libs)
+4. [mocks](#4-mocks)
+5. [spies](#5-spies)
+6. [resource management optimization](#6-resource-management-optimization)
+
```
tsuf239, line 21: I guess 3-6 will be out of scope for this iteration... WDYT?

docs/contributing/999-rfcs/2024-01-29-tests.md
```diff
@@ -0,0 +1,298 @@
...
+```js
+// file/global scope - preflight code only. Resources initialized here will be duplicated across outerScopes.
+let counter = new cloud.Counter();
+
+outerScope "bucket tests" {
+  /// preflight code only. Resources initialized here will be shared across innerScopes.
+  let bucket = new cloud.Bucket();
+
+  innerScope "bucket starts empty" {
```
MarkMcCulloh, line 51: When this topic has been brought up, usually there's mention of a preflight/inflight test concept, where the preflight test is the setup for the inflight tests. WDYT?

```wing
// breaking change, but like other declarations the phase can be inferred (preflight)
test "bucket tests" {
  /// preflight code only. Resources initialized here will be shared across tests here

  // explicitly inflight tests within the preflight test
  inflight test "bucket starts empty" {
  }
}
```

It's possible that users will be confused that it's a test within a test, although with good test output maybe it would be very intuitive. 
If it does seem confusing, maybe we can be more explicit about it being a setup or group. Just some ideas:

```wing
suite "bucket tests" {
  /// preflight code only. Resources initialized here will be shared across tests here
  
  preflight test "bucket as prop" {
  }

  // explicitly inflight tests within the preflight test
  inflight test "bucket starts empty" {
  }
}
```

```wing
test app "bucket tests" {
  /// preflight code only. Resources initialized here will be shared across tests here
  
  preflight test "bucket as prop" {
  }

  // explicitly inflight tests within the preflight test
  inflight test "bucket starts empty" {
  }
}
```

docs/contributing/999-rfcs/2024-01-29-tests.md
MarkMcCulloh: I think check/monitor is intentionally a different concept, meant for testing live applications. This is RFC seems like it's aiming closer towards unit testing.

docs/contributing/999-rfcs/2024-01-29-tests.md
```diff
@@ -0,0 +1,298 @@
...
+## 2. Lifecycle Hooks
+
+### Design
+
+A collection of 4 functions (beforeAll, beforeEach, afterAll, afterEach) allow the user to hook into the life cycle of tests to avoid repeating setup and teardown inflight code.
+They apply to the current scope:
+_global scope:_
+
+- beforeEach, afterEach are called before or after each outerScope- before running the first innerScope, and after running the last one.
+- beforeAll, afterAll are called before starting any of the innerScopes and after running them all.
```
MarkMcCulloh, line 110: There probably shouldn't be any testing-related functions in the global scope

docs/contributing/999-rfcs/2024-01-29-tests.md
ekeren: > I think check/monitor is intentionally a different concept, meant for testing live applications. This is RFC seems like it's aiming closer towards unit testing.

It is different concept but also very very similar to an `inflight test` especially when `inflight test`s do not run in isolation

I feel that in cases where things are very similar there is value in consolidating them into a single concept.

eladb: I am not sure I agree. The fact that two concepts use similar mechanics doesn't mean they should be the same concept.

`inflight test` is also very similar to just a `cloud.Function` but they are different concepts.

There is also a consideration of design effort. I am not sure we have a reasonable way to model the idea of "an inflight test that doesn't run in isolation" (ideas welcome), some sometimes it's just easier to say "a test always runs in isolation and if you want something that doesn't run in isolation use a monitor".

It's all just words.

ekeren: Here is my original idea for combining test (a wing program that contains both preflight and inflight code) and checks.

https://github.com/winglang/wing/issues/5540#issuecomment-1910665881

tsuf239:
stale

PR #5578, opened 2024-01-30
chore: rename ts4w -> @wingcloud/framework

MarkMcCulloh:
Copy-paste of naming rationale from slack:

#### Why an npm scope (@)?
- IDK if others share this feeling, but I really appreciate when things I interact with programmatically are in a scope
- if we add integrations for wing+js/ts related things, having a scope helps with trust and discoverability (e.g. @wingcloud/sentry or @wingcloud/nextjs)
#### `@wingcloud`?
- I wish it could be `@wing`, but alas it’s taken.
- I think this is a nice sounding name even though I dislike it in other contexts. In this context I feel like it can holistically cover wing-related things
#### framework ?
- My puristic argument is that this package is a “framework” (rather than library or sdk) because it calls your code rather than your code just calling it. So just call it “framework”!
- The name is terse for 3 reasons:
  - Short things/people are cool too
  - The scope provides context that this relates to wing
  - The fact that this is an npm package tells you it’s for js/ts


eladb (approved):
Ship it!

PR #5579, opened 2024-01-30
feat: empty `new` template for typescript and `test` function for ts

MarkMcCulloh:
Few changes:
- `wing new empty --language ts` works
  - Had to add a mechanism to switch the wing version in package.json
  - Eventually I assume we'll need a more full-blown template interpolation system
- In the `main` callback for ts, a `test` function is provided to serve as the proper way to do `std.Test`
  - This better matches the current wing experience, where tests are only allowed in an entrypoint. This can (and should) be revised depending on the evolution of testing that's currently being worked on.
- Removed `expect` export from ts lib
  - It's not very good compared to the builtin "node:assert" lib, so we shouldn't bother encouraging using it from ts
- Updated some random small docs


examples/ts-fixture/src/main.ts
```diff
@@ -1,7 +1,7 @@
-import { inflight, lift, main, std, cloud } from "@wingcloud/framework";
+import { inflight, lift, main, cloud } from "@wingcloud/framework";
 import * as winglib from "@winglibs/testfixture";
 
-main((app) => {
+main((app, test) => {
```
Chriscbr, line 4: 👍

libs/@wingcloud/framework/src/internal.ts
```diff
@@ -11,8 +11,8 @@ export async function compile(options: CompileOptions) {
 
   const program = ts.createProgram([options.entrypoint], {
     target: ts.ScriptTarget.ES2022,
-    module: ts.ModuleKind.Node16,
-    moduleResolution: ts.ModuleResolutionKind.Node16,
+    module: ts.ModuleKind.CommonJS,
+    moduleResolution: ts.ModuleResolutionKind.Node10,
```
Chriscbr, line 15: Node 10?

apps/wing/src/commands/init.test.ts
```diff
@@ -34,6 +34,32 @@ describe.each(templates)("new %s --language=wing", (template) => {
...
+
+  afterEach(() => {
+    console.log = log;
+  });
+
+  test(`wing new empty --language ts && wing test main.w`, async () => {
+    const workdir = await generateTmpDir();
+    process.chdir(workdir);
+
+    console.debug("HERE");
```
Chriscbr, line 52: Looks like leftover from debugging

apps/wing/project-templates/typescript/empty/main.ts
```diff
@@ -0,0 +1,19 @@
...
+    "Function",
+    inflight(async () => {
+      return "hello, world";
+    })
+  );
+
+  test(
+    "fn returns hello",
+    lift({ fn }).inflight(async ({ fn }) => {
+      const assert = await import("node:assert");
```
Chriscbr, line 15: Is this an optimization of sorts? I wonder if it be more natural to import node:assert at the top of the file

Chriscbr (approved):

apps/wing/project-templates/typescript/empty/main.ts
MarkMcCulloh: It's a necessary evil. Nothing from the outer scope can be used, including imports.

I have an idea to do some special hoisting for symbols that come from import statements instead, but for now this is needed.

apps/wing/src/commands/init.test.ts
MarkMcCulloh: I've been caught

libs/@wingcloud/framework/src/internal.ts
MarkMcCulloh: Yeah, It's the same as `ModuleResolutionKind.Node` (which is deprecated to tell people that this is not the modern way to do it)

I think these options are the closest to how wing actually needs it to work

eladb (commented):
Sweet!

PR #5587, opened 2024-01-31
fix: possible conflicting versions of "constructs" package for ts framework

MarkMcCulloh:
Just avoid ts framework relying on its own version of constructs


libs/@wingcloud/framework/src/main.ts
```diff
@@ -1,5 +1,7 @@
 import { platform, std } from "@winglang/sdk";
-import { Construct } from "constructs";
+import type {
+  Construct
+} from "@winglang/sdk/lib/core/types";
```
Chriscbr, line 4: Does a user also need to do this if they're creating Construct classes their app code? (wondering if the `empty` template needs to be updated as well)

Chriscbr (approved):

libs/@wingcloud/framework/package.json
```diff
@@ -23,9 +23,6 @@
     "registry": "https://registry.npmjs.org",
     "tag": "latest"
   },
-  "dependencies": {
```
eladb, line 26: FYI: if we ever take a dependency on constructs we should use `10.x`.

eladb (approved):

libs/@wingcloud/framework/src/main.ts
MarkMcCulloh: Hmm good point. I need to do some experiments to see what causes the least friction.

PR #5592, opened 2024-01-31
chore: update several jsii-based dependencies

MarkMcCulloh:
I went over every snapshot update and verified that the only change was to the versions (and one hash in awscdk's tests, and a new method on construct nodes in the lsp snaps)

- constructs ^10.2 -> ^10.3
- cdktf 0.17.0 -> 0.20.3
- @cdktf/provider-aws ^15 -> ^19 (internal dependency)
- aws-cdk-lib ^2.116.0 -> ^2.124.0

The main interesting one here is cdktf, which does have breaking changes between the two versions. However, I don't think any of them are relevant for us so I didn't mark this PR itself as a breaking change.


Chriscbr (approved):

MarkMcCulloh:
Well that's unfortunate, I was expecting better performance :/

MarkMcCulloh:
Despite the performance regression, I think this is still important to merge. I noticed that it's way too easy to run into conflicts with tf providers and cdktf because we are several versions behind. I'll get this updated and merged soon

MarkMcCulloh:
AWS Testing https://github.com/winglang/wing/actions/runs/7805818329

Issue #5526, opened 2024-01-24
Cannot use factory methods in entrypoint files


Chriscbr:
### I tried this:

This works:

```js
bring cloud;

class BucketFactory {
  pub static makeBucket(scope: std.IResource): cloud.Bucket {
    return new cloud.Bucket() in scope;
  }
}

class A {
  new() {
    let bucket = BucketFactory.makeBucket(this);
  }
}

let a = new A();
```

But using `makeBucket` at the top level doesn't work:

```js
bring cloud;

class BucketFactory {
  pub static makeBucket(scope: std.IResource): cloud.Bucket {
    return new cloud.Bucket() in scope;
  }
}

let bucket = BucketFactory.makeBucket(this);
```

### This happened:

error: Unknown symbol "this"

### I expected this:

No error

### Wing Version

0.54.58

### Node.js Version

20.9.0

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

eladb:
A few thoughts:

1. I also suggested somewhere to simply allow `this` in the entrypoint. As much as it might be weird, it's "correct" from the programming model perspective.
2. We discussed somewhere to consider a global `app` which represents the root app (`nodeof(this).app`). We can discuss an alternative name.
3. There's an analogy between constructors and static (especially factory) methods. Today we implicitly pass `scope` to all preflight constructors. I am wondering if that's something we can generalize to static methods.

Chriscbr:
I think any preflight code generated that uses `this` in an entrypoint file should already work -- it's just our type checker rejecting the use of `this`.

Implementing this should be straightforward. Define `this` as a symbol in the symbol environment, with a type of `constructs.IConstruct`.

This should compile:

```js
// entrypoint-file.main.w
class A {
  static foo() {
    log("it works here" + this.node.id);
  }
}

log("it also works here " + this.node.id);
```

And this should fail to compile:

```js
// non-entrypoint-file.w
class A {
  static foo() {
    log(this.node.id);
  }
}
```

eladb:
Let's do it! ☀️

monadabot:
Congrats! :rocket: This was released in [Wing 0.56.11](https://github.com/winglang/wing/releases/tag/v0.56.11).

PR #5594, opened 2024-01-31
feat(compiler): make "this" available in entrypoint files

hasanaburayyan:
Adds "this" to symbol environment of entrypoint files. Typed as `constructs.Construct`

closes: https://github.com/winglang/wing/issues/5526

## Checklist

- [ ] Tests added (always)


apps/wing/src/commands/test/__snapshots__/test.test.ts.snap
```diff
@@ -2,7 +2,13 @@
```
hasanaburayyan, line 2: Ah shoot, did not mean to commit this

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -389,18 +389,18 @@ error: Cannot access static property \\"m\\" from instance
    |     ^ Cannot access static property \\"m\\" from instance
 
 
-error: Unknown symbol \\"this\\"
-  --> ../../../examples/tests/invalid/access_static_from_instance.test.w:7:5
+error: Unknown symbol \\"instanceField\\"
+  --> ../../../examples/tests/invalid/access_static_from_instance.test.w:7:10
```
hasanaburayyan, line 393: So technically this is still just as invalid as it was before, however the error messages are different since now "this" is not an unknown symbol.

@Chriscbr @eladb just curious about your thoughts, I kinda wonder if it would be less confusing to introduce something like "app" (mentioned in the original PR).

Chriscbr: I don't have a strong intuition for which will be better DX, so I'm open to either

hasanaburayyan: Yea, I dunno thinking about it more my gut feeling is that "this" would be the better DX over something like "app" but yea maybe we just try `this` (pun intended) and change as needed

eladb (approved):
cool

Issue #4998, opened 2023-11-18
Array.sort() method support


subh-cs:
### Feature Spec

let arr:Array<num>=[2, 1, 3, 9, 6, 4];
arr.sort();
log("${arr}");  // it should print sorted array in ascending order, eg: [1, 2, 3, 4, 6, 9]

### Use Cases

.

### Component

Language Design, Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

revitalbarletz:
Issue taken by Lauty Suarez!

Chriscbr:
Sorting is certainly missing! thanks for opening this issue @subh-cs. I noticed a small bug in the issue -- the array you gave was immutable, so it shouldn't be sortable. But of course, `sort()` makes sense as a capability for `MutArray`s. I fixed it for your convenience :-)

For implementing this issue I think it would be forward thinking to define a semantics for the sorting. Even if we're using JavaScript's default `sort` implementation under the hood, we should note the essential edge cases in the docs to ensure the method works as people expect when they're coming from non-JS languages.

For example, if an array contains a collection of structs, how are the values sorted (supposing no comparison function is provided by the user)? We should include a note in the docstring how this edge case is handled.

I think it would also be handy to document whether or not the sort implementation is [stable or unstable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) since it's a detail relevant for implementing many algorithms.

ekeren:
@Chriscbr, there are two options of sort that I can think of:

```ts
// Option 1: Mutating  - relevant only for MutableArray
arr.sort();
// arr is not sorted 

// Option 2: Non-Mutating - relevant for Array and MutArray
let sorted = arr.sort();
// arr state remains the same 
```

Did we consider option 2?

revitalbarletz:
@subh-cs Are you looking into this one?

Warkanlock:
I'm working on this one

staycoolcall911:
Sounds good, then I'll assign you

Warkanlock:
I have a few comments regarding this change: 

1. Lack of a comparator function
2. A comparator function would imply the creation of dynamic typing and, therefore, we would run into the same problems I encountered when I tried to implement #5005. 

[...]

PR #5609, opened 2024-02-02
feat(sdk): implement sort() and sorted() methods on Array

Warkanlock:
Based on #4998

## Description

Both methods are implemented at the `Array` level, the new methods are `sort()` and `sorted()`; the main difference lies in the ability of sorted() to return a new array, and on the other hand, `sort` that mutates the original array.


libs/wingsdk/src/std/array.ts
```diff
@@ -311,4 +322,15 @@ export class MutArray {
     value;
     throw new Error("Macro");
   }
+
+  /**
+   * Returns the array, sorted and also mutated
+   *
+   * @macro $self$.sort()
```
tsuf239, line 329: @Chriscbr what do you think about passing a sorting function?

Warkanlock: I thought about a comparator, although JSII does not support generics, so we will need to rely on `any` for the comparator types

Warkanlock:
@MarkMcCulloh, first implementation of the changes you introduced last week; good to double-check if there's not something I missed from the implementation of the generics

libs/wingsdk/src/std/array.ts
```diff
@@ -125,6 +145,19 @@ export class Array {
     searchElement;
     throw new Error("Macro");
   }
+
+  /**
+   * Returns the array sorted without mutating the inital
```
Chriscbr, line 150:
```suggestion
   * Returns a sorted copy of the array.
```

libs/wingsdk/src/std/array.ts
```diff
@@ -311,4 +344,17 @@ export class MutArray {
     value;
     throw new Error("Macro");
   }
+
+  /**
+   * Returns the array, sorted and also mutated
```
Chriscbr, line 349:
```suggestion
   * Sorts the array, modifying it in-place.
```

libs/wingsdk/src/std/array.ts
```diff
@@ -125,6 +145,19 @@ export class Array {
...
+
+  /**
+   * Returns the array sorted without mutating the inital
+   *
+   * @macro ((arr, args) => { return args && args[0] ? [...(arr)].sort(args[0]) : [...(arr)].sort() })($self$, $args$)
+   *
+   * @param comparator Function used to determine the order of the elements.
+   * @returns same array, sorted
+   */
+  public sorted(comparator?: IComparator): Array {
```
Chriscbr, line 157: `sorted()` should also be available on MutArray, no?

Warkanlock: It's a good proposal; I implemented just sort() and sorted() according to what it was on the original spec but we can iterate it further

eladb (commented):
Good stuff!

libs/wingsdk/src/std/array.ts
Warkanlock: Interesting; I'll put it on my todo list

examples/tests/sdk_tests/std/array.test.w
```diff
@@ -399,3 +399,65 @@ test "removeFirst()" {
...
+//-----------------------------------------------------------------------------
+// sorted() for a non-mutable array
+
+test "sorted()" {
+    let arr = Array<num>[3, 1, 2];
+    let sortedArr = arr.sorted();
+
+    assert(sortedArr == Array<num>[1, 2, 3]);
+    assert(arr == Array<num>[3, 1, 2]);
+}
```
MarkMcCulloh, line 422: nit: 2 space indent

```suggestion
//-----------------------------------------------------------------------------
// sort() for a mutable array

test "sort()" {
  let mutArr = MutArray<num>[3, 1, 2];
  mutArr.sort();

  assert(mutArr == MutArray<num>[1, 2, 3]);
}

//-----------------------------------------------------------------------------
// sorted() for a non-mutable array

test "sorted()" {
  let arr = Array<num>[3, 1, 2];
  let sortedArr = arr.sorted();

  assert(sortedArr == Array<num>[1, 2, 3]);
  assert(arr == Array<num>[3, 1, 2]);
}
```

examples/tests/sdk_tests/std/array.test.w
```diff
@@ -399,3 +399,65 @@ test "removeFirst()" {
...
+    if (a > b) {
+        return 1;
+    }
+
+    return 0;
+  });
+
+  assert(sortedArr == Array<num>[1, 2, 3]);
+  assert(arr == Array<num>[3, 2, 1]);
+}
```
MarkMcCulloh, line 463: nit: style and indent

```suggestion
test "sortedWithComparator()" {
  let arr = Array<num>[3, 2, 1];
  let sortedArr = arr.sorted((a, b) => {
    if a < b {
      return -1;
    }

    if a > b {
      return 1;
    }

    return 0;
  });

  assert(sortedArr == Array<num>[1, 2, 3]);
  assert(arr == Array<num>[3, 2, 1]);
}
```

examples/tests/sdk_tests/std/array.test.w
```diff
@@ -399,3 +399,65 @@ test "removeFirst()" {
...
+
+    if (a > b) {
+        return 1;
+    }
+
+    return 0;
+  });
+
+  assert(mutArr == MutArray<num>[1, 2, 3]);
+}
```
MarkMcCulloh, line 442: nit: style and indent

```suggestion
test "sortedMutWithComparator()" {
  let mutArr = MutArray<num>[3, 2, 1];
  mutArr.sort((a, b) => {
    if a < b {
      return -1;
    }

    if a > b {
      return 1;
    }

    return 0;
  });

  assert(mutArr == MutArray<num>[1, 2, 3]);
}
```

examples/tests/sdk_tests/std/array.test.w
```diff
@@ -399,3 +399,65 @@ test "removeFirst()" {
...
+    if (a < b) {
+        return -1;
+    }
+
+    if (a > b) {
+        return 1;
+    }
+
+    return 0;
+  });
```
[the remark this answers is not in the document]
Warkanlock: Sounds good; I'll go ahead with implementing the `sorted` method in `MutArray` as well. In addition, I was trying to use an structure to sort using a field and I got an error of **property not found**, investigating a bit on the project I saw it was a bug a few months ago for another particular use case so this might be a bug on the compiler side

First of all, the compiler throws an error regarding `Unable to infer type` when using an `struct` and on the other hand, if you try to access the property of the object, it throws: `Property not found`

E.g:

```
  let person = Person {
    age: 25
  };

  let person2 = Person {
    age: 30
  };

  let person3 = Person {
    age: 35
  };

  let mutArr = MutArray<Person>[person3, person, person2];

  // implement a sort by age in desc order
  let sortedArr = mutArr.sorted((a, b) => {
    if a.age > b.age {
      return -1;
    }

    if a.age < b.age {
      return 1;
    }

    return 0;
  });
```

Errors generated after running the piece of code paste above:
image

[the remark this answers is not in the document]
Chriscbr: Does it work if you annotate the types?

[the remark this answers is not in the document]
Warkanlock: Yeah, with explicit type annotation inside the callback works as a charm although I thought I could be implicit about the type given I'm using generic at `MutArray` level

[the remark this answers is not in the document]
MarkMcCulloh: This looks like an inference bug. Replicated with this:

```wing
let sorted = (fn: ((num): void)?) => {};

sorted((a) => {});
//      ^ can't infer this
```

Created this here and I'll look into this now https://github.com/winglang/wing/issues/5680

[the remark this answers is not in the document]
Warkanlock: Thanks, Mark; additionally, a callback with annotations does not sort in the end, so my guess here is that the type checker is passing through, although failing silently because the array itself does not get sorted after passing the following callback

```
test "sortedMutWithComparatorAndStructs()" {
  let mutArr = MutArray<Person>[Person{age: 30}, Person{age: 40}, Person{age: 20}];

  // implement a reverse sort (instead of the default ascending sort)
  let sortedArr = mutArr.sorted((internalA, internalB) => {
    let a : Person = internalA;
    let b : Person = internalB;

    if a.age > b.age {
      return -1;
    }

    if a.age < b.age {
      return 1;
    }

    return 0;
  });

  log("{Json.stringify(sortedArr)}");
}
```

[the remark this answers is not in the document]
MarkMcCulloh: Are you saying that this snippet doesn't work when annotation? `log("{Json.stringify(sortedArr)}");` shows an unsorted list? My **PR** to fix the inference is merged but I'm not sure this is a type issue.

By the way, I really appreciate your patience with all the changes needed to get all this to work. Implementing stuff like this properly requires stretching the language in important places, so we're glad you're up for it.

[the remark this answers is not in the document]
Warkanlock: Exactly! No matter what callback I pass through, the array remains unaffected. Apparently, (debugging the .wing artifacts), I can replicate the non-sorting issue (in Javascript), so something weird might be going on in the translation step from the SDK to the Javascript closure.

This is the code generated using `struct` as types (it works for `num`, `string`, etc)
```
module.exports = function({  }) {
  class $Closure24 {
    constructor({  }) {
      const $obj = (...args) => this.handle(...args);
      Object.setPrototypeOf($obj, this);
      return $obj;
    }
    async handle() {
      const mutArr = [({"age": 30}), ({"age": 40}), ({"age": 20})];
      const sortedArr = ((arr, args) => { return args && args[0] ? [...arr].sort(args[0]) : [...arr].sort() })(mutArr, (async (a, b) => {
        if ((a.age > b.age)) {
          return (-1);
        }
        if ((a.age < b.age)) {
          return 1;
        }
        return 0;
      }));
      $helpers.assert($helpers.eq(mutArr, [({"age": 30}), ({"age": 40}), ({"age": 20})]), "mutArr == MutArray<Person>[Person{age: 30}, Person{age: 40}, Person{age: 20}]");
      $helpers.assert($helpers.eq(sortedArr, [({"age": 20}), ({"age": 30}), ({"age": 40})]), "sortedArr == Array<Person>[Person{age: 20}, Person{age: 30}, Person{age: 40}]");
    }
  }
  return $Closure24;
}
```

I will continue to investigate. As for patience, I'm here to learn how salsa is made, and all this back and forth makes me incredibly happy because I get to see with my own eyes how a language like this is made. It's a really fun experience for me and my ideal is to one day be able to contribute to these things without screwing up everyone's schedule 😆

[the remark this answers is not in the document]
revitalbarletz: @Warkanlock keep on the good work!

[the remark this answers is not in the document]
Warkanlock: Found the issue; it's the way "args" it's being read inside the `macro`

I'll investigate further

libs/wingsdk/src/std/array.ts
```diff
@@ -125,6 +145,19 @@ export class Array {
     searchElement;
     throw new Error("Macro");
   }
+
+  /**
+   * Returns a sorted copy of the array.
+   *
+   * @macro ((arr, args) => { return args && args[0] ? [...(arr)].sort(args[0]) : [...(arr)].sort() })($self$, $args$)
```
MarkMcCulloh, line 152:
```suggestion
   * @macro ($self$.sort($args$ ?? undefined))
```

- $args$ are the arguments passed as-is, not necessarily an array
- We don't actually need the wrapper function here, just calling sort is fine
- I think JS's sort handles `undefined` so it can just be passed in. adding `?? undefined` is just in case args is null which I think is possible

The other macros here can be changed similarly

libs/wingsdk/src/std/array.ts
```diff
@@ -125,6 +145,19 @@ export class Array {
     searchElement;
     throw new Error("Macro");
   }
+
+  /**
+   * Returns a sorted copy of the array.
+   *
+   * @macro ($self$.sort($args$ ?? undefined))
```
MarkMcCulloh, line 152:
```suggestion
   * @macro ($self$.sort($args$))
```

I checked the compiler code and this undefined actually shouldn't be necessary (my bad!)

examples/tests/sdk_tests/std/array.test.w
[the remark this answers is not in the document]
MarkMcCulloh: The issue is here, where the callback is async

```js
      const sortedArr = ((arr, args) => { return args && args[0] ? [...arr].sort(args[0]) : [...arr].sort() })(mutArr, (async (a, b) => {
        if ((a.age > b.age)) {
          return (-1);
        }
        if ((a.age < b.age)) {
          return 1;
        }
        return 0;
      }));
```

It's async because all inflight closures are async and `.sort()` can't handle an async callback. We can't just strip the async off (technically the closure could be `await`ing even though that would be a bad thing to do in a comparator), so we need a way to sort with an async comparator.

Warkanlock:
This PR is currently blocked, given that we will need more context and developments towards having phase independence.

More info here: https://github.com/winglang/wing/issues/435

staycoolcall911:
I'm doing some Housekeeping and closing this PR after chatting with @Warkanlock. 
#435 (phase independence) will not be implemented in the near term.

PR #5610, opened 2024-02-02
feat: prevent references to preflight values inflight and inline imports in typescript

MarkMcCulloh:
These work now!

```ts
import { inflight, main } from "@wingcloud/framework";
import assert from "node:assert";
import { addOne } from "./other";

const x = 2;

main((app, test) => {
  test(
    "test",
    inflight(async () => {
      // can reference modules imported at the top of the file
      assert.deepStrictEqual(1, 1);

      // This works for relative imports as well (they get resolved as absolute)
      const two = addOne(1);

      // This now causes an error at compile time, letting you know that `x` must be lifted
      // Nothing in the IDE yet, that's a whole other project
      console.log(x);
    })
  );
});
```

I updated typescript across the board initially because I was going to use https://github.com/nonara/ts-patch, but I realized using that package was not needed (for now).
I can revert the update if desired, but honestly why not keep it.


libs/@wingcloud/framework/src/internal.ts
```diff
@@ -26,17 +27,172 @@ export async function compile(options: CompileOptions) {
...
-  }
-  if (results.emitSkipped) {
+  let extraErrors: ts.Diagnostic[] = [];
+
+  const emitResult = program.emit(undefined, undefined, undefined, undefined, {
+    before: [
+      // This transformer finds all usages of `inflight` and does the following with the functions passed to it:
+      // 1. Replaces references to top-level imports with inline `require` calls
+      // 2. Throws an error if a reference to a non-import variable declared in an outer scope is used
+      (context) => {
```
eladb, line 38: Maybe extract to a separate function/file

libs/@wingcloud/framework/src/internal.ts
```diff
@@ -26,17 +27,172 @@ export async function compile(options: CompileOptions) {
...
   });
-  const results = program.emit();
 
-  for (const diagnostic of results.diagnostics) {
-    console.error(diagnostic.messageText);
-  }
-  if (results.emitSkipped) {
+  let extraErrors: ts.Diagnostic[] = [];
+
+  const emitResult = program.emit(undefined, undefined, undefined, undefined, {
```
eladb, line 33: Test for this?

At least minimal

eladb (approved):

Issue #5497, opened 2024-01-22
Add Slack Notifications on Wing new Quickstart Templates


hasanaburayyan:
### Use Case

To celebrate that users are creating new quickstart projects, Wing cli now notifies the Wing Slackbot that a new quickstart project has been created and historical statistics about the quickstart template.

Example:
🚀 A new "private-api-aws" project was just created! The total number of "private-api-aws" projects created so far is 2,340	

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

tsuf239:
Where should we save the total number? we need some sort of a db (or a counter) right? Should we open another AWS account?

revitalbarletz:
When a new project is created, we need to be able to get the user id so that we can correlate the id to the user email through the console sign in events.

hasanaburayyan:
Closing since we are able to use dashboards to correlate these ids

PR #5618, opened 2024-02-04
fix(sdk): add template notifications

tsuf239:
## Checklist
closes https://github.com/winglang/wing/issues/5497

Sending a notification whenever a new template event is created- which includes the template and the language names.
image

- Adding the cli options to be in a JSON form (and therefore we can use them in segment)
- Making sure they'll appear in the `cli__new` notification
- Adding a new slack destination in segment, it has to be enabled after merging this feature

The aggregated count is possible only for a higher tier in Segment. I talked to @revitalbarletz and we decided to make this data accessible as a part of a different view - that'll include more needed information as well.

- [ ] Tests added (always)


apps/wing/src/analytics/collectors/cli-collector.ts
```diff
@@ -28,7 +28,7 @@ export class CLICollector extends Collector {
     return {
       platform: platform.map((p: string) => basename(p)).join(","), // only report the platform name, not the full path
       target: determineTargetFromPlatforms(platform),
-      options: `${JSON.stringify(this.cmd.opts())}`,
```
hasanaburayyan, line 31: So I think this will result in a nested object right? If I recall correctly @ekeren advised against this as there are complications with nested JSON objects in segment. (unless Im misremembering)

ekeren: yes, nested json is not good

tsuf239: may I ask why?
I saw that segment handles them pretty well- and also you can use them as an additional filter-
without it we cannot parse the command option via segment (we won't be able to pass the language and the template to the slack notification, but as a json stringified string)

ekeren: segment deals with it well, but amplitude (which is where we actually consume the analytics via graphs) doesn't know how to handle them.

we can concat keys

e.g. 
```
{
  a: { 
    b: 
      {
        c: "v1"
      } 
    },
      d: "v2" 
  }
  d:"v3"
 }
```
==> 
```
{
a_b_c: "v1",
a_d: "v2",
d: "v3"
}
```

ekeren: Other consumers (like mixpanel) and google analytics are also not great with these nested json values...

tsuf239: added :) 
image
(The flattening is at our end now)

hasanaburayyan (approved):
Flattening seems like a good option here!

PR #5631, opened 2024-02-06
feat: jsii closure types from interfaces and allowing more generics in the sdk

MarkMcCulloh:
The purpose of this PR was to unblock more safe implementations of https://github.com/winglang/wing/issues/4998 and https://github.com/winglang/wing/issues/5005

To that end, the following changes were needed

- `@typeparam` annotation can now be used on interfaces in TS
- Generics methods can now take/return other generic classes/interfaces
- Generics methods can now take/return `[]`s of generic classes/interfaces
- `@callable` can be added to an interface in TS to turn the type into a function, where the first (and only) member of the interface will be used as the type signature

All means you could define something like this in TS

```ts
/**
* @typeparam T1
* @callable
*/
export interface IComparer {
  fn(x: T1, y: T1): num;
}

/**
* @typeparam T1
*/
class Array {
  sort(sorter: IComparer) { ... }
}
```

and then use in wing like this

```wing
[1, 2, 3].sort((x, y) => { ... })
```

Note on testing: I have no idea how to test the generic stuff without implementing the features. I partially implemented them to test this PR out, but nothing to commit.


ekeren:
Nice!

MarkMcCulloh:
Putting this in draft to rework the implementation

Warkanlock:
awesome @MarkMcCulloh

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -302,6 +302,63 @@ impl<'a> JsiiImporter<'a> {
 		self.register_jsii_type(&enum_fqn, &enum_symbol, enum_type_ref);
 	}
 
+	/// Import a JSII interface as a function instead.
+	/// These interfaces must have the @callable annotation and only one method defined, which will be the function signature.
+	fn import_closure(&mut self, jsii_interface: &wingii::jsii::InterfaceType) {
```
Warkanlock, line 307: @MarkMcCulloh just a question to understand in depth this change; would this new method allow closures coming from the `.jsii` compiled via the SDK? (as long as they have the `@typeparam` annotation in place)

Warkanlock: (assuming it was called from `import_interface`)

Warkanlock (approved):

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -302,6 +302,63 @@ impl<'a> JsiiImporter<'a> {
...
+
+		// TODO These should technically be "phase independent" by default, but wing currently doesn't have a way to create
+		// a phase independent function, so it's more useful to pick one for now.
+		let phase = Phase::Preflight;
+
+		let new_type_symbol = Self::jsii_name_to_symbol(&type_name, &jsii_interface.location_in_module);
+
+		let first_method = if let Some(methods) = &jsii_interface.methods {
+			if methods.len() != 1 {
+				panic!("Expected exactly one method defined in {jsii_interface_fqn}")
```
Chriscbr, line 321:
```suggestion
				panic!("Expected exactly one method defined in {jsii_interface_fqn} annotated with @callable")
```

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -302,6 +302,63 @@ impl<'a> JsiiImporter<'a> {
...
+			} else {
+				methods.first().unwrap()
+			}
+		} else {
+			panic!("Expected at least one method")
```
Chriscbr, line 326:
```suggestion
			panic!("Expected at least one method in {jsii_interface_fqn} annotated with @callable")
```

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -196,6 +197,31 @@ impl SymbolEnv {
...
+		phase: Phase,
+		statement_idx: usize,
+		type_params: Vec<UnsafeRef<Type>>,
+	) -> Self {
+		// Some sanity checks
+		// If parent is a type-environent this must be one too
+		assert!(
+			parent.is_none()
+				|| matches!(parent, Some(parent) if matches!(parent.kind, SymbolEnvKind::Type(_)) == matches!(kind, SymbolEnvKind::Type(_)))
+		);
```
Chriscbr, line 216: nit: refactor into a function since it's repeated above

libs/wingc/src/type_check.rs
```diff
@@ -5440,7 +5490,16 @@ impl<'a> TypeChecker<'a> {
...
 
-						let new_class = self.hydrate_class_type_arguments(env, WINGSDK_STRUCT, vec![type_]);
+						let new_class = self.hydrate_class_type_arguments(
+							env,
+							env
+								.lookup_nested_str(WINGSDK_STRUCT, None)
+								.unwrap()
+								.0
+								.as_type()
+								.unwrap(),
```
Chriscbr, line 5500: nit: Write a helper for this maybe?

Chriscbr (approved):

libs/wingc/src/type_check/jsii_importer.rs
MarkMcCulloh: If you create an interface with `@callable`, it will be imported as a function type instead of an interface:

```ts
/*
 * @callable
 */
export interface IFunkyFunction {
  fn: (arg: string) => number;
}
```

will become the wing function type:

```wing
(str): num
```

It can also use `@typeparam` to have generic args/return type

PR #5637, opened 2024-02-07
fix(sdk): cannot reference api.url within cloud.Api route on AWS

### LINKED ISSUE #2740: `cloud.Api` endpoint that uses `api.url` deploys to simulator but not to AWS
### I tried this:

```js
bring cloud;

let api = new cloud.Api();
api.get("/route", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
  log(api.url);
  return cloud.ApiResponse {
    status: 200,
    body: Json "ok"
  };
});
```

### This happened:

`wing test -t sim api_url.w` passes (it simulates the app, and runs all 0 of 0 tests)

But `wing test -t tf-aws api_url.w` fails when it tries to run `terraform apply`:

`Error: Cycle: aws_api_gateway_deployment.root_env0_cloudApi_api_deployment_FABCA7C9, aws_api_gateway_rest_api.root_env0_cloudApi_api_3A5A95BE, aws_api_gateway_stage.root_env0_cloudApi_api_stage_0D945F2C, aws_lambda_function.root_env0_cloudApi_cloudApiOnRequest3915c5ec_DC2C0B9C`

### I expected this:

Either the app should deploy, or I should get some error at compile time

### Is there a workaround?

_No response_

### Component

SDK

### Wing Version

0.18.1

### Wing Console Version

_No response_

### Node.js Version

18.14.1

### Platform(s)

MacOS

### Anything else?

Here's a breakdown of the dependency cycle:

1. In order to log `api.url`, the AWS Lambda function that is created has an environment variable which depends on the API Gateway stage invoke url:

```json
"environment": {
  "variables": {
    "CLOUD_API_C8499363": "${aws_api_gateway_stage.root_env0_cloudApi_api_stage_0D945F2C.invoke_url}",
    "WING_FUNCTION_NAME": "cloud-Api-OnRequest-3915c5ec-c8195997"
  }
},
```

2. The API Gateway stage depends to the API Gateway deployment ID 
3. The API Gateway deployment depends on the API Gateway rest api ID
4. The API Gateway rest api has the OpenAPI spec for the all endpoints, which refers to the ARN of the AWS Lambda function

### Community Notes

Chriscbr:
Fixes #2740

In this bug, when the user creates an API route whose inflight code references the API's url...

```js
let api = new cloud.Api();

api.get("/my_url", inflight () => {
  return {
    status: 200,
    body: api.url
  };
});
```

... when they deploy to the `tf-aws` platform, it causes resources to be created in a dependency cycle, making it impossible to deploy the app.

The reason this happens is that first, the `cloud.Api` is turned into an API Gateway, whose OpenAPI specification (list of all of the API routes) needs to reference the physical name of the `cloud.Function` (usually called the "ARN" in AWS-speak). Second, the `cloud.Function` is turned into an AWS Lambda function, which needs an environment variable to provide the URL of the API Gateway. Both resources want to use a property of the other resource that seems to only be available once it's created.

To unravel this dependency cycle, I updated the implementation of `cloud.Api` to take leverage of the fact that the ARNs of functions are well-defined in the CDK model, and can almost entirely[1] be determined at compile time. This allows the API Gateway to be deployed before the Lambda function without any issues.

[1] The only exception is that the AWS account name and region is needed, but these values can be obtained freely at deploy time.

## Checklist

- [x] Tests added (always)


eladb (approved):
Nice

PR #5641, opened 2024-02-07
feat(sdk): tf-aws reference existing bucket

hasanaburayyan:
Allows `bring tf-aws` and referencing existing buckets.

so this is now possible
```js
bring tf_aws;

let bucket = tf_aws.Bucket.fromBucketName("my-existing-bucket");

```

## Checklist

- [ ] Tests added (always)


examples/tests/valid/tfaws_reference_bucket.main.w
```diff
@@ -0,0 +1,12 @@
...
+bring util;
+
+
+if util.env("WING_TARGET") == "tf-aws" {
+  let bucket = tf_aws.Bucket.fromBucketName(this, "some-existing-bucket-name");
+
+  new cloud.Function(inflight (msg: str) => {
+    bucket.put("somefile.txt", "wow this goes in an existing bucket!");
+  });
+}
```
Chriscbr, line 12: The underscore looks a bit out of place to me since we've been sticking to camelCase for most Wing example code. @eladb thoughts? Who's the owner of Wing code style?
```suggestion
bring tfaws;
bring cloud;
bring util;


if util.env("WING_TARGET") == "tf-aws" {
  let bucket = tfaws.Bucket.fromBucketName(this, "some-existing-bucket-name");

  new cloud.Function(inflight (msg: str) => {
    bucket.put("somefile.txt", "wow this goes in an existing bucket!");
  });
}
```

tools/hangar/__snapshots__/test_corpus/valid/tfaws_reference_bucket.main.w_compile_tf-aws.md
```diff
@@ -0,0 +1,222 @@
...
+        "//": {
+          "metadata": {
+            "path": "root/Default/Default/some-existing-bucket-name_7dc22579",
+            "uniqueId": "some-existing-bucket-name_7dc22579"
+          }
+        },
+        "bucket": "some-existing-bucket-name"
+      }
+    }
+  },
```
Chriscbr, line 46: Nice 👍

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -46,20 +47,54 @@ export const BUCKET_PREFIX_OPTS: NameOptions = {
...
   private readonly notificationDependencies: ITerraformDependable[] = [];
 
-  constructor(scope: Construct, id: string, props: cloud.BucketProps = {}) {
+  constructor(scope: Construct, id: string, props: BucketProps = {}) {
     super(scope, id, props);
-
+    if (props._existingBucket) {
+      this.bucket = props._existingBucket;
+      return;
+    }
```
Chriscbr, line 97: Suppose we wanted to expose whether a bucket is public or not as a public property:

```js
let bucket = new cloud.Bucket(public: true);
assert(bucket.public == true);
```

I'm wondering how we would do that in the current implementation?

(food for thought) I think the way a lot of AWS CDK code does this is they define an interface, and create a new class that satisfies:

```ts
interface IBucket extends IResource {
  ...
}

class Bucket {
  public static fromBucketName(scope: Construct, bucketName: string): IBucket {
    class BucketReference extends Construct implements IBucket {
      ...
    }
    return new BucketReference(scope, ...);
  }
}
```

eladb:
```js
public class Bucket {
  public static fromBucketName(scope: Construct, bucketName: string): Bucket {
    return new BucketReference(scope, id, bucketName);
  }
}

class BucketReference extends cloud.Bucket {
  get public() {
    throw new Error("not supported for referenced bucket");
  }
}
```

☝️ Something like this

examples/tests/valid/tfaws_reference_bucket.main.w
eladb: Yes, this should be `tfaws` I think.

libs/wingsdk/src/target-tf-aws/bucket.ts
```diff
@@ -46,20 +47,54 @@ export const BUCKET_PREFIX_OPTS: NameOptions = {
...
-  private readonly bucket: S3Bucket;
-  private readonly public: boolean;
+  /**
+   * Create a new cloud.Bucket from an existing s3 bucket.
+   *
+   * @param scope scope to use for the bucket data lookup
+   * @param bucketName the name of the existing bucket to reference
+   * @returns A cloud.Bucket
+   */
+  public static fromBucketName(scope: Construct, bucketName: string): Bucket {
```
eladb, line 76: We also need `fromBucketArn()`

eladb (changes requested):

libs/wingsdk/src/target-tf-aws/bucket.ts
hasanaburayyan: So terraform doesnt actually support data lookups based on bucket arns, only bucket names: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/s3_bucket#argument-reference

I assume this is because Arns are not needed to identify uniqueness as S3 bucket names are globally unique.

Now we could technically still provide the API since the arn contains the bucket name, we would just parse the bucket name from the arn and look it up like that, but I wonder how useful it is to have both APIs?

eladb: We can start with just the name of that feels okay but in my experience the idiomatic way in AWS to reference resources is ARN so all of our `from` methods should have at least an import from ARN

hasanaburayyan:
Need to follow the example of: https://github.com/winglang/wing/pull/6001

staycoolcall911:
Closing for now - this was implemented in #6210

PR #5642, opened 2024-02-07
fix: possible keep-alive related issue in simulator

MarkMcCulloh:
testing some stuff

Not sure we need keep-alive enable in the simulator client. 
We aren't reusing the agent, so I don't think we're reusing sockets anyways.


libs/wingsdk/src/simulator/client.ts
```diff
@@ -20,17 +20,12 @@ export function makeSimulatorClient(url: string, handle: string) {
...
           method: "POST",
           headers: { "Content-Type": "application/json" },
           body: JSON.stringify(body),
-          dispatcher: new Agent({
-            keepAliveTimeout: 15 * 60 * 1000,
-            keepAliveMaxTimeout: 15 * 60 * 1000,
-            headersTimeout: 15 * 60 * 1000,
-            bodyTimeout: 15 * 60 * 1000,
-          }),
+          keepalive: false,
```
Chriscbr, line 28: IIRC one of the reasons I brought in undici was to customize some of these other fields that weren't available with the built-in fetch -- if we don't need them it could be worth trying to use the built-in `http` module instead

Chriscbr (approved):

libs/wingsdk/src/simulator/client.ts
MarkMcCulloh: fetch is still considered experimental in node versions we support and it will warn the users if they're not on a new enough version, so I think keeping undici for now is ok

MarkMcCulloh:
Closing this for now. Nothing I tried here appeared to help with the issue

Issue #901, opened 2022-12-14
Protection and tools for managing stateful resources and avoiding deletion


Chriscbr:
### RFC (Rev 1.)

Here's a proposal for a simple mechanism to prevent damage caused by logical name mapping changes.

If a resource is marked as `stateful`, it's Terraform resource name will be determined by looking up its path in a map stored in a `<entrypoint>.w.lock` file, next to the entrypoint file, which will be committed to the repository.

If an entry for this resource cannot be found in the lockfile, the compilation will fail and instruct the user to explicitly add an entry and provide a resource name.

Additionally, compilation will also fail if the file includes an entry that doesn't map to a stateful resource in the app.

This will ensure that:

1. Stateful resources are explicitly named.
2. If a stateful resource is moved within the app (e.g. it's path is changed), then users will easily be able to map the new path to the same stateful resource.
3. If a stateful resource is accidentally removed, users will have to acknowledge this by explicitly removing the resource entry from the lockfile.

We can offer some CLI commands to edit the lockfile, but that's not P1.

Here's an example.

Say I write this Wing called `hello.main.w`:

```js
bring cloud;

new cloud.Bucket();
```

Now, I compile:

```sh
$ wing compile -t tf-aws hello.main.w
ERROR: stateful resource "/root/Bucket" doesn't have a name in the lockfile.
Please edit `hello.main.w.lock`.
$ cat hello.main.w.lock
/root/Bucket: <NEW>
$ edit hello.main.w.lock
/root/Bucket: my_s3_bucket
$ wing compile -t tf-aws hello.main.w
$ # cool!
```

Now, let's say we refactored the app and we will move the same bucket into a class:

```js
bring cloud;

class MyStore {
  new() {
    new cloud.Bucket();
  }
}

new MyStore();
```

```sh
$ wing compile -t tf-aws hello.main.w
ERROR: stateful resource "/root/MyStore/Bucket" doesn't have a name in the lockfile.
ERROR: stateful resource "/root/Bucket" with the name "my_s3_bucket" was deleted.
Please edit `hello.main.w.lock`.
$ cat hello.main.w.lock
/root/Bucket: my_s3_bucket <DELETED>
/root/MyStore/Bucket: <NEW>
$ edit hello.main.w.lock
/root/MyStore/Bucket: my_s3_bucket
$ wing compile -t tf-aws hello.main.w
```

At some point we can be smarter and offer a nice DX:

```
It seems like the stateful resource "my_s3_bucket" was moved
[...]

PR #5658, opened 2024-02-08
fix(compiler)!: default object ids are volatile because of namespace

### LINKED ISSUE #5564: resource recreated when changing the library alias
### I tried this:

given the following code:

```wing
bring "@cdktf/provider-aws" as aws;
let distribution = new aws.cloudfrontDistribution.CloudfrontDistribution();
```

when it is changed to:
```wing
bring "@cdktf/provider-aws" as awsProvider;
let distribution = new awsProvider.cloudfrontDistribution.CloudfrontDistribution();
```

it will recreate the resource which might be unexpected

### This happened:

resource was deleted and recreated

### I expected this:

_No response_

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

eladb:
Use only the class name instead of the fully qualified name as the default preflight object ID because the namespace (e.g. `foo` in `foo.MyClass`) is a local name determined by the `bring` statement.

Fixes #5564

BREAKING CHANGE: Caution ⚠️ (and apologies 🙏)! This change is going to cause a replacement of any resource that didn't have an explicit identity. We are working to make it [safer and easier](https://github.com/winglang/wing/issues/901) to control IDs of resources (especially ones with state), but for now, if you wish to avoid the replacement, you'll need to add `as "cloud.Xxx"` to explicitly use the old naming. Don't hesitate to [ping us](https://t.winglang.io/slack) and we will help out with migration as needed.

## Checklist

- [x] Tests added (always)


Chriscbr (approved):
A worthy breaking change 👍

staycoolcall911:
We'll wait for #901 to be done before we release this breaking change

PR #5659, opened 2024-02-09
feat(sdk): support apig and lambda behind vpc

hasanaburayyan:
Supports deploying lamdbas and api gateways behind a vpc. also ensures that when API gateways are deployed behind vpcs that they use an VPC endpoint for invocation

to enable parameters `tf-aws/vpc_api_gateway` and `tf-aws/vpc_lambda` must be set

I tested the implementation in my own aws account as well:
Here we cannot hit the endpoint publicly
image

Here the lambda was able to invoke the API gateway
image

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/test/target-tf-aws/__snapshots__/bucket.test.ts.snap
+++ b/libs/wingsdk/test/target-tf-aws/__snapshots__/bucket.test.ts.snap
@@ -444,6 +444,14 @@ exports[`bucket with onCreate method 2`] = `
                 "id": "Code",
                 "path": "root/Default/Code",
               },
+              "ParameterRegistrar": {
+                "constructInfo": {
+                  "fqn": "constructs.Construct",
+                  "version": "10.3.0",
+                },
+                "id": "ParameterRegistrar",
+                "path": "root/Default/ParameterRegistrar",
+              },
               "aws": {
                 "constructInfo": {
                   "fqn": "cdktf.TerraformProvider",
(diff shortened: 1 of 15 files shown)


libs/wingsdk/test/target-tf-aws/__snapshots__/bucket.test.ts.snap
```diff
@@ -444,6 +444,14 @@ exports[`bucket with onCreate method 2`] = `
...
                 "path": "root/Default/Code",
               },
+              "ParameterRegistrar": {
+                "constructInfo": {
+                  "fqn": "constructs.Construct",
+                  "version": "10.3.0",
+                },
+                "id": "ParameterRegistrar",
+                "path": "root/Default/ParameterRegistrar",
+              },
```
hasanaburayyan, line 454: Not sure why these additions only just happened and weren't committed with the original PR to add platform parameter registration.

Chriscbr (commented):
looks great @hasanaburayyan!

What do you think would be a good way to test this?

hasanaburayyan:
> looks great @hasanaburayyan!
> 
> What do you think would be a good way to test this?

@Chriscbr great question 😂 I thought about just adding an ask test for this but I hesitated because it would greatly increase the ask spec test since VPCs take forever to tear down. Maybe that’s fine though

or perhaps it’s fine to just be satisfied with vpc configs in the terraform templates. 🤷‍♂️

Chriscbr:
Terraform config snapshot sounds great - E2E test is nice too, but I'm cool with leaving it out of scope since we could quickly run up against quotas with VPCs

Chriscbr (approved):

PR #5666, opened 2024-02-09
fix(compiler): panic when method has same name as a field

### LINKED ISSUE #1635: Naming a resource method with the same name as a field causes a compiler panic
### Community Note

### I tried this:

```js
// wing
bring cloud;

class A {
  cache_hits: cloud.Counter;
  new() {
    this.cache_hits = new cloud.Counter();
  }
  inflight cache_hits(): num {
    return this.cache_hits.peek();
  }
}
```

### I expected this:

A compiler error on `cache_hits` saying that there is already a method with the same name.

### Instead, this happened:

```
thread '<unnamed>' panicked at 'Expected method type to be a function signature', libs/wingc/src/type_check.rs:1752:38
stack backtrace:
unreachable
```

### Component:

Compiler

### Wing Version:

0.5.67

### Node.js Version:

18.12.1

### Platform(s):

MacOS

### Anything else?

_No response_

Chriscbr:
Fixes #1635

Fixing this required a bit of refactoring code around because even when there are member name conflicts, we want to make sure all of the code in the method still gets type checked. Type checking the body requires knowing the type signature of the method. But the method's correct type isn't available in the environment because of the name conflicts.

To solve this, I added some logic to keep track of a mapping of method name symbols to the method types within `type_check_class` (where `type_check_method` is called from). Since the keys of the map are symbols, and each symbol has a name and source location, the method types won't overwrite each other.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/invalid/class.test.w
+++ b/examples/tests/invalid/class.test.w
@@ -164,3 +164,32 @@ inflight class Jet extends Plane{
 
 new std.Resource();
 //^^^^^^^^^^^^^^^^ Cannot instantiate abstract class "Resource"
+
+// Class with field and method of the same name (field comes first)
+class C12 {
+  x: num;
+  x() {
+  //^ Symbol "x" is already defined
+    // The method body should still be type checked
+    2 + "2";
+    //^ Expected type to be "num", but got "str" instead
+    x == 5; // OK
+  }
+  new() {
+    this.x = 5;
+  }
+}
+
+// Class with method and field of the same name (method comes first)
+class C13 {
+  x() {
+  //^ Symbol "x" is already defined
+    // The method body should still be type checked
+    2 + "2";
+    //^ Expected type to be "num", but got "str" instead
+  }
+  x: num;
+  new() {
+    this.x = 5;
+  }
+}
--- a/examples/tests/invalid/class.test.w
+++ b/examples/tests/invalid/class.test.w
@@ -168,28 +168,50 @@ new std.Resource();
 // Class with field and method of the same name (field comes first)
 class C12 {
   x: num;
-  x() {
+  x(): num {
   //^ Symbol "x" is already defined
     // The method body should still be type checked
     2 + "2";
     //^ Expected type to be "num", but got "str" instead
     x == 5; // OK
+    return "hello";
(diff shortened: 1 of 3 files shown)


Chriscbr:
Screenshot from VS Code:

Screenshot 2024-02-09 at 6 47 34 PM

examples/tests/invalid/class.test.w
```diff
@@ -163,4 +163,55 @@ inflight class Jet extends Plane{
...
+
+// Class with field and method of the same name (field comes first)
+class C12 {
+  z: num;
+  z(): num {
+  //^ Symbol "z" is already defined
+    // The method body should still be type checked
+    2 + "2";
+    //^ Expected type to be "num", but got "str" instead
+    this.z == 5; // OK
```
Chriscbr, line 176: Within the body of `z() {...}`, `this.z` is typed as a `num` and not as a function. Is this correct? This can probably be debated but I think either behavior is OK since the user has a naming issue they need to fix

staycoolcall911: Yeah, I think it's ok either way. Let's give it a trt as you implemented and see how it goes

examples/tests/invalid/class.test.w
```diff
@@ -163,4 +163,55 @@ inflight class Jet extends Plane{
...
+    //^ Expected type to be "num", but got "str" instead
+    return "hello";
+    //^ Expected type to be "num", but got "str" instead
+  }
+  z: num;
+  new() {
+    this.z = 5;
+  }
+  z(): str {
+  //^ Symbol "z" is already defined
```
eladb, line 209: Can we improve this diagnostic? Something like "There's already a field in this class with this name"?

eladb (approved):

examples/tests/invalid/class.test.w
Chriscbr: Nice improvement idea 👍  I think solving this well requires some more refactoring so I'll table it for now

Fortunately the IDE and CLI output of `wing compile` will show you the location of the previous symbol definition so the experience isn't too bad

PR #5690, opened 2024-02-13
fix: invoke(nil) fails on aws for cloud.Function

MarkMcCulloh:
aws tests https://github.com/winglang/wing/actions/runs/7891384592

(passed running `wing test -f tf-aws` locally)

### DIFF
--- a/libs/wingsdk/src/shared-aws/function.inflight.ts
+++ b/libs/wingsdk/src/shared-aws/function.inflight.ts
@@ -22,6 +22,9 @@ export class FunctionClient implements IFunctionClient {
   public async invoke(payload?: string): Promise<string | undefined> {
     const command = new InvokeCommand({
       FunctionName: this.functionArn,
+      // If payload is undefined, pass json `null` as the payload to the function
+      // to ensure the received event will be `null` (which will be converted to `undefined` in the function code)
+      // If the Payload is undefined, the resulting event will instead be `{}`
       Payload: fromUtf8(payload ? JSON.stringify(payload) : "null"),
     });
     const response = await this.lambdaClient.send(command);
--- a/libs/wingsdk/src/shared-aws/function.inflight.ts
+++ b/libs/wingsdk/src/shared-aws/function.inflight.ts
@@ -25,7 +25,9 @@ export class FunctionClient implements IFunctionClient {
       // If payload is undefined, pass json `null` as the payload to the function
       // to ensure the received event will be `null` (which will be converted to `undefined` in the function code)
       // If the Payload is undefined, the resulting event will instead be `{}`
-      Payload: fromUtf8(payload ? JSON.stringify(payload) : "null"),
+      Payload: fromUtf8(
+        payload !== undefined ? JSON.stringify(payload) : "null"
+      ),
     });
     const response = await this.lambdaClient.send(command);
     return parseCommandOutput(response, this.functionArn);
--- a/libs/wingsdk/src/shared-aws/function.inflight.ts
+++ b/libs/wingsdk/src/shared-aws/function.inflight.ts
@@ -22,7 +22,7 @@ export class FunctionClient implements IFunctionClient {
   public async invoke(payload?: string): Promise<string | undefined> {
     const command = new InvokeCommand({
       FunctionName: this.functionArn,
-      Payload: payload ? fromUtf8(JSON.stringify(payload)) : undefined,
+      Payload: fromUtf8(payload ? JSON.stringify(payload) : "null"),
     });
     const response = await this.lambdaClient.send(command);
     return parseCommandOutput(response, this.functionArn);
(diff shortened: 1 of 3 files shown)


eladb (approved):
Is this also relevant in GCP/Azure?

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -22,7 +22,7 @@ export class FunctionClient implements IFunctionClient {
   public async invoke(payload?: string): Promise<string | undefined> {
     const command = new InvokeCommand({
       FunctionName: this.functionArn,
-      Payload: payload ? fromUtf8(JSON.stringify(payload)) : undefined,
+      Payload: fromUtf8(payload ? JSON.stringify(payload) : "null"),
```
MarkMcCulloh, line 25: aws needs a json value, otherwise the handler argument is `{}`

using the json `null` makes sense to speak AWS's language here

eladb, line 25: Add a message indicating that since we are stringifying the payload the string "null" cannot be injected. (I think)

eladb (approved):

MarkMcCulloh:
> Is this also relevant in GCP/Azure?

I'm not sure how their respective clients handle this, so maybe. Our current automated tests don't run against gcp/azure and I don't have it set up locally so I didn't try it

libs/wingsdk/src/shared-aws/function.inflight.ts
MarkMcCulloh: Added a comment. This is basically matching the sim's behavior where we treat json's `null` as `undefined`

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -22,7 +22,10 @@ export class FunctionClient implements IFunctionClient {
   public async invoke(payload?: string): Promise<string | undefined> {
     const command = new InvokeCommand({
       FunctionName: this.functionArn,
-      Payload: payload ? fromUtf8(JSON.stringify(payload)) : undefined,
+      // If payload is undefined, pass json `null` as the payload to the function
+      // to ensure the received event will be `null` (which will be converted to `undefined` in the function code)
+      // If the Payload is undefined, the resulting event will instead be `{}`
+      Payload: fromUtf8(payload ? JSON.stringify(payload) : "null"),
```
Chriscbr, line 28: what if the payload is empty string or 0? `payload ? JSON.stringify(payload) : "null"` would evaluate to "null" instead of "\"\"" or "0" respectively

libs/wingsdk/src/shared-aws/function.inflight.ts
```diff
@@ -110,7 +113,8 @@ function parseCommandOutput(
   if (!payload.Payload) {
     return undefined;
   } else {
-    return JSON.parse(toUtf8(payload.Payload));
+    const returnObject = JSON.parse(toUtf8(payload.Payload));
+    return returnObject === null ? undefined : returnObject;
```
Chriscbr, line 117: This shouldn't be needed -- `==` in the language treats null and undefined as the same (I believe that's the semantics we want)

Chriscbr: Maybe a bug with expect.equal?

MarkMcCulloh: I guess you're right that this is a bug with expect.equal though. Shouldn't we still avoid `null` bubbling through at runtime though?

libs/wingsdk/src/shared-aws/function.inflight.ts
MarkMcCulloh: 0 wouldn't be relevent since this value is either a string or undefined, but you're right about empty string (plus being explicit is good regardless)

libs/wingsdk/src/shared-aws/function.inflight.ts
Chriscbr: It's an open decision. Since Wing doesn't distinguish between null and undefined in its type system, I think there are two ways things can work when we emit and run JavaScript code:
1. enforce that only one of them is used (e.g. use "undefined" everywhere). This means if someone writes an extern function typed as returning `str?` and their JavaScript code implementation returns a null instead of undefined, then mysterious bugs may start occurring.
2. allow them to be used interchangeably. This puts a bit more burden on us (SDK and compiler internals need to be permissive to either appearing) but results in fewer errors for users, IMO. It's possible there are holes in this approach I'm not thinking of though.

In any case I'm cool with converting null's into undefined's by hand here as quick fix

MarkMcCulloh: In the context of this specific function in TypeScript, this ternary is technically just doing a better job of ensuring we are returning the correct thing. The return type here is `string | undefined`, not `string | undefined | null`.

PR #5705, opened 2024-02-14
chore(cli): make private-api template more interesting

hasanaburayyan:
This pr:
- Adds more interesting example to private-api template
- Adds more comments about what the code does
- Adds typescript version

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing/project-templates/typescript/private-api/main.ts
+++ b/apps/wing/project-templates/typescript/private-api/main.ts
@@ -21,6 +21,9 @@ import { cloud, lift, main } from "@wingcloud/framework";
  * And retrieve the note:
  * `aws lambda invoke --cli-binary-format raw-in-base64-out --function-name <function-name> --payload "\"n1\"" response.json`
  * `cat response.json`
+ * 
+ * If you have not made changes to this template you can find your function names in AWS using the following command:
+ * `aws lambda list-functions --query "Functions[?starts_with(FunctionName, 'Consumer')].FunctionName"`
  */
 main((root) => {
   let noteBucket = new cloud.Bucket(root, "Bucket");
@@ -28,7 +31,6 @@ main((root) => {
   const api = new cloud.Api(root, "Api");
 
   api.get('/note', lift({noteBucket}).inflight(async ({noteBucket}, request) => {
-    console.log("REQUEST:", request)
     let noteName = request.query.name; 
     const note = await noteBucket.get(noteName);
 
@@ -65,7 +67,6 @@ main((root) => {
       return "event is required `NAME:NOTE`"
     }
     const [name, note] = event!.split(":");
-    // send put request to the api
     const response = await fetch(`${url}/note/${name}`, {
       method: 'PUT',
       body: JSON.stringify(note)
--- a/apps/wing/project-templates/typescript/private-api/main.ts
+++ b/apps/wing/project-templates/typescript/private-api/main.ts
@@ -1,0 +1,86 @@
+import { cloud, lift, main } from "@wingcloud/framework";
+
+/**
+ * The example below is a simple note-taking app.
+ * It uses a cloud.Bucket to store notes and a cloud.Api to expose a RESTful interface.
+ * 
+ * The api has two endpoints:
+ * - GET /note?name=NAME to get a note by name
+ * - PUT /note/:NAME to save a note by name
+ * 
+ * The app also includes two cloud.Functions to consume the api.
+ * - Consumer-PUT reads a string like: `NAME:NOTE` and calls the api to save it.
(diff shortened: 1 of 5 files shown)


apps/wing/project-templates/typescript/private-api/wing.toml
```diff
@@ -0,0 +1,11 @@
...
+# vpc can be set to "new" or "existing"
+vpc = "new"
+# vpc_lambda will ensure that lambda functions are created within the vpc on the private subnet
+vpc_lambda = true
+# vpc_api_gateway will ensure that the api gateway is created within the vpc on the private subnet
+vpc_api_gateway = true
+# The following parameters will be required if using "existing" vpc
+# vpc_id = "vpc-123xyz"
+# private_subnet_id = "subnet-123xyz"
+# public_subnet_id = "subnet-123xyz"
```
tsuf239, line 11: Nice!!

apps/wing/project-templates/wing/private-api/main.w
```diff
@@ -1,20 +1,84 @@
...
+ * `aws lambda invoke --cli-binary-format raw-in-base64-out --function-name <function-name> --payload "\"n1:this is my note\"" response.json`
+ * `cat response.json`
+ * 
+ * And retrieve the note:
+ * `aws lambda invoke --cli-binary-format raw-in-base64-out --function-name <function-name> --payload "\"n1\"" response.json`
+ * `cat response.json`
+ *
+ * If you have not made changes to this template you can find your function names in AWS using the following command:
+ * `aws lambda list-functions --query "Functions[?starts_with(FunctionName, 'Consumer')].FunctionName"`
+ */
```
tsuf239, line 28: For me (as a non-patient non-native speaker) it seems like a lot of text, maybe it will be good to split it? lines 5-14 in this chunk, and lines 16-28 before the consumer functions
we don't add tests because they can't reach through the vpc?

apps/wing/project-templates/wing/private-api/wing.toml
```diff
@@ -1,4 +1,11 @@
...
+# vpc_lambda will ensure that lambda functions are created within the vpc on the private subnet
 vpc_lambda = true
-vpc_api_gateway = true
\ No newline at end of file
+# vpc_api_gateway will ensure that the api gateway is created within the vpc on the private subnet
+vpc_api_gateway = true
+# The following parameters will be required if using "existing" vpc
+# vpc_id = "vpc-123xyz"
+# private_subnet_id = "subnet-123xyz"
+# public_subnet_id = "subnet-123xyz"
```
tsuf239, line 11: By the way, just for general knowledge, we currently can't have two vpcs- and then put some lambdas under one of them and some under the other, right? Is it a common use case? (that someone that uses vpc tends to use more than one?)

hasanaburayyan: inter communication between VPCs is a common pattern for some architectures. This typically will require the use of either VPC peering or Transit Gateways. Id say thats probably out of scope since normally these architectures are inspired by strict organizational security concerns. Which means that the best we offer is, "you have a vpc that is peered with another? Kewl select existing and fill in the data"

As for defining multiple vpcs in the parameters and assigning different resources to each, that could be done but Id say thats a custom platform for the usecase.

apps/wing/project-templates/typescript/private-api/main.ts
MarkMcCulloh, line 1: Try running the default vscode ts formatter on this, there's some missing semicolons/whitespace

MarkMcCulloh (approved):

apps/wing/project-templates/wing/private-api/main.w
hasanaburayyan: Stylistically I worry that moving large block of text like that into the middle of the code might make reading the actual code more cumbersome to the user who is trying to see how each component is used/invoked.

Ill add this to the list of questions to maybe run by users during workshop to see how they feel about the templates.

apps/wing/project-templates/typescript/private-api/wing.toml
```diff
@@ -0,0 +1,11 @@
+[ tf-aws ]
+# vpc can be set to "new" or "existing"
+vpc = "new"
+# vpc_lambda will ensure that lambda functions are created within the vpc on the private subnet
+vpc_lambda = true
+# vpc_api_gateway will ensure that the api gateway is created within the vpc on the private subnet
+vpc_api_gateway = true
+# The following parameters will be required if using "existing" vpc
+# vpc_id = "vpc-123xyz"
+# private_subnet_id = "subnet-123xyz"
```
eladb, line 10: this is will be changed to an array right?

PR #5709, opened 2024-02-15
fix(platforms): tf-aws vpc does not support multiple subnets

hasanaburayyan:
These changes support providing multiple subnets (public and private) to the tf-aws vpc that is used for resources that require vpcs.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-tf-aws/function.ts
+++ b/libs/wingsdk/src/target-tf-aws/function.ts
@@ -244,7 +244,7 @@ export class Function extends cloud.Function implements IAwsFunction {
         ],
       });
       this.addNetworkConfig({
-        subnetIds: [app.subnets.private.id],
+        subnetIds: [...app.subnets.private.map((s) => s.id)],
         securityGroupIds: [sg.id],
       });
     }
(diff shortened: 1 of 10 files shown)


libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -244,7 +244,7 @@ export class Function extends cloud.Function implements IAwsFunction {
         ],
       });
       this.addNetworkConfig({
-        subnetIds: [app.subnets.private.id],
+        subnetIds: [...app.subnets.private.map((s) => s.id)],
```
hasanaburayyan, line 247: I need to write a test that shows multiple subnets attached when using `vpc = "existing"`

Chriscbr (approved):

libs/wingsdk/src/target-tf-aws/function.ts
eladb: Nice!

eladb (approved):
This is getting cooler and cooler

Issue #4604, opened 2023-10-19
Change `as "id"` to `is "id"`


eladb:
We want to use `as` for casting, so let's replace the current usage with `is`:

```js
new Bucket() is "myBucket";
new Bucket() is "yourBucket";
```

Thoughts?

skyrpex:
I kind of prefer the old fashioned way:

```
new Bucket(id: "yourBucket");
```

But I understand it leads to verbose and even possibly confusing paths...

So, `is` is fine to me.

Chriscbr:
Sounds good to me, I think unlocking "as" for casting would be nice.

yoav-steinberg:
What's the issue # for lang support for casting? Can help putting this in context.

eladb:
Considering @skyrpex's idea. We can synthetic optional keyword argument called `id` and then the usage will be:

```js
new R(id: "some_R")
```

It's a bit of a hack, but from a DX perspective it feels right.

Lancear:
What would this mean for struct expansion in function calls with an id field? Wouldn't this clash with the named arguments?

marciocadev:
Why not use `to` for casting instead of `as`?

eladb:
> What would this mean for struct expansion in function calls with an id field? Wouldn't this clash with the named arguments?

Yes, it would clash. We can decide one of the following:
1. To make `id` a reserved name for keyword arguments (could be a bit annoying)
2. If there's an explicit `id`, just use this value as the logical ID. Basically "embrace the clash".
3. Use something like `$id` or some other more unique name for the synthetic identifier.

At any rate, I am growing to like this direction because it does result in a very intuitive DX.

```js
new cloud.Bucket(id: "my-bucket");
new cloud.Queue();
new cloud.Function(id: "my-function", inflight () => {
  log("hey!");
});
```

I really hated the fact that the identifier was something you'd have to specify outside of the call.

eladb:
> Why not use `to` for casting instead of `as`?


`as` is commonly used for casting and I don't think we have a strong justification to shift the mental model here.

Lancear:
Then I would go for embrace the clash. Seems like the best DX for when it clashes.

marciocadev:
> I like `new R() id "some_R"` because it's explicit about what we're doing, setting a unique id for the resource.

I think the most practical way would be to replace the current as with id, as suggested by Yoav's syntax.

garysassano:
[...]

PR #5711, opened 2024-02-15
rfc: resource pinning

MarkMcCulloh:
https://github.com/winglang/wing/issues/901

[Rendered Version](https://github.com/winglang/wing/blob/mark/rfc-pin/docs/contributing/999-rfcs/2024-02-15-resource-pinning.md)


docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,159 @@
...
+
+Running `wing compile -t tf-aws` also succeeds with the following output and automatically generates a lockfile next to the entrypoint file.
+
+```shell
+# (output)
+[+pin] /root/CoolBucket
+```
+
+```json
+// main.w.lock.json
```
eladb, line 45:
```suggestion
// main.w.pin.json
```

semantically connect to `pin`

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,159 @@
...
+
+Given all this, I think it's important to avoid the word "stateful" and to provide tools to express the true intent for API consumers rather than as an intrinsic property of the resource.
+
+### pinning
+
+The `pin` function is used to mark a created resource as "pinned" in the resource tree. It takes a construct and returns the same construct, with metadata added to the tree to track the pin. This means that the user does not intend for the resource to be deleted or moved. Pinning is deep, meaning that pinning a parent inherently pins all of its children.
+
+This information will be available to the target platform/app which can choose what to do with this information. For example, the sim target can simply ignore this while the terraform target can manage a lockfile to inform the user when this changes between compilation and allow them to resolve any renames.
+
+### Lockfile API
```
eladb, line 131:
```suggestion
### Pinfile API
```

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,159 @@
...
+new cloud.Bucket() as "CoolBucket";
+```
+
+because they want to make sure the bucket doesn't get deleted by accident, they pin it
+
+```wing
+// main.w
+bring cloud;
+
+pin(new cloud.Bucket() as "CoolBucket");
```
eladb, line 32: kind of feels like this should be `pinned()` because it returns the object, right?

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,159 @@
...
+
+# Resource Pinning
+
+- **Author(s):**: @MarkMcCulloh
+- **Submission Date**: 2024-02-15
+- **Stage**: Proposal
+- **Stage Date**: 2024-02-15
+
+Introduces `pin` as a mechanism to protect resources from accidental deletion/moves in the logical application tree.
+
```
eladb, line 14: Please add a "Requirements" section at the top with a simple list of scenarios/use-cases/problems this RFC is trying to solve. It will help us verify that the solution addresses all the scenarios.

(there is a lot of info in the original issue that you can compile into this list).

eladb (changes requested):

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: I was assuming this lockfile could contain information other than pins eventually. That's me trying to future-proof (oops), so it's probably a dumb thing to do.


If this file must be semantically linked, perhaps the API should just be `lock` instead. WDYT?

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: It does return the object, but I think `pin()` is fine. Like `unsafeCast()`. I'm open to either since I can see why `pinned()` would make sense but chose the shorter verb option.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,185 @@
...
+- **Stage Date**: 2024-02-15
+
+Introduce a mechanism to protect resources from accidental deletion/moves in the logical application tree.
+
+## Requirements
+
+### Functional
+- User-ability in sources to signify intent for a resource to not be deleted or moved (e.g. "pinning")
+- Git-trackable way to store changes to pinned resources between compilations, per target platform (e.g. "pinfile")
+- User-ability to manually and declaratively override a pinned resource's path in the resource tree
```
eladb, line 20: I think there's a requirement to be able to control the resource id even if it's not pinned. There could be cases where a library is updated and I want to pin some low level resource within this library to a certain id without having being able to change the library code.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,185 @@
+---
+title: Resource Pinning
+description: Protection of resources from accidental deletion/moves
+---
+
+# Resource Pinning
```
eladb, line 6: I like the term ("pinning").

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,185 @@
...
+- **Author(s):**: @MarkMcCulloh
+- **Submission Date**: 2024-02-15
+- **Stage**: Proposal
+- **Stage Date**: 2024-02-15
+
+Introduce a mechanism to protect resources from accidental deletion/moves in the logical application tree.
+
+## Requirements
+
+### Functional
```
eladb, line 17: Is there a requirement to be able to control the policy? In CFN and TF there's a distinction between "disallow removal" and "leave orphan".

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,185 @@
...
+- User-ability in sources to signify intent for a resource to not be deleted or moved (e.g. "pinning")
+- Git-trackable way to store changes to pinned resources between compilations, per target platform (e.g. "pinfile")
+- User-ability to manually and declaratively override a pinned resource's path in the resource tree
+- Compilation failure when a pinned resource is deleted without explicit removal from the pinfile
+- Compilation failure when a pinned resource is moved without explicit mapping of a previous path to a new path
+- Compilation failure when a resource is no longer pinned in source but is still present in the pinfile
```
eladb, line 23: Resource replacement can be a result of logical id changes but *also* a result of changing certain attributes. This is engine+resource dependent.

I think we need to at least have an idea for how this interplays with the replacement behavior of the underlying provisioning engine.

For example, I would expect (maybe unjustifiably) that if I pin a resource then Terraform or CFN will never replace it, so if I modify an attribute that causes replacement, I would get an error as early as possible.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
eladb: I am wondering if the pinning should be on the id:

```js
new cloud.Bucket() as pinned("CoolBucket");
```

eladb (changes requested):

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: This was my first thought too, but symantically it's a little confusing (is pinned a function called on a string?). I also considered `as pinned "CoolBucket"`. Also I don't know where https://github.com/winglang/wing/issues/4604 is going to end up, which could effect this as well.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: Sure I think this makes sense and I kinda mention this in the out-of-scope section. I think the target can choose its own behavior here. The target itself knows which things are pinned, so it can do this sort of introspection/validation.

Should this be part of this RFC?

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: I'm not sure what this means in the context of the logical tree.

In the context of CFN/TF, I considered platform-level handling out of scope for this RFC. Presumably this is the sort of metadata that can be added to the pinfile next to.

eladb: Yes, sounds like the right direction to put this in the pinfile as additional attributes.

Might want to make sure the schema of this file allows it.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
eladb: I like the suggestion to include this in the pinfile as an attribute. Let's mention this in the RFC.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: Updated to use .pinned syntax

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: > For example, I would expect (maybe unjustifiably) that if I pin a resource then Terraform or CFN will never replace it, so if I modify an attribute that causes replacement, I would get an error as early as possible.

After thinking on this for a while I realized that this error still can't happen at compile time because we can't reliable store a real value for arbitrary attributes.

Consider the example of a bucket. The `name` attribute is likely something you'd want to store to ensure it doesn't changed, because that would cause a replacement. However, the name of a bucket is not always statically known because it could contain something like a token. So there would be nothing to pin without referencing a statefile or somehow hook into deployment.

There are two ways that pinning still overall solves the problem:

1. Provisioning features like `prevent_destroy` aren't perfect, but do totally fix this scenario because it's only a problem with mutation rather than resource orphaning. This means this would still be caught before deployment finishes at the very least. The platform can take care to add this (or equivalent) attribute to all pinned resources.
3. I added a section to the pinfile called `attributes`. This allows you to override the raw platform-level data for a resource. Instead of using this to catch errors, this serves as a way to override values.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: > In CFN and TF there's a distinction between "disallow removal" and "leave orphan".

Does TF have "leave orphan"? I mean you can technically remove a tf resource from the json and the statefile but terraform itself doesn't have a any way to declare that intention AFAIK.

If TF doesn't have it, I think it's okay leaving out for now. We can add it to the pinfile but it would likely not be implemented in the first version.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+```json
+// main.w.pin.json
+{
+  "version": "1",
+  "pinned": {
+    "tf-aws": {
+      "/root/CoolBucket": {
+        "type": "aws_s3_bucket",
```
eladb, line 70: let's use JSII FQNs here:

```suggestion
        "type": "@cdktf/provider-aws.s3Bucket...",
```

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+        "uniqueId": "CoolBucket",
+        "path": "/root/CoolBucket",
```
eladb, line 72: duplicated with the key, no?

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+
+```json
+// main.w.pin.json
+{
+  "version": "1",
+  "pinned": {
+    "tf-aws": {
+      "/root/CoolBucket": {
+        "type": "aws_s3_bucket",
+        "uniqueId": "CoolBucket",
```
eladb, line 71:
```suggestion
        "terraformId": "CoolBucket",
```

I think this should be explicit

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+`wing compile` and `wing test` still work as expected.
+
+`wing compile -t tf-aws` however fails with an error message
+
+```shell
+# (output)
+[-pin] /root/CoolBucket
+[+pin] /root/MyBucket/cloud.Bucket
+
+Error: The following pin has been moved or deleted:
```
eladb, line 108:
```suggestion
Error: The following pinned resource has been moved or deleted:
```

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+  - /root/CoolBucket
+
+If moved, update "path" property in the pinfile to the new path.
+If deleted, remove the entry from the pinfile.
+```
+
```
eladb, line 114:
```suggestion
> In the future we will add support for `wing pin mv` and `wing pin rm`
```

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+For now, Sally wants to avoid the destructive move of the bucket so she updates the `path` in the pinfile.
+
+```json
+// main.w.pin.json
+{
+  "version": "1",
+  "pinned": {
+    "tf-aws": {
+      "/root/CoolBucket": {
```
eladb, line 123: Weird... This should be the new path, no?

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+}
+
+let myBucket = new MyBucket();
+```
+
+Running `wing compile -t tf-aws` automatically updates the pinfile (because the resource is otherwise unchanged):
+
+```shell
+# (output)
+[-pin] /root/CoolBucket
```
eladb, line 157:
```suggestion
[unpinned] /root/CoolBucket
```

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+## Implementation
+
+### pinning
+
+The `pinned` property is used to mark a created resource as "pinned" in the resource tree. This is an inherent public property of all resources. This means that the user does not intend for the resource to be deleted or moved. Pinning is deep, meaning that pinning a parent inherently pins all of its children.
```
eladb, line 164: Let's put this under `node` instead of polluting all objects:

```js
nodeof(this).pinned = false;
```

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+This information will be available to the target platform/app which can choose what to do with this information. For example, the sim target can simply ignore this while the terraform target can manage a pinfile to inform the user when this changes between compilation and allow them to resolve any renames.
+
+### Annotated Pinfile
+
+```json
+{
+  "version": "1",
+  "pinned": {
+    "tf-aws": {
+      // The original logical path to the resource. These will only be leaves of the resource tree
```
eladb, line 175: not sure I understand why we need the "original" logical path? what's the value?

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+
+### pinfile API
+
+Platforms/Apps are provided with a pinfile API. The pinfile itself is a JSON file that contains the current state of the resource tree. The pinfile is created next to the entrypoint file and is named `<entrypoint>.pin.json`. The pinfile API provides a way to add pins to the pinfile, get a list of orphaned pins, and write the current state of the pinfile to disk.
+
+Minimal API:
+
+- `addPin(construct: Construct): void` - Adds a construct to the known pins for the current compilation
```
eladb, line 207: I am missing `app.move(from, to)` here

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+
+Statefulness is a useful concept when thinking about infrastructure but it can serve different purposes.
+The typical way to think about it is in terms of data, like a database or object storage, where the deletion of the infrastructure can be a *permanently destructive* operation. This is unlike something like a cloud Function, which can be deleted and recreated without the Function itself being "different" in any intentional way.
+
+While this is generally true, it's missing nuance. 
+Sometimes your statefulness is intentionally volatile so you don't care about watching it so closely. 
+Sometimes it isn't actually stateful but is treated as so by requiring additional mitigation to avoid downtime/corruption during replacement.
+Designing distributed systems requires balancing the trade-offs of statefulness and statelessness against factors like cost and reliability so it's not always obvious what the right choice is.
+
+With all this said, it's important to give consumers tools to express their intent for pinning even if a resource is typically considered stateful/stateless.
```
eladb, line 221:
```suggestion
With all this said, it's important to give users the tools to express their intent for pinning even if a resource is typically considered stateful/stateless.
```

eladb (dismissed):
a few minor notes

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: I like the simple verb, it's shorter (IMO `unpinned` is kinda an icky looking word) and makes it clear that the CLI just did an action for you. Not a deal breaker for me though.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: Is the JSII type useful here? terraform itself doesn't know what it is

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
eladb: Doesn't make sense to me to keep the original path once the resource is moved. We can change the mapping to be physical -> logical instead of logical -> physical and then it might make more sense.

MarkMcCulloh: Why not both?

> Even if we have tf attribute overrides, having an easy remapping is useful.

I think it's pretty common to want to logically remap, why not allow it alongside physical Instead of forcing physical remappings (there could be multiple for anything derived by the path)

eladb: Logical remapping is a pretty different use case and I "smell" that there are quite a lot of implications to it that we need to think about. I recommend *not* including it in this feature at the moment.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
Chriscbr: Yeah, it feels to me like knowing whether a set of attributes will result in a resource deletion / recreation is a turing-complete problem (it depends on how the underlying resource provider is implemented) so it might be worth leaving these kinds of edge cases to the provisioning engine.

For example, I could change a piece of Terraform config from `prop: "MyBucket12345-Backup" to `prop: "${other_bucket.bucket_name}-Backup" (or change my Wing code to cause that change), and it might _not_ result in a replacement since the Terraform provider can figure out that they resolve to the same string during `terraform plan` / `terraform deploy`. But there's no way the Wing compiler would be able to figure that out.

Providing an escape hatch where the Wing platform provider can include some extra information (a hash, or JSON object) in the pinfile for detecting sensitive changes could be worthwhile sounds ok (though I'd also be okay leaving it out if it seems like a capability we can add on later easily).

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+Introduce a mechanism to protect resources from accidental deletion/moves in the logical application tree.
+
+## Requirements
+
+### Functional
+- User-ability in sources to signify intent for a resource to not be deleted or moved (e.g. "pinning")
+- Provide default for a resource's "pinned" state in it's implementation
+- Git-trackable way to store changes to pinned resources between compilations, per target platform (e.g. "pinfile")
+- User-ability to manually and declaratively override a pinned resource's path in the resource tree
+- User-ability to manually and declaratively override a pinned resource's attributes used by the platform
```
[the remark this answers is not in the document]
MarkMcCulloh: Terraform attributes, or anything similarly low-level. This mechanism for this RFC does need to implemented at the platform level

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+  new() {
+    this.bucket = new cloud.Bucket();
+  }
+}
+
+let myBucket = new MyBucket();
+
+```
+
+`wing compile` and `wing test` still work as expected.
```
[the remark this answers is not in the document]
MarkMcCulloh: Agreed, since this is a platform-level concern I think being able to control it from there makes sense

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: Changed to JSII type to use as a marker

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: Removed this property for now, not sure if it's actually useful

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,239 @@
...
+  new() {
+    this.bucket = new cloud.Bucket();
+    this.bucket.pinned = false;
+  }
+}
+
+let myBucket = new MyBucket();
+```
+
+Running `wing compile -t tf-aws` automatically updates the pinfile (because the resource is otherwise unchanged):
```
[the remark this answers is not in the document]
MarkMcCulloh: If there are no pins left the file will just be deleted. If there were other pins the file just wouldn't have that entry in the map anymore.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: I have now removed any logical-path aspect of this feature

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,246 @@
...
+For now, Sally wants to avoid this destruction so she updates the pinfile to keep attributes the same 
+but updates the key to reflect the new path in the generated terraform.
+
+```json
+// main.w.pin.json
+{
+  "version": "1",
+  "pinned": {
+    "tf-aws": {
+      "aws_s3_bucket.MyBucket_AD8CE4AC": {
```
eladb, line 130: How does wing knows to associate this new name with the old name. In order to preserve the resource, the emitted Terraform must include the old name.

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,246 @@
...
+[-pin] aws_s3_bucket.CoolBucket
+```
+
+At this point, the pinfile is empty and is automatically deleted.
+
+## Implementation
+
+### pinning
+
+The `pinned` property is used to mark a created resource as "pinned" in the resource tree. This api is available on the a resource's Node (`nodeof(this).pinned`). This is an inherent public property of all resource nods. Setting this means that the user does not intend for the resource to be deleted or moved. Pinning is deep, meaning that pinning a parent inherently pins all of its children.
```
eladb, line 175:
```suggestion
The `pinned` property is used to mark a created resource as "pinned" in the resource tree. This api is available on the a resource's Node (`nodeof(this).pinned`). This is an inherent public property of all resource nodes. Setting this means that the user does not intend for the resource to be deleted or moved. Pinning is deep, meaning that pinning a parent inherently pins all of its children.
```

eladb (changes requested):

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
MarkMcCulloh: I can add the old path as a field here. I had that previously in the RFC and you mentioned that wasn't needed so I thought you knew something that I didn't 😁

docs/contributing/999-rfcs/2024-02-15-resource-pinning.md
```diff
@@ -0,0 +1,249 @@
...
+
+```json
+// main.w.pin.json
+{
+  "version": "1",
+  "pinned": {
+    "tf-aws": {
+      "aws_s3_bucket.MyBucket_AD8CE4AC": {
+        "type": "@cdktf/provider-aws.s3bucket.S3Bucket",
+        "originalPath": "aws_s3_bucket.CoolBucket",
```
eladb, line 131: This still feels wrong... I am sorry :-(

The new terraform resource name (`aws_s3_bucket.MyBucket_AD8CE4AC`) is not really something that I care about or understand as a user and not something that I need to carry forward with me.

Wing code "speaks" in logical node path terms. These are the IDs that users understand (and oftentimes assign themselves through `as X`).

I am sorry but I still think that this simply needs to be a map between *logical node paths* and *terraform resource names*.

Happy to talk more about this because it feels like something is misaligned in requirements/use case :-)

I also don't think we need the attributes of the resource here. This will be huge duplication with the Terraform output and I don't see why it is needed.

This is what I think this should look like:

```json
// main.w.pin.json
{
  "version": "1",
  "pinned": {
    "tf-aws": {
      "/MyBucket/Bucket": {
        "terraform_name": "aws_s3_bucket.CoolBucket",
        "type": "@cdktf/provider-aws.s3bucket.S3Bucket"
      }
    }
  }
}
```

MarkMcCulloh: The problem with logical->physical is that 1 logical can have multiple physical, no?
I'm probably mistaken, so I'll change that

> I also don't think we need the attributes of the resource here.

To prevent attributes from changing to avoid resource replacement. I can remove this if it's not needed.

> terraform_name

I think it's useful for the schema of this pinfile, other than the attributes, to be standard across platforms. Especially since it's not difficult to do. If it's not useful I can remove it.

eladb: > The problem with logical->physical is that 1 logical can have multiple physical, no?

I think this should refer to the logical path of the *leaf node*.

> I think it's useful for the schema of this pinfile to be standard across platforms

OK, although I am not sure this is something we need to abstract. This name is target specific. For example, in Terraform, it always has the TF type as a prefix, in CFN there are some characters that are not allowed, in K8S it has a strict length limit. In other targets it might not even be a string (god forbid)...

So from a design perspective, it actually makes more sense to give it a concrete name like `terraform_name` or `terraform_id` instead of trying to generalize.

MarkMcCulloh: > I think this should refer to the logical path of the leaf node.

Yeah that's what I meant. I think platforms can technically emit more than one physical thing given a single logical leaf.

eladb: Eventually, in CDKTF/AWSCDK/CDK8S there's a 1:1 mapping between L1s and resources in the target configuration. That's the "leaf node" that we should pin.

MarkMcCulloh: > Eventually, in CDKTF/AWSCDK/CDK8S there's a 1:1 mapping between L1s and resources in the target configuration

I understand that's how these CDKs work, but it is _technically_ possible for a single construct to spit out multiple things. Or I could even make my own type of CDK and not consider that restriction either. I thought it'd be useful for this pinfile to basically skip any possible abstractions and go straight to the actual output both for the source and destination of the mapping.

That being said, I'll change it to logical -> physical because it's obviously the 0.01% case

btw, are you okay with storing the attributes in the pinfile as well or should that be addressed?

Issue #4843, opened 2023-11-08
Implement updated set literal syntax


Chriscbr:
### Feature Spec

See https://github.com/winglang/wing/pull/4827

### Use Cases

Update compiler

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Chriscbr:
Marking this as a good first issue. The change requires updating a lot of Wing files but doesn't require touching too many deep compiler internals. I think the bulk of the changes need to be made in these files:
- https://github.com/winglang/wing/blob/main/libs/tree-sitter-wing/grammar.js (try combining the `array_literal` and `set_literal` rules into `array_or_set_literal`)
- https://github.com/winglang/wing/blob/main/libs/tree-sitter-wing/test/corpus/expressions.txt
- https://github.com/winglang/wing/blob/main/libs/wingc/src/parser.rs

Warkanlock:
I can take this one

Warkanlock:
> Marking this as a good first issue. The change requires updating a lot of Wing files but doesn't require touching too many deep compiler internals. I think the bulk of the changes need to be made in these files:
> 
>     * https://github.com/winglang/wing/blob/main/libs/tree-sitter-wing/grammar.js (try combining the `array_literal` and `set_literal` rules into `array_or_set_literal`)
> 
>     * https://github.com/winglang/wing/blob/main/libs/tree-sitter-wing/test/corpus/expressions.txt
> 
>     * https://github.com/winglang/wing/blob/main/libs/wingc/src/parser.rs
> 
>     * all wing files in `examples/` that use sets

Thanks for the context! I'll work on this as soon as I have time this week

MarkMcCulloh:
Completed in #5729

PR #5729, opened 2024-02-19
chore(wingc): initial commit to support new syntaxis for set definition

Warkanlock:
Based on #4843 

## Description

Implemented new syntaxis for the `Set` and `MutSet` keywords. From now on, sets will be defined using the identical syntaxes used for Arrays with the difference that there won't be any implicit definition of a set (before it was plain `{}`)

Example

```rust
let set : Set<num> = Set<num> [1,2,3];
```

I limited myself to adding more tests since this was more a functionality related to all the tests we already had in place.

### DIFF
--- a/libs/tree-sitter-wing/grammar.js
+++ b/libs/tree-sitter-wing/grammar.js
@@ -30,7 +30,7 @@ module.exports = grammar({
   precedences: ($) => [
     // Handle ambiguity in case of empty literal: `a = {}`
     // In this case tree-sitter doesn't know if it's a set or a map literal so just assume its a map
-    [$.json_map_literal, $.map_literal, $.set_literal],
+    [$.json_map_literal, $.map_literal, $.array_or_set_literal],
     [$.json_literal, $.structured_access_expression],
   ],
 
@@ -41,7 +41,7 @@ module.exports = grammar({
     // These modifier conflicts should be solved through GLR parsing
     [$.field_modifiers, $.method_modifiers],
     [$.class_modifiers, $.closure_modifiers],
-    [$.inflight_method_signature, $.field_modifiers]
+    [$.inflight_method_signature, $.field_modifiers],
   ],
 
   supertypes: ($) => [$.expression, $._literal],
@@ -54,7 +54,7 @@ module.exports = grammar({
     _semicolon: ($) => choice(";", $.AUTOMATIC_SEMICOLON),
     comment: ($) =>
       token(
-        choice(seq("//", /.*/), seq("/*", /[^*]*\*+([^/*][^*]*\*+)*/, "/"))
+        choice(seq("//", /.*/), seq("/*", /[^*]*\*+([^/*][^*]*\*+)*/, "/")),
       ),
 
     // Identifiers
@@ -62,7 +62,7 @@ module.exports = grammar({
       choice(
         $.nested_identifier,
         $._reference_identifier,
-        $.structured_access_expression
+        $.structured_access_expression,
       ),
 
     identifier: ($) => /([A-Za-z_$][A-Za-z_$0-9]*|[A-Z][A-Z0-9_]*)/,
@@ -75,8 +75,8 @@ module.exports = grammar({
         seq(
           field("object", $._type_identifier),
           // While the final "fields" identifier is optional in this grammar, upstream parsing will fail if it is not present
-          repeat(seq(".", optional(field("fields", $._type_identifier))))
-        )
+          repeat(seq(".", optional(field("fields", $._type_identifier)))),
+        ),
       ),
 
     nested_identifier: ($) =>
@@ -89,13 +89,13 @@ module.exports = grammar({
               $.expression,
               // This is required because of ambiguity with using Json keyword for both instantiation of Json
               // and Identifier for static methods.
-              $.json_container_type
-            )
+              $.json_container_type,
+            ),
           ),
           field("accessor_type", $.accessor),
           // While the "property" identifier is optional in this grammar, upstream parsing will fail if it is not present
-          optional(field("property", $._member_identifier))
-        )
+          optional(field("property", $._member_identifier)),
+        ),
       ),
 
     accessor: ($) => choice(".", "?."),
@@ -131,7 +131,7 @@ module.exports = grammar({
         "bring",
         optional(field("module_name", choice($.identifier, $.string))),
         optional(seq("as", field("alias", $.identifier))),
-        $._semicolon
+        $._semicolon,
       ),
 
     struct_definition: ($) =>
@@ -140,7 +140,7 @@ module.exports = grammar({
         "struct",
         field("name", $.identifier),
         optional(seq("extends", commaSep(field("extends", $.custom_type)))),
-        braced(repeat(field("field", $.struct_field)))
+        braced(repeat(field("field", $.struct_field))),
       ),
     struct_field: ($) =>
       seq(field("name", $.identifier), $._type_annotation, $._semicolon),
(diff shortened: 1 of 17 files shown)


libs/wingc/src/lsp/completions.rs
MarkMcCulloh, line 1: There are several string whitespace changes in this file. Was this done automatically? cargo fmt shouldn't change these since they're inside strings.

Undoing the whitespace changes should fix the tests

Warkanlock: It worked; now: wtf 👀

Warkanlock: Do the tests use the bytecode used by the string relative to and if there is a space will it compute more?

MarkMcCulloh: Yeah the completion tests `//^` to mean "get completions if the cursor were at the position pointed to by `^` above" which uses the actual string data to compute the bytes

See https://github.com/winglang/wing/blob/main/libs/wingc/src/lsp/completions.rs#L1258 and https://github.com/winglang/wing/blob/main/libs/wingc/src/lsp/sync.rs#L259-L274

Warkanlock: this is awesome; thanks so much Mark and sorry for breaking the tests in the first place LOL

examples/tests/invalid/container_types.test.w
MarkMcCulloh, line 1: There are still some sets in this file that use {} that need to updated. Make sure none of the tests are failing due to a parser error.

libs/tree-sitter-wing/grammar.js
```diff
@@ -617,36 +643,28 @@ module.exports = grammar({
...
-    array_literal: ($) =>
-      seq(
-        optional(field("type", $._builtin_container_type)),
-        "[",
-        commaSep(field("element", $.expression)),
-        "]"
-      ),
-    set_literal: ($) =>
+    _collection_literal: ($) => choice($.array_or_set_literal, $.map_literal),
+    array_or_set_literal: ($) =>
```
MarkMcCulloh, line 654: nit: I would just keep this as array_literal, conceptually we don't really have a set literal anymore

libs/tree-sitter-wing/grammar.js
```diff
@@ -679,6 +702,13 @@ function braced(rule) {
   return choice(seq("{", rule, "}"));
 }
 
+/*
+ * @param {Rule} rule
+ */
+function bracketed(rule) {
+  return choice(seq("[", rule, "]"));
```
MarkMcCulloh, line 709: Does this need to wrapped in a choice()? I'm noticing now that `braced()` also uses a seemingly unnecessary choice(). If you remove the choice from this and braced(), do all the tests still pass?

MarkMcCulloh (commented):
Overall looks good, just a couple nits and a fix for the test

libs/tree-sitter-wing/grammar.js
Warkanlock: interesting; I did this rename based on https://github.com/winglang/wing/issues/4843#issuecomment-1866956662

but you have a point that there's not set literal anymore in the codebase

libs/tree-sitter-wing/grammar.js
Warkanlock: Let me check!

examples/tests/invalid/container_types.test.w
Warkanlock: Hard to detect the invalid ones since were failing and passing the test 🤣

libs/tree-sitter-wing/grammar.js
Warkanlock: There's one test that's flaky since it passed OK in my local with the same set of changes, and it's not even related to the `Set` change.

image

image

Warkanlock: Indeed it worked

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1083,7 +1083,14 @@ Duration <DURATION>"
 `;
 
 exports[`container_types.test.w 1`] = `
-"error: Unknown parser error
+"error: Unexpected 'number'
+  --> ../../../examples/tests/invalid/container_types.test.w:3:24
+  |
+3 | let arr2 = Array<num> {1, 2, 3};
+  |                        ^ Unexpected 'number'
```
MarkMcCulloh, line 1090: The snapshots still look off in a few places.

In this case, I think we can remove this since it's specifically related to the {} syntax.

examples/tests/invalid/mut_container_types.test.w
```diff
@@ -7,10 +7,10 @@ let arr5: MutArray<num> = arr4;
 arr1.someMethod();
 
 //Set tests
-let s1 = MutSet<num>{1, "2", 3};
+let s1 = MutSet<num>[1, "2", 3];
 let s2 = MutSet<num>[1, "2", 3];
-let s3: MutSet<num> = {1, 1, 3}; // https://github.com/winglang/wing/issues/1117
-let s4 = MutSet<str>{"hi"};
+let s3: MutSet<num> = [1, 1, 3]; // https://github.com/winglang/wing/issues/1117
```
MarkMcCulloh, line 12:
```suggestion
let s3 = MutSet<num> [1, 1, 3];
```

tools/hangar/__snapshots__/invalid.ts.snap
Warkanlock: Oh, I thought snapshots got updated automatically. I'll do it

examples/tests/invalid/container_types.test.w
```diff
@@ -1,6 +1,6 @@
 //Array tests
 let arr1: Array<num> = [1, "2", 3];
-let arr2 = Array<num> {1, 2, 3};
+let arr2 = Array<num> [1, 2, 3};
```
MarkMcCulloh, line 3:
```suggestion
let arr2 = Array<num> [1, 2, 3];
```

examples/tests/invalid/immutable_container_types.test.w
```diff
@@ -18,7 +18,7 @@ m4.clear();
 // ^^^^^ Unknown symbol "clear"
 
 
-let s1: Set<Array<num>> = MutSet<Array<num>> {[1]};
+let s1: Set<Array<num>> = MutSet<Array<num>>[[1]];
 //                        ^^^^^^^^^^^^^^^^^^^^^^^^ Expected type to be "Set<Array<num>>", but got "MutSet<Array<num>>" instead
 let s2 = Set<str> {"a", "b", "c"};
```
MarkMcCulloh, line 23:
```suggestion
let s2 = Set<str> ["a", "b", "c"];
```

examples/tests/invalid/mut_container_types.test.w
MarkMcCulloh, line 1: In this file, please delete line 3 (`let arr2 = MutArray<num>{1, 2, 3};`)

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2451,6 +2507,20 @@ exports[`json.test.w 1`] = `
...
+   |  
+62 |     \\"1\\" => {[ 
+   | /-----------^
+63 | |     {
+64 | |       foo: \\"bar\\",
+65 | |       stuff: [],
+66 | |     }
+67 | |   //^ Missing required field \\"maybe\\" from \\"StructyJson\\"
+68 | |   ]}
+   | \\\\---^ Unexpected 'array_literal'
```
MarkMcCulloh, line 2521: Looks like a regression. The line in that file probably needs to be changed to this

```wing
let notJsonMissingFieldArray = {
  "1" => Set<Array<StructyJson>> [[ 
    {
      foo: "bar",
      stuff: [],
    }
  //^ Missing required field "maybe" from "StructyJson"
  ]]
};
```

Warkanlock: oh my, how many sets there were

MarkMcCulloh (approved):

PR #5730, opened 2024-02-20
feat(sdk): enable point in time recovery for ddb tables

skorfmann:
This enables [point in time recovery](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/PointInTimeRecovery.html) for DynamoDB tables in non test environments. The reasoning is, that this will create a final backup before destroying the tables, which is not desired in test environments.

I think in general, this is something every table in non test environments should have enabled though. This provides a window of 35 days of recovery. The final backup before deletion of the able is kept around for 35 days as well (at no additional cost). 

If users have more advanced backup needs, they could look into [AWS Backup](https://docs.aws.amazon.com/aws-backup/latest/devguide/whatisbackup.html), but this exceeds the scope of the table resource.

## Checklist

- [ ] Tests added (always)


staycoolcall911:
@skorfmann does it make sense to add that also to the `ex.DynamodbTable` resource?

PR #5732, opened 2024-02-20
fix(vscode): wrong link under goto definition of object property

yoav-steinberg:
The link displayed under "goto definition" of a type/obj property reference included the entire reference instead of just the property:
image
Fixed the return span in these cases. Looks better now.

## Checklist

- [ ] Tests added (always)


MarkMcCulloh (changes requested):
This behavior is intentional. The underline is not the entire reference, it's up-to the current symbol to basically tell you "this is the fully qualified place you're going to" like a regular URL

https://github.com/winglang/wing/assets/1237390/3fd4a1a9-0c15-4236-870f-625eb96d1877

This matches the `hover` highlight as well. Do you think this should be changed?

MarkMcCulloh (approved):

PR #5761, opened 2024-02-23
feat(compiler): class-specific error when accessing missing property

### LINKED ISSUE #2769: Give more specific error when assigning to a field that doesn't exist in a class
### Feature Spec

Currently, if you try assigning to a field in a class like so:

```js
bring cloud;

class MyBucket {
  init() {
  }
}
```

You get an "unknown symbol" error:

```
error: Unknown symbol "bucket"
  --> bla.main.w:5:10
  |
5 |     this.bucket = new cloud.Bucket();
  |          ^^^^^^ Unknown symbol "bucket"
```

Instead, we could make this error more specific:

```
error: Property "bucket" does not exist
  --> bla.main.w:5:10
  |
5 |     this.bucket = new cloud.Bucket();
  |          ^^^^^^ Property "bucket" does not exist
```

### Use Cases

Improve debugging experience

### Implementation Notes

_No response_

### Component

Compiler

### Community Notes

bjrnt:
Fixes #2769.

Given that I know pretty much nothing about compilers, I figured this seemed like the least intrusive change that still solves the problem. However, I can't tell if there are any obvious edge-cases I'm missing here.

Given the following code:

```ts
class MyClass {
  hello() {
    let a = this.myValue;
    this.myValue = 3;
    bob = 2;
    heh();
    this.do();
  }
}
```

The Wing compiler now returns the following errors:
```
Compilation failed with 5 errors
Error at test.w:5:18 | Property "myValue" does not exist
Error at test.w:6:10 | Property "myValue" does not exist
Error at test.w:7:5 | Unknown symbol "bob"
Error at test.w:8:5 | Unknown symbol "heh"
Error at test.w:9:10 | Property "do" does not exist
```

I wasn't sure if the error calling `do` a "property" is what we are after but it seems like the TypeScript compiler returns the same error for both cases, i.e. `Property 'do' does not exist on type 'MyClass'.`, so it shouldn't feel too unfamiliar to people.

## Checklist

- [x] Tests added (always) -- already exists but a different error is returned now

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -5750,6 +5750,10 @@ impl<'a> TypeChecker<'a> {
 			} else {
 				var.clone()
 			}
+		} else if let LookupResult::NotFound(property) = lookup_res {
+			self
+				.spanned_error_with_var(&property, format!("Property \"{property}\" does not exist"))
+				.0
 		} else {
 			self.type_error(lookup_result_to_type_error(lookup_res, property));
 			self.make_error_variable_info()
--- a/libs/wingsdk/src/core/lifting.ts
+++ b/libs/wingsdk/src/core/lifting.ts
@@ -21,7 +21,6 @@ export const INFLIGHT_INIT_METHOD_NAME = "$inflight_init";
  */
 const INFLIGHT_CLOSURE_HANDLE_METHOD = "handle";
 
-
 /**
  * The prefix used to name inflight closure object types.
  */
@@ -358,12 +357,14 @@ export function collectLifts(
  * Returns whether the given item is an inflight closure object.
  */
 function isInflightClosureObject(item: any): boolean {
-  return typeof item === "object" &&
+  return (
+    typeof item === "object" &&
     typeof item.constructor === "function" &&
     typeof item.constructor.name === "string" &&
     item.constructor.name.startsWith(INFLIGHT_CLOSURE_TYPE_PREFIX) &&
     item._liftMap !== undefined &&
     item._liftMap[INFLIGHT_CLOSURE_HANDLE_METHOD] !== undefined
+  );
 }
 
 /**
--- a/libs/wingc/src/jsify/snapshots/fails_if_referencing_unknown_field.snap
+++ b/libs/wingc/src/jsify/snapshots/fails_if_referencing_unknown_field.snap
@@ -2,4 +2,4 @@
 source: libs/wingc/src/jsify/tests.rs
 ---
 ## Errors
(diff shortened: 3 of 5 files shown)


libs/wingsdk/src/core/lifting.ts
yoav-steinberg, line 1: I don't see any real change in this file. Can you remove this from the pr?

libs/wingc/src/type_check.rs
```diff
@@ -5750,6 +5750,10 @@ impl<'a> TypeChecker<'a> {
 			} else {
 				var.clone()
 			}
+		} else if let LookupResult::NotFound(property) = lookup_res {
+			self
+				.spanned_error_with_var(&property, format!("Property \"{property}\" does not exist"))
+				.0
```
yoav-steinberg, line 5756: This will work, but what I don't like about it is that it created `lookup_result_to_type_error` specifically to generate meaningful error's based on a `LookupResult` but here we added an exception and handled the `NotFound` variant in a different way than what's currently implemented in `lookup_result_to_type_error`.
I think a better approach would be to extend how `lookup_result_to_type_error` handles `NotFound` conditions so it knows if the looked up symbol is a `ClassLike` property or not and produce the correct diagnostic.
So what I'd try is to add another argument to `lookup_result_to_type_error` which is the environment (`env`) in which the lookup happened. If that env turns out to be a class env we can say this is a property lookup error, and if it's not a class env then we can say this is an "unknown symbol". Makes sense?
Do you want to try this?

yoav-steinberg (changes requested):
Thanks, nice work. But I'd attempt a slightly different approach.

libs/wingsdk/src/core/lifting.ts
bjrnt: I believe this was added by the self mutation. I'll see if I can get rid of it with a rebase.

libs/wingc/src/type_check.rs
bjrnt: Makes sense! I was considering modifying `lookup_result_to_type_error` but decided against it since I figured it'd be a larger change. I agree that doing so would be a cleaner solution though. I'll give it a try.

bjrnt: I tried this approach but wasn't able to make any progress getting the information out of `env`. At this point, it'd probably take less time to write the code than to guide me to a solution, so I'll close this PR and stick to the areas I understand better.

yoav-steinberg: Ok. Hope it wasn't too frustrating. I'll take a look. Thanks a lot for giving it a try!

PR #5763, opened 2024-02-23
fix(console): openBrowser function not working as expected

skyrpex:
The previous implementation had references to unexisting variables, and also an applescript file was missing.

Now, browser tabs will be reused if possible.


apps/wing-console/console/app/scripts/openChrome.applescript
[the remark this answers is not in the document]
skyrpex: Yeah, you're right. Any suggestions for a directory? `assets`, `src/assets`?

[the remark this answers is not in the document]
skyrpex: For sanity reasons, in `assets` since we're bundling whatever is in `src` so the applescript won't be there anymore.

apps/wing-console/console/app/src/index.ts
```diff
@@ -144,5 +145,12 @@ export const createConsoleApp = async (options: CreateConsoleAppOptions) => {
       set(key, value) {},
     },
   });
+
+  if (options.open) {
+    // @ts-ignore
```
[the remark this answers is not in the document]
skyrpex: It was necessary when it really was a JS file, since TS would complain about the import being a JS...

apps/wing-console/console/app/scripts/open.mjs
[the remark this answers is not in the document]
skyrpex: Done.

apps/wing-console/console/app/scripts/dev.ts
MarkMcCulloh, line 1: Is it possible to use `.mts` for these scripts?

skyrpex: I'd love to use ESM everywhere but it becomes a problem when trying to import `src/open.ts`, which is a CJS file.

MarkMcCulloh: I thought you can `require` a cjs file from esm?

MarkMcCulloh (approved):

apps/wing-console/console/app/assets/openChrome.applescript
```diff
@@ -0,0 +1,95 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file at
+https://github.com/facebookincubator/create-react-app/blob/master/LICENSE
+*)
+
+property targetTab: null
+property targetTabIndex: -1
+property targetWindow: null
+property theProgram: "Google Chrome"
```
Chriscbr, line 12: What happens if I'm not using Chrome as my default browser or don't have it installed?

skyrpex: It goes back to the default behavior. In my case, I use Arc and does not actually reuse the browser tab (sadly).

apps/wing-console/console/app/scripts/dev.ts
skyrpex: You're right. My head is just exhausted 🤣

Issue #5754, opened 2024-02-22
Template "empty" is not available in wing.


sharuuu:
### I tried this:

wing new empty

### This happened:

Template "empty" is not available in wing. Please let us know you'd like to use this template in wing by opening an issue at https://github.com/winglang/wing/issues/!

### I expected this:

initilize a new project

### Wing Version

0.58.23

### Node.js Version

v21.6.2

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

hasanaburayyan:
hmm, with wing version `0.58.23` and node: `21.6.2` I am unable to reproduce.

monadabot:
Congrats! :rocket: This was released in [Wing 0.59.10](https://github.com/winglang/wing/releases/tag/v0.59.10).

PR #5780, opened 2024-02-26
fix(cli): template X is not available in wing

hasanaburayyan:
If a user's installation of wing somehow resulted in file the installation files not being owned by the user, there is a chance that the user running wing commands does not have full access to the template files. 

 For example running `sudo npm install -g winglang` would result in all the package files being owned by root, which the default would mean other users have read access but not write. For template quickstarts we should not care if the user has `write` access the important bit is just `read` so this addresses that issue

Closes: https://github.com/winglang/wing/issues/5754

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing/src/commands/init.ts
+++ b/apps/wing/src/commands/init.ts
@@ -1,7 +1,7 @@
 // This file and the main function are named "init" instead of "new"
 // to avoid a conflict with the "new" keyword in JavaScript
 import { exec } from "child_process";
-import { existsSync } from "fs";
+import { existsSync, constants } from "fs";
 import { copyFile, mkdir, readFile, readdir, writeFile } from "fs/promises";
 import { join, relative } from "path";
 import { promisify } from "util";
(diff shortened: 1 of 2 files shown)


apps/wing/src/commands/init.ts
```diff
@@ -99,7 +99,7 @@ export async function init(template: string, options: InitOptions = {}): Promise
   options.template = template;
   // Check if the template exists for the selected language
   const templatePath = join(PROJECT_TEMPLATES_DIR, language, template);
-  const templateExists = await exists(templatePath);
+  const templateExists = await exists(templatePath, constants.F_OK | constants.R_OK); // eslint-disable-line no-bitwise
```
MarkMcCulloh, line 102: Is just `R_OK` sufficient? I thought R_OK is both "it exists and is readable", but worth checking that

MarkMcCulloh (approved):

Issue #3390, opened 2023-07-12
Run tests with globally install `winglang` in CI


MarkMcCulloh:
To catch certain module resolution issues, it's better to test a full workflow via `npm i -g winglang`.

To do this locally and in CI, we need to make use of [verdaccio](https://verdaccio.org/) again. We should find a way to do this with as little overhead as possible

eladb:
Since this is an incident follow up, usually the best practice is to tag it as p1.

Let's also reference the incident itself.

PR #5786, opened 2024-02-27
fix: unable to resolve tokens inflight from awscdk platform

### LINKED ISSUE #5510: Resolution of tokens in the inflight function on the awscdk platform fails
### I tried this:

I created a website and tried to retrieve its URL in a function

```ts
bring cloud;

let website = new cloud.Website(path: "./website");

new cloud.Function(inflight() => {
  log(website.url);
});
```

### This happened:

When I execute the command:

```bash
$ CDK_STACK_NAME=my-stack wing compile main.w -t @winglang/platform-awscdk
$ cdk deploy --app target/main.awscdk/ 
```

the stack is created, but when I deploy, my function's print is this:
`2024-01-23T14:00:44.969Z	d94beaea-c655-47f9-8dba-a026e7dd4336	INFO	https://${Token[TOKEN.90]}`

### I expected this:

print the correct url

### Is there a workaround?

No

### Anything else?

This line adds the `CdkToken` class to the `registerTokenResolver`. The issue is that the `CdkToken` class is in the `awscdk` project, while the `registerTokenResolver` is implemented in `wingsdk`.

I believe the `registerTokenResolver` is not able to handle a class that does not exist in the project where it is defined.

https://github.com/winglang/wing/blob/ded4e13bfb6b221e44d96f53019c38237cfff61a/libs/awscdk/src/app.ts#L109

### Wing Version

0.54.14

### Node.js Version

v18.16.0

### Platform(s)

Linux

### Community Notes

MarkMcCulloh:
Fixes #5510

The awscdk platform was importing `registerTokenResolver()` directly from `@winglang/sdk/lib/core/tokens`. This was bypassing preflight logic to always force imports from the SDK to resolve from the same place because it was only checking for imports specifically from `@winglang/sdk` only. This caused it to load it as a different module which had a global that was unused by the rest of the SDK logic.

I was able to replicate the original issue and this fixed it in that same scenario.

Automated testing this is very difficult/slow because it requires a certain scenario that is very realistic: A global install of the SDK and a "local" install of a platform like awscdk. When you install `@winglang/platform-awscdk` it comes with its own copy of `@winglang/sdk`. In our hangar tests this gets deduped because they are installed alongside each other. Adding a way to test this is a pretty big tasks and should at least involve https://github.com/winglang/wing/issues/3390 as well

### DIFF
--- a/libs/wingcompiler/src/compile.ts
+++ b/libs/wingcompiler/src/compile.ts
@@ -317,13 +317,19 @@ async function runPreflightCodeInWorkerThread(
 ): Promise<void> {
   try {
     // Create a shimmed entrypoint that ensures we always load the compiler's version of the SDK
+    const sdkEntrypoint = require.resolve("@winglang/sdk");
     const shim = `\
 var Module = require('module');
 var original_resolveFilename = Module._resolveFilename;
-var WINGSDK_PATH = '${normalPath(require.resolve("@winglang/sdk"))}';
+var WINGSDK_PATH = '${normalPath(sdkEntrypoint)}';
+var WINGSDK_DIR = '${normalPath(join(sdkEntrypoint, "..", ".."))}';
 
 Module._resolveFilename = function () {
-  if(arguments[0] === '@winglang/sdk') return WINGSDK_PATH;
+  const path = arguments[0];
+  if(path === '@winglang/sdk') return WINGSDK_PATH;
+  if(path.startsWith('@winglang/sdk')){
+    arguments[0] = path.replace('@winglang/sdk', WINGSDK_DIR);
+  }
   return original_resolveFilename.apply(this, arguments);
 };
 
--- a/libs/wingcompiler/src/compile.ts
+++ b/libs/wingcompiler/src/compile.ts
@@ -321,14 +321,15 @@ async function runPreflightCodeInWorkerThread(
     const shim = `\
 var Module = require('module');
 var original_resolveFilename = Module._resolveFilename;
+var WINGSDK = '@winglang/sdk';
 var WINGSDK_PATH = '${normalPath(sdkEntrypoint)}';
 var WINGSDK_DIR = '${normalPath(join(sdkEntrypoint, "..", ".."))}';
 
 Module._resolveFilename = function () {
   const path = arguments[0];
-  if(path === '@winglang/sdk') return WINGSDK_PATH;
-  if(path.startsWith('@winglang/sdk')){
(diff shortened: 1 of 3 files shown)


libs/wingcompiler/src/compile.ts
```diff
@@ -317,13 +317,20 @@ async function runPreflightCodeInWorkerThread(
...
+var WINGSDK_PATH = '${normalPath(sdkEntrypoint)}';
+var WINGSDK_DIR = '${normalPath(join(sdkEntrypoint, "..", ".."))}';
 
 Module._resolveFilename = function () {
-  if(arguments[0] === '@winglang/sdk') return WINGSDK_PATH;
+  const path = arguments[0];
+  if(path === WINGSDK) return WINGSDK_PATH;
+  if(path.startsWith(WINGSDK)){
+    arguments[0] = path.replace(WINGSDK, WINGSDK_DIR);
+  }
```
hasanaburayyan, line 333: I love and hate this.

Its like is this "clever as fuck" or "hacky as fuck"... probably someone in the middle "works as fuck" 😁

turbo.json
```diff
@@ -11,6 +11,7 @@
     "tools/bump-pack/**",
     "!tools/bump-pack/node_modules/**"
   ],
+  "globalEnv": ["CI"],
```
hasanaburayyan, line 14: Why is this needed?

hasanaburayyan (approved):

libs/wingcompiler/src/compile.ts
hasanaburayyan: Feels like we might want to advertise this to potential platform builders

turbo.json
MarkMcCulloh: hangar makes use of this env variable to determine if it uses symlinks or if it'll install via tarballs. This ended up not mattering for testing this PR but it made sense to keep around anyways.

I should add this to hangar's config rather than the global one, so I'll change that.

libs/wingcompiler/src/compile.ts
MarkMcCulloh: This is a super hacky version of a [normal thing](https://nodejs.org/api/module.html#customization-hooks) that isn't stable yet

MarkMcCulloh: Ideally we should use the builtin thing or look into a more overhauled system of dealing with modules. e.g. I've been thinking about integrating something like https://github.com/unjs/jiti

Issue #5788, opened 2024-02-27
tsconfig compilerOptions not respected for @wingcloud/framework project


MarkMcCulloh:
### I tried this:

Add compilerOptions setting to tsconfig.json for a project using @wingcloud/framework

### This happened:

Setting is not used in generated output

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.59.15](https://github.com/winglang/wing/releases/tag/v0.59.15).

PR #5789, opened 2024-02-27
fix: compilerOptions for typescript project not respected

MarkMcCulloh:
Fixes #5786

There are a few options that are still not supported because it's either necessary (like `outDir`) or difficult (like `paths`). The user will be warned if they provide these options.

Also removed usage of `baseUrl` here because it's not needed and was causing a confusing error in some cases (if you have a file that has the same name as a node_module).


libs/@wingcloud/framework/src/internal.ts
```diff
@@ -6,32 +6,80 @@ export interface CompileOptions {
...
+        `Failed to parse tsconfig at ${tsconfigPath}: ${errors
+          .map((e) => e.messageText)
+          .join("\n")}`
+      );
+    }
+
+    for (const managedOption of MANAGED_TSCONFIG_OPTIONS) {
+      if (config?.compilerOptions?.[managedOption] !== undefined) {
+        console.warn(
+          `'${managedOption}' in tsconfig is managed by wing, value is ${compilerOptions[managedOption]}`
```
eladb, line 73: Maybe add the expected value?

eladb (approved):
Maybe add a wee test for this logic

staycoolcall911:
> Fixes #5786

@MarkMcCulloh - did you mean "Fixes #5788" in. the PR description?

libs/@wingcloud/framework/src/internal.ts
MarkMcCulloh: Changed the wording to show what the forced options are. The goal is that users shouldn't/can't provide them at all.

Issue #5794, opened 2024-02-28
add user_resource_interaction analytics event to Wing Console


ainvoner:
### Use Case

today we have analytics events for resources usage.
those events are being fired by both user interactions (clicking on resources actions in the Console) and by actions that triggered by other resources in my Wing application (for example: timer/scheduler that calls a function invoke)

we would like to keep track on only user interactions in order to measure their actual usage in the Console (and not their application behavior).

### Proposed Solution

add a new user_resource_interaction event for every actual user interaction with a resource.
this event will have the resource name and the action that was used for this resource.

### Implementation Notes

add a middleware to our Console tRPC router to catch only user interactions

### Component

Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

monadabot:
Congrats! :rocket: This was released in [Wing 0.59.28](https://github.com/winglang/wing/releases/tag/v0.59.28).

PR #5795, opened 2024-02-28
chore(console): add user resource interaction analytics event

ainvoner:
resolves: https://github.com/winglang/wing/issues/5794

## Checklist

- [ ] Tests added (always)


apps/wing-console/console/server/src/utils/createRouter.ts
```diff
@@ -120,6 +125,22 @@ const invalidateQueriesAfterMutation = middleware(async (options) => {
...
+      if (meta?.resource && meta?.action) {
+        analytics.track("user_resource_interaction", {
+          resource: meta.resource,
+          action: meta.action,
+        });
+      }
+    }
+    return await options.next();
+  },
+);
```
skyrpex, line 142: Instead of using placing `meta` properties on every route, maybe we could use `options.path`?

apps/wing-console/console/server/src/utils/createRouter.ts
```diff
@@ -76,6 +76,11 @@ export interface FileLink {
   column?: number;
 }
 
+export interface RouterMeta {
+  action?: string;
+  resource?: string;
+}
```
skyrpex, line 82: What about explicitly stating that this meta property is about analytics?

```suggestion
export interface RouterMeta {
  analytics?: {
    action: string;
    resource: string;
  };
}
```

apps/wing-console/console/server/src/utils/createRouter.ts
skyrpex: NVM, I think this is better: https://github.com/winglang/wing/pull/5795/files#r1506168813

apps/wing-console/console/server/src/utils/createRouter.ts
```diff
@@ -76,6 +76,13 @@ export interface FileLink {
   column?: number;
 }
 
+export interface RouterMeta {
+  analytics?: {
+    action?: string;
+    resource?: string;
+  };
+}
+
```
skyrpex, line 85: Does it make sense to allow undefined `action` and `resource` if `analytics` is defined? I think this better:

```suggestion
export interface RouterMeta {
  analytics?: {
    action: string;
    resource: string;
  };
}

```

apps/wing-console/console/server/src/utils/createRouter.ts
```diff
@@ -120,6 +127,22 @@ const invalidateQueriesAfterMutation = middleware(async (options) => {
...
+    const analytics = options.ctx.analytics;
+    if (analytics) {
+      const analyticsMeta = options?.meta?.analytics;
+      if (analyticsMeta && analyticsMeta?.resource && analyticsMeta?.action) {
+        analytics.track("user_resource_interaction", {
+          resource: analyticsMeta.resource,
+          action: analyticsMeta.action,
+        });
+      }
+    }
```
skyrpex, line 141: And then:

```suggestion
    const analytics = options.ctx.analytics;
    const meta = options.meta?.analytics;
    if (analytics && meta) {
      analytics.track("user_resource_interaction", {
        resource: meta.resource,
        action: meta.action,
      });
    }
```

skyrpex (changes requested):

skyrpex (approved):

PR #5807, opened 2024-02-29
fix(compiler): unknown property diagnostic is same as unknown symbol 

### LINKED ISSUE #2769: Give more specific error when assigning to a field that doesn't exist in a class

yoav-steinberg:
Fixes #2769
A revisit of #5761 
`x_instance.unkonwn_prop` resulted in:
```
Unknown symbol "unkonwn_prop"
```
And now results in:
```
Property "unkonwn_prop" doesn't exist in "X"
```

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/invalid/immutable_container_types.test.w
+++ b/examples/tests/invalid/immutable_container_types.test.w
@@ -1,7 +1,7 @@
 let m1 = Map<str>{"a" => "hi"};
 
 m1.set("a", "bye");
-// ^^^ Property "set" doesn't exist in "Map" (TODO: better error message https://github.com/winglang/wing/issues/1660)
+// ^^^ Member "set" doesn't exist in "Map" (TODO: better error message https://github.com/winglang/wing/issues/1660)
 
 let m2: Map<str> = MutMap<str> {};
 //                 ^^^^^^^^^^^^^^ Expected type to be "Map<str>", but got "MutMap<str>" instead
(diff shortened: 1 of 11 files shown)


examples/tests/invalid/immutable_container_types.test.w
```diff
@@ -1,33 +1,33 @@
 let m1 = Map<str>{"a" => "hi"};
 
 m1.set("a", "bye");
-// ^^^ Unknown symbol "set" (TODO: better error message https://github.com/winglang/wing/issues/1660)
+// ^^^ Property "set" doesn't exist in "Map" (TODO: better error message https://github.com/winglang/wing/issues/1660)
```
Chriscbr, line 4: Since `set` refers to a method (a function value) and not a property (a non-function value) the message feels off to me here. Perhaps we can change the error based on whether it's a method or property?

Or we can fall back to using the general term "member"

libs/wingc/src/type_check/symbol_env.rs
```diff
@@ -114,7 +114,8 @@ pub enum LookupResult<'a> {
 	/// A matching symbol was found but it's not public
 	NotPublic(reference([a], [SymbolKind]), SymbolLookupInfo),
 	/// The symbol was not found in the environment, contains the name of the symbol or part of it that was not found
-	NotFound(Symbol),
+	/// If the lookup environment was a type environment, the type is also included
+	NotFound(Symbol, Option<TypeRef>),
```
Chriscbr, line 118: Maybe clarify in the comment with an example of a lookup in a type environment and a lookup not in a type environment

Chriscbr (approved):
nice improvement!

eladb (approved):

examples/tests/invalid/immutable_container_types.test.w
yoav-steinberg: I think the general "member" is good enough here. I'll make the fix.

libs/wingc/src/type_check/symbol_env.rs
yoav-steinberg: added some wording, hope this helps

PR #5808, opened 2024-02-29
feat(console): restrict dragging the map when using the space key

skyrpex:
The changes in this commit restrict dragging the map using the space key. When the space key is pressed, the map becomes draggable and the user can move it around. The cursor also changes to a grab cursor when the space key is pressed. This provides a more intuitive way for users to interact with the map.


polamoros (approved):
Very cool!

PR #5821, opened 2024-02-29
feat(sdk): in-place app updates in simulator

eladb:
In order to improve iteration speed when updating an application, introduce a new simulator API to support in-place updates of an app in the simulator.

It is now possible to use `simulator.update(simDir)` without needing to stop the simulator and start a new instance. This method will analyze the diff between the currently loaded app and the new app defined in `simDir` and will create/update/delete resources accordingly.

This implies that if a resource was not updated, it won't be restarted across app updates.

The way this works is that when `update(simDir)` is called, we read the new `simulator.json` file and analyze the diff between the current resource configuration and the new one to determine which resources were *added* (exists only in the new config), *removed* (doesn't exist in the new config), *updated* (an attribute was changed) or *retained* (no change). Then, we:

1. Stop all *deleted* and *updated* resources.
2. Copy the state from all the *retained* resources (attributes and resolved properties).
3. Start all *new* and *updated* resources.

Modeled dependencies via an execution graph. This allows us to automatically replace any dependent resources and start in topological order.

Until we can deprecate `sim.State` (waiting for #5882), we sill have an initialization loop which allows late resolution of tokens.

Next step is to update the Wing Console to use this new API when a new version of the app is compiled.

### Future improvements

- Currently all updates are always a replacement (stop and start), but technically we can
  give resources an opportunity to do an in-place update (basically call `resource.update()`).
- We can add support for watching the sim directory and automatically 
  updating the app inside the simulator. Wow!

### Misc

The `api.cleanup()` method did not wait for the server to close.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -996,3 +1055,61 @@ export interface SimulatorServerResponse {
   /** The error that occurred during the method call. */
   readonly error?: any;
 }
+
+function planUpdate(current: BaseResourceSchema[], next: BaseResourceSchema[]) {
```
MarkMcCulloh, line 1059: Does this need to handle dependencies? Like, if A depends on B and A needs to be updated, doesn't that mean B may need to be updated too?

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -341,6 +368,35 @@ export class Simulator {
...
+    const resourceConfig = this.getResourceConfig(path);
+    let event: Trace = {
+      type: TraceType.RESOURCE,
+      data: { message: `${resourceConfig.type} deleted.` },
+      sourcePath: resourceConfig.path,
+      sourceType: resourceConfig.type,
+      timestamp: new Date().toISOString(),
+    };
+
+    this._addTrace(event);
```
skyrpex, line 397:
```suggestion
    this._addTrace({
      type: TraceType.RESOURCE,
      data: { message: `${resourceConfig.type} deleted.` },
      sourcePath: resourceConfig.path,
      sourceType: resourceConfig.type,
      timestamp: new Date().toISOString(),
    });
```

skyrpex (commented):
This looks very good! Can't wait to use it.

libs/wingsdk/src/simulator/simulator.ts
eladb: Yes, absolutely and definitely. I felt I missed something and originally thought about this as an issue but missed it.

The challenge here is that our simulator doesn't have a static dependency model.

Let me see what I can come up with.

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -291,6 +263,82 @@ export class Simulator {
...
+        // if we've tried too many times, give up (might be a dependency cycle or a bad reference)
+        if (next._attempts === 0) {
+          throw new Error(
+            `Could not start resource ${next.path} after ${START_ATTEMPT_COUNT} attempts. This could be due to a dependency cycle or an invalid attribute reference.`
+          );
+        }
+
+        // put back in the queue for another round
+        initQueue.push(next);
+      }
```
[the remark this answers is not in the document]
eladb: Yes, definitely makese sene to model this statically and then we won't need this reconsiliation loop. Both approaches are used in IaC (K8S doesn't have a static model either).

I am debating whether to refactor this right now or later. Not sure yet.

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -178,6 +186,125 @@ test("provides raw tree data", async () => {
   expect(treeData).toMatchSnapshot();
 });
 
+describe("in-place updates", () => {
```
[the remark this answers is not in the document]
eladb: +1

libs/wingsdk/src/simulator/simulator.ts
eladb: OK, added an execution graph. Should cover this.

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -291,6 +263,82 @@ export class Simulator {
...
+
+    const plan = planUpdate(
+      this._model.config.resources,
+      newModel.config.resources
+    );
+
+    // stop all *deleted* and *updated* resources
+    for (const c of [...plan.deleted, ...plan.updated]) {
+      await this.stopResource(c);
+    }
```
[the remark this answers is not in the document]
eladb: Done. `stopResource` will now stop all dependencies first.

libs/wingsdk/src/simulator/simulator.ts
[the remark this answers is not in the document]
eladb: Done.

libs/wingsdk/src/simulator/graph.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export interface Definition {
+  path: string;
+  deps?: string[];
+  props?: Record<string, any>;
+}
+
+class Node<T extends Definition> {
+  public readonly dependencies: Node<T>[] = [];
+  public readonly dependents: Node<T>[] = [];
```
MarkMcCulloh, line 11: Should these be `Set`s?

libs/wingsdk/src/simulator/graph.ts
```diff
@@ -0,0 +1,139 @@
...
+  public find(path: string): Node<T> | undefined {
+    return this.byPath[path];
+  }
+
+  public toposort(): Array<Array<string>> {
+    const result: Array<Array<string>> = [];
+    const pending = new Set(this.nodes.map((x) => x.path));
+    const started = new Set<string>();
+
+    let i = 0;
```
MarkMcCulloh, line 67: What is the purpose of `i`?

libs/wingsdk/src/simulator/graph.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    while (pending.size > 0 && i++ < 10) {
+      // create an array of all the nodes that are ready to be started. this means that either they have
+      // no dependencies or all of their dependencies have already been started.
```
MarkMcCulloh, line 71: The comments in this function should probably be revised since this function isn't "starting" anything

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -216,20 +468,8 @@ function removeLineNumbers(line?: string) {
...
-    let lastLine = lines.findIndex((line) =>
-      line?.includes("Simulator.runTest")
-    );
-    if (lastLine !== -1) {
-      lines = lines.slice(0, lastLine + 1);
-    }
-
-    error = lines.join("\n");
+    // take only the first line
+    error = result.error.split("\n")[0];
```
MarkMcCulloh, line 472: nit:

```suggestion
    error = result.error.message;
```

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -40,6 +38,13 @@ export interface SimulatorProps {
    * resources
    */
   readonly factory?: ISimulatorFactory;
+
+  /**
+   * Determins if resources should be started concurrently or not. Normally we would enable this in
+   * interactive environments where determinsm is not important.
+   * @default false
+   */
+  readonly concurrency?: boolean;
```
MarkMcCulloh, line 47: Now that we have dependencies, I'm not sure I see the harm in always doing concurrent starts. If the concern is that tests may fail, I would argue the tests are poorly written if they rely on the order in which resources start.

"start" in the simulator is like "deploy", and other provisioning engines deploy things concurrently when possible. They usually have a concurrently limit and maybe a way to opt out but that's more of a non-functional concern (e.g. API rate limits)

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -291,6 +271,76 @@ export class Simulator {
...
+      }
+    }
+  }
+
+  /**
+   * Updates the running simulation with a new version of the app. This will create/update/delete
+   * resources as necessary to get to the desired state.
+   * @param simDir The path to the new version of the app
+   */
+  public async update(simDir: string) {
```
MarkMcCulloh, line 294: Maybe this can be saved for a future optimization, but I feel like both `stop` and `update` can use similar logic as `start`. i.e. utilize topological order and concurrency.
In the case of `stop`, it'd be a reverse toposort

libs/wingsdk/src/simulator/graph.ts
eladb: Yes sure

libs/wingsdk/src/simulator/graph.ts
eladb: lol debugging...

Will remove

libs/wingsdk/src/simulator/graph.ts
eladb: Yah

libs/wingsdk/src/simulator/simulator.ts
eladb: We can make this true by default but pragmatically writing tests that are robust to concurrency is beyond the scope of this PR.

MarkMcCulloh: Are you talking about unit tests in wingsdk or tests users write in wing?

libs/wingsdk/src/simulator/simulator.ts
eladb: Both "update" and "delete" are using topological order implicitly.

Update calls startResources which starts in order.

Stop first stops it's dependents which implies it doesn't in reverse topo order

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -996,3 +965,61 @@ export interface SimulatorServerResponse {
...
+
+  // everything left in "current" is to be deleted
+  for (const config of Object.values(currentByPath)) {
+    deleted.push(config.path);
+  }
+
+  return { added, updated, deleted, retain };
+}
+
+function resourceByPath(list: BaseResourceSchema[]): {
```
eladb, line 1012: Really not needed

libs/wingsdk/src/simulator/simulator.ts
eladb: Unit tests in Wing SDK

libs/wingsdk/src/simulator/simulator.ts
MarkMcCulloh: True, the order is correct so I moreso mean the concurrency aspect, which only startResources is making use of

eladb: I'll add a note

libs/wingsdk/src/simulator/simulator.ts
MarkMcCulloh: Fair enough, I think it's something would should do at some point but I'm cool with keeping it out of this PR

libs/wingsdk/src/simulator/graph.ts
```diff
@@ -0,0 +1,139 @@
...
+  props?: Record<string, any>;
+}
+
+class Node<T extends Definition> {
+  public readonly dependencies: Node<T>[] = [];
+  public readonly dependents: Node<T>[] = [];
+  constructor(public readonly def: T) {}
+
+  public get path() {
+    return this.def.path;
```
eladb, line 15: Maybe we just need to store the path here instead of the entire object.

libs/wingsdk/test/simulator/__snapshots__/simulator.test.ts.snap
```diff
@@ -91,17 +91,7 @@ exports[`run single test > happy path 1`] = `
...
-    at Object.exports.handler ([abs])
-    at [abs]
-    at Script.runInContext (node:vm:<sanitized>)
-    at Object.runInContext (node:vm:<sanitized>)
-    at [src]/shared/sandbox.ts:<sanitized>
-    at new Promise (<anonymous>)
-    at Sandbox.call ([src]/shared/sandbox.ts:<sanitized>)
-    at Object.withTrace ([src]/simulator/simulator.ts:<sanitized>)
-    at TestRunner.runTest ([src]/target-sim/test-runner.inflight.ts:<sanitized>)",
+  "error": undefined,
```
eladb, line 94: Oops

MarkMcCulloh: My bad! I've made this mistake before so you'd think I'd remember that error is just a string...

eladb: I'll just use the first line

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -291,6 +271,76 @@ export class Simulator {
...
+
+    for (const wave of execution) {
+      if (this.concurrency) {
+        await Promise.all(wave.map((r) => this.startResource(r)));
+      } else {
+        for (const r of wave) {
+          await this.startResource(r);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
eladb: Thanks. I think for now I will simply remove the parallelism and we can add this later.

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -178,6 +186,250 @@ test("provides raw tree data", async () => {
   expect(treeData).toMatchSnapshot();
 });
 
+describe("in-place updates", () => {
```
Chriscbr, line 189: Add a test that checks retained resource state is retained (i.e. api.url should stay the same if the resource doesn't have to be stopped or updated)

libs/wingsdk/src/simulator/simulator.ts
eladb: I decided to get rid of the concurrency in the PR (both for "start" and "stop"). I'll leave it for a future optimization.

So basically, removed the `toposort` method and in `startResource`, I am first making sure all dependencies are started (simply recursive DFS).

Chriscbr (approved):

libs/wingsdk/src/simulator/graph.ts
eladb: Got rid of this method

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -996,3 +965,61 @@ export interface SimulatorServerResponse {
...
+      added.push(nextConfig.path);
+      continue;
+    }
+
+    // the resource is already in "current", if it's different from "next", it means it was updated
+    const state = (r: BaseResourceSchema) =>
+      JSON.stringify({
+        props: r.props,
+        type: r.type,
+      });
```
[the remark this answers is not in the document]
eladb: The dependent *are* taken into account but not here. If a resource is stopped (either because of an update or a delete), all of it's dependents are also stopped. Makes sense?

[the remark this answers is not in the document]
eladb: Added a comment.

libs/wingsdk/test/simulator/graph.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+test("fails on an indirect cyclic dependency", () => {
+  expect(() => {
+    new Graph([
+      { path: "a", deps: ["b"] },
+      { path: "b", deps: ["c"] },
+      { path: "c", deps: ["a"] },
+    ]);
+  }).toThrowError(/cyclic dependency detected: c -> a/);
+});
```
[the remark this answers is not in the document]
eladb: done.

libs/wingsdk/src/simulator/simulator.ts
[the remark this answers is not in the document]
eladb: Removed. Keeping this PR contained.

libs/wingsdk/test/target-sim/state.test.ts
```diff
@@ -5,7 +5,7 @@ import { Testing } from "../../src/simulator";
 import { State } from "../../src/target-sim";
 import { SimApp } from "../sim-app";
 
-test("state can be resolved at any time", async () => {
+test.skip("state can be resolved at any time", async () => {
```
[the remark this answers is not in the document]
eladb: I basically think this capability doesn't make sense now because we are moving from a dynamic dependency resolution model to a static one and this is basically testing the dynamic model.

Are you okay if we removed this?

On a different note, I think we need a refactor around `cloud.Service`, `State`, etc. It's a bit of a mess right now.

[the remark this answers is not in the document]
Chriscbr: Yeah, I haven't pieced together yet exactly how `State` should fit into all of this either, so that's cool with me 👍

libs/wingsdk/src/simulator/simulator.ts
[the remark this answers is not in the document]
Chriscbr: yes, that makes sense.

I only want to mention that `cloud.Deploy` has a semantics of "run this code every time my app is updated / I do a new deployment" and I'm not sure if `simulator.update()` is respecting this

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -178,6 +186,252 @@ test("provides raw tree data", async () => {
...
+      updated: ["root/Api1"],
+    });
+
+    expect(simTraces(sim)).toEqual([
+      "root/Api1 started",
+      "root/Api1/Endpoint started",
+      "root/Bucket1 started",
+      "in-place update",
+      "root/Api1/Endpoint stopped",
+      "root/Bucket1 stopped",
```
eladb, line 426: Add a test where a token value is resolved differently after update

libs/wingsdk/test/target-sim/state.test.ts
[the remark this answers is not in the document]
eladb: Yes, this `State` thing is horrific. We need to get rid of it. You were right!

libs/wingsdk/test/simulator/simulator.test.ts
eladb: done

libs/wingsdk/src/simulator/simulator.ts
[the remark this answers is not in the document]
eladb: Ok, node deps are now taken into account (added a test).

libs/wingsdk/test/target-sim/state.test.ts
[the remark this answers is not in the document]
eladb: Removed this test altogether for now.

libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -131,8 +130,13 @@ export class Api
 
   public async cleanup(): Promise<void> {
     this.addTrace(`Closing server on ${this.url}`);
-    this.server?.close();
-    this.server?.closeAllConnections();
```
eladb, line 135: @Chriscbr fyi (really hard to test but I ran into this)

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -996,3 +987,66 @@ export interface SimulatorServerResponse {
...
+    for (const resource of list) {
+      if (ret[resource.path]) {
+        throw new Error(
+          `unexpected - duplicate resources with the same path: ${resource.path}`
+        );
+      }
+      ret[resource.path] = resource;
+    }
+    return ret;
+  }
```
Chriscbr, line 1051: Super nit but if we are going to model dependencies then we could update the format of simulator.json to store the resources in an object/map instead of as an array like it is today

Chriscbr (approved):

libs/wingsdk/src/target-sim/api.inflight.ts
Chriscbr: More stability! 🙌  nice fix

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -178,6 +190,410 @@ test("provides raw tree data", async () => {
...
+    const bucket1 = new Bucket(app, "Bucket");
+    bucket1.addObject("url.txt", api1.url);
+
+    const stateDir = mkdtemp();
+    const sim = await app.startSimulator(stateDir);
+
+    const urlBeforeUpdate = await sim.getResource("root/Bucket").get("url.txt");
+
+    // remove the state directory otherwise Api reuses the port
+    fs.rmdirSync(stateDir, { recursive: true });
```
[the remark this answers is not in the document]
eladb: Cool beans

libs/wingsdk/src/simulator/simulator.ts
eladb: I rather not break the schema at this point :-)

libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -131,8 +130,13 @@ export class Api
...
+    return new Promise((resolve, reject) => {
+      this.server?.closeAllConnections();
+      this.server?.close((err) => {
+        if (err) return reject(err);
+        else return resolve();
+      });
```
MarkMcCulloh, line 138: I think the server should be closed before closeAllConnections.
.close() only prevents new connections.

eladb: done

skyrpex (approved):

PR #5823, opened 2024-02-29
Update eslint and enable `@typescript-eslint/consistent-type-imports`

skyrpex:
From `@typescript-eslint/consistent-type-imports` docs:

> TypeScript allows specifying a type keyword on imports to indicate that the export exists only in the type system, not at runtime. This allows transpilers to drop imports without knowing the types of the dependencies.


Chriscbr (approved):

MarkMcCulloh:
It's probably worthwhile to run `turbo default --concurrency 1` at the root to invoke projen across the repo to see if any changes get undone. Then `turbo lint` after that as well.

The github workflow will do all this too, but the iteration speed will be much faster locally

eladb (approved):

skyrpex:
Closing in favor of a more conservative and minimal approach: #5824.

PR #5825, opened 2024-03-01
fix: change the type of allowOrigin to string (#5741)

### LINKED ISSUE #5741: `Access-Control-Allow-Origin` must contain only one origin
### I tried this:

I created a new cloud api as follows:

```
let api = new cloud.Api(
  cors: true,
  corsOptions: {
    allowOrigin: ["https://example.com", "http://localhost:3000"],
  },
);
```

### This happened:

Then I tried to fetch my API endpoint, but I got the following CORS error.

```
Access to fetch at 'https://xyz.com/api' from origin 'http://localhost:3000' has been blocked by CORS policy:
Response to preflight request doesn't pass access control check:
The 'Access-Control-Allow-Origin' header contains multiple values, https://example.com,http://localhost:3000, but only one is allowed.
Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

### I expected this:

I expected the server to compare the `Origin` header to each of the origins in the `allowOrigin` array and to return the origin which matched. If none of the origins matched, then we can return an arbitrary origin. For example, we can always return the first origin in the array. If there are no origins in the `allowOrigin` array, then we should return `*`. Note that we should not return `null` because that's a security vulnerability.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

Here's a table that summarizes the behavior that I expected.

| `allowOrigin` | `Origin` Request Header | `Access-Control-Allow-Origin` Response Header |
|---|---|---|
| `["https://example.com", "http://localhost:3000"]` | `https://example.com` | `https://example.com` |
| `["https://example.com", "http://localhost:3000"]` | `http://localhost:3000` | `https://localhost:3000` |
| `["https://example.com", "http://localhost:3000"]` | `https://evil.com` | `https://example.com` |
| `[]` | `http://localhost:3000` | `*` |

### Is there a workaround?

I can manually add an additional `OPTIONS` endpoint for each of my API endpoints. However, that's cumbersome.

### Anything else?

I found the code which sets the `Access-Control-Allow-Origin` header.

https://github.com/winglang/wing/blob/aaace84a45b9abb867b4e195185d7dc3085a120f/libs/wingsdk/src/cloud/api.ts#L513-L524

As you can see, we are setting `Access-Contol-Allow-Origin` to `allowOrigin.join(",") || ""` which is clearly wrong because `Access-Control-Allow-Origin` should be a single origin, not multiple origins.

### Wing Version

0.58.8

### Node.js Version

v21.5.0

### Platform(s)

Linux

### Community Notes

aaditmshah:
Fixes #5741

The `Access-Control-Allow-Origin` header must contain only one origin. If we provide an array with more than one origin, then the origins are joined with commas. This causes the CORS preflight to fail because the browser expects only one origin. The proper solution to this problem would be to select one of the origins from the list. However, as a temporary fix, I changed the type of `allowOrigin` from `Array<string>` to simply `string`. The expectation is that the developer should only provide one origin. This prevents the CORS preflight from failing. However, it also disallows developers from having more than one origin.

*By submitting this pull request, I confirm that my contribution is made under the terms of the [Wing Cloud Contribution
License](https://github.com/winglang/wing/blob/main/CONTRIBUTION_LICENSE.md).*

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/docs/04-standard-library/cloud/api.md
+++ b/docs/docs/04-standard-library/cloud/api.md
@@ -521,7 +521,7 @@ let ApiCorsOptions = cloud.ApiCorsOptions{ ... };
 | <code><a href="#@winglang/sdk.cloud.ApiCorsOptions.property.allowCredentials">allowCredentials</a></code> | <code>bool</code> | Whether to allow credentials. |
 | <code><a href="#@winglang/sdk.cloud.ApiCorsOptions.property.allowHeaders">allowHeaders</a></code> | <code>MutArray&lt;str&gt;</code> | The list of allowed headers. |
 | <code><a href="#@winglang/sdk.cloud.ApiCorsOptions.property.allowMethods">allowMethods</a></code> | <code>MutArray&lt;<a href="#@winglang/sdk.cloud.HttpMethod">HttpMethod</a>&gt;</code> | The list of allowed methods. |
-| <code><a href="#@winglang/sdk.cloud.ApiCorsOptions.property.allowOrigin">allowOrigin</a></code> | <code>MutArray&lt;str&gt;</code> | The list of allowed allowOrigin. |
+| <code><a href="#@winglang/sdk.cloud.ApiCorsOptions.property.allowOrigin">allowOrigin</a></code> | <code>str</code> | The allowed origin. |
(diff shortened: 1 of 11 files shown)


staycoolcall911:
Assigning @tsuf239 as a reviewer for this PR.

tsuf239 (approved):
great work!

Issue #5819, opened 2024-02-29
Add commands for both Windows and Linux users


ayush2390:
### Use Case

It would be great if you can add commands both for Windows and Linux users in your documentation making it easier for the developers to know which command they have to run depending upon their machine and not get stuck.

For Eg-

For Windows users:

```
set BROWSER=none
npm run start
```

For Linux/Mac users:

```
BROWSER=none wing run backed/main.w
```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

tsuf239:
Just adding some more context, the issue raised when we tried to run this code on a windows machine: 
![image](https://github.com/winglang/wing/assets/39455181/4df7d689-3a13-4122-99ef-81166d7360f0)
because of those lines: https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-sim/react-app.ts#L18-L22

PR #5826, opened 2024-03-01
fix(sdk): `ex.ReactApp` start command with PORT env

meirdev:
Fix for #5819,

Adding an environment variable in Windows is different than Linux/Mac, the command:

`PORT=3001 npm run start`

Doesn't work, Windows assumes `PORT` is a command.

Instead, we can add the env with the [child_process.exec](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback) function and the `options.env` parameter.

_Note: `child_process.exec` is cross-platform but I haven't tested it on Windows._


tsuf239 (commented):
It looks great!
I added a flag trying to run the tests on windows, waiting for the results :)

tsuf239 (approved):

tsuf239:
@meirdev Can you please try to update from main? maybe it will help with the build error?

meirdev:
> @meirdev Can you please try to update from main? maybe it will help with the build error?

Still failing... although it now fails exactly where I changed, I see no reason why. This is also the same error that appeared earlier on the `redis.test.ts`.

eladb (approved):

PR #5865, opened 2024-03-07
feat(sdk): expose `endpoint` for `sim.DynamodbTable`

eladb:
Allow obtaining the endpoint used by the local DynamoDB service. This is needed in order to be able to create clients.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-sim/dynamodb-table.inflight.ts
+++ b/libs/wingsdk/src/target-sim/dynamodb-table.inflight.ts
@@ -30,6 +30,7 @@ export class DynamodbTable
     process.env.WING_DYNAMODB_IMAGE ?? "amazon/dynamodb-local:2.0.0";
   private readonly context: ISimulatorContext;
   private client?: DynamoDBClient;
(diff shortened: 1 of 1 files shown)


libs/wingsdk/src/target-sim/dynamodb-table.inflight.ts
```diff
@@ -51,14 +52,16 @@ export class DynamodbTable
         containerPort: "8000",
       });
 
+      this._endpoint = `http://0.0.0.0:${hostPort}`;
```
MarkMcCulloh, line 55: Why don't we use 127.0.0.1 or localhost here?

MarkMcCulloh (approved):

libs/wingsdk/src/target-sim/dynamodb-table.inflight.ts
eladb: I am not sure... That's how it's currently set up.

PR #5867, opened 2024-03-07
feat(sdk): cloud.Function concurrency for `sim` target

### LINKED ISSUE #168: Function `concurrency` prop - for the `sim` target
To support a `concurrency` option we need to implement it across all of the polycon implementations.

Adding support for `aws.Function` is simple, it should just be a single line of terraform config.

To add support for `sim.Function`, we need to model the option in our simulator. The simulator is single-threaded, but it's still possible for multiple inflight functions to technically be "active" by virtue of JavaScript's [event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop). To enforce concurrency constraints (e.g. `concurrency: 1` means no more than one inflight function can ever be active at a time) we need to track how many functions are active etc. and delay function invocations until there is available concurrency capacity.

### LINKED ISSUE #5549: Inflight handler caching allow shared mutable state between parallel function invocations
### I tried this:

```js
bring cloud;
bring util;

class Foo {
  inflight var n: num;
  inflight new() {
    this.n = 99;
  }

  pub inflight inc(): num {
    this.n += 1;
    return this.n;
  }

  pub inflight get(): num {
    return this.n;
  }
}

let foo = new Foo();

// a function that takes at least three seconds to run
  let n = foo.inc();
  util.sleep(3s);
  assert(n == foo.get());
});

test "Foo state is not shared between function invocations" {
  // start two invocations of fn, staggering them by 1 second
  fn.invokeAsync("");
  util.sleep(1s);
  fn.invoke("");
}
```

### This happened:

The test fails

### I expected this:

In clouds, FaaS invocations receive their own runtime environments where code is loaded into and executed. After an invocation is finished, it's possible the same loaded code will be reused -- but it's not possible for a single runtime environment to be used by two parallel invocations.

The Wing local simulator is violating this expectation, making it appear as if you can seamlessly share mutable state between concurrent `cloud.Function` invocations.

### Is there a workaround?

_No response_

### Anything else?

Regression from https://github.com/winglang/wing/pull/5478

### Wing Version

0.54.63

### Node.js Version

20.9.0

### Platform(s)

MacOS

### Community Notes

### LINKED ISSUE #5864: Inflight function clients are never cached
### I tried this:

```js
bring cloud;
bring expect;
bring util;

class Foo {
  inflight var n: num;
  inflight new() {
    this.n = 99;
  }

  pub inflight inc(): num {
    this.n += 1;
    return this.n;
  }
}

let foo = new Foo();

  let n = foo.inc();
  return "{n}";
});

test "single instance of Foo" {
  let x = fn.invoke("");
  let y = fn.invoke("");

  expect.equal(x, "100");

  log("{y}");
}
```

### This happened:

In the unit test, it always prints "100"

### I expected this:

Since the two invocation requests happen sequentially (one after the other), in most cloud providers the same container might be reused between the invocations to save the effort of spinning up a second container to run the user code. In such a case, the code's state from the previous request would be reused, and 101 would be printed in the example above. The simulator is not doing this.

### Is there a workaround?

_No response_

### Anything else?

Regression from #5554

Related to https://github.com/winglang/wing/issues/5549 and https://github.com/winglang/wing/issues/3244

### Wing Version

0.59.45

### Node.js Version

0.20.11

### Platform(s)

MacOS

### Community Notes

Chriscbr:
This PR implements a more precise system in the local simulator for simulating inflight function concurrency in cloud.Function, with the goal of unblocking support for function initialization caching (for example, to cache database connections between multiple requests).

Now, the simulation manages a pool of workers, each of which executes inflight code in an isolated Node.js process. Function initialization is performed once on a worker, and can then handle multiple sequential requests thereafter. If a worker is busy handling an existing request, then a new worker is spun up as long as it doesn't exceed the cloud.Function's configured `concurrency`. If a function times out, the worker process has to be stopped, so any initialization state is lost.

This diagram illustrates what could happen when your function receives 10 requests:

As your function receives more concurrent requests, the simulator scales up the number of execution environment instances in response.

Closes #168
Closes #5864
Closes #5549

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -9,63 +13,178 @@ import {
...
-        });
-      },
-    });
+    this.maxWorkers = props.concurrency;
+
+    this.createBundlePromise = this.createBundle();
+  }
+
+  // Create a bundle with the code that will be run by all workers.
+  private async createBundle() {
```
Chriscbr, line 39: We're now creating multiple sandboxes (one per worker), but we only need to do the code bundling once since all workers are executing the same. So I've moved this responsibility from the `Sandbox` class to the `FunctionClient` class.

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -9,63 +13,176 @@ import {
...
+    }
   }
 
   public async init(): Promise<FunctionAttributes> {
     return {};
   }
 
   public async cleanup(): Promise<void> {
-    await this.sandbox.cleanup();
+    await this.createBundlePromise;
```
MarkMcCulloh, line 83: Is this `await` needed?

examples/tests/valid/inflight_handler_singleton.test.w
```diff
@@ -34,9 +38,27 @@ test "single instance of Foo" {
   let z = fn2.invoke("");
 
   expect.equal(x, "100");
+
+  // the simulator intentionally reuses the sandbox across invocations
+  // but we can't trust that this will always happen on the cloud
```
MarkMcCulloh, line 43: Is the plan to eventually simulate the typical behavior of cloud providers where workers are occasionally evicted? It would be unfortunate to give users the false sense that they can rely on effectively stateful `cloud.Function`s.

libs/wingsdk/src/target-sim/function.ts
```diff
@@ -31,6 +32,7 @@ export class Function extends cloud.Function implements ISimulatorResource {
 
     // props.memory is unused since we are not simulating it
     this.timeout = props.timeout ?? Duration.fromMinutes(1);
+    this.concurrency = props.concurrency ?? 10;
```
MarkMcCulloh, line 35: I would expect the default max concurrency to be huge, like 100 or 1000. Actually I'd expect it to be "until you run out of memory" but setting some kind of limit is probably good.

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -9,63 +13,176 @@ import {
...
+      return newWorker;
+    }
+
+    return undefined;
+  }
+
+  private initWorker(): Sandbox {
+    if (!this.bundledFile) {
+      throw new Error("Bundle not created");
+    }
```
MarkMcCulloh, line 165: What if we centralize the bundling await here?

```suggestion
    await this.createBundlePromise;
```

libs/wingsdk/src/cloud/function.md
```diff
@@ -56,12 +56,12 @@ new cloud.Function(inflight () => {
 
 ## Function container reuse
 
-Most cloud providers will opportunistically reuse the function's container in additional invocations. It is possible
-to leverage this behavior to cache objects across function executions using `inflight new` and inflight fields.
+Most cloud providers will opportunistically reuse the function's container in additional invocations.
+It is possible to leverage this behavior to cache objects across function executions using `inflight new` and inflight fields.
```
MarkMcCulloh, line 60: imo it feels like a bad practice to encourage this way of thinking unless `cloud.Function` is intentionally designed as a "Stateful Function`.

examples/tests/sdk_tests/function/concurrency.test.w
```diff
@@ -0,0 +1,44 @@
...
+    c.inc();
+    util.sleep(5s);
+  }, concurrency: 1) as "concurrency fn";
+  
+  test "f1 concurrency limit reached" {
+    f1.invokeAsync();
+    try {
+      f1.invoke();
+    } catch e {
+      assert(e.contains("Too many requests, the function has reach its concurrency limit"));
```
eladb, line 19:
```suggestion
      assert(e.contains("Too many requests, the function has reached its concurrency limit"));
```

examples/tests/sdk_tests/function/concurrency.test.w
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  let q = new cloud.Queue();
+
+  q.setConsumer(inflight (message: str) => {
+    util.sleep(1s);
+    c.inc();
+  }, concurrency: 1, batchSize: 1);
+
+  test "queue applies backpressure to functions with limited concurrency" {
```
eladb, line 34: Beautiful

examples/tests/valid/inflight_handler_singleton.test.w
eladb: Eviction is a resource optimization. Ideally we should simulate the fact that when a function is not executing, it's process should be paused (`SIGSTOP`?) and resumed when another invocation needs to be performed.

libs/wingsdk/src/cloud/function.ts
```diff
@@ -40,6 +40,12 @@ export interface FunctionProps {
    * @default 30
    */
   readonly logRetentionDays?: number;
+
+  /**
+   * The maximum concurrent invocations that can run at one time.
+   * @default - platform specific limits
```
eladb, line 46: maybe worth mentioning the default in the simulator here?

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: Kind of feels like this is not coupled with `cloud.Function` because it will be needed by anyone who uses the sandbox. Perhaps make this a static method of `Sandbox` (`Sandbox.createBundle()`)?

And make this function return something like `SandboxBundle` which will be the input to the sandbox. No?

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: yeah looks off

libs/wingsdk/src/target-sim/function.inflight.ts
```diff
@@ -9,63 +13,176 @@ import {
...
   }
 
   public async save(): Promise<void> {}
 
   public async invoke(payload: string): Promise<string> {
     return this.context.withTrace({
       message: `Invoke (payload=${JSON.stringify(payload)}).`,
-      activity: () => this.sandbox.call("handler", payload),
+      activity: async () => {
+        await this.createBundlePromise; // ensure inflight code is bundled before doing any work
```
eladb, line 95: feels like this should be handled within the sandbox somehow... Maybe `SandboxBundle` holds this promise and the sandbox awaits it automatically?

libs/wingsdk/src/target-sim/function.inflight.ts
eladb: I think inside `Sandbox`.

libs/wingsdk/src/target-sim/function.ts
eladb: AWS has a pretty big default limit. I think 100 should be fine for the simulator.

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -166,6 +174,14 @@ export class Queue
         void fnClient
           .invoke(JSON.stringify({ messages: messagesPayload }))
           .catch((err) => {
+            // If the function is at a concurrency limit, pretend we just didn't call it
+            if (
+              err.message ===
+              "Too many requests, the function has reach its concurrency limit."
```
eladb, line 180:
```suggestion
              "Too many requests, the function has reached its concurrency limit."
```

libs/wingsdk/src/target-tf-gcp/function.ts
```diff
@@ -53,6 +54,12 @@ export class Function extends cloud.Function {
     // app is a property of the `cloud.Function` class
     const app = App.of(this) as App;
 
+    if (props.concurrency != null) {
+      throw new NotImplementedError(
+        "Function concurrency prop isn't implemented yet on the current target."
```
eladb, line 59:
```suggestion
        "Function 'concurrency' prop isn't implemented yet for GCP."
```

eladb (approved):
Great job. This is great!

examples/tests/valid/inflight_handler_singleton.test.w
MarkMcCulloh: Reusing processes is also a resource optimization.

My point is that sharing memory between invocations is not something users should be relying on.

libs/wingsdk/src/target-sim/function.inflight.ts
Chriscbr: Yeah, I can see how this looks odd. I think there's a good reason in this scenario, but let me know what you think.

The reason I added this was to fix a rare bug that was causing `wing test` to crash sometimes. Here's what was happening: 
- `wing test` is executed with cleanup enabled (the default option). The app being tested has a cloud.Function which is not invoked during the test.
- The simulator starts, runs the tests, and stops shortly thereafter.
- When the simulator is cleaning up each resource, `cleanup()` is called on the cloud.Function.
- Since `cloud.Function` starts bundling work asynchronously during the constructor, it's possible the bundling won't be finished when the resource is being cleaned up.
- All resources are stopped, the simulator stops, and `wing test` cleans up files like the `.wing` directory etc.
- The bundling work is eventually executed, but it can't find a file, and a gross error is printed to the terminal.

There isn't really a way to cancel promises you've created in Node.js, so the cleanest solution seemed like just to wait for the work to finish in all code paths.

Anyways -- I'll add a comment to the code :-)

apps/wingcli-v2/src/main.rs
```diff
@@ -145,6 +145,8 @@ fn run_javascript_node(source_file: &Utf8Path, target_dir: &Utf8Path, target: Ta
 	command.env("WING_PLATFORMS", target.to_string());
 	command.env("WING_SOURCE_DIR", source_dir);
 	command.env("WING_SYNTH_DIR", target_dir);
+
+	tracing::info!("Running command: {:?}", command);
```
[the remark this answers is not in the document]
Chriscbr: Yes but I want to keep this one (it's a helpful trace, won't be printed unless a certain flag is passed) :D

libs/wingsdk/src/cloud/function.md
Chriscbr: Hmm, yeah that's a good point. Maybe this deserves some more discussion (@ShaiBer also raised some questions about whether this makes sense to have as a functional concern in the cloud.Function API)

libs/wingsdk/src/target-sim/function.inflight.ts
Chriscbr: Yes that sounds like a good improvement

libs/wingsdk/src/target-sim/queue.inflight.ts
```diff
@@ -149,6 +149,14 @@ export class Queue
         if (!fnClient) {
           throw new Error("No function client found");
         }
+
+        // If the function we picked is at capacity, keep the messages in the queue
+        const hasWorkers = await (fnClient as any).hasAvailableWorkers();
```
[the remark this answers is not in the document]
Chriscbr: Yep, I just want be careful since I don't really want the code for the sim impl to be bundled / treated as a real dependency of this file. But I think `import type` might do the trick...

examples/tests/valid/inflight_handler_singleton.test.w
Chriscbr: For local testing I'm not sure if modeling eviction is that important, but if we're hosting apps using the simulator on Wing Cloud then I could see some eviction policy having merit. Sounds like a good idea for a future improvement

libs/wingsdk/src/cloud/function.ts
Chriscbr: OK

libs/wingsdk/src/target-sim/function.ts
Chriscbr: Ok - updated the default to 100

examples/tests/sdk_tests/function/concurrency.test.w
Chriscbr: Oops, thanks

libs/wingsdk/src/target-sim/function.inflight.ts
Chriscbr: I think you're right... I moved over this responsibility of waiting for the bundle to be available from `invoke` to `initWorker`, it should be a little bit better

PR #5882, opened 2024-03-11
feat(sdk): `sim.Resource`

eladb:
## Checklist

- [ ] Tests added (always)


eladb:
@Chriscbr is picking this up 🙏

eladb:
@Chriscbr [this issue](https://github.com/winglang/winglibs/pull/115) made me think that perhaps the API we are designing for `sim.Resource` is not good enough.

Now that cloud functions are running in a child process (Good Thing :tm:) means that the inflight API of the resource must traverse the process boundary.

So, in the WebSocket example, the `inflight sendMessage()` method cannot call `ws.send()` directly but must send a request to the child process that runs the server. The PR above implements this manually, but I am thinking that this is basically what we are doing with the proxy server to implement all the `cloud.Xxx` APIs, so kind of feels like there is some general mechanism that we can build that will make this seamless to implement.

(as long as inflight APIs are serializable, which is something we must enforce sooner rather than later because it had already bit me in the ass...)


The API we propose for `sim.Resource` will make it easy to miss this because it will give people the false sense that the inflight APIs are running in the same process as the `sim.Resource`... Anyway -- something here is not figured out and feels like we should go back to the drawing board here...

eladb:
A thought -- I am worried about overloading the execution model of inflight methods here. It will be very confusing if only in this particular case inflight methods are remote calls...

Feels like the "backend" implementation of the resource should be modeled as an `inflight class` and `sim.Resource` can provide the facilities for remote calls (and maybe that's something we can generalize even further):

```js
pub interface IResource {
  inflight onStart(): void;
  inflight onStop(): void;
}
```

```js
inflight class CounterResource impl sim.IResource {
  var counter: num;

  new() {
    this.counter = 0;
  }

  pub onStart() {
    // startup code
  }

  pub onStop() {
    // shutdown code
  }

  pub inc(n: num?): num {
    let p = this.counter;
    this.counter += (n ?? 1);
    return p;
  }

  pub peek(): num {
    return this.counter;
  }
}

class Counter {
  backend: sim.Resource;

  new() {
    // this is a "backend factory". it returns an inflight class that implements the
    // resource.
    let factory = inflight (): sim.IResource => {
      return new CounterResource();
    };

    this.backend = new sim.Resource(factory);
  }

  pub inflight inc(n: num?): num {
    let response = this.backend.call("inc", [n]);
    return num.fromStr(response);
  }

  pub inflight peek(): num {
    let response = this.backend.call("peek");
    return num.fromStr(response);
  }
}
```

We've been talking about supporting remoting at the language level at some point, but this requires further research and design :-)

eladb:
Closing for now. Someone will pick this up soon.

Issue #5758, opened 2024-02-22
Wing Libraries Can Include Implicit Platform Extensions


hasanaburayyan:
### Use Case

Wing libraries now support the ability to include a `platform.js` file that will be implicitly used whenever the library is brought.

For instance if a library supports multiple target implementations for a resource, then it can include a `platform.js` file that implements new instance based on the target.
```js
// platform.js

class LibPlatform {
	new instance(type, scope, id, props) {
		// determine concrete type
	}
}
```

This platform will be implicitly applied to the list of platforms for compilation.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

hasanaburayyan:
Im going to add some better use cases to the description for now I just needed to make the issue since I keep forgetting 😁

monadabot:
Congrats! :rocket: This was released in [Wing 0.60.9](https://github.com/winglang/wing/releases/tag/v0.60.9).

PR #5886, opened 2024-03-11
feat(platforms): support implicit platform files

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/5758

## Summary
This PR introduces the the loading of implicit platform files from either the app's source directory or any brought modules.

I.E. 

```
src/
├── main.w
└── wplatform.js
```

will result in `wplatform.js` being implicitly read as a platform

as well 

```
src/
├── main.w
├── lib/
│   ├── lib.w
│   └── wplatform.js
└── lib2/
    ├── lib.w
    └── wplatform.js
```
with a wing file like:
```js
bring "./lib" as myLib;
bring "./lib2" as myOtherLib;
```

will result in two platforms being read in.

This behavior also carries over for libraries, thus if a wing library contains a `wplatform.js` it will be implicitly loaded in when brought.

### Implications of ordering
Since platform order matters, the ordering I choose was to append implicit platforms (starting with source dir) to any explicitly defined platforms.

I.E. if a users runs `wing compile -t tf-aws -t custom-1` within a project structure like this:
```
src/
├── main.w
├── wplatform.js
└── lib/
    ├── lib.w
    └── wplatform.js
```

Then internally the order of platforms are: `tf-aws, custom-1, src/wplatform.js, src/lib/wplatform.js` 

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -21,6 +21,42 @@ export class PlatformManager {
...
+   * These are platforms that are not explicitly declared in the cli options
+   * but are implicitly available to the app.
+   *
+   * We look for platforms in the following locations:
+   * - The source directory
+   * - Any imported namespaces (provided by the wingc compiler output)
+   *
+   * To determine if a directory contains a platform, we check if it contains a file named "platform.js"
+   * TODO: Support platforms defined in Wing (platform.w) https://github.com/winglang/wing/issues/4937
+   */
```
hasanaburayyan, line 39: oops I still need to document this

libs/wingcompiler/src/compile.ts
```diff
@@ -164,8 +164,7 @@ export async function compile(entrypoint: string, options: CompileOptions): Prom
   if (!existsSync(synthDir)) {
     await fs.mkdir(workDir, { recursive: true });
   }
-
-  let preflightEntrypoint = await compileForPreflight({
+  let compileForPreflightResult = await compileForPreflight({
```
eladb, line 167: nit: Any reason these are not `const`?

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -21,6 +21,42 @@ export class PlatformManager {
...
+   * Retrieve all implicit platform declarations.
+   *
+   * These are platforms that are not explicitly declared in the cli options
+   * but are implicitly available to the app.
+   *
+   * We look for platforms in the following locations:
+   * - The source directory
+   * - Any imported namespaces (provided by the wingc compiler output)
+   *
+   * To determine if a directory contains a platform, we check if it contains a file named "platform.js"
```
eladb, line 37: I am wondering if we should use a more distinct name such as `wing-platform.js` (or `w-platform.js`, `wing_platform.js`, `platform.wing.js`, `wp.js`, `_wing_platform.js`).

@MarkMcCulloh what do you think?

libs/wingcompiler/src/compile.ts
hasanaburayyan: 😁 Yea I cant every remember what language Im in. I think const is fine here

libs/wingsdk/src/platform/platform-manager.ts
MarkMcCulloh: I think it'd be useful follow whatever convention we would use if these were wing files.
Other than that I don't feel super strong about a certain direction, but if we follow the theme of `main` and `test` then we'd have `platform`:

```
platform.w
whatever.platform.w

platform.js
whatever.platform.js
```

hasanaburayyan: Ill update the convention to 
```
platform.js
*.platform.js
```

eladb: Since this is implicitly loading these files I wanted to avoid people accidentally creating a `platform.js` or `platform.w` file and struggling to understand what the F...

Also, it's important that when people see this file they immediately understand its meaning.

Something I like about @MarkMcCulloh's suggestion is that it will be possible to put multiple platform files per directory, and that's required because a platform extension only applies to a single platform model.

So maybe just use `wplatform.js` as the extension:

```
tfaws.wplatform.js
awscdk.wplatform.js
wplatform.js
redis.awscdk.wplatform.js
```

A bit of a mouthful but better than hijacking "platform" :-)

hasanaburayyan: > A bit of a mouthful but better than hijacking "platform" :-)

😁 I wish I had a better suggestion that didnt go against the grain of the other existing conventions. Ill update to go with `wplatform` and learn to love it.

MarkMcCulloh: What if we always require at least two dots? `*.platform.js`. I believe this is how `test` files work as well.

I feel like it'd be hard to hit accidentally. Honestly I suspect platform files are going to be extremely common so even with `platform.js` this won't be an issue for users to understand.

Since we're likely going to be experimenting with more verbose CLI output over time, it'd be reasonable to also inform the users of which platforms are loaded which would further ease any confusion.

Sorry if it feels like I'm belaboring this point, I'm okay with `wplatform` to avoid more naming bikeshedding, but figured I'd finish the braindump.

docs/docs/02-concepts/03-platforms.md
```diff
@@ -38,6 +38,16 @@ The order in which platforms are evaluated is important.
 
 The first platform in the list is the primary platform, it is responsible for providing the Wing compiler with the base App that will be used to determine how resources are created, as well it will also lay the ground work for what target the rest of the platforms will need to be compatible with.
 
+#### Using Implicit Platforms
```
eladb, line 41:
```suggestion
#### Implicit Platforms
```

docs/docs/02-concepts/03-platforms.md
```diff
@@ -38,6 +38,16 @@ The order in which platforms are evaluated is important.
...
+
+Additionally, you can use naming conventions to implicitly define platforms that should be used. These platform files can be located in the root of your project or in a library that your project uses. The naming convention is as follows:
+```sh
+wplatform.js
+*.wplatform.js
+```
+
+For example, if you have a file named `custom.wplatform.js` in the root of your project, it will automatically be added to the list of platforms to be used when compiling your application. Its also important to note that implicit platforms are always loaded after the platforms specified in the `--platform` option.
+
```
eladb, line 50: Maybe add an example to show why this is useful. The original issue has some use case I think.

eladb (approved):

docs/docs/02-concepts/03-platforms.md
eladb: The description is this pull request has some nice documentation

libs/wingsdk/src/platform/platform-manager.ts
eladb: I think it makes sense to also support `wplatform.js` (without a prefix).

But @MarkMcCulloh if you feel strongly, I am also okay with just supporting `*.platform.js`.

@hasanaburayyan your call...

hasanaburayyan: lets roll with `wplatform` and `*.wplatform` for now, it grew on me overnight.

PR #5889, opened 2024-03-11
feat(compiler)!: stringify enums to their own names

Chriscbr:
It's common to want to be able to stringify enums in some way. While the exact mechanism hasn't been defined in the [Wing language spec](https://www.winglang.io/docs/language-reference#38-enumeration) yet, it's technically been possible to stringify enums through a bug in the compiler that allowed any type to be stringified:

```js
bring cloud;

enum Result {
  SUCCESS,
  FAILURE
}

log("{Result.SUCCESS}");
```

Previously this printed out "0". To align with the default behavior of many other languages, this PR updates the behavior so enums stringify by default to their own names (in this case, "SUCCESS").

With that, the PR fixes the gap that allowed all types to be stringified, so for example the following is now an error:

```js
class A {
  privateField: str;
  new() {
    this.privateField = "internal";
  }
}

let a = new A();
log("{a}");
```

This produces the error:

```
error: Expected type to be stringable, but got "A" instead
  --> main.w:9:7
  |
9 | log("{a}");
  |       ^ Expected type to be stringable, but got "A" instead
  |
  = hint: str, num, bool, json, and enums are stringable
```

If you needed this behavior, you can work around it by defining an [extern JavaScript function](https://www.winglang.io/docs/language-reference#52-javascript) that prints the value or calls "toString()", but please note the stringification of these types is subject to change.

This issue tracks the default stringification for more types: https://github.com/winglang/wing/issues/5890

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify/snapshots/enum_value.snap
```diff
@@ -52,8 +52,8 @@ class $Root extends $stdlib.std.Resource {
     super($scope, $id);
     const MyEnum =
       (function (tmp) {
-        tmp[tmp["B"] = 0] = ",B";
-        tmp[tmp["C"] = 1] = ",C";
+        tmp["B"] = "B";
+        tmp["C"] = "C";
         return tmp;
       })({})
```
Chriscbr, line 58: This is equivalent to how TypeScript emits:

```
enum MyEnum {
  B = "B",
  C = "C",
}
```

libs/wingc/src/type_check.rs
```diff
@@ -3220,6 +3219,18 @@ impl<'a> TypeChecker<'a> {
...
+		if actual_type.is_stringable() {
+			return actual_type;
+		}
+
+		let message = format!("Expected type to be stringable, but got \"{actual_type}\" instead");
+		let hint = "str, num, bool, json, and enums are stringable".to_string();
+		self.spanned_error_with_hints(span, message, vec![hint]);
+
+		self.types.string()
+	}
```
MarkMcCulloh, line 3232: nit: The return type doesn't seem to be used, is it needed?

```suggestion
	pub fn validate_type_is_stringable(&mut self, actual_type: TypeRef, span: &impl Spanned) {
		if !actual_type.is_stringable() {
		  let message = format!("Expected type to be stringable, but got \"{actual_type}\" instead");
		  let hint = "str, num, bool, json, and enums are stringable".to_string();
		  self.spanned_error_with_hints(span, message, vec![hint]);
		}
	}
```

(or if you want to avoid nesting, an early `return;`)

MarkMcCulloh (approved):

libs/wingc/src/type_check.rs
Chriscbr: 👍  Fixed

tools/hangar/__snapshots__/test_corpus/valid/enums.test.w_compile_tf-aws.md
```diff
@@ -54,9 +76,9 @@ class $Root extends $stdlib.std.Resource {
     super($scope, $id);
     const SomeEnum =
       (function (tmp) {
-        tmp[tmp["ONE"] = 0] = ",ONE";
```
eladb, line 57: I suspect this won't be compatible with how TypeScript/JSII represent enums...

libs/wingc/src/type_check.rs
```diff
@@ -3220,6 +3219,14 @@ impl<'a> TypeChecker<'a> {
 		first_expected_type
 	}
 
+	pub fn validate_type_is_stringable(&mut self, actual_type: TypeRef, span: &impl Spanned) {
```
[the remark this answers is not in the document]
Chriscbr: Good idea

tools/hangar/__snapshots__/test_corpus/valid/enums.test.w_compile_tf-aws.md
Chriscbr: I think it should work since this is pretty close to how enums are emitted in TypeScript (see screenshot) but I'll keep my eyes peeled for any breakages, unless you have ideas how to test this
 
Screenshot 2024-03-12 at 1 33 38 PM

PR #5896, opened 2024-03-12
chore: fix nondeterministic bucket sim test

### LINKED ISSUE #5895: Non-deterministic wingsdk unit test for bucket simulator
The snapshot for `update an object in bucket` wingsdk test here
https://github.com/winglang/wing/blob/main/libs/wingsdk/test/target-sim/bucket.test.ts#L37-L57

Can sometimes have an additional entry in the snapshot for `InvokeAsync`. The test should only run this once. The second `InvokeAsync` is due to the simulation ending causing the sandbox to exit with a SIGTERM.

MarkMcCulloh:
Fixes #5895

The goal of the test is to make sure we don't emit events for object creation twice. We can just avoid the problem of stopping the sim by doing it after getting the snapshot fails.

No matter what, the sim will be stopped due to the way the test SimApp works.


eladb (approved):
Cool

PR #5915, opened 2024-03-12
feat!: `app.parameters`

hasanaburayyan:
## Summary
BREAKING CHANGE: `app.platformParameters` has been changed to `app.parameters`

This PR now allows Wing authors to access the application's parameters through `app.parameters` for example:
```js
let parameters = nodeof(this).app.parameters;
```
This also means that now wing structs can easily be added to platform parameters through the use of `parameters.addSchema(Struct.schema())` as well structs can be read from the parameter list through `Struct.fromJson(parameters.read())`

## Checklist

- [x] Tests added (always)


examples/tests/valid/parameters.test.w
```diff
@@ -0,0 +1,23 @@
...
+    foo: {
+      type: "string"
+    },
+    meaning_of_life: {
+      type: "number"
+    }
+  }
+});
+
+let foo = registrar.getParameterValue("foo");
```
hasanaburayyan, line 17: I need to add a test of getting a value that is available, just need to tweak how the test would be generated in hangar

examples/tests/invalid/parameters.test.w
```diff
@@ -0,0 +1,18 @@
+let app = nodeof(this).app;
+
+let registrar = app.parameterRegistrar;
```
eladb, line 3: May I propose some name changes (at least on the Wing side):

```js
app.parameters.addSchema(...);
app.parameters.getValue("foo");
```

examples/tests/invalid/parameters.test.w
```diff
@@ -0,0 +1,18 @@
...
+
+registrar.addParameterSchema({
+  type: "object",
+  properties: {
+    foo: {
+      type: "string"
+    }
+  },
+  required: ["foo"]
+});
```
eladb, line 13: Of course:

```suggestion
struct MyParams {
  foo: str;
}

registrar.addParameterSchema(MyParams.schema());
```

eladb (changes requested):

examples/tests/invalid/parameters.test.w
hasanaburayyan: oh but of course!

examples/tests/invalid/parameters.test.w
hasanaburayyan: Yea honestly I think these are just good name changes even in the typescript side. Naming is hard

examples/tests/invalid/parameters.test.w
hasanaburayyan: This is actually pretty fucking cool, I added a short video to the PR description @eladb, feels like the puzzle pieces are starting to make a picture

hasanaburayyan:
I think I could actually get `Struct.fromParameters()` to work without to much effort

```js
struct AppParams {
  someMessage: str;
  someNumber: num;
}

let app = nodeof(this).app;

app.parameters.addSchema(AppParams.schema());
let appParams = AppParams.fromParameters(app.parameters);

log(appParams.someMessage);
```

examples/tests/valid/apps_with_params/nested_params/parameters.test.w
```diff
@@ -0,0 +1,24 @@
...
+struct MyParams {
+  houses: Array<House>;
+}
+
+
+let app = nodeof(this).app;
+
+app.parameters.addSchema(MyParams.schema());
+
+let myParams = MyParams.fromParameters(app.parameters);
```
eladb, line 20: `.fromParameters` feels a bit off. Not sure it makes sense to "pollute" all of our structs with a static method dedicated to reading `wing.toml` values. 

Maybe we can just use `fromJson`?

Maybe:

```js
MyParams.fromJson(app.parameters.json());
```

I am also wondering if the common case is to read all the parameters at once or portions of it... Feels like we need some kind of "sections" (e.g. `app.parameters.section("tf-aws")` which will return everything under the `[tf-aws]` section in the toml file...

eladb (changes requested):

examples/tests/valid/apps_with_params/nested_params/parameters.test.w
hasanaburayyan: > I am also wondering if the common case is to read all the parameters at once or portions of it...

Im imagining the common case here is that it will be to read portions of it. Which is why the behavior of `.fromParameters()` will only return a subset of the parameters based on the Struct's schema. While I think `app.parameters.section("tf-aws")` will work for using `.fromJson` it also feels a bit strange to grab portions by string values. 

I dunno, Im not sure if I find it too odd to have `.fromParameters` since parameters are part of the core platform system. From a platform builder perspective I think it offers a pretty smooth DX. Though I do understand the concern with additional static methods.

hasanaburayyan: Though actually maybe that means that `.fromJson` should work the same way, that given an object will only extract the fields that match its schema... I think that might be a bug. We dont error on extraneous fields (which is intentional) but we should ignore them and not assign them to the struct.

Sorry for the wall of text Im just brain dumping 😅 Im running on 0 caffeine from Ramadan so thoughts are lagging

hasanaburayyan: Okay I fixed `fromJson` to not just return the whole JSON if the validation is successful. As for adding `fromSection` I dont think we need to, since if you want your Struct to be parsed from a certain section of the wing.toml you can do
```js
Struct.fromJson(app.parameters.json().get("tf-aws"));
```

lemme know what you think @eladb and ignore most of my ramblings before these last 2 comments.

hasanaburayyan: Yea im aligned on just using `fromJson` now, ive made the changes

tools/hangar/src/paths.ts
```diff
@@ -5,6 +5,7 @@ export const repoRoot = path.resolve(__dirname, "../../..");
 export const examplesDir = path.join(repoRoot, "examples");
 export const testDir = path.join(examplesDir, "tests");
 export const validTestDir = path.join(testDir, "valid");
+export const appWithParamsDir = path.join(validTestDir, "apps_with_params");
```
MarkMcCulloh, line 8: We have a couple other folders in this dir that do something similar so we can probably leave out the `app_with_` part of this and just call the folder `parameters`.
I don't have a strong opinion on this though because examples folder is really due for a big makeover anyways.

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -41,12 +41,27 @@ export class ParameterRegistrar extends Construct {
...
   }
 
+  /**
+   * Return all parameters as Json
+   *
+   * @returns the schema as a string
+   */
+  public json(): Json {
+    return this._rawParameters as Json;
+  }
```
MarkMcCulloh, line 51: I don't really have an alternative solution, but we have to watch out with using `std` types outside of the `std` namespace. This `Json` class isn't really real since much of its behavior is technically created via macros. So if a consumer of the SDK outside of wing calls this function and attempts to call any of the macro functions they'll get an error.

Maybe that's okay for the platform namespace for now though.

MarkMcCulloh (approved):

docs/docs/04-standard-library/std/json.md
```diff
@@ -404,6 +404,25 @@ the Json object to validate.
...
+| **Name** | **Type** | **Description** |
+| --- | --- | --- |
+| <code><a href="#@winglang/sdk.std.JsonSchema.property.rawSchema">rawSchema</a></code> | <code>any</code> | The raw Json Schema definition. |
+
+---
+
+##### `rawSchema`<sup>Required</sup> <a name="rawSchema" id="@winglang/sdk.std.JsonSchema.property.rawSchema"></a>
+
+```wing
+rawSchema: any;
```
eladb, line 418: Why is this needed?

examples/tests/valid/apps_with_params/nested_params/parameters.test.w
```diff
@@ -0,0 +1,23 @@
...
+
+struct MyParams {
+  houses: Array<House>;
+}
+
+let app = nodeof(this).app;
+
+app.parameters.addSchema(MyParams.schema());
+
+let myParams = MyParams.fromJson(app.parameters.json());
```
eladb, line 19: Any reason these two APIs can't be one?

```suggestion
let myParams = MyParams.fromJson(app.parameters.json(schema: MyParams.schema()));
```

libs/wingsdk/src/core/app.ts
```diff
@@ -187,14 +187,11 @@ export abstract class App extends Construct implements IApp {
...
-      this._platformParameters = new ParameterRegistrar(
-        this,
-        "ParameterRegistrar"
-      );
+  public get parameters() {
+    if (!this._parameters) {
+      this._parameters = new ParameterRegistrar(this, "ParameterRegistrar");
     }
-    return this._platformParameters!;
+    return this._parameters!;
```
eladb, line 194: Call out this breaking change in the readme

libs/wingsdk/src/platform/parameter-registrar.ts
```diff
@@ -29,7 +29,7 @@ export class ParameterRegistrar extends Construct {
    * @param path the path of the parameter
    * @returns the value of the parameter
    */
-  public getParameterValue(path: string): any {
+  public getValue(path: string): any {
```
eladb, line 32:
```suggestion
  public value(path: string): any {
```

We are normally not using `get` this is not Java here

libs/wingsdk/src/platform/parameter-registrar.ts
eladb: Yes that's a good call. The typescript experience will not be nice.

For most std types there's a typescript equivalent so we just use these but we don't have for Json, ha?

I am wondering if we we should just use `any` for this and maybe annotate it somehow for Wing to know it's a `json`.

libs/wingsdk/src/std/json_schema.ts
```diff
@@ -25,11 +25,12 @@ export class JsonSchema {
     return new JsonSchema(schema);
   }
 
-  private jsonSchema: any;
+  /** The raw Json Schema definition */
+  public rawSchema: any;
```
eladb, line 29: Make this internal:

```suggestion
  /** @internal */
  public _rawSchema: any;
```

examples/tests/valid/apps_with_params/nested_params/parameters.test.w
hasanaburayyan: Im not sure I understand the implication of `paramters.json()` accepting a schema object. Does passing this schema in automatically add it to the applications parameters? It feels a bit off to have this accessor style method also be able to accept a parameter that mutates.

eladb: What's the use case of decoupling these APIs? It seems like you would always call `addSchema()` and then immediately afterwards you'd call `json()` to obtain the values.

So we might as well just have a single API that allows you to read the parameters and validate the schema. Maybe it should be `parameters.read(schema: MyParams.schema())` or something like this.

hasanaburayyan: Okay so I went ahead and changed things to use `parameters.read()` which can be called with or without options. I think there is still a use case for just wanting to read all the parameters in without adding a schema to the registrar. 

so both work: `parameters.read()` returns all parameters back as Json, and `parameters.read(schema: Foo.schema())` will only return back the parameters that are part of that schema, but also ensure the schema is added to the registrar.

eladb (approved):

eladb (approved):

PR #5935, opened 2024-03-13
feat(compiler): allow explicit lift qualifications of preflight objects

yoav-steinberg:
Fixes: #76
Creates a `lift` builtin function that can be used in inflight code to explicitly add lift qualifications to a method:
```wing
bring cloud;

bucket.addObject("k", "value");

let some_ops = ["put", "list"]; // We can define a list of ops in preflight code to be used when explicitly qualifying a lift

class Foo {
  pub inflight mehtod() {
    lift(bucket, some_ops); // Explicitly add some permissions to `bucket` using a preflight expression
    lift(bucket, ["delete"]); // Add more permissions to bucket using a literal
    log(bucket.get("k")); // Good old implicit qualification adds `get` permissions
    let b = bucket; // We can now use an inflight variable `b` to reference a preflight object `bucket`
    b.put("k2", "value2"); // We don't get a compiler error here, because explicit lifts are being used in the method disabling compiler qualification errors
    for k in b.list() { // `list` works on `bucket` because of explicit qualification and `b` references `bucket`
      log(k);
    }
    b.delete("k2"); // `delete` also works because of explicit qualification
    assert(bucket.tryGet("k2") == nil); `yay!`
  }
}

let foo = new Foo();

test "a test" {
  foo.mehtod();
}
```

## Checklist

- [x] Tests added (always)


libs/wingc/src/jsify/snapshots/access_property_on_value_returned_from_collection.snap
```diff
@@ -74,10 +74,10 @@ class $Root extends $stdlib.std.Resource {
       get _liftMap() {
         return ({
           "handle": [
-            [((obj, key) => { if (!(key in obj)) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })(s, "hello"), ["length"]],
+            [((obj, key) => { if (!(key in obj)) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })(s, "hello"), [].concat(...[["length"]])],
```
eladb, line 77: Maybe we don't need the ...?

```suggestion
            [((obj, key) => { if (!(key in obj)) throw new Error(`Map does not contain key: "${key}"`); return obj[key]; })(s, "hello"), [].concat(["length"])],
```

libs/wingc/src/jsify/snapshots/base_class_captures_inflight.snap
```diff
@@ -96,10 +96,10 @@ class $Root extends $stdlib.std.Resource {
       get _liftMap() {
         return ({
           "bar": [
-            [x, []],
+            [x, [].concat(...[])],
```
eladb, line 99: Nit: maybe concat is not needed here?

libs/wingc/src/jsify/snapshots/lift_binary_preflight_expression.snap
```diff
@@ -75,12 +75,12 @@ class $Root extends $stdlib.std.Resource {
       get _liftMap() {
         return ({
           "handle": [
-            [x, []],
-            [y, []],
+            [x, [].concat(...[])],
```
eladb, line 78: Would be nice even just to clean the diff

libs/wingc/src/lifting.rs
```diff
@@ -151,6 +152,12 @@ impl<'a> LiftVisitor<'a> {
 			udt_js
 		}
 	}
+
+	// Used for generating a js array represtining a lift qualificaiton (an operation done on a lifted preflight object)
+	// lift qualifcations are in array format so multile ops can be bunched together in some cases.
```
eladb, line 157:
```suggestion
	// lift qualifcations are in array format so multiple ops can be bunched together in some cases.
```

libs/wingc/src/type_check.rs
```diff
@@ -2021,9 +2021,44 @@ impl<'a> TypeChecker<'a> {
...
+
+		let str_array_type = self.types.add_type(Type::Array(self.types.string()));
+		self.add_builtin(
+			&UtilityFunctions::Lift.to_string(),
+			Type::Function(FunctionSignature {
+				this_type: None,
+				parameters: vec![
+					FunctionParameter {
+						name: "preflightObject".into(),
+						typeref: self.types.anything(), // TODO: maybe this has to be only preflight classes/interfaces??
```
eladb, line 2033: `std.IResource`?

libs/wingc/src/type_check.rs
```diff
@@ -2021,9 +2021,44 @@ impl<'a> TypeChecker<'a> {
...
+							For example, if the method accesses the `push` and `pop` and `age` members of a [cloud.Queue] object, the qualifications should be `[\"push\", \"pop\"]`."
+						),
+						variadic: false,
+					},
+				],
+				return_type: self.types.void(),
+				phase: Phase::Inflight,
+				// This builtin actually compiles to nothing in JS, it's a marker that behaves like a function in the type checker
+				// and is used during the lifting phase to explicitly define lifts for an inflight method
+				js_override: Some("".to_string()),
```
eladb, line 2052: Cute

libs/wingc/src/lifting.rs
```diff
@@ -197,10 +204,13 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
...
+			v.check_explicit_lift(node);
+
 			// Inflight expressions that evaluate to a preflight type are currently unsupported because
 			// we can't determine exactly which preflight object is being accessed and therefore can't
 			// qualify the original lift expression.
-			if expr_phase == Phase::Inflight && expr_type.is_preflight_class() && v.ctx.current_property().is_some() {
+			if expr_phase == Phase::Inflight && expr_type.is_preflight_class() && v.ctx.current_property().is_some() && !v.ignore_unknown_preflight_object_error() {
 				report_diagnostic(Diagnostic {
 					message: format!(
 						"Expression of type \"{expr_type}\" references an unknown preflight object, can't qualify its capabilities (see https://github.com/winglang/wing/issues/76 for details)"
```
eladb, line 216: Perhaps in this error we can hint that they should use an explicit lift...

Also this GitHub issue reference is no longer needed.

eladb (approved):
Very excited about this!

Chriscbr:
I'm playing around with the changes a bit on my local machine -- to be honest, I'm not sure I'm (personally) in love with the syntax.

I feel like it's potentially confusing that lift is a function that you call within an inflight function body (the body being where you put code that runs on the cloud), but instead it's really being used to provide declarative information about the enclosing function. For example, take a look at this code:

```js
bring cloud;

let b1 = new cloud.Bucket() as "b1";
let b2 = new cloud.Bucket() as "b2";

let selectBucket = inflight (i: num) => {
  if i % 2 == 0 {
    return b1;
  } else {
    return b2;
  }
};

  if false {
    lift(b1, ["put"]); // <-- calling "lift" inside of some control flow
    lift(b2, ["put"]);
  }
  let bucket = selectBucket(1);
  bucket.put("hello.txt", "world");
});
```

I think this code snippet could be pretty confusing to someone learning Wing for the first time, since many a user would rightfully assume the code inside the `if false` branch is going to be ignored. But as implemented by this PR, the statements are executed and the "put" permissions are added to the cloud.Function.

Likewise, when I read this code I'd expect that since lift looks like a function call, I could just refactor them into a separate function:

```js
let addLifts = () => {
  lift(b1, ["put"]);
  lift(b2, ["put"]);
};

  addLifts();
  let bucket = selectBucket(1);
  bucket.put("hello.txt", "world");
});
```

Unfortunately this doesn't work (it's a compile error). Even if we change it to "lift!()", it would still looks like a function call is being made.ed 

I feel like the mechanism (providing some way to declare sets of preflight objects and operations that should be added to the internal "lift map") seems like right direction, but my preference would be if we could iterate on the syntax a bit more before merging this feature. I see @eladb has already approved this so I don't know if this is open to discussion, but I figured I'd shoot my shot sharing the feedback.

eladb:
I agree with you @Chriscbr that this solution could be potentially confusing for the reasons you mentioned. These reasons came up when we discussed this last week.

There are many things we can do but I suspect most of them will require dedicated syntax to be designed.

What I like about this proposal (and the reason I approved the PR) is that it's a small incremental step which allows us to experiment with this idea of explicit lifting without introducing new dedicated syntax just yet. At this stage of the project, I am not sure the investment required to design a dedicated syntax for lifting is the right thing to focus on (especially because it has the potential to suck us all into this wonderful conversation).

I suggest is to start with this to unblock the use case, and continue the discussion for dedicated syntax asynchronously. I'd love to see some proposals and as usual I have some ideas ;-)

eladb:
One small tweak we can do (we discussed this as well) is to require that lift() will be used as a statement (i.e return type void) and that all lifts will be placed at the top of the closure definition.

libs/wingc/src/jsify.rs
```diff
@@ -1856,8 +1856,15 @@ impl<'a> JSifier<'a> {
...
-					let ops_strings = method_lift_qual.ops.iter().map(|op| format!("\"{}\"", op)).join(", ");
-					bind_method.line(format!("[{code}, [{ops_strings}]],",));
+					let ops = method_lift_qual.ops.iter().join(", ");
+					// To keep the code concise treat no ops, single op and multiple ops differenly here, although the multiple ops is the generic case
+					if method_lift_qual.ops.len() == 0 {
+						bind_method.line(format!("[{code}, []],"));
+					} else if method_lift_qual.ops.len() == 1 {
+						bind_method.line(format!("[{code}, {ops}],"));
+					} else {
+						bind_method.line(format!("[{code}, [].concat(...[{ops}])],"));
```
eladb, line 1866:
```suggestion
						bind_method.line(format!("[{code}, [].concat({ops})],"));
```

I think this should also work, no?

libs/wingc/src/lifting.rs
```diff
@@ -197,13 +204,16 @@ impl<'a> Visit<'a> for LiftVisitor<'a> {
...
+
 			// Inflight expressions that evaluate to a preflight type are currently unsupported because
 			// we can't determine exactly which preflight object is being accessed and therefore can't
 			// qualify the original lift expression.
-			if expr_phase == Phase::Inflight && expr_type.is_preflight_class() && v.ctx.current_property().is_some() {
+			if expr_phase == Phase::Inflight && expr_type.is_preflight_class() && v.ctx.current_property().is_some() && !v.ignore_unknown_preflight_object_error() {
 				report_diagnostic(Diagnostic {
 					message: format!(
-						"Expression of type \"{expr_type}\" references an unknown preflight object, can't qualify its capabilities (see https://github.com/winglang/wing/issues/76 for details)"
+						"Expression of type \"{expr_type}\" references an unknown preflight object, can't qualify its capabilities. Use the `lift` builtin to explicitly qualify the relevant preflight object"
```
eladb, line 216:
```suggestion
						"Expression \"{expr_type}\" references an unknown preflight object, can't qualify its capabilities. Use `lift()` to explicitly qualify the preflight object to disable this error."
```

yoav-steinberg:
> One small tweak we can do (we discussed this as well) is to require that lift() will be used as a statement (i.e return type void) and that all lifts will be placed at the top of the closure definition.

Here's where I stand after a lengthy talk with @Chriscbr, either add this :point_up: and merge this PR **or** take a look at this new rfc: #5951 and continue from there (or both?).

eladb:
The RFC will take time to stabilize. 

I vote for merging this pull request (with the requirements that all lifts are at the top). 

This is a stable incremental step forward, it will unblock this use case and will allow us to continue the syntax discussion over the RFC without blocking.

yoav-steinberg:
@Chriscbr @MarkMcCulloh this is basically ready, please read https://github.com/winglang/wing/pull/5935#issuecomment-2002362478 and if you don't object, I'll merge this.

PR #5951, opened 2024-03-14
rfc: explicit lift qualification

yoav-steinberg:
## Checklist

[rendered version](https://github.com/winglang/wing/blob/yoav/rfc-explicit_lift_qualification/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md)

Related to #76, #5935

- [ ] Tests added (always)

### DIFF
--- a/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
+++ b/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
@@ -135,7 +135,7 @@ inflight () => {
 Internally the argument to the `lift` block creates a `std.Qualifications` object which is added to the closures lift qualifications. 
 
 Note that multiple `lift` blocks in the code just create more `std.Qualifications`s for this closure. 
-The actuall qualification values (`b1` and `"put"`) are just appended to the closure/method and they don't strictly belong to the block.
+The actual qualification values (`b1` and `"put"`) are just appended to the closure/method and they don't strictly belong to the block.
 
 What does apply to the block are the disabling of qualification errors within the block, so in our example we can use the inflight expression `b`, and inside the block we also don't automatically add qualifications even if the compiler can technically do this. 
 This behavior ensures that we can create custom qualifications without worrying that the compiler with override this with something more general.
--- a/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
+++ b/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
@@ -40,7 +40,7 @@ When the compiler encounters an inflight expression that references a preflight 
 
 In the future we'd like to employ data flow analysis (probably through [reaching defintions analysis](https://en.wikipedia.org/wiki/Reaching_definition) resolve these errors.
 
-## Problems with current solution
+## Problems with the current solution
 
 We need a mechanism for explicitly qualifying lifted objects. Consider the following cases:
 
@@ -123,15 +123,15 @@ inflight () => {
   // In the following scope we disable qualification compilation errors due to unknown preflight object
   // and we also disable automatic qualification.
   // We explicitly qualify b1 and b2 with "put" since the compiler doesn't figure this out for us.
-  with lift_qualification {b1: ["put"], b2: ["put"]} {
+  lift {b1: ["put"], b2: ["put"]} {
     b.put("key", "value"); 
   }
 }
 ```
 
-Internally the argument to the `lift_qualification` keyword creates a `std.Qualifications` object wich is added to the closures lift qualifications. 
+Internally the argument to the `lift` block creates a `std.Qualifications` object wich is added to the closures lift qualifications. 
 
-Note that multiple `lift_qualification` blocks in the code just create more `std.Qualifications`s for this closure. 
+Note that multiple `lift` blocks in the code just create more `std.Qualifications`s for this closure. 
 The actuall qualification values (`b1` and `"put"`) are just appended to the closure/method and they don't strictly belong to the block.
 
 What does apply to the block are the disabling of qualification errors within the block, so in our example we can use the inflight expression `b`, and inside the block we also don't automatically add qualifications even if the compile can technically do this. 
@@ -177,7 +177,7 @@ let arr = [b1,b2];
 
 class A {
   inflight foo() {
-    with lift_qualifications qual { // Use the qualification object programatically defined in preflight code
+    lift qual { // Use the qualification object programatically defined in preflight code
       for b in arr {
         b.put("k99", "v"); // This key pattern should work for both buckets
       }
@@ -188,7 +188,7 @@ class A {
   }
 
   inflight bar() {
-    with lift_qualifications qual { // reuse `qual` for another method of `A`
+    lift qual { // reuse `qual` for another method of `A`
       let b = arr.at(randome(arr.len));
       b.put("k0", "bar");
     }
--- a/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
+++ b/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
@@ -175,7 +175,7 @@ let no_nesting = inflight () => {
 
 ## Next steps
 
-The next step will be to provide use library APIs for programatically creating and manipulating `std.Qualifications` objects and then reusing them. We can also inherit from them and to provide qualification objects that go beyond only the "operation" being performed on the object. Here's an example:
+The next step will be to provide use library APIs for programatically creating and manipulating `std.LiftQualification` objects and then reusing them. We can also inherit from them and to provide qualification objects that go beyond only the "operation" being performed on the object. Here's an example:
 
 ```js
 // stdlib
--- a/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
+++ b/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
@@ -132,9 +132,9 @@ inflight () => {
 }
 ```
 
-Internally the argument to the `lift` block creates a `std.Qualifications` object which is added to the closures lift qualifications. 
+Internally the argument to the `lift` block creates a `std.LiftQualificationsContainer` object which is added to the closures lift qualifications. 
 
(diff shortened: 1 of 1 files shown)


docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,198 @@
...
+
+// Define our explicit qualifications, b1 has a special key pattern when putting
+let b1Quals = new cloud.BucketPutCap(b1);
+b1Quals.setKeyPattern("k*");
+// b2 can put any key
+let b2Quals = new std.Qualification(b2, ["put"]);
+
+// Add them to our explicit qualifications
+quals.add(b1Quals);
+quals.add(b2Quals);
```
Chriscbr, line 174: The design of the `std.Qualifications` class seems the most risky to me. It could be worth reducing design surface area and seeing how far we can get without defining per-method qualification data (definitely a cool idea to consider for the future though).

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,198 @@
...
+let b2 = new cloud.Bucket() as "b2";
+inflight () => {
+  let var b = b1;
+  if true {
+    b = b2;
+  }
+  // In the following scope we disable qualification compilation errors due to unknown preflight object
+  // and we also disable automatic qualification.
+  // We explicitly qualify b1 and b2 with "put" since the compiler doesn't figure this out for us.
+  with lift_qualification {b1: ["put"], b2: ["put"]} {
```
Chriscbr, line 126: Some ideas other ideas for the naming could be `with capability <C> { ... }` or `using qualification <Q> { ... }`

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,198 @@
...
+let b1Quals = new cloud.BucketPutCap(b1);
+b1Quals.setKeyPattern("k*");
+// b2 can put any key
+let b2Quals = new std.Qualification(b2, ["put"]);
+
+// Add them to our explicit qualifications
+quals.add(b1Quals);
+quals.add(b2Quals);
+
+let arr = [b1,b2];
```
Chriscbr, line 176: I think this should unblock our collection use cases:

```js
// creating a capability for invoking "put" on every bucket in an array
let buckets: Array<cloud.Bucket> = ...
let bucketsCap = new std.Capability();
for b in buckets {
  bucketsCap.add(b, ["put"]);
}

// later

inflight () => {
  with capability bucketsCap {
    for b in buckets {
      b.put("key", "value");
    }
  }
}
```

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
MarkMcCulloh, line 126: Syntactically I'm a big fan of the approach over an expression-esque one.

MarkMcCulloh: Do we have an idea about what `with` or `using` is actually doing here from a user's perspective? Just the keyword I mean, not the entire construct.

If we don't have that fleshed out, I feel like we should just keep it as short as possible and avoid attributing any sort of "kind" like `lift_qualification`. i.e. just:

```js
with { b1: ["put"], b2: ["put"] } {
  // ...
}

let quals = new std.Qualifications();
with quals {
  // ...
}
```

or even bring back lift:

```js
lift { b1: ["put"], b2: ["put"] } {
  // ...
}
```

yoav-steinberg: > Do we have an idea about what with or using is actually
I kind of like python's [context manager pattern](https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers) for generalizing `with`, but sadly we're dealing with "comp-time" here. If you have any idea how to reconcile this and add a "context manager" to wing that'll work here, then `with` seems like good syntax. If not, then I agree, lets go with `lift { stuff } { .. }`.

eladb: +1 on `lift <map> { ... }`:

eladb (commented):
I like this direction :-)

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,198 @@
...
+  let var b = b1;
+  if true {
+    b = b2;
+  }
+  // In the following scope we disable qualification compilation errors due to unknown preflight object
+  // and we also disable automatic qualification.
+  // We explicitly qualify b1 and b2 with "put" since the compiler doesn't figure this out for us.
+  with lift_qualification {b1: ["put"], b2: ["put"]} {
+    b.put("key", "value"); 
+  }
```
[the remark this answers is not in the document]
yoav-steinberg: added these important notes, thanks!

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
yoav-steinberg: The input to `onLift` of an object is a host and a list of ops (strings). I think our goal should be that the input will be a list of qualifications/capabilities (not ops). A qualification can be a string but can also be something specific for that object, like a pattern for an argument, etc.
If this makes sense I'll add a note about it. But I agree this isn't the first step.
I do feel the more this is implemented in the SDK and less in the compiler, it'll be better in the long run.

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
yoav-steinberg: right, added to the example

Chriscbr (approved):
LGTM!

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,246 @@
...
+  lift {b1: ["put"], b2: ["put"]} {
+    b.put("key", "value"); 
+  }
+}
+```
+
+Internally the argument to the `lift` block creates a `std.Qualifications` object which is added to the closures lift qualifications. 
+
+Note that multiple `lift` blocks in the code just create more `std.Qualifications`s for this closure. 
+The actuall qualification values (`b1` and `"put"`) are just appended to the closure/method and they don't strictly belong to the block.
```
eladb, line 138:
```suggestion
The actual qualification values (`b1` and `"put"`) are just appended to the closure/method and they don't strictly belong to the block.
```

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,246 @@
...
+  let q = queue;
+  lift {bucket: ["get"]} {
+    b.get("key");
+    lift {queue: ["push"]} {
+      q.push("data");
+    }
+  }
+}
+
+let no_nesting inflight () => {
```
eladb, line 166:
```suggestion
let no_nesting = inflight () => {
```

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,246 @@
...
+4. Provide a framework that can be extended to programatically defining and reusing lift qualifications.
+
+## Naming
+
+"qualifiations" is used because it qualifies how the preflight object is being used. 
+Other words that we sometime used in the past and might be clearer are "capabilities" or "operations":
+* "capabilities" makes sense because it describes the capabilities the host needs in order to successfully run the inflight code.
+* "operations" makes sense because it describes the operations being performed on the preflight objects by the inflight code.
+
+In this proposal I'll use "qualifications", but I'd be happy to hear what you think.
```
eladb, line 100: Since this is a parameter of `lift`, I think we can't use the term without pairing it with the term "lift".

"Lift x"

`std.LiftQualification` is such a mouthful.

Other ideas:

- "lift specification", and then we can use `std.LiftSpec` 
- "lift definition" (`std.LiftDef`).
-  `LiftScope` (might be too overloaded with node scope). 
- `LiftContext` 👍


My favorite is `LiftContext`. It also feels like it will be a good vessel to add additional attributes to the lift in the future.

eladb (approved):
LGTM

(See suggested alternative naming).

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,246 @@
...
+  lift {bucket: ["get"], queue: ["push"]} {
+    b.get("key");
+    q.push("data");
+  }
+}
+```
+
+## Next steps
+
+The next step will be to provide use library APIs for programatically creating and manipulating `std.Qualifications` objects and then reusing them. We can also inherit from them and to provide qualification objects that go beyond only the "operation" being performed on the object. Here's an example:
```
eladb, line 178:
```suggestion
The next step will be to provide use library APIs for programatically creating and manipulating `std.LiftQualification` objects and then reusing them. We can also inherit from them and to provide qualification objects that go beyond only the "operation" being performed on the object. Here's an example:
```

eladb (approved):

docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md
```diff
@@ -0,0 +1,246 @@
...
+let b2 = new cloud.Bucket() as "b2";
+inflight () => {
+  let var b = b1;
+  if true {
+    b = b2;
+  }
+  // In the following scope we disable qualification compilation errors due to unknown preflight object
+  // and we also disable automatic qualification.
+  // We explicitly qualify b1 and b2 with "put" since the compiler doesn't figure this out for us.
+  lift {b1: ["put"], b2: ["put"]} {
```
yoav-steinberg, line 129: consider adding a build error if `"put"` doesn't exist on b1 (checks that the qualifiers really match existing methods).

Issue #3013, opened 2023-06-21
Using ESM in external preflight function fails


skorfmann:
### I tried this:

```
bring cloud;

class Utils {
  init() {}
  extern "./utils.js" toUpperCase(value: str): str;
}

let utils = new Utils();
utils.toUpperCase("world!");
```

```
import changecase from 'change-case'

export function toUpperCase(str) {
  return changecase.upperCase(str)
}
```

### This happened:

```
| [command]/usr/local/bin/wing compile -p /plugins/backend.s3.js -t tf-aws main.w
| Error: ERROR: Cannot use import statement outside a module
|
| target/main.tfaws.166271.tmp/.wing/preflight.js:15
|          }
|           toUpperCase(value)  {
| >>         return (require("/Users/sebastian/projects/monada/wing-github-action/examples/with-dependencies/./utils.js")["toUpperCase"])(value)
|          }
|          static _toInflightType(context) {
|
|     at compile (/usr/local/lib/node_modules/winglang/dist/commands/compile.js:117:19)
```

### I expected this:

It works, as it does for inflight functions (probably due to the esbuild bundling)

### Is there a workaround?

use commonjs where possible

### Component

Compiler

### Wing Version

0.21.5

### Node.js Version

v18.16.0

### Platform(s)

MacOS

### Anything else?

Changing the file extension to `.mjs` or setting `"type": "module"` doesn't change it either.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skyrpex:
I wonder:
- Do you have a package.json in the root of the project with `"type": "module"`? 
- Otherwise, did you try renaming your `.js` file to `.mjs`?

If none of the above conditions are met, the `.js` files are treated as commonjs modules.

skyrpex:
I tried renaming to `.mjs` and got this error:

```
Failed to compile.

ERROR: require() of ES Module /Users/cristian/Code/test/utils.mjs not supported.
Instead change the require of /Users/cristian/Code/test/utils.mjs to a dynamic import() which is available in all CommonJS modules.

target/index.wsim.116480.tmp/.wing/preflight.js:15
         }
         toUpperCase(value) {
>>         return (require("/Users/cristian/Code/test/./utils.mjs")["toUpperCase"])(value)
         }
         static _toInflightType(context) {
```

MarkMcCulloh:
[...]

PR #5967, opened 2024-03-17
feat: generate types for `extern` declarations

### LINKED ISSUE #5868: TypeScript `inflight`s (Enhanced `extern`)
### Use Case

To make use of existing huge/mature ecosystems, writing inflight functions in javascript/typescript would be a huge boost in the usability and adoption of the wing toolchain.

```ts
// handler.ts
// generated types
import type extern from "./handler.ts.extern"

export default {
  handler: async (event) => {
    console.log(event);
  }
} satisfies extern;
```

```wing
// main.w
bring cloud;

class MyResource {
  static inflight extern "./handler.ts" handler(event: str);
  myFunction: cloud.Function;
  new() {
    this.myFunction = new cloud.Function(MyResource.handler);
  }
}

new MyResource();
```

### Proposed Solution

Enhance the existing extern functionality with 2 features:

1. E2E type safety. The wing compiler should generate a type contract for the extern file to satisfy. This should include inflight clients for lifted resources.
2. A way to lift resources into the typescript. Related to https://github.com/winglang/wing/issues/76
3. Ability to pass extern function as arguments expecting inflight closures

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

MarkMcCulloh:
Closes #5868 (except the static part, but that's not critical for the use-case)

Compiling a wing project will now generate .d.ts files for the contract between wing and extern files. These .d.ts files are fully self-contained.

The code for this looks similar to existing dtsification code, but there are couple important distinctions that made it feel useful to separate it:
1. The new code only access to type information rather than the original ASTs
2. The new code is intentionally generating types in the same file as needed instead of producing references

Misc:
- No longer emit any files if any errors occur during compiliation
- Internally change extern to use Utf8PathBuff
- Converted some `examples` externs into typescript (I would love to do the rest, but need https://github.com/winglang/wing/issues/3013)

Note that this may be a breaking change: extern types must be consistent. So in one class you refer to an extern as returning a `str` and in another it returns a Json, that will now be a compiler error.


examples/tests/valid/dynamo.ts
```diff
@@ -11,15 +16,15 @@ export async function _putItem(tableName, item) {
   const response = await client.send(command);
   console.log(response);
   return;
-}
+};
 
-export async function _getItem(tableName, key) {
+export const _getItem: extern["_getItem"] = async (tableName, key) => {
```
eladb, line 21: Isn't there some typescript incantation that will apply this automatically to all extern declarations in this file?

examples/tests/valid/subdir/util.ts
```diff
@@ -0,0 +1,5 @@
+import type extern from "./util.extern";
+
+export const greet: extern["greet"] = async (name) => {
```
eladb, line 3: Once Yoav merges his PR, Can you add an example like this?

```js
let b = new cloud.Bucket();

class E {
  extern "./e.ts"
  pub static inflight f(b: cloud.Bucket);
}

inflight () => {
  lift(b, ["put"]);
  E.f(b);
};

// or

inflight () => {
  lift { b: ["put"] } {
    E.f(b);
  }
};

```

Curios to see the type annotations...

eladb (commented):
Cooooool

examples/tests/valid/dynamo.ts
MarkMcCulloh: Unfortunetly not https://github.com/microsoft/TypeScript/issues/38511.

The closest would be doing a default export with the whole module:

```ts
import type extern from "./blah.extern";

export default {
  method1() {
  }
  method2() {
  }
} satisfies extern
```

This does work now, however it adds a runtime cost to export it this way and it actually breaks esbuild's bundling in some of our examples.

---

Some magic that should be possible is to make `extern` an implicit global type for a given file. That way you can avoid the `import type extern ...`. I had trouble getting it working so I figured I'd leave it as a future dx improvement.

examples/tests/valid/subdir/util.ts
MarkMcCulloh: > Curios to see the type annotations...

The type won't be different yet, the .d.ts generation does not check any lifting information. I figured I would tackle that after his lift() pr is merged.

examples/tests/valid/dynamo.ts
eladb: Cool, thanks for the info

eladb (approved):

examples/tests/valid/capture_tokens.test.w
```diff
@@ -4,7 +4,7 @@ class MyResource {
   api: cloud.Api;
   url: str;
 
-  extern "./url_utils.js" pub static inflight isValidUrl(url: str): bool;
+  extern "./url_utils.ts" pub static inflight isValidUrl(url: str): bool;
```
Chriscbr, line 7: @MarkMcCulloh Just checking, so we've decided to officially support *.ts files? Should we note this in the documentation or language spec somewhere? (I'm cool with the choice, just wanna check if we're tracking it somewhere :^))

libs/wingc/src/dtsify/extern_dtsify.rs
```diff
@@ -0,0 +1,351 @@
...
+	known_types: HashMap<String, String>,
+	/// Because all types will be in the same namespace, if you types have the same name this map will be used to disambiguate them by incrementing the counter.
+	name_counter: HashMap<String, usize>,
+}
+
+/// Checks if given file has a valid extension to be considered an extern file
+pub fn is_extern_file(file: &Utf8PathBuf) -> bool {
+	if let Some(ext) = file.extension() {
+		match ext {
+			"js" | "cjs" | "mjs" | "jsx" | "ts" | "cts" | "mts" | "tsx" => true,
```
Chriscbr, line 36: Who woulda thunk a single language would result in so many file extensions / formats

PR #5984, opened 2024-03-18
fix(compiler): unable to chain ??

### LINKED ISSUE #1875: Unable to chain null coalescing operator (`??`)
### Community Note

### I tried this:

This doesn't work (@Chriscbr brought this up):

```js
let f = (x: str?, y: str?) => {
  let z = x ?? y ?? "hello";
//        ^^^^^^ Expected optional type, found "str"
  print("x is ${z}");
};
```

This works:

```js
let f = (x: str?, y: str?) => {
  let z = x ?? (y ?? "hello");
  print("x is ${z}");
};
```

This is also going to be defined in the spec (#436).

### I expected this:

To behave the same as JS's [Nullish_coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator.

### Instead, this happened:

See above.

### Component:

Compiler

### Wing Version:

_No response_

### Node.js Version:

_No response_

### Platform(s):

_No response_

### Anything else?

_No response_

### LINKED ISSUE #5944: The ?? operator cannot return an optional
### I tried this:

Wrote this code:
```ts
bring cloud;
let s1: str? = nil;
let s2: str? = nil;

let s3 = s1 ?? s2;
```

### This happened:

Got this error: 
> Expected type to be "str", but got "str?" instead

It was on the part after the `??` operator.

Even explicitely defining that s3 is optional didn't help:
This code still produced the same error:

```ts
bring cloud;
let s1: str? = nil;
let s2: str? = nil;

let s3: str? = s1 ?? s2;
```

### I expected this:

The code to compile an s3 to equal nil.

### Is there a workaround?

```ts
let var s3 = s1;
if (s3 == nil){
  s3 = s2;
}
```

But it is uglier, and requires s3 to be set to `var`

### Anything else?

I know it seems like a very small issue with a workaround, but for someone new to Wing, this can be the point that they lose interest and get back to a familiar language

### Wing Version

0.61.1

### Node.js Version

_No response_

### Platform(s)

MacOS

### Community Notes

Chriscbr:
Fixes #5944

This improves the type checking for expressions of the form `a ?? b` in two ways:
1. `b` is allowed to be an optional. So for example, if `x` and `y` are both `num?`, then `x ?? y` is allowed, and its type of the resulting expression is `num?` (reflecting that you may still end up with a nil). 
2. The subtyping is more permissive. For example, suppose you have two classes where `Derived` is a derived class from `Base`. (In other words, `Derived` is a subtype of `Base`). All of these type combinations are allowed:
	a. Optional(Base) ?? Optional(Derived) = Optional(Base)
	b. Optional(Base) ?? Derived = Base
	c. Optional(Derived) ?? Optional(Base) = Optional(Base)
	d. Optional(Derived) ?? Base = Base

## Checklist

- [x] Tests added (always)


yoav-steinberg (approved):
Looks good.
Even though having a `x ?? y` where both `x` and `y` are optionals feels somehow wrong to me.

libs/wingc/src/type_check.rs
```diff
@@ -2133,12 +2134,51 @@ impl<'a> TypeChecker<'a> {
 							// Left argument must be an optional type
 							if !ltype.is_option() {
 								self.spanned_error(left, format!("Expected optional type, found \"{}\"", ltype));
-								(ltype, phase)
-							} else {
-								// Right argument must be a subtype of the inner type of the left argument
-								let inner_type = *ltype.maybe_unwrap_option();
-								self.validate_type(rtype, inner_type, right);
-								(inner_type, phase)
```
yoav-steinberg, line 2141: I kind of see the logic in this straight forward "unwrap" implementation. Perhaps the original issue with #1875 is just a grammar right associativity thing.

Chriscbr: I just added a grammar test - [`fd14377` (#5984)](https://github.com/winglang/wing/pull/5984/commits/fd1437790e57a437b0448a2e4bc8e5bea59194b1)

If I look at the output, I think it's parsing (a ?? b ?? c) as ((a ?? b) ?? c) which I believe is correct

Chriscbr: On second thought, I think the operation is may be associative so the grouping of parens may not really matter all that much.

Chriscbr: I updated the PR description with a better explanation of the fixes

yoav-steinberg: note Rusts [`unwapr_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or), the result is always `T` and never an `Option<T>`.

Chriscbr: Ah I think I see what you mean now... yea changing the associativity should fix the original issue.

Chriscbr:
Superseding for now with https://github.com/winglang/wing/pull/5987

libs/wingc/src/type_check.rs
MarkMcCulloh: What's wrong with `a ?? b ?? c` -> `(a ?? b) ?? c`? Just a desire to avoid `??` resulting in an optional? 
To me this seems to solve the issue and also addresses https://github.com/winglang/wing/issues/5944. It also matches how JS works (although it's not necessary to match that)

Chriscbr: I added more context for the change in #5944. In the net this seems like a good change so let's try it

PR #5987, opened 2024-03-18
fix(compiler): cannot chain ?? expressions

### LINKED ISSUE #1875: Unable to chain null coalescing operator (`??`)

Chriscbr:
Fixes #1875

This fixes the issue by updating the precedent rules so the operation is right-associative. For now, expressions like `a ?? b` or `a ?? b ?? c` still have to return a non-optional value (like Rust's unwrap_or), but we can consider making it more flexible / open in the way that C# or Swift allows if appropriate in the future.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/valid/optionals.test.w
+++ b/examples/tests/valid/optionals.test.w
@@ -16,13 +16,14 @@ class Super {
 class Sub extends Super {
   new() { this.name = "Sub"; }
 }
-class Sub1 extends Super {
-  new() { this.name = "Sub"; }
+class SubSub extends Sub {
+  new() { this.name = "SubSub"; }
 }
 
 let optionalSup: Super? = new Super();
 let s = optionalSup ?? new Sub();
 assert(s.name == "Super");
+let s2 = optionalSup ?? optionalSup ?? new SubSub();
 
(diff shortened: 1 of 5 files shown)


eladb (approved):

yoav-steinberg (approved):
:+1:

Issue #5879, opened 2024-03-11
Connect to AWS infrastructure deployed on AWS, and interact with them from locally developed system


ekeren:
### Use Case

You can now develop only a portion of your system in wing and reference existing resources that are deployed on your remote dev environment.  This allows teams that have micro-stack to focus only on developing their own stack, while communicating with existing services. Wing Console support this behavior seamlessly, and although these resource are not simulated and are actually running on the cloud, you are able to view them and interact with them on the simulator. 

### Component

SDK, Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

garysassano:
I think this is related to #5170

ekeren:
@garysassano 💯 

@eladb, lets start small. I would like to see:
```wing
bring cloud;

let bucket = aws.Bucket.byId("id123");
let counter = new cloud.Counter(initial: 1);
let queue = new cloud.Queue();

queue.setConsumer(inflight (message: str) => {
  let index = counter.inc();
  bucket.put("wing-{index}.txt", "Hello, {message}");
  log("file wing-{index}.txt created");
});
```
And some mapping file:
```
id123: ARN
```

Where the Bucket is something that is already deployed on my dev account on AWS.
And the result should that:
- The code works 
- The bucket look like a bucket in wing console

ekeren:
The mapping file might be an overkill, but consider that different developers will have different mappings.

ainvoner:
@ekeren can we close it?

ekeren:
Yes. We can close this

Issue #5170, opened 2023-12-10
Support importing existing infrastructure


hasanaburayyan:
### Feature Spec

Wing resources can now be imported into your application code. This enables users to interact with resources that may have been created outside of the given Wing project. Whereas before Wing applications were expected to be entirely greenfield, now its easy to write poc, or one off jobs in Wing that interact with existing infrastructure in your cloud environment. 

### Use Cases

Import existing buckets, tables, redis, and other resources into your Wing application without having to provision them as part of your deployment process.

### Implementation Notes

One thought on implementation might be that resources in Wing have an optional `ImportOptions` property that contain any required fields that are needed to perform the import for the given Platform.

(Rough sketch maybe)
```js
bring cloud;

let b = new cloud.Bucket(import: {name: "my-existing-bucket"})
```

Related Slack thread: https://winglang.slack.com/archives/G04G3TCHE12/p1701893450410499

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

eladb:
Another thing to consider is what's the `sim` behavior of these resources. It could be:

1. Use a local simulated bucket.
2. Reference an S3 bucket in the cloud set up for development.

ekeren:
> let bucket = aws.Bucket.fromArn("arn:....");

We've just heard from a team deploying "micro-stacks" that are sharing resources on the network level and data storage level (like RDS, Shared S3 etc). Each stack has a different TF State file stored on S3.

It feels like this can be a common use case, and we should support it on the `cloud` namespace level. 
Something like:
```
let bucket = cloud.Bucket.from(j: Json) 
```

And the platform provider, using values or inputs ([rfc](https://github.com/winglang/wing/pull/5171)), is responsible to produce this Json, in the case of AWS it is probably `{ arn: '....' }`

eladb:
> we should support it on the `cloud` namespace level

Why does it matter to support this at the `cloud` abstract classes instead of the concrete resources?

ekeren:
> Why does it matter to support this at the cloud abstract classes instead of the concrete resources?

I don't think that the multi micro-stack applications that share a bucket means that my project needs to be bound to a specific cloud provider, the project can still be cloud agnostic...

eladb:
[...]

PR #6001, opened 2024-03-19
feat(sdk): `aws.QueueRef`

eladb:
Introduce `aws.QueueRef` as a way to add an external queue reference to an application:

```js
bring aws;
bring cloud;

let myQueue = new aws.QueueRef("arn:aws:sqs:...."); // <-- ARN of an existing SQS queue

  myQueue.push("new message");
});
```

This works both when running in the simulator and when running on AWS.

The Wing Console also includes some metadata about the queue such as the queue URL and the AWS Console URL:

image

### Issues

* Related to #5879 (only implements queue and we need additional resources).
* Related to #5170

### Misc

Added an option `link: bool` to `ui.FIeld` to indicate that the value returned is a link. We should make this a clickable link in the Wing Console.

## Checklist

- [x] Tests added (always)


MarkMcCulloh:
Is this a one-off solution for the related issue of referencing existing infra or is this intended to be the pattern to follow going forward?

Side note on the benchmark regressions: I'm willing to bet the additional ~90ms added to all the sim compilations are the imports added to `libs/wingsdk/src/shared-aws/queue.ts`. shared-aws is exported at the root so we'd loading all of these regardless of the target. We've previously avoided this dynamic imports for target-specific stuff, but not sure how easy that'd be here.

hasanaburayyan:
So firstly this is fucking cool, especially what it does in the console 🔥 

Thinking out loud here...

What about organizing these reference classes under a `reference` namespace so something like this:
```js
let outbox = new aws.reference.Queue("arn...");
```

eladb:
@MarkMcCulloh asked:

> Is this a one-off solution for the related issue of referencing existing infra or is this intended to be the pattern to follow going forward?

This is intended to be the pattern going forward for all external references, so we will have `aws.BucketRef`, `aws.TopicRef`, etc...

eladb:
@hasanaburayyan suggested:

> What about organizing these reference classes under a reference namespace so something like this: `aws.reference.Queue`

I am a flat lists and prefixes over namespace hierarchies as it's dramatically improves discoverability.

eladb:
@MarkMcCulloh, what would you recommend to do about the performance regressions (amazing that this was caught)?

libs/wingsdk/src/shared-aws/util.ts
```diff
@@ -0,0 +1,23 @@
+import { Token } from "cdktf";
```
MarkMcCulloh, line 1: > what would you recommend to do about the performance regressions (amazing that this was caught)?

Without doing any testing, my guess is that the issue is this. queue.ts imports this file, and having this at the top causes a `require("cdktf")` when you do `require("@winglang/sdk")`.

Try moving this inside isValidArn as a `require`

examples/tests/sdk_tests/queue/queue-ref.test.w
```diff
@@ -0,0 +1,54 @@
...
+    qref.pop();
+  });
+
+  test "can push to an external queue (QueueRef)" {
+    qref.push("m1");
+    qref.push("m2");
+    qref.push("m3");
+    util.waitUntil(() => {
+      return c.peek() == 3;
+    });
```
Chriscbr, line 30: Random design thoughts - should `util.waitUntil` throw by default if the condition is never met? (Right now there's a default timeout of around a minute, and if it fails, then it just returns `false`). And perhaps the throwing could be disabled with `throw: false`?

eladb: Mmm? It's supposed to throw after the timeout... Otherwise what's the point?

hasanaburayyan (approved):
LGTM, Im excited about this direction

examples/tests/sdk_tests/queue/queue-ref.test.w
eladb: OK, I am looking at the code and you are right. It's not throwing. Of course it needs to throw!

docs/docs/04-standard-library/aws/api-reference.md
```diff
@@ -146,6 +146,155 @@ The IAM certificate identifier value.
 ---
 
 
+### QueueRef <a name="QueueRef" id="@winglang/sdk.aws.QueueRef"></a>
```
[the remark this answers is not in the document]
eladb: I added a section in `queue.md` with information.

examples/tests/sdk_tests/queue/queue-ref.test.w
eladb: https://github.com/winglang/wing/issues/6006

PR #6008, opened 2024-03-19
fix(sdk)!: `util.waitUntil` is silent if condition is not met

### LINKED ISSUE #6006: `util.waitUntil` is not throwing after the timeout is elapsed?!!
### I tried this:

```js
bring util;

test "you should fail" {
  util.waitUntil(() => { return false; }, timeout: 1s);
}
```

### This happened:

The test PASSED!?!

### I expected this:

The test should fail if the condition is never met.

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

Chriscbr:
Closes #6006

`waitUntil()` is a built-in function in the `util` module for automatically polling and waiting until a condition is true (or until a timeout has been reached). From looking at how it was used in winglibs and example programs within the Winglang test suite, we saw there were a lot of cases where `waitUntil()` was called and the return value was silently ignored.

To make the API more intuitive, this PR changes the default behavior so that the function throws if the condition is never met within the timeout period, so it's no longer necessary to wrap the function in an `assert()` call for the most common use case. For the less-common case where you want to inspect whether the condition was reached and perform some action, it's now possible to set the `throws` option to `false` -- which is equivalent to switching to the old behavior.

BREAKING CHANGE: `util.waitUntil()` now throws if the condition is not met within the timeout period. To turn off this behavior, you can pass `throws: false` as a named option to the function.

## Checklist

- [x] Tests added (always)


MarkMcCulloh (approved):

eladb (approved):
🙏

PR #6009, opened 2024-03-20
fix: language server occasionally duplicates types a fails compilation

### LINKED ISSUE #6004: `extern type must be the same in all usages`
### I tried this:

Use extern

### This happened:

"First declared here" points to the same line :-):

```
extern type must be the same in all usages
hint: Change type to match first declaration: inflight (event: ApiRequest, ctx: AppendToStreamContext): ApiResponse
handlers.w(6, 82): First declared here
```

### I expected this:

_No response_

### Is there a workaround?

Reload VSCode

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

MarkMcCulloh:
Fixes #6004, I was able to replicate consistently and see that this change fixes it

Instead of attempting to reuse `Types`, let's just create a new one. This should cause all the existing types and symbol environments to be dropped, but that's okay because we immediately populate the new one with data.

No clue how to add a real test for this. The current rust-based tests don't have a multi-file system and I'm not sure this change is worth that effort to improve (as nice as it would be)

### DIFF
--- a/libs/wingc/src/lsp/sync.rs
+++ b/libs/wingc/src/lsp/sync.rs
@@ -150,8 +150,7 @@ fn partial_compile(
 	}
 
 	// Reset all type information
-	types.reset_expr_types();
-	types.reset_scope_envs();
+	types.reset_per_compile_data();
 
 	// -- TYPECHECKING PHASE --
 	let mut jsii_imports = vec![];
--- a/libs/wingc/src/lsp/sync.rs
+++ b/libs/wingc/src/lsp/sync.rs
@@ -120,7 +120,7 @@ pub fn on_document_did_change(params: DidChangeTextDocumentParams) {
 fn partial_compile(
 	source_path: &Path,
 	source_text: String,
-	mut types: &mut Types, // TODO: does this need to be shared between recompiles?
+	mut types: &mut Types,
 	jsii_types: &mut TypeSystem,
 	project_data: &mut ProjectData,
 ) {
@@ -150,7 +150,7 @@ fn partial_compile(
 	}
 
 	// Reset all type information
-	types.reset_per_compile_data();
+	*types = Types::new();
 
 	// -- TYPECHECKING PHASE --
 	let mut jsii_imports = vec![];
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -1664,12 +1664,6 @@ impl Types {
 		self.scope_envs[scope_id].expect("Scope should have an env")
 	}
 
-	pub fn reset_scope_envs(&mut self) {
-		for elem in self.scope_envs.iter_mut() {
-			*elem = None;
-		}
-	}
-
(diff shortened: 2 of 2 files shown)


libs/wingc/src/lsp/sync.rs
```diff
@@ -120,7 +120,7 @@ pub fn on_document_did_change(params: DidChangeTextDocumentParams) {
 fn partial_compile(
 	source_path: &Path,
 	source_text: String,
-	mut types: &mut Types, // TODO: does this need to be shared between recompiles?
+	mut types: &mut Types,
```
MarkMcCulloh, line 123: It's important to still keep `Types` around as long as possible in-between compilations because we need to access raw pointers to data it contains. But once we're about to compile again we know that we won't need the data inside it anymore.

libs/wingc/src/type_check.rs
```diff
@@ -1691,10 +1685,13 @@ impl Types {
...
-			*elem = None;
-		}
+	pub fn reset_per_compile_data(&mut self) {
+		self.type_for_expr.clear();
+		self.scope_envs.clear();
+		self.inferences.clear();
+		self.json_literal_casts.clear();
+		self.type_expressions.clear();
+		self.source_file_envs.clear();
 	}
```
Chriscbr: I'm wondering if the approach in this commit would be more efficient, since maybe we'd be saving some time spent reloading type information about jsii libraries from `TypeSystem` into `Types`. For example, [here](https://github.com/winglang/wing/blob/7f087c73f1f79a6af94324f3b91ed17f14b1e97b/libs/wingc/src/lib.rs#L229-L241) inside the top-level `type_check()` function called by both `compile()` and `partial_compile()`, we're resetting reloading information about the SDK and built-in functions etc., and in theory those calls could be skipped or made into no-ops during re-compiles.

Anyways it could be worth checking out if there's any performance difference between these approaches, but otherwise either looks okay to me (maybe throwing away `Types` each time will simply be more error-proof in the long run)

Chriscbr (approved):
LGTM. I'd be curious to learn how this fixed the bug (or what was the root cause?)

libs/wingc/src/type_check.rs
MarkMcCulloh: I think we could design around this, but it's tricky since we haven't been considering it strongly so far.
I'm pretty sure this approach causes a memory leak in the LSP for most type info because we'd never actually free anything that we don't have a good way to reuse.

Luckily the most expensive part of all of this (JSII importing) is saved already so I think it's okay to go with the safer/slower option for `Types` for now.

MarkMcCulloh:
> what was the root cause

root cause was `source_file_envs` in `Types` not being cleared in-between compilations. Over time `Types` has had more data that we weren't clearing between compilations, so this extern issue probably wasn't the only problem fixed here.

eladb (commented):
🎉

PR #6017, opened 2024-03-20
fix(sdk): unexpected end of JSON input error

### LINKED ISSUE #6010: Wing Console: `Unexpected end of JSON input`
### I tried this:

I don't know what steps to take to reproduce the problem exactly, it was when I tried to play a bit with `ex.React` to run a NextJS project.

### This happened:

`Unexpected end of JSON input` when I try to run the Console.

### I expected this:

_No response_

### Is there a workaround?

To remove the `.state` dir.

### Anything else?

As far as I can tell, the problem is that somehow an empty `.state/*/state.json` file is saved and the console tries to load it.

### Wing Version

0.61.18

### Node.js Version

v20.11.0

### Platform(s)

MacOS

### Community Notes

Chriscbr:
Fixes #6010

The underlying issue is caused by state files only getting partially written to disk when the simulator host process is killed in the middle of its shutdown. Due to the weird sequencing of operations, an empty file would end up being written (instead of no file or a file with valid JSON). Using synchronous calls for saving state files should reduce these kinds occurrences.

No tests were added because there isn't really a reliable way to kill a process precisely in the middle of a asynchronous Node.js function call. 

There's still a bug lurking where the console isn't letting the simulator finish shutting down, but this extra resiliency shouldn't hurt.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-sim/api.inflight.ts
+++ b/libs/wingsdk/src/target-sim/api.inflight.ts
@@ -162,7 +162,7 @@ export class Api
   }
 
   private async saveState(state: StateFileContents): Promise<void> {
-    await fs.promises.writeFile(
+    fs.writeFileSync(
       join(this.context.statedir, STATE_FILENAME),
       JSON.stringify(state)
     );
--- a/libs/wingsdk/src/target-sim/counter.inflight.ts
+++ b/libs/wingsdk/src/target-sim/counter.inflight.ts
@@ -38,7 +38,7 @@ export class Counter implements ICounterClient, ISimulatorResourceInstance {
   public async cleanup(): Promise<void> {}
 
   public async save(): Promise<void> {
-    await fs.promises.writeFile(
+    fs.writeFileSync(
       join(this.context.statedir, VALUES_FILENAME),
       JSON.stringify(Array.from(this.values.entries()))
     );
--- a/libs/wingsdk/src/target-sim/bucket.inflight.ts
+++ b/libs/wingsdk/src/target-sim/bucket.inflight.ts
@@ -15,15 +15,16 @@ import {
   BucketDeleteOptions,
   BucketGetOptions,
   BucketTryGetOptions,
+  BUCKET_FQN,
 } from "../cloud";
 import { deserialize, serialize } from "../simulator/serialization";
 import {
   ISimulatorContext,
   ISimulatorResourceInstance,
 } from "../simulator/simulator";
-import { Datetime, Json } from "../std";
+import { Datetime, Json, TraceType } from "../std";
 
-const METADATA_FILENAME = "metadata.json";
+export const METADATA_FILENAME = "metadata.json";
 
 export class Bucket implements IBucketClient, ISimulatorResourceInstance {
   private readonly _fileDir: string;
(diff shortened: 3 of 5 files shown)


libs/wingsdk/src/target-sim/api.inflight.ts
```diff
@@ -162,7 +162,7 @@ export class Api
   }
 
   private async saveState(state: StateFileContents): Promise<void> {
-    await fs.promises.writeFile(
+    fs.writeFileSync(
```
eladb, line 165: What's the motivation to change to sync?

(Curious)

libs/wingsdk/src/target-sim/counter.inflight.ts
```diff
@@ -38,7 +38,7 @@ export class Counter implements ICounterClient, ISimulatorResourceInstance {
   public async cleanup(): Promise<void> {}
 
   public async save(): Promise<void> {
-    await fs.promises.writeFile(
+    fs.writeFileSync(
```
eladb, line 41: Is there a way to also add some lower level protection against corrupted state files?

eladb (approved):

libs/wingsdk/src/target-sim/api.inflight.ts
MarkMcCulloh, line 165: I'm not against this change, but a more foolproof way to fix the issue would be to make the readFile more resilient instead right? Like if the file is corrupt/unparseable then it should remove it and assume no state is available.

Chriscbr: Here's what I believe is happening today:

1. User stops the Wing Console (via SIGTERM)
2. Wing Console asks the simulator to stop
3. Simulator starts stopping resources
4. The bucket's save() method is called.
5. The save method calls writeFile(). It creates and opens the file.
6. Node is waiting for I/O operations to finish.
7. The Wing Console process, seeing no more synchronous operations left on its call stack, exits. The writeFile() operation is unfinished so an empty (or partially written) file is saved to the state.

I haven't figured out the root cause of 7 yet (I've reached out to @skyrpex) so I opened this as a stop gap fix. Changing to sync means the file either gets written or it doesn't, which the bucket does know how to handle.

libs/wingsdk/src/target-sim/counter.inflight.ts
Chriscbr: Not sure 🤷‍♂️

It depends how much we want to design the API (so far I'm trying to avoid over-designing it). We could provide some set of formal APIs like `saveState(key: str, value: Json)` which all use safe operations under the hood, and also add some `validateState()` method that resources can implement which will be called before the state is read, etc.

My current thinking is to support resources like `cloud.Bucket`, simulator resources need to be able to read/write to their respective state directories whenever they want, and so the resource is responsible if it crashes or corrupts state somehow. But there might be alternatives I'm not thinking of

libs/wingsdk/src/target-sim/api.inflight.ts
Chriscbr: Sure -- yeah I was trying to fix the data getting corrupted, but we could also fix how we read the corrupted data. I can add that in

libs/wingsdk/src/target-sim/counter.inflight.ts
Chriscbr: High level I think first we should try what we can do to stop these cases of data corruption (and fix the issue in the console, assuming there's a bug there -- not addressed by this PR) and if that's not possible then we can double down on resiliency

Issue #5957, opened 2024-03-16
(containers) Support Exposing Multiple Ports On Container


hasanaburayyan:
### Use Case

It's common practice for containers to expose multiple ports when needed, especially in applications that require communication over different protocols or services running within the same container that listen on different ports.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

PR #6022, opened 2024-03-21
feat(sdk): introducing `sim.Container`

eladb:
The `sim.Container` resource allows running containers in the Wing Simulator.

```js
let c = new sim.Container(
  name: "http-echo",
  image: "hashicorp/http-echo",
  containerPort: 5678,
  args: ["-text=bang"],
);

test "send request" {
  http.get("http://localhost:{c.hostPort}");
}
```

There is also support for building containers from a local `Dockerfile`:

```js
new sim.Container(
  name: "my-service",
  image: "./my-service",
  containerPort: 8080,
);
```

Migrated the `redis` and `dynamodb` to use this new resource instead of starting/stopping their containers by hand.

### Motivation

Standardize how we manage containers in the Wing Simulator so we can improve the DX across the board and offer a nice UI in the Wing Console for this.

### Misc

Added a couple of utilities:

* `util.md5(dir)` - calculates the md5 hash on all the files in a directory (recursively).
* `util.glob(pattern)` - returns the files matching a glob pattern.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/container.ts
```diff
@@ -0,0 +1,144 @@
...
+import { BaseResourceSchema } from "../simulator/simulator";
+import { IInflightHost, Resource } from "../std";
+
+export const SIM_CONTAINER_FQN = fqnForType("sim.Container");
+export const HOST_PORT_ATTR = "host_port";
+
+/**
+ * Initialization properties for `sim.Container`.
+ */
+export interface ContainerProps {
```
hasanaburayyan, line 20: Im curious about 2 other props and whether they should be included, `restart` and `volumes`. I think its pretty trivial to allow sepcifing the restart option.

Since this is a sim only implementation, allowing `volumes` might be fine.

libs/wingsdk/src/target-sim/container.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * A name of a public Docker image to pull and run or a path to a local directory with a `Dockerfile`.
+   */
+  readonly image: string;
+
+  /**
+   * Internal container port to expose.
+   */
+  readonly containerPort?: number;
```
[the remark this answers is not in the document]
eladb: Yes, it's a common ask and we should add support for it - #5957

libs/wingsdk/src/target-sim/container.ts
eladb: I wanted to keep this PR to a minimum and then we will can add more features later. ok?

eladb: Eventually we might convert this to use `docker compose` and expose all of its features

libs/wingsdk/src/target-sim/container.md
```diff
@@ -0,0 +1,60 @@
...
+  containerPort: 5678,
+  args: ["-text=bang"],
+);
+
+test "send request" {
+  http.get("http://localhost:{c.hostPort}");
+}
+```
+
+There is also support for building containers from a local `Dockerfile`:
```
hasanaburayyan, line 37:
```suggestion
There is also support for building containers from a local directory with a `Dockerfile`:
```

libs/wingsdk/src/target-sim/redis.inflight.ts
```diff
@@ -15,23 +10,15 @@ export class Redis
...
 {
-  private containerName: string;
-  private readonly WING_REDIS_IMAGE =
-    process.env.WING_REDIS_IMAGE ??
-    // Redis version 7.0.9
-    "redis@sha256:e50c7e23f79ae81351beacb20e004720d4bed657415e68c2b1a2b5557c075ce0";
-  private readonly context: ISimulatorContext;
-
-  private connection_url?: string = undefined;
+  private connectionUrl?: string = undefined;
```
hasanaburayyan, line 13: Love the refactor

libs/wingsdk/test/target-sim/dynamodb-table.test.ts
```diff
@@ -146,12 +149,12 @@ test("update item", async () => {
         id: "S",
       },
       hashKey: "id",
+      hostPort: expect.any(String),
     },
     type: ex.DYNAMODB_TABLE_FQN,
   });
   await s.stop();
 
-  expect(listMessages(s)).toMatchSnapshot();
```
hasanaburayyan, line 154: intentional?

hasanaburayyan (approved):
LGTM

libs/wingsdk/test/target-sim/dynamodb-table.test.ts
eladb: yes, because now the messages have some non-deterministic output

libs/wingsdk/src/fs/fs.ts
```diff
@@ -461,6 +610,37 @@ export class Util {
...
+   *
+   * Built with the great `glob` package, based on https://www.npmjs.com/package/glob
+
+   * @param pattern The pattern to match.
+   * @param options Glob options.
+   * @returns List of matching files.
+   */
+  public static glob(pattern: string, options: GlobOptions = {}): string[] {
+    return glob.sync(pattern, options);
+  }
```
Chriscbr, line 624: I feel like we should be careful about blanket-inheriting JavaScript APIs with large existing surface areas. I have nothing against popular JavaScript libraries from an implementation perspective (they're clearly battle tested!) but I imagine their configuration options have acquired a lot of cruft. Regex and globbing are important enough utilities that I reckon there's value designing them well. Can we leave this outside of the PR?

libs/wingsdk/src/fs/fs.ts
```diff
@@ -461,6 +610,37 @@ export class Util {
...
+  public static md5(dir: string, globPattern: string = "**/*") {
+    const hash = crypto.createHash("md5");
+    const files = this.glob(globPattern, { nodir: true, cwd: dir });
+    for (const f of files) {
+      const data = fs.readFileSync(this.join(dir, f));
+      hash.update(data);
+    }
+
+    return hash.digest("hex");
+  }
```
Chriscbr, line 642: md5 taking a directory as a first argument seems like a secondary use case / kind of API, no? Normally I'd expect to directly pass the information I expect. Maybe we can leave this out of the PR...

Chriscbr (commented):
Nice improvement, `sim.Container` should make it easier to write a lot more winglibs 👍

libs/wingsdk/src/fs/fs.ts
eladb: This is `fs.md5()`, not `util.md5()`. Would that make sense more sense?

Chriscbr: ah, got it. Fair enough

libs/wingsdk/src/fs/fs.ts
eladb: I prefer not to punt this. I don't see us investing the time in the foreseeable future on designing this API, and I trust that `glob`, as you say, has been battle tested, so I am comfortable inheriting it.

Let me know if you are okay with this.

eladb: OK, reduced the API to the minimum I think it needed.

Chriscbr (approved):

Chriscbr (approved):

Chriscbr (approved):

Issue #3082, opened 2023-06-26
Simulate resource permissions


Chriscbr:
### Feature Spec

Wing's `sim` target today allows applications to be tested using simulated versions of cloud services. The simulation accurately reflects some of the limitations of deploying cloud applications today (for example, the fact that a resource must be deployed after any resources its configuration depends on), but not others. One noticeable difference is that permissions are not simulated. This means that if the Wing compiler generates classes with incorrect permissions, in many cases the only way to catch them is if you try to deploy and run the application on AWS.

### Use Cases

Discover permission bugs sooner and as part of our end-to-end testing of the Wing language. Most users and contributors assume "if it works on the simulator, then it will work on the cloud", and simulating permissions will make this gap smaller (at a small runtime cost).

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

garysassano:
Maybe a possible solution could be to use [Cedar Agent](https://github.com/permitio/cedar-agent) to manage permissions in the simulator?

garysassano:
Maybe a possible solution could be to use [OPA](https://github.com/open-policy-agent/opa) or [Cedar Agent](https://github.com/permitio/cedar-agent) to manage permissions in the simulator?

Chriscbr:
This has been implemented

PR #6037, opened 2024-03-21
chore(sdk): fix nondeterministic unit test snapshots

Chriscbr:
* Updates some pieces of the simulator to use the HTTP server for communicating with other resources where they weren't previously. This is a small piece of cleanup that will help with the implementation of https://github.com/winglang/wing/issues/3082.
* Fixes an issue where nondeterministic unit test snapshots were added to the SDK (see [here](https://github.com/winglang/wing/commit/885fb3af6e0b057cb893c82467d2f1c3db603bea#r140091022)).

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/schema-resources.ts
```diff
@@ -59,8 +59,8 @@ export interface ApiAttributes {
 
 /** Schema for cloud.Api.props.routes */
 export interface ApiRoute {
-  /** The path to handle. */
-  readonly path: string;
+  /** The HTTP path pattern to handle. */
+  readonly pathPattern: string;
```
Chriscbr, line 63: I changed the name of this field to be more distinctive so in the snapshot sanitizer, I can add it as an exclusion to the absolute-path-removing rule

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -99,9 +99,9 @@ export interface ISimulatorContext {
   readonly serverUrl: string;
 
   /**
-   * Find a resource simulation by its handle. Throws if the handle isn't valid.
+   * Obtain a client given a resource's handle.
    */
-  findInstance(handle: string): ISimulatorResourceInstance;
+  getClient(handle: string): any;
```
MarkMcCulloh, line 104: Was `ISimulatorResourceInstance` inaccurate? If so, maybe we can use `unknown` here to avoid accidental use as `any`.

MarkMcCulloh (approved):

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: I think so, yeah, since the interface currently models the simulator's lifecycle methods (init, cleanup, save) and we don't really want clients to be able to call those anyways. `unknown` sounds ok to me

PR #6039, opened 2024-03-22
chore(docs): compatibility matrix update

monadabot:
chore(docs): update compatibility matrix

[Workflow Run]: https://github.com/winglang/wing/actions/runs/8393479847

------

*Automatically created via the "matrix-update" workflow*


staycoolcall911:
@tsuf239 - It seems like there's an empty change here - can you please check why it happened? Thanks!

tsuf239:
> @tsuf239 - It seems like there's an empty change here - can you please check why it happened? Thanks!

my best guess is that someone probably edited this file manually and Prettier added an extra `\n` at the end of the file
(I'll add one at the auto-generation as well)

tsuf239:
opened a PR https://github.com/winglang/wing/pull/6041
(it might not close this one automatically- since the pr step is running only when there are changes, as far as I know)

Issue #5441, opened 2024-01-15
Invalid JSII library: failed to add property init to class


meirdev:
### I tried this:

Minimal (added by @eladb):

```js
bring "@cdktf/provider-docker" as tfdocker;

new tfdocker.container.Container();
```

```js
// npm i -D @cdktf/provider-docker

bring cloud;

bring "@cdktf/provider-docker" as tfdocker;

pub class Test {
    new() {
        let image = new tfdocker.image.Image(name: "ubuntu");
        let container = new tfdocker.container.Container(image: image.id, name: "test");
    }
}
```

### This happened:

```
Compiler bug during type-checking ('panicked at libs/wingc/src/type_check/jsii_importer.rs:585:22:
Invalid JSII library: failed to add property init to class: TypeError { message: "Symbol \"init\" already defined in this scope", span: WingSpan { start: WingLocation { line: 0, col: 0 }, end: WingLocation { line: 0, col: 0 }, file_id: "" }, annotations: [DiagnosticAnnotation { message: "previous definition", span: WingSpan { start: WingLocation { line: 0, col: 0 }, end: WingLocation { line: 0, col: 0 }, file_id: "" } }] }'), please report at https://www.winglang.io/contributing/start-here/bugs
```

### Anything else?

The relevant parts:

```js
@cdktf/provider-docker.container.ContainerConfig

        {
          "abstract": true,
          "docs": {
            "remarks": "If unset this will default to the `dockerd` defaults.\n\nDocs at Terraform Registry: {@link https://registry.terraform.io/providers/kreuzwerker/docker/3.0.2/docs/resources/container#init Container#init}",
            "stability": "stable",
            "summary": "Configured whether an init process should be injected for this container."
          },
          "immutable": true,
          "locationInModule": {
            "filename": "src/container/index.ts",
            "line": 129
          },
          "name": "init",
          "optional": true,
          "type": {
            "union": {
              "types": [
                {
                  "primitive": "boolean"
                },
                {
                  "fqn": "cdktf.IResolvable"
                }
              ]
            }
          }
        },
```

```js
@cdktf/provider-docker.container.Container

        {
          "docs": {
            "stability": "stable"
          },
          "locationInModule": {
            "filename": "src/container/index.ts",
            "line": 3785
          },
          "name": "init",
          "type": {
            "union": {
              "types": [
                {
                  "primitive": "boolean"
                },
                {
                  "fqn": "cdktf.IResolvable"
                }
              ]
            }
          }
        },
```

[...]

PR #6040, opened 2024-03-23
fix(compiler): can't instantiate preflight classes in static methods

### LINKED ISSUE #6019: Can't instantiate preflight class in a static method
### I tried this:

```wing
class Bar {}

class Foo {
  pub static createBar() {
    new Bar();
  }
}

Foo.createBar();
```

### This happened:

```
Error: Cannot read properties of undefined (reading 'addChild')
   --> temp/main.w:29:7
   | 
   | // B.bla();
   | 
29 | class Bar {}
   |       ^
at new Bar /home/yoav/wing2/temp/main.w:29:7
at createBar /home/yoav/wing2/temp/main.w:33:12
at /home/yoav/wing2/temp/main.w:37:1
```

### I expected this:

to work

### Is there a workaround?

Use closure function instead of static:
```wing
class Bar {}

let createBar = () => {
  new Bar();
};

createBar();
```
or use explicit scope:
```wing
class Bar {}
let root = this;

class Foo {
  pub static createBar() {
    new Bar() in root;
  }
}

Foo.createBar();
```

### Anything else?

The issue is that we default to using `this` as the scope of new resources (`new Bar()`). But `this` isn't available inside a static (in JS it refers to the type in a static context).
In these cases we need to use the parent scope.
This is related to #6016.

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

yoav-steinberg:
Fixed #6019
The fix adds an implicit `$scope` parameter to static methods so when they need to instantiate a preflight class they know from within which context they're being called.

Note the following exceptions to adding the implicit scope param to a static preflight method:
1) JSII imported methods. In the future we might want to support this via #6067.
2) `extern` methods, since the user implements these in JS/TS and doesn't expect an implicit scope parameter.
3) Static methods that override a method of the same name in a parent class that doesn't have an implicit scope argument (because one of the above reasons). We don't have clear definitions of static method overriding, but currently we do use this for overrding `onLiftType` inherited from the imported base `std.Resource` class. Opting out of implicit scoping for this case seemed like the easiest solution.

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -250,7 +250,7 @@ pub enum Type {
 	Enum(Enum),
 }
 
-pub const CLASS_INIT_NAME: &'static str = "init";
+pub const CLASS_INIT_NAME: &'static str = "init"; // TODO: change to "new"??
```
eladb, line 253: I think this should be something like `$new` so that we don't take up this symbol (to get rid of https://github.com/winglang/wing/issues/5441)

libs/wingc/src/type_check.rs
```diff
@@ -4956,11 +4968,34 @@ impl<'a> TypeChecker<'a> {
...
+		let method_sig = method_type
 			.as_mut_function_sig()
-			.expect("Expected method type to be a function")
-			.this_type = instance_type;
+			.expect("Expected method type to be a function");
+
+		// use the class type as the function's "this" type (or None if static)
+		method_sig.this_type = instance_type;
+
+		// For now all* static preflight methods require an implicit scope argument. In the future we may be smart and if
```
eladb, line 4979:
```suggestion
		// For now *all* static preflight methods require an implicit scope argument. In the future we may be smart and if
```

libs/wingc/src/type_check.rs
```diff
@@ -4956,11 +4968,34 @@ impl<'a> TypeChecker<'a> {
...
+		// the method doesn't instantiate any preflight classes then we can do away with it.
+		//
+		// *Special case: if we're overriding a method from a parent class, we assume its `implicit_scope_param`. Note that
```
eladb, line 4982:
```suggestion
		// *Special case*: if we're overriding a method from a parent class, we assume its `implicit_scope_param`. Note that
```

tools/hangar/__snapshots__/test_corpus/valid/baz.w_compile_tf-aws.md
```diff
@@ -24,7 +24,7 @@ class Baz extends $stdlib.std.Resource {
   constructor($scope, $id, ) {
     super($scope, $id);
   }
-  static baz() {
+  static baz($scope, ) {
```
eladb, line 27: aesthetic - can we avoid the trailing `,`?

libs/wingc/src/jsify.rs
```diff
@@ -1405,7 +1458,16 @@ impl<'a> JSifier<'a> {
 		}
 
 		let func_prefix = prefix.join(" ");
-		let mut code = new_code!(&func_def.span, func_prefix, "(", parameters.clone(), ")", arrow, "{");
+		let mut code = new_code!(
+			&func_def.span,
+			func_prefix,
+			"(",
+			implicit_parameters,
```
eladb, line 1465: I am assuming the reason the implicit parameters are before the explicit ones is in order to support variadic arguments, correct?

Worth mentioning

libs/wingc/src/jsify.rs
```diff
@@ -26,7 +26,7 @@ use crate::{
 		is_udt_struct_type,
```
eladb, line 26: Missing tests. Here are a few cases I would check:

* calling a static method from the entrypoint
* calling a JSII static method (that doesn't expect `$scope` as the first argument)
* calling a static method from another static method
* calling a static method from a constructor
* calling a static method from a constructor before calling `super()`

libs/wingc/src/jsify.rs
yoav-steinberg: Yes, and to look similar to ctors and to the construct pattern in general.

libs/wingc/src/type_check.rs
yoav-steinberg: Removed this TODO, since I see we have an issue for it.

examples/tests/valid/new_in_static.test.w
```diff
@@ -27,7 +47,36 @@ let scope = new c.Construct();
...
+  }
+}
+class Foo extends FooParent {
+  new() {
+    super(
+      // TODO: unsupported yet, see https://github.com/winglang/wing/issues/6016
+      // MyClass.createMyClassWithImplicitScope("from-ctor-before-super")
+      nil
+    );
+    MyClass.createMyClassWithImplicitScope("from-ctor");
```
eladb, line 79: I guess in this case we can just pass `$scope` instead of `this`, no?

eladb (approved):
Looks good!

examples/tests/valid/new_in_static.test.w
yoav-steinberg: Right, the challenge is correctly detecting this case and **not** passing `$scope` to all calls after `super()`.

Issue #5958, opened 2024-03-17
Increase coverage using automatic cloud snapshot tests


eladb:
I'd like to be able to capture a snapshot of the output when compiling to cloud targets and use it to protect against regressions in CI (without having to deploy to the cloud).

### RFC

See https://github.com/winglang/wing/pull/6044

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Chriscbr:
I think high level user story makes sense, I'm only unsure about the choice to to automatically take snapshots of every test the user writes. By default, Wing's syntax encourages you to reuse the same test fixtures or preflight components for multiple tests so that you don't have to write the same things over and over. So in the code below, the three tests would all produce identical snapshots, which is not great.

```js
let bucket = new cloud.Bucket();
let api = new cloud.Api();
let dynamodb = new dynamodb.Table();

test "scenario 1" {
  // ...
}

test "scenario 2" {
  // ...
}

test "scenario 3" {
  // ...
}
```

eladb:
> So in the code below, the three tests would all produce identical snapshots, which is not great.

I think a single `.test.w` file should product a single snapshot, not one per test.

Chriscbr:
@eladb Can this be closed per https://github.com/winglang/wing/pull/6163 ?

eladb:
Yes, we still need to run this in winglibs, but that's a different issue.

Issue #1214, opened 2023-01-22
Snapshot tests for `wing test`


Chriscbr:
### Community Note

> Please vote by adding a 👍 reaction to the issue to help us prioritize.
> If you are interested to work on this issue, please leave a comment.

### Feature Spec

Starting in Wing v0.x, `wing test` now saves test snapshots when the `matches_snapshot(json)` function is used:

```js
bring cloud;

new cloud.Function(inflight () => {
  let data = Json { a: 123, b: "boo" };
  assert(matches_snapshot(data));
}) as "test";
```

After the first time running the test, an `app.w.snap` file will be generated containing a sequence of recorded snapshot values associated with each test.

### Use Cases

Asserting that outputs generated by user code is as expected.

See https://blog.janestreet.com/the-joy-of-expect-tests/

### Implementation Notes

Do snapshots for preflight and inflight require different APIs or mechanisms?

Perhaps we can try implementing this in userland first (using `extern`)

### Component

SDK, Development Environment

staycoolcall911:
I believe this was done by @MarkMcCulloh

Chriscbr:
@staycoolcall911 my apologies - I think the issue might have not been explained well. 😅

I rewrote the description to explain how this would be part of Wing's built-in testing functions - let me know if it makes better sense.

eladb:
Snapshot format should be a markdown file that shows the output directory in a beautiful and attractive and GitHub readable format. 

Credits: @MarkMcCulloh

PR #6044, opened 2024-03-24
rfc: cloud snapshot tests

eladb:
An RFC for https://github.com/winglang/wing/issues/5958.

A proposal for a feature that will capture a snapshot of the synthesis output from `wing test` to reduce regressions in cloud implementations.

### DIFF
--- a/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
+++ b/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
@@ -57,6 +57,14 @@ We will also need `wing test --update-snapshot|-U -t tf-aws foo.test.w` to allow
 snapshots, because sometimes it doesn't make practical sense to have to deploy just in order to
 update the snapshot.
 
+The `--snapshots` or `-s` switch can be used to control behavior:
+
+ * `--snapshots=auto` - auto-detect based on CI flag (described above)
+ * `--snapshots=never` - disables snapshots altogether
+ * `--snapshots=dry` - will skip deployment and only update the snapshots.
+ * `--snapshots=wet` - forces a deployment even if `CI=1`.
+ * `--snapshots=assert` - only asserts that the snapshots have not changed
+
 ## Implementation Notes
 
 * We need to decide if we want the inflight JavaScript code to also be included or just trust the
--- a/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
+++ b/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
@@ -48,6 +48,8 @@ When I run `wing test -t tf-aws foo.test.w` in my dev environment, if the test s
 deploys and all test cases pass), the test framework will create `foo.test.w.tfaws.snap` which will
 include a nicely formatted (markdown?) snapshot of all the synthesis output from this test.
 
+> In a future version we can add support for placing all snapshots in an adjacent subdirectory.
+
 Now, when `wing test -t tf-aws foo.test.w` is executed in CI (`CI=1`), by default it won't actually
 go and deploy to the cloud. Instead, it will just synthesize the output and compare it to the
 committed `.snap` file. If the output is not the same, the test will fail with a nice diff
@@ -57,8 +59,8 @@ The `--snapshots` or `-s` switch can be used to control behavior:
 
  * `--snapshots=auto` - auto-detect based on CI flag (described above)
  * `--snapshots=never` - disables snapshots altogether
- * `--snapshots=dry` - will skip deployment and only update the snapshots.
- * `--snapshots=wet` - forces a deployment even if `CI=1`.
+ * `--snapshots=update` - skips deployment and only update the snapshots.
+ * `--snapshots=deploy` - forces a deployment even if `CI=1`.
  * `--snapshots=assert` - only asserts that the snapshots have not changed
 
 ## Implementation Notes
--- a/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
+++ b/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
@@ -53,10 +53,6 @@ go and deploy to the cloud. Instead, it will just synthesize the output and comp
 committed `.snap` file. If the output is not the same, the test will fail with a nice diff
 indicating that there was an unexpected regression.
 
-We will also need `wing test --update-snapshot|-U -t tf-aws foo.test.w` to allow only updating the
-snapshots, because sometimes it doesn't make practical sense to have to deploy just in order to
-update the snapshot.
-
 The `--snapshots` or `-s` switch can be used to control behavior:
 
  * `--snapshots=auto` - auto-detect based on CI flag (described above)
--- a/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
+++ b/docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
@@ -1,0 +1,69 @@
+---
+title: "#5958 Cloud Snapshot Tests"
+description: Spec and design for adding cloud snapshots to the Wing test framework
+---
+
+# Cloud Snapshot Tests
+
+- **Author(s)**: @eladb
+- **Submission Date**: 2024-03-24
+- **Stage**: Proposal
+- **Stage Date**: 2024-03-24
+
+## Requirements
+
+We would like to be able to capture a snapshot of the synthesized output when compiling to cloud
+targets, and use it to protect against regressions in CI, without having to deploy to the cloud.
+
+For example, let's say I am creating a Wing library with a resource that supports both `sim` and
+`tf-aws`. I am writing a bunch of tests for it. 
+
+I can run these tests in the simulator (`wing test`) and can also run them on AWS through `wing test
+-t tf-aws`.
+
+As my code **or its dependencies** are updated, I'd like to continuously run all these tests to make
+sure my library is not broken. Running the tests against `sim` is easy (I can just run `wing test`
+in my CI build environment), but it's not realistic to run all these tests against the real cloud
+target, given it could take a very long time and could also be very expensive.
+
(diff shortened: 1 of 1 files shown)


docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
hasanaburayyan, line 1: I think this feature probably makes sense to have. I cant say I have strong opinions on the design, but I like what is being proposed

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
```diff
@@ -0,0 +1,69 @@
...
+> This is a similar approach used successfully and scalably in the AWS CDK. See
+> [docs](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md) for details. The main
+> difference is that `cdk-integ` there was no way to add assertions to tests, so this is so much
+> cooler.
+
+When I run `wing test -t tf-aws foo.test.w` in my dev environment, if the test succeeds (e.g.
+deploys and all test cases pass), the test framework will create `foo.test.w.tfaws.snap` which will
+include a nicely formatted (markdown?) snapshot of all the synthesis output from this test.
+
+Now, when `wing test -t tf-aws foo.test.w` is executed in CI (`CI=1`), by default it won't actually
```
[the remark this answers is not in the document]
eladb: added

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
```diff
@@ -0,0 +1,73 @@
...
+committed `.snap` file. If the output is not the same, the test will fail with a nice diff
+indicating that there was an unexpected regression.
+
+The `--snapshots` or `-s` switch can be used to control behavior:
+
+ * `--snapshots=auto` - auto-detect based on CI flag (described above)
+ * `--snapshots=never` - disables snapshots altogether
+ * `--snapshots=dry` - will skip deployment and only update the snapshots.
+ * `--snapshots=wet` - forces a deployment even if `CI=1`.
+ * `--snapshots=assert` - only asserts that the snapshots have not changed
```
hasanaburayyan, line 62: maybe its just me but something feels off about `wing test -t tf-aws foo.test.w --snapshot=wet` since its the deployment aspect that is `wet` or `dry`.  Its essentially the "test" that is being performed in a dry or wet manner, not the snapshot.

I dont know If I have a better suggestion off the top of my head, but I almost feel like `wing test -t tf-aws foot.test.w --snapshot` should be the flag to force snapshot mode otherwise the deployment occurs.

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Implementation Notes
+
+* We need to decide if we want the inflight JavaScript code to also be included or just trust the
+  fact that if inflight code has changed, it should be reflected somehow in the asset names
+  referenced from the IAC model.
```
hasanaburayyan, line 68: Should this decision be made as part of the RFC?

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
eladb: I see what you are saying... 

I played around with the idea of a boolean flag but since there are multiple modes  here, you need more than just a boolean...

So we need a way to model all of these modes somehow... Any suggestion?

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
eladb: I think that initially we will just include the IAC code and see how it plays out.

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
hasanaburayyan: > So we need a way to model all of these modes somehow... Any suggestion?

hmm, not super sure. I mean the first few things that come to mind are either building a snapshot framework/library and control snapshots similar to how it works in jest.

Or maybe this is easier to model from a command perspective, `wing snapshot` 

Dunno need to think about it some

MarkMcCulloh (approved):
There probably needs to be more consideration intertwined with things like https://github.com/winglang/wing/pull/5576. When we move forward on that we should take this snapshotting into account.

I think there's enough here to move forward though. Especially since the hard part of this is the snapshot implementation itself which probably won't care much about how/where the lines are drawn for what a `test` is.

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
```diff
@@ -0,0 +1,73 @@
...
+We propose to add a feature to `wing test` which will capture a snapshot of the synthesized output
+of every `.test.w` file and store them in the repository.
+
+> This is a similar approach used successfully and scalably in the AWS CDK. See
+> [docs](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md) for details. The main
+> difference is that `cdk-integ` there was no way to add assertions to tests, so this is so much
+> cooler.
+
+When I run `wing test -t tf-aws foo.test.w` in my dev environment, if the test succeeds (e.g.
+deploys and all test cases pass), the test framework will create `foo.test.w.tfaws.snap` which will
```
Chriscbr, line 48: Would it make sense to put snapshot files in a `snapshots/` directory? I imagine if a project has multiple test files it could get inconvenient having all of the files intermixed with source files (sort of like `.ts`, `.js`, and `.d.ts` files intermixed in a TypeScript project).

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
```diff
@@ -0,0 +1,73 @@
...
+in my CI build environment), but it's not realistic to run all these tests against the real cloud
+target, given it could take a very long time and could also be very expensive.
+
+In some scenarios (such as the Wing SDK tests), users will prefer to setup a continuous integration
+workflow and run tests in the cloud, but for the most part running all my tests on the cloud for
+every commit is too expensive and requires a complicated setup (i.e. an AWS account), so I'd like to
+use a more lightweight approach...
+
+> NOTE: This is not https://github.com/winglang/wing/issues/1214 (which is about using snapshots for
+> assertions).
```
Chriscbr, line 35: Can you explain how the use case / problems solved by #1214 is different than the use case / problems addressed in the RFC? It seems to me like they're overlapping - maybe you can add some detail how they're different (both read to me like they're mechanisms that can be used to take snapshots of IaC configuration to make a test succeed or fail).

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
```diff
@@ -0,0 +1,73 @@
...
+go and deploy to the cloud. Instead, it will just synthesize the output and compare it to the
+committed `.snap` file. If the output is not the same, the test will fail with a nice diff
+indicating that there was an unexpected regression.
+
+The `--snapshots` or `-s` switch can be used to control behavior:
+
+ * `--snapshots=auto` - auto-detect based on CI flag (described above)
+ * `--snapshots=never` - disables snapshots altogether
+ * `--snapshots=dry` - will skip deployment and only update the snapshots.
+ * `--snapshots=wet` - forces a deployment even if `CI=1`.
```
Chriscbr, line 61: How would these snapshot flags interact with a more general purpose snapshot mechanism (like #1214) that doesn't require deploys?

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
eladb: Not sure that's the same thing as .js/.d.ts files because those are usually not checked into the repository.

Personally, it feels more discoverable that snapshots are side-by-side with their origin. We can add a feature later to put them under a subdirectory if that comes up.

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
eladb: I'll add some details, but they are simply different use cases. #1214 is about adding a general mechanism to write snapshot-based assertions in tests that developers author (e.g. I want to assert that response returned from a service hasn't changed). This RFC is about asserting that the Wing compiler outputs (IAC configs) haven't regressed after they've been tested in the cloud.

But I can see how this could be confusing and I am wondering if maybe we can either give the snapshots this RFC refers to a different name (any suggestions?) or change the CLI switch names to `--cloud-snapshots` or something like that, so that when we introduce #1214 we can leverage the `--snapshot` switch for them (because that would be more familiar).

Maybe we can call these "synthshots"?

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
hasanaburayyan: Eh maybe I’m over analyzing on this datapoint I’m fine to try this and see

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
eladb: See above

Chriscbr (approved):

hasanaburayyan (approved):

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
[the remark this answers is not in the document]
eladb: Cool, I'll change

docs/contributing/999-rfcs/2024-03-24-snapshot-tests.md
eladb: I'll keep the term "snapshots" for now.

PR #6047, opened 2024-03-24
chore: commit grammar.json in tree-sitter-wing

### LINKED ISSUE #6043: Tree-sitter: consider committing `grammar.json`
### Use Case

Install the grammar via `nvim-treesitter` (1.0, in development).

### Proposed Solution

While keeping `parser.c` out of version control is reasonable (and in fact recommended by upstream), the generated `grammar.json` is much smaller and has meaningful diffs. The benefit of keeping this in the repo is that downstream users (like nvim-treesitter) do not require `node` for building and can just do `tree-sitter g grammar.json && tree-sitter build -o parser.so` (which is what nvim-treesitter is transitioning to for 1.0).

### Implementation Notes

In addition, consider making proper releases including the generated artifacts (which is what nvim-treesitter will require for "stable" parsers). Upstream is in the process of adding reusable workflows and `tree-sitter` support to make this easier.

### Component

tree-sitter-wing

### Community Notes

MarkMcCulloh:
Fixes #6043

Grammar updates aren't happening too frequently, so this hopefully shouldn't cause much trouble.

### DIFF
--- a/libs/tree-sitter-wing/.gitattributes
+++ b/libs/tree-sitter-wing/.gitattributes
@@ -1,0 +1,1 @@
+/src/grammar.json linguist-generated
--- a/libs/tree-sitter-wing/.gitignore
+++ b/libs/tree-sitter-wing/.gitignore
@@ -1,4 +1,5 @@
 /src/*
 !/src/scanner.c
+!/src/grammar.json
 /log.html
 /*.wasm
--- a/libs/tree-sitter-wing/src/grammar.json
+++ b/libs/tree-sitter-wing/src/grammar.json
@@ -1,0 +1,4567 @@
+{
+  "name": "wing",
+  "word": "identifier",
+  "rules": {
+    "source": {
+      "type": "REPEAT",
+      "content": {
+        "type": "SYMBOL",
+        "name": "_statement"
+      }
+    },
+    "block": {
+      "type": "CHOICE",
+      "members": [
+        {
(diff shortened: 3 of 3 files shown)


libs/tree-sitter-wing/.gitattributes
```diff
@@ -0,0 +1 @@
+/src/grammar.json linguist-generated
```
staycoolcall911, line 1: I added the pr/e2e-full label on this PR, but maybe it shouldn't matter? If that's not needed, you can remove it before you merge that in.

staycoolcall911 (approved):

libs/tree-sitter-wing/.gitignore
```diff
@@ -1,4 +1,5 @@
 /src/*
 !/src/scanner.c
+!/src/grammar.json
```
eladb, line 3: will this get automatically updated by the self mutation in case it's changed?

eladb (approved):

libs/tree-sitter-wing/.gitignore
MarkMcCulloh: Yep it should

libs/tree-sitter-wing/.gitattributes
MarkMcCulloh: Shouldn't be needed, the file already exists in the built version

PR #6048, opened 2024-03-24
chore: deprecate `LegacySandbox`

eladb:
Migrate `cloud.Service` to use `Sandbox` instead of `LegacySandbox`.

Since we can't transmit a JavaScript function across the process boundary, I've modified the service entrypoint to expose `start()` and `stop()` methods which are called by `cloud.Service` in response to the respective lifecycle events.

For now, I've kept the contract of `cloud.Service` in tact, but I think we will want to change it so it accepts an object that has an `inflight start()` and and `inflight stop()` methods, but this is left for a future PR.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/cloud/service.ts
```diff
@@ -95,16 +95,25 @@ export class Service extends Resource implements IInflightHost {
     const lines = new Array<string>();
 
     lines.push('"use strict";');
-    lines.push("let $obj;");
-
-    lines.push("async function $initOnce() {");
-    lines.push(`  $obj = $obj || (await (${inflightClient}));`);
-    lines.push("  return $obj;");
+    lines.push("let $stop;");
```
MarkMcCulloh, line 98: style nit: imo it's much easier to read and write multiline templates. It's probably slightly faster at runtime too.

```js
`\
"use strict";
let $stop;
exports.start = async function() {
  if ($stop) {
    throw Error('service already started');
  }
  $stop = (await ((await ${inflightClient})).handle())) ?? (() => {});
};
exports.stop = async function() {
  if (!$stop) {
    throw Error('service not started');
  }
  await $stop();
  $stop = undefined;
};",`
```

libs/wingsdk/src/target-sim/service.inflight.ts
```diff
@@ -48,6 +39,7 @@ export class Service implements IServiceClient, ISimulatorResourceInstance {
   }
 
   public async cleanup(): Promise<void> {
+    await this.createBundlePromise;
```
MarkMcCulloh, line 42: Should this be inside `stop`? Maybe right before the call into the sandbox

libs/wingsdk/src/target-sim/service.inflight.ts
```diff
@@ -59,41 +51,57 @@ export class Service implements IServiceClient, ISimulatorResourceInstance {
...
-      // context. so we need to handle both options here. (in wing this is handled by the compiler).
-      if (typeof this.onStop === "function") {
-        await (this.onStop as any)();
-      } else {
-        await this.onStop.handle();
-      }
+    try {
+      await this.sandbox.call("stop");
+      await this.sandbox.cleanup();
+      this.running = false;
```
MarkMcCulloh, line 88: Maybe set running to false first in case stop is quickly called twice

```suggestion
      this.running = false;
      await this.sandbox.call("stop");
      await this.sandbox.cleanup();
```

MarkMcCulloh (approved):

PR #6051, opened 2024-03-25
fix(cli): handle shutdown events gracefully

skyrpex:
Previously, the `wing it` didn't wait for the Console to close and release its resources.


skyrpex:
⚠️ I'm not sure if this is necessary at all, though.

skyrpex:
I tried running an application that uses the dynamodb winglib. With the changes in this PR, the docker containers get stopped and removed correctly after Ctrl+C. Without the changes, the containers remain active.

eladb (approved):

Issue #5305, opened 2023-12-22
`sim.Resource` and `sim.Attribute`


Chriscbr:
### Use Case

Before (the API we have today):
```js
class MyService {
  state: sim.State;
  pub startTime: str;
  
  new() {
    this.state = new sim.State();

    new cloud.Service(inflight () => {
      let res = // create a websocket, etc.
      this.state.set("start_time", "2023-10-16T20:47:39.511Z");
      return () => { res.close(); };
    });

    this.startTime = this.state.token("start_time");
  }
}
```

After (proposed API):
```js
class MyService {
  state: sim.State;
  pub startTime: str;
  
  new() {
    this.state = new sim.State(inflight (ctx) => {
      let res = // create a websocket, etc.

      ctx.trace("custom message"); // new!
      fs.write(fs.join(ctx.statedir, "state.json"), "data"); // new!
      
      return {
        attributes: {
          "start_time": "2023-10-16T20:47:39.511Z",
        },
        teardown: () => { res.close(); },
      };
    });

    this.startTime = this.state.token("start_time");
  }
}
```

<details>
<summary>Alternative API</summary>

```js
class MyService extends sim.State {
  pub startTime: str;
  new() {
    this.startTime = this.token("start_time");
  }

  inflight res: Websocket;
  
  init(ctx: sim.Context): Map<str> {
    this.res = // create a websocket, etc.

    ctx.trace("custom message"); // new!
    fs.write(fs.join(ctx.statedir, "state.json"), "data"); // new!

    return Map<str>{ "start_time": "2023-10-16T20:47:39.511Z" };
  }

  teardown() {
    this.res.close();
  }
}
```

</details>

Benefits:
- Using simulator logger
- Setting/restoring simulator state for the resource
- Values can still be shared between init and teardown functions (for opening and closing long-lived objects like websockets or database connections)
- If sim.State returns a map without all the runtime attributes set (like `start_time`), we could detect the error sooner in the simulator

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

eladb:
Would it make sense to simply add this to the `cloud.Service`?

Chriscbr:
I wasn't sure -- here was my thinking:

[...]

PR #6079, opened 2024-03-27
fix(sdk): lift/permissions errors are not caught in simulator

### LINKED ISSUE #6002: sim resource should throw exception when operation was not lifted
### Use Case

As part of our guarantee that `if it works on your machine it works on the cloud` we have introduced a runtime error when a host is accessing an un-lifted operation. 

# Consider the following code

### `main.w`
```wing
bring cloud;
bring util;

let b = new cloud.Bucket();

class C {
  extern "./ex.ts" inflight pub static _writeToBucket(b: cloud.Bucket, file: str, content: str): void;
}
let liftAndWrite = new cloud.Function(inflight () => {
  lift(b, ["put"]);
  C._writeToBucket(b, "file.txt", "hello wing");
  }) as "liftAndWrite";
  
let justWrite = new cloud.Function(inflight () => {
  C._writeToBucket(b, "file.txt", "hello wing");
}) as "justWrite";

test "liftAndWrite should work" {
  liftAndWrite.invoke();
  util.waitUntil(() => {
    return b.exists("file.txt");
  }, timeout: 10s);
  assert(b.exists("file.txt"));
}

test "justWrite should not work" {
  justWrite.invoke();
  util.waitUntil(() => {
    return b.exists("file.txt");
  }, timeout: 10s);
  assert(!b.exists("file.txt"));
}
```
### `ex.ts`
```ts
import type extern from "./ex.extern";

export const _writeToBucket: extern["_writeToBucket"] = async (b, file, content) => {
  try {
    await b.put(file, content);
  } catch {
    
  }
}
```

## Testing on AWS 
```sh
% wing test  -t tf-aws main.w
✔ Compiling main.w to tf-aws...

✔ terraform init

✔ terraform apply

✔ Setting up test runner...

✔ Running tests...

pass ─ main.tfaws.630318 » root/Test.m4-Gz3rJlJ/env0/test:liftAndWrite should work 
pass ─ main.tfaws.630318 » root/Test.m4-Gz3rJlJ/env1/test:justWrite should not work
✔ terraform destroy

 
 
Tests 2 passed (2)
Test Files 1 passed (1)
Duration 1m43.62s
```
## Testing on sim
```sh
% wing test  main.w          
pass ─ main.wsim » root/env0/test:liftAndWrite should work 
fail ┌ main.wsim » root/env1/test:justWrite should not work
     │ Error: assertion failed: !b.exists("file.txt")
     │    --> main.w:31:3
     │    | util.waitUntil(() => {
     │    |   return b.exists("file.txt");
     │    | }, timeout: 10s);
     │ 31 | assert(!b.exists("file.txt"));
     │    | ^
     └ at handle /private/tmp/ts/main.w:31:3
 
 
Tests 1 failed | 1 passed (2)
Test Files 1 failed (1)
Duration 0m0.81s
```

### Proposed Solution

Throw an exception when an operation is not lifted in sim

### Implementation Notes

_No response_

### Component

Compiler, SDK

### Community Notes

Chriscbr:
Fixes #6002

To support the needs of cloud applications with many resources, Wing's compiler performs static analysis to identify all of the infrastructure resources that are used by different bits of runtime code. (In Wing parlance these are usually called preflight objects, and inflight functions, respectively). By identifying these relationships, the compiler can automatically generate applications whose resources have least-privileged access. For example, database read/write permissions only need to be added to the compute resources that perform queries on it.

This analysis is sufficient for many programs. But in general, obtaining the minimal set of resources needed for arbitrary inflight code is, in the worst case, as hard as the [halting problem](https://en.wikipedia.org/wiki/Halting_problem). To illustrate, the inflight code below only needs access to the preflight object `buckets[0]`, but it's impossible for any reasonable compiler to deduce this.

```js
bring cloud;

let buckets = [
  new cloud.Bucket() as "b1",
  new cloud.Bucket() as "b2"
];

  let var i = 10;
  while i > 0 {
    i -= 1;
  }
  buckets.at(i).put("key", "value");
});
```

Even if Wing's analyzer did provide a [covering set](https://en.wikipedia.org/wiki/Set_cover_problem) of the resources and permissions needed by inflight code (in other words, a set of permissions that might be broader than necessary), there are cases when users will want or need to scope down permissions.

https://github.com/winglang/wing/pull/5935 addresses some of these needs by letting the user annotate functions with additional lifting information:

```js
  lift(buckets.at(1), ["put"]);
  let var i = 10;
  while i > 0 {
    i -= 1;
  }
  buckets.at(i).put("key", "value");
});
```

This was an important change to unblock many use cases. Unfortunately, a side effect is that we've lost the guarantee that every program that compiles and works in the Wing simulator will also work in the cloud. That's because the simulator doesn't simulate permissions -- so even if bad `lift()` annotations are provided by the user, a program will still work in the simulator. Ouch!

As a rule of thumb, we want to maximize the extent to which Wing programs working on the simulator result in the same programs working on the cloud. Even though we can't catch these errors at compile time, they can be detected at runtime. To that end, this PR implements a basic permissions model within the Wing simulator. Whenever a resource tries calling a method on another resource, the API call will first be checked against a list of permissions (generated during preflight) to see if the request should go through. If not, an error will be raised to the caller. The end result is that as long as you're able to test the code path with a unit test, it should still work locally.

Here's an example of the error that will be surfaced to the user during `wing test` or in the Wing Console:

```
Error: Resource "root/Default/Function" does not have permission to perform operation "put" on resource "root/Default/b1".
   --> bucket.main.w:14:3
   | while i > 0 {
   |   i -= 1;
   | }
14 | buckets.at(i).put("key", "value");
   | ^
at /Users/chrisr/dev/wing-test/bucket.main.w:14:3
```

-----

### Implementation notes

**The good:** The main implementation is built around a new resource named `sim.Policy`. It stores a "principal" (a resource the policy applies to) and a list of statements (a list of resources and method names). It's probably not the only way to model this, but it felt like a decent starting point given that permissions are also represented with resources within CloudFormation and Terraform on most major cloud providers.

Policy checking is also straightforward. All requests between resources currently go through the simulator's internal HTTP server. We've added an additional piece of information that identifies who is the resource that's performing the method call, which is sent to the server.

**The mid:** How to store the information from `sim.Policy` in the simulator is an open question. This PR models `sim.Policy` like a resource (why not?). Except, nobody communicates with the policy directly. Instead, when the simulator is initializing or shutting down resources, if it sees a resource is a policy, it saves the resolved policy information into a policy registry. When we're checking if an API call should be allowed or not, it performs a lookup in the policy registry.

But does that mean `sim.Policy` has no effect on other resources? Not quite. First, an aside...

> Within the simulator, a resource can be identified in two ways. The first is with their construct path (a string like `root/MyApp/Storage/Bucket`), and the second is with a simulation-unique ID aka "handle" (a string like `sim-42`), which serves as an analog to physical resource names on cloud providers.

> To make this concrete, imagine you have a Wing app with a `sim.Container`, and you modify the container's image. When the simulator is updating to the new version of your app, the construct path for the resource will stay the same, but the resource itself needs to be destroyed and recreated, which results in a fresh simulation ID.

> Today this difference between these IDs isn't strictly necessary - it's possible we could use construct paths everywhere in the simulator. But using these IDs helps us detect dependency relationships in the resource graph, and if we wanted to model other kinds of application updates in the simulator (for example, Terraform's [`create_before_destroy` property](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#create_before_destroy)), then this distinction suddenly becomes very important. So for now I'm trying to keep it.

Each `sim.Policy` refers to resources via their simulation handles, as we do elsewhere in the simulator. But this means a `sim.Policy` resource can't be created until all of the resources it refers to have been created. This introduces some interesting edge cases (for example, how does a `cloud.Service` call other resources during start-up when its `sim.Policy` can't be created until `cloud.Service` has finished starting up??) I managed to solve most of these cases with auxiliary resources that run after the initial resource's creation.

**The bad:** In order for resources to make requests to the simulator server with their own handles, they need to know their own handles. Unfortunately, to expose this information in `ISimulatorContext`, I needed to refactor inflight simulation classes so that simulator context is passed to `init()` instead of to the JavaScript constructor. This results in some uglier code, but I don't think the distinction between constructors and `init()` will be relevant once more sim resources are implemented in Wing.

## Checklist

- [ ] Tests added (always)


Chriscbr:
(working on adding some tests)

yoav-steinberg (commented):
This seems very nice.
One thing I'm not sure of is the `sim.Policy` resource. I'm not sure how the mechanics of configuring the the simulator work, but it feels a bit awkward to create these resources just so we can use their configuration to set up the sim permission registry.
Couldn't we just store the statements in the resources themselves instead of adding the `sim.Policy` resource to each one?

Chriscbr:
> Couldn't we just store the statements in the resources themselves instead of adding the sim.Policy resource to each one?

I considered the approach, but I thought there could be other issues with this direction. I'll try to add some details.

For context, when you synthesize an app to the `sim` target, one of the main artifacts produced is a `simulator.json` file that lists each resource and some configuration for it. It looks like this:

```json
{
  "type": "@winglang/sdk.cloud.Topic",
  "path": "root/Default/Topic",
  "props": {}, // put permissions in here instead?
},
{
  "type": "@winglang/sdk.sim.Policy",
  "path": "root/Default/Topic/Policy",
  "props": {
    "principal": "${wsim#root/Default/Topic#attrs.handle}",
    "statements": [
      {
        "resourceHandle": "${wsim#root/Default/Topic/OnMessage0#attrs.handle}",
        "operation": "invokeAsync"
      }
    ]
  },
},
```

Let's take an example of a `cloud.Topic`. When someone writes `topic.onMessage(<inflight code>)` in Wing, a `cloud.Function` is created, and the topic needs permissions to call the function.

Instead of creating a separate Policy resource, we could have tried to inline the permissions for `Topic` directly within the configuration for `Topic`. But the permissions mention other resources, so it would introduce a dependency where the Topic can't be deployed until after the Function is deployed, which means for example that user code wouldn't be allowed to publish messages to the topic without creating a dependency cycle. (Maybe not allowing a topic listener to publish to the same topic is a reasonable restriction - but it's perfectly legit to do this on AWS.)

It's possible we could have worked around this by adding some exception that says that permission statements / information shouldn't be treated as dependencies, but my hunch is that might cause problems later. Hard to say for sure.

yoav-steinberg:
> It's possible we could have worked around this by adding some exception that says that permission statements / information shouldn't be treated as dependencies, but my hunch is that might cause problems later. Hard to say for sure.

I see how these are logically dependencies, but in practice what does this mean? Aren't these just strings (paths) which you check during permission checking on the server side of the Sim HTTP interface? If a `Topic` tries to invoke a `Function` that doesn't exist yet, you probably return an error anyways, so how does this introduce a new issue?

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -461,35 +526,35 @@ describe("in-place updates", () => {
...
       "root/Api1/Endpoint started",
+      "root/Api1/Policy started",
       "root/Bucket1 started",
+      "root/Bucket1/Policy started",
     ]);
 
-    const urlAfterUpdate = await (
-      sim.getResource("root/Bucket1") as IBucketClient
-    ).get("url.txt");
-    expect(urlAfterUpdate).not.toEqual(urlBeforeUpdate);
```
MarkMcCulloh, line 492: Was this part of the test suddenly (checking that the contents of the bucket are unchanged) broken?

libs/wingsdk/src/target-sim/service.ts
```diff
@@ -46,11 +63,59 @@ export class Service extends cloud.Service implements ISimulatorResource {
...
+ *
+ * Suppose a service puts an object in a bucket when it starts. The policy
+ * (sim.Policy) that grants the service bucket permissions only takes effect
+ * after the service is created. Because of this, the service needs to be
+ * started after both the service and the policy are created. Vice versa, the
+ * service needs to be stopped before the policy is deleted.
+ *
+ * @internal
+ */
+export class ServiceHelper extends Resource implements ISimulatorResource {
```
MarkMcCulloh, line 97: I feel like this should be avoidable but I can't think of a good way. Do you think requiring this is related to a more fundamental design issue and/or is it just a reasonable tiny bit of indrection that should be expected?

MarkMcCulloh (approved):
Nice approach and thoughtful execution 👍

libs/wingsdk/test/simulator/simulator.test.ts
Chriscbr: I noticed this test was printing some errors when I ran vitest, and when I dug into it, it was because the test clears the state directory while the simulator is still running, causing errors when the resources save their state to a non-existent directory. I'm not sure that part of the test made sense, so I removed them.

libs/wingsdk/src/target-sim/service.ts
Chriscbr: Not sure - I suspect we'll learn more once we implement https://github.com/winglang/wing/issues/5305. If we see these helpers become common then it I'd want to revisit this design, or probably just extend it with some way to model inline permissions.

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -568,9 +634,24 @@ export class Simulator {
       });
       req.on("end", () => {
         const request: SimulatorServerRequest = deserialize(body);
-        const { handle, method, args } = request;
+        const { caller, handle, method, args } = request;
         const resource = this._handles.tryFind(handle);
 
+        // Check if the caller has permission to call the method on the resource
```
eladb, line 640: very powerful that we have this proxy!

libs/wingsdk/src/target-sim/app.ts
```diff
@@ -137,6 +143,9 @@ export class App extends core.App {
       case SERVICE_FQN:
         return require.resolve("./service.inflight");
 
+      case SERVICE_HELPER_FQN:
```
eladb, line 146: ??

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -12,27 +12,23 @@ import { Util } from "../util";
...
 
-  private log(message: string) {
-    this.context.addTrace({
-      data: { message },
-      sourcePath: this.context.resourcePath,
-      sourceType: "container",
-      timestamp: new Date().toISOString(),
-      type: TraceType.LOG,
-    });
+  private get context(): ISimulatorContext {
```
eladb, line 23: Kind of feels like there might be some value in creating a base class for SimulatorResourceInstance that allows reusing `context`, `log`, etc.

libs/wingsdk/src/target-sim/service.ts
```diff
@@ -46,11 +63,59 @@ export class Service extends cloud.Service implements ISimulatorResource {
...
+/** @internal */
+export const SERVICE_HELPER_FQN = fqnForType("sim.ServiceHelper");
+
+/** @internal */
+export interface ServiceHelperProps {
+  readonly service: Service;
+  readonly autoStart: boolean;
+}
+
+/**
```
eladb, line 84: This doesn't feel right and makes me wonder if modeling policy as a resource is the right choice. Why not just add a `policy` attribute in the simulator schema to any inflight host (compute) resource (service, function, on-deploy) directly?

Not sure I fully understand the value of modeling policies as resources. What's the motivation?

eladb (changes requested):

eladb (approved):
@Chriscbr I don't want to block this PR, but I do feel like there is some design "smell" there in how policies are modeled. Having said that I don't think it's critical :-)

libs/wingsdk/src/target-sim/service.ts
Chriscbr: https://github.com/winglang/wing/pull/6079#issuecomment-2023734113

PR #6089, opened 2024-03-28
fix(sdk): cloud.Service shutdown code is ignored when Wing Console is stopped

### LINKED ISSUE #6018: "dangling" resources
### I tried this:

`myscript.sh`:

```bash
#!/usr/bin/env bash

python -m http.server $1
```

`dangling.js`:

```javascript
const child_process = require("node:child_process");

// npm i tree-kill
const kill = require("tree-kill");

exports.processEnv = () => {
  return { ...process.env };
};

exports.run = (port, env) => {
  const proc = child_process.spawn("./myscript.sh", [`${port}`], {
    env,
    detached: true,
  });

  proc.on("error", (error) => {
    console.log(error);
  });

  proc.stdout.setEncoding("utf8");
  proc.stdout.on("data", (data) => {
    console.log(data);
  });

  proc.stderr.setEncoding("utf8");
  proc.stderr.on("data", (data) => {
    console.log(data);
  });

  return () => {
    kill(proc.pid, "SIGKILL");
  };
};
```

`main.w`:

```typescript
bring cloud;
bring util;

class Dangling {
  static extern "./dangling.js" processEnv(): MutMap<str>;
  static extern "./dangling.js" inflight run(port: num, env: MutMap<str>): inflight(): void;

  new() {
    if !nodeof(this).app.isTestEnvironment {
      let env = Dangling.processEnv();

        return Dangling.run(8766, env);
      });
    }
  }
}

new Dangling();
```

### This happened:

The first time I run the code the server running from the script works as I expected.

Because there is no graceful way to close the `wing it main.w` except `Control+C,` the server is still running in the background, you can verify this with:

`ps aux | grep http.server`

And the next time I run `wing it main.w` I get the error:

`OSError: [Errno 48] Address already in use`

### I expected this:

For cleanup functions registered in `cloud.Service` to run when exiting the program.

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

0.61.18

### Node.js Version

v20.11.0

### Platform(s)

MacOS

### Community Notes

Chriscbr:
Fixes #6018

If the Wing Console is terminated by Ctrl + C, it gracefully waits for resources to be shut down (thanks to https://github.com/winglang/wing/pull/6051). However there's a latent bug since the kill signal is still propagated to child processes, like those running inflight code for cloud.Service. This changes the behavior so the kill signal is longer propagated, and the simulator will be fully responsible for shutting down its child processes.

If this doesn't solve all our issues (or runaway processes occur in some other ways), then we might need some kind of daemon for tracking processes created by the Wing simulator - but I'm hoping we might be able to avoid this kind of complexity for now.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/shared/sandbox.ts
+++ b/libs/wingsdk/src/shared/sandbox.ts
@@ -140,6 +140,10 @@ process.on("message", async (message) => {
     this.child = cp.fork(this.entrypoint, {
       env: childEnv,
       stdio: "pipe",
+      // keep the process detached so in the case of cloud.Service, if the parent process is killed
+      // (e.g. someone presses Ctrl+C while using Wing Console),
+      // we can gracefully call any cleanup code in the child process
+      detached: true,
       // this option allows complex objects like Error to be sent from the child process to the parent
       serialization: "advanced",
     });
--- a/libs/wingsdk/test/simulator/cleanup.test.ts
+++ b/libs/wingsdk/test/simulator/cleanup.test.ts
@@ -1,0 +1,91 @@
+import * as cp from "child_process";
+import { expect, test } from "vitest";
+import { Service } from "../../src/cloud";
+import { Testing } from "../../src/simulator";
+import { SimApp } from "../sim-app";
+
+const script = (simdir: string) => `
+const { simulator } = require("./");
+
+async function main() {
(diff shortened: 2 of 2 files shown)


MarkMcCulloh (approved):

eladb (commented):
nice

PR #6099, opened 2024-03-30
fix!: refactor platforms to expose global class factory

Chriscbr:
[Wing Platforms](https://www.winglang.io/docs/concepts/platforms) allow users to substitute in their own class definitions whenever a class is used in Wing code. To generate classes in JavaScript that extend a class defined by a Wing platform, the Wing compiler needs to generate code shaped something like this:

```
class Child extends getParentType() { ... }
```

where `getParentType()` is just some expression that obtains the parent class type from the Wing platform system. Surprisingly, JavaScript allows for putting expressions like function calls here (which is cool).

The compiler currently generates some code that no-ops, so classes defined by Wing platforms are never injected. We would like to instead generate code that accesses the current Wing platform, but it isn't possible because the only ways to obtain the platform are through the `App`.

To make a fix for this possible, this PR refactors our platform-related code so that we're now exposing the dependency injection factory through the [JavaScript global object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis). Wing generated code can obtain the target-specific version of any class it needs (if an implementation is available) by calling `globalThis.$ClassFactory.new(fqn, scope, id, ...args)`.

The unfortunate drawback of this change is that it pollutes the global namespace, and it means only a single Wing platform can be used for synthesizing apps in a given Node process. (If you want to compile apps to two different platforms, you'll likely need multiple node processes).

It's possible we could avoid this by making Wing-generated JavaScript exports parameterized -- however this may negatively affect the Wing TypeScript experience by introducing more steps for importing Wing libraries:

```
import { main, cloud } from "@wingcloud/framework";
import initPostgres from "@winglibs/postgres";

main((root) => {
  const postgres = initPostgres(root);
  new cloud.Bucket(root, "Bucket");
  new postgres.Database(root, "Database");
});
```

But such an alternative could cause problems if we want to eventually [make Wing libraries into JSII libraries](https://github.com/winglang/wing/issues/6655).

BREAKING CHANGE: Several platform-related APIs have been changed. `new()`, `newAbstract()`, `typeForFqn()`, and `tryNew()` methods have been removed from `core.App`. Instead, you can obtain a `ClassFactory` through `ClassFactory.of(scope)`, which has methods like `new()`, `tryNewInstance()` and `resolveType()`. `AppProps` now does not expect `newInstanceOverrides` anymore, and instead expects a `classFactory`.

## Checklist

- [ ] Tests added (always)


libs/@wingcloud/framework/src/main.ts
```diff
@@ -89,13 +89,15 @@ This is a Wing app and must be run through the Wing CLI (npm install -f winglang
   const rootId = process.env.WING_ROOT_ID;
 
   const $PlatformManager = new platform.PlatformManager({ platformPaths });
+  const $PolyconFactory = $PlatformManager.createPolyconFactory();
```
eladb, line 92: Wondering, do we really need another global? Why not just piggy back on `$PlatformManager`?

Chriscbr: I don't want to expose the entire surface of $PlatformManager to all preflight files, only the dependency injector (i.e. $PolyconFactory). If there's a need to give access to other platform things then we can always revisit or change it later

eladb:
OK... This is related to a discussion I had with @hasanaburayyan last month after we had our workshop where we created a custom platform provider and after a few iterations eventually just realized that the cleanest approach is to eventually just write a custom `App` class.

Looking at this PR gives me some clarity on what I think we should do...

I propose that we go back to the idea that the factory is the responsibility of **the app** and not the platform provider, and then we should just propagate the root app object to all modules (basically `$app` instead of the `$PolyconFactory`). This makes sense because the app object is kind of like the *global* scope in Wing apps, so having static access to it from everywhere is natural.

Then `$app.typeForFqn()` is available everywhere.

After this PR is merged we should continue to clean up the platform by decoupling the idea platform provider (which will now just be a class that implements `IApp`) and platform extensions (which can hook into the various stages and do stuff). But this is out of scope for this PR.

libs/@wingcloud/framework/src/main.ts
```diff
@@ -89,13 +89,15 @@ This is a Wing app and must be run through the Wing CLI (npm install -f winglang
   const rootId = process.env.WING_ROOT_ID;
 
   const $PlatformManager = new platform.PlatformManager({ platformPaths });
+  globalThis.$PolyconFactory = $PlatformManager.createPolyconFactory();
```
eladb, line 92: Perhaps defensively check that there isn't a factory already?

libs/awscdk/src/app.ts
```diff
@@ -167,47 +129,6 @@ export class App extends core.App {
     return this.synthedOutput;
   }
 
-  protected typeForFqn(fqn: string): any {
```
eladb, line 170: This is going to break users who leveraged this method to create a custom app

@hasanaburayyan ☝️

libs/awscdk/src/platform.ts
```diff
@@ -8,7 +36,62 @@ export class Platform implements platform.IPlatform {
   /** Platform model */
   public readonly target = "awscdk";
 
-  public newApp?(appProps: any): any {
+  public newApp(appProps: any): any {
     return new App(appProps);
   }
+
+  public newInstance(
```
eladb, line 43: I think we should deprecate this API because these two APIs are competing with each other.

When will `newInstance` get called? When will `typeForFqn` get called?

If I only override the former, then base classes will not be injected.

I think we should just keep the latter.

libs/wingc/src/jsify/snapshots/access_methods_and_properties_on_collections.snap
```diff
@@ -48,6 +48,8 @@ const $wing_is_test = process.env.WING_IS_TEST === "true";
 const std = $stdlib.std;
 const $helpers = $stdlib.helpers;
 const $extern = $helpers.createExternRequire(__dirname);
+const $PlatformManager = new $stdlib.platform.PlatformManager({platformPaths: $platforms});
+globalThis.$PolyconFactory = $PlatformManager.createPolyconFactory();
```
eladb, line 52: I still think we have too many layers and concepts here. Let's take this opportunity to remove redundant abstractions...

I would combine `PlatformManager` and `PolyconFactory` into a single concept.

Then:

```suggestion
globalThis.$platform = $stdlib.platform.load({ platformPaths: $platforms });
```

Then:

```js
globalThis.$platform.resolveType(fqn)
```


This will simplify all the hula hoops you need to do with .bind() and factories and managers and all that jazz.

What do you think?
E

libs/awscdk/src/platform.ts
```diff
@@ -8,7 +36,62 @@ export class Platform implements platform.IPlatform {
...
+  ): any {
+    const Type = this.typeForFqn(type);
+    if (Type) {
+      return new Type(scope, id, ...args);
+    }
+
+    return undefined;
+  }
+
+  public typeForFqn(fqn: string): any {
```
eladb, line 57:
```suggestion
  public resolveType(fqn: string): any {
```

Let's rename this one.

eladb (changes requested):
Let's take the opportunity and clean things up here.

libs/awscdk/src/app.ts
hasanaburayyan: Yea the other break would be in the `AppProps` interface where the factory is required.

libs/awscdk/src/platform.ts
hasanaburayyan: We discussed this a bit and both agreed that its probably safe to deprecate `newInstance()` but maybe do this in a followup PR. Just because there already a lot here, and this would require a decent amount of documentation updating as well.

libs/wingc/src/jsify/snapshots/access_methods_and_properties_on_collections.snap
hasanaburayyan: I’m unsure about collapsing the factory and platform manager. Maybe it’s the right call atm, but there’s still missing pieces of the platform system to implement, like platform compatibility which is an issue we also saw when different platforms have expectations on the type of App or provisioning engine

Also there is the discussion we have been having around defining differences in platforms and platform extensions. Maybe a manager isn’t that helpful with this (haven’t put much thought) but it might be better to leave for now

libs/awscdk/src/platform.ts
Chriscbr: Done

eladb:
@Chriscbr @hasanaburayyan how can I help get this through the finish line?

libs/wingc/src/jsify/snapshots/access_methods_and_properties_on_collections.snap
Chriscbr: Some of these directions are approached in https://github.com/winglang/wing/pull/6759 -- for now I'm going to try and scope the changes to just those that are needed for #6490

libs/awscdk/src/platform.ts
Chriscbr: https://github.com/winglang/wing/issues/6871

libs/compatibility-spy/test/spy.test.ts
```diff
@@ -14,6 +14,7 @@ describe("compatibility spy", async () => {
   const manager = new platform.PlatformManager({
     platformPaths: ["sim", join(__dirname, "../lib")],
   });
+  const factory = manager.createPolyconFactory();
```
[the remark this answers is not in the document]
Chriscbr: Ok - I'll rename

libs/@wingcloud/framework/src/main.ts
Chriscbr: Added

libs/awscdk/src/app.ts
Chriscbr: I've added notes about breaking changes to the PR description.

docs/contributing/999-rfcs/2022-06-14-polycons.md
```diff
@@ -238,7 +238,7 @@ Once the polycon is fully declared, we can define a concrete implementation.
 This is just a `Construct` that implements the desired Polycon's interface and takes the same properties in the constructor. 
 Many of these can exist and each can be composed of any type of construct (from any CDK!). 
 
-The concrete implementation will be chosen automatically through a `PolyconFactory` (see below) when instantiating the polycon.
```
eladb, line 241: We usually don't rewrite history. No need to update old RFCs, especially given it's completely out of date for many other reasons.

libs/@wingcloud/framework/src/main.ts
```diff
@@ -89,13 +89,15 @@ This is a Wing app and must be run through the Wing CLI (npm install -f winglang
   const rootId = process.env.WING_ROOT_ID;
 
   const $PlatformManager = new platform.PlatformManager({ platformPaths });
+  globalThis.$ClassFactory = $PlatformManager.createClassFactory();
```
eladb, line 92: Perhaps move this into the createApp() implementation so at least it's reused across all entry sites.

eladb, line 92: Add a defensive check that `$ClassFactory` is null

eladb (approved):
Hell yeah!

docs/contributing/999-rfcs/2022-06-14-polycons.md
Chriscbr: Oops, looks like this grep went a bit too far - thanks for catching it

libs/@wingcloud/framework/src/main.ts
Chriscbr: Ok, I've tried refactoring this a bit

libs/wingc/src/jsify.rs
```diff
@@ -573,46 +574,21 @@ impl<'a> JSifier<'a> {
...
-							"))(",
-							node_scope,
-							")"
-						)
-					} else {
-						new_code!(expr_span, "this.node.root.new(\"", fqn, "\", ", ctor, ", ", args, ")")
-					}
+					new_code!(
+						expr_span,
+						"globalThis.$ClassFactory.new(\"",
```
eladb, line 584: Is it possible to hide this global call behind a static method in the platform manager?

```suggestion
						"$stdlib.platform.PlatformManager.new(\"",
```

PR #6107, opened 2024-03-31
rfc: wing secrets through cli

hasanaburayyan:
RFC for creating application secrets via the Wing CLI

Demo of POC
https://www.loom.com/share/967b0047ae834442a32226cc3794fb84

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
+++ b/docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
@@ -29,7 +29,7 @@ Since secrets creation is platform specific, platforms can now implement a new h
 For example the `sim` platform implementation which needs to store secrets in a `.env` file, would look something like this:
 
 ```js
-public async createSecrets(secrets: { [key: string]: string }): Promise<string> {
+public async configureSecrets(secrets: { [key: string]: string }): Promise<string> {
   let existingSecretsContent = "";
   try {
     existingSecretsContent = fs.readFileSync('./.env', 'utf8');
--- a/docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
+++ b/docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
@@ -1,0 +1,101 @@
+---
+title: "#6105 Wing Secrets CLI"
+description: Creating Wing secrets from the CLI
+---
+
+# Wing Secrets CLI
+- **Author(s)**: @hasanaburayyan
+- **Submission Date**: 2024-03-31
+- **Stage**: Draft
+
+Creating secrets through the Wing CLI.
+
+## Background
+
+Wing applications often require secrets to be retrieved during runtime. These secrets are stored in platform specific secret stores, such as AWS Secrets Manager for `tf-aws` or a local `.env` file for the `sim` platform.
+
+Secrets must be configured before the application is run, and now the Wing CLI along with Wing platforms make it easy to configure secrets.
+
+### Out of Scope
+
+In this RFC a few things are out of scope:
+- Checking if the secrets exist in the platform's secret store when running `wing compile`
+- Reading secret values, for now we will only focus on creating secrets
+
+## Platform Hook
+
+Since secrets creation is platform specific, platforms can now implement a new hook `configureSecrets(secrets: { [key: string]: string }): string` which will be called by the Wing CLI to configure the secrets.
+
+For example the `sim` platform implementation which needs to store secrets in a `.env` file, would look something like this:
+
+```js
+public async createSecrets(secrets: { [key: string]: string }): Promise<string> {
+  let existingSecretsContent = "";
+  try {
+    existingSecretsContent = fs.readFileSync('./.env', 'utf8');
+  } catch (error) {}
+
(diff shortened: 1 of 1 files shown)


docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
```diff
@@ -0,0 +1,54 @@
...
+
+Creating secrets through the Wing CLI.
+
+## Background
+
+TODO
+
+## Sketch
+
+Running the blow command will result in an interactive prompt where you can input the Slack credentials:
```
Chriscbr, line 19:
```suggestion
Running the below command will result in an interactive prompt where you can input the Slack credentials:
```

docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
```diff
@@ -0,0 +1,54 @@
...
+
+```bash
+wing secrets main.w --create
```
Chriscbr, line 22: What would happen if you compiled `main.w` without running `wing secrets main.w --create` first? Are there errors raised because the secrets are missing? (do the errors tell you the names of the missing secrets?)

eladb, line 22:
```suggestion
wing secrets main.w
```

Why do we need the `--create` switch here?

eladb: Kind of feels like it would make sense to ask for these secrets interactively when running in the simulator without secrets in the `.env` file, no?

hasanaburayyan: My initial thought was that the default behavior of `wing secrets main.w` would list out required secrets. But default create is probably also fine.

eladb: As a design principle I usually prefer to optimize for the common use case, which is to set secret values I believe.

hasanaburayyan: > What would happen if you compiled main.w without running wing secrets main.w --create first? 

The same thing as today, the code compiles fine, and a runtime error occurs. In my opinion that might be the better DX, since I imagine if Im just writing some code that expects a secret, I might not want an error or warning on each compile as it would get noisey.

hasanaburayyan: > Kind of feels like it would make sense to ask for these secrets interactively when running in the simulator without secrets in the .env file, no?

Im thinking to scope this mvp to just secret creation, and not retrieval for now. I kinda worry that these checking for secret values on compile might get noisey/annoying.

docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
```diff
@@ -0,0 +1,116 @@
...
+    existingSecrets[key] = secrets[key];
+  }
+
+  const updatedContent = Object.entries(existingSecrets)
+    .map(([key, value]) => `${key}=${value}`)
+    .join('\n');
+
+  fs.writeFileSync('./.env', updatedContent);
+
+  return "Secrets saved to .env file";
```
eladb, line 56: I think a simple `console.log` here should be fine. What's the motivation of returning the output?

docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
```diff
@@ -0,0 +1,116 @@
...
+- Reading secret values, for now we will only focus on creating secrets
+
+## Platform Hook
+
+Since secrets creation is platform specific, platforms can now implement a new hook `configureSecrets(secrets: { [key: string]: string }): string` which will be called by the Wing CLI to configure the secrets.
+
+For example the `sim` platform implementation which needs to store secrets in a `.env` file, would look something like this:
+
+```js
+public async createSecrets(secrets: { [key: string]: string }): Promise<string> {
```
eladb, line 32: I'd call this `storeSecrets` or something like this.

eladb (approved):
LGTM

docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
hasanaburayyan: I agree that a `console.log` is probably fine, my initial motivation during the POC was to keep all the logging output and formatting within the cli command. I also figured requiring a return string might serve to motivate platform authors to provide useful response messages.

Im not married to the idea of the string return, happy to re-evaluate in the implementation PR

docs/contributing/999-rfcs/2024-03-31-wing-secrets-cli.md
hasanaburayyan: how about `configureSecrets`?

eladb: I don't think "store" implies it's an update, you can store something for the first time :-)

PR #6117, opened 2024-04-02
fix(sdk): functions and services are never updated in the simulator

### LINKED ISSUE #6116: `simulator.update()` doesn't reflect changes in functions
### I tried this:

In https://github.com/winglang/wing/pull/6069, the Console starts using the new `simulator.update()` method from the Simulator.

### This happened:

Using the `simulator.update()` method lead to a bug where new changes in functions weren't available anymore.

### I expected this:

The function code should be up to date in the Console instance after a call to `simulator.update()`

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

eladb:
Fixes #6116 by allowing a simulator resource to determine it's update "plan" when new configuration arrives.

For now, both `cloud.Function` and `cloud.Service` always return "REPLACE" which means that these resources will always get updated. In a future commit we can be more granular about the update by calculating the bundle hash.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/simulator/simulator.ts
+++ b/libs/wingsdk/src/simulator/simulator.ts
@@ -916,16 +916,13 @@ export class Simulator {
 
       case UpdatePlan.REPLACE:
         return true;
+
+      case UpdatePlan.AUTO:
+        const state = (r: BaseResourceSchema) =>
+          JSON.stringify({ props: r.props, type: r.type });
+
+        return state(oldConfig) !== state(newConfig);
     }
-
-    // the resource is already in "current", if it's different from "next", it means it was updated
-    const state = (r: BaseResourceSchema) =>
-      JSON.stringify({
-        props: r.props,
-        type: r.type,
-      });
-
-    return state(oldConfig) !== state(newConfig);
   }
 }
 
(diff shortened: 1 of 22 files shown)


libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -920,6 +1007,36 @@ export interface ISimulatorResourceInstance {
...
+
+  /**
+   * Determines the update plan for applying a new configuration for this resource.
+   *
+   * If this is not implemented, the default behavior is to automatically replace the resource if
+   * the new configuration is different from the current configuration.
+   *
+   * @param newConfig The new configuration to apply
+   */
+  plan(newConfig: BaseResourceSchema): Promise<UpdatePlan>;
```
Chriscbr, line 1019: Is the resource receiving a piece of config with all tokens resolved, or is it possible there are unresolved tokens in it? Which do you think should be the contract here?

Chriscbr (approved):

libs/wingsdk/src/simulator/simulator.ts
eladb: Tokens are unresolved, which I think is reasonable. Added to the docs.

PR #6123, opened 2024-04-03
chore(sdk): refactor inflight hosts to use inline policies in sim

Chriscbr:
Per a suggestion from #6079, this PR refactors the inflight host resources in the SDK (cloud.Service and cloud.Function) so that their policies are inlined. Other resources like `cloud.Api`, `cloud.Bucket`, `cloud.Topic`, `cloud.Schedule`, and `cloud.Queue` continue to use `sim.Policy`.

In general, `sim.Policy` is more flexible in the current implementation since using it doesn't automatically introduce a dependency on the resource(s) permissions are granted to. For example, a `Topic` can have external policy specifying that it can invoke a `Function`, without necessarily introducing a dependency between the `Topic` and `Function`.

But for `Function` or `Service`, this concern isn't as relevant since all of the permissions they use are derived from the inflight function(s) passed to them. So if a `Function` needs permissions for `Bucket`, it's going to also need an environment variable providing the name of the bucket, so a dependency between the two resources will be required either way.

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/simulator/graph.ts
```diff
@@ -1,9 +1,11 @@
+import { PolicyStatement } from "./simulator";
 import { resolveTokens } from "./tokens";
 
 export interface Definition {
   path: string;
   deps?: string[];
   props?: Record<string, any>;
+  policy?: PolicyStatement[];
```
eladb, line 8: Why not just model policy as a prop?

What's the motivation for the royal treatment?

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -796,6 +798,19 @@ export class Simulator {
     // allocate a handle for the resource so others can find it
     const handle = this._handles.allocate(path, resourceObject);
 
+    // if the resource is a policy, add it to the policy registry
+    if (resourceConfig.type === POLICY_FQN) {
```
eladb, line 802: Do we still really need this resource?

eladb (approved):
Sweet

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: > Do we still really need this resource?

I mentioned how sim.Policy behaves differently in the PR description:

> In general, `sim.Policy` is more flexible in the current implementation since using it doesn't automatically introduce a dependency on the resource(s) permissions are granted to. For example, a `Topic` can have external policy specifying that it can invoke a `Function`, without necessarily introducing a dependency between the `Topic` and `Function`.

eladb: I see. It makes sense

eladb (approved):

MarkMcCulloh (approved):

libs/wingsdk/src/simulator/graph.ts
Chriscbr: Nothing special - either probably works alright. My thinking was to reserve "props" for config unique to the individual resources

Issue #552, opened 2022-11-13
continuously ensure README.md and docs examples work


ekeren:
Following @staycoolcall911 [comment](https://github.com/winglang/wing/issues/552#issuecomment-1313597530) about the need to move the examples to a different file, and the fact that @eladb has decided to move the current README.md getting started guide to the winglang.io/docs section this is blocked by https://github.com/winglang/website/issues/66

DOD:
- [ ] fail to merge PR that breaks README.md examples
- [ ] run README.md examples with both yarn and npm (see #641)
- [ ] continuously monitor README.md examples in case they somehow break and notify alert on #549 slack channel when it does

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.53](https://github.com/winglang/wing/releases/tag/v0.74.53).

Issue #4166, opened 2023-09-12
Move `https://www.winglang.io/docs` to monorepo


MarkMcCulloh:
### Feature Spec

The documentation site is built and deployed from this repo from /docs

### Use Cases

- Ensures that documentation changes are built as part of PRs. This will notably catch broken links or otherwise invalid md
- Better CTRL-F  :)
- Monorepo all the things

### Component

Contributor Experience

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

PR #6149, opened 2024-04-04
feat(docs): test doc examples

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/552

## Summary
This PR introduces doc example testing. In its current state you can include the keyword `example` in the Markdown code block identifier similar to how we use the `playground` ability.

If you create a Wing example in the docs and want to have it tested in CI when code changes are made, then all you need to do is something like `"```javascript example" ` and this will mark the code block for testing. 

You can also mark an example for failure if you expect it to not compile by using example meta data like so 
`"```javascript example{valid: false"`

I started adding this to many of our doc examples and found quite a few were broken, so a good chunk of the changes in this PR fix broken examples

- You can also run `pnpm turbo test:doc_examples` from the root of the project to only generate and run doc example tests.

## Further considerations
- How to test examples that require other examples? (I.E. library examples)
- How to test typescript examples

## Checklist

- [x] Tests added (always)


hasanaburayyan:
@MarkMcCulloh @Chriscbr Mind taking a peek at this draft? Its just an initial poc on how we can test the examples in the docs. 

I added some additional considerations in the PR description for different types of examples that need tested

MarkMcCulloh:
When should these tests run?
In this impl, it will run when hangar runs. Right now, hangar does not run if there are only documentation changes. So we would want to change it so that hangar runs then too.
But that means we could unnecessarily be running the very slow tests for doc changes that may not need it.

My proposal would be to invert how this extracts tests from the docs. The docsite dir itself should have a build process that extracts wing files into a different place (examples/tests/docs or something) and then hangar can pick up any changes from that to determine if it needs to run any tests.

(only partially related) It'd be nice if the actual docsite build happened in this repo too https://github.com/winglang/wing/issues/4166

hasanaburayyan:
> When should these tests run? In this impl, it will run when hangar runs. Right now, hangar does not run if there are only documentation changes. So we would want to change it so that hangar runs then too. But that means we could unnecessarily be running the very slow tests for doc changes that may not need it.
> 
> My proposal would be to invert how this extracts tests from the docs. The docsite dir itself should have a build process that extracts wing files into a different place (examples/tests/docs or something) and then hangar can pick up any changes from that to determine if it needs to run any tests.
> 
> (only partially related) It'd be nice if the actual docsite build happened in this repo too #4166

@MarkMcCulloh 

So thinking about it a bit more, these example tests should run more often then just when doc changes occur. I.E. If someone makes a compiler or an SDK change that breaks an example we should know about it in the PR. So the examples can be updated.

So it doesnt feel like having hangar run for doc changes is great, but also if we just generate tests and hangar picks up on changes then we are missing coverage cases.

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_12/main.w
MarkMcCulloh, line 1: Since these are all pulled from other sources, we could add a .gitattributes to `examples/tests/doc_examples` with

```
*.w linguist-generated
```

docs/src/generate_examples_files.ts
```diff
@@ -0,0 +1,94 @@
...
+          const metaDataString = metaDataMatch[1]
+            .replace(/([a-zA-Z0-9_]+)\s*:/g, '"$1":') // Enclose property names in double quotes
+            .replace(/'/g, '"'); // Replace single quotes with double quotes if any
+          metadata = JSON.parse(metaDataString);
+        }
+      } catch (e) {
+        if (process.env.DEBUG == 1) {
+          console.log(e);
+        }
+        console.error(`Unable to parse metadata for example #${index + 1} in ${filePath}`);
```
MarkMcCulloh, line 59: Since this is build script I think it's fine to just spit out the error

```suggestion
        console.error(`Unable to parse metadata for example #${index + 1} in ${filePath}`, e);
```

docs/src/generate_examples_files.ts
```diff
@@ -0,0 +1,91 @@
...
+export interface Example {
+  code: string;
+  filePath: string;
+  exampleNumber: number;
+  metadata: ExampleMetadata;
+}
+
+const docExamplesDir = join(testDir, 'doc_examples');
+
+export function searchDirectoryForWingExamples(directory: string): Example[] {
```
Chriscbr, line 26:
```suggestion
export function findExamples(directory: string): Example[] {
```

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_12/main.w
Chriscbr, line 1: Is there any way to avoid committing these files? Are we ensuring the files are cleaned up when examples are removed?

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_15/main.w
Chriscbr, line 1: How come the testing process is split into two parts (generating files, and then testing them). Would it be possible to improve the contributing experience by writing a `describe()` test that both scans the `docs/` folder for Wing examples and then tests all of them?

tools/hangar/src/test_examples.ts
```diff
@@ -0,0 +1,39 @@
+import * as fs from "fs";
+import { join } from "path";
+
+export interface Example {
+  code: string;
+  filePath: string;
+  exampleNumber: number;
+}
+
+export function searchDirectoryForWingExamples(directory: string): Example[] {
```
Chriscbr, line 10: Duplicate code?

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_15/main.w
MarkMcCulloh: The benefit of splitting them up is that it avoids requiring `hangar` to run for arbitrary doc updates that don't update any tested examples

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_12/main.w
hasanaburayyan: Not sure if we can get around committing the files since we want to make sure that if a compiler or sdk change is made that these examples still get run.

As for the cleanup, yes the generate step blows away the existing examples before re-creating them all.

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_15/main.w
Chriscbr: Mmmm I see. Can we add some command or script to run everything together for contributors?

tools/hangar/src/test_examples.ts
hasanaburayyan: Ah probably is, I moved things around from my original PR probably duplicated something with copy-pasta. Ill take a peek.

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_15/main.w
hasanaburayyan: As of right now, all of this will happen with `pnpm build` (generation and testing) but I can add a command for those who only want to test doc example changes.

docs/turbo.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "compile": {
+      "outputs": [
+        "../examples/tests/docs_examples/**/*.w"
```
MarkMcCulloh, line 7:
```suggestion
        "../examples/tests/doc_examples/**/*.w"
```

examples/tests/doc_examples/invalid/01-preflight-and-inflight.md_example_15/main.w
hasanaburayyan: @Chriscbr @MarkMcCulloh Im no good with names, but you can now run `pnpm turbo test:doc_examples` to just generate the example tests from the examples and run them in hangar.

Im now a fan of Turbo 😁

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -28,7 +28,7 @@ Your preflight code runs once, at compile time, and defines your application's i
 
 For example, this code snippet defines a storage bucket using a class from the standard library:
 
-```js playground
+```js playground example
```
hasanaburayyan, line 31: Honestly Im wondering why we would not have all our examples openable in the `playground` and if that assumption is right, then maybe rather than add `example` I can key testing off `playground`

any thoughts here @Chriscbr @MarkMcCulloh

hasanaburayyan: Though.. maybe thats not the best approach atm, since this `example` keyword should eventually support library examples and typescript examples..

I may have talked myself out of this

Chriscbr: This looks alright to me 👍

MarkMcCulloh (approved):

PR #6152, opened 2024-04-04
fix: missing uncaught inflight exceptions when debugging

MarkMcCulloh:
During normal usage, when an unhandled exception occurs inflight we will prepare a stack trace for the user to show them where things went wrong. When debugging, the `throw` itself tells you it happened first. Since we were technically catching all inflight errors, the useful concept of "break on uncaught exception" didn't work.

With this change, the original errors remain uncaught and we can still communicate that error with the parent process.


Chriscbr (approved):

eladb (commented):
Sweet

PR #6157, opened 2024-04-05
feat: typescript and esm in preflight

### LINKED ISSUE #3013: Using ESM in external preflight function fails

MarkMcCulloh:
Fixes #3013

~TODO~
- [x] Check benchmarks to make sure happy path perf isn't hurt
- [x] Windows
- [x] ~Add e2e test to verify fix for `"type": "module"`~ Doesn't fix that issue anymore
- [x] Docs to officially sanction using ts in both preflight and inflight via extern


examples/tests/valid/extern_implementation.test.w
```diff
@@ -1,12 +1,12 @@
...
-  extern "./external_js.js" static inflight getUuid(): str;
-  extern "./external_js.js" static inflight getData(): str;
-  extern "./external_js.js" pub static inflight print(msg: str): void;
-  extern "./external_js.js" pub static preflightBucket(bucket: cloud.Bucket, id: str): Json;
+  extern "./external_ts.ts" pub static getGreeting(name: str): str;
+  extern "./external_ts.ts" static inflight regexInflight(pattern: str, text: str): bool;
+  extern "./external_ts.ts" static inflight getUuid(): str;
+  extern "./external_ts.ts" static inflight getData(): str;
+  extern "./external_ts.ts" pub static inflight print(msg: str): void;
+  extern "./external_ts.ts" pub static preflightBucket(bucket: cloud.Bucket, id: str): void;
```
Chriscbr, line 9: Is `.js` not supported anymore? (wondering why all of these are changed)

libs/wingcompiler/preflight.shim.js
```diff
@@ -0,0 +1,58 @@
...
+  throw new Error("Missing environment variable WING_JITI_ENTRYPOINT");
+}
+const entrypoint = JSON.parse(process.env.WING_JITI_ENTRYPOINT);
+delete process.env.WING_JITI_ENTRYPOINT;
+
+const jiti = require("jiti")(__filename, {
+  sourceMaps: true,
+  interopDefault: true,
+});
+jiti(entrypoint);
```
Chriscbr, line 56: Wouldn't this mean code produced by `wingc` needs to be executed through `jiti` in order to work (i.e. we're taking `jiti` as a hard dependency)? It looks like there's a substantial performance cost from the benchmarks.

I feel like we should be aiming to reduce the number of special requirements needed in order to run JavaScript `wingc` produced by wingc. Doing so would make the generated code interoperable with more JavaScript runtimes and make it easier to swap out Node.js with Deno or Bun etc. It would also make it a lot easier to debug (have a problem deep in some Wing project? Just modify the JS and run `node preflight.js`).

In the experimental Rust-based CLI, we're able to execute `preflight.js` via Node (just by passing a few extra environment variables -- it might be possible to refactor some of these ways), but it doesn't use jiti, so supporting TypeScript like this would break it.

examples/tests/valid/extern_implementation.test.w
MarkMcCulloh: It is. I only changed one file, there are other usages of .js extern.

Honestly though if extern supports .ts in both preflight and inflight, I expect most people to default to .ts.

libs/wingcompiler/preflight.shim.js
MarkMcCulloh: > Wouldn't this mean code produced by wingc needs to be executed through jiti in order to work (i.e. we're taking jiti as a hard dependency)?

Only if needs to run execute modules that didn't already work (typescript or esm)

> It looks like there's a substantial performance cost from the benchmarks.

True, locally the overhead wasn't so bad. I'm considering other options.

> I feel like we should be aiming to reduce the number of special requirements needed in order to run JavaScript wingc produced by wingc. Doing so would make the generated code interoperable with more JavaScript runtimes and make it easier to swap out Node.js with Deno or Bun etc.

This PR shouldn't harm compat with other runtimes, jiti works in other runtimes. If anything, the purpose is to try to make all runtimes work the same. Also in general I don't see harm in injecting special logic into emitted preflight js since its lifecycle is completely handled within wing's toolchain.

> (have a problem deep in some Wing project? Just modify the JS and run node preflight.js)

Is this a design goal? If so, then introducing jiti as part of preflight.js is not unreasonable. We would also need to add the overload to module resolution that's missing in preflight.js regardless of .ts support

Chriscbr: > Also in general I don't see harm in injecting special logic into emitted preflight js since its lifecycle is completely handled within wing's toolchain.

That's a fair point. I guess my concern's that adding support for ESM through this third party dependency (versus through a more mainstream means, like just relying on Node's native ESM support) could lead to unexpected compatibility or maintenance issues in the future.

The same also goes for running TypeScript files directly -- if we're promising interop with TypeScript, running it with `jiti` feels risky if only because it's not sanctioned by TypeScript or Node (though it looks like they are thinking on it - see https://github.com/nodejs/node/issues/43818). But if the risk seems small / is ok with you then I'm cool rolling with it.

----

Outta curiosity, is it possible to inline `jiti` directly within preflight.js? Also if it's only necessary for running ESM / TypeScript code, is it possible to adapt this somehow so we only pay for the cost of it when we're requiring those respective modules?

Further question: would this mean that inflight code (using imported ESM JavaScript libraries) also need to use `jiti`? Or else would there be some differences between how preflight and inflight code execute?

MarkMcCulloh: If we want ESM and typescript support to work in a predictable way, I think we'll always need something like jiti (even if we implement it ourselves).
For example, node has no builtin way of synchronously importing an ESM module. This is something bun supports that is useful for us.

> Outta curiosity, is it possible to inline jiti directly within preflight.js? Also if it's only necessary for running ESM / TypeScript code, is it possible to adapt this somehow so we only pay for the cost of it when we're requiring those respective modules or extern functions?

This will have to be the approach I think. I'm probably going to put this code into wingsdk to basically make this the compiler-internal way of loading externs.

> Further question: would this mean that inflight code (using imported ESM JavaScript libraries) also need to use jiti? Or else would there be some differences between how preflight and inflight code execute?

Inflight code would not need jiti because it has already been going through the same process (esbuild is transforming it), so if anything this better aligns how preflight and inflight execute. Now the primary difference is that inflight is bundled while preflight isn't. Jiti also uses babel by default, although I was going to try using esbuild instead.

eladb (approved):

examples/tests/invalid/extern_static.test.w
Chriscbr, line 1: Can we add a test checking that extern TypeScript methods work when they're defined in a non-entrypoint files?

Chriscbr (approved):

PR #6159, opened 2024-04-06
feat(sdk)!: add `regex` type to `std` library

### LINKED ISSUE #3644: Add Regex type to `String.replace()`
### Feature Spec

Currently `String.replace()` only supports `String` type for the `searchString` parameter, but also `Regex` type should be added once #1894 gets completed.

### Use Cases

Allow `String.replace()` method to use regular expressions.

### Implementation Notes

_No response_

### Component

SDK

### Community Notes

garysassano:
Closes #3644

## Checklist

- [x] Tests added (always)


examples/tests/sdk_tests/regex/regexp.test.w
```diff
@@ -0,0 +1,31 @@
+bring regex;
+bring expect;
+
+let r: regex.Regexp = regex.compile("p([a-z]+)ch");
```
Chriscbr, line 4: Can we make this a built-in type like the built-in `datetime` and `duration` types? I feel like regular expressions are probably handy enough to make it included without needing a `bring`.

Internally the TypeScript class can be named `std.RegExp` but some logic can be added to the compiler so that users only have to type `regexp`. If you grep for "datetime" in this file you'll see a few places where there's code for mapping types like "std.Datetime" into "datetime". We'll want to do the same thing so that Wing will translate code like this:

```js
let r: regexp = regexp.compile("p([a-z]+)ch");
```

into this, under the hood:
```js
let r: std.RegExp = std.RegExp.compile("p([a-z]+)ch");
```

We can then get rid of the `regex` module and things will be a tiny bit cleaner 🧼

Chriscbr (commented):
Looks great so far! I added a few comments

examples/tests/sdk_tests/regex/regexp.test.w
garysassano: I'm getting this after applying the changes:

```sh
error: Unknown symbol "regex"
```

libs/wingsdk/src/std/regex.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * Checks whether a given string matches a specified regex pattern.
+   * @param pattern The regex pattern to match against.
+   * @param text The input string to check for a match.
+   * @returns `true` if the pattern matches the input string, otherwise `false`.
+   */
+  public static match(pattern: string, text: string): boolean {
+    return new RegExp(pattern).test(text);
+  }
```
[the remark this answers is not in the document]
garysassano: I also thought it was kind of redundant, so let's just remove it.

garysassano:
@Chriscbr Wait, I'll push a fix.

PR #6163, opened 2024-04-07
Eladb/cloud snapshots

eladb:
When running `wing test -t TARGET` for a cloud target (e.g. `tf-aws`, etc), the framework will capture a snapshot of the synthesized output and store it next to the test file (only if the tests passed on the cloud).

Then, when running the same command within a CI system (`CI=1`), the framework will not actually deploy, but will only validate that the output hasn't changed.

A few CLI switches were added to control this behavior:

* `--snapshots=auto`   - auto-detect based on CI flag
* `--snapshots=never`  - disables snapshots
* `--snapshots=deploy` - forces a deployment even if `CI=1`.
* `--snapshots=update` - will skip deployment and only update the snapshots.
* `--snapshots=assert` - only asserts that the snapshots have not changed

Updated the `tf-aws-test.yml` workflow to use `--snapshots=never` since we actually
want to run these tests on the cloud.

## Checklist

- [ ] Tests added (always)


apps/wing/src/cli.ts
```diff
@@ -3,6 +3,7 @@ import { satisfies } from "compare-versions";
 
 import { optionallyDisplayDisclaimer } from "./analytics/disclaimer";
 import { exportAnalytics } from "./analytics/export";
+import { SNAPSHOTS_HELP } from "./commands/test/snapshots";
```
MarkMcCulloh, line 6: Command-specific imports are typically done dynamically, so putting this at the top here adds a bunch of imports that may never be needed depending on the command ran.

I get the sentiment of wanting to put the help text next to the command but as a quick fix for the benchmark I would say either to inline it in this file or put it as the lone export in another file.

eladb: OK, done.

apps/wing/src/commands/test/snapshots-help.ts
```diff
@@ -0,0 +1,17 @@
...
+Snapshots (s, --snapshots <mode>):
+  never  \t Snapshot are never captured
+  auto   \t Determines behavior based on the "CI" environment variable: "assert" if CI=1 or "deploy" otherwise
+  deploy  \t Execute tests on the target platform and update snapshots if all tests pass
+  assert \t verifies that the snapshot is up-to-date and fails the test is they are not
+  update \t Only update the snapshots without actually executing the tests on the target platform
+
+  When testing against a cloud target (e.g. -t tf-aws), if all tests pass, the compiler output 
+  will be captured under "<entrypoint>.snap.md".
+`;
```
MarkMcCulloh, line 11: I don't think we need the initial newline. Also, it looks like we can get rid of an extra space before one of the tabs. I didn't test locally though so I can't say for sure.

```suggestion
export const SNAPSHOTS_HELP = `\
Snapshots (s, --snapshots <mode>):
  never  \t Snapshot are never captured
  auto   \t Determines behavior based on the "CI" environment variable: "assert" if CI=1 or "deploy" otherwise
  deploy \t Execute tests on the target platform and update snapshots if all tests pass
  assert \t verifies that the snapshot is up-to-date and fails the test is they are not
  update \t Only update the snapshots without actually executing the tests on the target platform

  When testing against a cloud target (e.g. -t tf-aws), if all tests pass, the compiler output 
  will be captured under "<entrypoint>.snap.md".
`;
```

apps/wing/src/commands/test/snapshots-help.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const SNAPSHOTS_ERROR_HELP = [
+  "To update, run in a non-CI environment with cloud credentials or with '--snapshots=update'",
+  "To disable this behavior run with '--snapshots=never'",
+  "See https://www.winglang.io/docs/tools/cli#cloud-test-snapshots",
+].join("\n");
```
MarkMcCulloh, line 17: The description of `--snapshots=update` implies to me that the behavior should be the same regardless of being in CI or not.

```suggestion
export const SNAPSHOTS_ERROR_HELP = [
  "Update with '--snapshots=update'",
  "To disable this behavior run with '--snapshots=never'",
  "See https://www.winglang.io/docs/tools/cli#cloud-test-snapshots",
].join("\n");
```

apps/wing/src/commands/test/snapshots.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Creates a markdown snapshot of the synthesis output.
+ *
+ * @param baseName The base name of the test
+ * @param synthDir The directory containing the synthesis output
+ * @returns The snapshot content in markdown format
+ */
+async function createMarkdownSnapshot(baseName: string, synthDir: string) {
+  const fileMap: Record<string, string> = {};
+  const exclude = ["connections.json", "tree.json", "**/*.zip", "**/*.map"];
+  for await (const subpath of glob.iterate("**/*", {
```
MarkMcCulloh, line 138: Is there a more specific glob we can make here? Maybe only js and json files?

apps/wing/src/commands/test/snapshots.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    // take a snapshot of the synthesis output
+    const snapshot = await createMarkdownSnapshot(basename(snapshotFile), snapshotDir);
+
+    // update the snapshots if needed
+    switch (snapshotMode) {
+      case SnapshotMode.DEPLOY:
+      case SnapshotMode.UPDATE:
+        writeFileSync(snapshotFile, snapshot);
+        break;
```
MarkMcCulloh, line 85: Can we check the contents of the snapshot to see if it actually changed and avoid writing to it if not.

And if we do write to it, it'd be useful to let the user know as well.

apps/wing/src/commands/test/snapshots.ts
```diff
@@ -0,0 +1,178 @@
...
+  if (snapshotMode === SnapshotMode.NEVER) {
+    return;
+  }
+
+  const snapshotFile = `${entrypoint}.${target}.snap.md`;
+
+  await withSpinner(`Snapshot ${renderTestName(snapshotFile)}...`, async () => {
+    // we need to compile again in order for snapshots because we can't afford the snapshot to be
+    // based on a random root id (which is the default for tests)
+    const snapshotDir = await compile(entrypoint, options);
```
MarkMcCulloh, line 75: Should the `testing` compiler option also be set to true like it is for the test compilation.

apps/wing/src/commands/test/snapshots.ts
```diff
@@ -0,0 +1,178 @@
...
+  // skip if snapshots are disabled
+  if (snapshotMode === SnapshotMode.NEVER) {
+    return;
+  }
+
+  const snapshotFile = `${entrypoint}.${target}.snap.md`;
+
+  await withSpinner(`Snapshot ${renderTestName(snapshotFile)}...`, async () => {
+    // we need to compile again in order for snapshots because we can't afford the snapshot to be
+    // based on a random root id (which is the default for tests)
```
MarkMcCulloh, line 74: Does avoiding the random root id break anything? If not, why are we doing it in the first place?

apps/wing/src/commands/test/snapshots.ts
eladb: No, because this will result in non deterministic output (test apps have a random root for isolation).

apps/wing/src/commands/test/snapshots.ts
eladb: We need it for isolating test scopes.

apps/wing/src/commands/test/snapshots.ts
eladb: Ok

apps/wing/src/commands/test/snapshots.ts
eladb: Yes we can start with this

apps/wing/src/commands/test/snapshots.ts
MarkMcCulloh: [Here](https://github.com/winglang/wing/blob/main/apps/wing/src/commands/test/test.ts#L149-L153) both the random root id is set and a testing prop is true.
I understand avoiding the random root id for this, but the other prop seems important as well.

eladb: I am not sure I understand. Why does it make sense to set `testing` to `true` in this context? We are not really running the test, just synthesizing the output. I rather the output be as true to the production output as possible.

MarkMcCulloh: > We are not really running the test, just synthesizing the output.

Setting testing to true sets the WING_IS_TEST env to true, which sets isTestEnvironment in the app to true.

This flag is used in several places before running the tests, including being exposed in actual wing code. So it changes the resulting compilation.

> I rather the output be as true to the production output as possible.

The file being compiled could be a .test.w file, which means it probably doesn't have "production" output. `test {}` blocks are not included in production output. I figured the snapshot is reflecting the tests infrastructure rather than what'd you'd get from running "wing compile -t tf-aws entrypoint". The production infra isn't useful here because that's not what the test is using.

apps/wing/src/commands/test/snapshots.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  async function makeEntrypoint() {
+    const workdir = await generateTmpDir();
+    const entrypoint = join(workdir, "main.w");
+
+    await writeFile(
+      entrypoint,
+      `
+      bring cloud;
+      new cloud.Bucket();
+    `
```
MarkMcCulloh, line 131: It would be useful to see a `test {}` block in this example to get a more realistic file that someone may run `wing test` on.

MarkMcCulloh (approved):

Issue #6147, opened 2024-04-04
Very confusing error message when platform cannot be found


davidrupp:
### I tried this:

Working through https://www.winglang.io/docs/start-here/aws , substituting `awscdk` as the target (was `tf-aws`):

```
hello-wing % wing new empty
hello-wing % wing compile --platform awscdk main.w 
```

### This happened:

```
An error occurred while loading the custom platform: Error: ENOENT: no such file or directory, open '/Users/drupp/Developer/hello-wing/awscdk/index.js'
    at readFileSync (node:fs:455:20)
    at _loadCustomPlatform (/opt/homebrew/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/platform/platform-manager.js:203:52)
    at PlatformManager.loadPlatformPath (/opt/homebrew/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/platform/platform-manager.js:79:15)
    at /opt/homebrew/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/platform/platform-manager.js:97:18
    at Array.forEach (<anonymous>)
    at PlatformManager.createPlatformInstances (/opt/homebrew/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/platform/platform-manager.js:96:28)
    at PlatformManager.createApp (/opt/homebrew/lib/node_modules/winglang/node_modules/@winglang/sdk/lib/platform/platform-manager.js:103:14)
    at Object.<anonymous> (/Users/drupp/Developer/hello-wing/target/main.awscdk/.wing/preflight.js:86:31)
    at Module._compile (node:internal/modules/cjs/loader:1368:14)
    at Object..js (node:internal/modules/cjs/loader:1426:10) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/drupp/Developer/hello-wing/awscdk/index.js'
}
Error: Cannot read properties of undefined (reading 'newApp')
```

### I expected this:

AWS CDK target compiles correctly

### Wing Version

0.67.9

### Node.js Version

v21.7.2

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

davidrupp:
Just found https://www.winglang.io/docs/platforms/awscdk , which states:

> You will need to install the @winglang/platform-awscdk library in order to use this platform.
> $ npm i @winglang/platform-awscdk

[...]

PR #6176, opened 2024-04-08
fix(platforms): confusing error message when unable to load platform

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/6147

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/platform/platform-manager.ts
+++ b/libs/wingsdk/src/platform/platform-manager.ts
@@ -232,6 +232,9 @@ export function _loadCustomPlatform(customPlatformPath: string): any {
     script.runInContext(context);
     return new (platformModule.exports as any).Platform();
   } catch (error) {
+    if (process.env.DEBUG) {
+      console.error(error);
+    }
     const hint = customPlatformPath.includes(".")
       ? "Ensure the path to the platform is correct"
       : `Ensure you have installed the platform provider by running 'npm install ${customPlatformPath}'`;
--- a/libs/wingsdk/src/platform/platform-manager.ts
+++ b/libs/wingsdk/src/platform/platform-manager.ts
@@ -232,9 +232,9 @@ export function _loadCustomPlatform(customPlatformPath: string): any {
     script.runInContext(context);
     return new (platformModule.exports as any).Platform();
   } catch (error) {
-    const hint = customPlatformPath.includes(".") ?
-     "Ensure the path to the platform is correct" :
-     `Ensure you have installed the platform provider by running 'npm install ${customPlatformPath}'`
+    const hint = customPlatformPath.includes(".")
+      ? "Ensure the path to the platform is correct"
+      : `Ensure you have installed the platform provider by running 'npm install ${customPlatformPath}'`;
     console.error(
       `An error occurred while loading the custom platform: ${customPlatformPath}\n\n(hint: ${hint})`
     );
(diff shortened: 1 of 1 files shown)


MarkMcCulloh (approved):

MarkMcCulloh (changes requested):

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -232,9 +232,11 @@ export function _loadCustomPlatform(customPlatformPath: string): any {
     script.runInContext(context);
     return new (platformModule.exports as any).Platform();
   } catch (error) {
+    const hint = isScoped
+      ? `Ensure you have installed the platform package by running 'npm install ${customPlatformPath}'`
+      : "Ensure the path to the platform is correct";
     console.error(
-      "An error occurred while loading the custom platform:",
-      error
+      `An error occurred while loading the custom platform: ${customPlatformPath}\n\n(hint: ${hint})`
```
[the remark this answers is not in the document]
hasanaburayyan: I’m not sure how we could make using a non-existent platform clearer. In the original you get a big gnarly stack trace. I assume with these changes it might be clearer that they are using the wrong platform name, since it’s less verbose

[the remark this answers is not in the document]
MarkMcCulloh: Maybe instead of `isScoped`, check if the platform contains a `.`.
If it does, assume it's a path can give the path error. Otherwise give the error that the package should be installed

eladb, line 239: Please add instructions or doclink and suggest that the platform provider/extension might not be installed...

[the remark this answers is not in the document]
hasanaburayyan: Ah! Now I see the point you were making. 👍 Changed the check

hasanaburayyan: @eladb Just to be clear, when you suggest adding instructions are you referring to how to install platform extension? Currently in this PR if a platform is not installed, lets say `@winglang/platform-aws-cdk` the error message looks like this:
```bash
mywing compile -t @winglang/platform-awscdk
An error occurred while loading the custom platform: @winglang/platform-awscdk

(hint: Ensure you have installed the platform provider by running 'npm install @winglang/platform-awscdk')
```

MarkMcCulloh (approved):

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -232,9 +232,11 @@ export function _loadCustomPlatform(customPlatformPath: string): any {
     script.runInContext(context);
     return new (platformModule.exports as any).Platform();
   } catch (error) {
+    const hint = customPlatformPath.includes(".")
+      ? "Ensure the path to the platform is correct"
+      : `Ensure you have installed the platform provider by running 'npm install ${customPlatformPath}'`;
     console.error(
-      "An error occurred while loading the custom platform:",
-      error
```
eladb, line 237: Maybe we should print the error if DEBUG=1. Otherwise it will be hard to debug issues that are not a result of uninstalled provider.

Alternatively, separate the two cases (not found and other errors).

eladb (changes requested):

libs/wingsdk/src/platform/platform-manager.ts
hasanaburayyan: Good point, Ill add the log on DEBUG

eladb (approved):

eladb (approved):

eladb (approved):

eladb (approved):

PR #6197, opened 2024-04-10
fix(compiler)!: no way to detect if an interface implements a preflight class

### LINKED ISSUE #5872: No qualification error when accessing a preflight object via interface
### I tried this:

```wing
bring cloud;

bucket.addObject("key", "value");

interface ISomething {
  inflight ido(): void;
}

class Something impl ISomething {
  pub inflight ido() {
    log("bucket has {bucket.get("key")}");
  }
}

let s: ISomething = new Something();

test "test" {
  let is = s;
  is.ido(); // This should be a qualification error because we don't know who `is` is.
}
```
`wing test -t tf-aws main.w`

### This happened:

It compiled and ran, resulting in the following excepted runtime error:
```
fail ┌ main.tfaws.783602 » root/Test.g2vEv5bEZ9/env0/test:test
     │ Error: AccessDenied: Access Denied
```

### I expected this:

A compilation error:
`Expression of type "ISomething" references an unknown preflight object, can't qualify its capabilities (see https://github.com/winglang/wing/issues/76 for details)`

### Is there a workaround?

_No response_

### Anything else?

Currently we produce these errors when accessing a member of a preflight class type on an inflight expression. We need to extend this somehow to detect "preflight class interfaces" too. I'm not sure if this is trivial since interfaces don't have a phase? do they?

### Wing Version

0.59.46

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

yoav-steinberg:
Fixes #5872

Add a phase modifier `inflight` support to `interface` definitions. Now all preflight interfaces implicitly implement `std.IResouce` the same way that prefligth classes implicitly implement `std.Resource`. We can now explicitly qualify these interfaces with the `lift()` builtin and we can get qualification errors if an inflight variable is being used to access such an object (same as classes).

Note that inflight interfaces can be defined preflight with the keywork and all their methods will be implicitly `inflight`. Interfaces defined inflight are always inflight interfaces. Preflight interfaces may extend inflight interfaces, but not the other way around.

JSII imported interfaces are, at least for now, always preflight interfaces.

**Breaking change**:
Since inflight interfaces are new, code using interfaces in inflight that were defined in preflight scope will need to add the `inflight` modifier to those interfaces or the compiler will complain that you're trying to access an unknown preflight object:
```wing
inflight IMyIface { // Explicitly mark this interface as inflight so it can be used in inflight code without worrying about lifting
  do(): void; // No need to specify `do` is an inflight method, it's implicit from the interface type.
}
inflight () => {
  x: IMyIface = ....;
  x.do(); // This would be a qualification error without the above fix.
}
```

**Other fixes**
* Fixed span in diagnostic of wrongly typed methods implementing an interface, previously the span pointed to the implementing class instead of the specific method inside that class that mismatches the interface definition.
* Cleaned up some dead code.

## Checklist

- [x] Tests added (always)


libs/wingc/src/type_check.rs
```diff
@@ -1084,6 +1070,18 @@ impl TypeRef {
 		return false;
 	}
 
+	pub fn is_preflight_type(&self) -> bool {
```
Chriscbr, line 1073: `is_preflight_type` sounds like it might include preflight functions. Maybe `is_preflight_object` would describe it better?

libs/wingc/src/type_check.rs
```diff
@@ -4165,14 +4138,42 @@ impl<'a> TypeChecker<'a> {
 			};
 		}
 
-		// add methods from all extended interfaces to the interface env
 		let extend_interfaces = &interface_type.as_interface().unwrap().extends;
-		if let Err(e) = add_parent_members_to_iface_env(extend_interfaces, name, &mut interface_env) {
+
+		// If this is a preflifght interface and it doesn't extend any other preflight interfacs then implicitly make it extend
```
Chriscbr, line 4143:
```suggestion
		// If this is a preflight interface and it doesn't extend any other preflight interfacs then implicitly make it extend
```

libs/wingc/src/type_check.rs
```diff
@@ -4209,6 +4210,19 @@ impl<'a> TypeChecker<'a> {
 			})
 			.collect::<Vec<_>>();
 
+		// Verify impelmented interfaces are of valid phase for this class
```
Chriscbr, line 4213:
```suggestion
		// Verify implemented interfaces are of valid phase for this class
```

libs/wingc/src/type_check.rs
yoav-steinberg: changed to `is_preflight_object_type`

docs/docs/03-language-reference.md
```diff
@@ -1577,7 +1581,12 @@ of methods with different phases is not allowed as well.
 
 Interfaces represent a contract that a class must fulfill.
 Interfaces are defined with the `interface` keyword.
-Currently, preflight interfaces are allowed, while inflight interfaces are not supported yet (see https://github.com/winglang/wing/issues/1961).
+Interfaces may be either preflight interfaces or inflight interfaces.
+Preflight interfaces are defined in preflight scope and can contain both preflight and inflight methods.
+Only preflight classes may implement preflight interfaces.
+Inflight interfaces are either defined with the `inflight` modifier in preflight scope or simply defined in inflight scope.
+All methods of inflight interfaces are implicitly inflight (no need to use the `inflight` keyword).
+Both preflight and inflight classes may implement an inflight interface.
```
Chriscbr, line 1589: The reason a preflight class can implement an inflight interface might not be obvious
```suggestion
Since both preflight and inflight classes can have inflight methods defined inside them, they are both capable of implementing inflight interfaces.
```

examples/tests/invalid/interface.test.w
```diff
@@ -42,3 +44,26 @@ interface IMissingParamTypes {
...
+
+// Inflight interfaces can't extend JSII interfaces
+inflight interface IInflightExtendsJsii extends jsii_fixture.ISomeInterface {
+  method(): void;
+}
+
+// Inflight classes can't implement JSII interfaces
+inflight class CInflightImplJsii impl jsii_fixture.ISomeInterface {
+  pub method(): void {}
+}
```
Chriscbr, line 69: Would this also be an error?
```js
inflight interface IInflightIface {
  pub method(): void {}
}

inflight interface IInflightExtendsResource impl IInflightIface, std.IResource {
  pub method2(): void {}
}
```

Chriscbr (approved):
LGTM

examples/tests/invalid/interface.test.w
yoav-steinberg: Yes, `std.IResource` is a JSII imported interface and therefore it's a **preflight**  interface. So you get this error:
```
inflight interface IInflightExtendsResource extends IInflightIface, std.IResource {
                   ^^^^^^^^^^^^^^^^^^^^^^^^ Inflight interface IInflightExtendsResource cannot extend a preflight interface IResource
```

libs/wingc/src/type_check.rs
```diff
@@ -1605,6 +1573,17 @@ impl Types {
 			.expect("Resouce base class to be a type")
 	}
 
+	pub fn resource_base_interface(&self) -> TypeRef {
+		let resource_fqn = format!("{}.{}", WINGSDK_ASSEMBLY_NAME, WINGSDK_IRESOURCE);
+		self
+			.libraries
+			.lookup_nested_str(&resource_fqn, None)
+			.expect("Resouce base interface to be loaded")
```
staycoolcall911, line 1581:
```suggestion
			.expect("Resource base interface to be loaded")
```

libs/wingc/src/type_check.rs
```diff
@@ -1605,6 +1573,17 @@ impl Types {
...
+			.0
+			.as_type()
+			.expect("Resouce base interface to be a type")
```
staycoolcall911, line 1584:
```suggestion
			.expect("Resource base interface to be a type")
```

examples/tests/invalid/package.json
```diff
@@ -3,7 +3,8 @@
   "dependencies": {
     "cdktf": "0.20.3",
     "constructs": "^10.3",
-    "jsii-code-samples": "1.7.0"
+    "jsii-code-samples": "1.7.0",
+    "jsii-fixture": "workspace:^"
```
staycoolcall911, line 7: Just making sure you meant to add this

staycoolcall911 (commented):
Great cleanup!

examples/tests/invalid/package.json
yoav-steinberg: Yes, this relates to using the jsii fixture in the invalid tests.

libs/wingc/src/type_check.rs
yoav-steinberg: oops, I'll fix in some other PR...

libs/wingc/src/type_check.rs
yoav-steinberg: oops, I'll fix in some other PR...

PR #6205, opened 2024-04-10
feat(sdk): stream logs in `wing test`

### LINKED ISSUE #2096: Stream logs during wing unit tests
### Feature Spec

Currently, you only see the logs from a test after all tests finish. Starting today, logs can now be streamed during `wing test`. For example, suppose we have the following code:

```js
bring cloud;

resource Utils {
  init() {}
  extern "./utils.js" inflight sleep(seconds: num); // implementation omitted
}

let utils = new Utils();
  log("before");
  utils.sleep(1);
  log("after");
}) as "test";
```

Today, if you run `wing test hello.w`, after one second, "before" and "after" logs both be displayed.

Instead, "before" will be printed immediately, and "after" should be printed 1 second later.

### Use Cases

If a test takes several seconds, it's often preferable to show any logs associated with those tests as they're printed, instead of only when the test finishes.

### Implementation Notes

_No response_

### Component

SDK, CLI

### Community Notes

Chriscbr:
This PR introduces log streaming for the `wing test` command is used with the simulator. By default, any `log` statements printed by the user will be automatically streamed while the tests are running. (The first line of each log includes a string identifying which test it's associated with). More granular trace events from the individual resources can be displayed with the `--debug` flag.

The order of logs can vary and they each include timestamps, so the `--no-stream` flag can be passed to fall back to the old output style. This is used within the monorepo in places where we're snapshotting CLI output for catching regressions.

https://github.com/winglang/wing/assets/5008987/02c9e769-f6b6-4b17-8e5f-2418fcee7531

Closes #2096

## Checklist

- [x] Tests added (always)


apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,98 @@ async function runTestsWithRetry(
...
+        const severity = inferSeverityOfEvent(event);
+        const date = new Date(event.timestamp);
+        const hours = date.getHours().toString().padStart(2, "0");
+        const minutes = date.getMinutes().toString().padStart(2, "0");
+        const seconds = date.getSeconds().toString().padStart(2, "0");
+        const milliseconds = date.getMilliseconds().toString().padStart(3, "0");
+        const timestamp = `${hours}:${minutes}:${seconds}.${milliseconds}`;
+        let msg = "";
+        msg += chalk.gray(`[${timestamp}]`);
+        msg += LOG_STREAM_COLORS[severity](` [${SEVERITY_STRING[severity]}]`);
```
MarkMcCulloh, line 449: If !debug, is there more than 1 severity type? If so we can probably leave it out until we have different severities available to the user. It's definitely needed for debug where there are several kinds.

apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,98 @@ async function runTestsWithRetry(
...
+        const pathSuffix = event.sourcePath.split("/").slice(2).join("/");
+        const severity = inferSeverityOfEvent(event);
+        const date = new Date(event.timestamp);
+        const hours = date.getHours().toString().padStart(2, "0");
+        const minutes = date.getMinutes().toString().padStart(2, "0");
+        const seconds = date.getSeconds().toString().padStart(2, "0");
+        const milliseconds = date.getMilliseconds().toString().padStart(3, "0");
+        const timestamp = `${hours}:${minutes}:${seconds}.${milliseconds}`;
+        let msg = "";
+        msg += chalk.gray(`[${timestamp}]`);
```
MarkMcCulloh, line 448: Try this one to see if it's any good. I feel like the timestamp is a little too prominent.

```suggestion
        msg += chalk.dim(`[${timestamp}]`);
```

apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,98 @@ async function runTestsWithRetry(
...
+  }
+  return "verbose";
+}
+
+const SEVERITY_STRING = {
+  error: "ERROR  ",
+  warn: "WARNING",
+  info: "INFO   ",
+  debug: "DEBUG  ",
+  verbose: "VERBOSE",
```
MarkMcCulloh, line 388: The padding looks a little strange inside the brackets, how does it look at the end? i.e. `[VERBOSE]` and `[INFO]   `

apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,98 @@ async function runTestsWithRetry(
...
+        const date = new Date(event.timestamp);
+        const hours = date.getHours().toString().padStart(2, "0");
+        const minutes = date.getMinutes().toString().padStart(2, "0");
+        const seconds = date.getSeconds().toString().padStart(2, "0");
+        const milliseconds = date.getMilliseconds().toString().padStart(3, "0");
+        const timestamp = `${hours}:${minutes}:${seconds}.${milliseconds}`;
+        let msg = "";
+        msg += chalk.gray(`[${timestamp}]`);
+        msg += LOG_STREAM_COLORS[severity](` [${SEVERITY_STRING[severity]}]`);
+        msg += chalk.white(` "${testName}" » /${pathSuffix}`);
```
MarkMcCulloh, line 450: Is it obvious enough that the thing in quotes is the test name? Maybe we could do something else to emphasize it, like chalk.bold

apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,98 @@ async function runTestsWithRetry(
...
+        msg += "\n";
+        msg += chalk.whiteBright(`${event.data.message}`);
```
MarkMcCulloh, line 452: Not a deal breaker, but it'd be nice if we can avoid adding any ansi to the data itself since we could be overriding existing data.

MarkMcCulloh (commented):
Stoked for this. We should check that the output looks legible for the following cases

- long/multiline
- color disabled (e.g. NO_COLOR env set)

apps/wing/src/commands/test/test.ts
Chriscbr: Removed the `chalk.white` stuff

apps/wing/src/commands/test/test.ts
Chriscbr: Changed to bold 👍  looks pretty decent I think

apps/wing/src/commands/test/test.ts
Chriscbr: Done. dim does look a little darker on my screen

apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,102 @@ async function runTestsWithRetry(
...
   const { clean, testFilter, retry } = options;
+
+  let outputStream: SpinnerStream | undefined;
+  if (options.stream) {
+    outputStream = new SpinnerStream(process.stdout, "Running tests...");
+
+    // As of this comment, each Wing test is associated with an isolated environment.
+    // (All resources for test #0 are in root/env0/..., etc.)
+    // This means we can use the environment number to map each environment # to a test name,
+    // so when we receive a trace from the simulator, we can infer which test it's associated with.
```
Chriscbr, line 410: Note: when we want to add tests that share the same resources, it might add some trickiness because it might not be as clear which test a log or event should be associated with.

I imagine if/when we have this kind of capability, we might be able to associate a name with the larger test suite:

```
describe "simple cluster" {
  let r = new redis.Redis();
  
  test "set" { ... }
  test "get" { ... }
}
```

And then all logs can be associated with "simple cluster", or if the tests are run sequentially, we can try and annotate them based on which test is currently logging. Anyway, just food for thought.

tools/hangar/src/compatibility-spy.test.ts
```diff
@@ -8,7 +8,7 @@ compatibilityTestFiles.forEach((wingFile) => {
   test(wingFile, async ({ expect }) => {
     // "@winglang/compatibility-spy" will be installed in the tmpDir
     const platforms = ["sim", "node_modules/@winglang/compatibility-spy/lib"];
-    const args = ["test", "-o", "out.json"];
+    const args = ["test", "-o", "out.json", "--no-stream"];
```
MarkMcCulloh, line 11: Does it break if this flag isn't provided? Streaming logs should still work in a non-tty

apps/wing/src/commands/test/test.ts
Chriscbr: Ah, that's a good question. I think there are some cases. For example, if there are errors happening elsewhere in the stack, I imagine it could be good to surface those to the user by default:

```
bring cloud;
bring util;

let q = new cloud.Queue();

q.setConsumer(inflight (msg: str) => {
  log("Received message: " + msg);
  throw "chaos!";
});

test "setConsumer" {
  q.push("hello", "world");
  util.sleep(1s);
}
```

Here the test passes, but the queue consumer throws errors so they can be shown.

apps/wing/src/commands/test/test.ts
```diff
@@ -357,9 +364,102 @@ async function runTestsWithRetry(
...
+        if (testName === undefined) {
+          // This event is not associated with any test environment, so skip it.
+          return;
+        }
+        if (testFilter && !testName.includes(testFilter)) {
+          // This test does not match the filter, so skip it.
+          return;
+        }
+
+        const pathSuffix = event.sourcePath.split("/").slice(2).join("/");
```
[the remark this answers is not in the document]
Chriscbr: All of the logs/events have information associated with them in the current format (with larger apps and test names with more than one or two words it becomes pretty hard to read if it's all squeezed on one line). Not sure if that answers your question...
```
[09:50:58.162] [VERBOSE] setConsumer » /test:setConsumer/Handler
Initializing sandbox.
[09:50:58.205] [DEBUG]   my first test » /Bucket
Put (key=obj.txt).
[09:50:58.207] [INFO]    my first test » /test:setConsumer/Handler
hello
```

apps/wing/src/commands/test/test.ts
Chriscbr: Added

tools/hangar/src/compatibility-spy.test.ts
Chriscbr: There's a check so the streaming should work in a non-tty - I can try removing it from this test

Chriscbr:
Here are some screenshots of the latest revised formatting. The output is "short" when you run `wing test` without any flags:
Screenshot 2024-04-11 at 1 37 08 PM

and "full" when you run `wing test` with any debug flags:
Screenshot 2024-04-11 at 1 37 30 PM

tools/hangar/src/generated_test_targets.ts
```diff
@@ -78,7 +78,7 @@ export async function testTest(
 ) {
   const fileMap: Record<string, string> = {};
   const platforms = ["sim"];
-  const args = ["test"];
+  const args = ["test", "--no-stream"];
```
MarkMcCulloh, line 81: Is this arg needed for these tests as well?

MarkMcCulloh (approved):

tools/hangar/src/generated_test_targets.ts
Chriscbr: I was going to say yes since I think these tests include snapshots of stdout/stderr, but since the default logs don't include timestamps anymore I'm down to try snapshotting the new format as long as it's stable in CI. One sec...

MarkMcCulloh: Mostly I'm curious if we need to handle anything special if users run wing in CI. Even if we have to sanitize the output we should make sure it generally works/looks as expected in CI.

tools/hangar/__snapshots__/test_corpus/valid/resource.test.w_test_sim.md
```diff
@@ -2,12 +2,12 @@
...
-     └ counter is: 201
-pass ┌ resource.test.wsim » root/env1/test:dependency cycles
-     └ counter is: 201
- 
- 
+[INFO] test | counter is: 201
+[ERROR] dependency cycles | InvokeAsync (payload="foo") failure. Error: Process exited with code null, signal SIGTERM
+[ERROR] dependency cycles | InvokeAsync (payload="foo") failure. Error: Process exited with code null, signal SIGTERM
+pass ─ resource.test.wsim » root/env0/test:test             
+pass ─ resource.test.wsim » root/env1/test:dependency cycles
```
MarkMcCulloh, line 9: Flakey test? I think I've seen an issue with invokeasync  before

Chriscbr: I think I found a fix 🤞

PR #6217, opened 2024-04-11
fix: unable to hit most inflight breakpoints while debugging

### LINKED ISSUE #2546: Add an ability to debug with breakpoints
### Feature Spec

As a developer, I'd love to have the ability to put a breakpoint on winglang code in my IDE, to pause, play, inspect, evaluate and edit variables

### Use Cases

debugging

### Implementation Notes

_No response_

### Component

Development Environment

### Community Notes

### LINKED ISSUE #5986: Debugging inflight function with breakpoint in simulator causes timeout
### I tried this:

Set a breakpoint in a `test`, debug `wing test` and sit at the breakpoint for a minute

### This happened:

Test stops due to timeout

### I expected this:

Not sure, maybe one of these options:
- Much longer default timeout
- No timeout because we're debugging
- Easy way to quicky disable timeouts when debugging

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

### LINKED ISSUE #5988: Tests must be run multiple times to hit breakpoints when debugging `wing it`
### I tried this:

1. Create file with `test`
2. Set breakpoint inside test
3. Debug `wing it`
4. Run test in UI

### This happened:

Breakpoint was not hit. Running the test again hits the breakpoint.

### I expected this:

_No response_

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

### LINKED ISSUE #6036: Default debugger for vscode
### Use Case

VSCode can have a default debugger for a given language.
This debugger for wing should start the builtin javascript debugger attached to a new wing cli process. It must handle debugging both preflight and inflight (simulator) for all child processes.

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

MarkMcCulloh:
Fixes #5988
Fixes #2546
Fixes #6036
Fixes #5986

#5988 was a symptom of a deeper issue. In `wing it` or any other command, breakpoints would only work sometimes and depended on the order of inflights in a file and how many there were. The underlying issue is that the sourcemaps are too slow and may not finish loading in time to help the debugger attach breakpoints. At first I made a change in this PR to fix a problem with how esbuild bundles our sourcemaps. On it's own this change made many scenarios work that previously didn't.

That wasn't enough though, so I went through many other options. As part of this, I actually added debugger settings to vscode.

In the end I needed to add a hack where we simply have a small sleep before running inflight code.

TODO:
- [x] Check Windows sourcemaps aren't borked
- [x] Ensure minimal benchmark regression (might skip this process if debugger is not needed)
- [x] A few tests for the remapping logic
- [x] Test locally with a winglib
- [ ] Try to figure out better options than a sleep


libs/wingsdk/src/shared/bundling.ts
```diff
@@ -101,3 +94,100 @@ export function createBundle(
...
+/**
+ * Takes a bundled sourcemap and does the following fixes:
+ * - Normalizes paths in sources and sourceRoot
+ * - Removes duplicate sources and sourcesContent
+ * - Updates mappings to reflect the new source indices
+ *
+ * The duplicate sources come from esbuild's strange handling of multiple files being bundled that point to the same source (e.g. inflights that point to one .w file)
+ * See https://github.com/evanw/esbuild/issues/933
+ */
+function fixSourcemaps(sourcemapData: SourceMap): void {
```
MarkMcCulloh, line 114: if/when we do async esbuild, this would probably be much more efficient as a plugin

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -47,19 +47,29 @@ export class Sandbox {
       );
     }
 
+    let debugShim = "";
+    if (inspectorUrl?.()) {
+      // If we're debugging, we need to make sure the debugger has enough time to attach
+      // to the child process. This gives enough time for the debugger load sourcemaps and set breakpoints.
+      // See https://github.com/microsoft/vscode-js-debug/issues/1510
+      debugShim =
+        "\n  await new Promise((resolve) => setTimeout(resolve, 25));";
```
MarkMcCulloh, line 56: It's unfortunate that this may have to be the solution since it's arguably an upstream issue. When debugging I think user's are more accepting of some slowness so I may even increase this a little bit.

libs/wingsdk/src/shared/bundling.ts
Chriscbr, line 114: Can we write a small unit test for this?

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -47,19 +47,29 @@ export class Sandbox {
       );
     }
 
+    let debugShim = "";
+    if (inspectorUrl?.()) {
```
Chriscbr, line 51: Where is this defined? (for my own curiosity)

Chriscbr (approved):

libs/wingsdk/src/shared/sandbox.ts
MarkMcCulloh: [It comes directly from the inspector builtin module](https://nodejs.org/api/inspector.html#inspectorurl), I mapped it from `url` to be less ambigiguous.

PR #6218, opened 2024-04-12
fix(cli): `wing pack` does not enforce using `peerDependencies`

### LINKED ISSUE #5470: `wing pack` should enforce `peerDependencies` usage instead of regular dependencies
### Use Case

When wing libraries are compiled, the generated javascript is executed in the consumer's context. Because of that, transitive dependencies are suddenly expected to be actual dependencies of the consumer for the purpose of module resolution.

using `peerDependencies` more correctly models this relationship and should fix usage with pnpm and package managers that correctly enforce dependency closures.

### Proposed Solution

`wing pack` should error if `dependencies` is used in the package.json and instead suggest using `peerDependencies`

### Further clarification

If `A` depends on `B`, when you install `A` with npm your node_modules look like this:

```
node_modules
  A
  B
preflight.js
```

This means that preflight.js can import `B` even though it's not a direct dependency.

pnpm instead does something like this: (there are other misdirections involved but this is the gist)

```
node_modules
  A
    node_modules
      B
preflight.js
```

The js we emit makes direct references to `B` because it originally came from wing files in `A`, so they no longer resolve correctly, both at runtime and also during compilation because our resolver code isn't smart enough.

If you use peer dependencies, `pnpm install` will go back to creating a sufficiently flat node_modules with both A and B.

### Community Notes

garysassano:
Closes #5470

## Checklist

- [x] Tests added (always)

### DIFF
--- a/apps/wing/src/commands/pack.ts
+++ b/apps/wing/src/commands/pack.ts
@@ -123,7 +123,7 @@ export async function pack(options: PackageOptions = {}): Promise<string> {
     // check package.json has `dependencies`
     if (pkgJson.dependencies) {
       throw new Error(
-        `Wing libraries should use "peerDependencies" instead of "dependencies" in package.json.`
+        `Cannot create package with "dependencies" in package.json. Use "peerDependencies" instead.`
       );
     }
 
--- a/apps/wing/src/commands/pack.ts
+++ b/apps/wing/src/commands/pack.ts
@@ -120,8 +120,8 @@ export async function pack(options: PackageOptions = {}): Promise<string> {
       }
     }
 
-    // check package.json has `dependencies`
-    if (pkgJson.dependencies) {
+    // Check if package.json has non-empty "dependencies"
+    if (pkgJson.dependencies && Object.keys(pkgJson.dependencies).length > 0) {
       throw new Error(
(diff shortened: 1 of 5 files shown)


apps/wing/src/commands/pack.ts
```diff
@@ -120,6 +120,13 @@ export async function pack(options: PackageOptions = {}): Promise<string> {
       }
     }
 
+    // check package.json has `dependencies`
+    if (pkgJson.dependencies) {
```
MarkMcCulloh, line 124: Add a check to make sure there are values in this key (an empty dep map is weird but fine)

apps/wing/src/commands/pack.ts
```diff
@@ -120,6 +120,13 @@ export async function pack(options: PackageOptions = {}): Promise<string> {
...
+      throw new Error(
+        `Wing libraries should use "peerDependencies" instead of "dependencies" in package.json.`
+      );
```
MarkMcCulloh, line 127: We often write error messages as "problem, solution". So maybe

```suggestion
      throw new Error(
        `Cannot create package with "dependencies" in package.json. Use "peerDependencies" instead.`
      );
```

MarkMcCulloh (approved):

PR #6219, opened 2024-04-12
feat(sdk)!: change `cloud.Function` event type from `str` to `Json`

### LINKED ISSUE #6216: AWS Lambda input event is JSON and not string
### I tried this:

```js
bring cloud;

let fn = new cloud.Function(inflight (e) => {
  assert(Json.tryParse(e)?.get("foo")?.asNum() == 123);
});

test "send a json object to function" {
  fn.invoke(Json.stringify({ foo: 123 }));
}
```

This works in the simulator.

Now, deploy to AWS:

```sh
wing compile -t tf-aws
cd target/main.tfaws
terraform init
terraform apply
```

Now, let's try to call invoke this function:

```sh
aws lambda invoke --function-name Function-c852aba6 --payload $(echo '{ "foo": 123 }' | base64)  out.json
```

### This happened:

```
2024-04-11T16:29:33.931Z	ca9e1200-f286-4220-9b6e-f2ca7af93153	ERROR	Invoke Error 	
{
    "errorType": "TypeError",
    "errorMessage": "Cannot read properties of undefined (reading 'foo')",
    "stack": [
        "TypeError: Cannot read properties of undefined (reading 'foo')",
        "    at <anonymous> (/main.w:4:3)",
        "    at $Closure1.handle (/main.w:4:3)",
        "    at Runtime.exports.handler (/var/function_c852aba6.cjs:17:25)",
        "    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1173:29)"
    ]
}
```

### I expected this:

I kind of expected this to work, although I can understand why this doesn't work.

There's a discrepancy between the APIs of Lambda and `cloud.Function` where `cloud.Function` handlers are `str` and at AWS they are JSON values.

### Is there a workaround?

_No response_

### Anything else?

I think the right solution on Wing's end is to change the signature of `cloud.Function`'s handle to accept `Json?` as the input (and output).

This seems like the safest thing to do.

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

garysassano:
Closes #6216

BREAKING CHANGE: The input and return types for `cloud.Function` handlers has been changed from `str?` to `Json?` to align with the event formats used by major cloud providers like AWS, GCP, and Azure.

This is a breaking change for any existing code that assumes the input event is a `str`.

To migrate:
- Update `cloud.Function` handlers to expect a `Json?` value as input instead of a `str?`
- Remove any JSON parsing logic since the input will already be parsed into a `Json` value. To convert a `x: Json?` to `str`, you can write  `x?.asStr()!`.
- Update any tests or invocations of `cloud.Function` to pass in `Json` values instead of `str`

## Checklist

- [ ] Tests added (always)


eladb:
Please add a detailed description and a `BREAKING CHANGE` section to the PR description

garysassano:
The function invocation on AWS works now.

```sh
❯ aws lambda invoke --function-name Function-c852aba6 --payload $(echo '{ "foo": 123 }' | base64)  out.json
{
    "StatusCode": 200,
    "ExecutedVersion": "$LATEST"
}
```

garysassano:
I resolved the conflicts, but I don't have much knowledge on how the simulator resources work. Some of them require type adjusting.

tsuf239:
it seems like there are many compilation errors: 
image
I think you can reproduce it by running "pnpm run compile" at the SDK folder

Chriscbr:
Hey Gary, I wanted to give a heads up I've started working on update this PR and fix up some of the remaining tests. Stay tuned!

revitalbarletz:
@Chriscbr , @garysassano - what is the status of this one?

Chriscbr:
@revitalbarletz A few other features got bumped up in priority so I had to put my work on this on pause, but I'm aiming to update the PR and try doing some more testing for it later this week.

Chriscbr:
SDK spec tests: https://github.com/winglang/wing/actions/runs/9913437025/job/27390443979

Chriscbr:
Attempt fixing Azure tests: https://github.com/winglang/wing/actions/runs/9914506779

examples/tests/invalid/cloud_function_expects_inflight.test.w
```diff
@@ -1,7 +1,7 @@
 bring cloud;
 
-new cloud.Function((name: str): str => {
-  return "Hello {name}";
+new cloud.Function((name: Json?): Json? => {
+  return "Hello {name?.tryAsStr() ?? "world"}!";
```
eladb, line 4: Is this really needed? Why wouldn't this work?

```suggestion
  return "Hello {name ?? "world"}!";
```

eladb (approved):

examples/tests/invalid/cloud_function_expects_inflight.test.w
Chriscbr: That will compile, but `name` (if it's non-nil) is `Json`, which means it will be stringified as a JSON -- with quotes. This way JSON round-tripping behavior behaves well:

```js
let json1 = Json { a: 123 };
let str1 = "{json1}"; // {"a":123}
let parsed1 = Json.parse(str1); // OK!

let json2 = Json "hello";
let str2 = "{json2}"; // "hello" (with the quotes!)
let parsed2 = Json.parse(str2); // OK!
```

(open ended question) Do you think if `x` is Json, then `"{x}"` should do something different than `Json.stringify(x)`?

eladb: > (open ended question) Do you think if x is Json, then "{x}" should do something different than Json.stringify(x)?

Maybe... Pragmatically:

```js
let x: Json = "hello";

log("hello, {x}");
```

Would be nice if it printed:

```
hello, hello
```

skyrpex: > (open ended question) Do you think if x is Json, then "{x}" should do something different than Json.stringify(x)?

Worth a dedicated thread IMO. I don't know what's the desired behavior off the top of my head 🤔 .

libs/wingsdk/test/simulator/simulator.test.ts
```diff
@@ -818,7 +818,8 @@ describe("in-place updates", () => {
     ]);
   });
 
-  test("cloud.Api routes are updated", async () => {
+  // TODO: this test is flakey in CI
+  test.skip("cloud.Api routes are updated", async () => {
```
skyrpex, line 822: Maybe we could configure our vitest to retry some tests a few times... https://vitest.dev/config/#retry?

tools/hangar/__snapshots__/test_corpus/sdk_tests/function/aws-function.test.w_compile_tf-aws.md
```diff
@@ -184,17 +111,6 @@
...
-        "//": {
-          "metadata": {
-            "path": "root/Default/Default/FunctionAccessingContext/S3Object",
-            "uniqueId": "FunctionAccessingContext_S3Object_FCA6F30A"
-          }
-        },
-        "bucket": "${aws_s3_bucket.Code.bucket}",
-        "key": "<ASSET_KEY>",
-        "source": "<ASSET_SOURCE>"
-      },
```
skyrpex, line 197: Are all these objects not needed anymore because of the string to json change? 🤔

libs/wingsdk/test/simulator/simulator.test.ts
Chriscbr: Ah good idea, I'll try that out

tools/hangar/__snapshots__/test_corpus/sdk_tests/function/aws-function.test.w_compile_tf-aws.md
Chriscbr: I think this is just because I refactored the test a little bit - but yeah otherwise this would be pretty odd

examples/tests/invalid/cloud_function_expects_inflight.test.w
Chriscbr: I'll open a discussion. Let's address this when we're cleaning up the rest of the stringification system (toString methods etc.)

Chriscbr (approved):

skyrpex (approved):

PR #6231, opened 2024-04-14
fix: update tree-sitter to latest

MarkMcCulloh:
The primary goal of this PR is to update tree-sitter to the latest version. This resulted in:
- Pinned tree-sitter to a git commit dependency until https://github.com/tree-sitter/tree-sitter/commit/a7a47d561d4e64eaf226f93c4d68076afa67fdda is released. tree-sitter's release schedule is unpredictable and we don't publish any rust crates so this should be fine for now
- The `tree-sitter-traversal` dependency is removed since it doesn't support newer versions of tree-sitter, so I just added some code for the equivalent functionality
- Re-generated language bindings for tree-sitter. Due to a slight change in how tree-sitter builds it's hard to fully opt out of these (we do use the rust one). Luckily they should not require changing for a while
- Added some additional ts queries and query tests, taken from the neovim tree-sitter repo
- Changes to the scanner and rust parser to match the previous behavior of AUTOMATIC_SEMICOLON which had been broken
- Changes to the scanner and rust parser to match the previous behavior of named and positional arguments. I didn't get an exact match but I think most cases actually look a little better now

Most of the new files here are a result of the binding regeneration and can kinda be ignored.

Side note: Also noticed that the automatic semicolon insertion hasn't been working for a while so I added a fix for that (see `lsp.ts`)


apps/wing/src/commands/lsp.ts
```diff
@@ -136,9 +136,14 @@ export async function lsp() {
     for (const rd of raw_diagnostics) {
       if (rd.span) {
         const diagnosticUri = "file://" + rd.span.file_id;
+        let message = rd.message;
```
eladb, line 139: Test for this maybe?

eladb (approved):

clason:
> Pinned tree-sitter to a git commit dependency until https://github.com/tree-sitter/tree-sitter/commit/a7a47d561d4e64eaf226f93c4d68076afa67fdda is released. tree-sitter's release schedule is unpredictable and we don't publish any rust crates so this should be fine for now

Note that this doesn't help downstream users, who will have their own tree-sitter version (necessarily so, since they need to generate parsers for multiple different grammars). This PR will unbreak nvim-treesitter (until the next breaking change; tree-sitter is in a bit of flux right now), but it's not really sustainable. Not sure how to deal with this better, though; just wanted to make a note of this.

apps/wing/src/commands/lsp.ts
MarkMcCulloh: Unfortunately we don't have any fixture/system for e2e testing of the lsp and unit testing this particular function wouldn't be too useful imo

PR #6235, opened 2024-04-14
feat(cli): wing secrets

hasanaburayyan:
Intial poc code..

## TODO:
- [ ] Implement storeSecrets() on `tf-aws` platform
- [ ] Remove returning String from `storeSecrets()` api
- [ ] Dont print secrets.json file if no secrets in application
- [ ] some more things I need to re-review

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -110,52 +112,45 @@ export class PlatformManager {
...
+    for (const c of app.node.findAll()) {
+      if ((c as any)[SECRET_SYMBOL]) {
+        const secret = c as any;
+        secretNames.push(secret.name);
+      }
+    }
+
+    if (secretNames.length > 0) {
+      writeFileSync(
+        join(app.outdir, "secrets.json"),
```
eladb, line 133: const

apps/wing/src/commands/secrets.ts
```diff
@@ -0,0 +1,48 @@
...
+import { CompileOptions, compile } from "./compile";
+
+export interface SecretsOptions extends CompileOptions {
+  readonly list?: boolean;
+}
+
+export async function secrets(entrypoint?: string, options?: SecretsOptions): Promise<void> {
+  // Compile the program to generate secrets file
+  const outdir = await compile(entrypoint, options);
+  const secretsFile = path.join(outdir, "secrets.json");
```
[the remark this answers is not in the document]
hasanaburayyan: Great callout, yea ill update

libs/wingsdk/src/platform/platform-manager.ts
```diff
@@ -97,19 +104,14 @@ export class PlatformManager {
...
-
-    let appCall = this.platformInstances[0].newApp;
-
-    if (!appCall) {
-      throw new Error(
-        `No newApp method found on platform: ${this.platformPaths[0]} (Hint: The first platform provided must have a newApp method)`
-      );
+  public async storeSecrets(secretNames: string[]): Promise<any> {
+    if (!this.storeSecretsHook) {
+      throw new Error("No createSecrets method found on platform");
```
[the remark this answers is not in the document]
hasanaburayyan: @eladb not sure its that straight forward, originally I wanted to go with that option of just having platform[0] be the expected platform for storing secrets however I imagine that some `tf-aws` users might want to override this if they are using a custom platform that uses secure strings in SSM. 

So the behavior atm is the last platform that contains the `storeSecrets` hook is the one that is used. This way I could do `wing compile -t tf-aws -t my-secrets-platform`

So now to the error message, if someone compiles with `wing compile -t x -t y -t z` and none of them containe a `storeSecrets` hook, which platform do we complain about

In my most recent commits I update this error to be
```
throw new Error(`No storeSecrets method found on any platform`);
```

let me know what you think.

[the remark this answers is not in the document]
MarkMcCulloh: I like Elad's suggestion because I think for the average user the mention of `createSecrets` would be confusing. It's also not reasonably actionable since most people won't jump to implementing their own handling of this. That might be an error that shows up with `--verbose`.

As for which platform to mention in the error, would say we could mention all of them (with special handling for the common case of 1 platform)

libs/wingsdk/src/target-tf-aws/platform.ts
```diff
@@ -1,3 +1,9 @@
+import {
+  SecretsManagerClient,
+  GetSecretValueCommand,
+  CreateSecretCommand,
+  UpdateSecretCommand,
+} from "@aws-sdk/client-secrets-manager";
```
MarkMcCulloh, line 6: Could this be changed to a dynamic import that is only loaded if there are secrets?
This platform will be loaded for all tf-aws apps and unfortunately AWS SDKs are expensive. The benchmarks are showing an flat increase of almost ~100ms which is not negligible.

libs/wingsdk/src/target-sim/secret.inflight.ts
```diff
@@ -16,7 +15,8 @@ export class Secret implements ISecretClient, ISimulatorResourceInstance {
   private readonly name: string;
 
   constructor(props: SecretSchema) {
-    this.secretsFile = path.join(os.homedir(), ".wing", "secrets.json");
+    this.secretsFile = path.join(process.cwd(), ".env");
```
MarkMcCulloh, line 18: This env file should already be loaded by the CLI and available in the preflight process.env
Also, it's worth noting that it may not only contain secrets.

MarkMcCulloh: nvm this is the inflight file

libs/wingsdk/src/target-sim/platform.ts
```diff
@@ -11,4 +12,32 @@ export class Platform implements IPlatform {
...
+
+    for (const key in secrets) {
+      existingSecrets[key] = secrets[key];
+    }
+
+    const updatedContent = Object.entries(existingSecrets)
+      .map(([key, value]) => `${key}=${value}`)
+      .join("\n");
+
+    fs.writeFileSync("./.env", updatedContent);
```
MarkMcCulloh, line 39: Is it possible to ensure the loaded data stays in memory? Does secret.inflight.ts have access to the envs loaded in the simulator so that we don't need to deal with files?

libs/wingsdk/src/platform/platform-manager.ts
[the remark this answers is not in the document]
eladb: Perhaps the error could be:

```
Cannot find a platform or platform extension that supports storing secrets
```

eladb (approved):

libs/wingsdk/src/target-tf-aws/platform.ts
hasanaburayyan: Yea Im pretty sure we can use dynamic imports. Thank goodness for the benchmarks @MarkMcCulloh because I ALWAYS overlook using dynamic imports 😁

libs/wingsdk/src/target-sim/platform.ts
hasanaburayyan: Yea you are right this can all stay in memory

libs/wingsdk/src/target-sim/platform.ts
```diff
@@ -11,4 +12,32 @@ export class Platform implements IPlatform {
   public newApp(appProps: any): any {
     return new App(appProps);
   }
+
+  public async storeSecrets(secrets: Record<string, string>): Promise<void> {
+    let existingSecretsContent = "";
+    try {
+      existingSecretsContent = fs.readFileSync("./.env", "utf8");
```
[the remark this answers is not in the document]
hasanaburayyan: True, but without adding maybe a prefix to the environment variables as they are loaded in, im not sure how else you could identify what values are from .env file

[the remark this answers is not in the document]
MarkMcCulloh: True, this function is basically just worried about managing the env file anyways

libs/wingsdk/src/target-sim/platform.ts
MarkMcCulloh, line 39: Should the .env file be relative to the entrypoint rather than cwd?

hasanaburayyan: Yea actually I think I was using WING_SOURCE_DIR originally and ran into an issue with it not being set, forgot to give myself a TODO, but its fixed now

hasanaburayyan:
@mergify refresh

Issue #5916, opened 2024-03-12
`MutJson.del`


eladcon:
### Use Case

want to delete a key of a json object

```wing
let x = MutJson{a:1};
x.del("a");
```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

partha04patel:
If no one’s working on this/ no one minds, I could try giving this a shot!

staycoolcall911:
Sounds good, @partha04patel - you got it!
In case you have any questions, you're welcome to ask them here or on our [Slack’s `#dev` channel](https://t.winglang.io/slack)

staycoolcall911:
This was implemented in #6236

PR #6236, opened 2024-04-14
feat(sdk)!: add delete method for MutJson class

partha04patel:
## Checklist

- [x] Tests added (always)

BREAKING CHANGE: `has()` removed from static `Json`. It is now an instance method on `Json` objects.


partha04patel:
meant as resolution for https://github.com/winglang/wing/issues/5916

examples/tests/sdk_tests/std/json.test.w
```diff
@@ -178,4 +178,25 @@ test "deepCopy(), deepCopyMut()" {
   assert(copy != copyMut);
 
   assert(copyMut.get("object") == mutation);
+}
+
+test "delete() for MutJson" {
+  let mutObj = MutJson { x: 1, y: 2 };
+  mutObj.delete("x");
+  let assertThrows = (expected: str, block: (): void) => {
```
eladb, line 186: I think you can just use `Json.has(x)` instead to verify that the key was deleted. No?

partha04patel: I don't think I can since the has() method doesn't exist for the MutJson class from what I understand

partha04patel: has() method added to MutJson class as part of this PR!

MarkMcCulloh (approved):

libs/wingsdk/src/std/json.ts
```diff
@@ -453,4 +449,31 @@ export class MutJson {
...
+   * @macro (delete ($self$)[$args$])
+   *
+   * @param key The key
+   * @returns true if the given key is no longer present
+   */
+
+  public delete(key: string): boolean {
+    key;
+    throw new Error("Macro");
+  }
```
Chriscbr, line 465: What happens if `delete()` is called on something that isn't a map? like `Json true` or `Json [1, 2, 3]`? Can we add tests for this?

MarkMcCulloh: It should return true (matching the return comment). 
null is also technically valid json and that would throw here (bad), but most of our json methods wouldn't correctly handle null anyways.
Agree a test would be good, my b for approving without more tests.

partha04patel: I could try adding extra tests for some of those edge cases in another PR!

PR #6238, opened 2024-04-14
fix(vscode): wing debugger is not a default choice for .w files

MarkMcCulloh:
Missed this change in https://github.com/winglang/wing/pull/6217

Now vscode will automatically use the builtin debugger when you have a wing file open instead of asking you to pick one


staycoolcall911 (approved):
Amazing!

PR #6243, opened 2024-04-15
feat(console): map v2

skyrpex:
Changes the Console map's relationships and look & feel. Now, the relationships show the data flow of the application. In other words, it shows inflight to inflight relationships.

There are planned tasks to improve the map:

- Hide the implementation of complex nodes, and allow diving into these nodes to see their implementation
- Clean the implementation now that everything is clear on hindsight
- Avoid multiple connections to the same inflight from merging


apps/wing-console/console/app/demo/main.w
```diff
@@ -162,7 +162,6 @@ test "Add fixtures" {
   counter.inc(100);
 }
```
eladb, line 163: Does this have to be laid out horizontally? Is it possible to distribute them throughout the space?

Screenshot 2024-05-06 at 16 16 38

skyrpex: There's some configuration options we can use to hint the overall direction of the map. Can you share this app's code so I can try it?

skyrpex:
Changing to draft since I'm going to apply some changes.

apps/wing-console/console/app/demo/main.w
eladb: That's okay for now.

apps/wing-console/console/app/demo/main.w
```diff
@@ -162,7 +162,6 @@ test "Add fixtures" {
...
 
-
 class WidgetService {
```
eladb, line 165: @skyrpex I am trying this code:

```js
bring cloud;

class Foo {
  new() {
    let b = new cloud.Bucket();
  }
}

let foo = new Foo() as "foo";
for i in 0..10 {
  let bar = new Foo() as "bar-{i}";
  nodeof(foo).addConnection(source: foo, target: bar, name: "boom");
}
```

And I see a few issues:

1. I don't see the buckets
2. When I click on the connections it really show me the interaction panel for the connection.
3. The layout is weird. I expected something that is more organized...

image


I think it's important to make sure that *all connections are displayed*, not only the ones that connect to inflight methods!

skyrpex: > I don't see the buckets

Solved the buckets issue.

> When I click on the connections it really show me the interaction panel for the connection

Not sure what you mean...

> The layout is weird. I expected something that is more organized...

From what I've seen, this is controlled by the [Node Layering Strategy](https://eclipse.dev/elk/reference/options/org-eclipse-elk-layered-layering-strategy.html). If we use `MIN_WIDTH`, it's layered mostly horizontally like your screenshot. If we use `NETWORK_SIMPLEX`, it's layered mostly vertically but it's more organized.

There are more options for the layering strategy, and some of them are "advanced" which means that they are further controlled by specific options. My take right now is to use `NETWORK_SIMPLEX`, which is more vertical'ish but feels more organized:

image

And then, keep tweaking all these advanced configuration options until we find what we're looking for.

ainvoner: @skyrpex we can also choose some preferred layout presets and enable users to choose their favorite according to their applications using a simple dropdown

ainvoner: I'll open an issue so we will be able to discuss it

skyrpex:
Found a scrolling bug, will fix.

ainvoner (approved):

ainvoner:
Great job man!!!!

skyrpex:
> Great job man!!!!

Thank you!

eladb:
Amazing work!

PR #6247, opened 2024-04-16
feat: support file browser in bring ui

### LINKED ISSUE #6198: support file explorer in `bring ui`
### Use Case

It would be possible to create UI components that can list/delete/upload files and their content in a given local/remote directory, similar to what is available on `cloud.Bucket`

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

polamoros:
Resolves #6198


libs/wingsdk/src/ui/file-browser.ts
```diff
@@ -0,0 +1,159 @@
...
+  private readonly label: string;
+
+  constructor(
+    scope: Construct,
+    id: string,
+    label: string,
+    putHandler: IFileBrowserPutHandler,
+    deleteHandler: IFileBrowserDeleteHandler,
+    getHandler: IFileBrowserGetHandler,
+    listHandler: IFileBrowserListHandler
```
eladb, line 55: I think these should all be keywords ("props"):

```suggestion
    props: FileBrowserHandlers,
```

And then:

```js
extern interface FileBrowserHandlers {
  putHandler: IFileBrowserPutHandler;
  deleteHandler: IFileBrowserDeleteHandler;
  getHandler: IFileBrowserGetHandler;
  listHandler: IFileBrowserListHandler;
}
```

eladb (dismissed):

apps/wing-console/console/app/demo/main.w
```diff
@@ -1,13 +1,64 @@
...
+}
+
+class myBucket {
+  b: cloud.Bucket;
+  new() {
+    this.b = new cloud.Bucket();
+    new ui.FileBrowser("File Browser",
+      {
+        put: inflight (payload: str) => {
+          let data: PutData = PutData.fromJson(Json.parse(payload));
```
ainvoner, line 30: is there a way to avoid it and pass a struct instead of a str?

ainvoner:
@staycoolcall911 it is not only the console. i'm renaming it again

apps/wing-console/console/app/demo/main.w
ainvoner: there is also this one:
https://github.com/winglang/wing/pull/6219

ainvoner: @skyrpex but the implementation of the handlers is done in the custom resource `.w` file

apps/wing-console/console/server/src/router/app.ts
```diff
@@ -527,6 +525,95 @@ export const createAppRouter = () => {
         ) as IFunctionClient;
         await client.invoke("");
       }),
+
+    "app.invokeFileBrowserGetQuery": createProcedure
```
skyrpex, line 529: Maybe we could use different prefix and files to implement this (also, why `...GetQuery` instead of `get`?). For example, `router/file-browser.ts` and `fileBrowser.getObject`?

ainvoner: I agree, we should have a separate router file for each custom resource

polamoros: done!

ainvoner:
@Chriscbr thanks for help :) your handlers conversions works!

PR #6252, opened 2024-04-16
feat(compiler): index arrays, maps, and Json with `a[b]` syntax

### LINKED ISSUE #5702: Support array `[]` accessors
### Use Case

Instead of

```js
let arr = MutArray<str>["a", "b", "c"];
assert(arr.at(0) == "a");
arr.set(0, "A");
```

You can write

```js
let arr = ["a", "b", "c"];
assert(arr[0] == "a");
arr[0] = "A";
```

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

Language Design, Compiler

### Community Notes

Chriscbr:
Wing now supports a dedicated array indexing syntax `arr[index]` for getting and setting values within arrays, maps, JSON values, and strings. Here are some examples:

```js
// accessing values to perform simple lookups
let arr = MutArray<num>[3, 5];
assert(arr[0] == 3);
assert(arr[1] == 5);
assert(arr[-1] == 5);
assert(arr[-2] == 3);

arr[42]; // throws an error

// setting values
arr[0] = 42; 
arr[1] += 3.5;
```

Here are the types that can be indexed:

* `Array` and `MutArray` - accepts a `num` index
* `Map` and `MutMap` - accepts a `str` index
* `str` - accepts a `num` index
* `Json` and `MutJson` - accepts `num` and `str` index values

When accesses are performed, an error is thrown if the index is out of bounds or the key is invalid. If you want to obtain a value defensively (e.g. you want to see if a `Json` has a particular key or not, without a runtime exception), you can fall back to dedicated methods like `array.tryAt(0)` or `myjson.tryGet("key")` which return optional values.

Closes #5702

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/helpers.ts
```diff
@@ -55,6 +55,100 @@ export function unwrap<T>(value: T): T | never {
...
+
+  return obj[index];
+}
+
+export function assign(
+  obj: any,
+  index: string | number,
+  kind: "=" | "+=" | "-=",
+  value: any
+) {
```
MarkMcCulloh, line 104: nit: there's some duplication between `assign` and `lookup`

examples/tests/valid/indexing.test.w
```diff
@@ -0,0 +1,68 @@
+bring cloud;
+bring expect;
+
+let arr = Array<num>[1, 2, 3];
+
+assert(arr[0] == 1);
```
MarkMcCulloh, line 6: Spec update? Doc update would be nice as well, but IMO probably fine to get this PR in first.

MarkMcCulloh (approved):
Noice

examples/tests/valid/indexing.test.w
yoav-steinberg, line 1: 1. Maybe add tests with an expression as the index: `arr[1+2]` or `my_map[some_str.substring(1,1)]`.
2. Maybe some tests regarding indexing into an indexed value (to verify precedence works correctly): `arr[x][y][z]`.

libs/wingc/src/ast.rs
```diff
@@ -742,6 +742,8 @@ pub enum Reference {
 		property: Symbol,
 		optional_accessor: bool,
 	},
+	/// A reference to an accessed member of an object `expression[x]`
+	ElementAccess { object: Box<Expr>, index: Box<Expr> },
```
yoav-steinberg, line 746: I'd like to re-thing this as a "reference" and perhaps make `ElementAccess` just another `ExprKind`. References are are expressions that have a name but `ElementAccess` is just two expressions (object and index) with no name: `[1,2,3,4][0]` should evaluate to 1.

libs/wingc/src/type_check.rs
```diff
@@ -4485,30 +4493,112 @@ impl<'a> TypeChecker<'a> {
...
+				| Type::Optional(_)
+				| Type::Set(_)
+				| Type::MutSet(_)
+				| Type::Function(_)
+				| Type::Class(_)
+				| Type::Interface(_)
+				| Type::Struct(_)
+				| Type::Enum(_) => self.spanned_error(
+					variable,
+					format!("Unsupported reassignment of element of type {}", var_type),
```
yoav-steinberg, line 4571: Was there a test for this?

libs/wingc/src/type_check.rs
```diff
@@ -5793,9 +5900,97 @@ impl<'a> TypeChecker<'a> {
...
+					Type::Anything => (
+						ResolveReferenceResult::Location(self.types.anything(), self.types.anything()),
+						instance_phase,
+					),
+					Type::Unresolved => (
+						ResolveReferenceResult::Location(self.types.error(), self.types.error()),
+						instance_phase,
+					),
+					Type::Inferred(_) => (
+						ResolveReferenceResult::Location(self.types.error(), self.types.error()),
```
yoav-steinberg, line 5968: Is this really an error? Not sure how this works exactly.

yoav-steinberg (commented):
:sparkles:

eladb (approved):
hell yeah!

libs/wingc/src/ast.rs
yoav-steinberg: The difference being that anything after the `.` is a name. Json is the exception with it's dotted access to non-fields, but for any typed "object" access you always access a name. Even in json there's no real way to access an expression after the `.`: `my_json.this_cant_be_an_expression`.

MarkMcCulloh: > The difference being that anything after the . is a name.

True, but that's more of an optimization, both for our implementation and for readability. We could easily allow `a."b"` or even `a."{some string expression}"`. Not saying we should, but we could. 
The crux of my concern is that I'm not sure we're actually making use of idea that "References are expressions that have a name" anyways.

Again I'm actually in favor of breaking down `reference` into more useful parts. Like considering lvalues and rvalues for expressions and/or seeing `.` and `[]` as operators.

libs/wingc/src/type_check.rs
Chriscbr: I didn't, but I can try to add one. It's actually kinda tricky coming up with examples that produce this error, but it seems to correspond to degenerate cases like:

```
let var x = 3 + "hello";
x["hi"] = 4; // error: Unsupported reassignment of element of type unresolved
```

libs/wingc/src/type_check.rs
Chriscbr: I wasn't sure how to get inferred types working here so for now I've made it an error (I've added a test case just now). Open to improving this

libs/wingsdk/src/helpers.ts
```diff
@@ -55,6 +55,100 @@ export function unwrap<T>(value: T): T | never {
...
+  if (typeof index !== "string" && typeof index !== "number") {
+    throw new TypeError(
+      `Index must be a string or number (found "${typeof index}")`
+    );
+  }
+
+  if (typeof index === "number") {
+    if (
+      !Array.isArray(obj) &&
+      !Buffer.isBuffer(obj) &&
```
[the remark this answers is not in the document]
Chriscbr: Ah, I think copilot might generated this lol

libs/wingsdk/src/helpers.ts
Chriscbr: Refactored a little bit

examples/tests/valid/indexing.test.w
Chriscbr: 👍 Added a section to the language reference

examples/tests/valid/indexing.test.w
Chriscbr: Added

libs/wingc/src/ast.rs
yoav-steinberg: Right, I wasn't thinking of the left side of the assignment. I recall explaining to someone in the past that in our AST "reference" is actually "lvalue", so I'm convinced what you did is the best option given the current design.
Maybe we should change `Reference` to `LValue`..

PR #6258, opened 2024-04-17
fix(compiler): mutable lifted object may be modified inflight

### LINKED ISSUE #3069: prevent reassigning and mutating lifted objects
Following https://github.com/winglang/wing/pull/3064

It is now possible to capture reassignable/mutable objects
We need to prevent these captured object from being reassigned/modified in inflight

yoav-steinberg:
fixes #3069
The type checker will now modify the type of preflight expressions that are being used inflight to a non-mutable version of the same type: `MutArray` -> `Array`, `MutJson` -> `Json`...

Note that since we change the type of the expression we don't get descriptive error telling us that we're trying to mutate a lifted mutable object, instead we just get the same error we'd expect if the lifted object was non-mutable to begin with. This isn't ideal but it makes the fix simpler. Let me know what you think...

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/invalid/un_mut_lifted_objects.test.w
+++ b/examples/tests/invalid/un_mut_lifted_objects.test.w
@@ -10,10 +10,20 @@ st.add(3);
 let mp = MutMap<num>{"a" => 1};
 mp.set("a", 2);
 
+let opt_ar: MutArray<num>? = MutArray<num>[];
+opt_ar?.push(1);
+
+let recursive_ar = MutArray<MutArray<num>>[];
+recursive_ar.push(MutArray<num>[1]);
+recursive_ar.at(0).push(2);
+
 inflight() => {
   // Same code as above should be an error in inflight
-  ar.push(2); // Error: push doesn't exist in Array
+  ar?.push(2); // Error: push doesn't exist in Array
   j.set("a", 3); // Error: set doesn't exist in Json
   st.add(4); // Error: add doesn't exist in Set
   mp.set("a", 3); // Error: set doesn't exist in Map
+  opt_ar?.push(2); // Error: push doesn't exist in Array
+  recursive_ar.push(MutArray<num>[2]); // Error: push doesn't exist in Array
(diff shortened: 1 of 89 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -5599,6 +5604,29 @@ impl<'a> TypeChecker<'a> {
 			.map(|_| base_udt)
 	}
 
+	fn make_immutable(&mut self, type_: TypeRef) -> TypeRef {
+		match *type_ {
+			Type::MutArray(inner) => {
```
eladb, line 5609: what about `MutJson`?

eladb (commented):
Wow that was easier than expected...

libs/wingc/src/type_check.rs
yoav-steinberg: What about it? It's handled 4 lines down. Am I missing something?

eladb: No, I've missed it. Sorry!

eladb (commented):
Maybe add to docs?

examples/tests/invalid/un_mut_lifted_objects.test.w
```diff
@@ -0,0 +1,29 @@
...
+inflight() => {
+  // Same code as above should be an error in inflight
+  ar?.push(2); // Error: push doesn't exist in Array
+  j.set("a", 3); // Error: set doesn't exist in Json
+  st.add(4); // Error: add doesn't exist in Set
+  mp.set("a", 3); // Error: set doesn't exist in Map
+  opt_ar?.push(2); // Error: push doesn't exist in Array
+  recursive_ar.push(MutArray<num>[2]); // Error: push doesn't exist in Array
+  recursive_ar.at(0).push(3); // Error: push doesn't exist in Array
+};
```
Chriscbr, line 29: Maybe double check this works with the new array assignment syntax:

```
ar[0] = 5; // Error: Cannot update elements of an immutable Array
```

Chriscbr (approved):
Slick 🚢

PR #6278, opened 2024-04-19
fix(sdk): improve delete() method for MutJson class

partha04patel:
## Checklist

- [x] Tests added (always)


partha04patel:
Intended for having delete() work for non-map MutJson instances as brought up by @Chriscbr and @MarkMcCulloh from https://github.com/winglang/wing/pull/6236

docs/docs/04-standard-library/std/json.md
```diff
@@ -450,16 +450,16 @@ Convert Json element to string if possible.
...
 
 Removes the specified element from a map.
 
 ###### `key`<sup>Required</sup> <a name="key" id="@winglang/sdk.std.MutJson.delete.parameter.key"></a>
 
-- *Type:* str
+- *Type:* any
 
-The key.
+Either the key or index to delete.
```
hasanaburayyan, line 462: Im not sure we should allow deletion at an index. While it may seem like order is preserved in most JSON objects, this is not a guarantee, according to the [json specification](https://www.json.org/json-en.html) "An object is an unordered set of name/value pairs"

hasanaburayyan (changes requested):

docs/docs/04-standard-library/std/json.md
hasanaburayyan: It makes sense for Json Arrays, since their order is preserved but perhaps in these cases its better to just convert types

examples/tests/sdk_tests/std/json.test.w
```diff
@@ -186,4 +186,26 @@ test "delete() for MutJson" {
...
+    "true": true,
+    "false": false,
+    "object": {
+      "key1": "value1",
+      "key2": 2,
+      "key3": false,
+      "key5": [3, 2, 1]
+    }
+  });
+  assert(original.delete("key5")==true);
```
hasanaburayyan, line 210: This returning true feels off. Since "key5" is not a top level key of `original` it should be `original["object"].delete("key5")`

partha04patel: I can see why it feels off,  technically the returning true makes sense since "key5" was never present as a top level key of `original`, but `original["object"].delete("key5")` should also return true as well with it actually deleting the present key

revitalbarletz:
@partha04patel do you need any additional help here? seems like we are almost done!

PR #6286, opened 2024-04-21
fix(platforms): unable to perform awscdk context lookups

### LINKED ISSUE #6279: Unable to perform lookups in `awscdk` based platforms
### I tried this:

compile this code:
```js
// main.w
bring cloud;

let b = new cloud.Bucket();
```

using this platform:
```js
// platform.js
const ec2 = require("aws-cdk-lib/aws-ec2");
const cdk = require("aws-cdk-lib");
const platformCDK = require("@winglang/platform-awscdk/lib");
const core = require("@winglang/sdk/lib/core");

class Platform extends platformCDK.Platform {
  newApp(props) {
    return new platformCDK.App({
      ...props,
      stackFactory: (app, stackName) => {
        return new cdk.Stack(app, stackName, {
          env: {
            account: '123456789',
            region: 'us-east-1'
          }
        });
      }
    });
  }

  newInstance(type, scope, id, ...args) {
    if (type === "@winglang/sdk.cloud.Bucket") {
      const app = core.App.of(scope);
      
      const vpc = ec2.Vpc.fromLookup(app, 'VPC', {
        vpcId: 'my-vpc'
      })

      // Create a subnet just so we have something in the CF template
      new ec2.Subnet(scope, 'Subnet', {
        availabilityZone: 'us-east-1a',
        cidrBlock: '10.0.0.0/24',
        vpcId: vpc.vpcId
      })
    }
  }
}

module.exports = {
  Platform
}
```

by running `wing compile main.w -t platform.js`

### This happened:

The generated cloudformation template has a vpcID that comes from the dummy values in `awscdk` context lookup:
```json
{
...
"Subnet53F41B8F": {
   "Type": "AWS::EC2::Subnet",
   "Properties": {
    "AvailabilityZone": "us-east-1a",
    "CidrBlock": "10.0.0.0/24",
    "Tags": [
     {
      "Key": "Name",
      "Value": "hasanstack/Default/Default/Subnet"
     }
    ],
    "VpcId": "vpc-12345"
   }
  },
...
```

See more on Dummy Values here: https://github.com/aws/aws-cdk/blob/9eb21bd79ca69a1f3f0cabf7d31810941f17da8b/packages/aws-cdk-lib/core/lib/context-provider.ts#L101

### I expected this:

The lookup to have been successful

### Is there a workaround?

Not that Im aware	

### Anything else?

We think the issue is because we are calling `.synth()` directly when synthesizing a `awscdk` app, and the mechanism for performing lookups and storing the values in a context.json file is done through a feature in the CLI

How `awscdk` platform synthesizes: https://github.com/winglang/wing/blob/main/libs/awscdk/src/app.ts#L136

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

eladb:
In order to fix #6279, we need synthesis to be triggered by the CDK CLI instead of directly from the Wing CLI.

This is actually really easy to do. All you need is to create a `cdk.json` file (or use one from `cdk init`) and modify the `app` and the `output` options like so:

```json
{
  "app": "CDK_STACK_NAME=MyStack wing compile --platform @winglang/platform-awscdk main.w",
  "output": "target/main.awscdk"
}
```

Then, you can simply use `cdk deploy`, `cdk diff`, `cdk synth` as if it was a normal CDK app. No need to explicitly interact with the Wing CLI in this case.

To allow context lookups, we need to support specifying the stack's AWS environment (account/region), so two new environment variables have been added: `CDK_AWS_ACCOUNT` and `CDK_AWS_REGION`. These can be set together with `CDK_STACK_NAME` in the `cdk.json` file mentioned above.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/docs/055-platforms/awscdk.md
+++ b/docs/docs/055-platforms/awscdk.md
@@ -6,49 +6,160 @@ description: AWS CDK platform
 keywords: [Wing reference, Wing language, language, Wing language spec, Wing programming language, aws, awscdk, amazon web services, cloudformation]
 ---
 
-The `@winglang/platform-awscdk` [platform](../02-concepts/03-platforms.md) compiles your program for the AWS CDK (CloudFormation).
+The `@winglang/platform-awscdk` [platform](../02-concepts/03-platforms.md) compiles your program for
+the AWS CDK and deployed through the CDK CLI (and AWS CloudFormation).
+
+## Prerequisites
+
+* Install the AWS CDK (or use via `npx cdk`):
+  ```sh
+  npm i aws-cdk
+  ```
+
+* Install the Wing `awscdk` platform:
+  ```sh
+  npm i @winglang/platform-awscdk
+  ```
 
 ## Usage
 
-You will need to install the `@winglang/platform-awscdk` library in order to use this platform.
+Let's create `main.w` with our Wing program:
 
-```sh
-$ npm i @winglang/platform-awscdk
+```js
+bring cloud;
+
+new cloud.Bucket();
 ```
 
-This platform requires the environment variable `CDK_STACK_NAME` to be set to the name of the CDK
-stack to synthesize.
+> At this point, you can just run `wing it` (or `wing run`) to open the Wing Simulator.
 
-```sh
-$ export CDK_STACK_NAME="my-project"
-$ wing compile --platform @winglang/platform-awscdk [entrypoint]
+To use Wing with the AWS CDK, we will need to tell the CDK CLI to run `wing compile` as the CDK app
+and that the synthesis output is in the Wing's target directory.
+
+This can be done by creating a `cdk.json` file manually or through `cdk init` and editing the `app`
+and the `output` fields:
+
+```json
+{
+  "app": "CDK_STACK_NAME='MyStack' CDK_AWS_ACCOUNT='111111555555' CDK_AWS_REGION='us-east-1' wing compile --platform @winglang/platform-awscdk main.w",
+  "output": "target/main.awscdk",
+
+  // ... rest of cdk.json
+}
 ```
 
-## Parameters
+The `awscdk` platform uses the following environment variables as configuration options:
 
-The `CDK_STACK_NAME` environment variable specifies the name of the CDK stack to synthesize.
+* `CDK_STACK_NAME` (required) - sets the CloudFormation stack name to use.
+- `CDK_AWS_ACCOUNT` and `CDK_AWS_REGION` (optional) - the AWS environment for deployment and context
+  lookups (e.g. VPC lookups). The default is to use the AWS account region defined in the CLI
+  environment.
 
-## Output
+Now, the AWS CDK CLI will work as normal:
 
-The output includes both a AWS-CDK configuration file (under `target/<entrypoint>.awscdk`) and
-JavaScript bundles that include inflight code that executes on compute platforms such as AWS Lambda.
+* `npx cdk bootstrap` bootstrap your AWS account for AWS CDK use (once per account/region).
+* `npx cdk deploy` deploy the Wing app to your default AWS account/region.
+* `npx cdk synth` synthesize output to `cdk.out`.
+* `npx cdk diff` show a diff between your code and the deployed version
 
 ## Deployment
 
-To deploy your app, you will first need to install the [AWS CDK
-CLI](https://docs.aws.amazon.com/cdk/v2/guide/cli.html).
-
-If not previously done, you will need to bootstrap your environment (account/region):
+Before you can first deploy an AWS CDK app to your account, you'll need to bootstrap the account/region:
 
 ```sh
-$ cdk bootstrap --app target/app.awscdk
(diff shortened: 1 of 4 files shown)


docs/docs/055-platforms/awscdk.md
```diff
@@ -6,50 +6,161 @@ description: AWS CDK platform
...
+* `CDK_STACK_NAME` (required) - sets the CloudFormation stack name to use.
+- `CDK_AWS_ACCOUNT` and `CDK_AWS_REGION` (optional) - the AWS environment for deployment and context
+  lookups (e.g. VPC lookups). The default is to use the AWS account region defined in the CLI
+  environment.
+
+Now, the AWS CDK CLI will work as normal:
+
+* `npx cdk bootstrap` bootstrap your AWS account for AWS CDK use (once per account/region).
+* `npx cdk deploy` deploy the Wing app to your default AWS account/region.
+* `npx cdk synth` synthesize output to `cdk.out`.
```
MarkMcCulloh, line 62: Shouldn't the platform be responsible for this? The workflow described in this document no longer feels like a wing application, but rather an aws cdk one that happens to be written in wing.
Does the CLI expose anything programmatically we can use to solve this instead?

eladb: Wing doesn't take care of bootstrapping/deployment/diffs. Similarly to how in Terraform we use the Terraform CLI for this.

It is an AWS CDK application written in Wing for all intents and purposes.

The CDK CLI doesn't have a good programmatic API.

I believe this DX is more idiomatic for CDK users, which is the purpose of this platform basically.

eladb: We should provide something like `wing new awscdk` to make this seamless, but for now, I am fine with requiring people to edit `cdk.json`.

MarkMcCulloh: My comment was specifically marked on `synth` i.e. `compile`, where wing does take ownership. This PR would take `wing` out of that loop that we claim to be a part of. Obviously it's still involved, but it's no longer the "entrypoint".

Eventually, I think wing should take ownership of common things like `diff` and `deploy` too, but until then I think it benefits us to not lostt

hasanaburayyan: Im not in love with this flow, but pragmatically I understand why this is the direction to go for now. 

I dont want to get too crazy with platform hooks but in theory we could introduce a `compile()` hook to platforms, and the `awscdk` platform might be able to obfuscate all this from the users who would still just run `wing compile`

But thats just a "from the hip" thought.

>The CDK CLI doesn't have a good programmatic API.
^😢

hasanaburayyan (approved):
Thanks for tackling this one @eladb, I think we should ship this for necessity. I do want to continue the conversation around this flow and if we can/should make this more "wing" native feeling. Ill make some notes and add a topic proposal to a team time.

docs/docs/055-platforms/awscdk.md
eladb: You can still do `wing compile` and it will behave like today. But when you want to deploy, you just deploy with `cdk deploy`.

The only difference from the current experience is that instead of doing `cdk deploy --app target/main.awscdk` you just run `cdk deploy`. It's way more idiomatic and much nicer from every aspect as far as I am concerned.

eladb:
> continue the conversation

Happy to continue the discussion. 


The way I see it, the Wing CLI is currently focusing on compilation, local development and testing. Deployment is out of scope and ideally should be as idiomatic as possible (and as standard as possible) to the target platform.

Which is why I think that for CDK users this is better and more idiomatic DX because that's exactly how they work with any CDK language, and I don't think Wing should be different.

hasanaburayyan:
> The only difference from the current experience is that instead of doing cdk deploy --app target/main.awscdk you just run cdk deploy. It's way more idiomatic and much nicer from every aspect as far as I am concerned.

You know, when I first started my line of thinking here, my mind thought that the current deployment modus operandi was to run `aws cloudformation create-stack` as the best way to deploy the output artifacts. But this is `awscdk` target not `cf-aws` so shame on me.

docs/docs/055-platforms/awscdk.md
MarkMcCulloh: > You can still do wing compile and it will behave like today. But when you want to deploy, you just deploy with cdk deploy.

Sure, but `wing compile` won't work correctly so it has effectively been removed as the correct option. Even though it just affects context lookup, we're still basically saying "don't use `wing compile`"

eladb: This is true for all AWS CDK languages. If you just compile + synthesize a CDK program without going through the CDK CLI context providers won't work.

eladb: Not sure what's disappointing about it TBH, but ok.

eladb: Not sure what's disappointing about it TBH, but ok.

PR #6291, opened 2024-04-22
fix(console): `option+cmd+i` command leaves the map in drag mode

skyrpex:
When a user presses `option+cmd+i` to open the devtools, a spacebar keypress event is fired (on Chrome, at least). This unexpected keypress event leaves the map in drag mode and nodes can't be clicked until the spacebar is pressed again.

This PR checks the `event.altKey` property to conditionally ignore this unexpected keypress event.


polamoros (approved):

skyrpex:
Thanks to @polamoros for finding out the cause of the bug!

PR #6299, opened 2024-04-22
fix(sdk): store simulator paths as absolute path

Chriscbr:
Normalize the code within the SDK so that the paths to various simulator files are stored as absolute paths.

This is causing an issue where the feature added in https://github.com/winglang/wing/pull/6295 only works when running `wing it` and not in `wing test` because `wing test` initializes the simulator with a relative path.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/tools/hangar/__snapshots__/error.ts.snap
+++ b/tools/hangar/__snapshots__/error.ts.snap
@@ -63,7 +63,7 @@ at <ABSOLUTE>/inflight_stacktraces.test.w:22:3
 27 |   assert(x, \\"x is false\\");
    |   ^
 at <ABSOLUTE>/inflight_stacktraces.test.w:27:3
-fail ┌ inflight_stacktraces.test.wsim » root/env0/test:assert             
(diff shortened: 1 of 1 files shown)


examples/tests/error/inflight_stacktraces.test.w
```diff
@@ -12,7 +12,7 @@ test "expect.equal" {
 }
 
 test "bucket failed get" {
-  bucket.get("doesn't exist");
+  bucket.delete("doesn't exist", mustExist: true);
```
MarkMcCulloh, line 15: Why change this particular test?

If it is needed, the test name should reflect it

MarkMcCulloh (approved):

examples/tests/error/inflight_stacktraces.test.w
Chriscbr: The error raised by `bucket.get()` in the simulator includes an absolute file system path where we the simulator tried to read a file from, which was resulting in a non-deterministic test snapshot. Just being a little lazy... I'll fix the test name. :)

PR #6303, opened 2024-04-23
fix(compiler): mutable lifted object may be modified inflight, 2nd attempt

### LINKED ISSUE #3069: prevent reassigning and mutating lifted objects

yoav-steinberg:
Fixes #3069 (2nd attempt, instead of #6258)
The type checker will now modify the type of preflight expressions that are being used inflight to a non-mutable version of the same type: MutArray -> Array, MutJson -> Json...

Note that since we change the type of the expression we don't get descriptive error telling us that we're trying to mutate a lifted mutable object, instead we just get the same error we'd expect if the lifted object was non-mutable to begin with. This isn't ideal but it makes the fix simpler. Let me know what you think...

To get this working we needed to do the following changes to how we evaluate the phases of reference expressions:
* The phase of phase independent properties inflight will be inflight. This makes sure inflight calls like `preflight_array.copyMut()` will be we evaluated inflight so the side effect of turning the result to immutable won't happen. This is a major change to lifting because inflight calls like `preflight_array.at(0)` will also be executed inflight (no matter what args are passed to the call).
* There's a special exception to the above, struct fields: we need to keep them phase independent and not turn them to inflight when accessing them inflight. This is required so we can lift qualify methods called on struct field when those fields are preflight objects. This needs to be addressed in the future, because somehow struct fields being "phase independent" doesn't seem correct.
* Index access (`[i]`) phase calculation was wrong and didn't take into account the phase of the index. When the index phase is inflight the expression must be inflight too. When the index phase is independent and the instance is preflight we want to lift the entire expression and not only the instance so it can be correctly lift-qualified.
* All this means we don't need the (admittedly beautiful) logic of only turning phase independent method calls to inflight calls when they are passed some inflight state (instance or arg). Now we always treat them as inflight. This cleaned up some code.

## Checklist

- [x] Tests added (always)


eladb (commented):
looks good!

eladb (approved):

PR #6305, opened 2024-04-23
feat: external inflight typescript files

### LINKED ISSUE #6045: Implicitly typed extern expression
### Use Case

Javascript/Typescript for inflight handlers. There is currently a lot of ceremony involved in using extern as a way to consume JS/TS. When doing this, it's assumed that the user knows the types they'll have available when calling the external file. This is cumbersome because it's more typical to call the inflight when consuming APIs that define the types for you.

Ideally, the source of truth for types should be within wing but declared when used. If this is the case, then why not go all the way and implicitly determine the type for the extern.

This use case is different from https://github.com/winglang/wing/issues/6042, where unequivocally the source of truth is the external file itself.

Here's an example of the minimum solution to call a typescript function inflight:

```wing
bring cloud;

class Extern {
  pub static extern "./oncreate.ts" inflight onCreate(req: cloud.ApiRequest): cloud.ApiResponse;
}

bucket.onCreate(inflight (req) => {
  // wrapper function needed because static functions can't be passed as-is (bug)
  return Extern.onCreate(req);
})
```

### Proposed Solution

Introduce a global function or new syntax for "closure loaded from file". This syntax will be an expression that loads an external file, like our current extern implementation. 
The desired type contract the file is expected to implement is based on the expression's inferred type.

**(syntax is a placeholder)**

```wing
bring cloud;

bucket.onCreate(extern("./oncreate.ts"))

// error: extern "./oncreate.ts" already has type `(event: OnCreateEvent): void`
bucket.onUpdate(extern("./oncreate.ts"))
```

Compiling this will generate `./oncreate.extern.d.ts` like the current extern behavior, using the type that Bucket.onUpdate expects as a default export. The .ts file can then be implemented like so:

```ts
// oncreate.ts
import extern from "./oncreate.extern.d.ts";

export const handler: extern = async (key, type) => {
  console.log(`File Created: ${key}`);
}

export default handler
```

The return type of extern() is inferred, which means it must be resolved in the same scope. It also must resolve to a function type.

```wing
// error: unable to infer type of extern
extern("./oncreate.ts")

// error: extern from be a function
let x: str = extern("./oncreate.ts")
```

### Implementation Notes

We may still want a way to explicitly type the extern. We could reuse the generics syntax:

```wing
let func = extern<inflight (): void>("./handler.ts")
```

#### Inflight Handler

```wing
bring cloud;

let api = new cloud.Api();

api.get("/", extern("./handler.ts"));
```

```ts
// handler.ts
import extern from "./handler.extern.d.ts";

export const handler: extern = async (req) => {
  console.log(req.path);
  return {
    status: 200
  }
}

export default handler
```

#### Lifting

```wing
bring cloud;

let api = new cloud.Api();

api.get(inflight (req) {
  lift { bucket: ["put"] } {
    return extern("./handler.ts")(req, bucket);
  }
});
```

```ts
// handler.ts
import extern from "./handler.extern.d.ts";

export const handler: extern = async (req, bucketClient) => {
  console.log(req.path);
  bucketClient.put("hi", "hello");
  return {
    status: 200
  }
}

export default handler
```

#### Preflight

```wing
interface UrlHelpers {
  parseURL: (url: str): Json;
  formatURL: (url: Json): str;
}

let helpers: UrlHelpers = extern("./url-helpers.ts");
```

### Community Notes

MarkMcCulloh:
Fixes #6045

This PR introduces a new utility function `@inflight`. This allows you to create an inflight closure from an external file with specific lifts:

```wing
// main.w
bring cloud;

bucket.onCreate(@inflight("./bucket_create.ts",
  export: "default",
//        ^ Optional named export from the file, default is the default export
  lifts:[{ obj: queue, alias: "myQueue", ops: ["push"] }],
//       ^ object to lift, can be any preflight expression
//                     ^ Optional alias, by default, this will be the variable name passed to obj
//                                     ^ methods to lift, if not provided then all methods will be granted
));
```

```ts
// bucket_create.ts
import inflight from "./.bucket_create.inflight";

export default inflight(async ({ myQueue }, file) => {
//                               ^ inflight interface to your preflight queue
  await myQueue.push(file);
});
```


docs/docs/07-examples/10-using-javascript.md
```diff
@@ -4,7 +4,73 @@ id: using-javascript
...
+// main.w
+bring cloud;
+let bucket = new cloud.Bucket();
+
+bucket.onCreate(inflight (file) => {
+  log(file);
+});
+```
+
+Being able to write inflight wing alongside the preflight code is beautiful, but you may want to write the inflight function in a separate file and *language*. The special `importInflight` function can be used to create an inflight closure from a JavaScript/TypeScript:
```
eladb, line 21: I am not a fan of this new name. What was the motivation to change from `@inflight()`?

MarkMcCulloh: Me neither but without the @ I couldn't just call it `inflight`, have any other ideas? I was asked to keep the syntax change out of the PR to possibly revisit later.

eladb: So let's go with `@inflight` (and follow up by fixing all the rest of the builtins)

docs/docs/07-examples/10-using-javascript.md
```diff
@@ -4,7 +4,73 @@ id: using-javascript
...
+// bucket_create.ts
+import inflight from "./bucket_create.inflight";
+
+export default inflight(async ({}, file) => {
+//                                 ^ This is known to be a string, the first positional argument needed for `onCreate`
+  console.log(file);
+});
+```
+
+`wing compile` will generate `bucket_create.inflight.ts` while will contain all the information needed to have full end-to-end type safety.
```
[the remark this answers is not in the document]
MarkMcCulloh: Do you have a suggestion for other wording? Also I don't necessarily say it provides that, I say it "will contain all the information needed" for it, which isn't direct. I could simply add somewhere that typescript type checking should be done separately.

> the error will only appear within your IDE assuming you're using VS Code with an ambient TypeScript language server running in the background.

AFAIK this is how most people work with typescript and they're not constantly running `tsc`, so automatic type checking may not even be expected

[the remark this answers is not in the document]
Chriscbr: I might just change this to "which will contain all of the information need for TypeScript type checking" for now

Chriscbr (approved):

examples/tests/valid/intrinsics.test.w
```diff
@@ -1,13 +1,39 @@
...
+    return "message";
+  }
+  pub inflight done() {
+    counter.inc();
+  }
+}
+let example = new Example();
+let queue = new cloud.Queue();
+queue.setConsumer(@inflight("./inline_typescript.ts", 
+  lifts: { example: { lift: example, ops: ["getMessage", "done"] } }
```
eladb, line 31:
```suggestion
  lifts: { example: { obj: example, ops: ["getMessage", "done"] } }
```

This is so verbose...

Can we make this work?

```js
lifts: {
  example: ["getMessage", "done"],
  my_bucket: ["put"],
}
```

Since this is an intrinsic, I believe we can technically make this work, no? It does limit the user's ability to map the wing symbol to the context key, but I am not sure this is actually needed.

examples/tests/valid/intrinsics.test.w
```diff
@@ -1,13 +1,39 @@
...
+pub class Example {
+  pub inflight getMessage(): str {
+    return "message";
+  }
+  pub inflight done() {
+    counter.inc();
+  }
+}
+let example = new Example();
+let queue = new cloud.Queue();
```
eladb, line 29: Maybe add an example of lifting a primitive/collection as well.

libs/wingc/src/ast.rs
```diff
@@ -557,58 +557,59 @@ pub struct Intrinsic {
...
 impl Documented for IntrinsicKind {
 	fn render_docs(&self) -> String {
 		match self {
+			IntrinsicKind::Unknown => "".to_string(),
 			IntrinsicKind::Dirname => r#"Get the normalized absolute path of the current source file's directory.
 
 The resolved path represents a path during preflight only and is not guaranteed to be valid while inflight.
 It should primarily be used in preflight or in inflights that are guaranteed to be executed in the same filesystem where preflight executed."#.to_string(),
-			IntrinsicKind::Unknown => "".to_string(),
+			IntrinsicKind::Inflight => r#"TODO"#.to_string(),
```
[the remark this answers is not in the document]
MarkMcCulloh: This PR is in draft, I still have tests and docs to add.

examples/tests/valid/intrinsics.test.w
```diff
@@ -1,13 +1,39 @@
...
 let currentFile = fs.join(@dirname, filename);
 expect.equal(filename, fs.basename(currentFile));
-
 expect.equal(@dirname, fs.dirname(currentFile));
-
 expect.equal(bar.Bar.getSubdir(), fs.join(@dirname, "subdir"));
+
+// @inflight
+
+let counter = new cloud.Counter();
```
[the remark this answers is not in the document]
MarkMcCulloh: Even when there's no change to the types? I specifically wanted to make sure the file isn't updated in that case so I'll look into it.

libs/wingc/src/jsify.rs
```diff
@@ -693,7 +695,90 @@ impl<'a> JSifier<'a> {
...
+						extern_path
+					} else {
+						return new_code!(&expression.span, "");
+					};
+
+					let mut export_name = "default";
+
+					let mut lifts: IndexMap<String, (&Expr, Option<&Vec<Expr>>)> = IndexMap::new();
+					for x in &arg_list.named_args {
+						if x.0.name == "export" {
```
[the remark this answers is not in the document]
MarkMcCulloh: Yep, still in draft

libs/wingc/src/jsify.rs
```diff
@@ -693,7 +695,90 @@ impl<'a> JSifier<'a> {
...
+							if let ExprKind::Literal(Literal::String(s)) = &x.1.kind {
+								export_name = &s[1..s.len() - 1];
+							}
+						} else if x.0.name == "lifts" {
```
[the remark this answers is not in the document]
MarkMcCulloh: There is a std type and it does go through normal type checking.
This code here is not parsing it's just traversing the AST node.

The reason it's like this is because the lifts need to be statically known (to know which types to grab)

libs/wingc/src/jsify.rs
```diff
@@ -693,7 +695,90 @@ impl<'a> JSifier<'a> {
...
+							if let ExprKind::JsonLiteral { element, .. } = &x.1.kind {
+								if let ExprKind::JsonMapLiteral { fields } = &element.kind {
+									for field in fields {
+										let alias = field.0.name.clone();
+										if let ExprKind::JsonLiteral { element, .. } = &field.1.kind {
+											if let ExprKind::JsonMapLiteral { fields } = &element.kind {
```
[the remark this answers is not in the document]
MarkMcCulloh: I can probably get this specific case to work when the object is pointing to a Json literal, but it's worth noting that lifts still need to be statically known.

examples/tests/valid/intrinsics.test.w
MarkMcCulloh: I only did it this way because you mentioned wanting to map the symbol to a new context key, so I'm happy to change it to something more terse.
The bigger downside to me is we lose the ability to lift arbitrary preflight expressions (since it has to be a symbol). The workaround would be to use a `let` binding to get a symbol, which isn't horrible so I'll go with that.

eladb: cool beans!

(i always reserve the right to change my mind)

libs/wingc/src/jsify.rs
```diff
@@ -693,7 +695,90 @@ impl<'a> JSifier<'a> {
...
+				IntrinsicKind::Inflight => {
+					let arg_list = intrinsic.arg_list.as_ref().unwrap();
+					let path = if let ExprKind::Literal(Literal::String(ss)) = &arg_list.pos_args[0].kind {
+						// trim off the first and last character (the quotes)
+						let ss = &ss[1..ss.len() - 1];
+						let extern_path = Utf8Path::new(&ss);
+
+						let extern_path = normalize_path(extern_path, ctx.source_path);
+						if !extern_path.exists() {
+							todo!()
```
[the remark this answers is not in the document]
MarkMcCulloh: Yeah I haven't really decided on this so this PR has the lowest-effort version.
A auto-fixable error from the lsp makes sense as a follow-up to this PR.

examples/tests/valid/intrinsics.test.w
MarkMcCulloh: I made the syntax slightly less verbose but I struggled to implement the ideal one you proposed. A big problem is the DX for lifting primitives or not specifying ops didn't make sense because this isn't valid.

```wing
lifts: {
  example,
}
```

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
MarkMcCulloh: I'm starting with a simple compromise here: There is no longer an error if the file does not exist. Will create follow up issue for DX improvements.

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
MarkMcCulloh: After some attempts it looks like even getting this case to work would take some major changes so I'll leave it to a future improvement.
Honestly I'd rather skip to figuring out a way to allow dynamic lifts.

libs/wingc/src/dtsify/extern_dtsify.rs
```diff
@@ -42,44 +41,91 @@ pub fn is_extern_file(file: &Utf8PathBuf) -> bool {
...
-	) -> Self {
+	pub fn new(types: &'a Types) -> Self {
 		Self {
-			libraries,
-			extern_file,
-			extern_file_env,
+			types,
 			known_types: HashMap::new(),
 			name_counter: HashMap::new(),
 			hoisted_types: CodeMaker::default(),
```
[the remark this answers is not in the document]
MarkMcCulloh: Decided to leave it out of this one

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
eladb: What's dynamic lifts?

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
eladb: I am fine with this

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
MarkMcCulloh: not requiring statically-known arguments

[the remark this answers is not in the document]
eladb: Can you elaborate? What do you mean by statically known arguments?

examples/tests/valid/intrinsics.test.w
eladb: I see (we discussed supporting such syntax in the future, like in JavaScript, right?).

So the new syntax is:

```js
lifts: {
  example: { obj: example }
}
```

Correct?

MarkMcCulloh: No I wanted to still optimize the common case, so I went with an array instead:

```lifts
lifts: [
  { obj: example }
]
```

This will automatically use "example" as the alias. You can provide a different alias and ops in that object

eladb: I see. OK. Not loving it, but let's get this moving.

MarkMcCulloh: would you prefer

```wing
lifts: {
  example: { obj: example }
}
```

?

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
MarkMcCulloh: Kinda contrived example:

```wing
class SomeStuff {
  new(path: str, lifted: std.LiftDeclaration) {
    @inflight(path, lift: [lifted])
  }
}
```

When I say static I mean "while generating js" or "pre-preflight". During this time is when we have all the type information available to generate .d.ts files.

In the above example, the compiler doesn't know value `path` has so it can't immediately write the type info to disk.
In addition, inside `lifted` is the type of the object being lifted, the alias it will have, and the operations needed. All of these things are needed to generate the types but they will not be available to the compiler.

imo the solution is we need a preflight API for type information and then all this generation can happen during preflight.

examples/tests/valid/intrinsics.test.w
eladb: That's basically the original proposal (with `lift` => `obj`). Correct?

Would it make sense to use @yoav-steinberg's syntax?

```js
lifts: [
  { b: [put, list] }, // <-- special syntax
]
```

MarkMcCulloh: Also I was hoping to avoid giving intrinsics access to custom syntax, I'd much rather be able to compose existing language features in them

eladb: Yes, I like the idea of coming up with a language syntax to describing a lift specification (e.g. `myBucket^[ put, list ]`) which resolves to an `std.Lift` expression and can be used anywhere that takes `std.Lift`. Then, `@inflight` just takes in `Array<std.Lift>`.

But I am realizing this is not for this PR.

So maybe it makes more sense not to create yet another special syntax and just use your original proposal (I prefer the term `obj` over `lift`).

eladb:
```
^myBucket.[put,list]
^myBucket.put
^myBucket.{put|list}
```

MarkMcCulloh: What's wrong with the syntax currently used in this PR?

```wing
lifts: [
  { obj: example }
]
```

It's less verbose than the original and the syntax is not special. The way it's interpreted is special, but it's not causing any limitations for the implementation.

eladb: I can't do this, right?

```js
lifts: [
  { obj: x.y.z.someClosure() }
]
```

Or even this:

```js
lifts: [
  { obj: myBucket.arn }
]
```

Feels like it's not fully baked...

MarkMcCulloh: You can, you will just have to provide an alias (the compiler will give an error)

```wing
lifts: [
  { obj: myBucket.arn, alias: "bucketArn" }
]
```

I'll admit that I'm hoping to eventually replace this with something closer to what you were mentioning, but in the mean time this doesn't feel too bad

eladb (approved):

docs/docs/07-examples/10-using-javascript.md
```diff
@@ -4,7 +4,78 @@ id: using-javascript
...
+//                        ^ Relative (to current file) path to javascript or typescript file
+//                          Note: This must be a static string
+```
+
+`wing compile` will generate `bucket_create.inflight.ts` while will which will contain all of the information needed for TypeScript type checking and IDE support.
+With that, you can create the `bucket_create.ts` file:
+
+```ts
+// bucket_create.ts
+import inflight from "./bucket_create.inflight";
```
ekeren, line 39: @MarkMcCulloh , there is something counter intuitive for me in this code:

```ts
// bucket_create.ts
import inflight from "./bucket_create.inflight";

export default inflight(async ({ myQueue }, file) => {
```


Can you explain why do we need an inflight to wrap the actual function?

MarkMcCulloh: The primary reason is a DX improvement, there is currently no *functional* purpose for this since the function just returns the argument.

The dilemma here is that typescript (and javascript) don't provide an easy mechanism for "Given type X, make sure this file exports it".

I think the current extern mechanism suffers from this, so you have to write code like this:

```ts
import extern from "./blah.extern";

export const func: extern["func"] = (name) => {
  return `Hello, ${name}!`;
};
```

I think the `inflight()` function helps guide user behavior, although the function type is also exported so user's can choose to use the explicit typing approach as well.

examples/tests/valid/intrinsics.test.w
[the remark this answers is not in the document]
MarkMcCulloh: I fixed the issue with the infinite reload loop but there's still an issue where the initial creation of the inflight.ts file, so I'm going to fast-follow this PR with more improvements to that

PR #6311, opened 2024-04-24
fix(console): playwright install fails

polamoros:
Running `playwright install --with-deps` fails because it executes `apt update` in the background, which is failing on the `latest ubuntu-22.04` image release.

Related to [Issue #9733](https://github.com/actions/runner-images/issues/9733).

Thanks to the temporary fix provided [here](https://github.com/actions/runner-images/issues/9733#issuecomment-2074655660) 🙌.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,7 +129,8 @@ jobs:
 
       - name: Install Playwright
         working-directory: apps/wing-console/console/app
-        run: pnpm exec playwright install --with-deps
+        run: sudo rm /etc/apt/sources.list.d/microsoft-prod.list
+          pnpm exec playwright install --with-deps
 
       - name: Test
         run: pnpm test:ci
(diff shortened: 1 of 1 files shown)


tsuf239 (approved):

.github/workflows/build.yml
```diff
@@ -129,7 +129,8 @@ jobs:
 
       - name: Install Playwright
         working-directory: apps/wing-console/console/app
-        run: pnpm exec playwright install --with-deps
+        run: sudo rm /etc/apt/sources.list.d/microsoft-prod.list
+          pnpm exec playwright install --with-deps
```
staycoolcall911, line 133: Looks like there's still a problem with the `--with-deps` attribute:
https://github.com/winglang/wing/actions/runs/8815708146/job/24198200226?pr=6311

tsuf239:
seems to fail:
image

staycoolcall911 (approved):

PR #6320, opened 2024-04-24
fix(sdk): wing test fails if unable to clean up files

Chriscbr:
If a container mounts a volume to the state directory as described in https://github.com/winglang/wing/pull/6295, it can sometimes result in permission errors when the CLI is cleaning up files after `wing test` because the mounted directory may have different permissions.

```
EACCES: permission denied, scandir 'target/test/mount.test.wsim/.state/c892787d9270cdccc06c34c95f7f9738be92b19134'
```

This error only seems to happen on CI for me. I'm not sure of the cause but I'm guessing that in CI, docker runs in a different (rootless maybe?) mode than on macOS.

Unblocks https://github.com/winglang/winglibs/pull/207

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing/src/commands/test/test.ts
+++ b/apps/wing/src/commands/test/test.ts
@@ -524,7 +524,7 @@ async function testSimulator(synthDir: string, options: TestOptions) {
     try {
       rmSync(synthDir, { recursive: true, force: true });
     } catch (err) {
-      console.error(`Failed to clean up ${synthDir}: ${err} ${(err as any).stack}`);
+      console.warn(`Failed to clean up test directory: ${err}`);
     }
   } else {
     noCleanUp(synthDir);
--- a/apps/wing/src/commands/test/test.ts
+++ b/apps/wing/src/commands/test/test.ts
@@ -524,7 +524,7 @@ async function testSimulator(synthDir: string, options: TestOptions) {
     try {
       rmSync(synthDir, { recursive: true, force: true });
     } catch (err) {
-      console.warn(`Failed to clean up test directory: ${err}`);
+      console.warn(`Warning: unable to clean up test directory: ${err}`);
     }
   } else {
     noCleanUp(synthDir);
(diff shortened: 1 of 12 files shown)


Chriscbr:
@MarkMcCulloh @eladb I wasn't sure what the best way to handle this was to be honest (docker and unix permissions inexperience). My first thought was "if my local Wing tests pass, but the test runner couldn't clean everything up, maybe we should just make it a warning" - that was the easiest solution to implement, so I did that here.

But there are some other options, I just wasn't sure which would be worth the investment at the moment:
- Keep raising an error if cleanup fails, but we'll try harder to clean up the files. If the `rmSync` call fails, the code could try to recursively change the permissions on all files within the output directory and then try deleting everything again.
- Change `sim.Container` so it doesn't create weirdly permissioned files by parsing the `volumes` array given to its constructor, looking for local file paths, and try to change the permissions after the bind mount is created.
- Try to change the approach from https://github.com/winglang/wing/pull/6295 so instead of using bind mounts (mounting the container's file system directly to a place in your file system), we use named mounts. These are created with `docker volume create pgdata` and deleted with `docker volume create pgdata` and are used with a command like `docker run -v pgdata:/var/lib/postgresql/data postgres:15`. I'm not sure how to redesign `sim.Container` API around this though.

(These are all just ideas, I'm not sure which would work in practice)

apps/wing/src/commands/test/test.ts
```diff
@@ -521,7 +521,11 @@ async function testSimulator(synthDir: string, options: TestOptions) {
   }
 
   if (clean) {
-    rmSync(synthDir, { recursive: true, force: true });
+    try {
+      rmSync(synthDir, { recursive: true, force: true });
+    } catch (err) {
+      console.warn(`Failed to clean up test directory: ${err}`);
```
MarkMcCulloh, line 527: `warn` is just an alias to `log`, so it may be worth adding extra text and/or using chalk to add some color.

```suggestion
      console.warn(`[WARNING] Failed to clean up test directory: ${err}`);
```

MarkMcCulloh (approved):
I like the idea of using named volumes. It's a nice separation of something stateful and as a resource it would conveniently have its own lifecycle.
Not exactly sure how the APIs will look but it's the only solution that comes to mind that isn't hacky.

_Side note: Volumes even seem like a worthy abstraction for the `cloud` namespace_

Issue #6284, opened 2024-04-21
Postgres winglib does not retain state in simulator after reloading


staycoolcall911:
### Use Case

Postgres db is currently losing its state when the simulator is reloaded. 
The use case here is to use postgres populated with some test data and avoiding having to re-populate the data frequently.

### Proposed Solution

The [Postgres Winglib](https://github.com/winglang/winglibs/tree/main/postgres) `sim` implementation should save state to a local dir, similar to what was done for `cloud.Bucket`s in #5164

### Implementation Notes

See #5164 which added this capability to `cloud.Bucket`s

### Component

Libraries

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Chriscbr:
I tested this locally - the state is retained as long as `WING_ENABLE_INPLACE_UPDATES=true` is enabled in your shell.

We have an issue for removing this feature flag here https://github.com/winglang/wing/issues/6185, so I'll mark this resolved - @staycoolcall911 feel free to reopen if needed

ekeren:
I expect to have it serialized between wing ribs and not just in
hotreloading

ekeren:
I mean persistent between two new runs

ekeren:
You'll notice that I am using this flag in package.json

Chriscbr:
Ah, I misunderstood. Thanks @ekeren, makes sense 👍 

I'll look into this, it feels like there should be some way to do this with the `volumes` option in `sim.Container`

PR #6336, opened 2024-04-25
feat(sdk): reuse volumes for sim.Container

Chriscbr:
Another pass at solving #6284

The approach taken in https://github.com/winglang/wing/pull/6295 where we allow users to bind to the state directory is problematic because docker sometimes changes the permissions of directories on the user's file system after mounting them. This was causing `wing test` to fail (fixed in https://github.com/winglang/wing/pull/6320) and `wing pack` to fail as well.

This PR reworks the API so that data can be persisted via named volumes. Specifically, an anonymous docker volume can be specified, and Wing will automatically name it and reuse it across Wing Console runs. The change should fix the previous issues since named volumes are managed by docker.

The main downside of this approach is that deleting your `target` directory or the state files in `.state/` will not clean everything up all resources/files created by Wing from your machine, as Docker volumes may still exist. Eventually we can provide some facilities for automatically cleaning these, but we're leaving it intentionally out of scope until this general approach has been validated.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/examples/tests/sdk_tests/container/mount.test.w
+++ b/examples/tests/sdk_tests/container/mount.test.w
@@ -6,8 +6,6 @@ skipPlatforms:
 bring sim;
 bring util;
 
-// This test was added to check that "wing test" still works when sim.Container is mounted to the state directory
-
 // only relevant in simulator
 if util.env("WING_TARGET") == "sim" {
   let container = new sim.Container(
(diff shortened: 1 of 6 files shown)


eladb (approved):
This is very cool. Like this solution better than the magic env var.

PR #6345, opened 2024-04-26
feat(platforms): `tf-aws` implicit platform now has 2 private subnets

hasanaburayyan:
Several resources require multi-az support, such as RDS clusters. This PR adds an additional private subnet to the VPC that is created implicitly for users using tf-aws platforms.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-tf-aws/app.ts
+++ b/libs/wingsdk/src/target-tf-aws/app.ts
@@ -253,6 +253,15 @@ export class App extends CdktfApp {
       },
     });
 
+    const privateSubnet2 = new Subnet(this, "PrivateSubnet2", {
+      vpcId: this._vpc.id,
+      cidrBlock: "10.0.8.0/22", // 10.0.8.0 - 10.0.11.255
+      availabilityZone: `${this.region}b`,
+      tags: {
+        Name: `${identifier}-private-subnet-2`,
+      },
+    });
+
     // Create the internet gateway
     const internetGateway = new InternetGateway(this, "InternetGateway", {
       vpcId: this._vpc.id,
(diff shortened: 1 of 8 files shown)


libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -253,6 +253,15 @@ export class App extends CdktfApp {
...
     });
 
+    const privateSubnet2 = new Subnet(this, "PrivateSubnet2", {
+      vpcId: this._vpc.id,
+      cidrBlock: "10.0.8.0/22", // 10.0.8.0 - 10.0.11.255
+      availabilityZone: `${this.region}b`,
+      tags: {
+        Name: `${identifier}-private-subnet-2`,
+      },
+    });
```
hasanaburayyan, line 263: I could have used a loop for creating private subnets, but I like the verbose decorativeness for VPCs

eladb: Can you share why verbosity (and code duplication) is valuable?

Also feels like we will need some platform parameter to control the number of subnets, which will just make this even more cumbersome, no?

libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -300,6 +309,20 @@ export class App extends CdktfApp {
       },
     });
 
+    const privateRouteTable2 = new RouteTable(this, "PrivateRouteTable2", {
```
eladb, line 312: Please add some documentation under https://www.winglang.io/docs/platforms/tf-aws

eladb (changes requested):

libs/wingsdk/src/target-tf-aws/app.ts
hasanaburayyan: > Can you share why verbosity (and code duplication) is valuable?

Just when it comes to planning out a network, theres already so many complexities involved like ensuring there are not any overlapping CIDR ranges, multi-regency, and route tables. For instance there might arise a use case where we want some private subnets with no egress, but still ensure they have connectivity to other private subnets that do have egress and it feels like when this is done imperatively that its just harder to follow.

> Also feels like we will need some platform parameter to control the number of subnets, which will just make this even more cumbersome, no?

Yea it becomes more cumbersome, but I dont know if we should allow (yet) the ability to tweak this implicit VPC yet. If someone has specific VPC requirements my recommendation for now is create your own VPC and specify it in the `wing.toml` theres such a wide range of knobs to tweak with VPCs I think we need to look into making it a resources first and plan out the API a little more.

eladb (approved):

Issue #6346, opened 2024-04-26
`.env` Files Not Loaded When Compiling Outside Of WING_SOURCE_DIR


hasanaburayyan:
### I tried this:

main.w file:
```js
bring util;

log(util.env("MY_VALUE"));
```

.env file
```
MY_VALUE="abc123"
```

Compiled from one directory up `wing compile wing_app/main.w`

### This happened:

```bash
Error: Environment variable MY_VALUE not found.
  --> junk/main.w:3:1
  | bring util;
  | 
3 | log(util.env("MY_VALUE"));
  | ^
```

### I expected this:

To see logged "abc123"

### Is there a workaround?

Run compile from source directory

### Anything else?

Fundamentally the issue is that we load environment variables from the cli within a method called `runSubCommand` https://github.com/winglang/wing/blob/main/apps/wing/src/cli.ts#L24

This method calls `loadEnvVariables` as such:
```js
loadEnvVariables({
    mode: subCommand,
  });
```

The method `loadEnvVariables` can take an optional path to load variables from, but if not given one it will default to `process.cwd()` See code here: https://github.com/winglang/wing/blob/main/apps/wing/src/env.ts#L25

The challenge here is we dont really have a great way to grab the `entrypoint` file to determine the baseDir from within the `runSubCommand` 

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.73.0](https://github.com/winglang/wing/releases/tag/v0.73.0).

PR #6348, opened 2024-04-26
fix(cli)!: .env file not loaded when compiling outside sourceDir

hasanaburayyan:
Closes: https://github.com/winglang/wing/issues/6346

## Breaking Change
- Removed the ability to specify command specific `.env` files

## Summary
Basically just moves the loading of environment files to the subcommands. Previously this loading was happening for all commands, which really doesnt make sense since some commands like `lsp`, `docs`, `init` dont need to load those env files. 

Also I several subcommands call the compile subcommand (secrets, test, etc..) so having subcommand specific `.env` files is additional overhead that Im not sure what the usecase it solves is. If needed, its probably just easier to just use a `--env-file` flag

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-sim/secret.inflight.ts
+++ b/libs/wingsdk/src/target-sim/secret.inflight.ts
@@ -1,5 +1,3 @@
-import * as fs from "fs";
-import * as path from "path";
 import { SecretAttributes, SecretSchema } from "./schema-resources";
 import { ISecretClient, SECRET_FQN } from "../cloud";
 import {
@@ -11,18 +9,9 @@ import { Json, TraceType } from "../std";
 
 export class Secret implements ISecretClient, ISimulatorResourceInstance {
   private _context: ISimulatorContext | undefined;
-  private readonly secretsFile: string;
   private readonly name: string;
 
   constructor(props: SecretSchema) {
-    this.secretsFile = path.join(process.cwd(), ".env");
-
-    if (!fs.existsSync(this.secretsFile)) {
-      throw new Error(
-        `No secrets file found at ${this.secretsFile} while looking for secret ${props.name}`
-      );
-    }
-
     this.name = props.name;
   }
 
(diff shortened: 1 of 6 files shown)


libs/wingsdk/src/target-sim/secret.inflight.ts
```diff
@@ -11,18 +9,9 @@ import { Json, TraceType } from "../std";
 
 export class Secret implements ISecretClient, ISimulatorResourceInstance {
   private _context: ISimulatorContext | undefined;
-  private readonly secretsFile: string;
   private readonly name: string;
 
   constructor(props: SecretSchema) {
```
Chriscbr, line 14: Is it possible to have the secret values passed through `simulator.json` instead? I feel like the simulator's interface will be more robust if we can keep as many of the "simulation inputs" within the `target/app.wsim` folder that it gets passed as possible. Recall that to run the simulator, all you've needed to do is run this code more or less:

```js
let s = new sdk.Simulator({ simdir: "target/app.wsim" });
await s.start();
```

Chriscbr (changes requested):

libs/wingsdk/src/target-sim/secret.inflight.ts
Chriscbr: https://github.com/winglang/wing/issues/6349

hasanaburayyan: > Is it possible to have the secret values passed through simulator.json instead?

I imagine its totally possible, but I think that it makes sense to suggest it in a different PR for 2 reasons. 

First at its heart this PR still needs to go in to fix the issue of [`.env` files not being loaded](https://github.com/winglang/wing/issues/6346) and thats really what this fixes. The code deleted in this file was just dead (which is why I removed it, we dont actually use that file to read values in the `values()` method

As well, I think it might require some discussion, as there may be complexity there still around how we decide what to write into the `simulator.json` I.E. the DX right now is that users define their secrets for the simulator in the `.env` files, and they cant really just place these values i the generated output dirs.

hasanaburayyan: Yea Im gonna start working on a quick PR for: https://github.com/winglang/wing/issues/6349 and maybe get a discussion going around the implementation.

Chriscbr (approved):

PR #6350, opened 2024-04-26
chore: remove _supportedOps

### LINKED ISSUE #6353: Lifts are not recursive in TypeScript
### I tried this:

```ts
import { cloud, lift, main } from "@wingcloud/framework";

main((app, test) => {
  const myBucket = new cloud.Bucket(app, "Bucket");

  const putter = lift({ myBucket }).grant({ myBucket: ["put"] }).inflight(async (ctx) => {
    await ctx.myBucket.put("hello", "world");
  });

  test("invoke putter", lift({ putter }).inflight(async (ctx) => {
    await ctx.putter();
    return undefined;
  }));
});
```

### This happened:

```sh
$ wing test
[ERROR] (no test) | Invoke (payload=undefined). Error: Missing environment variable: BUCKET_HANDLE_a107cec5
fail ┌ main.ts.wsim » root/env0/invoke putter
     └ Error: Missing environment variable: BUCKET_HANDLE_a107cec5

Tests 1 failed (1)
Snapshots 1 skipped
Test Files 1 failed (1)
Duration 0m1.73s
```

### I expected this:

Since my test inflight closure lifts `putter`, and putter lifts `myBucket`, I expected this to work.

### Is there a workaround?

If I explicitly lift `myBucket` in the test inflight closure, it works.

But this is not practical because this issue can occur in a case where you don't have access to state that's being lifted by the inner call.

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

Chriscbr:
Fixes #6353

The information provided by this method is basically a lighter form of `_liftMap` and the duplication introduces some confusion while iterating on lifting code, so this finishes migrating some of the code over. Most of this is boilerplate that the compiler generates for you under ordinary circumstances.

Misc:
* Added some checks to `wing-api-checker` to ensure `@inflight` annotations are added to SDK resources written in TypeScript.

## Checklist

- [ ] Tests added (always)


libs/wingc/src/dtsify/mod.rs
```diff
@@ -207,9 +207,6 @@ impl<'a> DTSifier<'a> {
 			code.line(format!(
 				"[{TYPE_INTERNAL_NAMESPACE}.INFLIGHT_SYMBOL]?: {inflight_class_name};"
 			));
-			code.line(format!(
-				"_supportedOps(): {TYPE_INTERNAL_NAMESPACE}.OperationsOf<{inflight_class_name}>;"
-			));
```
yoav-steinberg, line 212: Are we on purpose not adding a `_liftMap` here?

libs/wingsdk/src/core/lifting.ts
```diff
@@ -258,23 +258,13 @@ export function collectLifts(
...
       matrixCache.set(obj, matrix);
-    } else if (
-      typeof obj === "object" &&
-      typeof obj._supportedOps === "function"
-    ) {
-      matrix = {};
-      for (const op of obj._supportedOps()) {
-        matrix[op] = new Map();
-      }
-      matrixCache.set(obj, matrix);
```
yoav-steinberg, line 277: :clap:

yoav-steinberg (approved):

libs/@wingcloud/framework/src/main.ts
```diff
@@ -76,6 +77,9 @@ This is a Wing app and must be run through the Wing CLI (npm install -f winglang
         new std.Test(this, name, handler, props);
       });
     }
+    get _liftMap(): LiftDepsMatrixRaw {
```
eladb, line 80: Would it make sense to make `_liftMap` optional and default to `{}`?

There were a bunch of times (can't recall the exact scenario) where I ran into an issue because this property was not implemented.

libs/wingsdk/src/std/resource.ts
```diff
@@ -167,12 +158,8 @@ export abstract class Resource extends Construct implements IResource {
...
   }
 
-  /**
-   * @internal
-   * */
-  public _supportedOps(): string[] {
-    return [];
-  }
+  /** @internal */
+  public abstract get _liftMap(): LiftDepsMatrixRaw;
```
eladb, line 162: Why does this have to be a getter?

```suggestion
  public readonly _liftMap?: LiftDepsMatrixRaw;
```

Maybe like this?

And then if not defined it defaults to `{}`... more robust

libs/wingsdk/src/target-sim/policy.ts
```diff
@@ -20,10 +21,13 @@ export interface PolicyProps {
 
 /**
  * Implementation of `sim.Policy`.
+ *
+ * @inflight `@winglang/sdk.sim.IPolicyClient`
```
eladb, line 25: Missing inflight type marker ?


(Is there a way to enforce this because we have a bunch of places where this is missing)

libs/wingsdk/src/target-sim/table.ts
```diff
@@ -2,6 +2,7 @@ import { Construct } from "constructs";
 import { ISimulatorResource } from "./resource";
 import { TableSchema } from "./schema-resources";
 import { bindSimulatorResource, makeSimulatorJsClient } from "./util";
+import { LiftDepsMatrixRaw } from "../core";
```
eladb, line 5: I would rename this type to `LiftMap`

eladb (approved):

eladb:
This is great and it will unblock #6353

libs/wingsdk/src/target-sim/table.ts
Chriscbr: Done

libs/wingsdk/src/target-sim/policy.ts
Chriscbr: Added some extra rules to the `wing-api-checker`.

libs/wingc/src/dtsify/mod.rs
Chriscbr: I wasn't sure. The tests are still passing after I removed it so maybe it isn't needed?

libs/wingsdk/src/std/resource.ts
Chriscbr: Sure, I'm down to try this

Issue #2637, opened 2023-05-23
Errors are being displayed in a different order from the line that generated the error in the code


marciocadev:
### Feature Spec

Some errors are being generated out of order compared to the generating code.

```bash
error: \\"y\\" cannot be initialized in the Inflight initializer
   --> ../../../examples/tests/invalid/class.w:45:10
   |
45 |     this.y = 1;
   |          ^ \\"y\\" cannot be initialized in the Inflight initializer

error: \\"x\\" cannot be initialized in the Preflight initializer
   --> ../../../examples/tests/invalid/class.w:40:10
   |
40 |     this.x = 1;
   |          ^ \\"x\\" cannot be initialized in the Preflight initializer
```

### Use Cases

It would be interesting to display the errors following the sequence of lines from the original code.

### Component

Compiler

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

yoav-steinberg:
We should probably sort by line number here https://github.com/winglang/wing/blob/bb52687ed23bf4a61b6752426a53d2d7359af244/apps/wing/src/commands/compile.ts#L98

PR #6351, opened 2024-04-27
fix(compiler): no error when missing `super()` call in constructor

### LINKED ISSUE #6332: super with argument should be required
### I tried this:

```wing
class RequiredField {
  pub name: str;
  new(name: str){
    this.name = name;
  }
}

class BreakRequirement extends RequiredField {

}

let breaks = new BreakRequirement();

test "main" {
  log(breaks.name);
}

```

### This happened:

The code compiled and printed `undefined`

### I expected this:

to fail to compile because `BreakRequirement` doesn't call `super(str)`

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

0.72.7

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

yoav-steinberg:
Fixes #6332
Note that this also adds a test for the implicit call to `super()` when no args are needed even if the child class has a ctor with no `super()` call in it, unlike typescript where you must call `super()` from a child ctor even if there are no args.

## Checklist

- [x] Tests added (always)


examples/tests/valid/class.test.w
Chriscbr, line 1: Can we add a test checking that it's possible to have statements before the super() statement (if we don't have one already)?

```js
class A {
  new(x: num) {}
}

class B extends A {
  new() {
    let x = 3 + 5;
    super(x);
  }
}
```

If we strictly enforce `super()` has to be the first statement, it can make some code a lot harder to express, at least from my experience working in TypeScript. Of course not all statements can be put before the `super()` statement - for instance, using `this` before `super()` isn't allowed. But it's ok if we add some kind of compiler check for that in a later PR.

Chriscbr (approved):
Good stuff!

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -1266,6 +1278,91 @@ error: Member \\"copyMut\\" doesn't exist in \\"MutSet\\"
...
+   |
+45 |   new() {}
+   |         ^^ Missing super() call as first statement of InflightChild4's constructor
+
+
+error: Expected 1 positional argument(s) but got 0
+   --> ../../../examples/tests/invalid/ctor_super.test.w:19:5
+   |
+19 |     super();
+   |     ^^^^^^^^ Expected 1 positional argument(s) but got 0
```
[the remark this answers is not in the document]
yoav-steinberg: The `compile.rs` rust example sorts the diags before printing them.
The typescript code in the cli's here doesn't seem to do that, know issue #2637.

examples/tests/valid/class.test.w
yoav-steinberg: We already have this requirement in a few other places (https://github.com/winglang/wing/blob/bb52687ed23bf4a61b6752426a53d2d7359af244/libs/wingc/src/jsify.rs#L1653) for example.

I agree it'll be good to get rid of this, perhaps another issue?

PR #6354, opened 2024-04-27
chore(sdk): deprecate `makeHandler()` and `convertBetweenHandlers()`

eladb:
We can now use `lift()` and `inflight()` in TypeScript instead of `makeHandler()` and `convertBetweenHandlers()`.

This is a great exercise for our TypeScript inflight support (and it's awesome :-)).

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/shared-aws/queue.ts
```diff
@@ -121,46 +125,35 @@ export class QueueRef extends Resource {
...
-        } catch (e) {
-          return e.message;
-        }
-      }`,
-      {
-        queue: {
-          obj: this,
-          ops: [QUEUE_URL_METHOD],
-        },
+    const awsConsoleHandler = lift({ queue: this }).inflight(async (ctx) => {
```
eladb, line 140: For some reason if I try to add a `grant({ queue: ["queueUrl"] })` I get:

```
Type '"queueUrl"[]' is not assignable to type '{ [K in keyof Lifted<this>]: Lifted<this>[K] extends AsyncFunction ? K : never; }[keyof Lifted<this>][]'.
  Type 'string' is not assignable to type 'Lifted<this>[keyof Lifted<this>] extends AsyncFunction ? keyof Lifted<this> : never'.ts(2322)
```

libs/wingsdk/src/core/lifting.ts
```diff
@@ -219,12 +219,6 @@ export function collectLifts(
   initialObj: any,
   initialOps: Array<string>
 ): Map<any, Set<string>> {
-  if (initialOps.includes(INFLIGHT_INIT_METHOD_NAME)) {
-    throw new Error(
-      `The operation ${INFLIGHT_INIT_METHOD_NAME} is implicit and should not be requested explicitly.`
-    );
-  }
```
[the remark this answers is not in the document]
Chriscbr: I'd recommend either of these options, if it's not too much of a hassle to include in this PR:

1. Fix this line that I recently changed (the likely cause of the regression)

https://github.com/winglang/wing/blob/b6d1f04b7c7d9036a100bd63178ed6134f6525c0/libs/wingsdk/src/core/inflight.ts#L224-L226

so that we filter out the $inflight_init key when generating automatic permissions:

```
Object.keys(...).filter(x => x !== "$inflight_init")
```

This would let us keep the error message (can revert the changes to lifting.ts)

2. Or, just filter out "$inflight_init" from `ops` before we call onLift() or onLiftType() here:

https://github.com/winglang/wing/blob/b6d1f04b7c7d9036a100bd63178ed6134f6525c0/libs/wingsdk/src/core/lifting.ts#L399-L428

libs/wingsdk/src/shared-aws/queue.ts
eladb: @MarkMcCulloh did you see this?

libs/wingsdk/src/core/utility-types.ts
```diff
@@ -14,13 +14,23 @@ export type PickNonFunctions<T> = Pick<
 /**
  * Represents a map of liftable objects with string keys.
  */
-export type LiftableMap = { [key: string]: Liftable };
+export type LiftableHash = { [key: string]: Liftable };
```
MarkMcCulloh, line 17: Why were these renames done?

nit: imo `LiftableHash` is a slightly worse name. I would say `LiftableHashMap` or `LiftableRecord`

libs/wingsdk/src/target-tf-gcp/app.ts
```diff
@@ -31,7 +31,7 @@ export interface AppProps extends CdktfAppProps {
    * The Google Cloud zone, used for all resources.
    * @see https://cloud.google.com/functions/docs/locations
    */
-  readonly zone: string;
+  readonly zone?: string;
```
MarkMcCulloh, line 34: This change seems a little out of place, why was it added?

libs/wingsdk/src/shared-aws/api.ts
```diff
@@ -78,20 +79,19 @@ export class ApiEndpointHandler {
...
-        corsHeaders: cloud.Api.renderCorsHeaders(corsOptions)?.defaultResponse,
-      }
-    );
+    return lift({ handler }).inflight(async (ctx, request) => {
+      return apigwFunctionHandler(
+        request as unknown as APIGatewayProxyEvent,
+        ctx.handler,
+        headers
+      ) as unknown as string;
+    });
```
MarkMcCulloh, line 95: `apigwFunctionHandler` is from an upper scope so this will break at runtime.

It may be possible to solve this by getting the path to that util (need to import `path` as well)

```suggestion
    return lift({ handler, utilPath: join(__dirname, "api-util") }).inflight(async (ctx, request) => {
      return require(ctx.utilPath).apigwFunctionHandler(
        request as unknown as APIGatewayProxyEvent,
        ctx.handler,
        headers
      ) as unknown as string;
    });
```

MarkMcCulloh:
Some failing sdk tests https://github.com/winglang/wing/actions/runs/8895505535/job/24425986210

libs/wingsdk/src/target-tf-gcp/app.ts
eladb: It was needed because there is code that doesn't pass `zone` and it failed complication. Not sure exactly how our main build passed.

libs/wingsdk/src/core/utility-types.ts
eladb: OK, changed to `LiftableRecord`

libs/wingsdk/src/shared-aws/api.ts
eladb: OK!

MarkMcCulloh:
Still 3 failing SDK test suites in aws:
- [api](https://github.com/winglang/wing/actions/runs/8908847211/job/24465176266)
- [endpoint](https://github.com/winglang/wing/actions/runs/8908847211/job/24465177119#step:12:1745)
- [http](https://github.com/winglang/wing/actions/runs/8908847211/job/24465178400#step:12:1915)

MarkMcCulloh (approved):
SDK Tests passing https://github.com/winglang/wing/actions/runs/8910509568

Issue #6340, opened 2024-04-26
Allow specifying compile output directory


eladb:
### Use Case

We have a use case where we want `wing compile -t TARGET` to write it's output to a custom output directory.

Specifically, we have a situation where Wing is used with the `awscdk` and integrated into a CI/CD system that assumes that the synthesized output is in the standard `cdk.out` and their CI/CD system currently does not respect the `output` option in `cdk.json`.

A temporary workaround for them is to `mv target/main.awscdk cdk.out` but that's ugly.

Ideally, we just want their `cdk.json` to be:

```json
{
  "app": "wing compile -t @winglang/platform-awscdk -o cdk.out main.w"
}
```

### Proposed Solution

Add an `-o` or `--output` options to the Wing CLI:

```sh
wing compile -t my-platform -o dist
```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

tsuf239:
There is a `-o`/`--output-file` on the test command for the JSON test result, should I change the letter there? (or it doesn't matter since `compile` is a different command?)

eladb:
doesn't matter because it's a different command

monadabot:
Congrats! :rocket: This was released in [Wing 0.73.19](https://github.com/winglang/wing/releases/tag/v0.73.19).

PR #6359, opened 2024-04-28
feat(cli): `compile --output`

tsuf239:
## Checklist
fixes: https://github.com/winglang/wing/issues/6340

- [ ] Tests added (always)


apps/wing/src/commands/compile.ts
```diff
@@ -49,6 +49,11 @@ export interface CompileOptions {
    * @default "./target"
    */
   readonly targetDir?: string;
+  /**
+   * The overrides the location to save the compilation output
+   * @default "./target"
```
eladb, line 54: The default is `./target/<entrypoint>.<target>` -- we should make sure this overrides the entire path and not just `./target`.

What I mean is that this should behave like this:

```sh
$ wing compile -t tf-aws main.w -o my-output
$ cd my-output
$ terraform init
$ terraform apply
```

eladb (commented):
please change the PR title to match the issue title

apps/wing/src/commands/compile.ts
tsuf239: it's only the comment that was wrong :)

libs/wingcompiler/src/compile.test.ts
```diff
@@ -12,60 +12,72 @@ export async function generateTmpDir() {
...
+    expect(stats.isDirectory()).toBeTruthy();
+    expect(artifactDir).toContain(targetDir);
+    expect(basename(artifactDir)).toEqual("enums.test.wsim");
+  });
 
-      const stats = await stat(artifactDir);
-      expect(stats.isDirectory()).toBeTruthy();
-      expect(artifactDir).toContain(targetDir);
-      expect(basename(artifactDir)).toEqual("enums.test.wsim")
+  test("should be able to override the target directory", async () => {
```
tsuf239, line 70: this is the only new test (rest is formatting)

libs/wingcompiler/src/compile.ts
```diff
@@ -63,6 +63,9 @@ export interface CompileOptions {
 
   // target directory for the output files
   readonly targetDir?: string;
+
+  // overrides the target directory for the output files
+  readonly output?: string;
```
eladb, line 68:
```suggestion
  readonly outputOverride?: string;
```

Add description of the default behavior to docstrings

tsuf239: I think there is a value in keeping it aligned to the cli (we pass all the options with a spread operator) 
looked at the cli tool and there is no way to change the variable :(

Issue #6363, opened 2024-04-28
`math - tf-aws` test fails with `no space left on device`


garysassano:
### I tried this:

Test run [in a PR](https://github.com/winglang/wing/actions/runs/8866141232/job/24347552655).

### This happened:

```sh
│ Error: Failed to install provider
│ 
│ Error while installing hashicorp/aws v5.41.0: write
│ .terraform/providers/registry.terraform.io/hashicorp/aws/5.41.0/linux_amd64/terraform-provider-aws_v5.41.0_x5:
│ no space left on device
```

### Is there a workaround?

Upgrade the GitHub runner's size or [free up some space](https://github.com/actions/runner-images/issues/2875).

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

tsuf239:
Gh action run can be found [here](https://github.com/winglang/wing/actions/runs/8866141232/job/24343706916)
Another error it throws is: 
```
│ Error: Inconsistent dependency lock file
│ 
│ The following dependency selections recorded in the lock file are
│ inconsistent with the current configuration:
│   - provider registry.terraform.io/hashicorp/aws: required by this configuration but no version is selected
│ 
│ To make the initial dependency selections that will initialize the
│ dependency lock file, run:
│   terraform init
╵
```

garysassano:
@tsuf239 The second error is normal. It expects to find the `hashicorp/aws` provider in the lockfile, but since the installation failed in the first place, it wouldn't be present.

monadabot:
Congrats! :rocket: This was released in [Wing 0.73.45](https://github.com/winglang/wing/releases/tag/v0.73.45).

PR #6381, opened 2024-04-30
feat(cli): `-p/ --parallel` option defines tests amount to be run on parallel

tsuf239:
Adding -b/ --batch option defines tests amount to be run on parallel,
as well as renaming the sdk spec tests file 
potentially fixes https://github.com/winglang/wing/issues/6363

## Checklist

- [ ] Tests added (always)


apps/wing/src/cli.ts
```diff
@@ -227,6 +227,7 @@ async function main() {
         .preset(3)
         .argParser(parseInt)
     )
+    .addOption(new Option("-b, --batch [batch]", "Number of tests to be executed on parallel"))
```
eladb, line 230: Add more details to this description and also describe the default behavior.

eladb (commented):
Please update the PR title and description according to the guidelines and fix capitalization.

apps/wing/src/cli.ts
```diff
@@ -227,6 +227,12 @@ async function main() {
         .preset(3)
         .argParser(parseInt)
     )
+    .addOption(
+      new Option(
+        "-b, --batch [batch]",
+        "Number of tests to be executed on parallel- if not specified will execute all at the same time"
```
MarkMcCulloh, line 233: This description and the PR description don't quite match the implementation.

If you have 10 tests with `-b 2`. If the first test takes 2 minutes and rest take 1 second, you'll waste a lot of time on the first batch. Ideally, the first one would run while the rest run 1-at-a-time in parallel with the first.

If possible, I would rename this option to `parallelism` and maybe use https://www.npmjs.com/package/p-limit (or a custom implementation if it's simple enough)

MarkMcCulloh (changes requested):

apps/wing/src/commands/test/test.ts
```diff
@@ -134,7 +140,20 @@ export async function test(entrypoints: string[], options: TestOptions): Promise
...
+  const totalBatches = Math.ceil(selectedEntrypoints.length / batchSize);
+  await PromisePool.withConcurrency(batchSize)
+    .for(selectedEntrypoints)
+    .onTaskFinished((_, pool) => {
+      const processed = pool.processedCount();
+      if (processed % batchSize == 0 || processed === selectedEntrypoints.length) {
+        log(`Done processing batch ${Math.ceil(processed / batchSize)}/${totalBatches}`);
+      }
+    })
+    .process(testFile);
```
MarkMcCulloh, line 156: Now that we're not doing batches, I think we can simplify this. I don't think it's important for the user to get feedback about the parallelism since they hopefully won't notice it.

```suggestion
  await PromisePool
    .withConcurrency(options.parallel ?? selectedEntrypoints.length)
    .for(selectedEntrypoints)
    .process(testFile);
```

Also should be consider a sane default max parallelism? Especially for cloud testing currently it's not a good idea to run too many tests at once.

apps/wing/package.json
```diff
@@ -52,6 +53,7 @@
     "npm-packlist": "^8.0.0",
     "open": "^8.4.2",
     "ora": "^5.4.1",
+    "p-limit": "^5.0.0",
```
MarkMcCulloh, line 56:
```suggestion
```

apps/wing/src/commands/test/test.ts
tsuf239: ok :) 
I was doing it for being able to see/test if it's working at all - we won't be able to track it otherwise (and also won't be able to write a test, unless spying promisePool)

what would be a good default limit? 10? 20?

MarkMcCulloh: It should be possible to test by having the tests access something shared as a way to check a lock, like a file. Then setting the parallelism to 1.

Not sure what a good default is, We could try 10.

apps/wing/src/cli.ts
```diff
@@ -235,6 +235,12 @@ async function main() {
         .preset(3)
         .argParser(parseInt)
     )
+    .addOption(
+      new Option(
+        "-p, --parallel [batch]",
+        "Number of tests to be executed on parallel- if not specified will start executing all at the same time"
```
MarkMcCulloh, line 241:
```suggestion
        "-p, --parallel <count>",
        "Number of tests to execute in parallel. 0 to run all at once."
```

Also we could add the default value in here as well

apps/wing/src/commands/test/test.test.ts
```diff
@@ -358,6 +358,39 @@ describe("retry option", () => {
...
+    `
+    );
+
+    await wingTest(["t1.test.w", "t2.test.w"], {
+      clean: true,
+      platform: [BuiltinPlatform.SIM],
+      parallel: 1,
+    });
+    const batchLogs = logSpy.mock.calls.filter((args) => args[0].includes("Processing batch"));
+    expect(batchLogs.length).toBe(0);
```
MarkMcCulloh, line 392: The "Processing batch" message is not ever printed so this test doesn't really verify anything

tsuf239: yes, I fixed it just now, I had to go somewhere last evening, and I got to it only now

apps/wing/src/cli.ts
tsuf239: zero isn't acceptable in this package- I can add it to our code, or make it behave the same as undefined and run 10 at once, what do you think?

MarkMcCulloh: Since it's a common option i've seen elsewhere, I think having "0 = no limit" is useful.

apps/wing/src/cli.ts
```diff
@@ -3,6 +3,7 @@ import { satisfies } from "compare-versions";
 
 import { optionallyDisplayDisclaimer } from "./analytics/disclaimer";
 import { exportAnalytics } from "./analytics/export";
+import { DEFAULT_PARALLEL_SIZE } from "./commands";
```
MarkMcCulloh, line 6: This import is causing us to eagerly import all the commands (and their dependencies).

This constant could be moved to another file.

apps/wing/src/cli.ts
```diff
@@ -235,6 +236,12 @@ async function main() {
         .preset(3)
         .argParser(parseInt)
     )
+    .addOption(
+      new Option(
```
MarkMcCulloh, line 240: I think the default value can be provided as an argument to this constructor as well

apps/wing/src/commands/test/test.test.ts
```diff
@@ -358,6 +358,80 @@ describe("retry option", () => {
...
+  `
+    );
+
+    const startingTime = Date.now();
+    await wingTest(["t1.test.w", "t2.test.w"], {
+      clean: true,
+      platform: [BuiltinPlatform.SIM],
+      parallel: 2,
+    });
+    expect(Date.now() - startingTime).toBeLessThan(4 * 1000);
```
MarkMcCulloh, line 433: This could be flakey if the test is under heavy load (so it'll be super slow)

We can try it though.

apps/wing/src/commands/test/test.ts
```diff
@@ -134,7 +140,11 @@ export async function test(entrypoints: string[], options: TestOptions): Promise
       });
     }
   };
-  await Promise.all(selectedEntrypoints.map(testFile));
+
+  await PromisePool.withConcurrency(options.parallel || selectedEntrypoints.length)
```
MarkMcCulloh, line 144: nit

```suggestion
  await PromisePool.withConcurrency(options.parallel ?? selectedEntrypoints.length)
```

MarkMcCulloh (approved):
🙌

apps/wing/src/commands/test/test.ts
tsuf239: actually not- if options.parallel is 0 we want no limit...

PR #6386, opened 2024-04-30
fix(platforms): `tf-aws` storing secrets as key/val instead of plain

hasanaburayyan:
I dunno why I thought to save them as key/val 😁 

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/target-tf-aws/platform.ts
```diff
@@ -87,7 +87,7 @@ export class Platform implements IPlatform {
         await client.send(
           new UpdateSecretCommand({
             SecretId: name,
-            SecretString: JSON.stringify({ [name]: value }),
+            SecretString: value,
```
hasanaburayyan, line 90: Contemplating the best way to test this.

tsuf239: how was `JSON.stringify({ [name]: value })` tested?

eladb (approved):
Nice! Maybe link to the Wingly episode just for the kicks

PR #6389, opened 2024-04-30
feat(sdk): sim.Resource

Chriscbr:
Introduces `sim.Resource`, a primitive for creating stateful resources running in the Wing simulator. Check out the new documentation [here](https://github.com/winglang/wing/blob/rybickic/sim-resource2/docs/docs/04-standard-library/sim/resource.md) for usage examples and an API reference.

Misc:
* Fix E2E tests for cloud.Function timeouts
* Converted the simulator implementation of `cloud.Counter` to use `sim.Resource`. It will be a bit of a manual process to do this for the other cloud resources.
* Improved the robustness of the log/event streaming facilities in the CLI. There were occasions where an error stack trace might not get printed to STDOUT because the formatting work was happening asynchronously.

Follow-up:
- [ ] Convert other `cloud` resource implementations to use `sim.Resource` internally
- [ ] Deprecate `sim.State`
- [ ] Deprecate `ctx.log(message, level)` in favor of `log(message, level: level)`
- [ ] Enforce serializable types for inflight methods on preflight classes
  - [ ] We could also consider loosening type constraints on `sim.IResource` implementations to allow private methods and fields to use non-serializable types.
- [ ] Improve DX around serializing and deserializing types

## Checklist

- [x] Tests added (always)


examples/tests/sdk_tests/resource/resource.test.w
```diff
@@ -0,0 +1,77 @@
...
+inflight class CounterBackend impl sim.IResource {
+  var counter: num;
+
+  new() {
+    this.counter = 0;
+  }
+
+  pub onStart() {
+    // startup code
+  }
```
eladb, line 12: We need to make sure there's a way to create/resolve tokens:

```suggestion
  pub onStart(ctx) {
    ctx.resolveToken("boom", value);
  }
```

Maybe this is a nice API for resolving tokens?

Also:
- how can you access things like the state directory of this resource?
- how does the resource emit logs (please include levels!)

(saying - let's try to make sure this API supports everything one needs in order to implement resources like a king).

examples/tests/sdk_tests/resource/resource.test.w
```diff
@@ -0,0 +1,77 @@
+bring sim;
+
+inflight class CounterBackend impl sim.IResource {
```
eladb, line 3: `sim.IResource` is slightly confusing because it's overloaded with `std.IResource` which I *guess* means something else. Maybe not?!

I am okay leaving it like this actually.

libs/wingsdk/src/target-sim/resource.inflight.ts
```diff
@@ -0,0 +1,89 @@
...
+    // TODO: support other update plans
+    return UpdatePlan.REPLACE;
+  }
+
+  public async call(method: string, args?: Array<Json>): Promise<Json> {
+    return this.context.withTrace({
+      activity: async () => {
+        return this.sandbox!.call("call", method, ...(args ?? []));
+      },
+      message: `${method}(${args?.join(", ")})`,
```
eladb, line 76: Wouldn't this be `${method)(undefined)` if `args` is `undefined`? You probably wanted `(args ?? []).join(", ")`

Or better yet add `= []` in the method signature.

libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -1,5 +1,202 @@
...
+  IInflightHost,
+  IResource as IStdResource,
+  Resource as StdResource,
+  Node,
+  Json,
+} from "../std";
+
+/**
+ * Contract that a resource backend must implement.
+ * @inflightinterface
```
eladb, line 20: Why not just `@inflight`?

libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -1,5 +1,202 @@
...
+  onStop(): Promise<void>;
+}
+
+// TODO: fill out IResourceContext
+
+/**
+ * Context for implementing a simulator resource.
+ */
+export interface IResourceContext {
+  // /**
```
eladb, line 41: There you go!

eladb (commented):
❣️

examples/tests/sdk_tests/resource/resource.test.w
Chriscbr: I agree (I'm not sure what's better)

An alternative could be sim.IBackend

eladb: Let's leave it with "resource". I think it's probably the same thing... :-)

examples/tests/sdk_tests/resource/on-stop.test.w
```diff
@@ -0,0 +1,29 @@
+bring sim;
+
+inflight class OnStopThrowerBackend impl sim.IResource {
+  pub onStart(ctx: sim.IResourceContext) {}
+  pub onStop() {
+    throw "unexpected error!";
+  }
```
Chriscbr, line 7: When a resource fails while stopping, there's a pretty good argument for gracefully handling the error so that other resources can also clean up afterwards and release any of their resources, so that's basically what we do.

examples/tests/sdk_tests/resource/call.test.w
```diff
@@ -0,0 +1,221 @@
...
+    this.ctx = nil;
+  }
+
+  pub onStart(ctx: sim.IResourceContext) {
+    this.ctx = ctx;
+  }
+  pub onStop() {}
+
+  pub throwsError() {
+    throw "Look ma, an error!";
```
eladb, line 31: :-)

examples/tests/sdk_tests/resource/call.test.w
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  pub printLogs() {
+    this.ctx!.log("an info log");
```
eladb, line 35: What happens if I just write `log()` without going through the context? Ideally we should capture these logs and it should Just Work ™️

examples/tests/sdk_tests/resource/call.test.w
```diff
@@ -0,0 +1,221 @@
...
+    return { "field1Length": arg1[0].field1.length };
+  }
+}
+
+class MyResource {
+  backend: sim.Resource;
+  new() {
+    this.backend = new sim.Resource(inflight () => {
+      return new MyResourceBackend();
+    });
```
eladb, line 78: Is it possible to pass `ctx` to the factory?

```suggestion
    this.backend = new sim.Resource(inflight (ctx) => {
      return new MyResourceBackend(ctx);
    });
```

examples/tests/sdk_tests/resource/on-start.test.w
```diff
@@ -0,0 +1,35 @@
+bring sim;
+
+inflight class OnStartThrowerBackend impl sim.IResource {
+  pub onStart(ctx: sim.IResourceContext) {
+    throw "unexpected error!";
+  }
```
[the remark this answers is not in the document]
eladb: I think that if a resource fails to start there's no real value in continuing the simulation. This PR #6408 improves this behavior.

examples/tests/sdk_tests/resource/on-stop.test.w
eladb: Agreed

examples/tests/sdk_tests/resource/tokens.test.w
```diff
@@ -0,0 +1,40 @@
+bring sim;
+
+inflight class BadTokenResolverBackend impl sim.IResource {
+  var ctx: sim.IResourceContext?;
+
+  new() { this.ctx = nil; }
+
```
eladb, line 7: Great job with all the tests!

libs/wingc/src/type_check.rs
```diff
@@ -4495,9 +4526,59 @@ impl<'a> TypeChecker<'a> {
 				}
 			}
 		}
+
+		// Check that if the class implements sim.IResource, then the
```
eladb, line 4530: I think this should be checked for all inflight methods, not just if the class implements `sim.IResource`...

libs/wingsdk/src/shared/sandbox.ts
```diff
@@ -4,12 +4,13 @@ import { readFile, stat } from "fs/promises";
...
 import { Bundle, createBundle } from "./bundling";
 import { processStream } from "./stream-processor";
+import { LogLevel } from "../target-sim";
 
 export interface SandboxOptions {
   readonly env?: { [key: string]: string };
   readonly context?: { [key: string]: any };
   readonly timeout?: number;
-  readonly log?: (internal: boolean, level: string, message: string) => void;
+  readonly log?: (internal: boolean, level: LogLevel, message: string) => void;
```
eladb, line 13: This will conflict with #6408 (but aligned!)

libs/wingsdk/src/target-sim/resource.inflight.ts
```diff
@@ -0,0 +1,149 @@
...
+      }
+    }
+    message += ")";
+    return message;
+  }
+
+  private addTrace(message: string, type: TraceType, level?: LogLevel) {
+    this.context.addTrace({
+      data: { message, level },
+      type,
```
eladb, line 143: See #6408 -- I put the `level` at the top level of `Trace`.

libs/wingsdk/src/target-sim/resource.md
```diff
@@ -0,0 +1,153 @@
...
+  c.inc();
+  log("counter is now {c.peek()}");
+});
+```
+
+### Serializability
+
+TODO
+
+### Lazy tokens
```
eladb, line 109:
```suggestion
### Late-bound Tokens
```

libs/wingsdk/src/target-sim/resource.md
```diff
@@ -0,0 +1,153 @@
...
+
+Consider a use case where there is an attribute of a simulated service that only gets resolved during initialization (e.g. the exposed port of a container).
+In order to create such resources, we need a way to obtain a lazy token that gets resolved during simulator initialization.
+
+Use the preflight method `resource.attrToken(key)` to obtain a token that can be used to reference the value of the attribute at runtime.
```
eladb, line 114: I would call these `resolveToken()` and `createToken()`

libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -1,5 +1,308 @@
...
+  resolveAttr(name: string, value: string): void;
+
+  /**
+   * Log a message at the current point in time. Defaults to `info` level.
+   *
+   * @param message The message to log.
+   * @param level The severity of the message.
+   * @inflight
+   */
+  log(message: string, level: LogLevel | undefined): void;
```
eladb, line 65: Why is this needed here? Can't we just support `log()`?

libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -1,5 +1,308 @@
...
+}
+
+/**
+ * The severity of a log message.
+ */
+export enum LogLevel {
```
eladb, line 71: This should be at the `Trace` level

libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -1,5 +1,308 @@
...
+            return prop;
+          }
+          return await prop.call($klass, ...args);
+        };
+
+        exports.stop = async function() {
+          if (!$klass) {
+            throw Error('resource not started');
+          }
+          await $klass.onStop();
```
eladb, line 214: Can we put `onStart` and `onStop` in constants so it's not that fragile?

Also - is it possible to put most of this code outside in some helper function that's properly compiled and unit-tested?

libs/wingsdk/src/target-sim/util.ts
```diff
@@ -71,6 +71,34 @@ export function makeSimulatorJsClient(filename: string, resource: Resource) {
 })()`;
 }
 
+export function makeSimulatorJsClientV2(filename: string, resource: Resource) {
+  const type = basename(filename).split(".")[0];
+  const env = makeEnvVarName(type, resource);
+  return `(function() {
+  let handle = process.env.${env};
```
eladb, line 78: can we move to a helper and not emit all this code?

libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -1,5 +1,308 @@
...
+          }
+          const attrs = {};
+          const ctx = {};
+          ctx.statedir = () => statedir;
+          ctx.resolveAttr = (name, value) => attrs[name] = value;
+          ctx.log = (message, level) => {
+            if (!level) level = 'info';
+            console.log(level + ': ' + message);
+          };
+          const client = ${inflightClient};
```
eladb, line 176: Definitely seems like you can pass `ctx` to the factory function instead of `onStart`

eladb (changes requested):

examples/tests/sdk_tests/resource/on-start.test.w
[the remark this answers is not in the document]
Chriscbr: Sounds good to me 👍

examples/tests/sdk_tests/resource/call.test.w
Chriscbr: Regular logs work as well, and the messages are treated as INFO level.

examples/tests/sdk_tests/resource/call.test.w
Chriscbr: Ah that's a good suggestion, let me try that -- should remove a bunch of ugliness if it works

Chriscbr: I guess if we take this idea to its logical conclusion than we can just rely on the constructor for starting resources and remove the `onStart` method, yeah?

libs/wingsdk/src/target-sim/resource.ts
Chriscbr: I see the argument for combining them. I'll open an a separate issue to keep tabs on this idea -- maybe we can add e.g. some kind of `levels` parameter to the built-in log function. I wanted to reduce the design surface area in this PR (there's already a lot going on) and there would might be some implementation details to work out since Node.js treats functions like `console.error` and `console.warn` the same as far as both emitting to STDERR for example.

libs/wingsdk/src/target-sim/resource.ts
Chriscbr: It seemed like `@inflight` was already being used to mean several things, but I'm OK with overloading it if that's your preference. Here's how the `@inflight` docstring works now when added to a TypeScript interface (per your suggestion):

- `@inflight foo.IBar` - means this interface is a preflight interface, but the inflight "half" of it is defined in `foo.IBar`
- `@inflight` (without a name) - means this is an inflight interface

Chriscbr: Implemented in the PR

eladb: I think they are all one and the same (concept)

libs/wingsdk/src/target-sim/resource.ts
Chriscbr: I'm not sure how to do this without adding more layers of runtime indirection. If you have ideas though, feel free to push something to the PR, it sounds like a nice improvement direction.

libs/wingc/src/type_check.rs
Chriscbr: Added a note in the PR description -- I think we can address this as a follow up

examples/tests/sdk_tests/resource/call.test.w
Chriscbr: Done

examples/tests/sdk_tests/resource/state.test.w
```diff
@@ -0,0 +1,45 @@
+// TODO: add a test for accessing uninitialized state
```
Chriscbr, line 1: TODO

libs/wingsdk/src/target-sim/util.ts
Chriscbr: https://github.com/winglang/wing/issues/6446

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/events.test.w_test_sim.md
```diff
@@ -2,6 +2,7 @@
 
 ## stdout.log
 ```log
+[ERROR] (no test) | Error: Cannot access context during class construction
```
Chriscbr, line 5: TODO - looks like a regression

libs/wingsdk/test/target-sim/__snapshots__/topic-producer.test.ts.snap
Chriscbr, line 1: TODO - this test is flakey

libs/wingsdk/src/target-sim/counter.backend.ts
eladb, line 1: I would keep this file name `counter.inflight.ts` (it's basically the same thing, no?)

examples/tests/sdk_tests/resource/call.test.w
MarkMcCulloh, line 1: Do these sdk tests need to be skipped in our cloud e2e tests?

libs/wingsdk/src/target-sim/queue.ts
```diff
@@ -167,6 +167,7 @@ export class QueueSetConsumerHandler {
         try {
           await ctx.handler($message.payload);
         } catch (error) {
+          // TODO: an error from user code is getting dropped - bad!
```
MarkMcCulloh, line 170: Since you already created an issue, you could add the link here

libs/wingsdk/src/target-sim/resource.ts
eladb: I agree that adding log levels to our `@log()` is out of scope for this PR (and we should add that support!).

But I am asking - if I write a resource backend and use `@log()`, where would this log go? Will it be emitted into the standard logging system?

If that's the case, I would just remove this `log()` method from `IResourceContext` for now and take care of supporting log levels through the standard logging APIs instead of introducing another API for logging just for resources.

eladb (approved):
Amazing job man!

MarkMcCulloh (approved):

examples/tests/sdk_tests/resource/call.test.w
Chriscbr: I think so, good callout

MarkMcCulloh: It's possible the tests would be marked as "unsupported". I forget exactly how that works but may be worth running the tests to see what happens

docs/docs/04-standard-library/sim/resource.md
```diff
@@ -0,0 +1,636 @@
...
+- enums
+- structs that contain only serializable types
+
+The `call` function on `sim.Resource` only supports arguments that are valid `Json` values.
+To convert a value to `Json`, use the `Json` constructor:
+
+```js
+let x: num = 5;
+let y: str = "a";
+this.backend.call("myMethod", [Json x, Json y]);
```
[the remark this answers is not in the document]
Chriscbr:
```
let x: num = 5;
let y: str = "a";
let z = [x, y];
```

This is invalid in Wing since array elements have to be a single type. For the compiler to know you want to create an array of Json, you'd have to either make the elements `Json` or specify the array as Json.

```
let z = [Json x, Json y];
let z = Json [x, y];
```

PR #6400, opened 2024-05-02
feat(compiler)!: explicit lift qualification statement

### LINKED ISSUE #6080: The syntax around `lift()` statements is confusing
### Use Case

This issue is for tracking the implementation for https://github.com/winglang/wing/pull/5951

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

Language Design

### Community Notes

yoav-steinberg:
fixes #6080

Initial implementation of the explicit lift statement as defined in the [rfc](https://github.com/winglang/wing/blob/main/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md).

```wing
inflight () => {
  let b = bucket; // Assing preflight object to inflight variable
  lift {bucket: put} {
    b.put("k","v"); // Within this block we can use the inflight variable because of the explicit lifting above
  }
  b.put("k","v"); // This is an error, because it's outside the explicit lift block and we can't figure out who `b` is
  
  // We can also define multiple lifts in a block and multiple ops per lift
  lift {bucket1: [put, get], bucket2: delete} { ... }
  
  // We can nest lift blocks
  lift {bucket1: put} {
    ..
    lift {bucket2: get} {
      ..
    }
  }

  // Type checker validates passed methods are really part of the inflight interface of the object
  lift {bucket: leak} {}
              //^ err: leak isn't an inflight method on type Bucket
}
```

Breaking change: this is instead of the `lift()` builtin macro that was used for the same purpose. 

## Checklist

- [x] Tests added (always)


libs/wingc/src/parser.rs
```diff
@@ -637,6 +638,43 @@ impl<'s> Parser<'s> {
 		})
 	}
 
+	fn build_lift_statement(&self, statement_node: &Node, phase: Phase) -> DiagnosticResult<StmtKind> {
+		// Lift statements are only legal in inflight
+		if phase != Phase::Inflight {
+			return self.with_error("Lift statements are only allowed in inflight phase", statement_node);
+		}
```
MarkMcCulloh, line 645: Is there a technical reason for this restriction? Could/Should this work?

```wing
// I could only think of a contrived example
let b = new cloud.Bucket();
lift {b: [get]} {
    b.get("blah");
  })
}
```

yoav-steinberg: Interesting. Today the lift mechanism stores lift information per method (in your example the implicit `handle` method). So when specifying `lift` or when actually doing something like `preflight_bucket.put()` the lift visitor knows in which method we're in and stores the relevant information for that method.

What you're suggesting implies that you can specify explicit lifts that are assigned to multiple methods:
```wing
lift {bucket: get} {
  class Foo {
    inflight foo_m() {
      let b = bucket;
      b.get("hey");
    }
  }
  class Boo {
    inflight boo_m() {
      let b = bucket;
      b.get("hey");
    }
  }
```

I'll need to modify the lift visitor to get that working..
I suggest a future improvement, if at all.

libs/wingc/src/type_check.rs
```diff
@@ -6288,6 +6253,76 @@ impl<'a> TypeChecker<'a> {
...
+				self.spanned_error(
+					&qual.obj,
+					format!("Expected a preflight object, but found {obj_phase} expression instead"),
+				);
+			}
+			// Make sure the object type is a preflight type
+			if !obj_type.is_preflight_object_type() {
+				self.spanned_error(
+					&qual.obj,
+					format!("Expected a preflight object type, but found {obj_type} instead"),
```
Chriscbr, line 6281: (nit) I feel like "preflight object type" could be confusing. Maybe we can mention something in the diagnostic as a hint, or reword this in some way

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -3847,25 +3879,39 @@ Duration <DURATION>"
...
-   --> ../../../examples/tests/invalid/simulator_permissions.test.w:20:8
+error: Unknown parser error
+   --> ../../../examples/tests/invalid/simulator_permissions.test.w:20:3
    |
 20 |   lift(buckets.at(0), [\\"list\\"]);
-   |        ^^^^^^^^^^^^^ Expected a preflight object as first argument to \`lift\` builtin, found inflight expression instead
+   |   ^^^^ Unknown parser error
+
+
+error: Expression of type \\"Bucket\\" references an unknown preflight object, can't qualify its capabilities. Use \`lift()\` to explicitly qualify the preflight object to disable this error.
```
Chriscbr, line 3910: Can we add a hint to this diagnostic that links to the language reference section? I imagine here is a good place: https://www.winglang.io/docs/concepts/inflights#explicit-lift-qualification

Chriscbr, line 3910:
```suggestion
error: Expression of type \\"Bucket\\" references an unknown preflight object, can't qualify its capabilities. Use a \`lift\` block to explicitly qualify the preflight object and disable this error.
```

libs/wingc/src/type_check.rs
```diff
@@ -6288,6 +6253,76 @@ impl<'a> TypeChecker<'a> {
...
+							self.spanned_error(
+								op,
+								"Only instance (non-static) members may qualify the lift".to_string(),
+							);
+						}
+					}
+					Some(_) => panic!("expected object envs to only have variables"),
+					None => self.spanned_error_with_annotations(
+						op,
+						format!("Object of type {obj_type} does not have an inflight member named {op}"),
```
Chriscbr, line 6306:
```suggestion
						format!("Object of type {obj_type} does not have an inflight member named \"{op}\""),
```

libs/wingc/src/parser.rs
```diff
@@ -637,6 +638,37 @@ impl<'s> Parser<'s> {
 		})
 	}
 
+	fn build_lift_statement(&self, statement_node: &Node, phase: Phase) -> DiagnosticResult<StmtKind> {
+		// Lift statements are only legal in inflight
+		if phase != Phase::Inflight {
+			return self.with_error("Lift statements are only allowed in inflight phase", statement_node);
```
Chriscbr, line 644:
```suggestion
			return self.with_error("Lift blocks are only allowed in inflight phase", statement_node);
```

PR #6402, opened 2024-05-03
fix: inflight() features in @wingcloud/framework do not work

MarkMcCulloh:
When the inflight() function was moved to the SDK, the special auto-import and extra validation provided by the ts framework stopped working. This PR makes the logic for detecting this special function slightly more rigorous by using jsdoc tags instead of relying on detecting the source module.


Chriscbr (approved):

Chriscbr (commented):
Test?

Issue #5426, opened 2024-01-05
Wing Console doesn't recover from a simulator crash


eladb:
### I tried this:

This is expected to fail because the token is never resolved:

```js
bring sim;

class Foo {
  pub x: str;
  new() {
    let s = new sim.State();
    this.x = s.token("never-resolved");
  }
}

let f = new Foo();

new cloud.Function(inflight () => {
  log(f.x);
});
```

Then, try to remove the `log(f.x)` line, and I would expect the Wing Console to recover.

### This happened:

Since the token is never resolved, the initialization of the simulator instance fails with `Could not start resource root/Default/cloud.Function after 10 attempts.`. This is expected;.

But then let's say you remove the `log(f.x)` line, but Wing Console is stuck in `compiling...`.

### I expected this:

Wing Console should recover after the code was fixed.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

staycoolcall911:
@eladb - #6408 was supposed to close this, no?

eladb:
I don’t think so. Maybe?

skyrpex:
Still an issue.

monadabot:
Congrats! :rocket: This was released in [Wing 0.76.4](https://github.com/winglang/wing/releases/tag/v0.76.4).

PR #6408, opened 2024-05-04
feat(sdk): containers logs and improved simulator stability

### LINKED ISSUE #6052: Wing Console fails to start but no diagnostics is emitted
### I tried this:

Run `wing it` against this commit: https://github.com/winglang/setup-cdk-ddbstream/commit/2acc7997987711f84ed3b59c155568d50f72fe3a

### This happened:

The Wing Console is stuck in "Loading Simulator" and never ends.

When running `wing test` I see some errors.

### I expected this:

I expect the Wing Console to provide real-time diagnostics if something is failing instead of hanging forever.

This is related to log streaming but as a general rule we should make sure we always provide feedback to the user if something in their app doesn't work. We cannot assume the app will not have issues :-)

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

eladb:
## Container improvements

Revisit how we are running containers in `sim.Container` to improve robustness and capture logs from docker build/pull commands and from running containers.

To do that we've changed how container lifecycle is managed by running it without `--detach` and monitoring the child process for exit codes, errors and stdio capture.

## Simulator stability improvements

Additionally, this fixes a few long standing issues in the simulator which caused it to hang when one of the resources failed to start (e.g. due to unresolvable token). To fix this, we track failures and if there were any failing resources, we stop all the started resources and throw an error. Additionally, in the CLI we needed to catch this error, stop the spinner and bail out from the test execution routine. I believe this fixes #6052 and #5426.

## Logging improvements

Add some initial support for log levels, because, come on, currently errors emitted from the simulator or resources are hidden by default because they are treated as "verbose". This includes the relevant Console fixes.

Tweak a little how logs are displayed when `DEBUG=1` so that log entries are separated by a newline and removed the indentation of the first line (basically like jest/vitest is doing).

## Other improvements

Hide the `cloud.Endpoint` nodes from the graph because it was a mistake (mine) to suggest to show them.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -194,6 +241,114 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+      get started() {
+        return started;
+      },
+      async kill() {
+        return new Promise((ok, ko) => {
+          if (!started) {
+            return ok();
+          }
+
+          child.kill("SIGKILL");
```
MarkMcCulloh, line 326: Ideally we should use SIGTERM first. If that doesn't work we can SIGKILL

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -55,50 +70,47 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+        });
+      } else {
+        this.addTrace(
+          `Image ${this.imageTag} not found, pulling...`,
+          TraceType.RESOURCE
         );
+        await this.docker("pull", [this.imageTag], {
+          stdout: true,
+          stderr: true,
+        });
```
MarkMcCulloh, line 92: Should stdout from pull/build be considered verbose?

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -126,53 +138,88 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+  }
+
   public async cleanup(): Promise<void> {
-    this.log(`Stopping container ${this.containerName}`);
-    await runCommand("docker", ["rm", "-f", this.containerName]);
+    this.addTrace(
+      `Stopping container ${this.containerName}`,
+      TraceType.RESOURCE
+    );
+    await this.child?.kill("SIGKILL");
```
MarkMcCulloh, line 221: The given code isn't used

```suggestion
    await this.child?.kill();
```

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -194,6 +241,114 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+          if (!started) {
+            return ok();
+          }
+
+          child.kill("SIGKILL");
+          child.once("error", ko);
+          child.once("exit", ok);
+        });
+      },
+      async join() {
```
MarkMcCulloh, line 331: nit: Maybe just call this `output` or `stdout`?

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -194,6 +241,114 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+      output.push(data);
+
+      if (logStdout) {
+        this.addTrace(data.toString(), TraceType.LOG, LogLevel.INFO);
+      }
+    });
+
+    if (logStderr) {
+      child.stderr.on("data", (data) => {
+        this.addTrace(data.toString(), TraceType.LOG, LogLevel.INFO);
```
MarkMcCulloh, line 300: Should this also have LogLevel.ERROR?

MarkMcCulloh: I know stderr is also often used for "verbose" output, but I feel like putting it at the same log level as stdout seems off

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -126,53 +138,88 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+      TraceType.RESOURCE,
+      LogLevel.INFO
+    );
 
     return {
       [HOST_PORT_ATTR]: hostPort,
     };
   }
 
+  private createEnvFile() {
```
MarkMcCulloh, line 187: This method can be async and use the regular fs functions instead of the `std` fs

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -194,6 +241,114 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
...
+    options: DockerSpawnOptions = {}
+  ): DockerProcess {
+    let logStdout = options.stdout ?? false;
+    const logStderr = options.stderr ?? false;
+    const onError = options.onError;
+
+    // can be used to hide container logs (used in our end to end tests)
+    if (process.env.WING_HIDE_CONTAINER_LOGS) {
+      logStdout = false;
+    }
```
MarkMcCulloh, line 265: Is this avoidable? Why do the logs need to be hidden in the e2e tests? Is there a way we can do filtering/redacting to make the output work fine in the e2e tests

tools/hangar/__snapshots__/test_corpus/sdk_tests/container/container.test.w_test_sim.md
```diff
@@ -2,6 +2,61 @@
...
+[INFO] get app | #6 sha256:86562bfdb01870649ea3dcc082af03ed25e8cdccb638638b8fa09ebc169a2995 2.34MB / 2.34MB 0.2s done
+#6 extracting sha256:96526aa774ef0126ad0fe9e9a95764c5fc37f409ab9e97021e7b4775d82bf6fa 0.1s
+#6 sha256:20bf5151d902500e94a168975a49a04897e1f895290d75782ff2da96becab400 451B / 451B 0.2s done
+[INFO] get app | #6 sha256:bd866a276fb26dd895780268b9622a1d92dc4393ab85054317bbd1ac40504459 7.34MB / 49.80MB 0.3s
+[INFO] get app | #6 sha256:bd866a276fb26dd895780268b9622a1d92dc4393ab85054317bbd1ac40504459 17.83MB / 49.80MB 0.5s
+#6 extracting sha256:96526aa774ef0126ad0fe9e9a95764c5fc37f409ab9e97021e7b4775d82bf6fa 0.3s done
+[INFO] get app | #6 sha256:bd866a276fb26dd895780268b9622a1d92dc4393ab85054317bbd1ac40504459 23.07MB / 49.80MB 0.6s
+[INFO] get app | #6 sha256:bd866a276fb26dd895780268b9622a1d92dc4393ab85054317bbd1ac40504459 33.55MB / 49.80MB 0.8s
+[INFO] get app | #6 sha256:bd866a276fb26dd895780268b9622a1d92dc4393ab85054317bbd1ac40504459 44.04MB / 49.80MB 1.0s
+[INFO] get app | #6 sha256:bd866a276fb26dd895780268b9622a1d92dc4393ab85054317bbd1ac40504459 49.28MB / 49.80MB 1.1s
```
MarkMcCulloh, line 42: With a default configuration I don't think I'd want to see these logs

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: I am not sure it is avoidable because build/pull logs are not deterministic. They also have a bunch of timestamps and other non-deterministic things in them. Trying to be pragmatic about it...

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: Since these are pretty long operations in many cases (and we don't have great visibility into the start up state, which we will improve), I figured for now they should be INFO so that users will understand what's going on instead of thinking the simulator is stuck. I am going to work with @skyrpex on improving logging and diagnosability later.

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: I don't think this should be an ERROR level. Despite it's name "STDERR" is not just for errors. ERROR logs in the Wing Console are going to be displayed with BIG RED HIGHLIGHTS so we should be careful to use them just when there's an actionable error from the system.

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: OK, moved to `writeFile` but still prefer to use our mkdtemp because it's nicer.

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: OK, I made the "VERBOSE" and add an "INFO" message that says pulling/building is happening.

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: I'll open an issue to follow up with this. Prefer to reduce the scope here.

tools/hangar/__snapshots__/test_corpus/sdk_tests/container/container.test.w_test_sim.md
eladb: Yes, changed to VERBOSE.

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: OK, Changed to SIGTERM for now. Let's see how this plays out. I think we have some timeout to kill stuff at the simulator level

libs/wingsdk/src/target-sim/container.inflight.ts
eladb: I think `join()` is the "proper" name for this [1], no? Because it blocks and waits for the child to finish.

[1]: https://stackoverflow.com/a/25391156

tools/hangar/__snapshots__/error.ts.snap
```diff
@@ -18,52 +18,7 @@ Duration <DURATION>"
 `;
 
 exports[`inflight_stacktraces.test.w 1`] = `
-"[ERROR] assert | Invoke (payload=undefined). Error: assertion failed: false
```
eladb, line 21: what happened here?

eladb:
@MarkMcCulloh ready for review

Chriscbr:
@eladb Hrm - it seems like errors that start from outside of the the unit tests aren't getting printed anymore, can we fix this?
<summary>sample wing app</summary>
<details>

```js
bring cloud;
bring util;

  throw "ah!";
});

test "my test" {
  log("my log");
  f.invokeAsync("");
  util.sleep(0.5s);
  log("my log 2");
}
```

</details>

Before:

Screenshot 2024-05-06 at 10 59 49 AM

After:

Screenshot 2024-05-06 at 11 00 08 AM

tools/hangar/__snapshots__/test_corpus/sdk_tests/bucket/rename.test.w_test_sim.md
```diff
@@ -2,7 +2,6 @@
 
 ## stdout.log
 ```log
-[ERROR] rename() | Copy (srcKey=no-such-obj.txt to dstKey=new-name.txt). Error: Source object does not exist (srcKey=no-such-obj.txt).
```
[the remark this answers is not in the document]
eladb: Yes, this is an internal error of the resource which is caught by the user, so it's actually a happy path and shouldn't pollute the logs with red errors. If it weren't thrown, it would be shown as an error.

libs/wingsdk/src/target-sim/container.inflight.ts
```diff
@@ -46,6 +49,20 @@ export class Container implements IContainerClient, ISimulatorResourceInstance {
   }
 
   public async init(context: ISimulatorContext): Promise<ContainerAttributes> {
+    try {
+      return await this.start(context);
+    } catch (e: any) {
+      this.addTrace(
+        `Failed to start container: ${e.message}`,
+        TraceType.RESOURCE,
```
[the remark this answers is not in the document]
eladb: I actually see a value in distinguishing between simulator, resource and user logs (in addition to the log level), so I wouldn't get rid of it too soon.

examples/tests/sdk_tests/container/pull-failure.test.w
```diff
@@ -0,0 +1,8 @@
+bring sim;
+bring cloud;
+
+let x = new sim.Container(
+  name: "pull-failure",
+  image: "foo://do-not-pull-me",
+  containerPort: 1234,
+);
```
MarkMcCulloh, line 8: For any .w files tested by hangar that use containers, add this to the top because we currently can only test them on linux:

```suggestion
/*\
skipPlatforms:
  - win32
  - darwin
\*/

bring sim;
bring cloud;

let x = new sim.Container(
  name: "pull-failure",
  image: "foo://do-not-pull-me",
  containerPort: 1234,
);
```

main.w
Chriscbr, line 1: Leftover from debugging

polamoros (approved):

tools/hangar/__snapshots__/test_corpus/sdk_tests/container/container.test.w_test_sim.md
```diff
@@ -2,9 +2,17 @@
...
+[INFO] get echo | Container hashicorp/http-echo started
+[INFO] get app | Building my-app:a9ae83b54b1ec21faa1a3255f05c095c from ./my-docker-image...
+[INFO] get app | Container my-app:a9ae83b54b1ec21faa1a3255f05c095c started
+[INFO] get echo | Container my-app:a9ae83b54b1ec21faa1a3255f05c095c started
+[INFO] get app | Container hashicorp/http-echo started
 [INFO] get echo | bang
 [INFO] get echo | 
 [INFO] get app | Hello, Wingnuts!
+[WARNING] get echo | Timeout waiting for container root/env0/my-app to shutdown, removing forcefully
+[WARNING] get app | Timeout waiting for container root/env1/my-app to shutdown, removing forcefully
```
MarkMcCulloh, line 15: Should this be happening every time?

eladb: Yes because this container  doesn't react properly to signals. But that's okay I like that we are testing this.

Issue #3730, opened 2023-08-07
Detect list of files used by a Wing project


Chriscbr:
We would like to have a general way to track changes to a Wing project so that the Wing Console and other CLI commands (like `wing compile --watch` or `wing test --watch`) can detect when a Wing project has changed and needs to be refreshed. The solution should also solve the case when a user adds a new .w file into their project (i.e. uses `bring` on a file that hasn't been compiled before)

It's possible this could be some kind of function exposed by `@winglang/compiler` that lists all files in the project given an entrypoint file passed as an argument (i.e. `main.w` or `*.main.w`), or a custom watcher that takes an entrypoint and explicitly notifies (e.g. using callbacks or an [EventEmitter](https://nodejs.dev/en/learn/the-nodejs-event-emitter/) when a file in the project has changed.

P1: detect changes whenever Wing source files have changed
P2: detect changes whenever the `package.json` file has changed, or when extern JS files have changed

tsuf239:
closed #3956 as a duplicate, when implementing this, please make sure that the solution also fits the case described there (creating a react website with wing file at the same directory level/ one folder down)

Chriscbr:
(Added the console/simulator labels since not having a good watching mechanism impacts the DX of those tools)

Chriscbr:
I ran into this while working on a winglib. The scenario was I was testing out an example, and I was importing a module from a parent directory, like in the snippet below.

```js
bring cloud;
bring "../hello.w" as hello;

new hello.Hello();
```

Unfortunately, when I made changes to "hello.w", the console didn't auto-update which was a bit confusing (also the "Reset" button doesn't re-compile the app so it didn't help).

PR #6416, opened 2024-05-05
feat(console): add --watch option to `wing it` to watch additional directory

MarkMcCulloh:
This allows users to watch arbitrary places for file changes. This is mostly an escape hatch for https://github.com/winglang/wing/issues/3730, but even when that is implemented there will probably continue to be a desire to watch arbitrary stuff.
 


apps/wing/src/cli.ts
```diff
@@ -144,6 +144,10 @@ async function main() {
     .argument("[entrypoint]", "program .w entrypoint")
     .option("-p, --port <port>", "specify port")
     .option("--no-open", "Do not open the Wing Console in the browser")
+    .option(
+      "-w, --watch <globs...>",
+      "Watch additional paths for changes. Supports globs and '!' for negations."
+    )
```
MarkMcCulloh, line 150: Would love to add a test for this, but I don't see a good place to do so. The CLI currently mocks the console for its tests, and the console server doesn't seem to have any tests to add to. I'm open to ideas for useful testing.

Issue #6412, opened 2024-05-05
Access AWS lambda context in wing 


eladcon:
### Use Case

I want to access some information that is available on the context variable passed to the handler (E.g. access function name or request id in AWS)

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Chriscbr:
Maybe we can start by exposing the info through an API like

```js
bring aws;

new cloud.Function(inflight (msg: str) => {
  let ctx = aws.context; // or aws.getContext();
});
```

eladb:
Sounds good, but I'd go with `aws.Function.context()` to associate with Lambda functions

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.2](https://github.com/winglang/wing/releases/tag/v0.74.2).

PR #6424, opened 2024-05-06
feat(sdk): aws.Function.context() for lambda functions

Chriscbr:
Closes https://github.com/winglang/wing/issues/6412

## Checklist

- [x] Tests added (always)


examples/tests/sdk_tests/function/aws-function.test.w
```diff
@@ -35,4 +36,14 @@ test "validates the AWS Function" {
     // If the test is not on AWS, it should not fail, so I am returning true.
     assert(true);
   }
-}
\ No newline at end of file
+}
+
+if target == "tf-aws" {
+  test "can access lambda context" {
+    let ctx = aws.context();
```
eladb, line 43:
```suggestion
    let ctx = aws.Function.context();
```

I think this to be semantically tied to `aws.Function`. 

It should also return an optional and resolve to `nil` on non-AWS functions and in the simulator.

eladb (changes requested):

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -583,9 +583,6 @@ impl<'a> JsiiImporter<'a> {
 		if let Some(properties) = jsii_interface.properties() {
 			for p in properties {
 				debug!("Found property {} with type {:?}", p.name.green(), p.type_);
-				if member_phase == Phase::Inflight {
-					todo!("No support for inflight properties yet");
-				}
```
Chriscbr, line 588: FYI @yoav-steinberg

examples/tests/sdk_tests/function/aws-function.test.w
Chriscbr: Done

MarkMcCulloh (approved):

libs/awscdk/src/function.ts
```diff
@@ -214,4 +214,25 @@ export class Function
...
+   */
+  protected _getCodeLines(handler: cloud.IFunctionHandler): string[] {
+    const inflightClient = handler._toInflight();
+    const lines = new Array<string>();
+    const client = "$handler";
+
+    lines.push('"use strict";');
+    lines.push(`var ${client} = undefined;`);
+    lines.push("exports.handler = async function(event, context) {");
+    lines.push("  globalThis.$awsLambdaContext = context;");
```
eladb, line 229: I am concerned that multiple concurrent invocations will corrupt this global state.

Can we use https://nodejs.org/api/async_context.html?

libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -425,7 +425,8 @@ export class Function extends cloud.Function implements IAwsFunction {
 
     lines.push('"use strict";');
     lines.push(`var ${client} = undefined;`);
-    lines.push("exports.handler = async function(event) {");
+    lines.push("exports.handler = async function(event, context) {");
```
eladb, line 428: We should reuse this code across the two aws platforms

eladb (changes requested):

libs/awscdk/src/function.ts
Chriscbr: @eladb can you explain? Concurrent invocations would have to be running on separate AWS Lambda instances, wouldn't they?

eladb: Yes, that might be guaranteed by the runtime but not enforced by the programming model. There is a reason why `ctx` is an argument to the handler and not a global.

I would say that the "correct" solution is to use the local storage, and perhaps that's something we want to formalize in wing as well.

but if it's too complex, I would at least add some defensive coding which nullifies the global at the end of the run (in a finally block) and verified that it is null at the beginning.

Chriscbr: IMO if AWS is providing this as a contract, then it should be alright to optimize our private implementations around it -- the only issue would be if we introduce a bug where we start using these AWS Lambda wrappers on platforms that aren't AWS Lambda.

I like the idea of adding some error checking so at least we can catch if this assumption is broken 👍

eladb (approved):

PR #6428, opened 2024-05-07
feat(compiler): docstrings

### LINKED ISSUE #4446: Docstrings
### Feature Spec

The Wing language now has a well-defined syntax for docstrings, also known as docstring comments. This is a subset of the language's comment syntax (`//` and `/* */`) that the compiler will record as part of the documentation for a symbol for the purposes of docs and displaying API documentation within IDEs. These comments are to be distinguished from free-floating comments that are not used for API documentation.

## Spec

TBD

### P1

- Conventions for annotating parameters of functions
- Conventions for annotating return values and throw behavior
- Conventions for annotating APIs as deprecated _(should this be a language keyword?)_

### P2

- Providing compiler warnings if it looks like a user tried to write a docstring-like comment but they made an error somehow (not a compiler error as long as it is a comment)
- Conventions for writing code examples within docstrings, that can be compiled
- Conventions for using Markdown or selective parts of Markdown syntax within docstrings (e.g. `[link text](link url)`)

### Useful reading

- https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html
- https://go.dev/blog/godoc
- https://peps.python.org/pep-0257/
- https://en.wikipedia.org/wiki/Javadoc
- https://developer.apple.com/documentation/xcode/writing-symbol-documentation-in-your-source-files

### Use Cases

- Generating API docs (markdown, HTML, etc.)
- Displaying documentation in IDEs (e.g. when you hover over a class or method)

### Implementation Notes

_No response_

### Component

Language Design

### Community Notes

yoav-steinberg:
Resolves #4446

docstrings can be specified with `///`:
```wing
/// This is a doc string for `Foo`
/// And it has another line to it
class Foo {
  /// This is a docstring for `x`
  x: num;
}
```
image

## more stuff
Added support for hovering over the ctor and the class name in in `new` expressions to display ctor docs (instead of class docs).

## Checklist

- [x] Tests added (always)


libs/wingc/src/lsp/hover.rs
```diff
@@ -427,4 +427,133 @@ Json.stringify({});
...
+
+	test_hover_list!(
+		class_doc,
+		r#"
+  /// Class doc
+  class Foo {
+      //^
+  }
+  "#,
+	);
```
MarkMcCulloh, line 439: Maybe add some other cases, like

```wing
/// Class doc

class Foo {}
```

```wing
/// Class doc

class Foo { /// member doc
  pub a: num;
}
```

```wing
/// Class doc
/// With multiline
/// ## And markdown
class Foo {
  pub a: num;
}
```

libs/wingc/src/lsp/hover.rs
```diff
@@ -427,4 +427,133 @@ Json.stringify({});
...
+
+	test_hover_list!(
```
MarkMcCulloh, line 441: These tests are great, but we should probably have one or two docstrings in our E2E tests as well. Just in case...

MarkMcCulloh (approved):

libs/wingc/src/fold.rs
```diff
@@ -208,6 +208,7 @@ where
 		kind,
 		span: node.span,
 		idx: node.idx,
+		doc: node.doc, // TODO: maybe `fold_doc`?
```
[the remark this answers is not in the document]
yoav-steinberg: was note to self, i'll remove

libs/wingc/src/lsp/hover.rs
yoav-steinberg: done

libs/wingc/src/visit.rs
```diff
@@ -286,8 +286,9 @@ where
 	V: Visit<'ast> + ?Sized,
 {
 	v.visit_symbol(&node.name);
-	for value in &node.values {
+	for (value, _doc) in &node.values {
 		v.visit_symbol(value);
+		// TODO: Visit _doc
```
staycoolcall911, line 291: @yoav-steinberg - was that TODO left on purpose?

PR #6434, opened 2024-05-07
rfc: intrinsics rfc and spec update

### LINKED ISSUE #6378: Compiler intrinsics (or, "special macros")
### Use Case

In wing there are currently several built-in globals, all of which are injected with their own macro functionalities. These can intentionally have behaviors different from those possible to implement in wing itself. Some are simply macros that emit code, while some interact with the compiler in more novel/deep ways (e.g. `lift`). Even the simple macros have access to information that user-defined code would typically not (types, source position, etc.).

These are incredibly useful, but their behavior is surprising when to the user they appear no different than any other symbol/variable. Additionally they allow you to do things like do variable shadowing and assign them to variables which can actually break their functionality.

### Proposed Solution

Introduce a new syntax to reflect these special behaviors.

I propose `@x()` as a "compiler intrinsic", with the following notes:
- Having an intrinsic called `@x` does not mean that there is a symbol called `x`. Intrinsics are their own separate concept.
- It will not be possible to write `let variable = @x;` or otherwise use `@x` as an expression on its own, to prevent the impression that this a singular value being referenced
- While the arguments/return types must be normal wing types, the kind of expressions that can be passed as arguments may have certain restrictions. For example, it would be common to only allow non-interpolated string for certain cases to ensure all information is available statically.

- Why `@`?
  - It's not a valid identifier character, so there is no ambiguity that this is a special thing
  - A reasonable-looking alternative is x!(...) a la rust. Unfortunately ! is already used as the postfix unwrap. I think having a construct that is <identifier><!> mean 2 different things is ambiguous for both users and the compiler.
  - Having a special character as a prefix is useful for IDE completions because all you have to do is type `@` and you see all the available options.
  - It looks nice in [zig for builtins](https://ziglang.org/documentation/master/#Builtin-Functions) so there is at least some precedent
  - `@` is used in js for decorators which has a similar vibe of being special in the eyes of the compiler. It would not be unreasonable to use @ for decorators of some kind in wing as well.

With this, we can convert our existing globals to intrinsics instead and go forward implementing new ones. Some existing issues that I think would benefit:
- `@file` - https://github.com/winglang/wing/issues/5449
- `@inflight` - https://github.com/winglang/wing/issues/6045

### Other thoughts

I think it would be reasonable to expose the ability to create these in user-space via some sort of compiler API. This would be amazing but the design and implementation that would be a pretty huge task. Even if we did, we could still use the same syntax for invoke them even though we probably shouldn't call them "intrinsic" anymore.

### Community Notes

MarkMcCulloh:
copy/paste of https://github.com/winglang/wing/issues/6378 to facilitate further discussion

Closes #6378


docs/docs/03-language-reference.md
```diff
@@ -572,20 +572,27 @@ The following is a list of supported indexable types:
...
+  - For example, the return type may change between inflight and preflight
+
+| Name            | Extra information                                                                                                         |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `@log()`        | logs str                                                                                                                  |
+| `@assert()`     | checks a condition and _throws_ if evaluated to false                                                                     |
+| `@unsafeCast()` | cast a value into a different type                                                                                        |
+| `@nodeof()`     | obtain the [tree node](./02-concepts/02-application-tree.md) of a preflight object                                        |
+| `@lift()`       | explicitly qualify a [lift](./02-concepts/01-preflight-and-inflight.md#explicit-lift-qualification) of a preflight object |
```
eladb, line 590: `@dirname` is another candidate.

I don't see a reason to limit this only to function-like builtins, right?

docs/contributing/999-rfcs/2024-05-07-intrinsics.md
```diff
@@ -0,0 +1,39 @@
...
+  - It's not a valid identifier character, so there is no ambiguity that this is a special thing
+  - A reasonable-looking alternative is x!(...) a la rust. Unfortunately ! is already used as the postfix unwrap. I think having a construct that is <identifier><!> mean 2 different things is ambiguous for both users and the compiler.
+  - Having a special character as a prefix is useful for IDE completions because all you have to do is type `@` and you see all the available options.
+  - It looks nice in [zig for builtins](https://ziglang.org/documentation/master/#Builtin-Functions) so there is at least some precedent
+  - `@` is used in js for decorators which has a similar vibe of being special in the eyes of the compiler. It would not be unreasonable to use @ for decorators of some kind in wing as well.
+
+With this, we can convert our existing globals to intrinsics instead and go forward implementing new ones. Some existing issues that I think would benefit:
+- `@file` - https://github.com/winglang/wing/issues/5449
+- `@inflight` - https://github.com/winglang/wing/issues/6045
+
```
eladb, line 36: If this is a breaking change, it's worth discussing how we plan to introduce. For example, if we decide that `log()` will change, we should soft-deprecate it and not hard-deprecate.

docs/docs/03-language-reference.md
```diff
@@ -572,20 +572,27 @@ The following is a list of supported indexable types:
...
+Intrinsic functions are a special call-like expressions built into the Wing compiler with
+the following properties (given an example intrinsic `@x()`):
+- It is not possble to write `let variable = @x;` or otherwise use `@x` as an expression on its own
+- `x` is not automatically a symbol that can be referenced
+- The arguments/return types must be representable Wing types, but can be more dynamic than user-defined functions
+  - For example, the return type may change between inflight and preflight
+
+| Name            | Extra information                                                                                                         |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `@log()`        | logs str                                                                                                                  |
```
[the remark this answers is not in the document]
MarkMcCulloh: It should be okay to allow the previous "global functions" and these new intrinsics to coexist during deprecation, sure. I figured the spec could focus only on the future state.

If you're proposing that `log()` should always exist, then that's worth discussing. I prefer killing all the magical globals, and pragmatically the extra `@` isn't sufficient friction to be concerned.
99% percent of the time I expect the keystroke count to remain the same. With `log()`, my keystrokes will often be `<l> <o> <g> <enter>` because there can be conflicting symbols. With `@log()`, I expect to only type `<@> <l> <o> <enter>` because there are less options after `@`. As a pragmatic choice we can even make `@log` come before `@lift` in the completions, then you would only type `<@> <l> <enter>`

docs/docs/03-language-reference.md
MarkMcCulloh: I wanted to avoid listing possible candidates here. Any additions can hopefully be added as they're finalized.

> I don't see a reason to limit this only to function-like builtins, right?

I don't think there is a technical feasibility reason, but my thought is that only allowing `@x()` helps avoid thinking that @x is a "value". I don't feel super strongly about this so open to loosening it.

docs/contributing/999-rfcs/2024-05-07-intrinsics.md
MarkMcCulloh: added note

docs/docs/03-language-reference.md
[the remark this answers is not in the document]
eladb: I am okay with that, but I think @staycoolcall911 had some concerns.

Definitely the soft-deprecation will allow us to experiment with this and see how it feels.

eladb (approved):

docs/contributing/999-rfcs/2024-05-07-intrinsics.md
```diff
@@ -0,0 +1,41 @@
...
+title: "Intrinsic Functions"
+description: Global compiler intrinsics
+---
+
+# Intrinsics
+
+- **Author(s)**: @MarkMcCulloh
+- **Submission Date**: 2024-05-07
+
+### Use Case
```
eladb, line 11: @Chriscbr can you take a look?

eladb (approved):

Chriscbr (approved):

docs/docs/03-language-reference.md
eladb: `@dirname` is a value

MarkMcCulloh: It doesn't have to be (`@dirname()`)

Also, per some other async discussion it's not going to be the initial approach for the relative path use case (coming soon)

MarkMcCulloh: I'm remembering now why I had a more technical reason for avoiding just `@x`.
Let's say we have an intrinsic `@x` that returns a function. That means `@x()` would call the returned function. Does this mean any values passed as arguments there are part of the intrinsic call (where the compiler has special behaviors) or not? The parser could not handle this ambiguity so we'd have to decide dynamically.

Similar thing applies if it return an object or interface of some kind, where does the "intrinsic" part of `@x.func()` end? You could say it ends at `@x`, but then why would something like `@inflight("file.ts")` behave differently (where the argument passed and the entire expression get treated differently)?

Chriscbr: I think supporting "value intrinsics" (intrinsics that don't need to be called) makes sense.

re: implementation -- let's give another try to get this working. Maybe we can try parsing intrinsics as a new expression variant in tree-sitter, with a grammar rule like

```
intrinsic := "@" identifier ( "(" arg_list ")" )?
```

Or, we could leave the grammar as is, and add some logic to `parser.rs` that turns the appropriate call expression or identifier expression into a dedicated IntrinsicCall or IntrinsicValue node.

Or, we could augment the tree-sitter grammar to allow identifiers to start with "@", and just add appropriate checks in the compiler so you can't define new symbols starting with "@", and associate the built-in intrinsic symbols in the type checker with a special "intrinsic" Type.

MarkMcCulloh: I have an approach that seems to work so I'll update this RFC to remove the restriction. For now it is basically the first approach you mention, to avoid `@` and intrinsic getting possibly mixed up with other expressions/references.

PR #6442, opened 2024-05-08
fix(compiler): invalid errors and false positives when checking for field initialization in ctors

### LINKED ISSUE #3978: cant update mutable inflight fields in constructor context
### I tried this:

compiling this:

```
bring cloud;
class C1 {  
    inflight var i: str;
    init() {
        let api = new cloud.Api();
        api.post("/k1", inflight (req) => {
            this.i = "1";
            log("i: ${this.i}");
            return {
                status: 200
            };
        });
    }

    inflight init() {
        this.i  = "0";
    }
}
```

notice `this.i` is updated in a `cloud.Api` handler 

### This happened:

getting this error:

```
error: "i" cannot be initialized in the preflight initializer
  --> test.w:7:18
  |
7 |             this.i = "1";
  |                  ^ "i" cannot be initialized in the preflight initializer
```

### I expected this:

to compile.

### Is there a workaround?

moving the handler code from the ctor to a regular method 

### Component

Compiler

### Wing Version

0.26.5

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

yoav-steinberg:
fixes #3978 

When checking that all fields are initialized in the ctor the visitor also delved into inner closures/methods and looked for field assignments inside them. This cause some errors like:
* Assuming a field is initialized if it's being assigned to inside an inner closure even if the closure is never called.
* If an inner closure is of a different phase that the ctor and the closure assigns to a field of it's own phase it would produce an error as if the ctor is trying to initialize a field it doesn't have access to.

Note that our field initialization code is far from perfect and this change will produce the following false negative:
```wing
class Foo {
  x: num;
  new() {
    () => { this.x = 1 }(); // In place execution: this initializes `x` but we'll now report an error forcing the user to be clearer about their intent.
  }
}
```

## Checklist

- [x] Tests added (always)

### DIFF
--- a/tools/hangar/__snapshots__/invalid.ts.snap
+++ b/tools/hangar/__snapshots__/invalid.ts.snap
@@ -880,21 +880,11 @@ error: Expected \\"S1\\" to be a class
    |                  ^^
 
 
-error: Class cannot extend itself
+error: Unknown symbol \\"C11\\"
    --> ../../../examples/tests/invalid/class.test.w:89:19
    |
 89 | class C11 extends C11 {
    |                   ^^^
-
-
-error: Symbol \\"C11\\" already defined in this scope
-   --> ../../../examples/tests/invalid/class.test.w:89:7
-   |
- 8 | class C11 {
-   |       --- previous definition
-   .
-89 | class C11 extends C11 {
-   |       ^^^
 
 
 error: Missing super() call as first statement of PaidStudent's constructor


examples/tests/invalid/class.test.w
```diff
@@ -5,6 +5,17 @@ class C1 {
 // ^ Preflight field "x" is not initialized
 }
 
+class C1_1 {
+  x:num;
+  new() {
+    () => {
+      // This doesn't count as an initialization because it's in a closure that might not be called
+      // and should produce an error.
+      this.x = 1;
```
MarkMcCulloh, line 14: Since this will be a somewhat subtle breaking change, could you compile the projects in https://github.com/winglang/winglibs to preemptively prepare any needed fixes? Just to avoid pagerduty alerts :)

MarkMcCulloh (approved):

examples/tests/invalid/class.test.w
yoav-steinberg: done

Chriscbr:
Nice catch!

PR #6443, opened 2024-05-08
feat: source directory intrinsic `@dirname`

### LINKED ISSUE #3736: `__dirname`
### Feature Spec

We're currently having the option to pass the relative or absolute path as the website directory path. 
That is done by using WING_SOURCE_DIR (and [soon](https://github.com/winglang/wing/pull/3723) via `App.of(<website scope>).sourceDir`)

When creating a multifile wing project this will only work if the website will be declared at the entry point...

Elad suggested this approach: 
> It feels like this should somehow be implicitly passed when new cloud.Website is called. Maybe we can add some magic 
 keyword argument called __dirname. If the compiler sees it, it will assign the file's directory to it implicitly.
>
>Something like:
>
>```
>new cloud.Website();
>
>// implies
>new cloud.Website(__dirname: __dirname);
>```

### Use Cases

multifile wing

### Implementation Notes

_No response_

### Component

Compiler, SDK

### Community Notes

MarkMcCulloh:
Fixes #3736

Implements the first compiler intrinsic per #6434. This removes the need for __dirname hacks used currently in the ecosystem. Also implements some basic support for intrinsic in completions/hover.


apps/vscode-wing/language-configuration.json
```diff
@@ -3,7 +3,7 @@
         "lineComment": "//",
         "blockComment": [ "/*", "*/" ]
     },
-    "wordPattern": "/([a-zA-Z_$][A-Za-z_$0-9]*)/",
+    "wordPattern": "([a-zA-Z_@][\\w_]*)",
```
yoav-steinberg, line 6: purposefully not considering `$` a valid char anymore?

examples/tests/valid/intrinsics.test.w
```diff
@@ -0,0 +1,23 @@
+bring fs;
+bring expect;
+bring "./subdir/bar.w" as bar;
+
+let path = "SHOULD_IGNORE";
+let filename = "intrinsics.test.w";
```
yoav-steinberg, line 6: of course we need another intrinsic for `@filename()`.

examples/tests/valid/intrinsics.test.w
```diff
@@ -0,0 +1,23 @@
...
+expect.equal(fs.dirname(currentFile), currentDir);
+
+let currentDirAlt = @path("./");
+expect.equal(currentDir, currentDirAlt);
+
+let upDir = @path("..");
+expect.equal(fs.dirname(currentDir), upDir);
+
+let packageJson = @path("package.json");
+fs.join(currentDir, "package.json");
```
yoav-steinberg, line 20: mssing an `expect.equals(packageJson, resultOfFsJoin);`?

libs/tree-sitter-wing/grammar.js
```diff
@@ -349,9 +350,19 @@ module.exports = grammar({
...
       ),
 
+    intrinsic: ($) =>
+      prec.right(
+        seq(
+          field("name", $.intrinsic_identifier),
+          optional(field("args", $.argument_list))
+        )
+      ),
+    intrinsic_identifier: ($) => /@[a-zA-Z]*/,
```
yoav-steinberg, line 364: No underscores or numbers? Should probably be same as any old identifier. Not critical I guess.

libs/tree-sitter-wing/test/corpus/expressions.txt
```diff
@@ -648,3 +648,17 @@ assert(x[y] == w[z]);
                   (reference_identifier))
                 (reference
                   (reference_identifier))))))))))
+
+================================================================================
+Debug panic
```
yoav-steinberg, line 653:
```suggestion
Intrinsic
```

yoav-steinberg (approved):
looks good

apps/vscode-wing/language-configuration.json
eladb: I agree with this change. We should reserve `$` for JavaScript implementation details.

examples/tests/valid/intrinsics.test.w
eladb: Why?

apps/vscode-wing/language-configuration.json
MarkMcCulloh: That part of the change was unintentional, I completely forgot that `$` is a valid identifier character.
I'll undo the change here even though I think I agree because it would belong in another PR.

examples/tests/valid/intrinsics.test.w
```diff
@@ -0,0 +1,23 @@
+bring fs;
+bring expect;
+bring "./subdir/bar.w" as bar;
+
+let path = "SHOULD_IGNORE";
+let filename = "intrinsics.test.w";
+let currentFile = @path(filename);
+expect.equal(fs.basename(currentFile), filename);
+
+let currentDir = @path(".");
```
eladb, line 10: TBH I am not sure why is this better than `__dirname`. I am assuming we are not checking that the referenced path actually exists, so effectively this is just `join(__dirname, ...)`. 

What am I missing?

I can see the benefit of something like `@asset()` which can be lifted and all but if this is just to reference something relative to the current directory of the source file, I don't think that's the right design.

I would also say the name is somewhat misleading and doesn't imply that this is about the current source file. 

My recommendation is to just go with `@dirname` and keep it simple and low surface for now and introduce something like `@asset()` if we really want to do the right thing.

eladb (changes requested):

examples/tests/valid/intrinsics.test.w
MarkMcCulloh: My issues with `@dirname`:
1. It would have to be `@dirname()` based the reasoning I proved in the other rfc
2. 99% percent of the time, you would immediately `join(@dirname(), "something")`, so let's design for the common case
3. This API mimics other types of "relative path" APIs we're going to have. `@asset()` `@inflight()` will look similar.
4. `@asset()` has more implications than `@path()` imo so I think it can coexist
5. This forces paths to be resolved upfront which should help to remind users that the path they are referencing is special.

> I would also say the name is somewhat misleading and doesn't imply that this is about the current source file.

This is fair, I'm open to alternative name but I still think the API is valuable. Any other names I could think of were either too long or ambiguous. `@resolve()` is the only other idea I kind of like but it implies there is more interesting resolution logic like node's. `@asAbsolute()` is another idea.

Also, I think the main reason that `@dirname()` doesn't seem misleading is because of our familiarity with JS. I can easily see that being ambiguously referring to the cwd.

> keep it simple and low surface

You've mentioned simplicity a few times and I don't see what is complicated about `@path()`. As you said it's basically just dirname + join and the implementation is already complete. The surface area is something that will have to be deal with with other intrinsic so it will soon be well trodden.

ekeren: > 99% percent of the time, you would immediately join(@dirname(), "something"), so let's design for the common case


I did a search in GitHub (public repositories) of all the occurrence of __dirname and noticed that in all occasions __dirname was used with join , so I think that @path() makes a lot of sense. Just as a side note , I have conducted this search a couple of days ago to prove to Mark that we need __dirname and @path but have seen that I am probably wrong. @path(.) will probably not be used a lot

ekeren: I also , like the @inflight(), I feel it sets us up for better cross language support

Chriscbr: I can see the argument for either side, `@path("./foo")` seems like an ok shorthand for `fs.join(@dirname, "./foo")` but I'm also ok with leaving it out for now if there are ambiguity issues. @MarkMcCulloh It's your call

MarkMcCulloh: reworked to `@dirname`

eladb (commented):
What happened?

😌

Issue #4785, opened 2023-11-03
Add more assertions to the expect module


skorfmann:
### Feature Spec

More capabilities for the expect module are making it easier to express test expectations for wing applications.

- `expect.toThrow` / `expect.doesNotThrow` // functions
- `expect.contains` / `expect.doesNotContain` // array, set, string, range
- `expect.match` / `expect.doesNotMatch` // string
- `expect.fail("message")` // always fails
- `expect.ok(condition[, "message"])`

Follow up for https://github.com/winglang/wing/pull/3976

### Use Cases

testing capabilites which are not expressable right now but pretty standard in other testing frameworks 

one use case from examples:

```
assert(response.body?.contains(expected) == true);
```

### Implementation Notes

So far the `expect` module is mirroring https://nodejs.org/api/assert.html - this will still be possible for `throw` and `match`. However, `includes` is not supported there. So this needs to be either implemented manually or we'll need to reach for another library.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

staycoolcall911:
Added `fail` and `ok` from #4789

revitalbarletz:
Srinu Madhav is trying to look at this one.

srinumadhavv:
Hey @skorfmann  as far as my understanding is this correct ?
```
expect.toThrow -> (fn: () => void, expected: string)  should accept the parameters and throws the expected error 
expect.doesNotThrow ->  doesNotThrow(fn: () => void) should accept the parameters and doesn't throw error
expect.match // string -> match(actual: string, expected: RegExp)  should accept the parameters and check if the string matches regex expression 
expect.doesNotMatch -> doesNotMatch(actual: string, expected: RegExp) should accept the parameters and check if the string doesn't match regex expression 
```
and also need some clarification on these methods 
```
expect.contains / expect.doesNotContain // array, set, string, range
expect.ok(condition[, "message"])
```

staycoolcall911:
Hi @srinumadhavv you seem to be correct. 
If you have doubts stick to Node's implementation, for example: https://www.geeksforgeeks.org/node-js-assert-throws-function/

In a similar manner `expect.contains` should succeed if the first argument is contained in the container type passed as the second argument: 
[...]

PR #6450, opened 2024-05-09
feat(sdk): add functions to expect module

srinumadhavv:
## Checklist

- [ ] Tests added (always)

**Evidences**
image
image

NOTE : As Mentioned by @Chriscbr  , this pr doesn't implement the following methods in issue #4785 
```
contains()
doesNotContain()
throws()
doesNotThrow() 
```
are left until we have a proper support for typing 

Once Merged this should close a part of #4785 for methods : `ok, fail, match, doesNotMatch`


tsuf239:
Can you please add a test written in Wing? 
similar to `examples/tests/sdk_tests/expect/assert.test.w` or in the same file
Can you please add a link to the issue? (if you add "fixes #issue_number" it will be closed automatically when the PR is merged)

srinumadhavv:
> Can you please add a test written in Wing? similar to `examples/tests/sdk_tests/expect/assert.test.w` or in the same file Can you please add a link to the issue? (if you add "fixes #issue_number" it will be closed automatically when the PR is merged)

Hey @tsuf239  , where should i add this "fixes #issue_number" is it in the description of the pr or anywhere else ?
and regd the test file i should add an example to each of the methods covered in this issue am i right ?

staycoolcall911:
> Hey @tsuf239 , where should i add this "fixes #issue_number" is it in the description of the pr or anywhere else ? and regd the test file i should add an example to each of the methods covered in this issue am i right ?

Yes and yes

libs/wingsdk/src/expect/assert.ts
```diff
@@ -62,5 +62,120 @@ export class Util {
...
+
+  /**
+   * Checks if a value contains another value.
+   * Supports arrays, sets, and strings.
+   * @param actual The value to test.
+   * @param expected The value to check for containment.
+   * @throws Will throw an error if the actual value does not contain the expected value.
+   * @returns void
+   */
+  public static contains<T>(actual: unknown, expected: T | string): void {
```
Chriscbr, line 137: The signature of this method might cause issues in Wing since the language doesn't support type parameters yet, and we don't want the method to be typed with `any` parameters.

I think we should hold off on adding the `contains()` and `doesNotContain()` methods until we have some way to model iterator / iterable types - see https://github.com/winglang/wing/issues/4036

In the meantime it should be possible to work around this by using another assertion:

```js
// main.w
bring expect;

let my_array = [3, 4, 5];
expect.ok(my_array.contains(6), "missing element 6");
```

Chriscbr (changes requested):

examples/tests/sdk_tests/expect/assert.test.w
```diff
@@ -169,6 +175,14 @@ let myClass = new MyClass(1, "hello");
 expect.equal(myClass, myClass);
 expect.notEqual(myClass, new MyClass(1, "hello world") as "yet another my class");
 
+test "fail test" {
+   try {
+    expect.fail("This is a failure");
+   }catch err{
+    expect.equal(err, "This is a failure");
+   }
+}
```
Chriscbr, line 184: Can we add some failing cases for the other new methods (checking that an error is thrown when `expect.match` fails or `expect.ok` fails)?

Chriscbr (commented):
LGTM! Just requested one small addition to the tests

examples/tests/sdk_tests/expect/assert.test.w
srinumadhavv: @Chriscbr Addressed the comment

Chriscbr (approved):

Issue #4596, opened 2023-10-19
Support for rename refactoring in VSCode


eladb:
### Feature Spec

The Wing VSCode extension now supports renaming symbols through the standard refactoring mechanism. Yey!

### Use Cases

Refactoring

### Component

IDE Extension

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

Lancear:
Sounds like a fun challenge! Would be interested at giving it a go :D

staycoolcall911:
Nice! I assigned the issue to you.
Feel free though to reach out with questions to the master of anything vscode related (among other things) @MarkMcCulloh

Lancear:
Heck yeah! Will ask when I need something!

Time to read through the lsp code :D

Lancear:
Back working on it, got the base case working. Debugging is a bit tricky with the stdin being used to communicate between the lsp server and client.

Lancear:
Noticed that 10s is seen as a reference as well due to the Duration.fromSeconds that is called behind the scenes 😅 
![image](https://github.com/winglang/wing/assets/52662014/7c9c3625-4bce-4f4d-87fd-2263a86ad1b8)

monadabot:
Congrats! :rocket: This was released in [Wing 0.73.46](https://github.com/winglang/wing/releases/tag/v0.73.46).

PR #6461, opened 2024-05-12
feat(vscode): rename refactor

tsuf239:
## Checklist
fixes: [#4596](https://github.com/winglang/wing/issues/4596)
https://www.loom.com/share/234a65f04ecb471b84621aa0cf27f66c

Created a new `rename_visitor` that has a list of symbol declarations and the references to them. it adds a symbol either as a reference or as a declaration during the visit.
when executing `create_text_edits` it chooses the symbol declaration and references the position belongs to and creates textEdits with the new_text passed by the LSP.
I also updated the `get_ranges` method to get all ranges in a certain line (and not only the first one)

Currently, the renaming of class properties and namespaces is unsupported:
- namespaces need a "customized" edit: `bring ex;` with `externals` as the new word won't become `bring externals;` but `bring ex as externals;`.
- it seems like properties aren't declared/defined as the other symbols- so the current procedure isn't working for them.

- [x] Tests added (always)


eladb:
Please include a detailed description with any important implementation notes

eladb (commented):
LGTM great job!

libs/wingc/src/lsp/rename_visitor.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::diagnostic::WingLocation;
+use crate::parser::RESERVED_WORDS;
+use crate::type_check::symbol_env::LookupResult;
+use crate::type_check::{SymbolKind, Types};
+use crate::visit::{visit_scope, Visit};
+use crate::visit_context::{VisitContext, VisitorWithContext};
+use crate::{ast::*, visit_context};
+pub struct LinkedSymbol<'a> {
+	symbol: Symbol,
+	references: Mutex<Vec<&'a Symbol>>,
```
MarkMcCulloh, line 14: This shouldn't be needed. All cases where we need to mutate the vector allows us to get a mutable reference to this value.

I wasn't sure about this so I had to try it locally, so hopefully it's okay I just push up the changes 🙏

libs/wingc/src/lsp/rename_visitor.rs
```diff
@@ -0,0 +1,167 @@
...
+			return;
+		}
+		if let Some(env) = self.ctx.current_env() {
+			match env.lookup_ext(symbol, None) {
+				LookupResult::Found(symbol_kind, lookup_info) | LookupResult::NotPublic(symbol_kind, lookup_info) => {
+					// TODO: remove to support rename-refactor of namespaces - after adjusting the edit
+					if matches!(symbol_kind, &SymbolKind::Namespace(_)) {
+						// as their rename include " as ___"
+						return;
+					}
```
MarkMcCulloh, line 52: We need to handle cases where the span points to an "invisible" span. This is mostly relevant for globals since they aren't necessarily defined in the current file, so you should not be able to rename them. Probably worth adding a test to make sure you can't rename a global.

```suggestion
					if matches!(symbol_kind, &SymbolKind::Namespace(_)) {
						// as their rename include " as ___"
						return;
					}
					if lookup_info.span.is_default() {
					  // This symbol is either generated or otherwise can't be renamed (like a global)
						return;
					}
```

libs/wingc/src/lsp/rename_request.rs
```diff
@@ -0,0 +1,422 @@
...
+    let buckets = [new cloud.Bucket()];
+    for bucket in buckets {
+      //------
+      bucket.addObject("bucket","bucket");
+    //------
+      log(bucket.toString());
+        //-----^
+    }
+		"#,
+		"t1"
```
MarkMcCulloh, line 305: Noticed this is the same in every test so it could be removed and just use a default value for now

libs/wingc/src/lsp/rename_request.rs
```diff
@@ -0,0 +1,422 @@
...
+	test_rename_request!(
+		variable_same_scope,
+		r#"
+let thing = "thing";
+  //-----
+let thing2 = thing;
+           //----^
+		"#,
+		"t1"
+	);
```
MarkMcCulloh, line 138: I love the tests!

libs/wingc/src/lsp/rename_visitor.rs
MarkMcCulloh: With this you probably don't need to check the `RESERVED_WORDS` anymore

MarkMcCulloh (approved):

libs/wingc/src/lsp/rename_visitor.rs
MarkMcCulloh: Like https://github.com/winglang/wing/pull/6461#discussion_r1598869383 I tested this locally but accidentally pushed the commit. Feel free to change/add stuff to my changes!

libs/wingc/src/lsp/rename_visitor.rs
tsuf239: sure! I tested that as well, thanks for the fix 🙏

libs/wingc/src/lsp/rename_request.rs
tsuf239: we're still checking it during the test :) 
I can of course change it to a constant within the test function, but as soon as we fix the namespace case- we'll have to add it again for the tests

libs/wingc/src/lsp/rename_visitor.rs
tsuf239: I tested it again and it seems great, thanks for pushing the commit!

Issue #6471, opened 2024-05-14
Use `sim.Resource` for simulator implementations of Wing SDK resources


Chriscbr:
- [ ] [Api](https://www.winglang.io/docs/standard-library/cloud/api)
- [ ] [Bucket](https://www.winglang.io/docs/standard-library/cloud/bucket)
- [x] [Counter](https://www.winglang.io/docs/standard-library/cloud/counter)
- [ ] [Domain](https://www.winglang.io/docs/standard-library/cloud/domain)
- [ ] [Endpoint](https://www.winglang.io/docs/standard-library/cloud/endpoint)
- [ ] [Function](https://www.winglang.io/docs/standard-library/cloud/function)
- [ ] [OnDeploy](https://www.winglang.io/docs/standard-library/cloud/on-deploy)
- [ ] [Queue](https://www.winglang.io/docs/standard-library/cloud/queue)
- [ ] [Schedule](https://www.winglang.io/docs/standard-library/cloud/schedule)
- [ ] [Secret](https://www.winglang.io/docs/standard-library/cloud/secret)
- [ ] [Service](https://www.winglang.io/docs/standard-library/cloud/service)
- [ ] [Topic](https://www.winglang.io/docs/standard-library/cloud/topic)
- [ ] [Website](https://www.winglang.io/docs/standard-library/cloud/website)

Chriscbr:
See the changes to `libs/wingsdk/src/target-sim/counter.inflight.ts` and `libs/wingsdk/src/target-sim/counter.ts` in https://github.com/winglang/wing/pull/6389 for an example.

PR #6490, opened 2024-05-15
fix(compiler): cannot lift a class that extends a cloud resource

### LINKED ISSUE #4203: Cannot lift a class which extends a cloud resource
### I tried this:

Some class extending a cloud resource which is used inflight:
```ts
bring cloud;

class ExtendingBucket extends cloud.Bucket {
  pub inflight put(name: str, data: str) {
    super.put(name, data);
  }
}

let b = new ExtendingBucket();

  b.put("hello.txt", "world");
});
```

### This happened:

An error is emitted:
```sh
ERROR: Unable to lift object of type Function

target/test/bla.main.wsim.706948.tmp/.wing/preflight.js:16
           return `
             require("./inflight.ExtendingBucket-1.js")({
>>             $cloud_Bucket: ${context._lift(cloud.Bucket)},
             })
           `;
```

### I expected this:

No error.

### Is there a workaround?

Can't think of one, but maybe there is...

### Component

Compiler

### Wing Version

0.31.1

### Node.js Version

18.15

### Platform(s)

MacOS

### Anything else?

_No response_

### Community Notes

### LINKED ISSUE #6093: Wing Platform overrides do not work on extended classes
### I tried this:

when extending in non-entrypoint files

// main.w
```js
bring "./lib.w" as lib;

let b = new lib.CustomBucket();
```

// lib.w
```js
bring cloud;

pub class CustomBucket extends cloud.Bucket {}
```

Produces preflight code like
```js
class CustomBucket extends (this.node.root.typeForFqn("@winglang/sdk.cloud.Bucket") ?? cloud.Bucket) {
```

The issue is that within this non-entrypoint preflight code, the class `this.node.root` happens at the top level where there is no `this`. 

We need to somehow provide the `App` or `PlatformManager` to non entrypoint files so they can call `app.node.root` 

### This happened:

resource instantiating does not go through proper polycon hook process.

### I expected this:

_No response_

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

### LINKED ISSUE #6851: Unable to extend `cloud.` classes in a module
### I tried this:

`mod.w`

```js
bring cloud;

pub struct BucketProps extends cloud.BucketProps {

}

pub class Bucket extends cloud.Bucket {
  new(props: BucketProps) {
    super(props);
  }

  pub inflight put(key: str, value: str) {
    log("put file please");
  }

  pub inflight get(key: str): str {
    return "return";
  }
}
```

`main.w`

```js
bring "./mod.w" as m;

let b = new m.Bucket();

test "put object" {
  b.put("hello", "world");
}
```

### This happened:

```
✘ [ERROR] Could not resolve "/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/@winglang/sdk/cloud" (originally "@winglang/sdk/cloud")

    target/test/main.wsim/.wing/handler_c879d6ca.sandbox.cjs:12:31:
      12 │         $cloud_Bucket: require("@winglang/sdk/cloud").Bucket,
         ╵                                ~~~~~~~~~~~~~~~~~~~~~

  The path "@winglang/sdk/cloud" was remapped to
  "/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/@winglang/sdk/cloud" using the
  alias feature, which then couldn't be resolved. Keep in mind that import path aliases are resolved
  in the current working directory.

node:internal/process/promises:391
    triggerUncaughtException(err, true /* fromPromise */);
    ^

Error: Build failed with 1 error:
target/test/main.wsim/.wing/handler_c879d6ca.sandbox.cjs:12:31: ERROR: Could not resolve "/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/@winglang/sdk/cloud" (originally "@winglang/sdk/cloud")
    at failureErrorWithLog (/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:1651:15)
    at /Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:1059:25
    at /Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:1004:52
    at buildResponseToResult (/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:1057:7)
    at /Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:1086:16
    at responseCallbacks.<computed> (/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:704:9)
    at handleIncomingPacket (/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:764:9)
    at Socket.readFromStdout (/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/esbuild/lib/main.js:680:7)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12) {
  errors: [
    {
      detail: undefined,
      id: '',
      location: {
        column: 31,
        file: 'target/test/main.wsim/.wing/handler_c879d6ca.sandbox.cjs',
        length: 21,
        line: 12,
        lineText: '        $cloud_Bucket: require("@winglang/sdk/cloud").Bucket,',
        namespace: '',
        suggestion: ''
      },
      notes: [
        {
          location: null,
          text: `The path "@winglang/sdk/cloud" was remapped to "/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/@winglang/sdk/cloud" using the alias feature, which then couldn't be resolved. Keep in mind that import path aliases are resolved in the current working directory.`
        }
      ],
      pluginName: '',
      text: 'Could not resolve "/Users/eladb/.npm-global/lib/node_modules/winglang/node_modules/@winglang/sdk/cloud" (originally "@winglang/sdk/cloud")'
    }
  ],
  warnings: []
}

Node.js v20.15.0
```

### I expected this:

To work

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

Chriscbr:
This PR aims to fix an issue where unexpected bugs happen when you try extending a class from the Wing SDK, like `cloud.Bucket`. The root cause is related to how Wings' inheritance is implemented in JavaScript.

Each cloud resource has a JavaScript class modeling its inflight API. The class needs to be constructed with target-specific arguments. For example, for the implementation of `cloud.Bucket`, there's a class definition `BucketClient` for each of the supported cloud targets (`tf-aws`, `tf-gcp`, and `tf-azure`). What's important to note is that each class needs to be instantiated at runtime with value that are specific to that cloud target. For example, `new aws.BucketClient(...)` expects an S3 Bucket name, while `new azure.BucketClient(...)` expects an Azure container name _and_ an Azure storage account name.

How does Wing ensure the right values are passed when the class is constructed? Currently, each preflight class has an `_toInflight()` method which returns a string of JavaScript code. That code instantiates the class and hardcodes in the necessary arguments. This method is code-generated by the compiler for most Wing classes.

But suppose we have a class `MyBucket` that extends `Bucket`. To instantiate such a class, we'd need to instantiate `MyBucket` with both the arguments needed by `MyBucket` AND the arguments needed by `Bucket`, and ensure that the `super()` call in `MyBucket`'s constructor passes the appropriate arguments to `Bucket`'s constructor. 

To support this, we're now exposing a new method in the TypeScript source code named `_liftedState()` which can be overridden by classes inside our TypeScript SDK implementation to specify what arguments they need when they're instantiated. When `_toInflight()` is called, it will automatically call `_liftedState()` to obtain the constructor arguments _and_ instantiate the class. But if you just call `_toInflightType()`, the class will be returned without instantiating an instance of it. These internal changes should not have any user-facing effects.

Fixes #4203
Fixes #6093
Fixes #6851

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/target-tf-aws/function.ts
```diff
@@ -319,10 +319,19 @@ export class Function extends cloud.Function implements IAwsFunction {
...
       __filename,
       "FunctionClient",
-      [`process.env["${this.envName()}"], "${this.node.path}"`]
+      undefined,
+      this._liftedFields()
     );
   }
 
+  /** @internal */
+  public _liftedFields(): Record<string, string> {
```
Chriscbr, line 328: This API name honestly doesn't make sense, maybe it should be `_inflightNewArgs()` or something like that

libs/wingsdk/src/target-tf-aws/counter.ts
Chriscbr, line 1: Probably need to hand-apply changes like this to all of the cloud resources in the SDK (for all cloud targets)

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -25,6 +25,16 @@ const NAME_OPTS: NameOptions = {
...
 export class Counter extends cloud.Counter implements IAwsCounter {
+  /** @internal */
+  public static _toInflightType(): string {
+    return core.InflightClient.forType(
+      __filename
+        .replace("target-tf-aws", "shared-aws")
+        .replace("counter", "counter.inflight"),
+      "CounterClient"
+    );
+  }
```
Chriscbr, line 36: In theory the function "toLiftableModuleType" should handle this / allow us to avoid having to implement _toInflightType everywhere, but I think the compiler is generating bad require paths or something like that. Needs more investigation...

libs/wingsdk/src/core/inflight.ts
```diff
@@ -44,13 +44,24 @@ export class InflightClient {
...
     clientClass: string,
-    args: string[]
+    args: string[] | undefined,
+    liftedFields?: Record<string, string>
   ): string {
     const inflightDir = dirname;
     const inflightFile = basename(filename).split(".")[0] + ".inflight";
+    let argsStr = "";
+    if (args !== undefined) {
+      argsStr = args.join(", ");
```
yoav-steinberg, line 54: can you explain when would we have `args` vs when we'll have `liftedFields`? Are these mutually exclusive?

libs/wingsdk/src/target-sim/counter.ts
```diff
@@ -45,6 +60,14 @@ export class Counter extends cloud.Counter {
     };
   }
 
+  /** @internal */
+  public _liftedFields(): Record<string, string> {
+    const env = makeEnvVarName("resource", this.backend);
+    return {
+      $handle: `process.env["${env}"]`,
+    };
+  }
```
yoav-steinberg, line 69: Guessing this can be shared by basically all sim resources.

libs/wingsdk/src/target-tf-aws/counter.ts
```diff
@@ -64,12 +74,15 @@ export class Counter extends cloud.Counter implements IAwsCounter {
 
   /** @internal */
   public _toInflight(): string {
-    return core.InflightClient.for(
-      __dirname.replace("target-tf-aws", "shared-aws"),
-      __filename,
-      "CounterClient",
-      [`process.env["${this.envName()}"]`, `${this.initial}`]
-    );
+    return core.InflightClient.forV2(Counter, this._liftedFields());
```
yoav-steinberg, line 77: can probably get rid of this in favor of `std.Resource._toInflight()`

libs/wingsdk/src/target-tf-aws/function.ts
yoav-steinberg: The thinking was: to create a client for resource r of type R you need some information from the preflight object r. This information is the unique information that isn't shared between all instances of R. Such information is typically (or always) stored in preflight fields of R (the env, ARN, special configs, etc).
In other words, what are the **fields** that are being lifted for all types R so they can be used inflight. These fields need to be lifted and stored (or used to initialized other inflight fields) in clients of R.

I feel `_inflightNewArgs` is missing what this essentially is, maybe `_liftedState()` makes more sense. We need this state to create a specific client for r.

libs/wingc/src/jsify.rs
```diff
@@ -1748,39 +1749,32 @@ impl<'a> JSifier<'a> {
...
+		code.close("};");
+		code.close("}");
+		code
+	}
 
-		code.line(format!(
-			"if (client.{CLASS_INFLIGHT_INIT_NAME}) {{ await client.{CLASS_INFLIGHT_INIT_NAME}(); }}"
-		));
-		code.line("return client;");
+	fn jsify_to_inflight_method(&self, class_name: &Symbol) -> CodeMaker {
```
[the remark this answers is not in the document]
Chriscbr: It might be possible through some JavaScript tricks (`this.constructor`), I'll give it a try

libs/wingsdk/src/target-sim/counter.ts
Chriscbr: Eventually - though for now the implementations may look different until we've closed https://github.com/winglang/wing/issues/6471

libs/wingsdk/src/target-tf-aws/function.ts
Chriscbr: I see, yeah `_liftedState` may be better

libs/wingc/src/jsify.rs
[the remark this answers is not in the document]
Chriscbr: Removed

libs/wingsdk/src/target-tf-aws/function.ts
Chriscbr: Updated

libs/wingsdk/src/target-sim/counter.ts
Chriscbr: It looks like this was correct (I copied the code into a function named "simulatorLiftedFieldsFor")

libs/wingsdk/src/core/inflight.ts
Chriscbr: I removed this method (supporting both params was just an interim fix so I could try to validate the initial solution for the PR)

libs/wingc/src/jsify.rs
```diff
@@ -1930,40 +1930,20 @@ impl<'a> JSifier<'a> {
 		res
 	}
 
-	fn jsify_to_inflight_method(&self, class_name: &Symbol, class_type: TypeRef) -> CodeMaker {
+	fn jsify_lifted_fields(&self, class_name: &Symbol, class_type: TypeRef) -> CodeMaker {
```
yoav-steinberg, line 1933: lets call this `jsify_listed_state`

yoav-steinberg:
excited about this one :) nice work.

Chriscbr:
It looks like the approach in this PR is missing one or two pieces -- it adds some code that calls "typeForFqn" to obtain the target-specific versions of classes. But this "typeForFqn" function isn't accessible from some places in preflight JavaScript like static methods because typeForFqn is currently associated with the app. (And, the app requires a root construct to be passed to it, whose code might require "typeForFqn"...). I opened a PR a while back for trying to unwind this cycle here (https://github.com/winglang/wing/pull/6099) so I'm going to try and see if I can bring it back to somewhere shippable...

libs/wingc/src/jsify/snapshots/allow_type_def_before_super.snap
```diff
@@ -125,16 +120,8 @@ class $Root extends $stdlib.std.Resource {
...
-                const BazClient = ${Baz._toInflightType()};
-                const client = new BazClient({
-                });
-                if (client.$inflight_init) { await client.$inflight_init(); }
-                return client;
-              })())
-            `;
+          _liftedState() {
+            return { ...(super._liftedState?.() ?? {}) };
           }
```
eladb, line 125: Since this is a no-op maybe we can omit this method if there is no state to lift...

libs/wingc/src/jsify/snapshots/base_class_captures_inflight.snap
```diff
@@ -48,8 +49,9 @@ const $helpers = require("@winglang/sdk/lib/helpers");
 const $macros = require("@winglang/sdk/lib/macros");
 module.exports = function({ $Base }) {
   class Derived extends $Base {
-    constructor({  }) {
-      super({  });
+    constructor($args) {
+      const {  } = $args;
+      super($args);
     }
```
eladb, line 55: Maybe we can omit this constructor if there is no state to lift

libs/awscdk/src/api.ts
```diff
@@ -229,10 +237,10 @@ export class Api extends cloud.Api implements IAwsApi {
   }
 
   /** @internal */
-  public _toInflight(): string {
-    return core.InflightClient.for(__dirname, __filename, "ApiClient", [
-      `process.env["${this.urlEnvName()}"]`,
-    ]);
+  public _liftedState(): Record<string, string> {
+    return {
+      $url: `process.env["${this.urlEnvName()}"]`,
```
eladb, line 242: Why do we need the $ here?

eladb (approved):

libs/wingc/src/jsify/snapshots/allow_type_def_before_super.snap
Chriscbr: Sure, yeah a little bit less code shouldn't hurt

(I'd honestly like it this to be hard contract / it's easier debugging our system when code fails fast when some contract isn't satisfied -- unfortunately this isn't an option here because we allow you to extend JSII classes, which don't have `_liftedState` methods.)

libs/awscdk/src/api.ts
Chriscbr: Hmm, yeah in retrospect the dollar signs aren't necessary. I think I did this originally just to mentally match up the output between what the compiler's generating for Wing source code and what's in the SDK while I was designing the solution, but in practice naming doesn't really matter.

libs/wingc/src/jsify/snapshots/base_class_captures_inflight.snap
Chriscbr: Done

examples/tests/valid/extend_counter.test.w
```diff
@@ -0,0 +1,49 @@
...
+    expect.equal(this.field2, 10);
+    expect.equal(this.field1, 5);
+    expect.equal(global_value, "yo");
+    super.inc();
+  }
+
+  pub inflight extra2(): str {
+    return "extra2";
+  }
+}
```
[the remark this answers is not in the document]
Chriscbr: This is a good callout, I need to think about this case some more...

libs/wingsdk/src/target-tf-aws/website.ts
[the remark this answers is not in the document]
Chriscbr: It should be OK here since `cloud.Website` doesn't have any inflight methods

examples/tests/valid/extend_counter.test.w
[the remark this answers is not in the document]
Chriscbr: I think overriding `inflight new()` can work, the only limitation might be that if you're extending a class in the SDK that's written in TypeScript, you won't be able to access fields defined in `cloud.Counter`'s inflight constructor.

Issue #4852, opened 2023-11-09
No signature help for `super()` calls


eladb:
### I tried this:

```js

class Base {
  init(hello: num) {}
}

class Derived extends Base {
  init() {
    super()
//        ^ no completion!
  }
}

new Derived();
```

### This happened:

No completion for the function call

### I expected this:

To see completions

### Component

IDE Extension

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

MarkMcCulloh:
Completions do show up (but need to be manually triggered)

<img width="689" alt="image" src="https://github.com/winglang/wing/assets/1237390/421ed7cb-9546-44fb-b196-f941fc41ec3a">

There is no signature help though, is that the issue?

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.4](https://github.com/winglang/wing/releases/tag/v0.74.4).

PR #6492, opened 2024-05-16
fix(vscode):  No signature help for `super()` calls

tsuf239:
fixes [#4852](https://github.com/winglang/wing/issues/4852)

I added another variable to catch the `StmtKind::SuperConstructor` we're on (and another one to catch the class scope), then I changed the `sig_data` to be initialized based on the new variable if exists.

image
image

## Checklist

- [x] Tests added (always)


libs/wingc/src/lsp/signature.rs
```diff
@@ -32,41 +34,70 @@ pub fn on_signature_help(params: lsp_types::SignatureHelpParams) -> Option<Signa
...
+								&Symbol {
+									name: if t.is_preflight_class() {
+										CLASS_INIT_NAME.into()
+									} else {
+										CLASS_INFLIGHT_INIT_NAME.into()
+									},
+									span: Default::default(),
+								},
+								None,
+							);
```
MarkMcCulloh, line 59: Can use `into()` to automatically create a Symbol here

```suggestion
							let init_lookup = t.as_class()?.env.lookup(
								&if t.is_preflight_class() {
										CLASS_INIT_NAME
									} else {
										CLASS_INFLIGHT_INIT_NAME
									}.into(),
								None,
							);
```

libs/wingc/src/lsp/signature.rs
```diff
@@ -177,10 +208,14 @@ pub struct ScopeVisitor<'a> {
...
 	/// The nearest expression before (or containing) the target location
 	pub call_expr: Option<&'a Expr>,
+	/// The nearest statement before (or containing) the target location - on a case we're on a super() call
+	pub call_stmt: Option<&'a Stmt>,
 	// The env of the found expression
 	pub call_env: Option<SymbolEnvRef>,
 	/// The current symbol env we're in
 	curr_env: Vec<SymbolEnvRef>,
+	/// the nearest class containing the call_stmt
+	pub class: Option<&'a Class>,
```
MarkMcCulloh, line 218: It should be possible to avoid storing this, the `call_env` will be a SymbolEnv whose parent will be the class environment.

So you should be able to use `call_env?.parent?.lookup`

MarkMcCulloh (approved):

libs/wingc/src/lsp/signature.rs
tsuf239: I tried it with call_env being assigned when the call_stmt is assigned and when the class is assigned, but none of them worked unfortunately :(

Issue #6500, opened 2024-05-16
simulator error when calling an inflight closure that returns a bucket


yoav-steinberg:
### I tried this:

```wing
bring cloud;

let my = new cloud.Bucket();

let foo = inflight (): cloud.Bucket => {
  return my;
};

test "test" {
  foo();
}
```

### This happened:

```
fail ┌ main.wsim » root/env0/test:test
     └ Error: Resource "root/env0/test:test/Handler" does not have permission to perform operation "then" on resource "root/env0/Bucket".
```

### I expected this:

to work

### Wing Version

0.73.47

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

yoav-steinberg:
Works on `tf-aws` platform

PR #6504, opened 2024-05-16
fix(compiler): incorrect lift qualification when accessing a property on a return value of an inflight closure 

### LINKED ISSUE #6501: incorrect lift qualification when acessing property on return value of inflight closure
### I tried this:

```wing
bring cloud;

let my = new cloud.Bucket();

inflight class Foo {
  pub do() {}
}

let foo = inflight (): Foo => {
  return new Foo();
};

test "test" {
  foo().do();
}
```

### This happened:

```
Error: Resource root/env0/$Closure1_0 does not support inflight operation do.
It might not be implemented yet.
```

### I expected this:

to run

### Is there a workaround?

We can assign the inflight closure to a temp variable and then call that temp variable. This along with an explicit `lift()` call works around the problem:
```
bring cloud;

let my = new cloud.Bucket();

inflight class Foo {
  pub do() {}
}

let foo = inflight (): Foo => {
  return new Foo();
};

test "test" {
  lift(foo,[]);
  let x = foo;
  x().do();
}
```

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

yoav-steinberg:
Fixes #6501

The lift visitor's context now adds information matching the current property with the actual object this property belongs to. This way when qualifying an expression we can verify the expression really belongs to the current property. The result is that something like `(expr_a + expr_b).do()` will qualify only the outer expression with `do()` (if it's a preflight expression). If only the inner expressions are preflight then they will be qualified without any property. This is more correct and doesn't assume any property needs to be qualified when we encounter a preflight expression during the lift visitor.

Specifically this helps us when `some_inflight_closure().do()` the `some_inflight_closure()` call expression isn't a preflight expression so it isn't lifted, but the `some_inflight_closure` part (without the `()` call) is lifted but without the `do()` qualification and defaults to the `handle` qualification which is what we want.

## Checklist

- [x] Tests added (always)


examples/tests/valid/lift_inflight_closure_returning_object_issue6501.test.w
```diff
@@ -0,0 +1,17 @@
+// Some inflight object with a method
+inflight class Foo {
+  pub do():num {
+    return 1337;
+  }
+}
+
+let foo = inflight (): Foo => {
```
eladb, line 8: Can you add a test similar to the bug report?

(Where the returned type is a preflight class)

eladb (approved):

examples/tests/valid/lift_inflight_closure_returning_object_issue6501.test.w
yoav-steinberg: I added a commented out test because it fail (on sim, not on aws) due to #6500
@Chriscbr any idea?

Chriscbr: I pushed a change with a fix. It's sort of a hack to deal with the fact that we're using these proxy objects that implicitly forward any method call like `x.blah()` into an HTTP call. When you do `await x` in JavaScript, if you don't have a Promise object, then for backwards compatibility JavaScript will try calling a method named "then", hence the errors.

yoav-steinberg: Excellent, thanks!

PR #6508, opened 2024-05-16
feat(sdk): custom inflight hosts for aws resources

eladb:
Previously, AWS resources could only be lifted into AWS Lambda functions. This changes allows any object that implements the `IAwsInflightHost` interface to be used as a host.

Misc: added `std.Resource.toInflight(obj)` which can be used to call the internal `_toInflight()` implementation of a preflight object.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-tf-aws/topic.ts
```diff
@@ -168,11 +169,9 @@ export class Topic extends cloud.Topic implements IAwsTopic {
...
 
   public onLift(host: IInflightHost, ops: string[]): void {
-    if (!(host instanceof Function)) {
-      throw new Error("topics can only be bound by tfaws.Function for now");
-    }
-
-    host.addPolicyStatements(...calculateTopicPermissions(this.topic.arn, ops));
+    AwsInflightHost.from(host)?.addPolicyStatements(
+      ...calculateTopicPermissions(this.topic.arn, ops)
+    );
```
Chriscbr, line 174: I wonder if it will be an issue if you try using e.g. an AWS bucket from a GCP function and the user only gets permission errors at deploy time. Something to keep in the back of our minds for later perhaps

Chriscbr (approved):

libs/wingsdk/src/target-tf-aws/topic.ts
eladb: Mmmm.. Good point. Perhaps `AwsInflightHost.from()` should not throw if the host is not compatible.

eladb: More defensive.

eladb: Fixed.

libs/wingc/src/type_check/jsii_importer.rs
```diff
@@ -864,10 +864,8 @@ impl<'a> JsiiImporter<'a> {
 	fn parameter_to_wing_type(&mut self, parameter: &jsii::Parameter) -> TypeRef {
 		let mut param_type = self.type_ref_to_wing_type(&parameter.type_);
 
-		// TODO variadic parameter support https://github.com/winglang/wing/issues/397
 		if parameter.variadic.unwrap_or(false) {
 			param_type = self.wing_types.add_type(Type::Array(param_type));
-			param_type = self.wing_types.add_type(Type::Optional(param_type));
```
eladb, line 870: This caused all variadic parameters to be optional, which is not a good look.

examples/tests/valid/external_ts.extern.d.ts
```diff
@@ -51,7 +51,7 @@ export interface MetadataEntry {
 export class Node {
   /** Add an ordering dependency on another construct.
   An `IDependable` */
-  readonly addDependency: (deps?: ((readonly (IDependable)[])) | undefined) => void;
+  readonly addDependency: (deps: (readonly (IDependable)[])) => void;
```
Chriscbr, line 54: Should it generate something like this?
```suggestion
  readonly addDependency: (...deps: (readonly (IDependable)[])) => void;
```

The compiler's `FunctionParameter` struct has a `variadic` boolean on it which can be used for this I think

eladb: Ooo love latent bugs

MarkMcCulloh: Created https://github.com/winglang/wing/issues/6512 since it's an existing issue.

The downside of boolean flags instead of an enum to force handling of different cases

eladb: Thanks @MarkMcCulloh

PR #6521, opened 2024-05-20
feat(compiler): support logging stringable values directly

Chriscbr:
Adds support for directly logging stringable values without needing to create a interpolated string:

```
let x = 5;

// before
log("{x}");

// now
log(x);
```

To implement the feature, I had to add a new type to the compiler's type system for representing stringable values. At the moment it more or less behaves like `str | num | bool | ...`. For now it's not really something we have concrete plans to expose in userland (to let users write functions with stringable inputs) but it's something we can consider introducing in the future.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/apps/vscode-wing/syntaxes/wing.tmLanguage.json
+++ b/apps/vscode-wing/syntaxes/wing.tmLanguage.json
@@ -160,7 +160,7 @@
         },
         {
           "name": "support.class.wing",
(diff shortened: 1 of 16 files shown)


eladb (approved):
LGTM!

Not directly related to this PR - can we make an `std.IStringable` interface (with a `toString()` method)  so that it will be possible for any type to be stringable?

Issue #6525, opened 2024-05-21
api gateway UI loose the endpoint while editing body


ekeren:
### I tried this:

Notice that the "Send" Button gets disabled because the URL gets unselected 

Uploading bug.mov…

### This happened:

see videm

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.3](https://github.com/winglang/wing/releases/tag/v0.74.3).

PR #6527, opened 2024-05-21
fix(console): api gateway ui loose the endpoint while editing body

polamoros:
Fixes https://github.com/winglang/wing/issues/6525

### DIFF
--- a/apps/wing-console/console/ui/src/ui/api-interaction.tsx
+++ b/apps/wing-console/console/ui/src/ui/api-interaction.tsx
@@ -316,28 +316,10 @@ export const ApiInteraction = memo(
       ],
     );
 
-    const loadRoutesFromOpenApi = useCallback(() => {
-      const routes = getRoutesFromOpenApi(openApiSpec);
-      setRoutes(routes);
-      const currentRouteExists = routes.some(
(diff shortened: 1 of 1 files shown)


apps/wing-console/console/ui/src/ui/api-interaction.tsx
```diff
@@ -316,28 +316,10 @@ export const ApiInteraction = memo(
       ],
     );
 
-    const loadRoutesFromOpenApi = useCallback(() => {
```
ainvoner, line 319: it handled the use case when the route was removed from the code while it is selected in the UI.
after your change, the routes are getting updated but the routes string still remains visible

ainvoner (changes requested):

apps/wing-console/console/ui/src/ui/api-interaction.tsx
polamoros: The route is being removed from the dropdown, but I wouldn't remove it from the input. 

For example, if the user wrote a custom route `/route-that-never-existed`, we would clean it up as well.

ainvoner (approved):

PR #6536, opened 2024-05-22
feat(sdk): support api props

eladcon:
It's now possible to set function props such as env vars for api handlers

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/examples/tests/sdk_tests/api/env.test.w
+++ b/examples/tests/sdk_tests/api/env.test.w
@@ -3,7 +3,6 @@ bring expect;
 bring http;
 bring util;
 
-
 let api = new cloud.Api();
 
 api.get("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
@@ -13,13 +12,10 @@ api.get("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
     status: 200,
     body: foo
   };
-}, { env: { FOO: "bar" } });
+}, env: { FOO: "bar" });
 
 test "can access env vars in handler" {
     let url = api.url + "/path";
     let res = http.get(url);
     expect.equal(res.body, "bar");
 }
-
-
-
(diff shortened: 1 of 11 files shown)


Chriscbr (approved):

examples/tests/sdk_tests/api/env.test.w
```diff
@@ -0,0 +1,25 @@
...
+let api = new cloud.Api();
+
+api.get("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
+  let foo = util.env("FOO");
+
+  return cloud.ApiResponse {
+    status: 200,
+    body: foo
+  };
+}, { env: { FOO: "bar" } });
```
eladb, line 16:
```suggestion
}, env: { FOO: "bar" });
```

I believe you don't need the `{}`, correct?

libs/awscdk/src/api.ts
```diff
@@ -49,17 +49,14 @@ export class Api extends cloud.Api implements IAwsApi {
     method: string,
     path: string,
     inflight: cloud.IApiEndpointHandler,
-    props?: cloud.ApiGetOptions
+    props?: cloud.FunctionProps,
```
eladb, line 52: I think this should still be `ApiGetOptions`, but it should extend `FunctionProps`. Otherwise, if we want to add any options it will be a breaking change.

libs/wingsdk/src/cloud/api.ts
```diff
@@ -587,42 +588,42 @@ export class Api extends Resource {
 /**
  * Options for Api get endpoint.
  */
-export interface ApiGetOptions {}
```
eladb, line 590: Perhaps introduce: `ApiEndpointOptions`:

```js
struct ApiEndpointOptions extends FunctionProps {}

struct ApiGetOptions extends ApiEndpointOptions {}
struct ApiPostOptions extends ApiEndpointOptions {}
// ...
```

eladb (approved):

Issue #5943, opened 2024-03-14
`cloud.Api`: misleading error message for endpoint path


meirdev:
### I tried this:

```
bring cloud;

let api = new cloud.Api();

api.get("/:id/comments", inflight (req) => {
});

```

### This happened:

```
An error has occurred:
Failed to compile.

Error: Invalid path /:id/comments. Url parts can only contain alpha-numeric chars, "-", "_" and ".". Params can only contain alpha-numeric chars and "_".
```

### Anything else?

All parts of the URL match the requirement. the regex can be updated to validate the URL but the real problem is that the first part of the URL is a variable. (see here https://github.com/winglang/wing/issues/5210#issuecomment-1987192944)

### Wing Version

0.59.46

### Node.js Version

v20.11.0

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.15](https://github.com/winglang/wing/releases/tag/v0.74.15).

PR #6543, opened 2024-05-23
fix(sdk): api path- collision with `/{proxy+}`

tsuf239:
fixed: 
#5943
#5210

tf-aws
image

aws-cdk
image
(the delete test passed in the next run- it was only failing because something in the env couldn't be destroyed)
image
and the cycle one doesn't work on main as well :)

## Checklist

- [ ] Tests added (always)


tsuf239:
I noticed that the regex wasn't including the "/" path- so I fixed it and realized that we don't have an api test that includes only a root path "/"- so I changed the `post.test.w` to include one. it passes both in tf-aws and the sim
image

examples/tests/sdk_tests/api/post.test.w
```diff
@@ -19,15 +20,14 @@ api.post("/path", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
...
-  let fetchResponse: http.Response = http.post(url, method: http.HttpMethod.POST, headers: { "content-type" => "application/json" }, body: Json.stringify(body));
+  let response: http.Response = http.post(api.url, headers: { "content-type" => "application/json" }, body: Json.stringify(body));
+  let fetchResponse: http.Response = http.post(api.url, method: http.HttpMethod.POST, headers: { "content-type" => "application/json" }, body: Json.stringify(body));
 
-  assert(response.body == Json.stringify(body));
-  assert(response.status == 200);
-  assert(response.url == url);
+  expect.equal(response.body , Json.stringify(body));
+  expect.equal(response.status , 200);
+  expect.match(response.url , api.url);
```
tsuf239, line 28: this is match since the response url in the sim includes an extra "/" at the end

examples/tests/sdk_tests/api/root_path_vars.test.w
```diff
@@ -0,0 +1,47 @@
...
+  expect.equal(resWithId.body, "id is 123");
+
+  let resWithComments = http.get("{api.url}/123/comments");
+  expect.ok(resWithComments.ok);
+  expect.equal(resWithComments.body, "id is 123");
+
+  let unknownRes = http.get("{api.url}/123/comments/unknown-path");
+  expect.equal(unknownRes.status, 404);
+  expect.match(unknownRes.body, "Error");
+}
```
Chriscbr, line 47: What happens in this case?
```suggestion
  let unknownRes = http.get("{api.url}/123/comments/unknown-path");
  expect.equal(unknownRes.status, 404);
  expect.match(unknownRes.body, "Error");
  
  let unknownRes2 = http.get("{api.url}/123/unknown-path");
  expect.equal(unknownRes2.status, 404);
  expect.match(unknownRes2.body, "Error");
}
```

Chriscbr (approved):

libs/wingsdk/src/cloud/api.ts
```diff
@@ -432,6 +430,40 @@ export class Api extends Resource {
...
+        partA.match(/^:.+?$/) &&
+        partB.match(/^:.+?$/) &&
+        partA[i] !== partB[i]
+      ) {
+        return true;
+      }
+    }
+
+    return false;
+  }
```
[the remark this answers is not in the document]
tsuf239: I've added some tests for the sibling paths in the `libs/wingsdk/test/target-sim/api.test.ts` - since it was already a part of the api class, and I didn't want to take it out if it's only used there- especially since we don't have any of those in the other resources, and I didn't know where to locate the test file

PR #6571, opened 2024-05-28
fix(compiler): expected type to be "Array", but got "Array" instead

### LINKED ISSUE #3822: Expected type to be "Array", but got "Array" instead
### I tried this:

```js
assert([1,2] == [{ a : 12 }])
```

### This happened:

```
error: Expected type to be "Array", but got "Array" instead
  --> wing/test.w:1:8
  |
1 | assert([1,2] == [{ a : 12 }]);
  |        ^^^^^^^^^^^^^^^^^^^^^ Expected type to be "Array", but got "Array" instead
```

### I expected this:

A better error

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

tsuf239:
Fixes #3822 
I considered a few approaches and concluded that the lightest one would be passing the parent type into the validation function, and then using it (if not None) in the diagnostic message.
Updated the validation function calls accordingly.
Currently supported only for the `Equal` and `NotEqual` binary operators

## Checklist

- [x] Tests added (always)


Chriscbr (approved):

examples/tests/invalid/equality.test.w
yoav-steinberg, line 1: * Maybe also add a test with optionals as the values
```
let a: num? = nil;
let b: str? = nil;
assert([a] == [b]);
```
* Maybe add a test with nested arrays:
```
assert([[1]] == [["1"]]);
```

yoav-steinberg, line 1: Do we have similar errors when the types are maps/sets?
```
let a = {"a" => 1};
let b = {"b" => "b"};
assert(a == b);
```
If so then we should test this too.

PR #6586, opened 2024-05-29
fix(sdk): deep equality for sets fails in Node v20

### LINKED ISSUE #4444: Deep equality for sets fails in Node v20
### I tried this:

```ts
let a = MutSet<Map<str>> {{"a" => "b", "c" => "d"}};
let b = a.copy();
assert(a == b);

let c = b.copyMut();
assert(a == c);

a.add({"e" => "f"});
c.add({"e" => "f"});
assert(a == {{"a" => "b", "c" => "d"}, {"e" => "f"}});
assert(c == {{"a" => "b", "c" => "d"}, {"e" => "f"}});
assert(a == c);
```

### This happened:

Assertion failed with the following error (it is the final assert that is causing this):
```
ERROR: assertion failed: a == c

target/test/main.wsim.354620.tmp/.wing/preflight.js:17
       {((cond) => {if (!cond) throw new Error("assertion failed: a == {{\"a\" => \"b\", \"c\" => \"d\"}, {\"e\" => \"f\"}}")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(a,new Set([({"a": "b","c": "d"}), ({"e": "f"})]))))};
       {((cond) => {if (!cond) throw new Error("assertion failed: c == {{\"a\" => \"b\", \"c\" => \"d\"}, {\"e\" => \"f\"}}")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(c,new Set([({"a": "b","c": "d"}), ({"e": "f"})]))))};
>>     {((cond) => {if (!cond) throw new Error("assertion failed: a == c")})((((a,b) => { try { return require('assert').deepStrictEqual(a,b) === undefined; } catch { return false; } })(a,c)))};
     }
   }
```

### I expected this:

For the assertions to pass without any errors.

### Is there a workaround?

Run the tests by removing the final assertion, like so:
```ts
let a = MutSet<Map<str>> {{"a" => "b", "c" => "d"}};
let b = a.copy();
assert(a == b);

let c = b.copyMut();
assert(a == c);

a.add({"e" => "f"});
c.add({"e" => "f"});
assert(a == {{"a" => "b", "c" => "d"}, {"e" => "f"}});
assert(c == {{"a" => "b", "c" => "d"}, {"e" => "f"}});
// assert(a == c);
```

Which passes as expected: ` Tasks:    12 successful, 12 total`

### Component

Compiler

### Wing Version

0.35.5

### Node.js Version

v20.8.0

### Platform(s)

Linux

### Anything else?

Originally noticed this issue because of [a CI failure](https://github.com/winglang/wing/actions/runs/6423981885/job/17444538461#step:7:336) for the PR  #4432. This behavior only happens with node v20 as seen in the pipeline and upon further testing in local environment, it is confirmed that it still passes (the snippet I provided here, as well as the one that caused the ci failure) in node v18, but fails in node v20.

### Community Notes

tsuf239:
Fixes #4444 
Replaced the native assert module import to an [external npm package](https://www.npmjs.com/package/assert) for the problematic file.

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/helpers.ts
```diff
@@ -1,9 +1,13 @@
...
 // so avoid importing anything heavy here.
-import { deepStrictEqual, notDeepStrictEqual } from "node:assert";
 import * as path from "node:path";
 import type { Construct } from "constructs";
 import type { Node } from "./std/node";
+// since we moved from node:18 to node:20 the deepStrictEqual doesn't work as expected.
+// https://github.com/winglang/wing/issues/4444
+// therefore we're using an npm package called "assert" instead.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const { deepStrictEqual, notDeepStrictEqual } = require("external-assert");
```
MarkMcCulloh, line 10: I don't think it's worthwhile to take on this new dependency just for deepStringEqual. This `helpers.ts` file is included in every inflight/preflight so this module will end up being bundled everywhere.

Instead it may be simple enough to implement/vendor the function ourselves.
Here's an example in [vitest](https://github.com/vitest-dev/vitest/blob/d6700bbd895e63776263b206ec73ccbb858a1b94/packages/expect/src/jest-utils.ts#L76) (some of it can be easily removed)

Worst case scenario it may be best to find a package that only deals with `deepStringEqual` instead of the entire `assert` module.

tsuf239: took the function stright from node 18: https://github.com/nodejs/node/blob/v18.x/lib/internal/util/comparisons.js
(and added dozens of deepEquality tests, thanks for chatgpt)

libs/wingsdk/src/helpers.ts
```diff
@@ -1,14 +1,17 @@
 // Code in this file will be automatically included in all inflight code bundles,
 // so avoid importing anything heavy here.
-import { deepStrictEqual, notDeepStrictEqual } from "node:assert";
+import { notDeepStrictEqual } from "node:assert";
 import * as path from "node:path";
 import type { Construct } from "constructs";
+// since we moved from node:18 to node:20 the deepStrictEqual doesn't work as expected.
+// https://github.com/winglang/wing/issues/4444
+// therefore we're using a local version of the comparison from node 18: https://github.com/nodejs/node/blob/v18.x/lib/internal/util/comparisons.js
```
MarkMcCulloh, line 8: This note should probably just be in `equality.ts` instead.

libs/wingsdk/src/equality.ts
MarkMcCulloh, line 1: Could this be moved to the `util` folder?

libs/wingsdk/src/equality.ts
```diff
@@ -0,0 +1,475 @@
+// taken from node 18 https://github.com/nodejs/node/blob/v18.x/lib/internal/util/comparisons.js
```
MarkMcCulloh, line 1: Since this is an older version, are there bugs here that we are relying on? I understand the newer one doesn't work, but do we know what we're missing out on explicitly?

MarkMcCulloh (approved):

libs/wingsdk/src/equality.ts
tsuf239: what they fixed was taking into account the reference to an object when comparing it- if there are two objects one is the clone of the other, and we add a key (also an object) of the same name and the same structure and different reference, it will fail 
if the parent objects have the same structure and different references - it will pass
(at least that is the impression I got while working on it)

libs/wingsdk/src/helpers.ts
tsuf239: changed it, the link appears now only in the `equality.ts` file

PR #6589, opened 2024-05-29
fix(compiler): better error message when putting in parenthesis in a catch block

### LINKED ISSUE #5945: Confusing error message when putting error in parenthesis in try-catch block
### I tried this:

```ts
try {

} catch (e) {

}
```

### This happened:

Got this error: 
> Missing `catch` or `finally` blocks for this try statement

### I expected this:

Any other message basically. The error is not that there is no catch statement, but that `catch (e)` should be `catch e`

### Is there a workaround?

_No response_

### Anything else?

Another seemingly small issue, but that users will encounter often as co-pilot generates try-catch blocks like this

### Wing Version

0.61.1

### Node.js Version

_No response_

### Platform(s)

MacOS

### Community Notes

tsuf239:
Fixes #5945 
```
try {
  log("Hello World");
} catch (e) {
//      ^^^ Unexpected parentheses in catch block. Use 'catch e' instead of 'catch (e)'.
  log("error");
}

try {
  log("Hello World");
} catch (longError) {
//      ^^^^^^^^^^^ Unexpected parentheses in catch block. Use 'catch longError' instead of 'catch (longError)'.
  log("error");
}
```
image

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/tree-sitter-wing/grammar.js
+++ b/libs/tree-sitter-wing/grammar.js
@@ -55,7 +55,12 @@ module.exports = grammar({
       choice(braced(optional(repeat($._statement))), $.AUTOMATIC_BLOCK),
     _semicolon: ($) => choice(";", $.AUTOMATIC_SEMICOLON),
     comment: ($) =>
-      token(choice(seq(/\/\/[^\/]/, /.*/), seq("/*", /[^*]*\*+([^/*][^*]*\*+)*/, "/"))),
+      token(
+        choice(
+          seq(/\/\/[^\/]/, /.*/),
+          seq("/*", /[^*]*\*+([^/*][^*]*\*+)*/, "/")
+        )
+      ),
 
     doc: ($) => seq("///", field("content", $.doc_content)),
     doc_content: ($) => /.*/,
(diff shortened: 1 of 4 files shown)


libs/tree-sitter-wing/grammar.js
```diff
@@ -334,7 +353,15 @@ module.exports = grammar({
         optional(
           seq(
             "catch",
-            optional(field("exception_identifier", $.identifier)),
+            optional(
```
yoav-steinberg, line 356: This is extra grammar just for more specific errors. I guess this is ok, and it's probably the easiest way to deal with this.
I wonder if this is another hint that we need to eventually move to our own parser? @MarkMcCulloh I wonder what's your take on this?

yoav-steinberg (approved):
:heavy_check_mark:

libs/tree-sitter-wing/grammar.js
MarkMcCulloh: I'm definitely in favor of us doing our own parser for error-correcting cases like this and others.

> This is extra grammar just for more specific errors

The way I see it: If we were to write our own parser, we would definitely add code to specifically detect this case. Adding it to our ts grammar is currently our only way to do that.

It's worth noting we have the same issue in several of our syntaxes wrt to parens:

```wing
for (x in [2]) {}
if (let x = 2) {}
```

I think our parser should be able to avoid exploding for all of these

Issue #5212, opened 2023-12-13
Invoke from the simulator does not work


meirdev:
### I tried this:

Get the request data when the user clicks invoke:

```
bring cloud;

let api = new cloud.Api();

api.post("/test", inflight (req: cloud.ApiRequest): cloud.ApiResponse => {
  if let requestBody = req.body {
    let data = Json.parse(requestBody);

    return cloud.ApiResponse {
      status: 200,
      body: data.get("message").asStr()
    };
  }

  return cloud.ApiResponse {
    status: 400,
    body: "error"
  };
});
```

Invoke message:

```
{"body": "{\"message\": \"hello\"}"}
```

### This happened:

It is not possible to access the body of the request, an error (400) always returns.

### I expected this:

I expect to receive:

```json
{
  "success": true,
  "response": {
    "status": 200,
    "body": "hello"
  }
}
```

### Anything else?

It looks like the code is getting the request object as a string. And I think the problem is here:

https://github.com/winglang/wing/blob/e64d002ff01ad64d9f0680587f790d1b7b3959d1/libs/wingsdk/src/target-sim/function.inflight.ts#L56

If I change to this:

```typescript
return sb.call("handler", payload);
```

It works:

<img width="818" alt="Screenshot 2023-12-13 at 21 33 24" src="https://github.com/winglang/wing/assets/21178966/05b03c76-f9f4-4645-afd9-a18c27ecc88c">

But if I send a message not in JSON format:

```
hello
```

I get this error:

<img width="1068" alt="Screenshot 2023-12-13 at 21 36 33" src="https://github.com/winglang/wing/assets/21178966/5e76fd70-6d1a-4776-96c5-940abfbb863c">

### Wing Version

0.51.16

### Node.js Version

v21.2.0

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

Chriscbr:
Thanks for the bug report @meirdev. As I understand it, basically you're observing that even though an HTTP request to `cloud.Api` in the simulator ends up calling `invoke()` on `cloud.Function` internally, if you pass the expected `ApiRequest` structure directly to `cloud.Function` in the Wing console, it doesn't seem to work.

[...]

PR #6595, opened 2024-05-30
feat(sdk): show `cloud.Api` endpoint functions

skyrpex:
This change makes `cloud.Api` endpoint functions visible in the Console's map view.


skyrpex:
@eladb not sure if this is the solution you were looking for, I remember this comment here https://github.com/winglang/wing/issues/5212#issuecomment-1854646682.

The problem is that invoking the endpoint functions directly is strange since it's expecting a specific API payload format.

IMO it would be interesting if we detected that they are `cloud.Api` functions and provided some kind of schema for the event payload, but not sure how much work is required for that to happen.

eladb (approved):
Might be good for the time being

PR #6599, opened 2024-05-30
fix(compiler): can't pass inflight closure as parameter to `super()` call

### LINKED ISSUE #4925: Can't pass an inflight closure to a `super` ctor call
### I tried this:

```js
class Base {
  new(handler: inflight (): void) { }
}

class Derived extends Base {
  new() {
    super(inflight () => { });
  }
}

new Derived();
```

### This happened:

```
➜  winglibs git:(main) wing compile ~/Documents/test.main.w 
ERROR: There is already a Construct with name 'Derived' in $Root [Default]

hint: Every preflight object needs a unique identifier within its scope. You can assign one as shown:

> new cloud.Bucket() as "MyBucket";

For more information, see https://www.winglang.io/docs/concepts/application-tree

../../Documents/target/test.main.wsim.543615.tmp/.wing/preflight.js:11
       class Base extends $stdlib.std.Resource {
         constructor($scope, $id, handler) {
>>         super($scope, $id);
         }
         static _toInflightType(context) {
```

### I expected this:

To work

### Is there a workaround?

_No response_

### Component

Compiler

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

_No response_

### Community Notes

yoav-steinberg:
fixes #4925

Now you can pass inflight closures and also any `new PreflightClass()` as parameters to a `super()` call in a ctor.
What I did is detect these cases and made sure the implicit scope passed to the construct ctor is `$scope` and not `this` which isn't available before `super()` is called.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/examples/tests/valid/inflight_closure_as_super_param.test.w
+++ b/examples/tests/valid/inflight_closure_as_super_param.test.w
@@ -27,9 +27,8 @@ let c = new Derived() as "derived";
 assert(nodeof(c.f).path.endsWith("derived/in_derived"));
 // Make sure the instance created in the super call is scoped to the parent (root)
 assert(!nodeof(c.f_base).path.endsWith("derived/in_root"));
-let in_root = regex.compile("root\\/env[0-9]+\\/in_root");
-log(nodeof(c.f_base).path);
-assert(in_root.test(nodeof(c.f_base).path));
+let appPath = nodeof(this).path;
+assert(nodeof(c.f_base).path == "{appPath}/in_root");
 
 test "boom!" {
   assert(c.h() == "boom!");
--- a/examples/tests/valid/inflight_closure_as_super_param.test.w
+++ b/examples/tests/valid/inflight_closure_as_super_param.test.w
@@ -1,0 +1,35 @@
+class Foo {}
+
+class Base {
+  pub h: inflight (): str;
+  pub f_base: Foo;
+  new(handler: inflight (): str, f: Foo) { 
+    this.h = handler;
+    this.f_base = f;
+  }
+}
+
+class Derived extends Base {
+  pub f: Foo;
+  new() {
+    super(inflight (): str => { 
+      return "boom!";
+    }, new Foo() as "in_root");
+    this.f = new Foo() as "in_derived";
+  }
+}
+
+let c = new Derived() as "derived";
+// Make sure that instances created in a ctor are scoped to the instance they were created in
+// This is related to this test because the transformed inflight closure is also an instance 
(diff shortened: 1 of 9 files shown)


examples/tests/valid/inflight_closure_as_super_param.test.w
```diff
@@ -0,0 +1,35 @@
...
+    this.f_base = f;
+  }
+}
+
+class Derived extends Base {
+  pub f: Foo;
+  new() {
+    super(inflight (): str => { 
+      return "boom!";
+    }, new Foo() as "in_root");
```
Chriscbr, line 17: For the case of arbitrary preflight classes, the issue of not being able to pass a newly-constructed class to super() is interesting because the normal construction pattern (where the child class receives the "this" scope) also isn't possible in TypeScript:

```ts
class MyConstruct extends Construct {
  constructor(scope: Construct, id: string, helper: Construct) {
    super(scope, id);
    // do something with helper...
  }
}

class MySuperConstruct extends MyConstruct {
  constructor(scope: Construct, id: string) {
    let c = new Construct(this, "Helper"); // error: 'super' must be called before accessing this
    super(scope, id, c);
  }
}
```

As you observed, it's possible to reuse the "scope" to make it work:

```ts
let c = new Construct(scope, "Helper");
```

But it's worth noting this technically changes the tree structure so the construct being created is now a sibling instead of a child. Not necessarily a big deal.

Having the compiler do this automatically sounds like a fair compromise given that I don't think there's any other workaround. Indeed, I think the only other way to access the implicit "$scope" argument in Winglang right now requires the `nodeof()` function right now - but `nodeof` expects a scope, so it doesn't really help:

```
class A {
  new(bucket: cloud.Bucket) {
  }
}

class B extends A {
  new() {
    super(new cloud.Bucket() in nodeof(this).scope!); // error: using this before "super" has been called
  }
}
```

I feel like it could be worth documenting in the language reference or [here](https://github.com/winglang/wing/blob/main/docs/docs/02-concepts/02-application-tree.md) that classes constructed inside `super()` statements will be siblings instead of children in the tree - what do you think?

Chriscbr (approved):
Nice fix! 🙌

examples/tests/valid/inflight_closure_as_super_param.test.w
yoav-steinberg: :+1:  Agree. I see the docs only contain information about preflight class "naming" and not about "scoping", I'll add a section and including this edge case.

eladb (approved):
Sweet!

PR #6602, opened 2024-05-30
feat: show connections for token references

### LINKED ISSUE #6433: Treat token references as connections (and show them in the Console)
### Use Case

When a resource references an attribute of another resource, it introduces a relationship between them, similarly to the relationship created when inflight code calls a method on another preflight resource.

I want to see these relationships (connections) in the Wing Console!

The fact that the relationship is introduced through a reference to an attribute and not a method call doesn't make it less of a relationship.

Before:
image

After:

image

### Proposed Solution

We should identify tokens when we call `toSimulator()` in the resource configuration (simulator initially, but can also do it for terraform), and simply add a connection to the `tree.json`.

It should be straightforward to add something like this.

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

### LINKED ISSUE #6607: `invokeAsync` connection is not displayed in Wing Console
### I tried this:

```js
let fn1 = new cloud.Function(inflight () => {}) as "fn1";

let fn2 = new cloud.Function(inflight () => {
  fn1.invokeAsync();
}) as "fn2";
```

### This happened:

image

### I expected this:

To see the connection between the two functions

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

eladb:
Resolves #6433 by adding a connection edge when resources reference each other through tokens (e.g. a function references `api.url`). The fact that these are implicit references doesn't mean they are not references...

## Console UI Tweaks

* Remove the `()` from the connection operations because oftentimes they do not represent an actual method name. If we want to distinguish between operations and other types of connections, we should add this to the model and then display.
* Only show the operation when there's a `sourceOp` in the connection. Do not default to the connection name.
* In the navigation pane, change the label from "Access" to "Connections".
* Restore connections for `invokeAsync()`. Not sure why they were removed (resolves #6607)
* Change the node title of topics created for bucket notifications to `OnXxx` instead of `onxxx` (all lowercase).
* Change the lift-based connection name to `call`.
* Change the title for the queue consumer function to `Consumer` instead of `SetConsumer()`, also `Tick` (for schedules), `Subscriber` (for topics).

## Misc

* Make the `source` property of `nodeof(x).addConnection()` optional and default to `this`.

## Checklist

- [ ] Tests added (always)


skyrpex:
image

These kind of connections are visually awkward and confusing to me (from parent to child). That's why I suggested a few days ago to use pseudo-containers like these:

image

We keep a hint of the hierarchy but the connections are easier to read.

skyrpex:
But that's subject for a different PR/proposal.

skyrpex:
> For now, maybe we could make it so the connections from parent to children go from west to east. I'll try amending the PR.

Nevermind, a proper refactor of the construct map nodes would be needed. It's a bit of a mess right now and can't easily do it.

libs/wingsdk/src/target-sim/app.ts
```diff
@@ -265,6 +269,35 @@ export class App extends core.App {
...
+      resolveTokens(resource.props, (to) => {
+        // skip references to the "handle" of the target resource because it would be reflected by
+        // the connections created by inflight method calls.
+        if (to.attr !== "handle") {
+          core.Connections.of(this).add({
+            source: map[from],
+            target: map[to.path],
+            targetOp: to.attr,
+            name: "<ref>",
+          });
```
Chriscbr, line 294: Food for thought -- this approach doesn't produce the "sourceOp" so we're losing information about what method(s) are referencing to the other resource. For example if we have a class like this:

```wing
bring cloud;

class A {
  bucket: cloud.Bucket;
  api: cloud.Api;
  new() {
    this.api = new cloud.Api();
  }
  pub inflight foo() {
    this.bucket.list();
  }
  pub inflight bar() {
    log(this.api.url);
  }
}

let a = new A();
```

Then we can show the specific relationship:

Screenshot 2024-05-31 at 3 51 38 PM

WDYT?

I think I got a minimal version working - here's a git patch if you feel up to incorporating it into your PR:

```patch
diff --git a/libs/wingsdk/src/std/resource.ts b/libs/wingsdk/src/std/resource.ts
index 3573d0d7e..7257da780 100644
--- a/libs/wingsdk/src/std/resource.ts
+++ b/libs/wingsdk/src/std/resource.ts
@@ -1,7 +1,9 @@
 import { Construct, IConstruct } from "constructs";
 import { App, LiftMap } from "../core";
 import { AbstractMemberError } from "../core/errors";
+import { parseToken } from "../simulator/tokens";
 import { Node } from "../std";
+import { isSimulatorToken } from "../target-sim/tokens";

 /**
  * A resource that can run inflight code.
@@ -219,6 +221,9 @@ function addConnectionsFromLiftMap(
   baseOp?: string
 ) {
   for (const [op, liftEntries] of Object.entries(liftData)) {
+    if (baseOp ?? op === "$inflight_init") {
+      continue;
+    }
     for (const [dep, depOps] of liftEntries) {
       if (Construct.isConstruct(dep) && !(dep instanceof AutoIdResource)) {
         // case 1: dep is an ordinary resource
@@ -234,6 +239,19 @@ function addConnectionsFromLiftMap(
       } else if (hasLiftMap(dep)) {
         // case 2: dep is an inflight
         addConnectionsFromLiftMap(construct, dep._liftMap, baseOp ?? op);
+      } else if (isSimulatorToken(dep)) {
+        // case 3: dep is a simulator token
+        const { path, attr, prop } = parseToken(dep);
+        const target = Node.of(construct)
+          .findAll()
+          .filter((c) => c.node.path === path)[0];
+        Node.of(construct).addConnection({
+          source: construct,
+          sourceOp: baseOp ?? op,
+          target,
+          targetOp: prop ?? attr,
+          name: "ref",
+        });
       }
     }
   }
```

Chriscbr (approved):
Nice 👍  I added an optional suggestion in a comment

libs/wingsdk/src/target-sim/app.ts
eladb:
```sh
$ git apply <<HERE
PASTE THE ABOVE
HERE
```

Magic!

eladb: @Chriscbr I am not sure how this would work... The lift map only includes information about inflight method calls, not references of preflight values. Right?

eladb: I am leaving the previous implementation for now.

skyrpex (approved):

libs/wingsdk/src/target-sim/app.ts
Chriscbr: > The lift map only includes information about inflight method calls, not references of preflight values. Right?

@eladb Here's the lift map for the example snippet above:

```
      get _liftMap() {
        return ({
          "foo": [
            [this.bucket, ["list"]],
          ],
          "bar": [
            [this.api.url, []],
          ],
          "$inflight_init": [
            [this.api.url, []],
            [this.bucket, []],
          ],
        });
      }
```
The token (this.api.url) is a reference to a preflight value and you can see it's included in the lift map -- that's how I was able to produce the screenshot in my earlier comment.

I think there's a bug in the UI where the console automatically displays everything with trailing parentheses, so it looks like `url()` instead of `url` in the screenshot - maybe that's the confusion?

PR #6611, opened 2024-05-31
poc(console): add support for ui.graph

polamoros:
@Chriscbr 
image


skyrpex:
Can you change the colors? The contrast ratio is too low

image

PR #6612, opened 2024-05-31
feat(sdk): fewer replacements for cloud.Service and sim.Resource

Chriscbr:
Extends the changes made in https://github.com/winglang/wing/pull/6577 for reducing the time it takes for the simulator to update to cloud.Service and sim.Resource.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-sim/resource.inflight.ts
+++ b/libs/wingsdk/src/target-sim/resource.inflight.ts
@@ -1,7 +1,7 @@
 import * as path from "path";
 import { IResourceClient, SIM_RESOURCE_FQN } from "./resource";
 import { SimResourceAttributes, SimResourceSchema } from "./schema-resources";
-import { Bundle } from "../shared/bundling";
+import { Bundle, isBundleInvalidated } from "../shared/bundling";
 import {
   SandboxMultipleConcurrentCallsError,
   Sandbox,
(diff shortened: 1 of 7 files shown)


libs/wingsdk/src/target-sim/resource.inflight.ts
```diff
@@ -136,9 +136,23 @@ export class Resource implements IResourceClient, ISimulatorResourceInstance {
...
+  public async plan(invalidated: boolean): Promise<UpdatePlan> {
+    // TODO: support customizing update plans
+    if (invalidated) {
+      return UpdatePlan.REPLACE;
+    }
+
+    // Check if any of the bundled files have changed since the last bundling
+    const bundleInvalidated = await isBundleInvalidated(
+      this.resolvedSourceCodeFile,
+      this.bundle!,
```
eladb, line 148: Maybe as a future refactor, would it make sense to encapsulate this behavior into `Bundle`?

And then it would be something like:

```js
if (bundle.isInvalidated()) ...
```

Or something like this...

Could be nice to also encapsulate `ensureBundled` into this class so there's less duplication across consumers.

libs/wingsdk/test/sim-app.ts
```diff
@@ -10,6 +10,7 @@ import { App } from "../src/target-sim/app";
  * @see AppProps
  */
 export interface SimAppProps {
+  readonly outdir?: string;
```
eladb, line 13: Add a doc comment explaining the default

eladb (approved):

libs/wingsdk/src/target-sim/resource.inflight.ts
Chriscbr: Yeah, I agree it it makes sense to make Bundle into its own stateful class etc. I'll open an issue

eladb (approved):

Issue #6538, opened 2024-05-22
API for creating Wing clients in other languages


Chriscbr:
### Use Case

The use case is for providing a seamless development experience for teams where application developers and platform engineers have separate responsibilities, and where application developers aren't dealing with Wing or infrastructure.

As a platform engineer I'd like to use Wing to create the preflight infrastructure for developers in my organization, and then produce client libraries in TypeScript (or other languages) for accessing that infrastructure within other pieces of code. Imagine for example that the application developers write TypeScript or Python running in containers in their own repo, and that repo's code is already being automatically deployed with GitOps via ArgoCD or other means.

As an application developer I'd like to consume the TypeScript package published by the platform team and write application code that interacts with these resources (using whichever methods or permissions have been granted by the platform team). If I'm running my TypeScript code / container locally I expect the Wing clients to connect to the Wing simulator and interact with the simulated resources, and when my code is deployed to the AWS I expect the clients to connect to the real cloud resources

_(credit goes to @eladb for raising this use case)_

### Proposed Solution

On the platform engineer side, I can write a Wing application that looks something like this:

```js
bring cloud;
bring std;

let b = new cloud.Bucket();

let client = new std.TypeScriptClient();
client.lift(b, as: "myBucket", with: ["put"]);

new TerraformOutput(name: "WingClientContext", client.context);
```

TypeScriptClient will produce a JavaScript npm package with TypeScript typings for the client (with package.json, index.ts, index.d.ts etc.). By exporting the client context through Terraform output, I can expose the deploy-time configuration needed by the Wing Client. As a platform engineer I can then mount that configuration to the containers through a ConfigMap or other means.

On the application developer side, I can use the Wing client with `npm i @acme/wing-clients` and using it like so:

```js
import { createClient } from "@acme/wing-clients";

async function main() {
  const client = await createClient(); // reads from process.env.WING_CLIENT_CONTEXT
  await client.myBucket.put("key", "value");
}
```

In this simple example, only myBucket (a class from the Wing SDK) is being exposed, but it's possible to package user-defined classes / abstractions as well.

### Component

Compiler, SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Discord](https://t.winglang.io/discord).

eladb:
[...]

PR #6614, opened 2024-06-02
rfc: generate TypeScript clients 

MarkMcCulloh:
For https://github.com/winglang/wing/issues/6538

[Rendered RFC](https://github.com/winglang/wing/blob/mark/ts-client-rfc/docs/contributing/999-rfcs/2024-06-02-ts-client.md)


docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,121 @@
...
+An important consideration for this discussion is allowing the generated client to reference resources defined from a separate wing application.
+To facilitate this separation, it seems necessary to introduce a mechanism to export/import deployed data between Wing apps.
+The rest of this proposal does it via `cloud.Outputs`, but that likely deserves its own RFC.
+
+## Example Situation
+
+Acme Co. has a platform team and several application teams. Their platform team manages Kubernetes clusters in AWS with additional AWS resources as needed.
+Application teams write applications in TypeScript and deploy them to these shared Kubernetes clusters.
+The centralized infrastructure definitions are written in Wing in their own repository.
+The applications are written in TypeScript but also have Wing files for infrastructure/deployment.
```
eladb, line 27: > but also have Wing files for infrastructure/deployment.

What's that? Can you elaborate?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,121 @@
...
+
+To add a new application, the platform team needs to ensure that the application has the necessary permissions to interact with the shared resources. To do this, a kubernetes namespace and service account is created:
+
+```wing
+let namespace = cluster.createNamespace("app123");
```
eladb, line 32: Let's assume users don't need wing to define k8s resources

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,121 @@
...
+Additionally, it exports a wing library with all the necessary types and runtime code to look up that information.
+This library is published to an internal registry for other teams.
+
+After the central infrastructure is deployed, the application team can import the information into their own wing application:
+
+```wing
+bring "@acme/core-infra" as coreInfra;
+
+let workload = new Workload(
+  namespace: coreInfra.namespace,
```
eladb, line 73: Let's assume wing is not used to define k8s resources.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,121 @@
...
+
+Now the application will have the necessary permissions to interact with the shared bucket.
+
+However, we need to create a client for TypeScript to actually make use of these permissions.
+The environment where the client will run needs to know where to resolve the bucket.
+To do this, the client can attach to any inflight host. This takes care of preparing it for the client to run.
+Luckily, the workload exposes this host for us:
+
+```wing
+let client = new TypeScriptClient(
```
eladb, line 87: Feels like this needs to be defined at the infra wing level and published to be used by the application.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,121 @@
...
+import { sharedBucket } from "../.wing";
+
+await sharedBucket.put("key", "value");
+```
+
+To run this locally, `wing it` is used for both the root infrastructure and then the application.
+
+### Implementation Thoughts
+
+#### Skipping cloud.Outputs?
```
eladb, line 109: I don't think we need Outputs but we need a way to define a client and somehow retrieve (after/during deployment) the blob of "binding context" this client needs in order to be initialized, and leave the job of propagating this to the user.

```js
let client = new TypeScriptClient(...);


new CfnOutput(
  name: "ClientContext", 
  value: client.bindContext
);

// or

new k8s.ConfigMap(
  name: "ClientContext", 
  value: client.bindContext
)
```

eladb (changes requested):
Good start. 

I think it will be useful to limit the scope of this RFC to the low level APIs and not attempt to provide a full end to end DX. This is because we want this to support different types of stacks (k8s, VMs, serverless, etc) and we need to first focus on getting the mechanism right.

We can then add layers of DX to support specific use cases.

Here are some useful assumptions I would take:

1. Infra is in a separate repo and written in Winglang
2. App is in a separate repo and can be any TypeScript app (running locally, in k8s, serverless). No winglang in this repository.
3. Wing is not used to deploy any app related resources to the cloud nor to run the app in the simulator.
4. Infra is always deployed before the app.

Something like this.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: I was expecting wing to be used to deploy the application, because otherwise it's not clear to me how (or why) wing fits into the workflow.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: They won't use wing to define k8s resources on either the core infra or the application side? So they would only use wing to define other resources?

Wing could be used to create e.g. an IAM role, and then the platform team (or app team) could manually reference this role in their manifests. This wouldn't the RFC overall, to me that would still necessitate the `cloud.Outputs` to somehow get this information out.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: If wing is not used to deploy the application, how could the application know what information is needed?

If your answer is "the client", that's fine but the client itself still needs to look up the information. If that's the responsibility of the user, what help have we actually given them?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: How would the client know how to resolve the information?
I know that's basically the point of this RFC, but if the answer is "that's completely up to the user" then this RFC could be shortened to "TypeScriptClient writes types and inflight clients to disk, then the users can figure out how to make use of that"

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: > leave the job of propagating this to the user.

This is basically 90% of the work involved in this feature.
If both the process of binding the data from the source and reading the data from the client requires explicit actions by the user, then all we're doing is generating types and writing our existing JS clients to disk with invisible requirements.

> we need a way to define a client and somehow retrieve (after/during deployment)

My proposal here is how I think that could work. Do you think it wouldn't? I understand your original proposal, I just don't know how it could actually work without giving the user a bad experience. I especially don't know how this would work with local development if the core infrastructure is running in the simulator.

The client can only resolve the necessary information in one way. Does it resolve it from somewhere on disk? environment variable?
If you intend to use a ConfigMap, they it would make sense to be on disk. If you're consuming this through a terraform/cfn output, then you might what it to be through environment variables. But since the client can only work one way, this choice only adds friction.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: In this workflow wing will only be used to define the preflight (infra) while the app will be deployed by other tools like ArgoCD for example.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: It's a combination of "the client" and some "binding context" that is outputted by the deployment engine and somehow passed to the client upon initialization.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: Yes. Ideally this is a simple low level mechanism but it's not enough to just write stuff to disk because some binding context is only resolved during deployment (eg bucket ARN). I agree that one major piece is generating the client and the other is some kind of binding context protocol that is created during deployment exported and then passed to the app when it starts so it can give it to the client.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: See above comments.

We need to define the low level protocol and I think that should be the scope of this RFC.

It makes sense to include some examples in an appendix on how this low level protocol can be used to deploy apps in various environments.

Maybe include an example for:
1. K8s
2. Lambda
3. EC2

It would be valuable to look at three different examples to make sure the low level mechanism holds water.

But the goal is not to focus on the DX of each one of these examples, but rather to define the low level protocol.

Makes sense?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: Yes, permissions are platform-specific and definitely require some solution in the RFC. 

Maybe when `TypeScriptClient` is compiled to AWS it would provision an IAM policy resource of some kind so that the app's role could use.

Maybe the user will be required in this case to specify the name of the IAM role/policy to provision in `wing.toml` so that the app can reference.

MarkMcCulloh:
Closing for now to rework

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+```wing
+bring cloud;
+bring clients;
+bring fs;
+
+let client = new clients.TypeScript(
+  name: "@acme/infra-client",
+  contextResolver: inflight () => {
+    return fs.readJson("/var/run/wing-client-context.json");
+  },
```
MarkMcCulloh, line 57: Not sure if there should be a default implementation, I feel like there are too many options to pick one

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+To lift a resource into the client to make it accessible to consumers, add a `lift` expression:
+
+```wing
+bring cloud;
+
+let bucket = new cloud.Bucket();
+
+client.lift(lift { bucket: [put] });
```
MarkMcCulloh, line 41: `.lift(lift ...)` looks weird, but `.add(lift ...)` felt imprecise

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+  "version": "1",
+  "env": {
+    "BUCKET_HANDLE_a107cec5": "abcxyz"
+  }
+}
+```
+
+The client itself is effectively a proxy of an unseen/unavailable `IInflightHost`, so any expressivity that interface has must be handled by the client.
+
+### Permissions
```
MarkMcCulloh, line 136: I'd love to tackle permissions here, but I struggled to think of a solution that doesn't explode the scope of this RFC

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+
+If you're using terraform and deploying to AWS, you will need the bucket name and the correspond IAM policy.
+There is no abstract API available to store/retrieve this information currently as this is handled only by target-specific subtypes of `IInflightHost`.
+
+To keep this simple, we can provide and a helper API in the shared aws module (and other platforms, as needed) to generate the necessary policies by exposing the lift information (i.e. the `_liftMap`) from the client.
+
+```wing
+bring aws;
+
+let policies = aws.generatePolicyStatements(client.lifts);
```
MarkMcCulloh, line 157: I believe this functionality basically exists already, so the only real work here is to expose it

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+    b.put("key", "value");
+  }
+}
+```
+
+With some changes, with syntax could also be used as a way to create a preflight container for lifted data.
+
+```wing
+let b = new cloud.Bucket();
+let lifts = lift { b: [put] };
```
MarkMcCulloh, line 198: I considered proposing `@lift({ b: [put] })`, but I'm not sure it was valuable to discuss this change here since really effect the behavior. I expect the change about what `lift` can do to be more controversial anyways.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+### Type generation with `lift`
+
+Wing's compiler has type information available for use, but only before preflight execution occurs.
+This makes an API like the following not feasible to implement using logic similar to the existing `extern` or `@inflight`:
+
+```wing
+client.lift("bucket", bucket, ["put"]);
+```
+
+While executing preflight, the knowledge about the type of the `bucket` instance is lost, so generating the actual client code will have no way to reference it.
```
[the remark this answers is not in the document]
MarkMcCulloh: > What if we simply had the Wing compiler generates a toInflightDts() method on every generated class that returns a string of its TypeScript DTS types?

This implies we would need to eagerly generate .d.ts for all types, which is a waste of compiler time when not used. This would have to include any referenced JSII types too, since the generated .d.ts files will need to reference them.

Additionally, the final types need to know what has been lifted. If you only lift `put` on bucket, the entire bucket client type should not be available.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+/// aws-cdk-lib
+new awscdk.CFNOutput(
+  name: "ClientContext",
+  value: client.bindContext
+);
+```
+
+This could be used to create a user/role that is eventually usable within the k8s cluster as a service account.
+
+### AWS Lambda
```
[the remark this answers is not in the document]
MarkMcCulloh: In the sentence here I do mention creating a role and using a k8s service account. I didn't write it out because at that point it feels out of scope for the RFC.
I can specifically add a note about `aws.generatePolicyStatements(client.lifts)` as you mentioned, but writing out the actual usage of policy statements wouldn't really change how the clients works.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+
+let bucket = new cloud.Bucket() as "shared";
+
+let client = new clients.TypeScript(
+  name: "@acme/infra-client",
+  contextResolver: inflight () => {
+    let ctxRaw = util.getEnv("WING_CLIENT_CONTEXT");
+    let ctxDecoded = util.fromBase64(ctxRaw);
+    return Json.parse(ctxDecoded);
+  },
```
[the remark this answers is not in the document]
MarkMcCulloh: I suspect custom resolvers will be common since the user will have to manually deal with the storing the context. I do like the idea of premade ones though.

Not sure this actually works right now, but WDYT about static inflights instead of an enum?

```wing
  contextResolver: clients.Resolver.FROM_ENV // <-- This is an inflight function
```

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: I mention the issues with that below and why I think it'd be good to make changes to the lift syntax

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
Chriscbr: > This implies we would need to eagerly generate .d.ts for all types, which is a waste of compiler time when not used. This would have to include any referenced JSII types too, since the generated .d.ts files will need to reference them.

Maybe I got confused -- I saw the RFC says "An initial implementation of this can re-use the existing .d.ts generation, and the std.LiftQualificationsContainer can store a path to that generated file to be used as-is when generating the full client in preflight". Does this mean we will still be generating .d.ts files in wingc for every `lift { }` expression? Or would it mean it's being generated during preflight runtime (sounds like it could be slower than doing it in compiled Rust)?

> Additionally, the final types need to know what has been lifted. If you only lift put on bucket, the entire bucket client type should not be available.

Don't we handle this in the TypeScript framework already? with the Pick/Omit stuff?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
Chriscbr: Either sounds nice to me 👍

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
MarkMcCulloh: > Does this mean we will still be generating .d.ts files in wingc for every lift { } expression?

Yes. The important part of the `lift` syntax is that our compiler can analyze the types and ops statically and emit the precise type information during justification to be used later. This would not be possible with an api like `client.lift(bucket, ["put"])` without a relatively rigorous reflection system to do the same thing.

A big part of the motivation for direction is that I think the changes to `lift` can be used elsewhere, like unifying how lifting works in `@inflight` and resources like `cloud.Website`

> Don't we handle this in the TypeScript framework already? with the Pick/Omit stuff?

It does, but the context is different. That is operating on fully packed winglibs in preflight.
The clients are inflight-only, and should ideally not even contain any irrelevant types at all. Additionally, the clients will not even have dependencies since they'll ideally be bundled like all our other inflight code (which I should probably mention in this do, oopsie).

My hunch is that getting `lift` to work here would be less work (both implementation and design-wise) than getting reflection to work well enough to support the use case.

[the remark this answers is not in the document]
Chriscbr: Ok, got it. yeah it would be cool to see a sketch showing more of how std.LiftQualificationsContainer would look like but it sounds like it could be handy

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: Not sure where this inflight closure will be executed. 

I think we should simply make `context` a string token and then it can be referenced.

eladb (changes requested):
Haven't gone through the whole thing yet but this is really good!

👍

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+let client = new clients.TypeScript(
+  name: "@acme/infra-client",
+  contextResolver: inflight () => {
+    return fs.readJson("/var/run/wing-client-context.json");
+  },
+);
+
+// The consumer of the client will need to retrieve/link this data after deployment
+let clientContext = new cloud.Bucket() as "ctx";
+clientContext.addObject("ctx", Json.stringify(client.context))
```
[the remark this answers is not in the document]
eladb: Or even just a stringified json

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+
+The client code is the same as the generated javascript as if it were used within wing itself!
+
+## Implementation Details
+
+### Client context data
+
+The client context data is a JSON object that contains all the necessary information to connect to the resources.
+For now, the only information that needs to be stored are environment variables.
+This corresponds to the only API available on `IInflightHost`, which is `addEnvironment`, which is how these values are added:
```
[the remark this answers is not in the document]
MarkMcCulloh: This RFC should discuss changing IInflightHost and and also all inflight clients? 
They currently use environment variables, and this proposal was just going to reuse those inflights.

I understand what you mean, the way clients work now does not need to rely completely on env, but I didn't think we needed to change all that here in order for a TypeScriptClient to work.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+
+```json
+{
+  "version": "1",
+  "env": {
```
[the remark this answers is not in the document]
MarkMcCulloh: I don't think it's overloading the term, we do currently expect our inflight to run in an place with the concept of environment variables.
We can change this of course, it's just the case right now.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+The client code is the same as the generated javascript as if it were used within wing itself!
+
+## Implementation Details
+
+### Client context data
+
+The client context data is a JSON object that contains all the necessary information to connect to the resources.
+For now, the only information that needs to be stored are environment variables.
```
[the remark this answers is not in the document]
MarkMcCulloh: Currently they are, inflight clients and IInflightHost explicitly rely on environment variables. Conceptually they don't need to, but as mentioned here I'm matching the current IInflightHost

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: Mmmmm. I see, so the intent is that when a client is instantiated by the consuming app, it will mutate the current process' environment so that the inflight clients can be initialized.

I think we should change our implementation of `cloud.Function` (and any other inflight host) to use `TypeScriptClient` under the hood, so there's actually no real value in keeping the current environment variables.

But tactically, I can see the incremental value. Let's make sure this is discussed in the RFC.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
Chriscbr: > Let's add a way to define this as constructor props as well

What does this mean?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+For example, when using
+
+```wing
+client.lift(lift { bucket: [put] });
+```
+
+it would be helpful to retrieve the necessary information about the `bucket` and the `put` operation.
+
+If you're using terraform and deploying to AWS, you will need the bucket name and the correspond IAM policy.
+There is no abstract API available to store/retrieve this information currently as this is handled only by target-specific subtypes of `IInflightHost`.
```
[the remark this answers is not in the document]
MarkMcCulloh: What's the benefit of this approach over the current one in this proposal?

The end result is the same either way (you can get the necessary policy data) but the DX of `aws.generatePolicyStatements(client.lifts)` is both easy and powerful to me. I think it helps convey "You need to take care of making sure these permissions are applied." because you're the one that has to extract them in the first place.

The client is also otherwise completely platform agnostic in the same way an inflight is (i.e. it's more of a set of instructions rather than a "resource"), which seems pretty useful.
I'm not over-indexing on this agnosticism, but I don't see the benefit of getting rid of it.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,298 @@
...
+
+As a platform engineer I'd like to use wing to create the preflight infrastructure for developers in my organization, and then produce client libraries in TypeScript for accessing that infrastructure within other pieces of code.
+Imagine for example that the application developers write TypeScript running in containers in their own repo, and that repo's code is already being automatically deployed with GitOps via ArgoCD or other means.
+
+As an application developer I'd like to consume the TypeScript package published by the platform team and write application code that interacts with these resources (using whichever methods or permissions have been granted by the platform team). If I'm running my TypeScript code / container locally I expect the wing clients to connect to the wing simulator and interact with the simulated resources, and when my code is deployed to the AWS I expect the clients to connect to the real cloud resources.
+
+## Proposal
+
+The `clients` module has a `TypeScript` class used to create a client. When this is instantiated, `wing compile` will produce a publishable JavaScript npm package with TypeScript type declarations.
```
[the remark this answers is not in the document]
MarkMcCulloh: I'm open to suggestions. Any idea about what feels off?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: What if start with only the constructor prop? Allowing the `.lift()` API is probably a more complicated implementation because it would require a way to combine the final lifts together (including the types).

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: I see two main differences between the proposed approach (let's call it "A") and my proposal (let's call it "B").

1. Approach A is not platform agnostic. From a layering perspective, the way I see it everything that has to do with IAM policies is a platform concern, which implies that we don't want it to be explicit in the application code because it is different across platforms. Currently, the way to customize platformic concerns in Wing is either by creating a custom platform provider or using `wing.toml`. If we require the user to call `aws.generatePolicyStatements` explicitly in their Wing program, the program becomes platform-specific.

2. In approach A the user can just forget to create an IAM policy/role. It's on them to know somehow to call this method, generate the IAM policies and provision the resources. Approach B leads the user into the "pit of success".

I sense that there will be other aspects of these clients that are platform-specific, so it makes sense to design them like any other `cloud.*` resource and enable platform providers and extensions to control their behavior.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,297 @@
...
+
+The client does not directly/automatically handle permissions needed for the arbitrary host it could run on.
+There is no expectation that the deployer of the host is also using wing so there is no shared system to transfer permission knowledge.
+
+For example, when using
+
+```wing
+let client = new clients.TypeScript(
+  name: "@acme/infra-client",
+  lifts: @lift({ bucket: [put] }),
```
eladb, line 115: 1. How does syntax work if I want to lift `myArray[0]` for example?
2. How do I lift multiple objects?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: One way to evaluate names is to see if they read like proper english. "new clients typescript" doesn't read well. (in contrast see "new cloud bucket" which reads beautifully).

I suspect the plural form is off. I am also not sure about the term "client" here. It's a bit overloaded. I was thinking of "bindings" maybe, but not sure...

Maybe the namespace should be `ts` and then it's `new ts.Client()` or `new ts.Bindings()`.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
MarkMcCulloh: > Approach A is not platform agnostic.
> If we require the user to call aws.generatePolicyStatements explicitly in their Wing program, the program becomes platform-specific.

Neither `A` nor `B` will result in a platform agnostic app. Even if a `B` creates a policy, some other AWS specific resource (i.e. role or user) will need to be created to attach the policy to. In many cases I expect this to be done in the same wing app.

> In approach A the user can just forget to create an IAM policy/role. It's on them to know somehow to call this method, generate the IAM policies and provision the resources. Approach B leads the user into the "pit of success".

In `B`, a policy is created but it's up to the user to use it. A policy does nothing on its own. Even worse, the policy is hidden so now creating a role to attach it to is a disconnected process.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
MarkMcCulloh: I'm fine with `typescript.Client`

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: Both of these questions/answers are part of the original syntax. We can improve them but I think that belongs in its own discussion. A benefit of reusing `lift` here is that any improvements to it will benefit its original inflight usage and the usage proposed here.

> How does syntax work if I want to lift myArray[0] for example?

It can be used as is:

```wing
@lift({ buckets[0]: [put] })
```

If the resulting name isn't good, `let` can be used

```wing
let bucket = buckets[0];
@lift({ bucket: [put] })
```

> How do I lift multiple objects?

The existing syntax supports this already

```wing
@lift({ bucket1: [put], bucket2: [get] })
```

eladb: Ok, we need to figure out a better solution. This is still not good.

MarkMcCulloh: Will that block this RFC?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: In approach B, the app code stays the same across different platforms. It will behave differently based on the target platform, and the behavior can be customized by the platform provider without needing to change the app code.

That's a fundamental design tenet for Wing and we should make sure we don't break the layering.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
MarkMcCulloh: > Let's make sure this is discussed in the RFC.

I don't see how this is in scope of the RFC. This RFC is a proposal within an architecture we already have. I don't disagree with your point about environment variables but I feel like the discussion in these threads is a sufficient record and any desire for a wider change can be in its own PR.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
MarkMcCulloh: In approach A, the app code stays the same across different platforms **and** it doesn't behave differently across platforms, so the there's no need to customize the behavior across platforms. This does not break any layering because there's no layer necessary.

To move this forward I will go ahead with your approach.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
eladb: No, it's for the `@lift` RFC :-)

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: OK, thanks! I think it will really help to flush out a couple of end-to-end examples and see how things look & feel.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+For example, it could be stored in a bucket:
+
+```wing
+bring cloud;
+bring clients;
+bring fs;
+
+let client = new typescript.Client(
+  name: "@acme/infra-client",
+);
```
eladb, line 61: I wish we could set the default node id to be based on `name` because if there are multiple clients, it will look like:

```js
new typescript.Client(name: "foo") as "foo";
new typescript.Client(name: "bar") as "bar";
```

This is a common issue... @Chriscbr I believe we discussed this in the context of revisiting the `as X` syntax. Maybe we can at least allow classes to determine the default node ID?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+// Will use the environment variable WING_CONTEXT to resolve the context
+const client = new Client();
+
+// Optionally, you can pass the context directly to bypass the environment variable
+// const client = new Client({ context: "..." });
+
+await client.bucket.put("key", "value");
+```
+
+The client code is the same as the generated javascript as if it were used within wing itself!
```
eladb, line 90: Indeed this is exciting! (`!`)

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: I am okay to punt this, but just want to make sure we are aware that we are introducing an abstraction leak and that's debt we should clean up.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
eladb: See above.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+}
+```
+
+The client itself is effectively a proxy of an unseen/unavailable `IInflightHost`, so any expressivity that interface has must be handled by the client.
+
+### Permissions for the simulator
+
+The client is effectively an `IInflightHost`, so it any resource lifted into it registers the permissions it needs.
+This is implemented on a per-target basis to expose the permissions in a useful way.
+For the `sim` target this problem is (currently) much more difficult.
```
eladb, line 117: We have permission modeling for `sim`, no? Should work like `cloud.Function`s.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+As an MVP, the simulator can be configured to ignore permissions via `wing.toml`.
+Even though this is unsafe and cause cause local/cloud differences, wing is intentionally not claiming ownership of the client's host environment.
+
+### Type generation with `@lift`
+
+Wing's compiler has type information available for use, but only before preflight execution occurs.
+This makes an API like the following not feasible to implement using logic similar to the existing `extern` or `@inflight`:
+
+```wing
+client.lift("bucket", bucket, ["put"]);
```
eladb, line 130: Why not use lift expressions here?

```suggestion
client.lift(@lift({ bucket: ["put"] });
```

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+  }
+}
+```
+
+With some changes, with syntax could also be used as a way to create a preflight container for lifted data.
+
+```wing
+let b = new cloud.Bucket();
+let lifts = @lift({ b: [put] });
+//  ^^^^^ is a std.LiftQualificationsContainer
```
eladb, line 154: Why is this called "container"?

```suggestion
//  ^^^^^ is a std.LiftQualifications
```

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+This is very similar to the `lift` mechanism in `@wingcloud/framework`
+
+```ts
+const b = new cloud.Bucket(app, "Bucket");
+const lifts = lift({ bucket }).grant({ bucket: ["put"] });
+```
+
+In the [RFC for explicit lift qualifications](https://github.com/winglang/wing/blob/main/docs/contributing/999-rfcs/2024-03-14-explicit-lift-qualification.md), there is a mention of a `std.LiftQualificationsContainer`.
+
+If this data structure is augmented to also include type information, it could be used to generate the client during preflight.
```
eladb, line 166: This feels like the right direction.

A good way to evaluate if we succeeded would be to implement `typescript.Client` purely in user land.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+
+The client package will contain the following:
+
+- package.json
+
+Minimal package.json with the name and version. For MVP, the package will not have any dependencies because the client code is self-contained.
+
+- index.js
+
+Exports the `Client` class. In this class constructor, the context is read from the environment variable `WING_CONTEXT` (or constructor arg) and is used to set the environment variables for the current process. Each lifted object will then be instantiated which utilizes the existing `_toInflight` logic and will consume all the necessary environment variables.
```
eladb, line 179: > and is used to set the environment variables for the current process. 

This is an implementation detail. Worth mentioning that it might change in the future.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+
+(Future Improvements: The lifted clients could be lazily instantiated?)
```
eladb, line 181: This is not ideal in many cases because it hurts determinism at runtime.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+
```
eladb, line 182: Don't we need the constructor to run in an async context in order to call the async inflight initializers?

I suspect this will need to be something like:

```js
import { Client } from "@acme/infra";

const client = await Client.new();
await client.myBucket.put(...);
```

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+- index.d.ts
+
+Exports the `Client` class, with fields for each lifted object. Each field will reference a type from `lifts.d.ts`.
+
+- lifts.d.ts
```
eladb, line 187: What's the benefit of separating the `d.ts` files?

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+
+To extract the necessary permissions, we set up the wing.toml file to create a new policy:
+
+```toml
+[tf-aws]
+client_policy = "AcmeClientPolicy"
+```
+
+After this policy is created, a role and associated service account must be created for application pods to use it.
+This is only needed for `tf-aws` deployment so it can either but inside a check to the `WING_TARGET` environment variable or in a separate class/file:
```
eladb, line 222:
```suggestion
This is only needed for `tf-aws` deployment so it can either be inside a check to the `WING_TARGET` environment variable or in a separate class/file:
```

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+
+```wing
```
eladb, line 224: I think this part of the example is a bit confusing.

This part would not likely be in the Wing application code but rather somewhere in the platform infrastructure definitions.

For the sake of the example it makes more sense to write it in Terraform (HCL) instead of Wing.

(the fact that you can use Wing to generate the TF is beside the point).

This will make it easier to understand the clear boundary between app and platform, which is the purpose of this PR.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
```diff
@@ -0,0 +1,354 @@
...
+[tf-aws]
+client_policy = "AcmeClientPolicy"
+```
+
+Now the application developers need to setup the lambda with two steps:
+
+1. Map the SSM parameter as an environment variable, which is a directly supported feature in AWS
+2. Attach the created policy to the lambda role
+
+For local development, the owners of the central infrastructure can provide a script to lookup the SSM parameter and add the environment variable to the lambda either via an env or whatever deployment process being used.
```
eladb, line 354: One way to make sure we got this design right is to refactor `cloud.Function` (and any other compute resources) to use this layer under the hood. Eventually this should converge into a single layer that allows runtime code to interact with Wing resources.

eladb (approved):
Awesome progress!

LGTM to me, with some caveats...

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: If you're running a simulator in one process, you can't later "deploy" something else to it. Our sim permissions model only works for things deployed into it.
Tackling this would be a big lift so I lift it out for now

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: This section describes why we can't use the former syntax

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: It's not a necessity, I just expect this to be the initial implementation.
The lifts.d.ts is probably the exact same file generated from `std.LiftQualifications`, while index.d.ts has types for the Client (which will re-export from lifts.d.ts)

They could be combined, we'd just have to deal with the possibility of name conflicts.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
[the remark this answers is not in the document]
MarkMcCulloh: This leak is not exposed to the user, the context value is opaque. I'm writing about it explicitly here because this porting of the RFC is discussing implementation details.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: This is indeed a section discussion implementation detail. I can add "for now"

skyrpex:
Here's an additional use case: I'd like to use Wing code for inflights, but I'd like to use the AWS CDK constructs.

```wing
bring "aws-cdk-lib" as cdk;
bring "@aws-sdk/client" as sdk;

// This is a Wing construct wrapping a CDK Bucket.
class CustomBucket {
   bucket: cdk.aws_s3.Bucket;

   new() {
      this.bucket = new cdk.aws_s3.Bucket();
   }

   inflight client: sdk.Client;

   inflight new() {
      this.client = new sdk.S3();
   }

   inflight putObject(key: string, body: string) {
      this.client.send(new PutObject(
         Bucket: this.bucket.bucketName,
         Key: key,
         Body: body,
      ));
   }
}

// This is a Wing construct wrapping a CDK Lambda.
class CustomFunction {
   func: cdk.aws_lambda.Function;

   new(handler: inflight (): any) {
      // This lift method will generate a compiled JS file for `handler` along with its permissions.
      let code = lift(handler);

      // Here we create a CDK Lambda and manually bundle the generated JS code for the `handler` inflight.
      this.func = new cdk.aws_lambda.Function(
         code: bundleWithEsbuild(
            filename: code.filename,
            handler: code.handler,
         ),
      );

      // Here we can generate the CDK policies by iterating the permission objects.
      for permission in code.permissions {
         // In this example, the `handler` inflight only interacted with our Bucket class.
         permission.resource; // typeof Bucket
         permission.action; // "putObject"

         // new cdk.aws_iam.PolicyStatement(...);
      }
   }
}

let bucket = new CustomBucket();
new CustomFunction(inflight () => {
   bucket.putObject("hello.txt", "world!");
});
```

Long story short, I'm asking for a built-in `lift(handler: inflight (...any):any))` method that gives me both a JS file or tree, and the resources that the handler interacted with. Developers will be able to easily integrate with any cloud provider. AFAIK there's currently no way to use Wing for inflights without using `cloud.Function`.

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: After some prototyping I realized this actually works really well without much effort, I updated the RFC to reflect the reality

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: Removed the note

docs/contributing/999-rfcs/2024-06-02-ts-client.md
MarkMcCulloh: Changed to hcl here, and also used plain kubernetes yaml for the part right after this for similar reasons.

Issue #6622, opened 2024-06-03
feat(console): show resource status


skyrpex:
The intent of this change is to improve the Console diagnostics when a construct fails to initialize during inflight.

![image](https://github.com/winglang/wing/assets/1077520/1e491fe6-2848-4897-ad93-39643be08085)

- The map view will show a small indicator for every construct that failed to initialize (e.g. red circle with exclamation mark), or that it's depending on a resource that failed (e.g. gray circle with a pause icon)
- The tree view will also show small indicators
- The inspector view will also show indicators
- The logs view will show the initialization errors

Also, the simulator should emit these events for every resource:

- starting, error, stopping, started, stopped

Proposal: #6593.

skyrpex:
@eladb @Chriscbr I have a problem. I'm trying to implement the new statuses of resources: starting, error, stopping, started, stopped. But see the following example:

```wing
bring sim;
new sim.Resource(inflight (ctx) => {
  throw "oops";
}) as "MyResource";
```

The "oops" error is logged after `MyResource` is successfully started, and there's no way for me to infer that it really was an initialization error afterwards.

I'm currently trying to figure out if I can change how this works and run these initialization inflights before marking a resource as "started", but may need a hand.

Chriscbr:
I think we need to update the simulator so when it's initializing resources, if any resource fails to start, we should stop all resources that were started instead of continuing to initialize other resources. (It seems like I'm able to find some examples where this doesn't happen) @eladb WDYT?

skyrpex:
I wouldn't stop all resources but any resources that depend on resources that failed to initialize.

It'd be beneficial to allow interacting with whatever is working from the app IMO, and also to detect more than one problem in different unrelated resources, too.

eladb:
I am with Chris. Let's keep this transactional and simple.

I am not sure I see much value in interacting with an app that didn't fully initialize properly.

(For example, in CloudFormation if an update fails, it reverts back all the updated resources)

skyrpex:
Fair enough, will do it tha way 👍🏻

PR #6625, opened 2024-06-03
fix(console): runtime errors do not appear in the logs

skyrpex:
In some instances, runtime errors when invoking `cloud.Function` won't appear in the logs, or they may appear under Verbose logs.

Related to #6622 but doesn't close it.

### DIFF
--- a/libs/wingsdk/src/simulator/simulator.ts
+++ b/libs/wingsdk/src/simulator/simulator.ts
@@ -931,7 +931,7 @@ export class Simulator {
               status: "failure",
             },
             type: TraceType.RESOURCE,
-            level: LogLevel.VERBOSE,
+            level: LogLevel.ERROR,
             sourcePath: resourceConfig.path,
             sourceType: resourceConfig.type,
             timestamp: new Date().toISOString(),
--- a/tools/hangar/__snapshots__/error.ts.snap
+++ b/tools/hangar/__snapshots__/error.ts.snap
@@ -30,7 +30,52 @@ Duration <DURATION>"
 `;
 
 exports[`inflight_stacktraces.test.w 1`] = `
-"fail ┌ inflight_stacktraces.test.wsim » root/env0/test:assert              
+"[ERROR] assert | Error: assertion failed: false
+  --> ../../../examples/tests/error/inflight_stacktraces.test.w:7:3
+  | let bucket = new cloud.Bucket();
+  | 
+  | test \\"assert\\" {
+7 |   assert(false);
+  |   ^
+at <ABSOLUTE>/inflight_stacktraces.test.w:7:3
+[ERROR] expect.equal | Error: Expected values to be strictly deep-equal:
+
+1 !== 2
+
+   --> ../../../examples/tests/error/inflight_stacktraces.test.w:11:3
+   | }
+   | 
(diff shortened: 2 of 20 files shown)


skyrpex:
@ainvoner not sure if the snapshots should have those errors, though. Not sure where they come from 🤔

skyrpex:
@MarkMcCulloh can you please review? 🙏🏻

skyrpex:
I restored the consoleLogger.ts so it only logs to the Console UI errors with `source=user`. There's a problem, though, and it's that when invoking a function from the UI, the error will be logged as `source=compiler`. Will check if I can fix that.

```wing
bring cloud;
// The trace is emitted with source=compiler.
new cloud.Function(inflight (ctx) => {
  throw "oops";
});
```

skyrpex:
We may want to revisit our logging system in the Console, since it may have lost its meaning. Anyways, we are now logging runtime errors properly in the Console logs view.

As a side effect, additional error logs appear in the test snapshots (which seems the correct behavior to me).

tools/hangar/__snapshots__/error.ts.snap
```diff
@@ -30,7 +30,52 @@ Duration <DURATION>"
...
 exports[`inflight_stacktraces.test.w 1`] = `
-"fail ┌ inflight_stacktraces.test.wsim » root/env0/test:assert              
+"[ERROR] assert | Error: assertion failed: false
+  --> ../../../examples/tests/error/inflight_stacktraces.test.w:7:3
+  | let bucket = new cloud.Bucket();
+  | 
+  | test \\"assert\\" {
+7 |   assert(false);
+  |   ^
+at <ABSOLUTE>/inflight_stacktraces.test.w:7:3
```
MarkMcCulloh, line 40: This is the same as the error printed below, I don't think we want this duplicated

tools/hangar/__snapshots__/error.ts.snap
```diff
@@ -30,7 +30,52 @@ Duration <DURATION>"
...
+
+   --> ../../../examples/tests/error/inflight_stacktraces.test.w:11:3
+   | }
+   | 
+   | test \\"expect.equal\\" {
+11 |   expect.equal(1,2 );
+   |   ^
+at <ABSOLUTE>/inflight_stacktraces.test.w:11:3
+[ERROR] bucket failed delete | Error: Object does not exist (key=doesn't exist).
+[ERROR] bucket failed delete | Error: Object does not exist (key=doesn't exist).
```
MarkMcCulloh, line 53: This is a "caught" exception. By default I don't think we want to see these printed

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -931,7 +931,7 @@ export class Simulator {
               status: "failure",
             },
             type: TraceType.RESOURCE,
-            level: LogLevel.VERBOSE,
+            level: LogLevel.ERROR,
```
MarkMcCulloh, line 934: Not sure what to do here, but I think we need to somehow distinguish caught error and uncaught errors

skyrpex:
Thanks for the feedback, Mark. I'm closing this down to gather my thought and also some data because I feel there's something fundamentally wrong with the current logging setup.

Issue #6584, opened 2024-05-28
IDE extension rename bug


eladb:
### I tried this:

See below.

### This happened:

https://github.com/winglang/wing/assets/598796/177ec8d3-2159-4e0b-9ac2-60a236ad46c4

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

staycoolcall911:
Reproduced the issue with latest wing version 0.74.23 and this repo: https://github.com/winglang/translator-example/tree/main

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.58](https://github.com/winglang/wing/releases/tag/v0.74.58).

Issue #6547, opened 2024-05-23
rename changes argument instead of only parameter


ekeren:
### I tried this:

have the following code
```wing
class Example {
  var b: cloud.Bucket;
  new(c: cloud.Bucket){
    this.b = new cloud.Bucket();
  }

  something(b: cloud.Bucket) {
    this.b = b;
    //   ^ put cursor on this
  }
}
```

### This happened:

I put the cursor on the code inside `something` method
```wing
    this.b = b;
    //   ^ put cursor on this
```

and renamed to renamed, the result is:
```wing
bring cloud;

class Example {
  var b: cloud.Bucket;
  new(c: cloud.Bucket){
    this.b = new cloud.Bucket();
  }

  something(renamed: cloud.Bucket) {
    this.renamed = renamed;
  }
}

````

Notice that:
1. the actual member name wasn't renamed
2. The argument of something was renamed

Instead, lets prevent this

### Wing Version

0.74.8

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

eladb:
Yes, encountered this too and was too lazy to submit an issue.

What is our strategy for implementing renames? Isn't it based on the compiler references? If not, why?

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.58](https://github.com/winglang/wing/releases/tag/v0.74.58).

Issue #6546, opened 2024-05-23
vscode "Rename failed to apply edits"


ekeren:
### I tried this:

tried to use rename for variable `k`
```wing
bring cloud;

class Example {
  b: cloud.Bucket;
  new(b: cloud.Bucket){
    this.b = new cloud.Bucket();
    let k = "";
    log(k);
  }
}
```

and got:
<img width="1539" alt="image" src="https://github.com/winglang/wing/assets/1727147/b4c2dcce-090d-4a75-9050-f362d7572b30">

Notice that when trying to rename `k` in this code, it works:
```wing
bring cloud;

class Example {
  // b: cloud.Bucket;
  new(c: cloud.Bucket){
    // this.b = new cloud.Bucket();
    let k = "";
    log(k);
  }
}
```

### This happened:

NA

### Wing Version

0.74.8

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.58](https://github.com/winglang/wing/releases/tag/v0.74.58).

Issue #6484, opened 2024-05-15
rename refactor for class/struct/interface members


tsuf239:
### Use Case

currently, we can't rename class properties, structs keys, and interface keys since it wasn't supported on #6461 

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

eladb:
Can we at least not give the user a false sense that this is supported until this actually works?

Currently the feature feels broken...

tsuf239:
> Can we at least not give the user a false sense that this is supported until this actually works?
> 
> Currently the feature feels broken...

yes, as soon as we implement this: https://github.com/winglang/wing/issues/6564

PR #6637, opened 2024-06-05
feat(vscode): rename class and struct members

tsuf239:
fixes [#6546](https://github.com/winglang/wing/issues/6546)
fixes [#6547](https://github.com/winglang/wing/issues/6547)
fixes [#6484](https://github.com/winglang/wing/issues/6484)
fixes [#6584](https://github.com/winglang/wing/issues/6584)

I added some more visitors and cases to the `rename_request` to enable renaming of class members and struct properties 

## Checklist

- [x] Tests added (always)


libs/wingc/src/lsp/rename_visitor.rs
```diff
@@ -168,6 +207,40 @@ impl<'a> Visit<'a> for RenameVisitor<'a> {
...
+				}
+			}
+			StmtKind::Class(c) => {
+				for (m, ..) in &c.methods {
+					if m.name == "new" {
+						continue;
+					}
+					self.linked_symbols.push(LinkedSymbol {
+						symbol: m.clone(),
+						references: vec![],
```
tsuf239, line 234: if we try to lookup class methods it doesn't resolve to the interface decleration I think, so we need to find some other way to connect to them. I tried to use `c.implements` but it doesn't seems to work- and I don't want to add additional data structures if possible, what am I missing here? 
(I can resolve it on a following PR or in the current one, depends on the complexity of the solution )

tsuf239: if it's possible I think that the logic of 
```
if class.implements[method]- add the base interface method as a linked symbol (if not exists)
    and reference extending interfaces, up to the class' method
if not- add it as a linked symbol
```
and on the interfaces add just the methods that don't exist yet

MarkMcCulloh: > I tried to use c.implements but it doesn't seems to work

What part of it didn't work? I think you're right that class methods don't have a direct linkage to the interface they implement (if any), but it should be possible to lookup in the interfaces in c.implements.

libs/wingc/src/lsp/rename_request.rs
```diff
@@ -419,4 +439,217 @@ test "pow" {
...
+  new(name: str) {
+    this.name = name;
+  }
+
+  inflight pub jump(height: num) {}
+  pub sleep(d: duration) {
+    //----^
+    log("{this.name} sleeps");
+  }
+}
```
MarkMcCulloh, line 589: If this class were public, renaming the method would cause errors in any other files that use it. I suspect this will be surprising for most people using the rename functionality.

Until we implement multi-file renames, maybe would should prevent renames of _public_ classes/structs? I'm not sure if it's better to give the user the option to do it anyway and let them fix it since they'll still get the compiler error.

libs/wingc/src/lsp/rename_visitor.rs
```diff
@@ -168,6 +207,40 @@ impl<'a> Visit<'a> for RenameVisitor<'a> {
...
+				for field in &s.fields {
+					self.linked_symbols.push(LinkedSymbol {
+						symbol: field.name.clone(),
+						references: vec![],
+					})
+				}
+			}
+			StmtKind::Class(c) => {
+				for (m, ..) in &c.methods {
+					if m.name == "new" {
```
MarkMcCulloh, line 229: The constants `CLASS_INIT_NAME` and `CLASS_INFLIGHT_INIT_NAME` can be used here to check for both types of constructors

libs/wingc/src/lsp/rename_visitor.rs
```diff
@@ -29,17 +29,17 @@ impl<'a> RenameVisitor<'a> {
...
+			.any(|s: &LinkedSymbol<'a>| symbol.same(&s.symbol) || s.references.iter().any(|r| symbol.same(r)))
 	}
 
-	fn add_reference_symbol(&mut self, symbol: &'a Symbol) {
+	fn add_reference_symbol(&mut self, symbol: &'a Symbol, symbol_env: Option<&UnsafeRef<SymbolEnv>>) {
 		// symbols that appear in let/if lef statements will point to a prev declaration of a variable of the same name if exists
 		// this is why we add them in advance during visit_statement
 		// the other condition is for "this" that points to the "new" keyword for some reason
-		if self.is_symbol_linked(symbol) {
+		if self.is_symbol_linked(symbol) || symbol.name == "this" {
```
MarkMcCulloh, line 39: Does this need to handle "super" as well?

libs/wingc/src/lsp/rename_visitor.rs
```diff
@@ -154,10 +154,49 @@ impl<'a> Visit<'a> for RenameVisitor<'a> {
...
+				let object_type = self.types.get_expr_id_type_ref(object.id);
+				if let Some(expr_type) = object_type.as_class() {
+					self.add_reference_symbol(property, Some(&UnsafeRef::from(&expr_type.env)));
+				}
+				if let Some(expr_type) = object_type.as_struct() {
+					self.add_reference_symbol(property, Some(&UnsafeRef::from(&expr_type.env)));
+				}
+				if let Some(expr_type) = object_type.as_interface() {
+					self.add_reference_symbol(property, Some(&UnsafeRef::from(&expr_type.env)));
+				}
```
MarkMcCulloh, line 192: I think as_env() can be used here to simplify things

```suggestion
				if let Some(inner_env) = object_type.as_env() {
					self.add_reference_symbol(property, Some(inner_env));
				}
```

Or even this, if the logic is the same

```suggestion
				self.add_reference_symbol(property, object_type.as_env());
```

MarkMcCulloh (approved):

libs/wingc/src/lsp/rename_visitor.rs
tsuf239: super seems to be handled already (probably never included)

libs/wingc/src/lsp/rename_visitor.rs
tsuf239: That's a good one! added :)

libs/wingc/src/lsp/rename_visitor.rs
tsuf239: opened #6693 I'll try again there

libs/wingc/src/lsp/rename_request.rs
tsuf239: opened #6694

PR #6642, opened 2024-06-05
fix(sdk): remove redundant test constructs from construct tree

yoav-steinberg:
This relates to #3168

Our framework duplicates the test environment for each `test` statement. The environment includes the tests themselves, so the `cloud.Function` resources created for each test are duplicated for each test. So given n tests we have n^2 cloud functions. This blows up the number of resources. This PR removes the unused `Test` (`cloud.Function`) resources generated in each test environment making test deployment much faster.

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/std/test-runner.ts
```diff
@@ -40,18 +40,51 @@ export class TestRunner extends Resource {
...
         new Root(app, "Default");
       }
     }
   }
 
+  /**
+   * Scan the given construct tree and disconnect nodes that don't match the predicate.
+   * @internal
+   */
+  public static _disconnectTree(
```
eladb, line 75: This feels wrong... I think there's already logic in the TestRunner that conditionally adds the `cloud.Function` only in the relevant environment. Perhaps there's a bug there?

See: https://github.com/winglang/wing/blob/c800787c2dac74a32079426fcfddf3acb058dc1c/libs/wingsdk/src/std/test-runner.ts#L118

eladb: At any rate, I would highly recommend against removing nodes from the construct tree. It's a very flaky business. Instead we just need to make sure that the `cloud.Function` of the test will only be created in the test environment it belongs.

An alternative approach to the `_addTestFunction()` we have is to expose a public method on `std.Test` called something like `enableTest()` and call it on the relevant test for each enviornment.

eladb (changes requested):

libs/wingsdk/src/std/test-runner.ts
yoav-steinberg: > This feels wrong... I think there's already logic in the TestRunner that conditionally adds the cloud.Function only in the relevant environment.

You're right, kind of surprising implementation. Anyway I'm scrapping this. :man_shrugging:

PR #6643, opened 2024-06-06
feat(console): implement resource running state indicators

### LINKED ISSUE #6622: feat(console): show resource status

skyrpex:
Closes #6622.


libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -867,25 +919,36 @@ export class Simulator {
...
+    try {
+      const attrs = await resourceObject.init(context);
+
+      this.state[path].attrs = {
+        ...this.state[path].attrs,
+        ...attrs,
+      };
+    } catch (error: any) {
+      initError = error;
+    }
```
Chriscbr, line 932: I think there are other parts of the simulator expecting to catch and handle initialization errors - I wonder if they need to be updated somehow somehow to be compatible with this change

https://github.com/winglang/wing/blob/cdfab81a04acc494f84d01b74838e8cba72fb964/libs/wingsdk/src/simulator/simulator.ts#L279-L309

skyrpex: What I've seen is that constructs (e.g. resource.inflight.ts) swallowed their errors, emitted an error trace and returned an empty attrs object.

If we want to shut everything down if a construct fails to initialize, the UX may be impacted (e.g. no bucket can be inspected while any construct failed).

apps/wing-console/console/ui/src/features/map-view.tsx
```diff
@@ -253,9 +274,15 @@ const ConstructNode: FunctionComponent<PropsWithChildren<ConstructNodeProps>> =
...
-                highlight && "outline-4 border-sky-400 dark:border-sky-500",
+                "outline outline-0",
+                !highlight &&
+                  !hasError &&
+                  "border-slate-200 dark:border-slate-800",
+                highlight && !hasError && "border-sky-400 dark:border-sky-500",
+                highlight && "outline-4",
+                !hasError && "outline-sky-200/50 dark:outline-sky-500/50",
+                hasError &&
+                  "border-red-500 dark:border-red-800 outline-red-200/50 dark:outline-red-500/50",
```
polamoros, line 285: The contrast is not good for nodes with errors in dark mode.
image

skyrpex: It's brighter in dark mode now.

skyrpex:
I think I'm also going to bubble upwards loading statuses (e.g. a child is loading, so the parent will show loading as well).

eladb:
image

In the side panel, instead of the red dot, can we add another "Label" called "Status" with a text status and a red dot?

skyrpex:
@eladb something like this?

image

eladb:
> @eladb something like this?
> 
> image

Yes! And it should be easy to click on this and see the errors (e.g. through the log filter when it will be implemented).

eladb:
@skyrpex can you share a screen recording of how this looks with the "loading" animation? I don't see this in the preview environment.

skyrpex:
> @skyrpex can you share a screen recording of how this looks with the "loading" animation? I don't see this in the preview environment.


https://github.com/winglang/wing/assets/1077520/fcce1c97-20b0-4597-859a-1c900ed22ebe

eladb (approved):

ainvoner (approved):

Issue #6645, opened 2024-06-06
Option to keep nodes expanded in node display


polamoros:
### Use Case

Some apps have a single container resource that is shown collapsed by default.

<img width="1512" alt="image" src="https://github.com/winglang/wing/assets/5547636/0a93f3c2-08df-4a65-a44a-86ab4c6cb782">

should look like
<img width="1512" alt="image" src="https://github.com/winglang/wing/assets/5547636/0492ea10-137b-4ac2-965e-cb7f4696200e">

### Proposed Solution

Add `display.collapsible` option.

### Component

Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Discord](https://t.winglang.io/discord).

eladb:
I think it's okay if we add `display.expandable` but I would also just add a heuristic to the Console so that if there's only a single root node, it will be expanded by default.

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.45](https://github.com/winglang/wing/releases/tag/v0.74.45).

PR #6653, opened 2024-06-06
feat(console): option to keep nodes expanded in node display

polamoros:
Resolves https://github.com/winglang/wing/issues/6645

This PR adds a new display setting to keep a node expanded by default:

```
let rateSchedule = new cloud.Schedule(cloud.ScheduleProps{
  rate: 5m
}) as "Rate Schedule";
nodeof(rateSchedule).expanded = true;
```

Also expands automatically if there is one single node


apps/wing-console/console/app/demo/main.w
```diff
@@ -106,6 +107,7 @@ let table = new ex.Table(
 let rateSchedule = new cloud.Schedule(cloud.ScheduleProps{
   rate: 5m
 }) as "Rate Schedule";
+nodeof(rateSchedule).collapsible = false;
```
eladb, line 110:
```suggestion
nodeof(rateSchedule).expanded = true;
```

I would use `expanded` instead of `collapsable`

libs/wingsdk/src/core/tree.ts
```diff
@@ -86,6 +86,11 @@ export interface DisplayInfo {
    * The icon of the resource in the UI.
    */
   readonly icon?: string;
+
+  /**
+   * Whether the resource is collapsible in the UI.
+   */
+  readonly collapsible?: boolean;
```
eladb, line 93:
```suggestion
  /**
   * The default view of this resource in the UI. By default, nodes are collapsed,
   * so set this to `true` if you want the node to be expanded by default.
   *
   * @default false
   */
  readonly expanded?: boolean;
```

eladb (commented):
I don't think we need a way to *block* someone from collapsing and expanding a node. Instead of I would change the semantic to `expanded` (defaults to `false`) and then if I want a node to be expanded by default then I can set this to `true`.

apps/wing-console/console/app/demo/main.w
polamoros: done 👍

eladb (approved):

apps/wing-console/console/ui/src/services/use-explorer.tsx
```diff
@@ -84,6 +85,32 @@ export const useExplorer = () => {
     setSelectedItems([selectedNode]);
   }, [selectedNode, setSelectedItems]);
 
+  const initialExpandedItems = useMemo(() => {
```
skyrpex, line 88: instead of creating a memo, you couñd move this code inside the use effect below

polamoros: changed 👍

libs/wingsdk/src/std/node.ts
```diff
@@ -89,6 +89,14 @@ export class Node {
    */
   public icon?: string;
 
+  /**
+   * The default view of this resource in the UI. By default, nodes are collapsed,
+   * so set this to `true` if you want the node to be expanded by default.
+   *
+   * @default false
+   */
+  public expanded?: boolean;
```
ainvoner, line 98: I think it will be a very strange behavior if external libs with custom resources will have the ability to change the default expanded state of a node.
This of the following use case:
 - loading the app, resource A is expanded by default.
 - The user collapse resource A since they would like to focus on a different area in their app.
 - User changes the code, the console reloads - resource A is expanded again.
 
 I don't like it. If we are trying to handle a case where my app has only 1 resource and it looks strange being collapsed. lets handle this use case in the Console.

ainvoner (changes requested):

libs/wingsdk/src/core/tree.ts
```diff
@@ -86,6 +86,14 @@ export interface DisplayInfo {
    * The icon of the resource in the UI.
    */
   readonly icon?: string;
+
+  /**
+   * The default view of this resource in the UI. By default, nodes are collapsed,
```
skyrpex, line 91: > The default view of this resource in the UI.

I'm not sure that phrase makes sense 🤔

polamoros: :)

apps/wing-console/console/ui/src/services/use-explorer.tsx
```diff
@@ -1,6 +1,6 @@
 import { ResourceIcon } from "@wingconsole/design-system";
 import type { ExplorerItem } from "@wingconsole/server";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
```
skyrpex, line 3: I think this is now unused.

skyrpex (approved):

PR #6656, opened 2024-06-08
chore: avoid full wasm optimizations in development

MarkMcCulloh:
The `-O3` level optimization for wingc's wasm is pretty slow but worth keeping in for the final build.
For local dev it's still valuable to run wasm-opt with some optimization as it greatly reduces the size of the wasm file, which then improves performance when JS loads it. The base level optimization runs in like 500ms on my machine so it's not much of a problem.

### DIFF
--- a/libs/wingc/package.json
+++ b/libs/wingc/package.json
@@ -2,7 +2,7 @@
   "name": "@winglang/wingc",
   "private": true,
   "scripts": {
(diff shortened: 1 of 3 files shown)


yoav-steinberg (approved):
thanks :+1:

tsuf239:
I've changed some code in `jsify.rs` on a local branch and `pnpm wing test` and `pnpm wing compile` don't seem to include those changes, I think it's related :(

Issue #4470, opened 2023-10-09
Unexpected optional chaining behavior with macro functions


skyrpex:
### I tried this:

```ts
struct Result {
  item: Json?;
}

test "KO: fails to compile" {
  let result = Result {};

  assert(result.item?.get("id")?.asStr() == nil);
  //                ^ I shouldn't need the "?" here.
  // If we remove it, it fails to compile.
}

test "KO: runtime error" {
  let result = Result {};

  assert(result.item?.get("id")?.asStr() == nil);
  // Unknown error: Cannot read properties of undefined (reading 'id')
  // 
  // It should return `nil`. If `item` exists, we should be able to chain without additional "?".
}

test "OK: optional chaining behavior" {
  let result = Result {
    item: {
      id: "user1",
    },
  };

  assert(result.item?.get("id")?.asStr() == "user1");
  //                           ^ Still, the "?" shouldn't be necessary.
}
```

[...]

PR #6659, opened 2024-06-09
fix(compiler): optional macro chaining

tsuf239:
closes [#4470](https://github.com/winglang/wing/issues/4470)
closes [#6662](https://github.com/winglang/wing/issues/6662) as well! 🎉 

I inserted a default value for optional maps, json, sets, and structs
below you can find the compilation output of the following wing code, comments and newlines added later and not a part of the original compilation output.

```wing
struct Result {
  item: Json?;
  items: Array<Json>?;
  mapItem: Map<str>?;
  mapItems: Array<Map<str>>?;
  setItem: Set<num>?;
  setItems: Array<Set<num>>?;
  structItem: Result?;
  structItems: Array<Result>?;
}

test "optional chaining macros" {
  let result = Result {};

  expect.equal(result.item?.tryGet("id"), nil);
  expect.equal(result.item?.has("id"), false);

  expect.equal(result.items?.tryAt(0)?.tryGet("id"), nil);
  expect.equal(result.items?.tryAt(0)?.has("id"), false);

  expect.equal(result.mapItem?.tryGet("a"), nil);
  expect.equal(result.mapItem?.has("id"), false);

  expect.equal(result.mapItems?.tryAt(0)?.tryGet("id"), nil);
  expect.equal(result.mapItems?.tryAt(0)?.has("id"), false);

  expect.equal(result.setItem?.size, nil);
  expect.equal(result.setItem?.has(6), nil);

  expect.equal(result.setItems?.tryAt(0)?.size, nil);
  expect.equal(result.setItems?.tryAt(0)?.has(6), nil);

  expect.equal(result.structItem?.item, nil);
  expect.equal(result.structItems?.tryAt(0)?.item, nil);
}
```

```js
       // expect.equal(result.item?.tryGet("id"), nil);
      (await $expect_Util.equal(((result.item ?? {}))?.["id"], undefined));
      // expect.equal(result.item?.has("id"), false);
      (await $expect_Util.equal(((obj, key) => { return obj.hasOwnProperty(key); })((result.item ?? {}),"id"), false));

      // expect.equal(result.items?.tryAt(0)?.tryGet("id"), nil);
      (await $expect_Util.equal((((result.items ?? []).at(0) ?? {}))?.["id"], undefined));
      // expect.equal(result.items?.tryAt(0)?.has("id"), false);
      (await $expect_Util.equal(((obj, key) => { return obj.hasOwnProperty(key); })(((result.items ?? []).at(0) ?? {}),"id"), false));
     
      // expect.equal(result.mapItem?.tryGet("a"), nil);
      (await $expect_Util.equal(((result.mapItem ?? new Map()))["a"], undefined));
      // expect.equal(result.mapItem?.has("id"), false);
      (await $expect_Util.equal(("id" in ((result.mapItem ?? new Map()))), false));

      // expect.equal(result.mapItems?.tryAt(0)?.tryGet("id"), nil);
      (await $expect_Util.equal((((result.mapItems ?? []).at(0) ?? new Map()))["id"], undefined));
       // expect.equal(result.mapItems?.tryAt(0)?.has("id"), false);
      (await $expect_Util.equal(("id" in (((result.mapItems ?? []).at(0) ?? new Map()))), false));

       // expect.equal(result.setItem?.size, nil);
      (await $expect_Util.equal(result.setItem?.size, undefined));
      // expect.equal(result.setItem?.has(6), nil);
      (await $expect_Util.equal((await result.setItem?.has?.(6)), undefined));

      // expect.equal(result.setItems?.tryAt(0)?.size, nil);
      (await $expect_Util.equal((result.setItems ?? []).at(0)?.size, undefined));
      //  expect.equal(result.setItems?.tryAt(0)?.has(6), nil);
      (await $expect_Util.equal((await (result.setItems ?? []).at(0)?.has?.(6)), undefined));

      //  expect.equal(result.structItem?.item, nil);
      (await $expect_Util.equal(result.structItem?.item, undefined));
     // expect.equal(result.structItems?.tryAt(0)?.item, nil);
      (await $expect_Util.equal((result.structItems ?? []).at(0)?.item, undefined));
```

the other thing that needs to be changed in the issue's example is to change the `get` to `tryGet` since the key we try to reach is non-existent. 

## Checklist

- [ ] Tests added (always)


tools/hangar/__snapshots__/test_corpus/valid/optional_macros.test.w_compile_tf-aws.md
```diff
@@ -0,0 +1,111 @@
...
+      (await $expect_Util.equal(((result.mapItem ?? new Map()))["a"], undefined));
+      (await $expect_Util.equal(("id" in ((result.mapItem ?? new Map()))), false));
+      (await $expect_Util.equal((((result.mapItems ?? []).at(0) ?? new Map()))["id"], undefined));
+      (await $expect_Util.equal(("id" in (((result.mapItems ?? []).at(0) ?? new Map()))), false));
+      (await $expect_Util.equal(result.setItem?.size, undefined));
+      (await $expect_Util.equal((await result.setItem?.has?.(6)), undefined));
+      (await $expect_Util.equal((result.setItems ?? []).at(0)?.size, undefined));
+      (await $expect_Util.equal((await (result.setItems ?? []).at(0)?.has?.(6)), undefined));
+      (await $expect_Util.equal(result.structItem?.item, undefined));
+      (await $expect_Util.equal((result.structItems ?? []).at(0)?.item, undefined));
```
Chriscbr, line 29: I'm not sure this is the right solution. By falling back to an empty value, like an empty array or empty object, I think it can result in the wrong runtime behavior:

```
let s: Set<num>? = nil;
let a = s?.toArray();
log(a?.length ?? -1);
```

If "a" evaluates to an empty array, the log function will print 0 instead of -1, which is a bug. "a" should evaluate to "nil" in Wing, or "undefined" in JavaScript. That way "a?.length" will evaluate to "undefined" and the log function will show "-1", like in TypeScript:

```js
class Foo {
  toArray() { return []; }
}

function makeOptionalFoo(): Foo | undefined {
  return undefined;
}

let x = makeOptionalFoo();
let y = x?.toArray();
console.log(y?.length ?? -1); // "-1"
```

I think we need the emitted code to check if the expression is null or undefined, and only run the macro code when it's not one of those values. Perhaps something like

```js
foo?.bar(args)
// ->
(($value) => $value != undefined ? <bar-macro-code> : $value)(foo, args)
```

Chriscbr (changes requested):

tsuf239:
here is the new compilation output:
```
 // expect.equal(result.item?.tryGet("id"), nil);
await $expect_Util.equal(result.item === undefined ? undefined : result.item?.["id"], undefined);
 // expect.equal(result.item?.has("id"), nil);
      (await $expect_Util.equal((result.item === undefined ? undefined : ((obj, key) => { return obj.hasOwnProperty(key); })(result.item,"id")), undefined));

 // expect.equal(result.items?.tryAt(0)?.tryGet("id"), nil);
      (await $expect_Util.equal((result.items?.at(0) === undefined ? undefined : (result.items?.at(0))?.["id"]), undefined));
 // expect.equal(result.items?.tryAt(0)?.has("id"), nil);
      (await $expect_Util.equal((result.items?.at(0) === undefined ? undefined : ((obj, key) => { return obj.hasOwnProperty(key); })(result.items?.at(0),"id")), undefined));

  // expect.equal(result.mapItem?.tryGet("a"), nil);
      (await $expect_Util.equal((result.mapItem === undefined ? undefined : (result.mapItem)["a"]), undefined));
  // expect.equal(result.mapItem?.has("id"), nil);
      (await $expect_Util.equal((result.mapItem === undefined ? undefined : ("id" in (result.mapItem))), undefined));

  // expect.equal(result.mapItems?.tryAt(0)?.tryGet("id"), nil);
      (await $expect_Util.equal((result.mapItems?.at(0) === undefined ? undefined : (result.mapItems?.at(0))["id"]), undefined));
  // expect.equal(result.mapItems?.tryAt(0)?.has("id"), nil);
      (await $expect_Util.equal((result.mapItems?.at(0) === undefined ? undefined : ("id" in (result.mapItems?.at(0)))), undefined));

  // expect.equal(result.setItem?.size, nil);
      (await $expect_Util.equal(result.setItem?.size, undefined));
  // expect.equal(result.setItem?.has(6), nil);
      (await $expect_Util.equal((await result.setItem?.has?.(6)), undefined));

  // expect.equal(result.setItems?.tryAt(0)?.size, nil);
      (await $expect_Util.equal((result.setItems === undefined ? undefined : result.setItems?.at(0))?.size, undefined));
  // expect.equal(result.setItems?.tryAt(0)?.has(6), nil);
      (await $expect_Util.equal((await (result.setItems === undefined ? undefined : result.setItems?.at(0))?.has?.(6)), undefined));

  // expect.equal(result.structItem?.item, nil);
      (await $expect_Util.equal(result.structItem?.item, undefined));
  // expect.equal(result.structItems?.tryAt(0)?.item, nil);
      (await $expect_Util.equal((result.structItems === undefined ? undefined : result.structItems?.at(0))?.item, undefined));

```

tools/hangar/__snapshots__/test_corpus/valid/optional_macros.test.w_compile_tf-aws.md
```diff
@@ -0,0 +1,111 @@
...
+module.exports = function({ $expect_Util }) {
+  class $Closure1 {
+    constructor({  }) {
+      const $obj = (...args) => this.handle(...args);
+      Object.setPrototypeOf($obj, this);
+      return $obj;
+    }
+    async handle() {
+      const result = ({});
+      (await $expect_Util.equal((result.item === undefined ? undefined : (result.item)?.["id"]), undefined));
```
Chriscbr, line 16: This looks a lot better. I think there's only one issue left -- the part of the expression before the `?.` is emitted twice, so it will get evaluated twice by accident:

```
let makeArray = (): Array<num> => {
  log("hi");
  return [1, 2, 3];
};

log(makeArray()?.contains(2) ?? true); // prints "hi" twice
```

Chriscbr (changes requested):

tsuf239:
reached to the cleanest solution by changing map's `has` into the js version (abstract) and json's `has` to accept optional (there is nothing we can do in the justification process to get more precise result, more than changing the macros their self to be more precise)

tsuf239:
```
 // expect.equal(result.item?.tryGet("id"), nil);
 (await $expect_Util.equal((result.item)?.["id"], undefined));
//  expect.equal(result.item?.has("id"), false);
(await $expect_Util.equal(((obj, key) => { return obj?.hasOwnProperty(key); })(result.item,"id"), undefined));

  // expect.equal(result.items?.tryAt(0)?.tryGet("id"), nil);
      (await $expect_Util.equal((result.items?.at(0))?.["id"], undefined));
  // expect.equal(result.items?.tryAt(0)?.has("id"), false);
 (await $expect_Util.equal(((obj, key) => { return obj?.hasOwnProperty(key); })(result.items?.at(0),"id"), undefined));

  // expect.equal(result.mapItem?.tryGet("a"), nil);
 (await $expect_Util.equal((result.mapItem)?.["a"], undefined));
  // expect.equal(result.mapItem?.has("id"), false);
      (await $expect_Util.equal((await result.mapItem?.has?.("id")), undefined));

  // expect.equal(result.mapItems?.tryAt(0)?.tryGet("id"), nil);
      (await $expect_Util.equal((result.mapItems?.at(0))?.["id"], undefined));
  // expect.equal(result.mapItems?.tryAt(0)?.has("id"), false);
      (await $expect_Util.equal((await result.mapItems?.at(0)?.has?.("id")), undefined));

  // expect.equal(result.setItem?.size, nil);
      (await $expect_Util.equal(result.setItem?.size, undefined));
  // expect.equal(result.setItem?.has(6), nil);
      (await $expect_Util.equal((await result.setItem?.has?.(6)), undefined));

  // expect.equal(result.setItems?.tryAt(0)?.size, nil);
      (await $expect_Util.equal(result.setItems?.at(0)?.size, undefined));
  // expect.equal(result.setItems?.tryAt(0)?.has(6), nil);
      (await $expect_Util.equal((await result.setItems?.at(0)?.has?.(6)), undefined));

 // expect.equal(result.structItem?.item, nil);
      (await $expect_Util.equal(result.structItem?.item, undefined));
 // expect.equal(result.structItems?.tryAt(0)?.item, nil);
      (await $expect_Util.equal(result.structItems?.at(0)?.item, undefined));     
```

tools/hangar/__snapshots__/test_corpus/valid/optional_macros.test.w_compile_tf-aws.md
tsuf239: added

Chriscbr:
> reached to the cleanest solution by changing map's `has` into the js version (abstract) and json's `has` to accept optional (there is nothing we can do in the justification process to get more precise result, more than changing the macros their self to be more precise)

I still think we should look into handle this automatically through jsification. Here's an example of another macro that's still not handled, so it fails to compile:

```js
let makeArray = (): Array<num> => {
  log("makeArray called");
  return [1, 2, 3];
};
let makeArray2 = (): Array<num>? => {
  log("makeArray2 called");
  return nil;
};

log(makeArray().at(1));
log(makeArray2()?.at(1) ?? -1);
```

We have over 80 macros in the standard library (and growing), and writing tests for each one in an optional chaining position feels like effort worth avoiding.

------

I'm wondering if you can give another try with the code generation approach from this comment? https://github.com/winglang/wing/pull/6659#discussion_r1632581331

Btw - I wonder if we can make the emitted code a bit more readable by, say, generating a global function for each macro (I believe a lot of JavaScript bundlers do this):

```js
"use strict";
// since Array.at is used in the Wing source code and the method is a macro,
// the compiler emits a global function named `__Array_at`
// if `Array.at` is called multiple times, `__Array_at` is only added once
let __Array_at = function (skipIfNil, $self$, ...$args$) {
  // optional chaining behavior
  if (skipIfNil && $self$ == undefined) return $self$;
  // return <macro text>;
  return ((arr, index) => { if (index < 0 || index >= arr.length) throw new Error("Index out of bounds"); return arr[index]; })($self$, $args$);
}
const makeArray = (() => {
  console.log("makeArray called");
  return [1, 2, 3];
});
const makeArray2 = (() => {
  console.log("makeArray2 called");
  return undefined;
});
console.log(__Array_at(false, makeArray(), 1));
console.log(__Array_at(true, makeArray2(), 1) ?? (-1));
```

tsuf239:
@Chriscbr do you think it's better to add those in the inflight file itself? or to parse and bundle the macro fn files in the SDK itself, and then import them straight from there? (less repetitive work for the compiler and potential clashes, in a tradeoff of very-slightly bigger bundle size )

Chriscbr:
> @Chriscbr do you think it's better to add those in the inflight file itself? or to parse and bundle the macro fn files in the SDK itself, and then import them straight from there? (less repetitive work for the compiler and potential clashes, in a tradeoff of very-slightly bigger bundle size )

Parsing and bundling all of the macro fns into one file also sounds like a good approach to me.

tsuf239:
```
// expect.equal(result.item?.tryGet("id"), nil);
(await $expect_Util.equal($macros.__Json_tryGet(true, result.item, "id"), undefined));
//  expect.equal(result.item?.has("id"), nil);
(await $expect_Util.equal($macros.__Json_has(true, result.item, "id"), undefined));

// expect.equal(result.items?.tryAt(0)?.tryGet("id"), nil);
(await $expect_Util.equal($macros.__Json_tryGet(true, $macros.__Array_tryAt(true, result.items, 0), "id"), undefined));
// expect.equal(result.items?.tryAt(0)?.has("id"), nil);
(await $expect_Util.equal($macros.__Json_has(true, $macros.__Array_tryAt(true, result.items, 0), "id"), undefined));

// expect.equal(result.mapItem?.tryGet("a"), nil);
(await $expect_Util.equal($macros.__Map_tryGet(true, result.mapItem, "a"), undefined));
 // expect.equal(result.mapItem?.has("id"), nil);
(await $expect_Util.equal((await result.mapItem?.has?.("id")), undefined));
 
 // expect.equal(result.mapItems?.tryAt(0)?.tryGet("id"), nil);
(await $expect_Util.equal($macros.__Map_tryGet(true, $macros.__Array_tryAt(true, result.mapItems, 0), "id"), undefined));
// expect.equal(result.mapItems?.tryAt(0)?.has("id"), nil);
(await $expect_Util.equal((await $macros.__Array_tryAt(true, result.mapItems, 0)?.has?.("id")), undefined));

// expect.equal(result.setItem?.size, nil);
(await $expect_Util.equal(result.setItem?.size, undefined));
// expect.equal(result.setItem?.has(6), nil);
(await $expect_Util.equal((await result.setItem?.has?.(6)), undefined));

// expect.equal(result.setItems?.tryAt(0)?.size, nil);
(await $expect_Util.equal($macros.__Array_tryAt(true, result.setItems, 0)?.size, undefined));
// expect.equal(result.setItems?.tryAt(0)?.has(6), nil);
(await $expect_Util.equal((await $macros.__Array_tryAt(true, result.setItems, 0)?.has?.(6)), undefined));

// expect.equal(result.structItem?.item, nil);
(await $expect_Util.equal(result.structItem?.item, undefined));
// expect.equal(result.structItems?.tryAt(0)?.item, nil);
(await $expect_Util.equal($macros.__Array_tryAt(true, result.structItems, 0)?.item, undefined));
     
 let calls = 0;
      const makeArray = (async () => {
        calls = (calls + 1);
        return [1, 2, 3];
      });

// expect.ok(makeArray()?.contains(2) ?? false);
(await $expect_Util.ok(($macros.__Array_contains(true, (await makeArray()), 2) ?? false)));
// expect.equal(calls, 1);
(await $expect_Util.equal(calls, 1));

```

libs/wingsdk/scripts/generate-macros.mts
```diff
@@ -0,0 +1,45 @@
...
+          if (skipIfNil && $self$ === undefined) return $self$;
+          return ${macro}
+        }
+        `;
+      }
+    }
+  }
+}
+
+project.createSourceFile("lib/macros.js", macroFileContent, {
```
Chriscbr, line 41: Is it possible to generate it as a typescript file inside `src/` so that the file will be saved to git? If there are type checking issues, maybe they can be ignored by generating `// @ts-ignore` or by adding the file name to the excludes section in `tsconfig.json`.

apps/wing/src/commands/pack.test.ts
```diff
@@ -186,10 +186,18 @@ describe("wing pack", () => {
...
+    await fs.symlink(
+      require.resolve("@winglang/sdk/lib/index.js"),
+      join(outdir, "package", "node_modules", "@winglang", "sdk", "index.js")
+    );
+
     await fs.symlink(
-      require.resolve("@winglang/sdk"),
-      join(outdir, "package", "node_modules", "@winglang", "sdk")
+      require.resolve("@winglang/sdk/lib/macros.js"),
+      join(outdir, "package", "node_modules", "@winglang", "sdk", "lib", "macros.js")
```
Chriscbr, line 200: See my other comment -- if `src/macro.ts` is compiled with the rest of the files (and exported in `src/index.ts`) then maybe we won't need to special case it here

libs/wingsdk/src/std/array.ts
```diff
@@ -42,7 +42,7 @@ export class Array {
   /**
    * Get the value at the given index, returning nil if the index is out of bounds.
    *
-   * @macro $self$.at($args$)
+   * @macro $self$?.at($args$)
```
Chriscbr, line 45: I'm not sure if this is needed
```suggestion
   * @macro $self$.at($args$)
```

libs/wingsdk/src/std/json.ts
```diff
@@ -159,7 +159,7 @@ export class Json {
   /**
    * Checks if a Json object has a given key
    *
-   * @macro ((obj, key) => { return obj.hasOwnProperty(key); })($self$,$args$)
+   * @macro ((obj, key) => { return obj?.hasOwnProperty(key); })($self$,$args$)
```
Chriscbr, line 162: Same here
```suggestion
   * @macro ((obj, key) => { return obj.hasOwnProperty(key); })($self$,$args$)
```

Chriscbr (approved):

libs/wingsdk/scripts/generate-macros.mts
tsuf239: it can create a js file and save it as a ts, there are plenty of unsolvable (tried!) type-checking issues, and the file itself- since it's actually js, and un-linted, looks very ugly :( 
Do you wish to commit it anyway? for verification?

Chriscbr: I see - let's try leaving it as is then

Issue #6124, opened 2024-04-03
Wing Console application “reset” capability


eladb:
### Use Case

I’d like to be able to reset the state and completely restart my wing application in the wing console.

This is in case the state is “dirty” or I want to recover from a bad update (can be caused by eg a malfunctioning `cloud.Service` for example).

### Proposed Solution

A nice “reset” button in a place that makes sense.

### Component

Wing Console

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

ainvoner:
and analytics for how many times this button was clicked

monadabot:
Congrats! :rocket: This was released in [Wing 0.74.60](https://github.com/winglang/wing/releases/tag/v0.74.60).

PR #6665, opened 2024-06-10
feat(console): wing console application reset capability

polamoros:
Resolves https://github.com/winglang/wing/issues/6124


polamoros:
Thoughts?

@skyrpex @ainvoner @eladb 

image

image

apps/wing-console/console/server/src/router/app.ts
```diff
@@ -53,6 +53,13 @@ export const createAppRouter = () => {
         config: ctx.layoutConfig,
       };
     }),
+    "app.restart": createProcedure.mutation(async ({ ctx }) => {
```
eladb, line 56: I can't see this log when I am playing around with this...

eladb (changes requested):
@polamoros I am not a fan of the new header.

I think the "RESET" button should be on the top right of the map view:

Screenshot 2024-06-10 at 16 20 31

And it should just say "Reset".

When clicking on it, there should be a message box:

```
Are you sure you want to reset all state and restart the application?

YES/NO
```

The Discord button can stay on the footer AFAIK for now.

polamoros:
> @polamoros I am not a fan of the new header.
> 
> I think the "RESET" button should be on the top right of the map view:
> 
> Screenshot 2024-06-10 at 16 20 31
> 
> And it should just say "Reset".
> 
> When clicking on it, there should be a message box:
> 
> ```
> Are you sure you want to reset all state and restart the application?
> 
> YES/NO
> ```
> 
> The Discord button can stay on the footer AFAIK for now.

I don't like the header either. Regarding moving the reset button to the map controls, I don't think it's the correct solution. We are restarting the application, not just the map. 
I would prefer moving it to the footer or adding a menu on the right side of the header (if we decide to keep the header).

Here's an example
image


Regarding the Discord button, @ainvoner wants to place it in a more visible location.

skyrpex:
> @polamoros I am not a fan of the new header.
> 
> I think the "RESET" button should be on the top right of the map view:
> 
> Screenshot 2024-06-10 at 16 20 31
> 
> And it should just say "Reset".
> 
> When clicking on it, there should be a message box:
> 
> ```
> Are you sure you want to reset all state and restart the application?
> 
> YES/NO
> ```
> 
> The Discord button can stay on the footer AFAIK for now.

It doesn't make sense to me to have such a functionality inside the map view: the reset action affects the whole application, not just the map view. Also, it may signal to the users that it only resets the view (same as the [] button on the right side).

Even if a new header is almost empty, it makes sense to place the reset button there IMO.

polamoros:
What about moving it to the left side of the footer?

I think it makes more sense.

image

skyrpex:
We could also merge the functionality of the reset button and the Status label. They are similar.

skyrpex:
It's strange that the Status label isn't updated when the app is restarting 🤔

apps/wing-console/console/server/src/router/app.ts
polamoros: it's a verbose message
image

skyrpex:
I think the restart button should be also disabled when the simulator is loading (e.g. after editing the wing file).

eladb (commented):
LGTM

apps/wing-console/console/ui/src/layout/default-layout.tsx
```diff
@@ -20,6 +20,7 @@ import { Explorer } from "../ui/explorer.js";
 import { ResourceMetadata } from "../ui/resource-metadata.js";
 import { LogsWidget } from "../widgets/logs.js";
 
+import { Header } from "./header.js";
```
skyrpex, line 23: Unused.

apps/wing-console/console/ui/src/layout/status-bar.tsx
```diff
@@ -65,11 +82,36 @@ export const StatusBar = ({
           </span>
         </div>
       </div>
+      {/*center*/}
+      <div className="w-full flex justify-center items-center"></div>
```
skyrpex, line 86:
```suggestion
      <div className="grow"></div>
```

You can use `grow` so it occupies all remaining space.

apps/wing-console/console/ui/src/layout/status-bar.tsx
```diff
@@ -65,11 +82,36 @@ export const StatusBar = ({
...
       {/*right side*/}
-      <div className="w-full flex space-x-1 justify-end">
+      <div className="w-full flex gap-2 items-center justify-end">
```
skyrpex, line 88:
```suggestion
      <div className="flex gap-2 items-center">
```

It isn't necessary for the div to grow, and `justify-end` doesn't have any effect here either.

apps/wing-console/console/ui/src/layout/status-bar.tsx
```diff
@@ -29,18 +33,31 @@ export const StatusBar = ({
...
         theme.text1,
         theme.border3,
-        "py-1 px-4 flex text-2xs w-full relative z-10",
+        "px-4 flex items-center text-2xs w-full relative z-10",
       )}
     >
       {/*left side*/}
-      <div className="w-full flex space-x-6">
-        <div title={wingVersion} className="truncate space-x-1 min-w-[7rem]">
+      <div className="w-full flex gap-4 items-center">
```
skyrpex, line 55:
```suggestion
      <div className="flex gap-2 items-center">
```

No need to grow. Let's standardize the gap to 2 (in sync with the rest of the status bar.

apps/wing-console/console/ui/src/layout/status-bar.tsx
```diff
@@ -29,18 +33,31 @@ export const StatusBar = ({
...
+
   return (
     <footer
       className={classNames(
-        theme.bg3,
+        theme.bg1,
         theme.text1,
         theme.border3,
-        "py-1 px-4 flex text-2xs w-full relative z-10",
+        "px-4 flex items-center text-2xs w-full relative z-10",
```
skyrpex, line 51:
```suggestion
        "px-4 flex gap-2 items-center text-2xs w-full relative z-10",
```

Standardize gap. This gap would apply if the width of the window is small.

apps/wing-console/console/ui/src/layout/status-bar.tsx
```diff
@@ -65,11 +82,36 @@ export const StatusBar = ({
...
+              application?
+            </p>
+            <div className="flex justify-around gap-2">
+              <Button onClick={() => setShowRestartModal(false)}>Cancel</Button>{" "}
+              <Button onClick={restart} dataTestid="restart-simulator-button">
+                Reset
+              </Button>
+            </div>
+          </div>
+        </Modal>
```
skyrpex, line 114: Better to move it out of the flex container. Even though it uses portals behind the hood, I think the flex gap may apply when the modal is visible.

skyrpex (changes requested):

apps/wing-console/console/ui/src/layout/status-bar.tsx
polamoros: done!

skyrpex (approved):

PR #6676, opened 2024-06-10
chore(repo): upgrade vitest to v1.6.0

### LINKED ISSUE #6675: Update to vitest v1 throughout monorepo
Currently it's not possible to use the vitest VS Code extension in the repo since Vitest v1.4.0 or newer is required.

As of this posting we're using Vitest v1.3.1 in some parts of the repo, and v0.34 in others.

~Vitest v1 doesn't support changing directories in worker threads, so we'll need to find a workaround or rewrite tests in the repo that use APIs like `process.chdir()` (see https://github.com/vitest-dev/vitest/issues/1436)~

Edit: It looks like the CLI options changed (`--no-threads` isn't available anymore), so now you can use `--pool forks` as a workaround

Chriscbr:
Closes #6675

I thought for a hot minute that vitest deprecated thread-based workers and so this change would have required a lot more refactoring, thankfully they still support fork-based workers, it's just under a different option now. (Many tests within the Wing CLI package depend on `process.chdir()`).

## Checklist

- [x] Tests added (always)


package.json
```diff
@@ -25,7 +25,8 @@
   },
   "pnpm": {
     "overrides": {
-      "mime": "^3.0.0"
+      "mime": "^3.0.0",
+      "rollup": "^4.18.0"
```
MarkMcCulloh, line 29: Instead of this, another option is to change the tsconfig.json in apps/wing to have this at the end:
```
  "exclude": ["src/**/*.test.ts"]
```

This is a good idea anyways because it avoids including our tests in the final `dist` (I just noticed that 😱)

MarkMcCulloh: I just tried it locally and it worked for me. I tried to do this vitest update a while ago and wasn't able to fix this, but finally saw this comment https://github.com/rollup/rollup/issues/5199#issuecomment-2100534676

Chriscbr: Sure, yeah let me try that (also it seems like this override isn't working like I expected)

apps/wing/tsconfig.json
```diff
@@ -21,8 +21,8 @@
     "emitDecoratorMetadata": true,
     "declaration": true,
     "baseUrl": ".",
-    "esModuleInterop": true
+    "esModuleInterop": true,
   },
   "include": ["src/**/*"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", "src/**/*.test.ts"]
```
MarkMcCulloh, line 27: nit: `node_modules` isn't needed here ("exclude" only filters based on "include")

MarkMcCulloh (approved):
Thanks for doing the update 🙏 

For the current CI error with the benchmarks, you should be able to import and extend JsonReporter from "vitest/reporters" in `tools/hangar/src/benchmarking/reporter.ts`. This export was previously hidden weirdly inside `BenchmarkReportsMap.json`

apps/wing/tsconfig.json
Chriscbr: Ah - makes sense!

polamoros (approved):

PR #6686, opened 2024-06-12
chore(repo): upgrade vitest (attempt 2)

### LINKED ISSUE #6675: Update to vitest v1 throughout monorepo

Chriscbr:
https://github.com/winglang/wing/pull/6676 was reverted due to some unexpected build errors. This PR brings back the code with fixes to the benchmark script and to stack trace output on Windows machines.

Huge thanks to @MarkMcCulloh for the help getting this PR over the finish line.

Closes #6675

## Checklist

- [ ] Tests added (always)


polamoros (approved):

apps/wing-console/console/app/src/index.ts
```diff
@@ -65,8 +65,8 @@ export const createConsoleApp = async (options: CreateConsoleAppOptions) => {
     analyticsAnonymousId: analyticsStorage.getAnonymousId(),
     analytics,
     async requireSignIn() {
-      if (options.requireSignIn === false) {
-        return false;
+      if (options.requireSignIn) {
+        return options.requireSignIn;
       }
```
skyrpex, line 70: This is removing a very important behavior: skipping the sign in prompt when the env variable `NO_SIGN_IN` is `true`.

polamoros: isn't this `requireSignIn` flag used to force the modal behaviour? I think it is not passed by default so it will go through the default behavior.

skyrpex: Ahh yes, but `if (options.requireSignIn !== undefined)`

polamoros: this should be 
```
if (options.requireSignIn !== undefined) {
  return options.requireSignIn;
}
```

PR #6698, opened 2024-06-13
feat(sdk): allow implementing `sim.Resource` without shutdown activities

### LINKED ISSUE #6697: Allow implementing `sim.Resource` without an `onStop` method
### Use Case

I want to be able to define a `sim.Resource` without a stopping method:

```js
let resource = new sim.Resource(inflight (ctx) => {
  log("initializing...");
  ctx.resolveToken("foo", "bar");
});

let foo = resource.createToken("foo");
```

### Proposed Solution

The factory closure that `sim.Resource` should not require a return value. If none is provided, then it should behave as if there's simply no `onStop` method.

### Implementation Notes

_No response_

### Component

_No response_

### Community Notes

eladb:
Fixes #6697 by changing the return type of the resource factory to optional.

## Checklist

- [x] Tests added (always)


libs/wingsdk/src/target-sim/resource.ts
```diff
@@ -277,7 +277,7 @@ export interface IResourceFactoryClient {
    * Function that will be called to create the resource.
    * @inflight
    */
-  handle(context: IResourceContext): Promise<IResource>;
+  handle(context: IResourceContext): Promise<IResource | undefined>;
```
Chriscbr, line 280: Can we add something to the docs maybe? As a rule of thumb when we're allowing callback functions to return nothing, I feel like it's worth making it clear what the nil behavior means

Chriscbr (approved):
LGTM.

It looks like the main change here is that the factory function expected by sim.Resource no longer needs to return a backend. Put another way, the factory will fall back to a no-op backend. Maybe the PR title should be updated? (Right now if the factory does return a backend, it will still need to implement the onStop method technically, so it could be misleading.)

To make that optional, I think we'd have to make the onStop method optional in this interface or split it into a separate IResourceWithOnStop interface: 

https://github.com/winglang/wing/blob/d5fadbc9395a7b4012c9b0198bdd26b75e78e1d4/libs/wingsdk/src/target-sim/resource.ts#L24-L29

libs/wingsdk/src/target-sim/resource.ts
eladb: Done

PR #6699, opened 2024-06-14
fix(compiler): adding statements before super()

### LINKED ISSUE #6387: Cannot add statements before `super()` statement
### I tried this:

```js
class A {
  new(x: num) {}
}

class B extends A {
  new() {
    let x = 3 + 5;
    super(x);
  }
}
```

### This happened:

```
error: Call to super constructor must be first statement in constructor
  --> wing/main.w:8:5
  |
8 |     super(x);
  |     ^^^^^^^^^ Call to super constructor must be first statement in constructor
```

### I expected this:

I expected the example to compile. This kind of code is supported in TypeScript:

```js
class A {
  constructor(foo: number) {}
}

class B extends A {
  constructor() {
    let x = Math.random();
    super(x);
  }
}
```

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

0.73.13

### Node.js Version

20.11.1

### Platform(s)

MacOS

### Community Notes

marciocadev:
This PR allows statements to be executed before calling `super()`, with the exception of any statement that includes the reserved word `this` or `super` and the `super()` call is only valid when called in the top scope of the constructor.

The PR also cleans up some redundant `super()` related diagnostics generated in _parser.rs_ they were also handled (and more correctly) in _type_check.rs_.

Closes #6387

## Checklist

- [x] Tests added (always)


examples/tests/invalid/invalid_call_before_super.w
```diff
@@ -0,0 +1,44 @@
...
+//  ^^^^^^^^^^^^^^ This command cannot be executed before calling super()
+    super(5);   
+  }
+}
+
+class D extends A {
+    new(x: bool) {
+      if x {
+        super(1);
+//      ^^^^^^^^^ Call to super constructor can only be done from within class constructor
```
yoav-steinberg, line 28: This error message is a bit misleading because actually are within the class ctor, perhaps something like:
"Call to super constructor can only be done at the top level scope of the constructor"

libs/wingc/src/jsify.rs
```diff
@@ -1868,12 +1868,14 @@ impl<'a> JSifier<'a> {
...
-			false
-		};
+		// Check if some of the statements is the super constructor call, if not we need to add one
+		// as the first statement.
+		let mut super_called = false;
+		for s in init_statements.statements.iter() {
+			if matches!(s.kind, StmtKind::SuperConstructor { .. }) {
+				super_called = true;
+			}
+		}
```
yoav-steinberg, line 1878: nit
```suggestion
		let super_called = init_statements
			.statements
			.iter()
			.any(|s| matches!(s.kind, StmtKind::SuperConstructor { .. }));
```

libs/wingc/src/type_check.rs
```diff
@@ -4667,34 +4667,25 @@ new cloud.Function(@inflight("./handler.ts"), lifts: { bucket: ["put"] });
...
-									"super() call must be the first statement of {}'s constructor",
-									ast_class.name
-								),
-							);
+						// Check if one of the statements before the super() call is invalid
+						for i in 0..idx {
+							if self.type_check_invalid_stmt_before_super(&ctor_body.statements[i]) {
+								self.spanned_error(
+									&ctor_body.statements[i],
+									"This command cannot be executed before calling super()".to_string(),
```
yoav-steinberg, line 4681: It might be worth mentioning **why** this statement is illegal. Something like:
"Statements referencing `this` or `super` cannot be executed before calling `super()`"

yoav-steinberg, line 4681: I'm guessing you don't need `to_string` here:
```suggestion
									"This command cannot be executed before calling super()",
```

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2971,6 +2971,63 @@ exports[`invalid compile directory 1`] = `
 "
 `;
 
+exports[`invalid_call_before_super.w 1`] = `
+"error: Call to super constructor can only be done from within class constructor
```
Chriscbr, line 2975: Perhaps
```suggestion
"error: super() can only be called in the top-level of a class constructor
```

libs/wingc/src/type_check.rs
```diff
@@ -4667,34 +4667,25 @@ new cloud.Function(@inflight("./handler.ts"), lifts: { bucket: ["put"] });
...
-							self.spanned_error(
-								super_call,
-								format!(
-									"super() call must be the first statement of {}'s constructor",
-									ast_class.name
-								),
-							);
+						// Check if one of the statements before the super() call is invalid
+						for i in 0..idx {
+							if self.type_check_invalid_stmt_before_super(&ctor_body.statements[i]) {
```
yoav-steinberg, line 4678: suggestion: for me these types of functions make more sense when they are positive rather than negative:
```suggestion
							if !self.type_check_valid_stmt_before_super(&ctor_body.statements[i]) {
```

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2971,6 +2971,63 @@ exports[`invalid compile directory 1`] = `
...
+
+
+error: Call to super constructor can only be done from within class constructor
+   --> ../../../examples/tests/invalid/invalid_call_before_super.w:30:9
+   |
+30 |         super(2);
+   |         ^^^^^^^^^
+
+
+error: This command cannot be executed before calling super()
```
Chriscbr, line 2989:
```suggestion
error: "super()" must be called before accessing "this" in the constructor of a derived class
```

examples/tests/valid/statements_before_super.w
```diff
@@ -0,0 +1,13 @@
...
+  a: num;
+  new(a: num) {
+    this.a = a;
+  }
+}
+
+class B extends A {
+  new() {
+    let x = 3 + 5;
+    super(x);
```
Chriscbr, line 11: Nice 👍

libs/wingc/src/type_check.rs
```diff
@@ -4778,6 +4769,38 @@ new cloud.Function(@inflight("./handler.ts"), lifts: { bucket: ["put"] });
...
+					CalleeKind::Expr(e) => {
+						matches!(e.kind, ast::ExprKind::Reference(Reference::InstanceMember { .. }))
+					}
+					_ => false,
+				};
+			}
+			_ => false,
+		}
+	}
+	fn type_check_invalid_stmt_before_super(&mut self, stmt: &Stmt) -> bool {
```
yoav-steinberg, line 4790: The logic here seems to simplistic, or to narrow and only handles certain statements. I'd expect this to be a function that creates some visitor object to recursively run over all inner expression of the statement and search for a reference to `this` or ` CalleeKind::SuperCall` in an `ExprKind::Call`.
This needs to be a recursive visitor because there might be some inner scope (like an `if` block) that references `this` or calls `super.something()`.

```rust
//psuedo code
let mut check = CheckValidBeforeSuperVisitor::new();
check.visit_stmt(stmt);
return check.stmt_valid;
```
The implementation of the `Visitor` trait needs to internally do:
* implement `visit_expr` and set `stmt_valid` to false if `matches!(expr.kind, ExprKind::Call { calee:  CalleeKind::SuperCall(_), .. }`
* implement `visit_reference` and set `stmt_valid` to false if the reference matches `Reference::Identifier(Symbol {name: "this", ..})`
* Make sure not to dive into class definitions, so inner references to `this` or super method calls remain valid. This can be done by impelmenting `visit_class` in our visitor and just retuning without looking into the inner parts of the class.

Let me know if this makes sense or if something isn't clear...

yoav-steinberg (changes requested):

marciocadev:
I changed a lot to use the `Visitor` pattern, and I think it's better now.

libs/wingc/src/type_check.rs
```diff
@@ -4667,36 +4669,41 @@ new cloud.Function(@inflight("./handler.ts"), lifts: { bucket: ["put"] });
...
-							.expect("at least the super call stmt");
-						if idx != expected_idx {
-							self.spanned_error(
-								super_call,
-								format!(
-									"super() call must be the first statement of {}'s constructor",
-									ast_class.name
-								),
-							);
+						// Check if one of the statements before the super() call is invalid
```
Chriscbr, line 4678:
```suggestion
						// Check if any of the statements before the super() call are invalid
```

libs/wingc/src/type_check.rs
```diff
@@ -4667,36 +4669,41 @@ new cloud.Function(@inflight("./handler.ts"), lifts: { bucket: ["put"] });
...
+					super_ctor_loc_check.visit_scope(ctor_body);
+					// Redundant super() calls diagnostics
+					for span in super_ctor_loc_check.redundant_calls.iter() {
+						self.spanned_error_with_annotations(
+							span,
+							"super() should be called only once",
+							vec![DiagnosticAnnotation::new(
+								"First super call occurs here",
+								super_ctor_loc_check.first_call.as_ref().unwrap(),
+							)],
```
Chriscbr, line 4700: Nice span tracking

Chriscbr (approved):
Nice work @marciocadev and @yoav-steinberg!

yoav-steinberg (approved):

Issue #1306, opened 2023-01-29
cloud.Service - AWS implementation


staycoolcall911:
https://www.winglang.io/contributing/rfcs/2023-01-20-wingsdk-spec#service

hasanaburayyan:
@Chriscbr @eladb Im curious, the spec for this calls for using AppRunner over ECS in AWS due to complexity of VPC maintenance in SDK. Since then we have implemented Redis which introduced framework for deploying vpc and launching components within a VPC. Would Apprunner still be the preference here? 

I personally dont have experience with AppRunner, so no strong arguments against it, but ECS w/ fargate seems to be pretty standard place.

eladb:
ECS/Fargate is my preference too

Chriscbr:
yeah, App runner might have just been something new / shiny I saw that thought might fit the bill. If implementing this with ECS seems like a easier starting point then I would go for that 👍

(In the future we can offer multiple implementations of different resources, like a Counter implemented with DynamoDB and a Counter implemented with Redis, so we can also do the same for Service)

garysassano:
I think you could probably ease the burden of managing VPC with the new [VPC Lattice](https://aws.amazon.com/vpc/lattice/) service.

garysassano:
I think you could probably ease the burden of managing VPC with [VPC Lattice](https://aws.amazon.com/vpc/lattice/).

garysassano:
AWS App Runner is actually just an abstraction built on top of ECS Fargate to make it easier to use for developers.

Slide taken from re:Invent 2022:

![aws-compute-evolution](https://github.com/winglang/wing/assets/10464497/fb91f4f7-ceb6-404c-b882-2008558b59f0)

monadabot:
Congrats! :rocket: This was released in [Wing 0.77.3](https://github.com/winglang/wing/releases/tag/v0.77.3).

PR #6705, opened 2024-06-14
feat(sdk): support `cloud.Service` for `tf-aws` using Amazon ECS

hasanaburayyan:
Example of what works so far:
https://www.loom.com/share/5230c4c7608f49c791a819f2d322cbf4

Closes: https://github.com/winglang/wing/issues/1306

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/shared-aws/service.inflight.ts
```diff
@@ -0,0 +1,49 @@
...
+    public readonly clusterName: string,
+    public readonly serviceName: string,
+    public readonly client = new ECSClient({})
+  ) {}
+  
+  public async start(): Promise<void> {
+    const command = new UpdateServiceCommand({
+      cluster: this.clusterName,
+      service: this.serviceName,
+      desiredCount: 1,
```
eladb, line 16: I think we need to extend `cloud.Service` to include some scaling parameters.

libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -344,4 +358,85 @@ export class App extends CdktfApp {
...
+  /**
+   * The ECR Repository for the App
+   */
+  public get ecr(): EcrRepository {
+    if (this._ecr) {
+      return this._ecr;
+    }
+
+    const ecr = new EcrRepository(this, "Ecr", {
+      name: "my-ecr-repo", // TODO: make this configurable
```
eladb, line 371: Let's base the default on the root ID of the app (`{root.id}-images`).

And do this: https://github.com/winglang/wing/issues/6709

libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -344,4 +358,85 @@ export class App extends CdktfApp {
...
+
+  /**
+   * The ECR Authorization Token for the App
+   */
+  public get ecrAuth(): DataAwsEcrAuthorizationToken {
+    if (this._ecr_auth) {
+      return this._ecr_auth;
+    }
+
+    if (!this._ecr) {
```
eladb, line 386:
```suggestion
    // make sure we also have an ECR repository in this app
    if (!this._ecr) {
```

libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -344,4 +358,85 @@ export class App extends CdktfApp {
...
+  /**
+   * The ECS Cluster for the App
+   */
+  public get ecsCluster(): EcsCluster {
+    if (this._ecsCluster) {
+      return this._ecsCluster;
+    }
+
+    this._ecsCluster = new EcsCluster(this, "EcsCluster", {
+      name: "my-ecs-cluster", // TODO: make this configurable
```
eladb, line 432: Use `root.id` as the base for the default.

libs/wingsdk/src/target-tf-aws/app.ts
```diff
@@ -344,4 +358,85 @@ export class App extends CdktfApp {
...
+      ]
+    });
+
+    return this._dockerProvider;
+  }
+
+  /**
+   * The ECS Cluster for the App
+   */
+  public get ecsCluster(): EcsCluster {
```
eladb, line 426: Kind of feels like this shouldn't be part of the `App`. I foresee different providers for `cloud.Service` (e.g. EKS) and then it would make sense to manage the clusters at the provider level.

So perhaps this singleton should be decoupled from the app:

```js
let cluster = tfaws.EcsCluster.getOrCreate(this);
cluster.haveFun();
```

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+import { IamRole } from "../.gen/providers/aws/iam-role";
+import { SecurityGroup } from "../.gen/providers/aws/security-group";
+import { Image } from "../.gen/providers/docker/image";
+import { CloudwatchLogGroup } from "../.gen/providers/aws/cloudwatch-log-group";
+import { IInflightHost } from "../std";
+import { Lazy } from "cdktf";
+import { LiftMap } from "../core";
+
+
+export class Service extends cloud.Service implements IAwsInflightHost {
```
eladb, line 21: Add some commentary on how this is implemented

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+  private service: EcsService;
+
+  constructor(
+    scope: Construct,
+    id: string,
+    handler: cloud.IServiceHandler,
+    props: cloud.ServiceProps = {}
+  ) {
+    super(scope, id, handler, props);
+    // We probably need to create the ECR per app (App.of(this).ecr)
```
eladb, line 35: Seems like this is what you are doing no?

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+
+    this.workdir = App.of(this).workdir;
+    this.wrapperEntrypoint = join(this.workdir, `${this.assetName}_wrapper.js`);
+    this.dockerFileName = `Dockerfile_${this.assetName}`;
+    let app = App.of(this) as App;
+
+    // This forces the lazy loading of the properties (TODO: find a better way to do this)
```
eladb, line 42: Wouldn't just referencing these below do the trick?

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+    app.ecrAuth;
+    app.dockerProvider;
+    app.vpc;
+    
+    let image = new Image(this, "DockerImage", {
```
eladb, line 47:
```suggestion
    const image = new Image(this, "DockerImage", {
```

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+              "Resource": "*"
+            },
+            {
+              "Effect": "Allow",
+              "Action": [
+                "ecr:BatchGetImage",
+                "ecr:GetDownloadUrlForLayer",
+                "ecr:GetAuthorizationToken"
+              ],
+              "Resource": "*"
```
eladb, line 94: I think you should be able to be more specific with this since you are creating the image repository, no?

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+                Statement: this.policyStatments
+              });
+            }
+
+            return JSON.stringify({
+              Version: "2012-10-17",
+              Statement: [
+                {
+                  Effect: "Allow",
+                  Action: "none:null",
```
eladb, line 132: Do we really need this weird null statement? Can't we just create a policy without statements?

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+          },
+        ],
+      }),
+      inlinePolicy: [{
+        name: `${this.assetName}-inline-policy`,
+        policy: Lazy.stringValue({
+          produce: () => {
+            this.policyStatments = this.policyStatments ?? [];
+            
+            if (this.policyStatments.length !== 0) {
```
eladb, line 120: Coding style - bail out early for the non-happy path:

```js
if (length === 0) {
  return empty policy;
}

// happy path here!
return non-empty policy
```

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+                  name: key,
+                  value: value,
+                };
+              }
+              ),
+              logConfiguration: {
+                logDriver: "awslogs",
+                options: {
+                  "awslogs-group": `/ecs/${this.assetName}`,
+                  "awslogs-region": "us-east-1", // TODO: Can I ignore this?
```
eladb, line 167: Alternatively you should be able to use `app.region` (if we don't have a property like this we should definitely create one and use a TF data resource to obtain the current region).

Here's how you can do this:

https://github.com/winglang/winglibs/blob/main/containers/aws.w

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+          }
+        ])
+      }) as any,
+      networkMode: "awsvpc",
+      requiresCompatibilities: ["FARGATE"],
+      memory: "512",
+      cpu: "256",
+    });
+
+    let subnetIds = app.subnets["private"].map((subnet) => subnet.id);
```
eladb, line 180: Would it make sense to add `app.privateSubnets`? The `"private"` magic string is risky.

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+      vpcId: app.vpc.id,
+      egress: [
+        {
+          fromPort: 0,
+          toPort: 0,
+          protocol: "-1",
+          cidrBlocks: ["0.0.0.0/0"],
+        }
+      ],
+      ingress: [
```
eladb, line 192: We should extend `cloud.Service` to allow specifying the incoming ports.

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+          protocol: "-1",
+          cidrBlocks: ["0.0.0.0/0"]
+        }
+      ]
+    });
+
+    this.service = new EcsService(this, "Service", {
+      name: this.assetName,
+      cluster: app.ecsCluster.arn,
+      desiredCount: 1,
```
eladb, line 205: This should be configurable through `cloud.Service`.

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+      this.policyStatments.push({
+        Action: policy.actions,
+        Effect: policy.effect ?? "Allow",
+        Resource: policy.resources
+      })
+    }
+  }
+
+  addNetwork(config: NetworkConfig): void {
+    // TODO: handle networking
```
eladb, line 234: definitely required.

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+    console.log("Adding network", config);
+  }
+
+  _liftMap?: LiftMap | undefined;
+
+  /** @internal */
+  public _preSynthesize(): void {
+    super._preSynthesize();
+    let wrapper = `
+const service = require("${resolve(this.entrypoint)}");
```
eladb, line 244: Eventually this should be extracted into some common helper so it can be used across different container providers, but that's fine for now.

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    let clusterArn = (App.of(this) as App).ecsCluster.arn;
+    if (
+      ops.includes(cloud.ServiceInflightMethods.START) ||
+      ops.includes(cloud.ServiceInflightMethods.STOP)
+    ) {
+      host.addPolicyStatements({
+        actions: ["ecs:UpdateService"],
+        resources: [`${clusterArn}`], // TODO: add the service specific ARN (I think I can just do /this.service.name but need to test)
```
eladb, line 297: TODO

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+      });
+    }
+
+    if (ops.includes(cloud.ServiceInflightMethods.STARTED)) {
+      host.addPolicyStatements({
+        actions: ["ecs:DescribeServices"],
+        resources: [`${clusterArn}`], // TODO: add the service specific ARN (I think I can just do /this.service.name but need to test)
```
eladb, line 304: TODO

libs/wingsdk/src/target-tf-aws/service.ts
```diff
@@ -0,0 +1,338 @@
...
+      });
+    }
+
+
+    host.addEnvironment(this.envName(), this.service.name);
+    host.addEnvironment("ECS_CLUSTER_NAME", (App.of(this) as App).ecsCluster.name); // TODO: Should I add hash to name?
```
eladb, line 310: See my note about naming the singletons above

eladb (approved):
This is very exciting!

libs/wingsdk/src/target-tf-aws/service.ts
hasanaburayyan: Yea it probably does make sense, we have this magic string in other places as well so Ill update it everywhere.

libs/wingsdk/src/target-tf-aws/app.ts
hasanaburayyan: Yupp, 😁 Im working my way through cleaning up PRs and removing my hard coded "make it work" code

libs/wingsdk/src/target-tf-aws/service.ts
hasanaburayyan: Ah my appologies the comment is actually only valid for app.dockerProvider. Since the code below does not reference the docker provider directly, but its required to create the docker image and push to ECR.

libs/wingsdk/src/shared-aws/service.inflight.ts
hasanaburayyan: @eladb I went back and forth on adding it, do you think its a functional concern to the developer?

libs/wingsdk/src/target-tf-aws/service.ts
hasanaburayyan: Im not sure we do this same thing for `function.ts` with a comment that says: 
```
// policy must contain at least one statement, so include a no-op statement
```

So I assume its needed, but I can test

libs/wingsdk/src/target-tf-aws/service.ts
hasanaburayyan: Yea forgor to remove comment after I did it 😁

libs/wingsdk/src/target-tf-aws/service.ts
hasanaburayyan: Im working on a followup PR to adjust the way `cloud.Service` works theres a limitation in the current API, Ill plan to make these configurations there.

Basically when a return function is expected, we have an issue where a service that is long running I.E.

```js
let s = new cloud.Service(inflight () => {
  while true {
    let index = counter.inc();
    bucket.put("file-{index}.txt", "Hello, world!");
    util.sleep(10s);
  }

	// This will never get returned
  return inflight () => {
    for file in bucket.list() {
      bucket.delete(file);
    }
  };
});
```
The program loop never finishes, so calling stop on the service has no function to run when the service shuts down.

libs/wingsdk/src/target-tf-aws/service.ts
hasanaburayyan: Yea I have been thinking about that, but its kinda useless (in wing) unless `cloud.Service` has a `url` property. See comment bellow about followup PR, I think service needs some other tweaks as well.

PR #6721, opened 2024-06-17
fix(compiler)!: this.node and nodeof(this) have different behaviors

### LINKED ISSUE #5840: this.node and nodeof(this) give different behaviors
### I tried this:

```js
class A {
  new() {
    this.node.hidden; // type checking error
    nodeof(this).hidden; // OK
    
    log("root:   " + this.node.root.node.path);
    log("nodeof: " + nodeof(this).root.node.path);
  }
}

new A();

test "a" {
  assert(2 == 2);
}

test "b" {
  assert(2 != 3);
}

```

### This happened:

Two visible differences in behavior

1. We see `this.node` does not have a hidden field, but the node returned by `nodeof()` does
2. If we comment out the previous error and run the tests, we see the following is logged:

root:   root
nodeof: root/env0
root:   root
nodeof: root/env1

This shows the "root" values are different also.

### I expected this:

No difference / or only one of the syntaxes to be allowed

### Is there a workaround?

We recommend only using `nodeof(this)`

### Anything else?

`nodeof(this)` returns a special Node instance, not the `Node` instance that normally exists on classes from the [constructs](https://github.com/aws/constructs) framework. Some of the differences in behavior are because

Perhaps we can hide the node field from classes extending `Construct` somehow

### Wing Version

0.59.12

### Node.js Version

20.9.0

### Platform(s)

MacOS

### Community Notes

Chriscbr:
To avoid confusion between different APIs and different runtime behavior between the `std.Node` and `constructs.Node` types, we avoid exposing `constructs.Node` in the language where possible.

Fixes #5840

BREAKING CHANGE: `.node` is no longer a directly available field on preflight classes and `IConstruct` values. Use the global function `nodeof()` for accessing tree nodes instead.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/type_check.rs
+++ b/libs/wingc/src/type_check.rs
@@ -969,7 +969,7 @@ impl TypeRef {
 		}
 	}
 
-	pub fn as_mut_struct(&mut self) -> Option<&mut Struct> {
+	pub fn as_struct_mut(&mut self) -> Option<&mut Struct> {
 		if let Type::Struct(ref mut st) = **self {
 			Some(st)
 		} else {
@@ -984,7 +984,7 @@ impl TypeRef {
 			None
 		}
 	}
-	fn as_mut_interface(&mut self) -> Option<&mut Interface> {
+	fn as_interface_mut(&mut self) -> Option<&mut Interface> {
 		if let Type::Interface(ref mut iface) = **self {
 			Some(iface)
 		} else {
@@ -1008,7 +1008,7 @@ impl TypeRef {
 		}
 	}
 
-	pub fn as_mut_function_sig(&mut self) -> Option<&mut FunctionSignature> {
+	pub fn as_function_sig_mut(&mut self) -> Option<&mut FunctionSignature> {
 		if let Type::Function(ref mut sig) = **self {
 			Some(sig)
 		} else {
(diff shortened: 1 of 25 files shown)


libs/wingc/src/type_check.rs
```diff
@@ -1970,6 +1971,40 @@ impl<'a> TypeChecker<'a> {
...
+			.libraries
+			.lookup_nested_str(&CONSTRUCT_BASE_INTERFACE, None)
+			.expect("constructs.IConstruct not found in type system")
+			.0
+			.as_type()
+			.expect("constructs.IConstruct was found but it's not a type");
+		let iface = constructs_iface
+			.as_interface_mut()
+			.expect("constructs.IConstruct was found but it's not a class");
+		iface.env.symbol_map.remove("node");
```
eladb, line 1992: Maybe out `"node"` in a const

libs/wingc/src/type_check.rs
```diff
@@ -1970,6 +1971,40 @@ impl<'a> TypeChecker<'a> {
...
+			.libraries
+			.lookup_nested_str(&CONSTRUCT_BASE_CLASS, None)
+			.expect("constructs.Construct not found in type system")
+			.0
+			.as_type()
+			.expect("constructs.Construct was found but it's not a type");
+		let class = constructs_class
+			.as_class_mut()
+			.expect("constructs.Construct was found but it's not a class");
+		class.env.symbol_map.remove("node");
```
eladb, line 2005: Ditto

libs/wingc/src/type_check.rs
```diff
@@ -6805,14 +6845,19 @@ where
 	match lookup_result {
 		LookupResult::NotFound(s, maybe_t) => {
 			let message = if let Some(env_type) = maybe_t {
-				format!("Member \"{s}\" doesn't exist in \"{env_type}\"")
+				format!("Member \"{s}\" does not exist in \"{env_type}\"")
 			} else {
 				format!("Unknown symbol \"{s}\"")
 			};
+			let mut hints = vec![];
+			if s.name == "node" {
```
eladb, line 6853: Ditto

eladb (approved):
Yey!

libs/wingc/src/type_check.rs
Chriscbr: Updated 👍

PR #6729, opened 2024-06-17
feat(sdk): bucket multipart upload  

eladcon:
## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/cloud/bucket.ts
+++ b/libs/wingsdk/src/cloud/bucket.ts
@@ -477,6 +477,13 @@ export interface IBucketClient {
    * @inflight
    */
   multipartUpload(key: string): Promise<string>;
+
+  /**
+   * Complete a multipart upload to a given key in the bucket.
+   * @param uploadId The upload id for the multipart upload
+   * @inflight
+   */
+  completeMultipartUpload(uploadId: string): Promise<void>;
 }
 
 /**
@@ -592,4 +599,6 @@ export enum BucketInflightMethods {
   RENAME = "rename",
   /** `Bucket.multipartUpload` */
   MULTIPART_UPLOAD = "multipartUpload",
+  /** `Bucket.completeMultipartUpload */
+  COMPLETE_MULTIPART_UPLOAD = "completeMultipartUpload",
 }
--- a/libs/wingsdk/src/cloud/bucket.ts
+++ b/libs/wingsdk/src/cloud/bucket.ts
@@ -293,7 +293,7 @@ export interface BucketGetOptions {
 /**
  * Options for `Bucket.tryGet()`.
  */
-export interface BucketTryGetOptions extends BucketGetOptions {}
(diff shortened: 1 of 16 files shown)


libs/wingsdk/src/cloud/bucket.ts
```diff
@@ -457,6 +469,21 @@ export interface IBucketClient {
...
    */
   rename(srcKey: string, dstKey: string): Promise<void>;
+
+  /**
+   * Initiate a multipart upload to a given key in the bucket.
+   * @param key The key to upload to
+   * @returns The upload id for the multipart upload
+   * @inflight
+   */
+  multipartUpload(key: string): Promise<string>;
```
eladb, line 479: Is the term "multipart" really helping here? I think the common developer won't really care that this is the underlying mechanism.

The docstrings should indicate that this will begin a multipart upload, but not needed in the name I think.

The return type can also be used to encode more meaning after you've landed on the right method. 

I would also return a struct (with the upload id inside) instead of a string so it's easier to correlate the start/stop.

So maybe something like this:

```js
struct MultipartUpload {
  id: str;
}

startUpload(key: str): MultipartUpload;

completeUpload(upload: MultipartUpload): void
```

eladb, line 479: Where do I specify the content?

libs/wingsdk/src/shared-aws/bucket.inflight.ts
```diff
@@ -410,6 +413,21 @@ export class BucketClient implements IAwsBucketClient {
         });
         break;
       case BucketSignedUrlAction.UPLOAD:
+        if (opts?.uploadId !== undefined) {
```
eladb, line 416: How is this related to signed urls?

libs/wingsdk/src/shared-aws/permissions.ts
```diff
@@ -160,6 +162,12 @@ export function calculateBucketPermissions(
   ) {
     actions.push("s3:CopyObject");
   }
+
+  // multipart upload signed url
+  if (ops.includes(cloud.BucketInflightMethods.SIGNED_URL)) {
```
eladb, line 167: ??

eladb (changes requested):

eladb:
Removed the word "support" from the title

libs/wingsdk/src/cloud/bucket.ts
yoav-steinberg: this just initiates an upload, you the either use a signed url and an http put or the `uploadPart` method to specify the content.

libs/wingsdk/src/shared-aws/bucket.inflight.ts
yoav-steinberg: You may generate a signed url specifically to upload a part of a multipart upload. When there's a multipart upload object attached to the sign request options then we know you want a signed url for a multipart upload.

Issue #6756, opened 2024-06-21
Intermittent CI failures on windows (Cannot find module)


MarkMcCulloh:
### I tried this:

Run `sdk_tests/resource/call.test.w` on Windows in CI

(Example https://github.com/winglang/wing/actions/runs/9618038572/job/26531168752)

### This happened:

```
Error: R] resource can log messages at different levels | Error: an error log
Error: R] resource.call times out if the method takes too long | Error: Resource is not running (it may have crashed or stopped)
Error: R] resource.call times out if the method takes too long | Error: Process exited with code null, signal SIGTERM
Error: R] resource.call times out if the method takes too long | Error: node:internal/modules/cjs/loader:1148
Error: R] resource.call times out if the method takes too long | Error:   throw err;
Error: R] resource.call times out if the method takes too long | Error:   ^
Error: R] resource.call times out if the method takes too long | Error: 
Error: R] resource.call times out if the method takes too long | Error: Cannot find module 'D:/a/wing/wing/examples/tests/sdk_tests/resource/target/test/call.test.wsim/.wing/resource_c823d293.sandbox.cjs.bundle/index.cjs'
Error: R] resource.call times out if the method takes too long | Error: 
Error: R] resource.call times out if the method takes too long | Error: 
Error: R] resource.call times out if the method takes too long | Error: 
Error: R] resource.call times out if the method takes too long | Error: 
Error: R] resource.call times out if the method takes too long | Error:   code: 'MODULE_NOT_FOUND',
Error: R] resource.call times out if the method takes too long | Error:   requireStack: []
Error: R] resource.call times out if the method takes too long | Error: }
Error: R] resource.call times out if the method takes too long | Error: 
Error: R] resource.call times out if the method takes too long | Error: Node.js v20.14.0
```

### Anything else?

Seems to have started with the update to vitest v2 for hangar

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

PR #6757, opened 2024-06-21
chore: disable flakey test on Windows

Chriscbr:
The test named "resource.call times out if the method takes too long" has been flaking since we merged in https://github.com/winglang/wing/pull/6686. Specifically, it sometimes logs additional errors:

```
Resource is not running (it may have crashed or stopped)
Process exited with code null, signal SIGTERM
node:internal/modules/cjs/loader:1148
  throw err;
  ^

Cannot find module 'D:/a/wing/wing/examples/tests/sdk_tests/resource/target/test/call.test.wsim/.wing/resource_c823d293.sandbox.cjs.bundle/index.cjs'

  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v20.14.0
```

I have a suspicion the test is already broken in some way because even when I try running the test on my local system, the test logs "Resource is not running (it may have crashed or stopped)" instead of an error saying that the method call timed out: https://github.com/winglang/wing/blob/56010c4f7c5934e7fabdefb8fd91e8896f624f29/libs/wingsdk/src/target-sim/resource.inflight.ts#L179

This PR disables the test until we're able to solve the underlying issue.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/examples/tests/sdk_tests/resource/call.test.w
+++ b/examples/tests/sdk_tests/resource/call.test.w
@@ -218,13 +218,16 @@ if util.env("WING_TARGET") == "sim" {
     r1.printLogs();
   }
 
-  test "resource.call times out if the method takes too long" {
-    let var msg = "";
-    try {
-      r1.methodTakesLongTime();
-    } catch err {
-      msg = err;
+  // This test is disabled on Windows because it is known to be flakey on that platform
+  if util.os() != "win32" {
+    test "resource.call times out if the method takes too long" {
+      let var msg = "";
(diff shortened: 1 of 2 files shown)


MarkMcCulloh (approved):
https://github.com/winglang/wing/issues/6756

Chriscbr:
Ah, if I only disable it on Windows then the snapshots assertion will fail. I'll disable it on all platforms for now

hasanaburayyan (approved):

PR #6760, opened 2024-06-24
chore: update minor/patch across repo

MarkMcCulloh:
Resolves all audit/dependabot notices (except 1, which will be ignored anyways)

Updating typescript caused an interesting error in the SDK (see changes to AssertionError) but otherwise the update was uneventful. As part of this update I made changes to hangar to ignore cdktf version number changes in the future since it's not an interesting change.


eladb (approved):
Impressive.

We should get automatic updates in place. It will be much less disruptive to continuously update deps than one off every once in a while.

skyrpex (approved):

PR #6803, opened 2024-06-28
chore: add docs about building liftable abstractions

Chriscbr:
Adding some docs about parts of the Wing lifting system that the docs doesn't cover yet but is used in a lot of winglibs today.

## Checklist

- [ ] Tests added (always)


docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
 Statements within a `lift` block are exempt from the compiler's analyzer that tries to determine preflight object usage automatically.
 If an inflight method is directly or indirectly called within a `lift` block without sufficient resource qualifications, it may result in errors at runtime.
 
+## Building liftable abstractions
+
+Compute environments where inflight code is executed are called **inflight hosts**.
```
boyney123, line 339: Should we have a link to inflight hosts somewhere? Or I think as you said move the title to reflect inflight hosts. First time the term is used, maybe the section should lead with this in the title.

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+For example, AWS Lambda, Fly.io machines, and Fargate containers are all examples of compute environments that can run inflight code.
```
boyney123, line 340: I think it's helpful to link out to these concepts, link to AWS Lambda, Fly.io, and Fargate docs somewhere. Allows people to deep deeper if they want to understand more.

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+
+In order to run a piece of inflight code, the inflight host may need to be set up in a certain way.
+For example, to run an inflight function that reads from a storage bucket, the inflight host may require permissions to read from the bucket, or may require certain environment variables to be set, or may require network policies must be configured.
+
+Preflight classes can be used to encapsulate these requirements.
+Any preflight class can implement a method named `onLift` that is called when the class is used in inflight code, where requirements can be added to the inflight host.
```
boyney123, line 346: Do we have any more info on onLift where we can link to?

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+        throw "Unsupported inflight host type";
+      }
+    } else {
+      // no requirements for other operations
+    }
+  }
+}
+```
+
+Here, `Model` has one public inflight method.
```
boyney123, line 388: Maybe I'm misunderstanding does it have one or two? `invoke` and `printModelId`?

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+Inside the `onLift` method, the class first checks if the "invoke" method was one of the operations requested by the inflight host.
+It then check if inflight host that's trying to use `Model` is an AWS Lambda function, and if so, it adds a policy statement to the Lambda function that allows it to invoke the model.
```
boyney123, line 390:
```suggestion
It then checks if the inflight host is an AWS Lambda function, and if so, it adds a [policy statement](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html) to the Lambda function that allows it to invoke the model.
```

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+
+Under the hood, the compiler will call `onLift` on the `Model` class once for each inflight host that uses it.
+`onLift` should not be called directly.
```
boyney123, line 393: `onLift should not be called directly.`, should we have a sentence to explain why?

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,82 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
 Statements within a `lift` block are exempt from the compiler's analyzer that tries to determine preflight object usage automatically.
 If an inflight method is directly or indirectly called within a `lift` block without sufficient resource qualifications, it may result in errors at runtime.
 
+## Building liftable abstractions
```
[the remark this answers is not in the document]
boyney123: Makes sense I think, but I feel there is a section missing to define inflight hosts? (unless I can't see it somewhere)

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+
+Compute environments where inflight code is executed are called **inflight hosts**.
+For example, AWS Lambda, Fly.io machines, and Fargate containers are all examples of compute environments that can run inflight code.
+
+In order to run a piece of inflight code, the inflight host may need to be set up in a certain way.
```
boyney123, line 342:
```suggestion
In order to run a piece of inflight code, the inflight host may need to be configured.
```

docs/docs/02-concepts/01-preflight-and-inflight.md
```diff
@@ -334,6 +334,83 @@ Within the first clause of the `lift` block, a list of qualifications on preflig
...
+For example, to run an inflight function that reads from a storage bucket, the inflight host may require permissions to read from the bucket, or may require certain environment variables to be set, or may require network policies must be configured.
```
boyney123, line 343: Think it's worth linking to https://www.winglang.io/docs/standard-library/cloud/bucket for `storage bucket`?

I'm a fan of helping people out when these terms may be new to them.

docs/docs/02-concepts/01-preflight-and-inflight.md
Chriscbr: I updated the title -- I think this is the first time we introduce the concept or make it explicit. It could be worth giving the article a broader refresh in the future and see if the concept should be introduced earlier, not sure

docs/docs/02-concepts/01-preflight-and-inflight.md
Chriscbr: Oops, you're right. Fixed

docs/docs/02-concepts/01-preflight-and-inflight.md
Chriscbr: Maybe? It think reason mainly comes down to the intent of the API as a hook that's meant to be used by the framework - similar to how React lets you define `componentDidMount()` methods on classes, but they aren't meant to be called directly. It's an example of the [hollywood principle](https://deviq.com/principles/hollywood-principle). If you think there's a sentence or clause I could add to clarify this, I'm open to ideas.

There might be technical reasons to avoid calling it directly, like that it could lead to a host getting 
"configured" multiple times, but these reasons aren't set in stone and also might confuse people more.

docs/docs/02-concepts/01-preflight-and-inflight.md
Chriscbr: Fixed

docs/docs/02-concepts/01-preflight-and-inflight.md
Chriscbr: Sure thing, done

docs/docs/02-concepts/01-preflight-and-inflight.md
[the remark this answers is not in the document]
Chriscbr: I updated the section title

docs/docs/02-concepts/01-preflight-and-inflight.md
Chriscbr: I think so -- I just pushed a change so I think we should be generating some API reference docs where we specify the `onLift` method in the SDK. So there's another place where folks can refer to the method signature / find it in the docs.

I should mention there might be an issue if we try to do the same for `onLiftType`, because the method needs to be implemented right now as a static method. But it's not possible to define an interface that enforces a class implements a static method. For now I'm ok looking over this gap 😅

docs/docs/02-concepts/01-preflight-and-inflight.md
boyney123: OK makes sense to me now you explained it, also TIL about the hollywood principle 😄

boyney123 (approved):

PR #6815, opened 2024-07-01
feat(sdk): add timeout property to http requests

### LINKED ISSUE #6791: Add ability to specify a timeout for `http` methods
### Use Case

The ability to specify a timeout on http methods similar to axios, e.g.

`await axios.get('https://example.com/api', {
      timeout: 5000 // Timeout in milliseconds (5 seconds)
 });`

### Proposed Solution

As an additional method parameter or an extra setting in http.RequestOptions so that it will be available on all relevant methods including `http.fetch()`, `http.get()`, `http.post()`, etc.

### Implementation Notes

_No response_

### Component

Libraries

### Community Notes

tsuf239:
closes #6791

tf-aws:
image

sim:
image

tf-azure (unsupported):
image

tf-gcp (unsupported):
image

## Checklist

- [ ] Tests added (always)


libs/wingsdk/src/http/http.ts
```diff
@@ -134,6 +135,11 @@ export interface RequestOptions {
    * @default about:client
    */
   readonly referrer?: string;
+  /**
+   * Timeout for terminating a pending request. None if undefined.
+   * @default undefined
```
Chriscbr, line 140: Can you check if Node's `fetch` has a default timeout? I think it might be 30s - but if there isn't any then let's add "no timeout"
```suggestion
   * @default - 30s
```
```suggestion
   * @default - no timeout
```

examples/tests/sdk_tests/http/fetch.test.w
```diff
@@ -24,15 +25,42 @@ api.get("/target", inflight () => {
...
+});
+
+test "fetch" {
+  // "redirect is 'follow' by default"
+  let redirectDefault = http.get("{api.url}/redirect");
+  log(redirectDefault.body);
+  expect.equal(redirectDefault.status, 200);
+  expect.equal(redirectDefault.body, "I am the target");
+
+ // "redirect 'manual'"
```
Chriscbr, line 43: nit: indent
```suggestion
  // "redirect 'manual'" 
```

libs/wingsdk/src/http/http.ts
Chriscbr, line 1: (optional) I wonder if we can add a `http.md` file here and include an example, like we do for the cloud resources with `cloud/bucket.md`?

Chriscbr (approved):
LGTM 🙌

libs/wingsdk/src/http/http.ts
tsuf239: there isn't any- otherwise I could just add the property :)

libs/wingsdk/src/http/http.ts
tsuf239: https://www.winglang.io/docs/standard-library/http/api-reference looks pretty comprehensive to me, as a developer I will be able to use the HTTP module after looking at those (especially that the links are clickable and working)
image
image

I have no idea how to encourage people to search for it if needed, or if it's "findable" enough, would you search for "HTTP" if you were new to wing and wanted to send an HTTP request? 
(@boyney123 WDYT?)

boyney123: Yeah I think it's fine here tbh, although when we add examples/patterns page this is a perfect example for that, will make a note of this in a collection of patterns we can add.

PR #6830, opened 2024-07-02
fix(sdk): customized rootId for sim

### LINKED ISSUE #6598: --rootId option does nothing in `sim`
### I tried this:

```wing
log(nodeof(this).app.node.id);
```

Then run `wing compile --rootId hello`

### This happened:

Outputs "root"

### I expected this:

Outputs "hello"

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

tsuf239:
fixes #6598
checked the console too and it doesn't seem to break, the `rootid.test.w` file and its tests work.
## Checklist

- [ ] Tests added (always)


apps/wing/src/commands/compile.ts
```diff
@@ -30,7 +30,7 @@ export interface CompileOptions {
    *
    * @default "Default"
    */
-  readonly rootId?: string;
```
eladb, line 33: Why?

apps/wing/src/commands/test/test.ts
```diff
@@ -195,12 +195,12 @@ async function testOne(
 
   let results: std.TestResult[] = [];
   if (shouldExecute) {
+    options.rootId = options.rootId ?? `Test.${nanoid(10)}`;
```
eladb, line 198: Not sure I understand why this change is needed?

It's better to put the id in a local variable and reuse than to mutate the options object

apps/wing/src/commands/test/test.ts
```diff
@@ -195,12 +195,12 @@ async function testOne(
...
     const synthDir = await withSpinner(
       `Compiling ${renderTestName(entrypoint)} to ${target}...`,
       async () =>
         compile(entrypoint, {
           ...options,
-          rootId: options.rootId ?? `Test.${nanoid(10)}`,
```
eladb, line 203: This will cause the id to be non deterministic

eladb (changes requested):

apps/wing/src/commands/test/test.ts
```diff
@@ -192,20 +192,20 @@ async function testOne(
   // determine snapshot behavior
   const snapshotMode = determineSnapshotMode(target, options);
   const shouldExecute = snapshotMode === SnapshotMode.NEVER || snapshotMode === SnapshotMode.DEPLOY;
+  const testOptions = { ...options, rootId: options.rootId ?? `Test.${nanoid(10)}` };
```
Chriscbr, line 195: @eladb @tsuf239 When should random test IDs be used?

my understanding so far:
- if a random ID was used every time someone ran `wing test`, then [snapshot tests](https://www.winglang.io/contributing/rfcs/2024-03-24-snapshot-tests) would be broken since they would produce different snapshots each time
- for cloud testing (currently only in our monorepo, but maybe users will want this done for them in the future), if we don't randomize the ID when running `wing test`, then it's more likely for tests to fail due to spurious resource name conflicts

tsuf239: it's already in use in the cloud targets (I believe it updates the snapshots each time, from my experience- since the snapshots were invented since the randomization was there before), 
this PR is for making it behave the same way in the sim 
second point is very true :) we added this randomization to be able to run the tests in parallel

Chriscbr: > I believe it updates the snapshots each time
I see, I wonder if it's a latent bug?

examples/tests/valid/this.test.w
```diff
@@ -1,7 +1,12 @@
+/*\
+args:
+  - --rootId
+  - root
+\*/
```
Chriscbr, line 5: Why is this needed? is the default ID not "root" anymore? I'm not sure I understand how randomizing the ID is needed to fix https://github.com/winglang/wing/issues/6598

tsuf239: You are right about the randomization, this was reverted, but the args here are needed for the tf-aws compilation hangar test- the default one there is `Default` and not `root`

apps/wing/src/commands/test/test.ts
tsuf239: yes, as far as I know the snapshots are not even committed: https://github.com/winglang/wing/blob/b1d01275367b3eb2476528ec56d7b29d193492b4/examples/tests/sdk_tests/.gitignore#L1

Chriscbr (approved):

Chriscbr (approved):
LGTM

PR #6831, opened 2024-07-02
feat(console): react to envfile changes

### LINKED ISSUE #6349: Wing Console does not reflect changed secrets
### I tried this:

I created this app:

```js
bring cloud;

let s = new cloud.Secret(name: "secret1");

  log(s.value());
});
```

with this `.env` file in my directory:

```
secret1=foo
```

I ran `wing run` to open the Wing Console, and invoked the function, and it worked. Then I edited my Wing file and .env file to the following:

```
secret1=bar
```

I invoked the function, and it still prints "foo".

Then I edited the Wing file and .env file to the following:

```js
let s = new cloud.Secret(name: "secret2");
```

```
secret2=bar
```

I invoked the function, and it says the secret does not exist.

### This happened:

See above

### I expected this:

I expected the simulator to automatically reload based on the secret information since my app recompiled each time I edited the .w file.

### Is there a workaround?

Stop and restart Wing Console

### Anything else?

_No response_

### Wing Version

0.72.12

### Node.js Version

20.12.2

### Platform(s)

MacOS

### Community Notes

skyrpex:
Fixes #6349.


skyrpex:
Duplicates an `env.ts` file from the cli to the console. Need to refactor into a shared package.

skyrpex:
@MarkMcCulloh in order to share the `loadEnvVariables` functionality between the wing cli and the console server, the quickest solution would be moving the `loadEnvVariables` implementation to the `@wingconsole/server` package and exposing it so the wing cli just `import { loadEnvVariables } from "@wingconsole/server`. But feels odd importing such a function from the console server package, right?

Another solution is to keep the `loadEnvVariables` duplicated, but feel like we could get into inconsistencies parsing the env file at some point.

The last solution that comes to my mind is to actually create a `@winglang/dotenv` package in `libs/dotenv` and consume it from both wing cli and console server. The only gotcha is the complexity of adding new packages to the CI (must add lots of stuff to the yml files...).

MarkMcCulloh:
We could expose this functionality in wingsdk instead, there are already similar shared things like this (stacktrace rewriting, Test and TestRunner, etc.).

This feels a little off but I think the root of that is that the responsibility of "watching" should eventually be moved out of the console anyways, `wing compile` and `wing test` would benefit from a similar watch mode eventually imo.

skyrpex:
> We could expose this functionality in wingsdk instead, there are already similar shared things like this (stacktrace rewriting, Test and TestRunner, etc.).

Good call, will do that.




> This feels a little off but I think the root of that is that the responsibility of "watching" should eventually be moved out of the console anyways, `wing compile` and `wing test` would benefit from a similar watch mode imo.

I feel the same. Kind of related to the compiler emitting files that should be watched, too. Let's talk about this one day.

polamoros (approved):

PR #6832, opened 2024-07-02
chore(cli): remove `awscdk` from builtin

hasanaburayyan:
looks like oversight from when we moved `awscdk` to be external

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing/src/cli.ts
+++ b/apps/wing/src/cli.ts
@@ -169,7 +169,7 @@ async function main() {
     .argument("[entrypoint]", "program .w entrypoint")
     .option(
       "-t, --platform <platform> --platform <platform>",
-      "Target platform provider (builtin: sim, tf-aws, tf-azure, tf-gcp, awscdk)",
+      "Target platform provider (builtin: sim, tf-aws, tf-azure, tf-gcp)",
       collectPlatformVariadic,
       DEFAULT_PLATFORM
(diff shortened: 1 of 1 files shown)


Chriscbr (approved):

apps/wing/src/cli.ts
```diff
@@ -169,7 +169,7 @@ async function main() {
     .argument("[entrypoint]", "program .w entrypoint")
     .option(
       "-t, --platform <platform> --platform <platform>",
-      "Target platform provider (builtin: sim, tf-aws, tf-azure, tf-gcp, awscdk)",
```
eladb, line 172: We should figure out how we improve the discoverability of non built in platforms?  

Maybe we can add here a link to a doc page with a list as a simple first step?

@boyney123 what do you think?

boyney123: Yeah I like the idea, maybe we can add a `List of platforms` page to the new platforms section.

image

Also the new patterns/examples page I'm working on, maybe we can also add a filter by platform too, this will also help with discoverability.

I will add a page into the docs with a table/list of platforms as a first step.

boyney123: Added a new here @eladb https://github.com/winglang/wing/pull/6859, let me know what you think

Issue #6754, opened 2024-06-21
Cannot see full dates for logs in Wing Console


Chriscbr:
### I tried this:

Looking at the logs of an app that I've had running for multiple days (on the Wing Cloud hosted platform)

### This happened:

If I have an app running for multiple days, the Wing Console logs only show the hours, minutes, and seconds.

<img width="449" alt="Screenshot 2024-06-21 at 2 30 16 PM" src="https://github.com/winglang/wing/assets/5008987/5fe91c64-dd76-435b-8faf-22089d54dfb3">

### I expected this:

To show the full date (either by default, or by letting me turning on some setting)

### Wing Version

0.75.3

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.76.6](https://github.com/winglang/wing/releases/tag/v0.76.6).

PR #6833, opened 2024-07-03
feat(console): show full date for log entries

polamoros:
Resolves https://github.com/winglang/wing/issues/6754

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing-console/console/ui/src/features/logs-pane/console-logs.tsx
+++ b/apps/wing-console/console/ui/src/features/logs-pane/console-logs.tsx
@@ -262,15 +262,17 @@ export const ConsoleLogs = memo(
   }: ConsoleLogsProps) => {
     const { theme } = useTheme();
 
-    const useLongDateFormat = useMemo(() => {
-      if (!logs[0]?.timestamp) {
-        return false;
+    const [useLongDateFormat, setUseLongDateFormat] = useState(false);
+
+    useEffect(() => {
+      if (!logs[0]?.timestamp || useLongDateFormat) {
+        return;
       }
       const now = new Date();
       const logDate = new Date(logs[0].timestamp);
 
-      return logDate.getDate() !== now.getDate();
-    }, [logs]);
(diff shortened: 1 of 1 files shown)


apps/wing-console/console/ui/src/features/logs-pane/console-logs.tsx
```diff
@@ -234,6 +262,16 @@ export const ConsoleLogs = memo(
...
+    const useLongDateFormat = useMemo(() => {
+      if (!logs[0]?.timestamp) {
+        return false;
+      }
+      const now = new Date();
+      const logDate = new Date(logs[0].timestamp);
+
+      return logDate.getDate() !== now.getDate();
+    }, [logs]);
+
```
skyrpex, line 274: I think we could improve this implementation. The problem I see now is that this hook will be executed every time there's a new log entry, but as soon as we know we need to use the long date format, we don't have to check again.

I propose using a `const [useLongDateFormat, setUseLongDateFormat] = useState(false);` and an effect that depends on `useLongDateFormat, logs`. If `useLongDateFormat` is true, no need to check; otherwise, check if the date in the timestamps is different and update with `setUseLongDateFormat` as appropiate.

apps/wing-console/console/ui/src/features/logs-pane/console-logs.tsx
```diff
@@ -129,7 +157,7 @@ const LogEntryRow = memo(
                 "flex-shrink-0 select-none pointer-events-none",
               )}
             >
-              {dateTimeFormat.format(log.timestamp)}
+              {formatDate(log.timestamp)}
```
skyrpex, line 160: Everytime there's a new log entry, all previou log entries will be formatted again, right?

Why don't we add a `<LogEntry />` component and use a memo so we format the date only once?

skyrpex (changes requested):

skyrpex (approved):

Issue #6552, opened 2024-05-23
Weekly releases digest on Discord


Chriscbr:
### Use Case

On the Wing Discord server (https://t.winglang.io/discord) there's currently a #breaking-channels channel that publishes messages whenever breaking changes are published to the toolchain or to Wing libraries.

It could be handy to also get notified about other changes, but to have them as some kind of periodic, weekly digest.

### Proposed Solution

The digest could be produced through a weekly job that identifies all commits (from the Wing monorepo and winglibs repo) made in the past week. Commits that start with "chore" could be skipped or ignored since they usually reflect non-user-facing changes.

### Component

Other

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Discord](https://t.winglang.io/discord).

staycoolcall911:
Related to #2933

Chriscbr:
Implemented in https://github.com/Chriscbr/wing-releases-discord-bot/commit/f456a4ae90e0b99a430ecba11a3c96722b685f2b

staycoolcall911:
Nice!!!

Chriscbr:
I think there's some bug causing the cron function to not be executed, so it wasn't published this past Sunday

Chriscbr:
The fix has been added in Wing v0.76.10 but it hasn't applied since I think Wing Cloud is still running on an older version. @polamoros @eladcon how much effort do you estimate it would take to update Wing Cloud?

PR #6844, opened 2024-07-03
fix(sdk): cron events not triggered on simulator

Chriscbr:
Needed for https://github.com/winglang/wing/issues/6552

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/libs/wingsdk/src/target-sim/schedule.inflight.ts
+++ b/libs/wingsdk/src/target-sim/schedule.inflight.ts
@@ -24,7 +24,6 @@ export class Schedule
 
   constructor(props: ScheduleSchema) {
     this.interval = parseExpression(props.cronExpression, { utc: true });
-    this.scheduleFunction();
   }
 
   private get context(): ISimulatorContext {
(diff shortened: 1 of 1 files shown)


libs/wingsdk/src/target-sim/schedule.inflight.ts
```diff
@@ -36,19 +35,32 @@ export class Schedule
...
     this.intervalTimeout = setTimeout(() => {
       this.runTasks();
       this.scheduleFunction();
-    }, this.nextDelay(this.interval));
+    }, delay);
   }
 
   public async init(context: ISimulatorContext): Promise<ScheduleAttributes> {
     this._context = context;
+    this.scheduleFunction();
```
MarkMcCulloh, line 63: Is there a feasible test to make sure the trigger happens? Our schedule resource lacks granularity so I imagine there would be some pain

MarkMcCulloh (approved):

libs/wingsdk/src/target-sim/schedule.inflight.ts
Chriscbr: We've got some end to end tests that test for e.g. the schedule activating every minute, I think here the issue was just with date calculation / subtraction.

There's _probably_ a way to check the behavior of date-based crons (that doesn't take hours to run) by writing a test where a cron is dynamically chosen during preflight based on the current time that `wing test` is executed, but it sounds like a headache in practice

Issue #6579, opened 2024-05-28
UX for creating a tunnel in Wing Console is confusing and potentially risky


eladb:
### I tried this:

Create a tunnel in Wing Console

### This happened:

* It's very hard to discover this feature (unless you know to click the tiny icon and all that).
* We should get the user to approve that they understand the security implications of a tunnel into their local machine.
* It takes a few seconds to get the tunnel created and the app restarted and there's no clear indication what's happening (perhaps some kind of modal loading window or some other indication that the app is being restarted).

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

ekeren:
> It takes a few seconds to get the tunnel created and the app restarted and there's no clear indication what's happening (perhaps some kind of modal loading window or some other indication that the app is being restarted).

+100
And plus 1 for the other stuff

boyney123:
Just adding to this, I'm still new to the whole thing, and agree on what's already been said.

Also stopping my wing console with an active tunnel kills the tunnel (expected), when I start the console again the tunnel is back to active. But only a small indication in the UI that this is the case. 

Could this catch people out? Folks that open a tunnel, close the console, and open in the future to have an open tunnel they forgot about? Just a thought...

polamoros:
I'm adding some UI/UX improvements to the endpoints here
https://github.com/winglang/wing/issues/6796

Only missing this part:

> We should get the user to approve that they understand the security implications of a tunnel into their local machine.

skyrpex:
> We should get the user to approve that they understand the security implications of a tunnel into their local machine.

Maybe we could use our local config object to store whether the user approved the use of tunnels or not, so we only show it once.

> It takes a few seconds to get the tunnel created and the app restarted [...]

Not sure if we could also make it faster somehow? Is there room for quick improvements there, @eladcon ?

polamoros:
Regarding the warning, I would show the modal the first time a user exposes an endpoint.

<img width="1512" alt="image" src="https://github.com/winglang/wing/assets/5547636/da3ba828-448a-459c-8545-28c1419f8f1c">

Once the user accepts this modal, we won't show it again (as @skyrpex suggested).

skyrpex:
[...]

PR #6852, opened 2024-07-05
feat(console): notify of potential risks when exposing endpoints

polamoros:
Resolves https://github.com/winglang/wing/issues/6579


apps/wing-console/console/design-system/src/modal.tsx
```diff
@@ -12,6 +12,25 @@ export interface ModalProps {
   children?: ReactNode;
 }
 
+export const ModalFooter = ({ children }: { children: ReactNode }) => {
```
skyrpex, line 15: Try to leave the `ModalProps` interface and the `Modal` component together. I'd move ModalFooter and ModalBody before those.

skyrpex (changes requested):

apps/wing-console/console/ui/src/features/endpoints-pane/endpoint-tree.tsx
```diff
@@ -1,86 +1,125 @@
...
+
+  const endpointsSortedList = useMemo(() => {
+    return (
+      endpointList.data?.sort(
+        (a, b) =>
+          a.exposeStatus.localeCompare(b.exposeStatus) ||
+          a.label.localeCompare(b.label),
+      ) ?? []
+    );
+  }, [endpointList.data]);
```
skyrpex, line 122: Careful here, the `Array.prototype.sort` mutates the array. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort.

Maybe we should clone before mutating.

```suggestion
  const endpointsSortedList = useMemo(() => {
    return (
      [...endpointList.data ?? []].sort(
        (a, b) =>
          a.exposeStatus.localeCompare(b.exposeStatus) ||
          a.label.localeCompare(b.label),
      )
    );
  }, [endpointList.data]);
```

skyrpex (changes requested):

apps/wing-console/console/ui/src/features/endpoints-pane/endpoint-tree.tsx
polamoros: fixed 👍

skyrpex (approved):

PR #6863, opened 2024-07-08
fix(console): empty space at the start of long log lines

skyrpex:
The commit 2860cad0f6acc523a19f5e66ab871c227382b94b introduced an error where long log lines are visualized with an empty space at the beginning.


polamoros (approved):

skyrpex:
This is the error:

image

PR #6867, opened 2024-07-08
fix(console): some resources are marked as started even though they are stopped

### LINKED ISSUE #6858: Console: some resources are marked as "Started" even though they aren't
```wing
bring cloud;
bring util;
bring sim;

let service1 = new sim.Resource(inflight () => {
  util.sleep(5s);
  throw "Oops";
}) as "Service1";

let service2 = new sim.Resource(inflight () => {
  util.sleep(5s);
  throw "Oops";
}) as "Service2";

```

Notice that `Service1` begins to start, and `Service2` is marked as started until `Service1` finished. Then, `Service2` is marked as starting.

You can use the reload button in the status bar to trigger it again.

https://github.com/winglang/wing/assets/1077520/424bb3f0-8cb9-4b62-8fc7-9bb6d3466a2d

We have to revisit the implementation of `libs/wingsdk/src/simulator/simulator.ts` and `apps/wing-console/console/server/src/router/app.get-hierarchichal-running-state.ts`, which defaults undefined running states to "started".

skyrpex:
This changeset fixes the running state of constructs that don't have any kind of runtime code.

It basically visits all remaining constructs after successfully starting all resources and marks them as started.

Fixes #6858.

### DIFF
--- a/apps/wing-console/console/server/src/router/app.get-hierarchichal-running-state.ts
+++ b/apps/wing-console/console/server/src/router/app.get-hierarchichal-running-state.ts
@@ -20,7 +20,7 @@ export const getHierarchichalRunningState = (
   }
 
   const runningState = (node?.resourceConfig?.attrs?.["runningState"] ??
-    "started") as RunningState;
+    "stopped") as RunningState;
 
   if (node.children?.length > 0) {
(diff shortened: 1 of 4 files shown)


skyrpex:
Will implement in a different way after a discussion with @Chriscbr

Issue #6818, opened 2024-07-01
Console: Test runner fails when running more than one test at a time


skyrpex:
The error seems to be related to using multiple simulator instances sharing the same output directory. The message is `Cannot stop a simulation that is still starting.`.

skyrpex:
Related to #5349.

skyrpex:
@polamoros please have a look and suggest a solution for this problem here.

monadabot:
Congrats! :rocket: This was released in [Wing 0.77.13](https://github.com/winglang/wing/releases/tag/v0.77.13).

PR #6876, opened 2024-07-09
fix(console): test runner fails when running more than one test at a time

polamoros:
Resolves https://github.com/winglang/wing/issues/6818


apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,58 @@
...
+}) => {
+  let testCompiler: Compiler;
+  let simfilePath = "";
+
+  const createSimulatorInstance = async (simfile?: string) => {
+    if (!simfile && !simfilePath) {
+      throw new Error("No simfile provided");
+    } else if (!simfile) {
+      simfile = simfilePath;
+    }
```
skyrpex, line 22: I believe you can get rid of the else here because the throw acts as an early return.

apps/wing-console/console/server/src/index.ts
```diff
@@ -216,8 +184,7 @@ export const createConsoleServer = async ({
   simulator.on("started", () => {
     appState = "success";
 
-    // Clear tests when simulator is restarted
-    testsStateManager().setTests([]);
+    testRunner.initialize();
```
skyrpex, line 187: Is it necessary to initialize the test runner after the simulator starts?

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,58 @@
...
+    const testSimulator = createSimulator();
+    await testSimulator.start(simfile);
+    await testSimulator.reload();
```
skyrpex, line 25: Is it necessary to reload just after a start?

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,58 @@
+import type { Simulator } from "../../wingsdk.js";
+import { createCompiler, type Compiler } from "../compiler.js";
+import { createSimulator } from "../simulator.js";
+
+export const createSimulatorManager = ({
+  wingfile,
+  platform,
+  watchGlobs,
+}: {
```
skyrpex, line 9: Can you explain what's the purpose of this manager?

apps/wing-console/console/server/src/utils/test-runner/test-runner.ts
```diff
@@ -0,0 +1,259 @@
...
+  runAllTests(): void;
+
+  // Register a callback to be called when the tests change.
+  onTestsChange(callback: () => void): void;
+
+  // Initialize the test runner.
+  initialize(): void;
+
+  // Stop the test runner.
+  stop(): void;
```
skyrpex, line 47: I don't think we need these. Initialization should happen automatically when calling any method, and it should also stop automatically after no use.

I can see the use of a `forceStop` method, though, when the server is being closed.

apps/wing-console/console/server/src/utils/test-runner/test-runner.ts
```diff
@@ -0,0 +1,259 @@
...
+      status: response.error ? "error" : "success",
+      time: response.time,
+      datetime: Date.now(),
+    });
+
+    runOnTestChangeCallbacks();
+    simulator.stop();
+  };
+
+  const runAllTests = async () => {
```
skyrpex, line 202: Calling `runTest` and `runAllTests` in sequence should work without crashing. I suggest that both of these methods use some kind of shared test queue.

apps/wing-console/console/server/src/utils/test-runner/test-state-manager.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+    initialized: () => {
+      return initialized;
+    },
+    restart: () => {
+      initialized = false;
+      tests = [];
+    },
+  };
+};
```
skyrpex, line 30: I don't understand what's this for 🤔

skyrpex (changes requested):

apps/wing-console/console/server/src/index.ts
polamoros: Yes if we want to restart the tests when there is a change in the wingfile (that causes a restart on the simulator)

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
polamoros: outdated

apps/wing-console/console/app/demo/main.w
```diff
@@ -8,11 +8,11 @@ let errorService = new cloud.Service(inflight () => {}) as "ErrorService";
 
 let errorResource = new sim.Resource(inflight () => {
   util.sleep(5s);
-  throw "Oops";
+  //throw "Oops";
```
polamoros, line 11: For some reason, the testing sim is not working for resources that throw exceptions :) .

working on it

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+import uniqueid from "lodash.uniqueid";
+
+import type { Simulator } from "../../wingsdk.js";
+import { createCompiler, type Compiler } from "../compiler.js";
+import { createSimulator } from "../simulator.js";
+
+const getSimulatorInstance = async (simfile: string) => {
+  const stateDir = mkdtempSync(
+    join(tmpdir(), `pwing-console-test-${uniqueid()}`),
+  );
```
skyrpex, line 14: Maybe we could use node packages only. See: https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

```js
import { mkdtemp } from 'node:fs/promises';
import { join } from 'node:path';
import { tmpdir } from 'node:os';

try {
  await mkdtemp(join(tmpdir(), 'foo-'));
} catch (err) {
  console.error(err);
}
```

apps/wing-console/console/server/src/utils/test-runner/test-state-manager.ts
```diff
@@ -0,0 +1,34 @@
...
+      const index = tests.findIndex((t) => t.id === test.id);
+      if (index === -1) {
+        tests.push(test);
+      } else {
+        tests[index] = test;
+      }
+      onTestsChange(test.id);
+    },
+  };
+};
```
skyrpex, line 34: I don't see the need for this file 🤔 It isn't really adding any value as far as I can see.

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      });
+    });
+  };
+
+  // Run a callback with a simulator instance.
+  const useSimulatorInstance = async <T>(
+    callback: (simulator: Simulator) => Promise<T>,
+  ): Promise<T> => {
+    const simulator = await getSimulator();
+    const result = await callback(simulator);
```
skyrpex, line 68: If the callback throws, the simulator will be kept running forever.

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+import { createSimulator } from "../simulator.js";
+
+const getSimulatorInstance = async (simfile: string) => {
+  const stateDir = mkdtempSync(
+    join(tmpdir(), `pwing-console-test-${uniqueid()}`),
+  );
+
+  const testSimulator = createSimulator({
+    stateDir,
+  });
```
skyrpex, line 18: Maybe use the SDK simulator directly instead of our wrapper. I think we don't need anything from the wrapper.

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      simulator.stop();
+    } catch (error) {
+      console.log(error);
+    }
+    return result;
+  };
+
+  const forceStop = () => {
+    testCompiler?.stop();
+  };
```
skyrpex, line 79: Maybe this should stop running simulators too.

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+  wingfile,
+  platform,
+  watchGlobs,
+}: {
+  wingfile: string;
+  platform?: string[];
+  watchGlobs?: string[];
+}) => {
+  let testCompiler: Compiler;
+  let testSimulator: Simulator;
```
skyrpex, line 37: I think it's not necessary to have this floating reference to a simulator.

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
skyrpex: The comments don't explain the behavior. From what I can tell:

- `createSimulatorManager ` will create one compiler (watcher) for the given wingfile when `useSimulatorInstance` is called for the first time
- The compiler will be kept running
- `useSimulatorInstance` will create a dedicated simulator and stop it at the end of the callback

apps/wing-console/console/server/src/utils/test-runner/simulator-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      testSimulator = await getSimulatorInstance(simfilePath);
+      return testSimulator;
+    }
+
+    testCompiler = createCompiler({
+      wingfile,
+      platform,
+      watchGlobs,
+      testing: true,
+    });
```
skyrpex, line 52: Maybe we could create the compiler at the beginning.

skyrpex (changes requested):

skyrpex (approved):

PR #6885, opened 2024-07-10
fix(console): console log levels are not persistent across restarts

### LINKED ISSUE #2806: Console log levels are not persistent across restarts
### I tried this:

Changed log level
Restarted the console

### This happened:

Log level reverted to default

### I expected this:

Log level to persist

### Is there a workaround?

_No response_

### Component

Wing Console

### Wing Version

_No response_

### Wing Console Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Anything else?

Related to #2552

### Community Notes

eladcon:
Using local storage to save some of the console view settings across app restarts

Fixes #2806

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/apps/wing-console/console/ui/src/features/localstorage-context/localstorage-context.tsx
+++ b/apps/wing-console/console/ui/src/features/localstorage-context/localstorage-context.tsx
@@ -33,7 +33,6 @@ export const AppLocalStorageProvider = (
       return hash;
     }
   }, [props.storageKey]);
-  console.log("AppLocalStorageProvider", { key });
   return (
     <AppLocalStorageContext.Provider
       value={{


apps/wing-console/console/ui/src/use-localstorage.ts
```diff
@@ -0,0 +1,79 @@
...
+  useLayoutEffect,
+  useEffect,
+  useMemo,
+} from "react";
+
+import { trpc } from "./trpc.js";
+
+const noop = () => {};
+
+export const useLocalStorage = <T>(
```
skyrpex, line 16: We have `react-use` installed. Is it possible to use https://github.com/streamich/react-use/blob/master/docs/useLocalStorage.md?

skyrpex:
I've noticed a delay in the stuff that uses the wingfile as part of the storage key... I'd like to avoid the delay, it feels very wrong.

Off the top of my head I think we could inject the wingfile into the HTML so it's available instantly to the JS.

apps/wing-console/console/ui/src/use-localstorage.ts
```diff
@@ -0,0 +1,79 @@
...
+  key: string,
+  initialValue: T,
+): [T, Dispatch<SetStateAction<T>>, boolean, () => void] => {
+  const wingfileQuery = trpc["app.path"].useQuery();
+  const wingfile = useMemo(() => {
+    return wingfileQuery.data;
+  }, [wingfileQuery.data]);
+
+  return useLocalStorage(wingfile ? `${wingfile}.${key}` : key, initialValue);
+};
```
skyrpex, line 79: I think we should change how this work a bit:

- Move the whole `wingfile` thing away
- Create a `AppLocalStorageProvider` that accepts a `storageKey` and setups a context with that key
- Change `useAppLocalStorage` so it uses the context above. If the context is not set, throw
- Use the `AppLocalStorageProvider` in the `Console.tsx` or `App.tsx` file, and display a loader until the `wingfile` isn't fetched
- Use a hash of `wingfile` as the `storageKey` instead of the whole `wingfile` string

skyrpex: With the implementation above we will avoid:

- Random behavior meanwhile the wingfile is being fetched
- Making the `use-localstorage` too dependent on the `app.path` / `app.wingfile` key

skyrpex (changes requested):

skyrpex (changes requested):

skyrpex (approved):

Issue #5857, opened 2024-03-06
Deprecate `ex` modules from Wing SDK


Chriscbr:
### Use Case

Most of these have been implemented as third party winglibs now, so they can be deprecated.

### Proposed Solution

Mark the classes/types as deprecated, announce it on our community channels, then remove them after a month.

### Component

SDK

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out on the #dev channel in the [Wing Slack](https://t.winglang.io/slack).

monadabot:
Congrats! :rocket: This was released in [Wing 0.78.0](https://github.com/winglang/wing/releases/tag/v0.78.0).

PR #6903, opened 2024-07-12
feat!: sunset ex.Table and ex.Redis

Chriscbr:
Closes https://github.com/winglang/wing/issues/5857

BREAKING CHANGE: `ex.Redis` and `ex.Table` have been removed from the standard library in favor of dedicated libraries maintained in Winglang. We recommend migrating apps to use the [redis](https://github.com/winglang/winglibs/tree/main/redis) winglib for Redis support, and the [dynamodb](https://github.com/winglang/winglibs/blob/main/dynamodb) or [postgres](https://github.com/winglang/winglibs/tree/main/postgres) winglibs for functionality similar to `ex.Table`.

## Checklist

- [x] Tests added (always)


eladb (approved):

docs/docs/04-standard-library/ex/redis.md
```diff
@@ -1,458 +0,0 @@
----
-title: Redis
```
staycoolcall911, line 2: @Chriscbr - I believe the `Redis` winglib suppurts only the `sim` target, while the `ex.Redis` module supports both. I would sunset `ex.Redis` only after we migrate support for `tf-aws` and/or `awscdk` in the `Redis` winglib. See https://github.com/winglang/winglibs/blob/main/redis/redis.w

Chriscbr: Do we know of anyone using the Redis module in production right now? I understand the value in supporting users, but if the APIs aren't being unused I feel like it's better to avoid prolonging this cleanup.. the less code we have to maintain the faster we can ship.

skyrpex (approved):

boyney123:
Looks like these also need removing from the docs? I believe these parts are automated for the standard lib? @Chriscbr ?

Chriscbr:
@boyney123 Yeah you're right, I'll look into removing them

PR #6908, opened 2024-07-15
fix(console): preflight logs don't appear in the logs panel

### LINKED ISSUE #3698: log function logs to console twice
### I tried this:

Inside the hello.w 
------------------
log("2+2");

$ wing it hello.w

### This happened:

Output was logged twice
> 2+2
> 2+2 

So what is happening is if we use wing it Hello.w
and Hello.w contains
------------------------
log("2+2");
Output :
2+2
2+2

But now in the already opened file in Vscode and i edit the Hello.w but adding say Log("2+3") below the first log statement;

Hello.w
----------
log("2+2")  // already present
log("2+3")
and press save
It prints correctly (ie. Both things once)
2+2
2+3

### I expected this:

I expected it once 
> 2+2

I am running in on VSCode

### Is there a workaround?

_No response_

### Component

Language Design, Compiler, SDK

### Wing Version

0.24.44

### Node.js Version

v18.16.0

### Platform(s)

Windows

### Anything else?

Screenshot 2023-08-03 224605
Screenshot 2023-08-03 231806

### Community Notes

skyrpex:
Moves the preflight log entries (those emitted when compiling wing code) to the console logs panel. Previously, they were emitted through stdout to the terminal that launched the console.

In order to achieve so, adds an optional `preflightLog` option to the `compile` function in the `@winglang/compiler` package. By default, preflight logs are emitted through `console.log`.

Fixes #3698.


libs/wingcompiler/src/compile.ts
```diff
@@ -290,7 +293,7 @@ npm i @wingcloud/framework
     const arg = `${normalPath(props.entrypointFile)};${normalPath(props.workDir)};${normalPath(
       props.wingDir
     )}`;
-    props.log?.(`invoking %s with: "%s"`, WINGC_COMPILE, arg);
+    // props.log?.(`invoking %s with: "%s"`, WINGC_COMPILE, arg);
```
skyrpex, line 296:
```suggestion
    props.log?.(`invoking %s with: "%s"`, WINGC_COMPILE, arg);
```

skyrpex:
This is how these new entry logs are:

image

polamoros (approved):
Is there any way to filter the compiler logs only?

PR #6911, opened 2024-07-15
fix(sdk): use lockfiles to prevent multiple simulators on the same state directory

### LINKED ISSUE #5799: `wing it` in the same directory uses the same state directory
### I tried this:

Running `wing it` twice in the same directory

### This happened:

The simulations use the same state files

### I expected this:

I expected the simulation runs to use separate state, or to see a warning / error of some kind.

Each simulated resource is manages its own simulated files, and having more than one simulation of the same resource touching the same files may result in race conditions, data loss, or hard to resolve bugs.

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

0.59.12

### Node.js Version

20.11.0

### Platform(s)

MacOS

### Community Notes

skyrpex:
Uses a lockfile to prevent multiple simulators on the same state directory, which would cause issues. The lockfile is periodically updated in order to check for staleness.

Additionally, adds a `--statedir "mystatedir"` flag to `wing it` that allows users changing the simulator state directory.

Fixes #5799.


libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -314,6 +323,36 @@ export class Simulator {
...
   }
 
+  private async acquireStateLock() {
+    if (this.lockfile) {
+      return; // already locked
+    }
+
+    await mkdir(this.statedir, { recursive: true });
+    try {
+      this.lockfile = await openFileHandle(this.lockfileFilename, "wx");
```
Chriscbr, line 333: I wonder what we can do to make the mechanism more robust to crashes (maybe storing a process ID or timestamp inside that can be read? not sure)

libs/wingsdk/src/simulator/simulator.ts
Chriscbr, line 1: Test?

Chriscbr (approved):

libs/wingsdk/src/simulator/simulator.ts
skyrpex: I've been looking out there and found https://www.npmjs.com/package/proper-lockfile. They check for staleness by constantly updating the lockfile, and use `mkdir` instead of `open` to make it network-compatible.

Not sure how a process ID in the lockfile would help, though.

I think we are good if we just implement the staleness check. No need for network compatibility, IMO. What do you think?

libs/wingsdk/src/simulator/simulator.ts
skyrpex: On it 💪🏻

libs/wingsdk/src/simulator/simulator.ts
Chriscbr: Oh I see, so basically checking if it's stale based on its timestamp. That sounds good to me 👍

libs/wingsdk/src/simulator/lockfile.ts
```diff
@@ -0,0 +1,125 @@
...
+    return false;
+  }
+};
+
+const UPDATE_INTERVAL = 1000;
+const STALE_THRESHOLD = 5000;
+
+export interface LockfileProps {
+  readonly path: string;
+  readonly onCompromised?: () => void;
```
Chriscbr, line 21: Maybe this can be async?

libs/wingsdk/src/simulator/simulator.ts
```diff
@@ -225,9 +226,17 @@ export class Simulator {
...
 
+  private lockfile: Lockfile;
+
   constructor(props: SimulatorProps) {
     const simdir = resolve(props.simfile);
     this.statedir = props.stateDir ?? join(simdir, ".state");
+    this.lockfile = new Lockfile({
+      path: join(this.statedir, ".lock"),
+      onCompromised: () => {
+        this.stop().catch(() => {});
```
Chriscbr, line 237: This would cause any exceptions raised by stop() to be caught / hidden from the user, I wonder if we should avoid that? Maybe
```suggestion
        void this.stop();
```

libs/wingsdk/src/simulator/lockfile.ts
```diff
@@ -0,0 +1,125 @@
...
+  try {
+    await fs.stat(path);
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+const UPDATE_INTERVAL = 1000;
+const STALE_THRESHOLD = 5000;
```
Chriscbr, line 17: (optional) Add a short description for each?

libs/wingsdk/src/simulator/lockfile.ts
```diff
@@ -0,0 +1,125 @@
...
+          if (this.lastMtime) {
+            try {
+              const stats = await fs.stat(this.path);
+              if (stats.mtimeMs !== this.lastMtime) {
+                this.markAsCompromised(
+                  "Lockfile was updated by another process"
+                );
+                return;
+              }
+            } catch (_error) {
```
Chriscbr, line 85: Maybe it's PTSD but I get weary whenever we're dropping error information since the only other way for a user to get it would be to change the source code and rebuild the project. Can we log the original error some way? (perhaps only if DEBUG=1 or something...)

libs/wingsdk/src/simulator/lockfile.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (await exists(this.path)) {
+      try {
+        const stats = await fs.stat(this.path);
+        if (Date.now() - stats.mtimeMs > STALE_THRESHOLD) {
+          await fs.rm(this.path);
+        }
+      } catch (error) {
+        throw new Error("Failed to check lockfile status: " + error);
+      }
+    }
```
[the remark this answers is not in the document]
Chriscbr: I suppose there could be an edge case where the file exists but the permission has changed.

Chriscbr (approved):

libs/wingsdk/src/simulator/lockfile.ts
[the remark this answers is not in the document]
skyrpex: I guess we could assume it got compromised then, too?

libs/wingsdk/src/simulator/simulator.ts
skyrpex: My intention was to hide problems like "Can't stop simulator if it's not running" but that should never happen in the first place... So good call, will remove that.

libs/wingsdk/src/simulator/lockfile.ts
skyrpex: Yeah, will change it.

libs/wingsdk/src/simulator/lockfile.ts
skyrpex: 👍🏻 I believe there's already a log mechanism in place, will have a look.

skyrpex: Now, if there's a related error, it's passed to the onCompromised callback so the simulator can log it.

libs/wingsdk/src/simulator/lockfile.ts
skyrpex: Done.

libs/wingsdk/src/simulator/simulator.ts
skyrpex: The simulator now logs these errors too 👍🏻

polamoros (approved):

Issue #6129, opened 2024-04-03
Winglibs don't work with yarn or pnpm


eladcon:
### I tried this:

try `@winglibs/cognito` with yarn as the package manager 

### This happened:

```
An error has occurred:
Failed to compile.

error: Unable to load "@cdktf/provider-aws": Module not found in "/private/tmp/a2/node_modules/@winglibs/cognito/platform/tfaws.w"
  --> node_modules/@winglibs/cognito/platform/tfaws.w:3:1
  |
3 | bring "@cdktf/provider-aws" as aws;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Unable to load "@cdktf/provider-aws": Module not found in "/private/tmp/a2/node_modules/@winglibs/cognito/platform/tfaws.w"

```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

MarkMcCulloh:
Wing's ecosystem makes heavy use of peer dependencies. yarn does not have a builtin way to install peer dependencies. In addition to that, the now-default PnP mechanism that yarn uses is likely not something we're going to support anytime soon.

Given that, I don't think yarn is something we will support. npm is the only officially supported package manager for now while pnpm and bun both unofficially work as well.

A workaround if yarn must be used, it must not be with the PnP nodeLinker mode and any peer dependencies must be manually installed. For example, to use `@winglibs/cognito` you must run `yarn add @winglibs/cognito @cdktf/provider-aws cdktf constructs`

To close out this issue we should officially document package manager support level.

ekeren:
Does this issue only surface when installing a @winglib? Or are there issues if I install wing with yarn add (not globally)

Chriscbr:
In the meantime we might be able to soften the blow of these errors by raise a compiler warning if there's a `yarn.lock` or `pnpm-lock.yaml` file in the project.

mingfang:
this is ironic because the recommended tool for building wing itself was pnpm.
https://www.winglang.io/contributing/start-here/development

PR #6961, opened 2024-07-29
feat(compiler): warning if unsupported package manager is used

Chriscbr:
Adding a small warning to help users encountering issues using winglibs with other JavaScript package managers. (https://github.com/winglang/wing/issues/6129). Screenshot of what the diagnostic looks like:

Screenshot 2024-07-29 at 7 34 50 PM

I ended up also removing some code for injecting panics into the compiler that has seldom been used in my experience (it introduced some snapshot regressions and removing the code seemed like the simpler fix).

## Checklist

- [ ] Tests added (always)


libs/wingc/src/lib.rs
```diff
@@ -156,6 +156,23 @@ pub unsafe extern "C" fn wingc_malloc(size: usize) -> *mut u8 {
 	}
 }
 
+/// Check if the project has a pnpm or yarn or bun lockfile, and emit a warning if it does.
+fn emit_warning_for_unsupported_package_managers(project_dir: &Utf8Path) {
+	let lockfiles = ["pnpm-lock.yaml", "yarn.lock", "bun.lock", "bun.lockb"];
```
MarkMcCulloh, line 161: nit: `const`

libs/wingc/src/lib.rs
```diff
@@ -156,6 +156,23 @@ pub unsafe extern "C" fn wingc_malloc(size: usize) -> *mut u8 {
...
+	for lockfile in &lockfiles {
+		let lockfile_path = project_dir.join(lockfile);
+		if lockfile_path.exists() {
+			report_diagnostic(Diagnostic {
+				message: "The current project has a pnpm/yarn/bun lockfile. Wing hasn't been tested with package managers besides npm, so it may be unable to resolve dependencies to Wing libraries when using these tools. See https://github.com/winglang/wing/issues/6129 for more details.".to_string(),
```
MarkMcCulloh, line 166: nit: `const`, or maybe `static` since it's a `String`

libs/wingc/src/lib.rs
```diff
@@ -272,6 +291,8 @@ pub fn compile(
 		&mut asts,
 	);
 
+	emit_warning_for_unsupported_package_managers(&project_dir);
```
MarkMcCulloh, line 294: To avoid the cost of multiple syscalls for the average case, what if this check only happens if there are errors? Ideally, we'd only want to check this if we specifically fail package resolution.

libs/wingc/src/parser.rs
```diff
@@ -2254,11 +2255,6 @@ impl<'s> Parser<'s> {
 			"json_literal" => self.build_json_literal(&expression_node, phase),
 			"struct_literal" => self.build_struct_literal(&expression_node, phase),
 			"optional_unwrap" => self.build_optional_unwrap_expression(&expression_node, phase),
-			"compiler_dbg_panic" => {
-				// Handle the debug panic expression (during parsing)
-				dbg_panic!();
-				Ok(Expr::new(ExprKind::CompilerDebugPanic, expression_span))
-			}
```
MarkMcCulloh, line 2261: 🥳

libs/wingc/src/lib.rs
```diff
@@ -156,6 +156,23 @@ pub unsafe extern "C" fn wingc_malloc(size: usize) -> *mut u8 {
 	}
 }
 
+/// Check if the project has a pnpm or yarn or bun lockfile, and emit a warning if it does.
+fn emit_warning_for_unsupported_package_managers(project_dir: &Utf8Path) {
```
MarkMcCulloh, line 160: Inverting the check for the success case (look for a `package-lock.json` instead) could be useful since there's only one case that is supposed to work and many cases that might not.

A `package-lock.json` file may not be present in cases where there are no dependencies, but ideally we shouldn't even be checking this in that case.

MarkMcCulloh (approved):
Left some comments but I understand we probably don't want to overcomplicate this so it LGTM as-is 👍

libs/tree-sitter-wing/grammar.js
```diff
@@ -463,9 +461,6 @@ module.exports = grammar({
         )
       ),
 
-    compiler_dbg_panic: ($) => "😱",
```
eladb, line 466: If I recall the whole idea of this "feature" (or backdoor) was to test that panics are handled properly and diagnostics are reported for them.

Maybe reconsider removing them.

eladb (approved):

libs/wingc/src/lib.rs
Chriscbr: > could be useful since there's only one case that is supposed to work and many cases that might not

I guess I'm OK with the possibility that this list might become outdated (since it's easy to add onto). Like you said, if the user has no dependencies (or maybe they just aren't saving lock files), package-lock.json might not be present so I'm not sure how we'd avoid the check otherwise.

MarkMcCulloh: It's not only about outdated cases, but cases this doesn't handle like workspaces since the lockfile would only present at the root, not necessarily next to a package.json

libs/wingc/src/lib.rs
Chriscbr: > Ideally, we'd only want to check this if we specifically fail package resolution.

Perhaps. But even if you use pnpm initially and only add a winglib later, getting the warning earlier in your journey as a first-time user might be better, I'd argue.

For now I'm optimizing for the pain of users being unable to compile their Wing apps, since that's a critical point where users can drop off (they might get confused by the error, have no idea how to fix it, and give up before asking on Discord or GitHub). Minimizing the number of warnings for happy Wing users who insist on using a different package manager is a secondary issue.

libs/tree-sitter-wing/grammar.js
MarkMcCulloh: Could be interesting to instead elevate this past a debug feature and turn it into an intrinsic ala `panic!` or `todo!`

libs/wingc/src/lib.rs
Chriscbr: Mmm I see. Yeah, fair enough, we can probably do more here (maybe do a lookup for the project's node_modules and check if there's a .pnpm folder, or only raise the error when a module lookup error occurs like you suggested) -- definitely open to improving it.

Let's merge this in for now, if you wanna try and fine tune the warning then I'll totes take the contribution

libs/tree-sitter-wing/grammar.js
Chriscbr: Noted.

(I think there are better ways to test the custom panic handler / CLI error message that don't require inserting special syntax and plumbing it through all of the stages of the compiler)

PR #6975, opened 2024-08-01
fix(console)!: not all diagnostics are errors

Chriscbr:
The `compile` function from the `@winglang/compiler` package throws even if the compilation was successful (e.g. the compiler emitted a warning). As result of that behavior, the console shows a blue screen of death when the compiler emits a warning.

This changeset refactors `compile` so it doesn't throw anymore, but returns the output directory (if possible) along with the diagnostics and any preflight errors.


skyrpex:
Looks good, although I'd like to change the console part a bit. We only want to log/report errors if `shouldCompileAgain` is false.

PR #6980, opened 2024-08-02
feat: internal access modifier

### LINKED ISSUE #4156: `internal` access modifier
### Feature Spec

`internal` access modifier can be used anywhere the `pub` access modifier can be applied (on type definitions, methods, fields, etc.).
When a type or method or other API element has the `internal` access modifier, it behaves like it's public (`pub`) within the current Wing library or module, but it will appear private when the library or module is imported by other Wing source code.

### Use Cases

```wing
class A {
  internal static do_a() {};
}
class B {
  public static do_b() {
    A.do_a(); // We don't want access to `do_a` outside this file, but we do need access to it in this exported public method
  }
}
```

### Implementation Notes

_No response_

### Component

Language Design, Compiler

### Community Notes

Chriscbr:
Closes #4156

It's been a while since I've worked with the symbol lookup subsystem. This area is in dire need of refactoring, unfortunately the system is a little too broad to take on in this PR (I'm also not sure what the immediate fix is off the top of my head). Some of the issues:
- It's a bit confusing which `lookup_xxx` method should be called in different scenarios
- The logic for handling on access modifiers errors is split between two places (once inside `type_check.rs` for determining if field/method accesses are valid, and once inside `symbol_env.rs` for determining if looking up types within environments are valid)
- The logic also needs to be added/maintained separately for the LSP.

There are also some edge cases I intentionally avoided getting side tracked with to focus on just getting the initial feature working. For example, there might be cases we should look towards other languages for how to handle, like in what situations is it okay to subclass a method as internal.

I did some manual QA to validate that autocompletions work as expected with the language server.

## Checklist

- [x] Tests added (always)


tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2843,6 +2843,104 @@ Test Files 1 failed (1)
 Duration <DURATION>"
 `;
 
+exports[`internal_access_modifiers.test.w 1`] = `
+"error: Expected identifier "fixture" to be a variable, but it's a namespace
+  --> ../../../examples/tests/invalid/internal_access_modifiers.test.w:3:9
+  |
+3 | let e = fixture.FavoritePlanets.MARS;
+  |         ^^^^^^^
```
Chriscbr, line 2851: Confusing error message if you try to access an internal enum

tools/hangar/__snapshots__/invalid.ts.snap
```diff
@@ -2843,6 +2843,104 @@ Test Files 1 failed (1)
...
+   |
+34 | internal interface InternalInterface {}
+   |                    ----------------- defined here
+
+
+error: Expected an interface, instead found type "unresolved"
+  --> ../../../examples/tests/invalid/internal_access_modifiers.test.w:6:22
+  |
+6 | class FakeClass impl fixture.InternalInterface {}
+  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^
```
Chriscbr, line 2870: Unneeded extra error when you extend an internal interface

eladb (approved):
Horray!

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: https://github.com/winglang/wing/issues/6985

tools/hangar/__snapshots__/invalid.ts.snap
Chriscbr: https://github.com/winglang/wing/issues/6986

PR #6988, opened 2024-08-05
feat(compiler)!: deprecate `@inflight` intrinsic

Chriscbr:
This PR deprecates `@inflight()`, an intrinsic function that was added to the Wing language with the aim of streamlining the process of integrating existing TypeScript code into Wing programs.

Today, the language currently supports two mechanisms for interoperating with TypeScript code: `extern` methods and `@inflight`. (See docs: https://github.com/winglang/wing/blob/9c36ce5fd386123744079db65e7e9b22f356a16a/docs/api/03-language/10-using-javascript.md). The main differences between them are:
- `extern` requires you to specify the signature of the function you're using in Wing, while signature of the `@inflight` function can be inferred
- `@inflight` allows you to interact with Wing resources from TypeScript, while `extern` only allows passing non-resource types (e.g. primitives, collections, functions)

The `@inflight` syntax has a few DX problems that make it tricky to use, without clear paths to resolution. The main issues are:
- It's hard for users to correctly set up the type-linkage between Wing and TypeScript.
- Type checking errors inside TypeScript files won't result in compilation errors to your Wing project. To provide a guarantee like this, we would require coupling our compilation process to that of TypeScript's.
- By exposing Wing resources to TypeScript through the "lifts" object, users to lose the permission-inferred experience that they get when writing inflight code in Winglang. If users specify the wrong permissions or "ops", the errors will only be caught at runtime.

The `@inflight` syntax was a valuable experiment and it's possible a variation on this will return in the future.

In the mean time, we're planning to focus on improving the experience of making TypeScript code available to Wing by bringing the type systems closer together, and improving the ergonomics of the language's `extern`, and exploring other ideas like supporting mechanisms for importing TypeScript libraries directly into Wing.

BREAKING CHANGE: `@inflight()` has been removed. We recommend using the existing `extern` method syntax for directly referencing JavaScript or TypeScript code that does not use Wing resources.

## Checklist

- [x] Tests added (always)


eladb (approved):
👍

PR #6999, opened 2024-08-06
fix: expected type to be `Array<Json>`, but got `Array<Json>` instead

### LINKED ISSUE #5421: Expected type to be "Array<Json>", but got "Array<Json>" instead
### I tried this:

```js
let extra = MutArray<Json>["bar"];
let base = [Json "foo"];
base.concat(extra.copy());
```

### This happened:

```
error: Expected type to be "Array<Json>", but got "Array<Json>" instead
  --> main.w:3:13
  |
3 | base.concat(extra.copy());
  |              ^^^^^^^^^^ Expected type to be "Array<Json>", but got "Array<Json>" instead
```

### I expected this:

No error

### Is there a workaround?

If I change the second statement to
```
let base = Array<Json>["foo"];
```
then it compiles.

### Anything else?

_No response_

### Wing Version

0.54.28

### Node.js Version

20.9.0

### Platform(s)

MacOS

### Community Notes

tsuf239:
fixes #5421
in addition to the examples in the issue, I added one with `MutMap` (since it was also missing).
I noticed that `Set` and `MutSet` aren't legal `Json` values, is that right @Chriscbr? (or should I add them as well) 

## Checklist

- [ ] Tests added (always)


Chriscbr:
> I noticed that Set and MutSet aren't legal Json values, is that right @Chriscbr?

That's correct - `Set` and `MutSet` in Wing map to the JavaScript [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) type, which can't be used as plain JSON.

examples/tests/valid/json-types.test.w
```diff
@@ -0,0 +1,25 @@
...
+
+let map = MutMap<str>{"a" => "1"};
+map.set("b", "2");
+
+let takeJson = (obj: Json) => {
+  
+};
+
+takeJson({ values: arr });
+takeJson({ values: map });
```
Chriscbr, line 25: These `takeJson` calls shouldn't be valid. MutArray isn't safe to convert directly to Json (for the same reason `MutArray` can't passed to a function that expects an `Array`). But it would be ok to implicitly convert a `MutArray` to `MutJson`.

Chriscbr (changes requested):

examples/tests/valid/json-types.test.w
tsuf239: Fixed-
could not get to the implicit conversion since `MutJson` and `Json` behave strangely- #7002 #7003

Chriscbr (approved):

Issue #6915, opened 2024-07-16
unable to use `util.exec` in preflight


eladb:
### I tried this:

```
bring util;

let x = util.exec("ls", []);
log(x.stdout);
```

### This happened:

```
undefined
```

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

monadabot:
Congrats! :rocket: This was released in [Wing 0.83.1](https://github.com/winglang/wing/releases/tag/v0.83.1).

Chriscbr:
It looks like continues to error on Windows machines - see this build error (https://github.com/winglang/wing/actions/runs/11226171454/job/31329344016):

```
 × src/test_corpus/doc_examples/35-exec-processes.md_example_1/main.w.test.ts > wing test -t sim 1710ms
   → Error: Program not found: spawnSync bash ENOENT
   --> main.w:14:15
   | let output3 = util.exec("bash", ["--norc", "--noprofile", "-c", "echo $WING_TARGET $ENV_VAR"], { inheritEnv: true });
   | 
   | // exec with current working directory
14 | let output4 = util.exec("bash", ["--norc", "--noprofile", "-c", "echo Hello"], { cwd: "/tmp" });
   |               ^
at D:/a/wing/wing/tests/doc_examples/valid/35-exec-processes.md_example_1/main.w:14:15
Tests 1 failed (1)
Snapshots 1 skipped
Test Files 1 failed (1)
Duration 0m0.98s
```

PR #7019, opened 2024-08-14
fix(sdk): unable to use `util.exec` in preflight

tsuf239:
fixes: #6915

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/examples/tests/sdk_tests/util/exec.test.w
+++ b/examples/tests/sdk_tests/util/exec.test.w
@@ -15,9 +15,9 @@ let assertThrows = inflight (expected: str, block: (): void) => {
 
 let output1 = util.exec("echo", ["-n", "Hello, Wing!"]);
   
-  expect.equal(output1.stdout, "Hello, Wing!");
-  expect.equal(output1.stderr, "");
-  expect.equal(output1.status, 0);
+expect.equal(output1.stdout, "Hello, Wing!");
+expect.equal(output1.stderr, "");
+expect.equal(output1.status, 0);
 
 
 test "exec()" {
--- a/examples/tests/sdk_tests/util/exec.test.w
+++ b/examples/tests/sdk_tests/util/exec.test.w
@@ -13,6 +13,12 @@ let assertThrows = inflight (expected: str, block: (): void) => {
   assert(error);
 };
 
+let output1 = util.exec("echo", ["-n", "Hello, Wing!"]);
+  
+  expect.equal(output1.stdout, "Hello, Wing!");
+  expect.equal(output1.stderr, "");
+  expect.equal(output1.status, 0);
+
 
 test "exec()" {
   // "exec() with valid program"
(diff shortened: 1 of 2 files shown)


examples/tests/sdk_tests/util/exec.test.w
```diff
@@ -13,6 +13,12 @@ let assertThrows = inflight (expected: str, block: (): void) => {
   assert(error);
 };
 
+let output1 = util.exec("echo", ["-n", "Hello, Wing!"]);
+  
+  expect.equal(output1.stdout, "Hello, Wing!");
+  expect.equal(output1.stderr, "");
+  expect.equal(output1.status, 0);
```
eladb, line 20:
```suggestion
expect.equal(output1.stdout, "Hello, Wing!");
expect.equal(output1.stderr, "");
expect.equal(output1.status, 0);
```

libs/wingsdk/src/util/util.ts
```diff
@@ -223,11 +222,11 @@ export class Util {
    * @param opts `ExecOptions`, such as the working directory and environment variables.
    * @returns A struct containing `stdout`, `stderr` and exit `status` of the executed program.
    */
-  public static async exec(
+  public static exec(
```
eladb, line 225: @Chriscbr what do you think? It's not ideal that exec will be sync in inflight.

Is there a plan to support preflight/inflight variants of utils?

Chriscbr: I agree. Right now the rough plan is to allow unphased functions to have separate preflight and inflight implementations, at least if they're implemented with JavaScript. Check out the RFC we have open for more details: https://github.com/winglang/wing/pull/1711

For now I'm cool with trying out this change just to be pragmatic since I'm not we support this with our TypeScript-based SDK (but it will be nice to get this working eventually)

libs/wingsdk/src/util/util.ts
```diff
@@ -239,10 +238,10 @@ export class Util {
     };
 
     try {
-      const { stdout, stderr } = await execFilePromise(program, args, execOpts);
+      const stdout = execFileSync(program, args, execOpts);
       return {
         stdout: stdout.toString(),
-        stderr: stderr.toString(),
+        stderr: "",
```
Chriscbr, line 244: Maybe add a TODO comment or open an issue to note this

Chriscbr (approved):

libs/wingsdk/src/util/util.ts
tsuf239: There is no real limitation I think- the `stderr` is thrown in the catch statement if exists- and passed to the user :) 
I think that it yields the same result of the `promisify` function

Chriscbr: @tsuf239 I thought even if a command succeeds it can still emit to both stdout and stderr?

PR #7036, opened 2024-08-20
feat(compiler)!: interfaces, enums, and public classes must be defined at the top-level

### LINKED ISSUE #5542: Disallow interfaces definitions except at the top-level
### Use Case

For the same reason we disallow structs except at the top-level of a file (see https://github.com/winglang/wing/issues/5135). Interfaces only define contracts (they don't contain any source code) so there aren't strong use cases to support defining interfaces within if branches or as a result of code execution etc.

```js
bring math;

if math.random() < 0.5 {
  interface Foo {
    foo(): str;
  }
} else {
  interface Foo {
    foo(): num;
  }
}
```

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

Compiler

### Community Notes

### LINKED ISSUE #7035: `pub` or `internal` types can be declared inside local scopes
### I tried this:

```js
for i in 0..1 {
  pub class MyClass {}
  new MyClass();
}
```

### This happened:

The code compiles

### I expected this:

I expected an type checking error to be raised. There's no way for classes defined within inner scopes (within if statements, for loops, functions, or other classes) to be possibly consumed by other files, and it's possible to statically detect this based on where the class is defined in the code, so we should indicate to the user that their code is invalid.

The same logic should also apply for structs, enums, and interfaces as well.

### Is there a workaround?

_No response_

### Anything else?

_No response_

### Wing Version

0.82.7

### Node.js Version

20.11.1

### Platform(s)

_No response_

### Community Notes

Chriscbr:
Closes #7035
Closes #5542

To encourage more readable code and reduce the number of edge cases we might have to handle in the compiler implementation, we add the constraint that that interfaces, enums, and any exported classes have to be defined at the top level. This requirement is already in place for structs.

It's a bit unclear whether we want to apply the same restriction to classes, since classes can have code inside of them (unlike interfaces, enums, and structs) and you can certainly find plenty of examples where folks define anonymous classes inside functions in other languages, so for now we're continue to allow them, though support for it isn't as rigorously tested. We only require that if you define a class in a local context, it must be private, since it wouldn't be possible to access them in other Wing files.

BREAKING CHANGE: interfaces, enums, and classes marked with "pub" or "internal" are now required to be defined at the top-level of Wing programs. Please let us know if you encounter issues due to this change.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/docs/api/05-language-reference.md
+++ b/docs/api/05-language-reference.md
@@ -1365,6 +1365,7 @@ If no catch block exists among caller functions, the program will terminate.
 Structs are loosely modeled after typed JSON literals in JavaScript.  
 Structs are defined with the `struct` keyword.  
 Structs are "bags" of immutable data.
+Structs must be defined at the top-level of a Wing file.
 
 Structs can only have fields of primitive types, preflight classes, and other structs.  
 Array, set, and map of above types is also allowed in struct field definition.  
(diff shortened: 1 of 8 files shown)


skyrpex (approved):

eladb:
+1 not to do this for classes at the moment.

PR #7043, opened 2024-08-22
feat(console): use log files instead of in-memory logs

### LINKED ISSUE #7024: Use log files instead of in-memory logs in the console
The console currently stores every log entry in memory. This can lead to out of memory errors in the application runs long enough.

The proposed change includes storing the log entries in a log file, maybe using [jsonl](https://jsonlines.org/) as the format. The log consumer will use a handle to retrieve newer log entries.

This change is part of the console environments feature.

skyrpex:
This changeset refactors the console so it uses a file to store the logs (previously, logs were stored in memory). The log retrieval mechanism still reads the whole file, but some code capable of streaming line-based files was added for future references.

Closes #7024.


skyrpex:
Now I just need to make sure the react query's infinite query actually fetches the most recent data...

skyrpex:
Now I just need to make sure the react query's infinite query actually fetches the most recent data...

PR #7047, opened 2024-08-22
chore(compiler): store FQN information for types

### LINKED ISSUE #7041: Store FQN information throughout compiler
In Wing, user-defined types (like classes and enums) are automatically organized hierarchically based on the directory structure of the user's source code. As a human-readable way to understand the location of types (and correspondingly, how to import them), we associate each type or namespace with a "FQN" or fully-qualified name. For example, the `Bucket` class that comes included in the Wing cloud library has an FQN of `@winglang/sdk.cloud.Bucket`. The FQN is a dotted string formed by the name of the library, the names of any namespaces it belongs to, and the name of the type.

To implement the language so far we didn't need to store or manage this information in the compiler, but we're reaching a point where not having FQNs is starting to cause friction. One example is that the Wing platform system heavily depends on using FQN strings in preflight code, so we're limited by what code the compiler can generate today. Another example is that for generating docs, we'd like to know the FQN of every type and of every namespace.

To ensure we generate unique and correct FQNs throughout a Wing compilation, it seems like the best place to store this FQN information is directly within the structs in our compiler that correspond to namespaces and types. Some challenges I anticipate are described below.

## Name conflicts

Today, Wing allows you to have two classes, interfaces, or enums with the same name in a given namespace, by defining them in sibling files within the same directory. An error is only raised if the parent directory is imported with a `bring` statement and both types are public. [1] Even when the error is avoided (by making one of the types non-public or not bringing the directory), it's possible the type system would end up associating these types with the same FQN, which ought to be avoided.

Potential solutions:
1. Do not generate FQNs for non-public types. In Rust, we'd have to store the FQN as `fqn: Option<String>`. (Namespaces aren't affected by naming conflicts, so they can still have `fqn: String`).
2. Associate non-public types in the compiler with FQNs that are differentiated in some way (e.g. `@winglibs/momento.InternalClass#0294db` - here, the hash is generated based on the name of the source file).
3. Disallow user-defined types in the same directory from having the same name so that we can guarantee each type can be given a unique FQN. I think this is fine in terms of design, but it might result in a less consistent experience due to various edge cases: 
    - When you run `wing compile main.w`, the compiler doesn't try to read and parse all files in your current directory; it only reads the `main.w` file and any files it depends on. This means an error might not be raised if one of the files is currently unused.  It's also possible you might get an error when you open both files in your IDE, but when you compile the project no errors are raised.
    - If we detected two classes in different files with the same name, it could be tricky to ensure an error is reported on both classes.

Another edge case to consider is that the compiler technically has some limited support for nested classes right now. FQNs for such classes may be ambiguous or not serve a concrete purpose. 

[1] Another scenario is that the parent directory is packaged as a Wing library and both classes are public.

## Implementation approach

One snag we might hit is that we generally type check files in reverse-dependency order -- that is, if file A brings in types from file B, we type check file B before file A. This makes things tricky for us since that generally means the deepest files in our project are type checked first, and thus we need to calculate the FQNs for types in those files first.

Possible solutions:
1. Calculate the FQN of types based on a source file's relative path from the root of the project or library it's in. This means we need the parser to calculate what the root file of each project or library is so we can use it at type checking time.
2. Have the parser calculate a FQN for every source `File`. During type checking, we have information about what `File` we're type checking, so we can just retrieve the base FQN and append the name of the class or enum we're processing to calculate the FQN.

Chriscbr:
Closes #7041

TODO: add FQN information to enums, interfaces, structs

## Checklist

- [x] Tests added (always)

### DIFF
--- a/libs/wingc/src/lib.rs
+++ b/libs/wingc/src/lib.rs
@@ -224,10 +224,6 @@ pub unsafe extern "C" fn wingc_compile(ptr: u32, len: u32) -> u64 {
 	}
 	let source_path = Utf8Path::new(split[0]);
 	let output_dir = split.get(1).map(|s| Utf8Path::new(s)).expect("output dir not provided");
-	let project_dir = split
-		.get(2)
-		.map(|s| Utf8Path::new(s))
-		.expect("project dir not provided");
 
 	if !source_path.exists() {
 		report_diagnostic(Diagnostic {
@@ -240,7 +236,7 @@ pub unsafe extern "C" fn wingc_compile(ptr: u32, len: u32) -> u64 {
 		return WASM_RETURN_ERROR;
 	}
 
-	let results = compile(project_dir, source_path, None, output_dir);
+	let results = compile(source_path, None, output_dir);
 
 	if let Ok(results) = results {
 		string_to_combined_ptr(serde_json::to_string(&results).unwrap())
@@ -286,13 +282,36 @@ pub fn type_check(
 	tc.type_check_file_or_dir(file, scope);
 }
 
-pub fn compile(
-	project_dir: &Utf8Path,
-	source_path: &Utf8Path,
-	source_text: Option<String>,
-	out_dir: &Utf8Path,
-) -> Result<CompilerOutput, ()> {
-	let source_package = as_wing_library(project_dir).unwrap_or_else(|| DEFAULT_PACKAGE_NAME.to_string());
+/// Infer the root directory of the current Wing application or library.
+///
+/// Check the current file's directory for a wing.toml file or package.json file,
+/// and continue searching up the directory tree until we find one.
+/// If we run out of parent directories, fall back to the first directory we found.
+pub fn find_nearest_wing_project_dir(source_path: &Utf8Path) -> Utf8PathBuf {
+	let initial_dir = if source_path.is_dir() {
+		source_path.to_owned()
+	} else {
+		source_path.parent().unwrap().to_owned()
+	};
+	let mut current_dir = initial_dir.as_path();
+	loop {
+		if current_dir.join("wing.toml").exists() {
+			return current_dir.to_owned();
+		}
+		if current_dir.join("package.json").exists() {
+			return current_dir.to_owned();
+		}
+		if current_dir == "/" {
+			break;
+		}
+		current_dir = current_dir.parent().unwrap();
+	}
+	return initial_dir;
+}
+
+pub fn compile(source_path: &Utf8Path, source_text: Option<String>, out_dir: &Utf8Path) -> Result<CompilerOutput, ()> {
+	let project_dir = find_nearest_wing_project_dir(source_path);
+	let source_package = as_wing_library(&project_dir).unwrap_or_else(|| DEFAULT_PACKAGE_NAME.to_string());
 	let source_path = normalize_path(source_path, None);
 	let source_file = File::new(&source_path, source_package.clone());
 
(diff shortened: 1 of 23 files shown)


libs/wingc/src/lib.rs
```diff
@@ -286,15 +283,42 @@ pub fn type_check(
...
+		if current_dir == "/" {
+			break;
+		}
+		current_dir = current_dir.parent().unwrap();
+	}
+	return initial_dir;
+}
+
+pub fn compile(source_path: &Utf8Path, source_text: Option<String>, out_dir: &Utf8Path) -> Result<CompilerOutput, ()> {
+	let project_dir = find_nearest_wing_project_dir(source_path);
```
Chriscbr, line 314: I updated the main `compile` function so it's now responsible for looking through your file system for the root project you're in. The main motivation is to unify the logic with the LSP -- in the LSP, we don't know where the root of the project is, which makes it tricky for us to know where is the "root" is for the purposes of automatically calculating FQNs according to the user's directory structure. So in both places, we now try to look for a package.json or wing.toml and if we find one, we treat that as our project root.

There's a chance this heuristic will be too aggressive 🤷 - but I think the best way to go about this is eventually to lean into it and to recommend building projects with `wing.toml` files.

libs/wingc/src/parser.rs
```diff
@@ -262,7 +280,10 @@ fn parse_wing_file(
 
 	// Parse the source text into an AST
 	let parser = Parser::new(&source_text.as_bytes(), source_file.to_owned());
-	let (scope, dependent_wing_paths) = parser.parse(&tree_sitter_root);
+	let (scope, dependent_wing_paths, found_library_roots) = parser.parse(&tree_sitter_root);
+	for (name, path) in found_library_roots {
+		library_roots.insert(name, path);
+	}
```
Chriscbr, line 286: My end goal here was when we're done with parsing, we should have a list of all of the libraries we import (plus the current project or library) and their root paths, so we can use that info in the type checker for calculating FQNs.

Now where do we actually create or generate that data? It's not that clear what the easiest way to add this was -- I ended up just passing it around in the parser functions we have here, but I could imagine it might be better if we had some more flexible API that let us just query for this information or just hand the parser a callback that should be called every time a new library is encountered, etc. It might be worthwhile to see how other compilers handle this.

tools/hangar/__snapshots__/test_corpus/valid/bring_local_dir.test.w_compile_tf-aws.md
```diff
@@ -106,13 +106,13 @@ class $Root extends $stdlib.std.Resource {
     const w = $helpers.bringJs(`${__dirname}/preflight.widget-1.cjs`, $preflightTypesMap);
     const subdir = $helpers.bringJs(`${__dirname}/preflight.subdir2-6.cjs`, $preflightTypesMap);
     $helpers.nodeof(this).root.$preflightTypesMap = $preflightTypesMap;
-    const widget1 = new w.Widget(this, "widget1");
+    const widget1 = globalThis.$ClassFactory.new("rootpkg.subdir2.inner.Widget", w.Widget, this, "widget1");
```
Chriscbr, line 109: "rootpkg" is the default name we give for the assembly if the project has no name since every FQN needs one. It was an arbitrary choice (though I wanted to distinguish it from the word "root" which is used by itself in other parts of the SDK).

libs/wingc/src/type_check.rs
```diff
@@ -304,6 +304,9 @@ pub struct Namespace {
 
 	/// Where we can resolve this namespace from
 	pub module_path: ResolveSource,
+
+	/// The fully qualified name of this namespace
+	pub fqn: String,
```
eladb, line 309: Horray!

PR #7051, opened 2024-08-26
feat(console): use http batch streams

skyrpex:
The HTTP batch stream links in TRPC can resolve individually as soon as their responses finished streaming, making the batch stream faster than a normal batch.

### DIFF
--- a/apps/wing-console/console/ui/src/Console.tsx
+++ b/apps/wing-console/console/ui/src/Console.tsx
@@ -1,10 +1,9 @@
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import {
-  httpBatchLink,
   wsLink,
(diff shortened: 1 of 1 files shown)


Chriscbr (approved):

apps/wing-console/console/ui/src/Console.tsx
```diff
@@ -61,34 +60,15 @@ export const Console = ({
...
-            condition(op) {
-              return op.path.startsWith("test.");
-            },
-            true: httpLink({
-              url: trpcUrl,
-            }),
-            false: httpBatchLink({
-              url: trpcUrl,
-            }),
+          false: unstable_httpBatchStreamLink({
```
eladb, line 70: nice!

PR #7063, opened 2024-08-30
chore: huge monorepo refactor!

### LINKED ISSUE #5992: Repo organization: move tests out of `examples`
Currently all of our tests are under `/examples/tests`. That's a bit weird.

Can we just move them to `/tests` ?

Chriscbr:
This is a big change!

To improve the contributing experience, this PR refactors the structure of our monorepo so that individual packages from `libs/` and `apps/` are now all filed into `packages/` according to their npm package name.

The main goal here is to make it easier to find and discover packages when you get build errors. Today, we use [turborepo](https://turbo.build/repo/docs) to orchestrate build commands across these separate packages, and it's helpful in that even if you run a command that compiles several packages, errors that are raised will always be prefixed with the package's name. However, the monorepo has over two dozen packages, and their folder names are often separate from their package names. This makes it hard to locate a package based on its name in the build log.

To date, there also hasn't been any major benefits to splitting the packages into libraries and apps -- the distinction is a bit murky since it's possible for a package to function as both an entrypoint and as a library. It's a lot more convenient to be able to navigate to a package without needing to know whether it's in "examples" or "libs" or "apps".

Test fixtures have also been moved into the `packages/` which should make them easier to find, and `examples/tests/` has been moved to `tests/`.

All console packages are now in the `wing-console/` directory (separated since it has a source-available license).

The `tools/` packages are still in their own directory, but it could make sense to move them to `packages/` in the future as well.

Closes #5992

## Checklist

- [x] Tests added (always)


eladb (dismissed):
Ship it!

MarkMcCulloh:
Something external to update after this change: https://github.com/winglang/tree-sitter-wing/blob/main/sync.sh

That repo is effectively a read-only clone of the tree-sitter-wing package, so the directory in this sync script needs to be updated

boyney123:
image

🚀!

PR #7078, opened 2024-09-04
feat: `@app` and `@filename` intrinsics

### LINKED ISSUE #6058: Expose `app` global variable
### Use Case

I've been thinking about this more, and I'm starting to feel like maybe it could make sense to expose `app` as a global variable as a substitute for `this` (added in https://github.com/winglang/wing/pull/5594). Here's a design sketch:

```js
// main.w

app; // std.IApp

class MyFactory {
  pub static makeBucket() {
    new cloud.Bucket() in app; // can be used as a scope
  }
}

// instead of util.env("WING_TARGET")
if app.platform == "sim" {
  log("app compiled to simulator");
} else {
  log("app compiled to another platform");
}
```

`app` can also be available in module files, but it would need a few restrictions (see below). It could be emitted in JavaScript as `App.of(this)`.

```js
// mod.w

class Store {
  new() {
    if app.platform == "sim" {
      // ...
    } else {
      // ...
    }
  }
  
  static myStaticMethod {
    app; // error: Cannot use `app` in static context
  }
}

class Subclass extends Store {
  new() {
    app; // error: Cannot access `app` in a subclass before super() has been called
    super();
  }
}
```

The motivation is a few-folded:
- After we added https://github.com/winglang/wing/pull/5594 I started to think maybe "this" in entrypoint files could be confusing
- "this" in entrypoint files is typed as `IConstruct`, but it would be more useful to have access to app-related facilities, like the compilation platform. Supporting both `this` and `app` might be more confusing though (not sure)

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

Language Design

### Community Notes

Chriscbr:
Closes #6058

This PR introduces the `@app` and `@filename` intrinsics to the language. `@filename` is a complement to the existing `@dirname` intrinsic, and `@app` provides a way to access the globally accessible app. `@app` exposes some meta information that can be useful, like whether the app is being compiled as a test or not, and the platform parameter system. In the CDK world, the app is also the "root" of the tree that all resources get added to.

## Checklist

- [x] Tests added (always)


Chriscbr:
Mmm... I thought `this` wasn't necessary, but I found an example where it still seems like it's load-bearing. Basically, if you compile your Wing app to have multiple tests, `this` will take on different values, so this is valid:

```
bring cloud;
bring expect;

class BucketFactory {
  pub static makeBucket(scope: constructs.IConstruct): cloud.Bucket {
    let bucket = new cloud.Bucket() in scope;
    // apply customizations to bucket...
    return bucket;
  }
}

let bucket = BucketFactory.makeBucket(this);

test "can use bucket" {
  bucket.put("hello", "world");
  expect.equal(bucket.list().length, 1);
}

test "can use other bucket" {
  bucket.put("yo", "sup");
  expect.equal(bucket.list().length, 1);
}
```

But if you were to change it to:

```
let bucket = BucketFactory.makeBucket(@app);
```

Then the test can't be run on AWS, because in preflight it tries creating two buckets with the same scope and name.

I'm not totally sure how to resolve this, so I'll revert a few of the changes and leave in `this` for now.

eladb (changes requested):
I think these need to be documented somewhere

Chriscbr:
@eladb missed that, thanks - I added it to the language ref

PR #7122, opened 2024-09-12
feat: `bytes` type

### LINKED ISSUE #7118: `bytes` type
### Use Case

A lot of Wing code that's written today works with structured data like strings, floats, and JSON, with less of a focus on low level bit twiddling and performance optimizations -- in part because the compiler translates most code into JavaScript.

But for working with APIs in the real world, it's sometimes necessary to have some way to represent opaque binary data so that you can save it to disk or pass it to low level functions. For example, file contents aren't always strings, and not being able to represent their opaque data during program execution limits the kinds of programs you can work. This has also been a limitation for working with object storage services like S3 in Wing.

### Proposed Solution

A dedicated `bytes` type.

With first-class `bytes` support, we can provide APIs that work with bytes directly -- for example, to support downloading and uploading raw file data:

```js
bring http;
bring util;

let downloadImageAndSaveToBucket = inflight (imageUrl: str): str => {
  let response = http.get(imageUrl);

  if response.status == 200 {
    let key = util.uuidv4();
    bucket.putBytes(key, response.body.bytes());
  } else {
    throw "received status code: {response.status}";
  }
};
```

(for larger files, you may want to save the file to disk first and then perform a multi-part upload)

### Implementation Notes

This will help with addressing issues like:
- https://github.com/winglang/wing/issues/1686
- https://github.com/winglang/wing/issues/5881

### Component

Compiler

### Community Notes

Chriscbr:
Closes #7118 

This implements the immutable variant of `bytes`. The mutable version is left for follow up work as it's not as critical to support.

## Checklist

- [x] Tests added (always)


skyrpex:
Any plans on adding streaming APIs? The current methods only load entire files... what about big files that need to be processed in steps, for example?

Chriscbr:
> Any plans on adding streaming APIs? The current methods only load entire files... what about big files that need to be processed in steps, for example?

I left it out of the base `bytes` RFC to avoid more scope creep, but I'm totally with you, Wing should data streaming (at least in inflight). I'll open an issue for further design discussion. 

I think since `bytes` is implemented in terms of JavaScript's `Uint8Array` (which is supported by Node.js and browsers), any further streaming additions should be compatible with this API. But we can adjust or revisit if needed

PR #7146, opened 2024-09-19
fix(compiler): unexpected StripPrefixError

Chriscbr:
Fixes a bug where the compiler can panic in certain situations when files from different directories are loaded. Specifically, the issue happens when a Wing file tries importing another Wing file from a directory that is not a sibling or in a sibling directory.

The issue is related to a facility of the compiler responsible for calculating a FQN (fully-qualified name) for every public type. To calculate the FQN for a public type like a class, we need to know where the file is relative to its package. For example, when a file like `/home/node_modules/@winglibs/dynamodb/aws/table.w` is type checked, the compiler needs to know that it belongs to the library named `@winglibs/dynamodb` and that the root of `@winglibs/dynamodb` is `/home/node_modules/@winglibs/dynamodb`. This way, types defined inside `aws` will get added to the `aws` namespace, and types defined at the root should get added to the root namespace, and if a type is defined outside of the project root, it's an error.

Today, the way we identify what package each source file belongs to is by looking up the file system and seeing if there's a package.json or wing.toml file somewhere up the file tree. But we don't enforce all projects to have these files and we want the language server to work even when these files don't exist. In such cases, we decide that all files that aren't part of a well-defined package are part of an implicit "default" package, and the root of this default package is whatever is the common root of all of its files.

This PR should also slightly improve the stability of the language server.

## Checklist

- [x] Tests added (always)

### DIFF
--- a/packages/@winglang/wingc/src/type_check.rs
+++ b/packages/@winglang/wingc/src/type_check.rs
@@ -7496,27 +7496,58 @@ fn lookup_known_type(name: &'static str, env: &SymbolEnv) -> TypeRef {
 /// ```
 pub fn calculate_fqn_for_namespace(package_name: &str, package_root: &Utf8Path, path: &Utf8Path) -> String {
 	let normalized_root = normalize_path(&package_root, None);
-	let normalized = normalize_path(&path, None);
-	if normalized.starts_with("..") {
-		panic!(
-			"File path \"{}\" is not within the package root \"{}\"",
-			path, package_root
-		);
+	let normalized_path = normalize_path(&path, None);
+	if normalized_path.starts_with("..") {
+		report_diagnostic(Diagnostic {
+			message: format!(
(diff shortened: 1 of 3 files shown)


packages/@winglang/wingc/src/type_check.rs
```diff
@@ -7496,27 +7496,58 @@ fn lookup_known_type(name: &'static str, env: &SymbolEnv) -> TypeRef {
...
+				"File path \"{}\" is not within the package root \"{}\"",
+				path, package_root
+			),
+			span: None,
+			annotations: vec![],
+			hints: vec![],
+			severity: DiagnosticSeverity::Error,
+		});
+		return package_name.to_string();
 	}
```
Chriscbr, line 7512: Not sure if changing these panics into diagnostics is right per se, but it should result in fewer LSP crashes

Issue #4106, opened 2023-09-06
[Feature Request] Having a typeof operator in Winglang 


harsh9607:
### Feature Spec

I was wondering if we could have a `typeof `Operator like we have in javascript.
We could have something like
```
let x  = 42 ;
log( typeof(x));
```
returns number;

 
 
 
 

### Use Cases

https://www.winglang.io/docs/language-reference

It is part of Roadmap 

![image](https://github.com/winglang/wing/assets/29411800/60ec321b-24c6-4c8a-a80c-c550c1b07316)

 Reflection: add an extended typeof operator to get type information.

### Implementation Notes

It could be built upon the typeof operator of Javascript. 

### Component

Language Design

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

staycoolcall911:
@MarkMcCulloh, @tsuf239, @hasanaburayyan, @Chriscbr - would you suggest implementing this similar to:
1. [util.sleep()](https://www.winglang.io/docs/standard-library/util/api-reference#@winglang/sdk.util.Util.sleep)
2. Our built-in [log()](https://www.winglang.io/docs/language-reference#12-utility-functions)
3. Any other suggestion on how to implement this?

ekeren:
I am not sure we want to implement this without a broader discussion of reflection and also typeof in javascript is not compatible with strongly typed languages... (where the typeof `bucket` could be `cloud.Bucket` or maybe some `IBucket` - thanks @Chriscbr )

hasanaburayyan:
Yea I think for sure its something that would be useful to have and we probably want it. However, its kinda complicated because of the way `typeof` works in javascript. Maybe we add this to an upcoming team time discussion? Im not sure how exactly implementation would work, somehow we need to map the javascript back to the Wing type.

staycoolcall911:
Makes sense - I'll add to a next team time so we can discuss it.

Chriscbr:
@ainvoner After re-reading this issue, if I understand it correctly this might be a separate feature from the `typeof` type operator that TypeScript introduces for use inside type annotations and type parameters (e.g. `someFunc<typeof otherFunc>()`). Their docs explains it better than I can https://www.typescriptlang.org/docs/handbook/2/typeof-types.html

PR #7148, opened 2024-09-19
rfc: preflight type reflection

Chriscbr:
Related to https://github.com/winglang/wing/issues/4106

Proposal for an extension to Winglang that adds support for type reflection. Examples and an implementation sketch is included.

## Checklist

- [ ] Tests added (always)

### DIFF
--- a/docs/api/05-language-reference.md
+++ b/docs/api/05-language-reference.md
@@ -1280,9 +1280,158 @@ let rawData: bytes = fs.readBytes("path/to/file");
 fs.writeBytes("path/to/file", rawData);
 ```
 
+#### Roadmap
+
+The following features are not yet implemented, but we are planning to add them in the future:
+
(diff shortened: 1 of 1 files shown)


eladb (approved):
Beautiful! No comments! Ship it!

Chriscbr:
@eladb An extra note...

There's really two kinds of reflection that we could imagine supporting.

In the examples I've shown in the RFC, a specific type is always given to the function. This is pretty powerful, as the examples show. It's less risky to implement since the `Type` value that's returned is known by the time the JavaScript is generated.

But one could imagine a more powerful reflection macro that lets you pass an arbitrary value/variable. (It could be a separate macro like `@typeOf(...)`, or it could be part of the same macro). I think this macro would be more tricky to implement since it could requires more cross-cutting changes to our code generation (say, by annotating preflight values with type info via hidden properties or something like that...), though it would be mighty cool.

PR #7151, opened 2024-09-20
feat: `@type` intrinsic

### LINKED ISSUE #7150: Implement type reflection intrinsics
### Use Case

`@type` as described in the language reference here: https://www.winglang.io/docs/api/language-reference#115-type-reflection

### Proposed Solution

_No response_

### Implementation Notes

_No response_

### Component

Compiler

### Community Notes

Chriscbr:
Closes #7150

Adds for the new `@type` intrinsic function. This function can be passed a type, and it will give you back an object (`std.reflect.Type`) with information about it -- for example, whether it's a struct or class, what properties or fields it has, etc. You can use this to validate request data structures at runtime, generate schemas for APIs or databases, or for simply asserting on the structure of your code.

Future work:
- Support doc reflection
- Fix support for variadic functions
- Open discussion about accessing private fields or methods via reflection
- Open issue about explicit "preflight" annotation
- Open issue about explicit `TypeId` enum

## Checklist

- [x] Tests added (always)


packages/@winglang/wingc/src/jsify.rs
```diff
@@ -2283,3 +2285,65 @@ fn escape_javascript_string(s: &str) -> String {
...
+				func_code.append(jsify_type_reflection(&p.type_annotation, &p.type_annotation.span));
+				func_code.append(", ");
+			}
+			func_code.append("], ");
+			func_code.append(jsify_type_reflection(&t.return_type, &t.return_type.span));
+			func_code.append(")");
+
+			new_code!(expr_span, "std.Type._ofFunction(", func_code, ")")
+		}
+		TypeAnnotationKind::UserDefined(_udt) => todo!(),
```
eladb, line 2347: TODO

packages/@winglang/tree-sitter-wing/grammar.js
```diff
@@ -41,7 +41,6 @@ module.exports = grammar({
 
     // These modifier conflicts should be solved through GLR parsing
     [$.field_modifiers, $.method_modifiers],
```
eladb, line 43: Todo: Add docs

eladb (approved):
Beautiful!

Chriscbr:
It looks like we need to be smart about how we emit these reflection classes, otherwise we'll run into code generation issues. For example, supposing you have structs that refer to each other like these:

```js
struct S1 {
  child: S2?;
}
struct S2 {
 child: S1?;
}
```

Then the naive way of generating meta-objects will emit something like this:

```js
new std.StructType("S1", {
  child: new std.OptionalType.of(
    new std.StructType("S2", {
      child: new std.StructType("S1", <infinity>)
    })
  )
});
```

The example above is contrived, but in practice there are many recursive relationships between classes in Wing libraries and the SDK via method parameters and return types etc.

Chriscbr:
The way I'm thinking of solving this is to "collect" the list of types referenced through helper functions in `jsify.rs`, assigning names to each of them, and then to emit somewhere around the beginning of preflight.js looking something like:

```js
// prelude: we instantiate classes for all types, without their contents filled
let $s1_type = new std.StructType("S1");
let $s2_type = new std.StructType("S2");

// after: emit code to fill in type information, with cycles allowed
$s1_type.fields = { child: new std.OptionalType($s2_type) };
$s2_type.fields = { child: new std.OptionalType($s1_type) };
```

This code that defines reflection data for named types (classes, structs, interfaces, enums) can go into some kind of prelude section of `preflight.js`. Or even in a separate file. Then, wherever `@type(MyStruct)` is actually called, it will just emit a reference to the variable `$s1_type`.

Chriscbr:
I've been making a little bit more progress. To do the approach above, we basically need to do some recursion on types. When we generate the javascript reflection code for S1, it needs to emit the reflection code for S2, and to emit that reflection code, we'd need to emit reflection code for S1. To avoid recursing indefinitely, we need to have a base case where if a type has already been explored (we've generated code for it), then we stop and reuse the existing result.

But we can't do that easily today because `Type` / `TypeRef` do not implement the Eq/Hash traits. Put another way, we have no way to tell if two types are equal. Part of the reason why it hasn't been easy to define equality on types is because we create duplicate type objects even for identical function signatures - there's no interning process going on. It's possible we can add a stub implementation of Eq/Hash that handles enough of the cases and works around this, but it might be worthwhile to see if we can change how we're storing types to avoid.

packages/@winglang/wingc/src/jsify.rs
```diff
@@ -2146,6 +2186,397 @@ impl<'a> JSifier<'a> {
...
+		if let Some((variable_name, _, _)) = type_reflection_definitions.get(&type_) {
+			return new_code!(expr_span, format!("$types.{variable_name}"));
+		}
+		std::mem::drop(type_reflection_definitions);
+
+		// Generate a unique name for referring to this type in the $types object
+		let mut type_variable_counter = self.type_variable_counter.borrow_mut();
+		*type_variable_counter += 1;
+		let type_variable_name = format!("t{}", type_variable_counter);
+		std::mem::drop(type_variable_counter);
```
Chriscbr, line 2291: Give the variables short friendly names before the unique number to make them more debuggable

Chriscbr:
Wow - this recursion problem really punches back!

So it seems like I fixed all of the issues with generating types that may have cyclic references in preflight. I've copied an example of a `types.cjs` file generated with my local build:

<details>

```
const std = require("@winglang/sdk").std;
const $types = {};
$types.t1 = std.Type._ofInterface(new std.InterfaceType("MyInterface", "examples-valid.MyInterface"));
$types.t2 = std.Type._ofInterface(new std.InterfaceType("BaseInterface", "examples-valid.BaseInterface"));
$types.t3 = std.Type._ofInterface(new std.InterfaceType("IResource", "@winglang/sdk.std.IResource"));
$types.t4 = std.Type._ofInterface(new std.InterfaceType("IConstruct", "constructs.IConstruct"));
$types.t5 = std.Type._ofInterface(new std.InterfaceType("IDependable", "constructs.IDependable"));
$types.t6 = std.Type._ofInterface(new std.InterfaceType("IHostedLiftable", "@winglang/sdk.std.IHostedLiftable"));
$types.t7 = std.Type._ofInterface(new std.InterfaceType("ILiftable", "@winglang/sdk.std.ILiftable"));
$types.t8 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t9 = std.Type._ofInterface(new std.InterfaceType("IInflightHost", "@winglang/sdk.std.IInflightHost"));
$types.t10 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t11 = std.Type._ofStr();
$types.t12 = std.Type._ofVoid();
$types.t13 = std.Type._ofArray(undefined, false);
$types.t14 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t15 = std.Type._ofClass(new std.ClassType("MyClass", undefined));
$types.t16 = std.Type._ofClass(new std.ClassType("Resource", "@winglang/sdk.std.Resource"));
$types.t17 = std.Type._ofClass(new std.ClassType("Construct", "constructs.Construct"));
$types.t18 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t19 = std.Type._ofAny();
$types.t20 = std.Type._ofBool();
$types.t21 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t22 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t23 = std.Type._ofArray(undefined, false);
$types.t24 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t25 = std.Type._ofArray(undefined, false);
$types.t26 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t27 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t28 = std.Type._ofFunction(new std.FunctionType(std.Phase.PREFLIGHT));
$types.t29 = std.Type._ofEnum(new std.EnumType("MyEnum", "examples-valid.MyEnum", {"VARIANT1": new std.EnumVariant("VARIANT1"), "VARIANT2": new std.EnumVariant("VARIANT2"), }));;
$types.t30 = std.Type._ofStruct(new std.StructType("MyStruct", "examples-valid.MyStruct"));
$types.t31 = std.Type._ofStruct(new std.StructType("Base1", "examples-valid.Base1"));
$types.t32 = std.Type._ofStruct(new std.StructType("Base2", "examples-valid.Base2"));
$types.t33 = std.Type._ofNum();
$types.t34 = std.Type._ofArray(undefined, false);
$types.t35 = std.Type._ofOptional(undefined);
$types.t36 = std.Type._ofJson();
$types.t1.data.bases.push($types.t2.data);
$types.t1.data.methods["method1"] = new std.Method("method1", false, $types.t14.data);
$types.t2.data.bases.push($types.t3.data);
$types.t3.data.bases.push($types.t4.data);
$types.t3.data.bases.push($types.t6.data);
$types.t3.data.methods["onLift"] = new std.Method("onLift", false, $types.t8.data);
$types.t4.data.bases.push($types.t5.data);
$types.t6.data.bases.push($types.t7.data);
$types.t6.data.methods["onLift"] = new std.Method("onLift", false, $types.t8.data);
$types.t8.data.params.push($types.t9.data);
$types.t8.data.params.push($types.t13.data);
$types.t8.data.returns = $types.t12;
$types.t9.data.bases.push($types.t3.data);
$types.t9.data.methods["addEnvironment"] = new std.Method("addEnvironment", false, $types.t10.data);
$types.t10.data.params.push($types.t11.data);
$types.t10.data.params.push($types.t11.data);
$types.t10.data.returns = $types.t12;
$types.t13.data.child = $types.t11;
$types.t14.data.returns = $types.t12;
$types.t15.data.base = $types.t16.data;
$types.t15.data.interfaces.push($types.t1.data);
$types.t15.data.properties["field1"] = new std.Property("field1", $types.t11);
$types.t15.data.methods["method1"] = new std.Method("method1", false, $types.t27.data);
$types.t15.data.methods["method2"] = new std.Method("method2", true, $types.t28.data);
$types.t15.data.methods["onLift"] = new std.Method("onLift", false, $types.t22.data);
$types.t15.data.methods["onLiftType"] = new std.Method("onLiftType", true, $types.t24.data);
$types.t15.data.methods["toInflight"] = new std.Method("toInflight", true, $types.t26.data);
$types.t15.data.methods["isConstruct"] = new std.Method("isConstruct", true, $types.t18.data);
$types.t15.data.methods["toString"] = new std.Method("toString", false, $types.t21.data);
$types.t16.data.base = $types.t17.data;
$types.t16.data.interfaces.push($types.t3.data);
$types.t16.data.methods["onLift"] = new std.Method("onLift", false, $types.t22.data);
$types.t16.data.methods["onLiftType"] = new std.Method("onLiftType", true, $types.t24.data);
$types.t16.data.methods["toInflight"] = new std.Method("toInflight", true, $types.t26.data);
$types.t16.data.methods["isConstruct"] = new std.Method("isConstruct", true, $types.t18.data);
$types.t16.data.methods["toString"] = new std.Method("toString", false, $types.t21.data);
$types.t17.data.interfaces.push($types.t4.data);
$types.t17.data.methods["isConstruct"] = new std.Method("isConstruct", true, $types.t18.data);
$types.t17.data.methods["toString"] = new std.Method("toString", false, $types.t21.data);
$types.t18.data.params.push($types.t19.data);
$types.t18.data.returns = $types.t20;
$types.t21.data.returns = $types.t11;
$types.t22.data.params.push($types.t9.data);
$types.t22.data.params.push($types.t23.data);
$types.t22.data.returns = $types.t12;
$types.t23.data.child = $types.t11;
$types.t24.data.params.push($types.t9.data);
$types.t24.data.params.push($types.t25.data);
$types.t24.data.returns = $types.t12;
$types.t25.data.child = $types.t11;
$types.t26.data.params.push($types.t3.data);
$types.t26.data.returns = $types.t11;
$types.t27.data.returns = $types.t12;
$types.t28.data.returns = $types.t12;
$types.t30.data.bases.push($types.t31.data);
$types.t30.data.bases.push($types.t32.data);
$types.t30.data.fields["base1"] = new std.Property("base1", $types.t20);
$types.t30.data.fields["base2"] = new std.Property("base2", $types.t20);
$types.t30.data.fields["field1"] = new std.Property("field1", $types.t33);
$types.t30.data.fields["field2"] = new std.Property("field2", $types.t34);
$types.t31.data.fields["base1"] = new std.Property("base1", $types.t20);
$types.t32.data.fields["base2"] = new std.Property("base2", $types.t20);
$types.t34.data.child = $types.t35;
$types.t35.data.child = $types.t36;
module.exports = $types;
```

</details>

All of this works fine. We define the types first, then fill in the types' contents.

But it appears that separately, there's a problem with lifting `std.Type` instances from preflight to inflight. Namely, calling `toInflight()` sometimes blows the stack frame.

Let's look back at the example before:

```
struct S1 {
  child: S2?;
}

struct S2 {
 child: S1?;
}
```

And suppose we lift a type reflection:

```
let s1 = @type(S1);

inflight () => {
  log(s1.kind); // should print "struct"
}
```

In preflight, `s1` is an instance of `std.Type` storing inside a `std.StructType`. To use it inflight, we have to generate some JS code that will get included in the inflight function, through `toInflight()`. What does the `toInflight()` implementation look like? The current naive implementation looks approximately like this:

```js
  /** @internal */
    const args = [
      `"${this.name}"`,
      this.fqn ? `"${this.fqn}"` : "undefined",
      arrayToInflight(this.bases), // helper function that calls _toInflight() on each parent struct reflection
      mapToInflight(this.fields), // helper function that calls _toInflight() on each field reflection
    ];
    return `new std.StructType(${args.join(", ")})`;
  }
```

But as you can imagine, this means `s1._toInflight()` will end up calling `s2._toInflight()` which will then call `s1._toInflight()` and so on.

The solution isn't as obvious to me. One idea could be to do some kind of runtime tree traversal -- so we can find all unique types -- then define some variables for them, then call an altered method like `toInflightWithContext()` on each type so that it can refer to this map of variables (stopping the endless recursion) when it sees another type.

Another idea could be to totally revise how we're code generating these helper classes. But I'm not sure where I'd start with that to be honest.

packages/@winglang/sdk/src/std/type.ts
```diff
@@ -0,0 +1,641 @@
...
+
+  /** @internal */
+  public static _ofFunction(f: FunctionType): Type {
+    return new Type("function", f);
+  }
+
+  /**
+   * What kind of type this is.
+   */
+  public readonly kind: string;
```
Chriscbr, line 127: Make this an enum

```suggestion
  public readonly kind: TypeId;
```

eladb:
> Another idea could be to totally revise how we're code generating these helper classes. But I'm not sure where I'd start with that to be honest.

A sketch of a thought: instead of only generating types that are accessed, perhaps we could generate a single in-memory data structure that includes the *entire* app type system. Basically a big-ass JSON dump of the application's type info that the compiler holds in-memory. Then `@type(x)` will just access this at runtime (preflight-time).

So this:

```js
class Foo {}

let x = @type(Foo);
```

Becomes this:

```js
const $types = require("./types.json");

const x = std.Type._of($types["fqn.of.Foo"]);
```

MarkMcCulloh:
> instead of only generating types that are accessed, perhaps we could generate a single in-memory data structure that includes the entire app type system.

Depending on how this is done, this could be a huge performance hit with JSII. Unless the original jsii manifest could be reused somehow.

> But it appears that separately, there's a problem with lifting std.Type instances from preflight to inflight. Namely, calling toInflight() sometimes blows the stack frame.
> But as you can imagine, this means s1._toInflight() will end up calling s2._toInflight() which will then call s1._toInflight() and so on.

@Chriscbr Maybe this is too far, but what if the `std.Type` API were changed so that all access to child type information is done through a preflight method instead of a property? iirc this is how the Type class in C# reflection works https://learn.microsoft.com/en-us/dotnet/api/system.type?view=net-8.0 (perhaps due to similar recursion problems?)

There's a hit to ergonomics, but I think it avoids the eager/cyclic lifting

```wing
struct S1 {
  child: S2?;
}

struct S2 {
 child: S1?;
}

let s1 = @type(S1);
let s2 = s1.getField("child");
let allFields = s1.getFields();

inflight () => {
  log(s1.kind); // should print "struct"
  log(s2.kind); // should print "struct"
}
```

packages/@winglang/sdk/src/std/type.ts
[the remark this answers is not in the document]
Chriscbr: Done

packages/@winglang/sdk/src/std/type.ts
Chriscbr: Might leave this as a future improvement for now

Issue #6465, opened 2024-05-13
Cloud deployment fails if root directory contains spaces


polamoros:
### I tried this:

create an empty directory with a name that includes spaces `my directory` 

`mkdir "my directory"`

Create a wing test file:

```
//my directory/main.test.w

bring cloud;

let bucket = new cloud.Bucket();

test "Bucket is empty" {
  assert(bucket.list().length == 0);
}
```

and run `wing test -t tf-aws main.test.w` 

### This happened:

```
✔ Compiling main.test.w to tf-aws...

✔ terraform init

✖ terraform apply

Command failed: terraform apply -auto-approve
╷
│ Error: opening S3 object source (assets/Test78LWzuS1GQ_env0_testBucketisempty_Handler_Asset_9A744A14/966A2F55D4BD8FD662E1A92D8BB84CD9/archive.zip): open assets/Test78LWzuS1GQ_env0_testBucketisempty_Handler_Asset_9A744A14/966A2F55D4BD8FD662E1A92D8BB84CD9/archive.zip: no such file or directory
│
│   with aws_s3_object.Test78LWzuS1GQ_env0_testBucketisempty_Handler_S3Object_9FB2AC36,
│   on main.tf.json line 145, in resource.aws_s3_object.Test78LWzuS1GQ_env0_testBucketisempty_Handler_S3Object_9FB2AC36:
│  145:       }
│
╵

✔ terraform destroy


Tests 1 failed (1)
Snapshots 1 skipped
Test Files 1 failed (1)
Duration 0m23.74s
```

### I expected this:

The test to pass.

### Is there a workaround?

no

### Anything else?

The test passes on the simulator

```
>_ wing test main.test.w

pass ─ main.test.wsim » root/env0/test:Bucket is empty

Tests 1 passed (1)
Snapshots 1 skipped
Test Files 1 passed (1)
Duration 0m0.30s
```

### Wing Version

0.73.41

### Node.js Version

v20.11.1

### Platform(s)

MacOS

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

skyrpex:
It worked to me:

```
wing test -t tf-aws main.test.w


✔ Compiling main.test.w to tf-aws...

✔ terraform init

✔ terraform apply

✔ Setting up test runner...

✔ Running tests...

[...]

PR #7188, opened 2024-10-01
fix(sdk): cloud deployment fails if root directory contains spaces

tsuf239:
fixes: #6465 

terraformAsset can't handle file paths that contain spaces- so I created the bundle zip manually- ONLY if there's a spaced path- since it hurts the benchmarks a bit

## Checklist

- [ ] Tests added (always)


tsuf239:
sdk test run: https://github.com/winglang/wing/actions/runs/11124550295

tsuf239:
another workflow run after a small change: https://github.com/winglang/wing/actions/runs/11127000644

packages/@winglang/sdk/src/target-tf-aws/function.ts
```diff
@@ -284,15 +287,29 @@ export class Function extends cloud.Function implements IAwsFunction {
...
+        path: bundle.directory,
+        type: AssetType.ARCHIVE,
+      });
+
+      this.assetPath = asset.path;
+    }
 
     this.bundleHash = bundle.hash;
-    this.assetPath = asset.path;
   }
```
Chriscbr, line 313: Is there a reason we need to still use TerraformAsset? Can we bundle the zip manually in all cases?

Chriscbr (approved):

packages/@winglang/sdk/src/target-tf-aws/function.ts
tsuf239: The manual zip makes the process longer in more than 25%- so the benchmark tests are failing (this is why I split it), I think we can consider remove it in the future

